PR #772, opened 2012-07-02
Automatically set icon size to default image size if not specified

dbmikus:
If no Icon size is provided, sets default Icon size to size of image loaded in.

Default Icon size is no longer a fixed dimension. If user does not provide a
default size, it will set the Icon size to the actual image size upon icon load.
Also, automatically configures iconSize, shadowSize, iconAnchor, and iconPopup
if values are not provided.

Fixed Icon image margins to actually place the middle bottom of the icon at the
specified map location.

Signed-off-by: Dylan Mikus <dbmikus@gmail.com>

- - -
I resubmitted this commit with the changes squashed into one and with the removal of auto-configuration of endline stuff. This change only changes the ability to autosize default icon sizes so images aren't squashed by default.


IvanSanchez:
Closing as incredibly old (0.4, srsly?!)

Issue #942, opened 2012-08-26
Adding layers with different CRS


kengu:
Hi, i've read [issue #693](https://github.com/CloudMade/Leaflet/issues/693) regarding support for adding layers with different CRS. Although it might not be a common use case, it does have it's uses. Mashups is one example, where users "shop" base layers from a host of different sources which might not all support a common CRS. Rescue services often depend on a host of different map sources (ranging from GIS-specialist data to public available services), which are combined (mashup) to give a better overview of the situation (my use case). Other frameworks does support this, but lack the excellent support for mobile devices and the lightweight characteristics of Leaflet. I do understand that this present some problems, but since project and reproject is supported, I believe it should be possible to move from single CRS to multiple CRS definitions.

I've cloned Leaflet and will look into this my self. If I succeed, I'll send a pull request for my solution.

PR #945, opened 2012-08-26
Added support for different CRS for each L.TileLayer.WMS instance

kengu:
This is a partial implementation of issue #942 which only affects L.TileLayer.WMS.

It turned out to be pretty easy to add support for different CRS for each L.TileLayer.WMS instance. I just had to add a small reproject method to L.Map which default to old algorithm in L.TileLayer.WMS.getTileUrl when L.TileLayer.WMS.option.crs (L.CRS instance) is not set (default: undefined). When set, L.Map.reproject uses this to reproject given point and zoom (L.Point) from L.Map.options.crs to given L.CRS (in this case L.TileLayer.WMS.options.crs is set, otherwise L.Map.options.crs as before).

_Example usage_

``` javascript
var map = L.map('map'),
    url = ... ,
    options = { ... , crs: L.CRS.EPSG4326};
L.tileLayer.wms(url, options).addTo(map);
```

which force `EPSG:4326` to be used in WMS requests instead of `EPSG:3857` which `map` now by default uses.

### DIFF
--- a/src/layer/tile/TileLayer.WMS.js
+++ b/src/layer/tile/TileLayer.WMS.js
@@ -35,24 +35,30 @@ L.TileLayer.WMS = L.TileLayer.extend({
 	},
 
 	onAdd: function (map) {
-
+		
+		// Get CRS from layer options is set, use CRS from map otherwise
+		var crs = (typeof this.options.crs === 'undefined' ? map.options.crs : this.options.crs);
+		
+		// This might contain a L.CRS, which should be removed (will not be overwritten is key is 'srs')
+		delete this.wmsParams.crs;
+		
 		var projectionKey = parseFloat(this.wmsParams.version) >= 1.3 ? 'crs' : 'srs';
-		this.wmsParams[projectionKey] = map.options.crs.code;
-
+		this.wmsParams[projectionKey] = crs.code;
+		
 		L.TileLayer.prototype.onAdd.call(this, map);
 	},
 
 	getTileUrl: function (tilePoint, zoom) { // (Point, Number) -> String
 
 		var map = this._map,
-			crs = map.options.crs,
+			crs = this.options.crs,
 			tileSize = this.options.tileSize,
 
 			nwPoint = tilePoint.multiplyBy(tileSize),
 			sePoint = nwPoint.add(new L.Point(tileSize, tileSize)),
 
-			nw = crs.project(map.unproject(nwPoint, zoom)),
-			se = crs.project(map.unproject(sePoint, zoom)),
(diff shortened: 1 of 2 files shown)


arctelix:
@mourner I'm a little confused on how to properly implement a TileLayer.WMS with a crs that differs from the map's crs.  Currently i do have it working but, only with the addition of @mhotby 's modification.  Without that, it seems impossible to implement.  Was this not the point of your integration of the `crs` option in the `TileLayer.WMS`?

For example If i define two TileLayers as follows:

1) Open Stree Maps with EPSG:3857

    osm = new L.TileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png')

2) and a WMS with EPSG:2263

     nycCrs = new L.Proj.CRS("EPSG:2263" etc..)
     nycMap = L.tileLayer.wms('http://maps.nyc.gov/geowebcache/service/wms/', {
         crs:nycCrs,
         tileSize: 512,
         etc...}

Define a map with the default, osm compatible crs of EPSG3857:

    var map = L.map('map', {
        layers:[osm]}

    var baseMaps = {
        "nyc": nycMap,
        "osm": osm};

    L.control.layers(baseMaps).addTo(map);

With the configuration above all requests to the nycMap WMS server fail due to incorrect scale and or origin issues.

However If i use the `setCrs()` method and and 'baselayerchange' event listener suggested by mhotby (modified for my setup):

    L.Map.prototype.setCrs = function(newCrs) {
        this.options.crs = newCrs;
    }

    map.on('baselayerchange', function(layer) {
        if(layer.name == 'osm')
            map.setCrs(L.CRS.EPSG3857);
        else if(layer.name == 'nyc')
            map.setCrs(nycCrs);
        map.setView(map.getCenter(),map.getZoom());
    });

Now everything works as expected.  I'm just wondering if this is what you intended or if i have missed something in my configuration.  Also, if this is the correct way to implement it , why are mhotby 's mods not implemented along with the TileLayer.WMS.options.crs.  It seems to be incomplete without them...

Thanks for everything, leaflet is nothing short of awesome!

perliedman:
@arctelix are you trying to mix more than one CRS on the same map? That has been discussed a couple of times, for example see #2553.

My opinion on this is that CRS switching should not be supported by Leaflet core, since I can't see it working in an intuitive way. It could, however, probably quite easily be be supported by a plugin. So far, no one have written such a plugin, as far as I know.

I was actually a bit surprised to learn about the WMS functionality added by this PR, since it seems to add at least partial support for multiple CRS!

arctelix:
Thanks!

PR #1232, opened 2012-12-23
POC of a i18n system

yohanboniface:
Hi!

While working on a i18n feature for [Leaflet.Storage](https://github.com/yohanboniface/Leaflet.Storage), I though it should be more convenient to have all Leaflet plugins share the same system. That's why a suggest to add it in Leaflet core instead of making a plugin. Making a plugin (say Leaflet.I18n) means making it a dependency for all the other plugins, so I think it will not work.

So here is a suggestion.

The API is:
- to make a string translatable, just call `L._` (shortcut for `L.i18n`), like this:

```
L._("The phrase I want to be translatable");
L._("A complex sentence with a {variable}", {variable: "my value"});
```
- one or more `locale/{locale_code}.js` should be provided, something like:

```
var mylocale = {
    "Simple phrase to translate": "Une simple phrase à traduire",
    "A sentence with a {variable}": "Une phrase avec une {variable}",
};
```
- the locale must register itself, doing something like:

```
L.registerLocale('fr', mylocale);
```

 This can be done more than once. For example, every plugin can add it's translations for some locale without overriding all the already registered strings. However, if the same string for the same locale is registered twice, the last wins.
- end user must load the `locale/` files provided by Leaflet, or plugins
- end user must set his locale, doing so:

```
L.setLocale('fr');
```

And then all the translatable strings available will be translated :).

Of course, i18n is optional: unless one locale has been set, the behaviour is unchanged.

Insolved:
- Not sure of the better way to make the locale available in the `dist/` module. I just think that each end user must only load the locales he wants

Of course, some documentation of this will be needed, if merged.

Thanks for reviewing!

Yohan

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -5,6 +5,7 @@ var deps = {
 		      'core/Class.js',
 		      'core/Events.js',
 		      'core/Browser.js',
+		      'core/I18n.js',
 		      'geometry/Point.js',
 		      'geometry/Bounds.js',
 		      'geometry/Transformation.js',
--- a/debug/leaflet-include.js
+++ b/debug/leaflet-include.js
@@ -7,6 +7,7 @@
 		'core/Class.js',
 		'core/Events.js',
 		'core/Browser.js',
+		'core/I18n.js',
 
 		'geometry/Point.js',
 		'geometry/Bounds.js',
--- a/spec/runner.html
+++ b/spec/runner.html
@@ -29,6 +29,7 @@
 	<script type="text/javascript" src="suites/core/UtilSpec.js"></script>
 	<script type="text/javascript" src="suites/core/ClassSpec.js"></script>
 	<script type="text/javascript" src="suites/core/EventsSpec.js"></script>
+	<script type="text/javascript" src="suites/core/I18nSpec.js"></script>
 
 	<!-- /geometry -->
 	<script type="text/javascript" src="suites/geometry/PointSpec.js"></script>
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -1,0 +1,33 @@
+describe("I18n", function() {
+
+    beforeEach(function() {
+        var fr = {
+            "Simple phrase to translate": "Une simple phrase à traduire",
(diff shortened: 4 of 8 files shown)


rastapopougros:
So, many years later, it seems @yohanboniface was right… no trans system directly inside core (and promotion of it) = no translations nowhere.

Zoom in, Zoom out always only in english, no arabic, chinese, french, spanish…
Same for "Full Screen" which is a very common plugin. And so on…

Sometimes we can change strings manually via options, but it means every single software that uses Leaflet, has to re-implement i18n and above all *have its own list of translations*, instead of mutual locales files.

I'm just a user/integrator, so to quote @yohanboniface : sad for humans :(
(and +1 for his request, still relevant)

tmcw:
> above all have its own list of translations, instead of mutual locales files.

I'm not seeing how this could possibly be any different. What's the alternative to plugins managing their own locale files - a central repository? Would that central repository contain all hundreds of Leaflet plugins, including the dead ones? Would it have a point release every time any of those plugins updated a source string?

You can read this situation as either

- Leaflet made the wrong choice, if it had adopted a translation system, plugins would have adopted it
- Nobody in the plugin ecosystem has stepped up and promoted a translation system, or set an example by translating their plugin

I find it easier to believe the latter.

rastapopougros:
> What's the alternative to plugins managing their own locale files

It's a misunderstanding. I talked about other software which include and use Leaflet, absolutely not about plugins.

It's precisely inside each Leaflet plugins, using a common API provided by the core, that I would love to have translations (and inside the core package for "Zoom in" and "Zoom out").

For the moment, those other softwares or all projects (website etc), they all have to have their own translations for each core leaflets string + plugins strings, duplicated in hundreds, or thousands. Instead of having those translations in leaflet core, and in leaflet plugins, only once.

Concretely, each project with non english humans have to translate "Zoom out", "Full Screen", etc, in arabic, french, chinese… and stores and maintains those translations only inside that project, and overload strings for core and plugins to incorporate them.

I was talking about
- having arabic or spanish translation of "Zoom out" only once in the whole community (=> in the core)
- and having arabic or spanish translation of "Full Screen" only once in the whole community (=> in this plugin, and same for all other plugins)
And not in each project who uses them. :(

Issue #82, opened 2011-05-30
Problem with polylines/polygons over the date line


ghost:
Hi guys, when a polyline straddles the W180 line of longitude, crossing from one side to the other, the polyline does not continue on the other side of the W180 line, but is drawn back on the other side of the "current iteration" of the world.

Here is an example, polyline has 3 points draw from left to right. The first two polypoints are drawn on the left side of the W180 line, but the third which is meant to be just past the W180 line is not drawn to the right of the W180 line. It is being drawn back to the left where the current world begins:

http://fswidgets.com/beta/leaflet_test.html

JoeMatulka:
I very much enjoy using Leaflet. I was wondering if there is any other work done on this issue other than the work around.

Keep up the good work! :smile:

wmleler:
It was a while ago so I’m not sure, but I think they fixed this issue.  —wm

RhezaBoge:
i'm sorry for my english.

there is a function that i found on leaflet 0.7.7  ("wrap" Conversion Methods). it has made me solved this problem with some additional code. 

but i can't run this function again on the newest leaflet (currently 1.0.1). i don't know why.. so for a while, i just use leaflet 1.0.1 but i change little bit code on "wrap function" (use leaflet 0.7.7 code). let me know if I missed anything or had some incorrect code for "wrap" methods on the newest leaflet. or maybe "wrap" methods has been removed or changed?

callumacrae:
<img width="1263" alt="screen shot 2016-10-17 at 14 22 43" src="https://cloud.githubusercontent.com/assets/472830/19440148/a8b8a34c-9478-11e6-9738-079f77bd045d.png">

I'm getting this with leaflet 1.0.1. Looks similar to previous screenshots of this issue: https://github.com/mbostock/topojson/issues/242

chrisregner:
I'm sorry, but I'm confused. @mourner said this:

> I removed LatLng wrapping/clamping by default alltogether (there's now an explicit LatLng wrap method for this). So all similar problems should be solved now in theory.

What I understood from this that I shouldn't need any work around. But I still experience the issue:
 
![image](https://user-images.githubusercontent.com/26005158/30159980-1d86b838-93fd-11e7-8b5f-069f0f772583.png)

So I tried doing the workaround he suggested:

>...the current workaround is to change the definition of points that are not drawing properly e.g. from this:
> `new L.LatLng(60, -150)`
> to this:
[...]

Issue #33, opened 2011-04-13
Buggy zoom animation on Android (2.2, Nexus One)


mourner:
Android animates transform in quite a weird way... Needs checking out.

PR #1293, opened 2013-01-18
Proposal for solving "crossing date line geometries" problem

### LINKED ISSUE #2645: Polygon Points across the dateline
I am pretty sure this issue would have been brought up previously, but I have looked and looked and I can't seem to see the issue resolved, or find a direct thread about the issue.

The Bug itself is when the polygons go across the ITD or International Date Line
First of all, the things I have tried to stop the issue:
1) using the continuous world flag.
2)directly going above 180 degrees

All the above results in the map looking like this: 
screenshot - 230414 - 16 23 15

I am using rhumb lines to calculate the polygon so it will be impossible for me to directly go above 180.

yohanboniface:
**Warning: This is full R&D, POC, not mergeable at all in this state!**

This follows the [discussion we had](https://github.com/Leaflet/Leaflet/commit/f5cfab73a19e8976289b85cce8a6e04a66fe6c00#commitcomment-2386191) about longitude wrapping and geometries [crossing date line](https://github.com/Leaflet/Leaflet/issues/82). With the 0.5 release (\o/) I've started thinking about this again.

As you already know, **_my**_ opinion is that removing wrapping from `L.LatLng` was a good move, but removing it from default Leaflet behaviour was not such a good one.

My main points are:
- this breaks consistency with other GIS, like PostGIS (which doesn't accept longitudes over 180 when in  geography mode)
- this doesn't totally solve the "cross date line" problem

I mean, for example, just trying to display a polygon around New Zealand will not give expected results. Say:

```
        L.polygon([
            [-40, 176],
            [-45, 178],
            [-46, -178],
            [-40, -175],
        ]).addTo(map);
```

This will result in a huge polygon crossing the whole world, instead of a small one crossing the date line.

So thinking about this, I went to this point:
- Leaflet **must** use acute coordinates internally, and **must** be able to expose them to third party GIS
- Leaflet _should_ try to **display** things as close as possible as how _humans_ expect them

In others words: _data_ is for machines, _displaying_ is for humans.

So I thought about some heuristical algorithm to determine where some ambiguous point has to be displayed.

This algorithm is very simple, but in the same time quite intrusive, and actually I don't fully like my current implementation, but here is what it looks like:
- when projecting a point, try to give a context to disambiguate
- this context is by default the map center
- and when displaying a path, use previous point of the path to contextualize the next one

And than you can successfully display the above Polygon. :)

As previously said, the actual implementation lacks of some refactoring, it's just a proof-of-concept. The aim is to continue the discussion, and not just complaining, also making some proposals ;)

So, when you have some time, I'd like your opinion on this proposal. Hence, if you think some point merit to dig in and not just everything has to be trashed, I can work on it to reach a satisfactory state.

fixes #2645


MuellerMatthew:
@mikeatlas Looking at the above dateline example you provided from Oct 24th 2014, it seems that the polyline for the top two points on the map should actually go over the top of the map when crossing the dateline, and not straight across which is a geographically longer route.  

I think the solution to this needs to include both the max 180degree rule, but ideally it should also have a check when drawing lines on whether it is shorter to cross over a pole.  Whenever I fly between the US and China I have gone over the north pole because that is the shortest route between NYC and Beijing. Using your solution if i plotted both points on the map and drew a line, then the route would appear to go horizontally across the map instead of up and over.

mikeatlas:
@MuellerMatthew I'm certainly no expert, my solution is a "visual" workable hack, never claimed it was truly correct or would work for non web-mercator projections. I imagine in the future with the arctic circle melting, plus the rising stakes/claims between russia and canada over "northern routes" - it wont just be airplane flights that need better north-pole projection handling- imagine -180 to 180 at 85lat? non web mercator -a circle, of course! 

D3.js has some great support for non-web-mercator projections, but I think leaflet's biggest strengths are 

-free
-easy to use for JS beginners
-easy to make 1-off relatively static maps with low #s of objects

Even Leaftlet 1.0 doesn't seem to have any IDL dateline solution integrated, nor have I ever seen it used for anything other than WGS84 (I don't have a GIS background either) projection. 

Perhaps we'll see something change in Leaflet 2.0 in a few years, I expect WebGL to be far more prevalent by then, and possibly many libraries like Leaflet may chose to re-implement or take a new approach to rendering web maps with widespread WebGL available. I've already seen a few non-free WebGL-only javascript attempts at Google-Earth like platforms (alpha status).

mikeatlas:
Also @mourner and @yohanboniface if you're considering re-approaching this (`mourner added the needs-discussion label on May 8`) and (`yohanboniface referenced this pull request in umap-project/umap on Aug 2 POIs should be displayed outside the center of the world map #33`)

I'd like to add that non-web-mercator projection may take on increasingly important roles in web-based mapping (along with the availability and support of WebGL) specifically for reasons mentioned above (the north passage becoming a reality; cross-polar flights already happen daily...)

linuskamb:
Is there any hope that this will be fixed some day? I have been hoping to use Leaflet for a tsunami warning information app for the Pacific and National Tsunami Warning Centers to distribute warning information to emergency management personnel, but, given that tsunamis are largely a Pacific across-the-dateline phenomenon, this is a show-stopper for me.  I use datasources that are not under my control, and I cannot possibly hack all the longitude values.

ValentinH:
Also +1

landswellsong:
Subscribed and went to write an ugly hack

dwy904:
Has this issue been fixed so far ?

clawfrank:
Aside from the four thousand commits between this branch and master, is performance still the blocker here? @lemon-tree 's suggestion to make this an option seems like a good compromise.

> 
> 
> Has any progress been made on this at all recently?
> 
> From what I can gather, the performance concerns have stalled this feature, but could it not be implemented as a option flag that is off by default?

PR #1309, opened 2013-01-24
Support more granularity in zoom levels

lapo-luchini:
Support more granularity in zoom levels (e.g. zoomStep=0.5) that can be used, with providers that produce those levels, to have smaller changes in zoom level after a pinch operation, or a better fit with fit-to-bounds operations.


lapo-luchini:
Change merged from #2382 is much more comprehensive, but it lacks the ability to support tile layers with support for fractional zoom.
I'll try and do a new PR for that.

PR #1430, opened 2013-02-19
Allow classes to be constructed without `new`

jfirebaugh:
This uses the "instanceof trick" from [John Resig](http://ejohn.org/blog/simple-class-instantiation/) to ensure that
`new Foo(…)` and `Foo(…)` both return a newly constructed instance.

This eliminates the need for separate camel-case functions that wrap the
constructor (e.g. `L.map`). The subtleties of that approach are a source
of confusion and frustration for new users.

I suggest that the documentation be changed to exclusively reference a
`new L.Map()`/`L.Map()` style. Obviously we'll have to keep the camel-case
variants around for backward compatibility at least until a major version
release.

This is not quite ready to merge: several specs are failing. The issue
is that there's no way to make e.g. `new L.LatLng(undefined)` behave like
`L.latLng(undefined)` (returning undefined). Is this an important behavior?
For example, what's the rationale for 84d537b8046fd6ee397d20d8f5154d9812b63502?

### DIFF
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -2,31 +2,40 @@ describe("Class", function() {
 
 	describe("#extend", function() {
 		var Klass,
-			constructor,
+			initializer,
 			method;
 
 		beforeEach(function() {
-			constructor = jasmine.createSpy("Klass constructor");
+			initializer = jasmine.createSpy("Klass initializer");
 			method = jasmine.createSpy("Klass#bar method");
 
 			Klass = L.Class.extend({
 				statics: {bla: 1},
 				includes: {mixin: true},
 
-				initialize: constructor,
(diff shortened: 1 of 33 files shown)


akauppi:
To @mourner (Mar 2013): 

>docs/tutorials/examples to reflect only one way to do things (and that would be using factories without new anywhere)

This would indeed be helpful. I adopted that style, but have consistently wondered why much of the code uses `new` and capitals, instead. Like if there's a performance advantage or something. Being consistent about this would indeed cut the newbie confusion.

(wow, I missed the issue's 2 years birthday 🍰🍰)

mourner:
We're not exposing the `new Class` way of creating things in the docs now, so I think we can safely punt on this. Great to have this PR in the archive though, in case we change our minds :)

Issue #712, opened 2012-05-29
toGeoJSON


jfgirard:
Hi, 
I'm working with L.GeoJSON and the editing tools.  I didn't find a built-in way to extract the GeoJSON from a layer when it's time to save back to server.

Here is what I did. The target variable is from the click event (e.target):

```
toGeoJSON: function(target) {
    if (target instanceof L.Marker) {
        //Point
        return {
            coordinates: this.latLngToCoords(target.getLatLng()),
            type: 'Point'
        }
    } else if (target instanceof L.MultiPolygon || target instanceof L.MultiPolyline) {
        //MultiPolygon and MultiLineString
        var multi = [];
        var layers = target._layers;
        for (var stamp in layers) {
            multi.push(this.toGeoJSON(layers[stamp]).coordinates);
        }
        return {
            coordinates: multi,
            type: (target instanceof L.MultiPolygon) ? 'MultiPolygon': 'MultiLineString'
        };
    } else if (target instanceof L.Polygon) {
        //Polygon
        var coords = this.latLngsToCoords(target.getLatLngs());
        return {
            coordinates: [coords],
            type: 'Polygon'
        };
    } else if (target instanceof L.Polyline) {
        //Linestring
        var coords = this.latLngsToCoords(target.getLatLngs());
        return {
            coordinates: coords,
            type: 'LineString'
        };
    } else if (target instanceof L.FeatureGroup) {
        //Multi point and GeometryCollection
        var multi = [];
        var layers = target._layers;
        var points = true;
        for (var stamp in layers) {
            var json = this.toGeoJSON(layers[stamp]);
            multi.push(json);
            if (json.type !== 'Point') {
                points = false;
            }
        }
        if (points) {
            var coords = multi.map(function(geo){
                return geo.coordinates;
            });
            return {
                coordinates: coords,
                type: 'MultiPoint'
            };
        } else {
            return {
                geometries: multi,
                type: 'GeometryCollection'
            };
        }
    }
},

latLngToCoords: function(latlng) {
    return [latlng.lng, latlng.lat];
},

latLngsToCoords: function(arrLatlng) {
    var coords = [];
    arrLatlng.forEach(function(latlng) {
        coords.push(this.latLngToCoords(latlng));
    },
    this);
    return coords;
}
```

I don't know where to include such a function in the leaflet framework...  But I think it would be very useful working with edit tools.

Jeff

PR #1462, opened 2013-02-26
Add #toGeoJSON to various layer types (#712)

jfirebaugh:
Two shortcomings of this approach:
1. Features and FeatureCollections aren't round tripped. `L.GeoJSON.geometryToLayer`
   discards feature properties and I'm not seeing a clear way to keep those around for
   reserialization.
2. MultiPoints aren't round tripped. `L.GeoJSON.geometryToLayer` creates a FeatureGroup
   of Markers when it encounters a MultiPoint, but my implementation of
   `FeatureGroup#toGeoJSON` always creates a GeometryCollection.

@jfgirard's implementation in #712 contains a heuristic: if converting each layer
in a FeatureGroup to GeoJSON produces only Points, it returns a MultiPoint, otherwise
it returns a GeometryCollection. I don't like this solution, because it results in
the opposite problem: a GeometryCollection containing only Points can't be round tripped.

Instead I think Leaflet should have a dedicated MultiMarker class in the style of
MultiPolyline and MultiPolygon. But implementing that is tricky, because it needs to
support L.GeoJSON's `pointToLayer` option.

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -1,7 +1,7 @@
 describe("L.Marker#toGeoJSON", function () {
 	it("returns a Point object", function () {
 		var marker = new L.Marker([10, 20]);
-		expect(marker.toGeoJSON()).toEqual({
+		expect(marker.toGeoJSON()).to.eql({
 			type: 'Point',
 			coordinates: [20, 10]
 		});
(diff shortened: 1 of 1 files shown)


MenschMarcus:
Hey folks, I have trouble using this function. You provide me with a full example how to use a layer.toGeoJson() or layer.toJson() function? I can not find it anywhere in the api :/

Issue #4, opened 2011-04-01
Ability to hide/show markers and layer groups


mourner:

rwwagner90:
+1 I need this

Jakobud:
FYI, this is all solved when using the new custom pane management and creation. You can use css to easily hide/show the pane DOM elements. It will be available for v1.0.

rwwagner90:
@Jakobud I am using 0.8-dev, which appears to have some support for custom panes. However, the changelog isn't very specific on how to use this functionality. I have a FeatureGroup, which I am storing shapes in, and I would like to have that FeatureGroup live in a custom pane, instead of the overlay pane. I tried this:
```js
var filterPane = map.createPane('filterPane');
var drawnItems = new L.FeatureGroup({pane: filterPane});
map.addLayer(drawnItems);
```

This does create the new filter pane, as I wanted it to, but then it adds the FeatureGroup to the overlay pane, and not the filter pane, as I specified.

Jakobud:
This is because in the current 0.8-dev branch there was not yet support for adding vector elements to custom panes. They were still added to the overlay pane because the map only had a single renderer. 

Several months ago I created a custom branch with support for vectors in custom panes. See this pull request here: #2786 

It works really great and I am currently using this custom branch for a project without any issues. However @mourner wants the code refactored and changed up a bit because of I broke some backwards compatibility regarding plugins and some other stuff in the code. So take it with a grain of salt.

rwwagner90:
@Jakobud I implemented a solution using custom renderers. It's not as clean as it would be if I could just set the pane for the FeatureGroup and be done with it, but it does work. I would like to see your PR merged in, so we could do vector pane support out of the box. How goes the refactoring of it?

Jakobud:
Eh I'm kind of in a holding pattern at the moment. I'm not sure how to do some of what @mourner is suggesting. Please feel free to look at yourself and contribute. The current PR is not going to make it in since he wants these changes.

Essentially my solution works like this:
1. Create new pane
2. Create layer and in the layer options, specify the custom pane
3. If it's a vector element, a new renderer will be added to the custom pane and used for the vector element

The solution @mourner would like to see:
1. Create new pane
2. Create new renderer and add it to a custom pane
3. Create new vector layer and instead of `.addTo(map)` do `addTo(renderer)`

[...]

PR #1516, opened 2013-03-14
showing and hiding of markers/paths/groups, #4

snkashis:
to avoid adding and removing of layers when toggling of visibility, not destroying, is preferred.

includes svg and (rough) canvas support. two debug pages(regular,and canvas) included for testing purposes, which have timers set demonstrating behaviors.

most likely needs work regarding svg class handling.

Fixes issue #4


php4fan:
How was this finally solved? Can't find anything in the docs

PR #1572, opened 2013-04-08
Compatibility with browserify

jfirebaugh:
In a browserify package, `this` will refer to a node-style exports object, not
window. So just reference `window` directly.


dougajmcdonald:
Is this fix in a particular release? still seeing the issue in 0.7.5

Issue #1509, opened 2013-03-08
Target zoom in zoomstart event data


Jakobud:
I wanted to run an event that would tell me what zoom level I was zooming to, when zooming:

```
map.on('zoomstart', function(event)
{
  console.log(event.target);
  console.log(event.target._animateToZoom);
}
```

When loading up a map, the first time I zoom, this returns `undefined`.  The next time I zoom (and every time after that) it returns the expected result.

The `undefined` is returned even though I inspect the `event.target` from the previous `console.log` and I can see that the value is there and that it's clearly not `undefined`. 

Any clue as to why I'm seeing this behavior?

mourner:
Closing, see https://github.com/Leaflet/Leaflet/pull/1600#issuecomment-77186793

PR #1600, opened 2013-04-16
Include view information on view changed events

kamalmarhubi:
*NB This PR is to initiate discussion and is not intended to be merged in its present state.*

Events like `move*` and `zoom*` could tell their handlers a bit more about what is changing.  I propose changing their documented type from Event to ViewEvent. The additional data would look something like this:

``` js
{ from:
    { center: {lat: 45.5081, lng: -73.555} // a LatLng object
    , zoom: 13
    }
, to:
    { center: {lat: 45.5081, lng: -73.555} // a LatLng object
    , zoom: 14
    }
}
```

Then the handlers can refer to `event.from.zoom` or `event.to.center` and so on instead of having to poke at `event.target`, where some of this information is only available through undocumented internal properties.

This would solve #1509, possibly among others.

You can see a start on this for the `zoomstart` event in this PR.  This is the only event I need for what I'm trying to achieve, but I'd like to see this implemented.  I'm hoping for some feedback before proceeding any further.

Questions:
* Which events need this information? I list `move*`, `zoom*`, `viewreset`, maybe `load`. Others?
* What information should be included? I've added lat/lng and zoom, but the pixel offsets might be of interest too.
* Is the naming / data layout right?
* Are there any reasons not to do this?

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -531,7 +531,12 @@ L.Map = L.Class.extend({
 			this.fire('movestart');
 
 			if (zoomChanged) {
-				this.fire('zoomstart');
+				// _resetView is called on initialization; getCenter throws if the map isn't initalized,
+				// hence the _loaded check here.  This should probably be done differently.
+				this.fire('zoomstart',
+						{ from: !this._loaded ? {} : { center: this.getCenter(), zoom: this._zoom},
+						  to:   { center: center, zoom: zoom }
+						});
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


mourner:
I'm currently refactoring the way move/zoom events work — see #3278. I've come to realize that we can't include target zoom data in `zoomstart` event, because this event is fired not just when starting a usual zoom animation, but also in other cases when the target zoom isn't available, or the animation curve is not defined — e.g. when you start pinching a map with your fingers on mobile, `zoomstart` should fire, but you don't know where you end up. Specifically for animation, there's now a `zoomanim` event that contains target `center` and `zoom` in event data. So I'm closing the issue.

kamalmarhubi:
That makes sense. Thanks for the update.

Issue #1616, opened 2013-04-24
Control animation in setView/panTo/setZoom/fitBounds/etc.


mourner:
Ability to force animation on/off in view change calls like `setView` and `fitBounds`.
Combined ticked from #1420, #340, #1164, #483.

PR #1617, opened 2013-04-24
Control animation in setView/panTo/setZoom/fitBounds/etc.

mourner:
Implements #1616. Since zooming and panning have different animation mechanisms, I propose the following API:

``` js
map.setView(center, zoom, <ZoomPanOptions>);
map.fitBounds(bounds, <ZoomPanOptions>);
map.panTo(center, <PanOptions>);
map.setZoom(center, <ZoomOptions>);
...
```

ZoomPanOptions:

``` js
{
  // true will reset the view to clear state completely (without any animations etc.)
  reset: false,
  zoom: <ZoomOptions>, // will be used for zoom change animation
  // will be used for panning animation (also includes duration, easing etc.)
  pan: <PanOptions>,
  // passing animate will be an equivalent of passing animate to both zoom and pan options
}
```

Both `zoomOptions` and `panOptions` have `animate` option. Panning behavior:

``` js
// not specifying `animate` will cause animated pan if animating not more 
// than a screen away or reset the view if it is
map.setView(center, zoom); 
map.panTo(center);

// animate: true will force it to always animate pan if possible
map.setView(center, zoom, {pan: {animate: true}}); 
map.panTo(center, {animate: true});

// animate: false will turn off animation, just shifting the map pane
// if not more than a screen away, resetting the view otherwise
map.panTo(center, {animate: false});

// panBy is similar to panTo but it never does the pan distance check,
// so it either animates or shifts map pane depending on animate options
map.panBy(offset, {animate: false});
```

Zooming behavior:

``` js
// no `animate` will cause animated zoom if zoom origin is inside the current view
map.setView(center, zoom);

// animate: true will attempt animation disregarding where the zoom origin is
map.setView(center, zoom, {zoom: {animate: true});

// animate: false will always cause full view reset when zooming
map.setView(center, zoom, {zoom: {animate: false});
```

It may look a bit complicated at first but it's the simplest API I could think of that covers all the use cases.

@danzel @tmcw @jfirebaugh Comments? Suggestions?

### DIFF
--- a/src/map/anim/Map.PanAnimation.js
+++ b/src/map/anim/Map.PanAnimation.js
@@ -14,7 +14,7 @@ L.Map.include({
 			this._panAnim.stop();
 		}
 
-		if (this._loaded && !options.reset) {
+		if (this._loaded && !options.reset && options !== true) {
 
 			if (options.animate !== undefined) {
 				options.zoom = L.extend({animate: options.animate}, options.zoom);


hon:
Is there a way to set animation option's in L.map? One won't invoke `setVeiw` under some circumstances.

Issue #791, opened 2012-07-10
Center positions for controls


kekscom:
For UI controls in general, other positions than viewbox corners would be appreciated.
Suggestion: allo top, right, bottom, left as well. Especially on mobile, Zoom is best placed right, bottom right or bottom.

Addition: once multitouch has been detected, zoom controls could be removed - at least by some "auto" setting.

williamjriley:
It would actually be very simple to add this functionality, by adding two lines to this code in leaflet.js:

Beginning at line 4900 I believe?
`t("top", "left"),`
`t("top", "right"),`
`t("bottom", "left"),`
`t("bottom", "right"),`

The two lines:
`t("top", "center"),`
`t("bottom", "center")`

This would allow for use of positions 'topcenter' and 'bottomcenter'

Then simply add css class for 'leaflet-center':
`.leaflet-center {`
`    left: 50%;`
`    transform: translate(-50%, 0%);`
`}`

perliedman:
Hi, thanks for the suggestion.

However, this is not as simple as it might seem. See #5554, which is an attempt to add this. Unfortunately, we still support browsers that do not have CSS `transform`, so not easy to add at this point.

Also, for the future, please do not refer to line numbers in the build Leaflet file in `dist`, but preferably the actual source file.

williamjriley:
Fair enough. Apologies for the line number, but I was in the middle of a work-day when I posted, so I didn't have time to go through the source. For my personal needs, we are not supporting IE < 9, and so I am okay with just making this change in a local copy and using it. I feel like it would still be good to add this functionality, and advise that the extra positions are only available in browsers that support transform, but that's only my opinion. Thank you for the response!

PR #1652, opened 2013-05-05
Allow for more control positions

snkashis:
New positions including "topcenter","bottomcenter", "middleleft", "middleright".

New debug test page included.

References #791

### DIFF
--- a/debug/map/alt_controls.html
+++ b/debug/map/alt_controls.html
@@ -1,0 +1,54 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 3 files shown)


MuellerMatthew:
Instead of trying to define alternative control positions, why not add a generic option such as `position: 'undefined'` to allow users to add a custom control to the map which does not fall inside of the "_controlCorners" boxes, and which is then placed inside a generic div container which overlays the entire map.  It would be the users job to use css, etc to define how the control element actually fits into the container.

Issue #1798, opened 2013-06-27
Scaling tiles after supported zoom levels


DrYSG:
I am using Leaflet for offline Map viewing (see my demo of using PouchDB for offline tile storage at: http://gis.stackexchange.com/questions/62070/offline-slippy-map-tiles-database-for-leaflet/62073#62073 )

As a base layer I am using the NASA BlueMarble image. The full scale image at 8 levels of zoom is "only" 186MB, 43K tiles of JPG. I would like to generate this to 10 or 11 zoom levels, but it is too heavy for offline.

But for my use case (as a backdrop for a few selected small footprint aerial photos at zoom 11) it would be fine to just zoom the level 8 zoom pixels. (users don't care about grainy backdrops, just full res aerials).

Does this sound like a reasonable request for the core product?

Someone did suggest I look at: ZoomTMSLayer: https://github.com/tilemapjp/ZoomTMSLayer but I do not have TMS layers. What I have are EPSG4326 LL84 tiles in a PouchDB database, and I am using the FunctionalLayer plugin for leaflet to display.

PR #1802, opened 2013-06-27
TileLayer maxNativeZoom option for scaling tiles on unsupported zoom levels

mourner:
Implements #1798. Suppose you have a tile layer with maxZoom 18, but you need it up to 20 and OK with scaling of tiles. Then you would do this:

``` js
var tiles = L.tileLayer(..., {
    maxZoom: 20,
    maxNativeZoom: 18
});
```

Tested and seems to work well for both simple tile layers and WMS layers.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -287,7 +287,7 @@ L.TileLayer = L.Class.extend({
 		if (!this._map) { return; }
 
 		var map = this._map,
-			bounds = map.getPixelBounds(),
+		    bounds = map.getPixelBounds(),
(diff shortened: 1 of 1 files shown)


nothingisnecessary:
Sorry if off-topic, but seems related: Is there a 'tweening' option for tile services that do not provide tiles at each zoom level? (For example, on offline mobile devices with local tile servers where space is at a premium and may only store tiles at every other zoom level). See related SO question: http://stackoverflow.com/questions/28904160/how-to-use-chunky-coarse-grained-zoom-levels-with-leaflet-maps

PR #1828, opened 2013-07-04
Added nuspec file

resnyanskiy:
Source file for nuget package (http://nuget.org/packages/leaflet)


davidr-halomicrosystems:
Hi there, will Leaflet get an update on the package?

jieter:
In #4341, you'll see that Leaflet maintainers do not maintain the NuGet package...

PR #2272, opened 2013-12-08
fix regression from #2250 (map.getSize)

Zverik:
With #2250 I have unknowingly broken a chain of `invalidateSize(); fitBounds()` calls when map is not loaded, but the container was resized in meantime. It is a complex use-case, involving `getSize()` calls when map is not yet loaded, and using `fitBounds()` for initialization, instead of `setView()`. See [this jsfiddle](http://jsfiddle.net/uA7Pe/1/) (by uncommenting ).

This PR removes `map._sizeChanged` property, since it is basically a parameter to `getSize()` method, and adds `refresh` parameter to `setSize()`. Methods that can call `setView()` (including `fitBounds()`) now call `getSize(true)`, so they always use current values for container size.

I hope this can be merged in *stable* for 0.7.2 release.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -73,7 +73,7 @@ L.Map = L.Class.extend({
 
 	setZoomAround: function (latlng, zoom, options) {
 		var scale = this.getZoomScale(zoom),
-		    viewHalf = this.getSize().divideBy(2),
+		    viewHalf = this.getSize(true).divideBy(2),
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Closing as incredibly old. Send a new PR against the `stable` branch, which holds the 0.7.x code, if you think this is still an issue.

PR #2310, opened 2013-12-17
Evented base class to use in favor of mixing in event methods

### LINKED ISSUE #2280: Ditch Events mixin, make Evented base class instead
Currently we copy all Events functions into each Leaflet class. For some reason I initially thought standalone mixins are a great idea in Leaflet, but now I see that the `includes` property is only used to include `Events`, and most of the Leaflet classes are evented. 

So it seems to me that it makes more sense to ditch the `includes` concept and make the event methods inherited instead of mixed in. This will improve memory footprint because the methods will be reused from the prototype chain, and will look cleaner in the code:

``` js
// proposed (or we could call it EventEmmiter)
L.Layer = L.Evented.extend({...});

// old
L.Layer = L.Class.extend({
    includes: L.Mixin.Events, ...});
```

People will still be able to to use the methods manually without inheriting if there's a need:

``` js
L.Evented.prototype.on.call(this, 'myevent', handleEvent);
L.Evented.prototype.fire.call(this, 'myevent');
```

cc @tmcw @danzel

mourner:
Closes #2280

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -1,16 +1,9 @@
 describe('Events', function () {
-	var Klass;
-
-	beforeEach(function () {
-		Klass = L.Class.extend({
-			includes: L.Mixin.Events
-		});
-	});
 
 	describe('#fireEvent', function () {
 
 		it('fires all listeners added through #addEventListener', function () {
-			var obj = new Klass(),
+			var obj = new L.Evented(),
 				spy1 = sinon.spy(),
 				spy2 = sinon.spy(),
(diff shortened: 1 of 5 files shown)


voltagex:
What's the current thoughts on this? There's very little documentation around on how to make an Evented Control, this would make it a lot simpler.

voltagex:
Were all Controls ever going to be Evented? Is the recommended way to make custom controls that support stuff like .on() still to use a Mixin?

mourner:
@voltagex the previous approach also works, so something like this: `MyControl.include(L.Events)`.

PR #2346, opened 2014-01-03
Image placeholder appears on map if tile load error

lesserj:
'leaflet-tile-loaded' class was being applied even if a tile failed to load (404 error).  This would cause an image placeholder to appear on the map.

Here is an example:
image

I fixed this by only adding the class if no error.


nkovacs:
This has now bitrotted. I'll open a new pull request.

PR #2347, opened 2014-01-04
Add an editorconfig file

mattiasb:
From http://editorconfig.org/:
"EditorConfig helps developers define and maintain consistent coding
styles between different editors and IDEs. The EditorConfig project
consists of a file format for defining coding styles and a collection of
text editor plugins that enable editors to read the file format and
adhere to defined styles."

I tried my best to derive the specific settings from the Leaflet source
code.
I'm a bit unsure whether Leaflet uses lf or crlf though, there seems to
be a bit of both. :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+
+[**.{json}]
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+
+[**.{js,css,html,md}]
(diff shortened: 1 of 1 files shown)


anetz89:
Adding the configuration at a higher level is a good idea, but it forces you to create a master folder. 
What about adding ".editorconfig" to .gitignore? This would allow everyone to add his own editorconfig to the root folder without adding "noise" to the repository.

Maybe it's time to rethink this issue, as editorconfig is more established after three years.

If not, here's a configuration that fullfills the current lint configuration:

```
# http://editorconfig.org
root = true

[*]
indent_style = tab
indent_size = 4
end_of_line = crlf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

PR #2362, opened 2014-01-09
Added rotate functionality to Marker

jeffsmale90:
Based largely on RotatedMarker.js from Leaflet.PolylineDecorator by
Benjamin Becquet.

https://github.com/bbecquet/Leaflet.PolylineDecorator/blob/master/src/L.RotatedMarker.js

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -56,4 +56,33 @@ describe('DomUtil', function () {
 	// describe('#setPosition', noSpecs);
 
 	// describe('#getStyle', noSpecs);
+
+	describe('#setAnchoredTransform', function () {
+		it('does not rotate or scale if angle is 0 and scale is 1', function () {
+			L.DomUtil.setAnchoredTransform(el, L.point(5, 6), L.point(1, 2), 0, 1);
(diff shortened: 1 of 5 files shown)


DrYSG:
any progress here?

DrYSG:
I tried all three of the solutions below. Since I don't need compatibility with old browsers, I went with the simplests (jacobtoye). But the issue I found was that popUp comments which are located at:
```
            var myIcon = L.icon({
                iconUrl: 'images/carCursor.png',
                iconSize: [64, 64],
                iconAnchor: [30, 20],
                popupAnchor: [30, -32]
            });
```
are also being rotated, and this results in the pop-up covering my marker (the contents of the popup is about 12 lines of text.


https://github.com/bbecquet/Leaflet.PolylineDecorator/blob/master/src/L.RotatedMarker.js

https://github.com/shramov/leaflet-plugins/blob/master/layer/Marker.Rotate.js

https://github.com/jacobtoye/Leaflet.transformmarker/blob/master/Leaflet.transformmarker.js

perliedman:
Closing this off, it's not likely anyone will update this to the current code base and there are plugins that solve this for those who need the functionality.

PR #2389, opened 2014-01-16
Add click option to layers control

frankrowe:
Adds a `click` option to the Layers Control, which disables mouseover expanding and enables click expanding.

Layers control is collapsed on map click, which is the same behavior as on tablets.


IvanSanchez:
Automatic click/touch expansion of `L.Control.Layers` has been implemented since this was reported.

PR #2436, opened 2014-02-05
Update ImageOverlay.js

robinComa:
Adding setBounds method for imageOverlay. Objectives : set img url and bound (before: remove layer and create layer).


allthesignals:
I'm confused, is setBounds not usable on imageOverlay?

PR #2496, opened 2014-02-28
Use proper minus sign (U+2212)

andrewshadura:
Using minus instead of a dash is better because it's usually of the same size as plus, so the buttons look symmetric.


deltab:
@mourner: When you tried it, did you equalize the font size? Currently the font size is set slightly larger for the zoom-out button, so it oversizes the minus sign.

PR #2522, opened 2014-03-06
Options for the handler

AndriiHeonia:
Now you can not only enable or disable map handlers, you can set options for them. See example in debug/map/handler.html.


IvanSanchez:
As of 1.0.0, map handlers take options from the map they belong to. I don't think we'll be changing this behaviour in the near future.

PR #2744, opened 2014-06-17
performance: no need to update the layer if it’s set to noClip and has a...

fab1an:
...lready been simplified+updated.

Saves ~0.5s on iOS7, on certain operations like dragging the map around.

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -52,6 +52,11 @@ L.Polyline = L.Path.extend({
 	initialize: function (latlngs, options) {
 		L.setOptions(this, options);
 		this._setLatLngs(latlngs);
+
+	    // 0: init
+		// 1: projected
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This might still be useful, but there are merge conflicts. @fab1an , do you think this is worth rebasing on top of (at least) 1.0.0-rc1?

fab1an:
Done.

fab1an:
Addendum: I rebased this onto master, it that ok as well?

IvanSanchez:
@fab1an Sure, as long as things are mostly in sync, that's fair.

I'm wondering if this should be done with a `this._painted` flag instead - a state with values [0, 1, 2] might be a bit more confusing.

fab1an:
I don't know why I did it with the numbers. I think to see the ordering, but feel free to change it.

PR #2755, opened 2014-06-21
refactoring of events, it should be about 4 times faster now

fab1an:
I reworked the events-system, this way it's about 4 times faster. Removing listeners for 3000 layers took 160ms on iPad, now it takes ~40ms.

Most performance comes from these observations:
* concatenating object-keys is really slow: http://jsperf.com/member-associative-array-syntax-vs-dot-syntax/3
* concise syntax for initializing object-properties does too many lookups: http://jsperf.com/short-object-init

Remarks:
* I removed the second code path with the simple arrays for simplicity's sake. The memory-overhead by using a couple, or even thousands of simple objects is negligible. If we want to improve in that direction GC-pauses, e.g. because of creating and destroying hundreds of thousands of L.points during projection are the larger problem.

* In the old code there was a bug, if you removed a no-context listener, event[type] would still be declared and listens would return true, even if no listener would be in the array anymore.

* There's another bug in the old code, which is in the new one as well. The whole "don't fire during removed" doesn't work if no function is supplied during removal and all listeners for the type are removed. I didn't change it for now.

* There were a couple of functional differences between the with and the without/context listeners. In the old code you could add the same event/fn pair without context twice, which wasn't possible with a context. 

* I found this commit https://github.com/Leaflet/Leaflet/commit/dcb6f74ef5c5c222de85655e1e39a6be506049f4 but can't understand how it guarantees the execution order of events. Order of property traversal is never guaranteed in JavaScript!

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -247,17 +247,32 @@ describe('Events', function () {
 			obj.removeEventListener('test', spy, foo2);
 
 			expect(obj.listens('test')).to.be(false);
+			
+			//Add and remove a listener without context
+			obj.addEventListener('test', spy);
+			obj.removeEventListener('test', spy);
+
+			expect(obj.listens('test')).to.be(false);
 		});
 
 		it('makes sure an event is not triggered if a listener is removed during dispatch', function () {
 			var obj = new L.Evented(),
-			    spy = sinon.spy();
+			    spy = sinon.spy(),
+				spy2 = sinon.spy(),
+				spy3 = sinon.spy(),
+				foo = {};
 
+			/* without context */
 			obj.addEventListener('test', function () { obj.removeEventListener('test', spy); });
 			obj.addEventListener('test', spy);
 			obj.fireEvent('test');
 
 			expect(spy.called).to.be(false);
+			
+			/* with context */
+			obj.addEventListener('test2', function () { obj.removeEventListener('test2', spy2, foo); }, foo);
+			obj.addEventListener('test2', spy2, foo);
+			obj.fireEvent('test2');
 		});
 	});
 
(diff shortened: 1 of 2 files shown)


nmccready:
@mourner is this still relevant?

mourner:
It's relevant, but I didn't dig deeper into the changes here — the events code is very sensitive and introduce subtle bugs easily, so this PR needs to be reviewed very thoroughly.

fab1an:
Patch-author here. The problem is also that the original code has several bugs, which I listed in my comment.

nmccready:
Cool I was checking to make sure this ticket didn't get lost. It seems like an important one.

fab1an:
Is it still relevant? I'm especially interested in the bugs/edge cases in the original code, shouldn't those be worked on?

IvanSanchez:
@fab1an I think the performance problems from thousands of layers in 0.7.x have been covered by the huge work on event delegation from #3032. 

(Which means the whole events code has changed since, which also means a huge merge conflict.)

fab1an:
Looks really awesome, but still: What about the bugs or inconsistent behaviour I mentioned in my initial message here?

IvanSanchez:
I guess those should be split off to separate bug reports (with reproducible cases running on the latest `master`, etc etc).

fab1an:
How should I add the test-cases? With a pull-request that fails?

IvanSanchez:
Most times it's enough to demonstrate using jsbin, plunkr, codepen or http://playground-leaflet.rhcloud.com/. A failing unit test is definitely going the extra mile :-)

fab1an:
First https://github.com/Leaflet/Leaflet/issues/4474

fab1an:
I still think it's worth removing the second codepath with the arrays for simplicity's sake.

mourner:
Closing in favor of #4709 (which includes the changes)

Issue #2779, opened 2014-07-06
0.8-dev, can't add polyline to a custom pane?


Jakobud:
I have the following:

```
map.createPane("myCustomPane");
L.polygon([[0,0],[5,5],[3,3]], { pane: "myCustomPane" });
```

Doesn't show up.  Pane's container is empty.  Looking through the source, it looks like Markers can be easily added to panes.  Can polylines not be added?

`L.Polygon` extends `L.Polyline` extends `L.Path` extends `L.Layer`, which has `overlayPane` set in the options:

https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L5

Am I missing something here?

Adding markers to panes works fine:
```
map.createPane("myCustomMarkerPane");
L.marker([0,0], { pane: "myCustomMarkerPane" });
```

zhxsh:
Keep only one map,it solves my big problem,thank you very much

Issue #1742, opened 2013-06-07
Custom Panes and Pane Management


Jakobud:
It would be excellent to be able to create/delete/manage custom map panes and then be able to specify which panes your layers get added to.  This would allow fine tuned control over css changes to groups of layers at once, allow you to solve z-index issues, and this would even be an excellent solution for #1734.

What has brought about this, is that I use a lot of SVG objects on my map application.  There are certain svg elements that I want to be above others all the time (with regards to z-index).  However svg elements don't use z-index, and instead the elements added to the DOM last are the ones shown on top.  The problem is that I add and remove a lot of the SVG elements to the map during normal usage of the mapping application.  So this constantly changes which SVG elements are on top all the time.

Custom map panes would solve this problem and give the user a LOT more fine tuned control over their groups of layers and how they are displayed and interacted with.

Any thoughts on this?

rwwagner90:
@mourner Does this only work with markers? I was trying to use it for a FeatureGroup, and had no luck.
```js
var filterPane = map.createPane('filterPane');
var drawnItems = new L.FeatureGroup({pane: filterPane});
```
The shapes I add to the FeatureGroup still get drawn in the overlay-pane.

rwwagner90:
To answer my own question, it appears FeatureGroup does not take an options object. You must set the options explicitly, after creation.

```js
var drawnItems = new L.FeatureGroup();
drawnItems.options.pane = 'filterPane';
```

mourner:
@rwwagner90 it does, it just accepts a string rather than reference that I see in your sample.

rwwagner90:
@mourner I thought, based on the sample code above, either would work?

mourner:
Hmm, I should look into why it didn't...

rwwagner90:
I actually was incorrect, when I said that it worked setting it to the string. I had changed something else to make it work. Setting the pane in a FeatureGroup appears to have no effect on where the shapes are drawn. They are still drawn in the overlay layer. However, using the renderer works.
```js
var renderer = (L.SVG && L.svg()) || (L.Canvas && L.canvas());
map.createPane('filterPane');
renderer.options.pane = 'filterPane';
draw = {
      polyline: {
        metric: true,
        shapeOptions: {
          renderer: renderer,
          color: color,
          weight: 5,
          fillOpacity: 0
        }
      },
      polygon: {
        allowIntersection: false,
        showArea: true,
        drawError: {
          color: '#b00b00',
[...]

PR #2786, opened 2014-07-10
Pane management: huge refactoring and rewrite!

Jakobud:
I've been messing with the master branch recently, eager to see the new Pane management features but was disappointed to see that you could not yet assign vector elements to custom Panes (#2779).

Well I started diving into the code to see what I could learn and improve and here are the results.  I have completely rewritten the way that Panes work.  Currently, Panes are simply DOM elements/containers, for example:

```
<div class="leaflet-pane leaflet-marker-pane"></div>
```

The problem is that, in the current state, Leaflet cannot put vector elements on any pane other than the `overlayPane`.  The reason is that the map only uses a single Renderer, who's container would always show up on the `overlayPane`, even if you assigned a different or custom pane to a vector element as you created it.  So therefore, all vector elements get added to the `overlayPane` since that is where the renderer's `svg` (or canvas or whatever) container was.

The solution is that Leaflet needed the ability to use multiple renderers, one for each pane that any vector elements existed on.  Keeping track of multiple renders on panes being just DOM elements was tricky...

### A new class.... L.Pane
Therefore I created a new class/object `L.Pane`.  With this commit, panes are no longer simple DOM elements.  They are now objects, that have a container and (potentially) a renderer.  Moving to this type of setup allowed me to easily keep track of which panes already had renderers when adding vector elements to the map.  Also, the Pane class approach will allow for some more flexibility and additions in the future.  

- `map.createPane()`, `map.getPane()` and `map.getPanes()` are still used the same way.
- `map.getPane()` will automatically create a pane for you if it doesn't exist.
- `map.paneExists()` check if a pane exists.
- To add a DOM element to a pane, use `myPane.add(myDomElement)`
- To remove a DOM element from a pane, use `myPane.remove(myDomElement)`
- If you need to access the DOM element/container of a pane use `myPane.getContainer()`.  This is used several places in the source, for example, whenever the `mapPane`'s container needed to be transformed when zooming and panning the map.
- Access a pane's container's css styles: `myPane.style.zIndex = 10`.
- Vector layers can now be assigned to any pane.  When adding a vector layer to a pane.`map.getRenderer()` will now fetch the renderer from the vector layer's assigned pane.  If a renderer on that pane already exists, it will use that.  If not, it will create it.
- If you don't specify a pane for Vector layers, they end up on the `overlayPane` by default, just like they currently do.

@mourner I don't know what plans you had exactly for Panes, but I've put a lot of time and thought into this and I think this is a really good approach. It adds new features (vectors can now use custom panes too!!) and also using Panes as objects will allow for some more features in the future (maybe hooking panes into map events).  Give it a shot!  What do you think?

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -88,6 +88,10 @@ var deps = {
 		desc: 'Extends LayerGroup with mouse events and bindPopup method shared between layers.'
 	},
 
+	Pane: {
+		src: ['dom/Pane.js'],
+		desc: 'Pane management'
+	},
 
 	Path: {
 		src: [
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -59,7 +59,7 @@ describe("Marker", function () {
 			expect(oldIcon.parentNode).to.be(null);
 
 			expect(marker._icon.src).to.contain('marker-icon.png');
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode).to.be(map._panes.markerPane.getContainer());
 		});
 
 		it("changes an image to a DivIcon", function () {
@@ -73,7 +73,7 @@ describe("Marker", function () {
 			expect(oldIcon.parentNode).to.be(null);
 
 			expect(marker._icon.innerHTML).to.contain('Inner1Text');
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode).to.be(map._panes.markerPane.getContainer());
 		});
 
 		it("reuses the icon/shadow when changing icon", function () {
@@ -87,8 +87,8 @@ describe("Marker", function () {
 			expect(oldIcon).to.be(marker._icon);
 			expect(oldShadow).to.be(marker._shadow);
 
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
-			expect(marker._shadow.parentNode).to.be(map._panes.shadowPane);
+			expect(marker._icon.parentNode).to.be(map._panes.markerPane.getContainer());
+			expect(marker._shadow.parentNode).to.be(map._panes.shadowPane.getContainer());
 		});
 	});
 
--- a/src/dom/Pane.js
+++ b/src/dom/Pane.js
@@ -13,7 +13,7 @@ L.Pane = L.Class.extend({
 		// Create the container
 		this._container = L.DomUtil.create('div', className, container || map._mapPane.getContainer());
 
-		// Shortcut for accessing Pane's container's styles
+		// Shortcut for accessing DOM container styles
 		this.style = this._container.style;
 	},
 
@@ -41,14 +41,14 @@ L.Pane = L.Class.extend({
 	},
 
 	// Add DOM element to pane
-	add: function (container) {
+	appendChild: function (container) {
 		this._container.appendChild(container);
 	},
 
 	// Remove DOM element from pane
-	remove: function (container) {
+	removeChild: function (container) {
 		this._container.removeChild(container);
-	}
+	},
 
 });
 
--- a/src/dom/Pane.js
+++ b/src/dom/Pane.js
@@ -6,13 +6,36 @@ L.Pane = L.Class.extend({
 
 	initialize: function (map, name, container) {
 		this._name = name;
+		this._renderer = null;
 
+		// Define the class name for the container
 		var className = 'leaflet-pane' + (name ? ' leaflet-' + name.replace('Pane', '') + '-pane' : '');
+
+		// Create the container
 		this._container = L.DomUtil.create('div', className, container || map._mapPane.getContainer());
+
+		// Shortcut for accessing Pane's container's styles
+		this.style = this._container.style;
+	},
+
+	getName: function() {
+		return this._name;
 	},
 
 	getContainer: function () {
 		return this._container;
+	},
+
+	getRenderer: function (layer) {
+		var renderer = this._renderer;
+
+		// Create renderer if it doesn't exist
+		if ( !renderer ) {
+			renderer = this._renderer = (L.SVG && L.svg()) || (L.Canvas && L.canvas());
+			renderer.options.pane = this._name;
+		}
+
+		return renderer;
 	},
 
 	add: function (container) {
@@ -22,7 +45,6 @@ L.Pane = L.Class.extend({
 	remove: function (container) {
 		this._container.removeChild(container);
 	}
-
 
 });
 
--- a/src/dom/Pane.js
+++ b/src/dom/Pane.js
@@ -1,0 +1,23 @@
+/*
+ * L.Pane is used for pane management
+ */
+
+L.Pane = L.Class.extend({
+
+	initialize: function (map, name, container) {
+		console.log('initializing pane object: ' + name);
+
+		this._name = name;
+
+		var className = 'leaflet-pane2' + (name ? ' leaflet-' + name.replace('Pane', '') + '-pane' : '');
+		this._container = L.DomUtil.create('div', className, container || map._mapPane);
+
+		console.log(this);
+	}
+
+
+});
+
+L.pane = function (map, name, container) {
+	return new L.Pane(map, name, container);
+};
--- a/src/dom/Pane.js
+++ b/src/dom/Pane.js
@@ -6,7 +6,6 @@ L.Pane = L.Class.extend({
 
 	initialize: function (map, name, container) {
 		this._name = name;
-		this._renderer = null;
 
 		// Define the class name for the container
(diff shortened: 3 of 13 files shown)


mourner:
OK, starting to look at this again as we'd want to get the pane management API stable for the first beta release already, so that subsequent updates are mostly fixes.

Jakobud:
I will need to sit down and get my head wrapped around this again. It's been a while. Last we left off I think we need to get rid of the "auto-create panes" magic as well as figure out if it's possible to do all of this without breaking any backwards compatibility, since the proposed features makes panes Objects that have a DOM container, vs the current Panes are DOM containers approach.

mourner:
Yeah. My proposed way to deal with this is described in https://github.com/Leaflet/Leaflet/pull/2786#issuecomment-61964836

Jakobud:
Okay, so per your suggestions, I should be getting rid of the Panes as an object and try to keep them as DOM elements (in order to preserve backwards compatibility) but somehow reference the renderer as a custom attribute on the DOM object, right? Something like `data-renderer="myRenderer"`?

I have several questions. First of all, currently renderers are not referenceable by name or even stored anywhere by name. So this functionality would have to be added, similar to how panes are stored and referenced by name. Does that sound right?

Also, you said a pane can have more than one renderer. First of all, how would that work if I am adding a renderers as a data attribute to the DOM object? Would I have to hack together something like this:

`data-renderers="myRenderer1|myRenderer2"`

Also, if a pane can have multiple renderers, then that means whenever someone adds a vector layer to a pane, they would have to specify not just the pane, but also the renderer.

perliedman:
I like the API outlined by @mourner in https://github.com/Leaflet/Leaflet/pull/2786#issuecomment-61964836. It simple, elegant and mostly intuitive (at least as intuitive as I think multiple renderers and custom panes can be). Also, more importantly, it is as far as I can tell, backwards compatible with all basic use cases.

Regarding @Jakobud's comment above, I don't think @mourner's approach needs to store which renderer or renderers are used for a given pane: each renderer knows which pane to use (assigned on creation), and layers are added to an explicit renderer, or the map's implicit renderer, not to a pane.

Jakobud:
With that API would users need to add vector layers to renderers and then also to the map?

```
var polygon = L.polygon(...).addTo(renderer).addTo(map);
```

But this is only for vector layers. If you are adding like a marker, then you have to add it to a pane and then the map:

```
var marker = L.marker(...).addTo(map);  // but which pane?
// Alternatively:
var marker = L.marker(...).addTo(pane).addTo(map);
```

The thing that is confusing about that approach is that users are going to be confused how to add things to the map. If some layers you add to a renderer and some you add to a pane (and then some do you add them just straight to the map?) that would be confusing more most people I would think...

I'm either missing something or I'm not convinced on that approach yet. It seems like we should only be adding layers to the map like we currently do:

```
var marker = L.marker(...).addTo(map);
```

Which would add the layer to a predefined pane and/or renderer based off of what kind of layer it is, just like how things are now. But you can optionally specify a certain pane or renderer if you wanted to pick a different one. That seem more straightforward.

bjorm:
Sorry to barge in on this discussion, but this feature -- if it does what I think it does -- is quite critical for me. To be sure, will this feature enable me to arbitrarily stack tiles and SVG layers like in the screen below?
layering

perliedman:
@Jakobud for most people, this change will make no difference, since you can still just add layers to the map. These layers will use the map's default renderer, and everything will work exactly as it is now.

If you check the example in @mourner's original comment, you can see that you only add a layer to the renderer:

```
var renderer = L.canvas({pane: 'myPane'}).addTo(map);
var polygon = L.polygon(...).addTo(renderer);
```

The renderer, acting sort of like `LayerGroup` will display the layer in the map, since the renderer has been added to the map.

I'm not sure how markers fit into this. To be sure, the default has to be like it is now: adding a marker to the map adds them to the marker pane. Since markers don't use renderers, they don't fit into the renderer API. I guess you could add an option to specify which pane to add the marker to, but maybe there's a more elegant way?

mourner:
Yes, what Per says. As for markers — all layers accept a `pane` option (with the exception of vector layers). 99% of the users won't care about this change at all — they'll continue to work with markers and polylines as they did before. What I like about the renderer/pane/layer relation I describe above is that it doesn't introduce new concepts at all. It works like a LayerGroup. If you need to combine some vectors into a group that's rendered by a separate renderer in a custom pane, you do that explicitly.

Jakobud:
Ah okay I think I see what you guys are getting at finally. Let me take some time with this and get back to you.

@bjorm Yes, it will allow you to freely add vector layers to as many different custom panes as you like. And since they are on different panes, you will be able to adjust their display order via css (z-index, etc). I have this current branch working well for a project but I'm going to be refactoring it per the above discussion.

Jakobud:
I came across a hurdle last night:

```
map.createPane('testPane');
var renderer = L.svg({pane: 'testPane'});
var polygon = L.polygon([[0,0],[1,1],[1,0]]);
polygon.addTo(renderer);  // <--- Can accept renderer object OR map object
```

So L.Path's `addTo()` has to accept either an L.Map object or an L.Renderer object. How do I differentiate between the two?

Jakobud:
Ah wait a minute. Just realized it doesn't matter. I just need to add something like this to `L.Path`:

```
        addTo: function (object) {
		object.addLayer(this);
		return this;
	},
```
Then add the appropriate `addLayer()` function to `L.Renderer` (or `L.Canvas`/`L.SVG`)... Okay I think this sounds doable now.

Jakobud:
Well guys... I'm having a hard time figuring out how to do this. It's just not clear. When you `addTo(renderer)`, my thought process was that the layer goes through the normal process of being added to the map. But when it comes time to add the vector to the renderer here:

```
// L.Path
onAdd: function () {
		this._renderer = this._map.getRenderer(this);
		this._renderer._initPath(this);

		// defined in children classes
		this._project();
		this._update();

		this._renderer._addPath(this);
	},
```
Then it would know that you specified a renderer and it would use that renderer instead of the map's renderer. But then I'm back to the point where I need to be able to differentiate between a `map` object an a `renderer` object so that I know I'm adding to the renderer instead of the map.

Also, once I get past that hurdle, should the layer still be added to the map and have a L.stamp ID generated for it etc? Should you be able to do `map.remove(myVectorLayer)`? Or would you have to recreate all that logic on the `L.Renderer` class?

If you guys can give me some guidance I would appreciate it. There is just a lot in the source that I'm not sure what it's purpose is. @mourner not enough source comments!! :-)

Jakobud:
@mourner, so over on #1742 @rwwagner90 just showed me something that I didn't realize was possible: When creating a vector layer, **you can specify the renderer in the layer options**. That right there might be the best solution for all of this. AND it already exists in the code.

#### Workflow:

```
// Create custom pane
map.createPane('myPane');

// Create custom renderer, add it to custom pane
var myRenderer = L.svg({pane: 'myPane'});

// Create vector layer that will now end up on custom pane in your renderer
var polygon = L.polygon([[0,0],[1,1],[1,0]], {renderer: myRenderer}).addTo(map);
```

I need to set this up and test it more tonight, but according to this workflow logic, all the functionality to do what we need to do is already in 0.8-dev, even without this PR... Have you guys tried this or messed with it? It seems like a very logical workflow.

#### Room for improvement:

There are a couple inconsistent and non-straightforward things I don't like:

1. If you create a vector layer and specify a custom pane in the options, this will be ignored. A vector layer will always go wherever the renderer is. Doesn't matter what pane you specify.
2. You specify custom panes for layers by their string name `{pane: 'myCustomPane'}` but when specifying renderers you have to specify the renderer object: `{renderer: myRenderer}`. Perhaps it would be wise to keep track of and reference renderers on the `map` object in the same way that we keep track of and reference panes. I don't think that would be hard to do. Something like `map.getRenderer('myRenderer')`.

Thoughts? Opinions?

Jakobud:
After more testing I'm pretty convinced this is the way to go. This approach allows you to add any number of renderers to any number of custom panes and freely add vector layers to any of those renderers:

```
// Create pane
map.createPane('testPane');

// Create two renderers both on the same pane
var renderer = L.svg({pane:'testPane'}).addTo(map);
var renderer2 = L.svg({pane:'testPane'}).addTo(map);

// Create two vector layers, one on each pane
var polygon = L.polygon([[0,0],[1,0],[0,1]], {renderer:renderer}).addTo(map);
var polygon2 = L.polygon([[0,0],[1,0],[0,1]], {renderer:renderer}).addTo(map);
```

Works perfectly as expected.

Jakobud:
I have an idea for something that might make this more straightforward... before I get into this, @mourner can you give me an example of why you want people to be able to make multiple renderers on a pane? What is the purpose of that? Just to be flexible?

mourner:
No, the ability to make multiple renderers on a pane isn't a strict requirement. I'm only concerned about easy and consistent API and compatibility.

Jakobud:
Closing this. I'm going to be opening a new pull request soon with improved and backwards compatible code for this feature.

PR #2800, opened 2014-07-18
Modify L.DivIcon so that the html option can be either a string OR a HTML Node

justinmanley:
I modified L.DivIcon so that the `html` parameter passed in the `options` object can be either a string or an HTML object.  This is purely for the convenience of application + plugin developers.  For example:

``` javascript
var icon, marker;
....
this._textarea = L.DomUtil.create('textarea', 'sample-application-textarea'),
icon = new L.DivIcon({ html: textarea });
marker = new L.Marker(L.latLng(10, 10), { icon: icon });
...
```

This will make it more convenient for plugin developers to store a reference to the DOM node that is being set as the icon.  This pull request also sets a `_container` property on the `DivIcon`, making it easier for plugin developers to access the `<div>` container holding the icon's HTML.

I modeled this on [`L.Popup#_updateContent`](https://github.com/Leaflet/Leaflet/blob/master/src/layer/Popup.js#L171), since popups already exhibit similar overloading (they accept either a string or an HTML element).

I also added a test in `/spec/suites/layer/marker/DivIconSpec.js`.

### DIFF
--- a/spec/suites/layer/marker/DivIconSpec.js
+++ b/spec/suites/layer/marker/DivIconSpec.js
@@ -1,0 +1,14 @@
+describe("DivIcon", function() {
+	var map;
+
+	beforeEach(function () {
+		map = L.map(document.createElement('div')).setView([0, 0], 0);
+	});
+
+	describe("#createIcon", function() {
+		it ("Accepts an HTMl node", function() {
+			var div = document.createElement('div'),
+				icon = new L.DivIcon({ html: div });
+		});
+	});
+});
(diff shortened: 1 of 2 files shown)


perliedman:
I'm closing this as too old to be useful now.

We've actually merged accidentally merged #5517 with equivalent functionality before finding @mourner's comment here. If we really want to go down this path, this comment is a good starting point for how to do it: https://github.com/Leaflet/Leaflet/pull/2800#issuecomment-49528071

Issue #2675, opened 2014-05-08
Bugs after initializing a TileLayer with a bad url


bennlich:
If you initialize an L.TileLayer with an empty or bad url, and then use `setUrl` to reset the url to something good, the TileLayer displays properly, but it won't fire any `'loading'` events when panning and loading new tiles.

In 0.7 this seems to be because `_tilesToLoad` gets screwed up by the bad url (somehow it becomes negative), but I haven't verified that this is the cause in 0.8.

Here's a fiddle: http://jsfiddle.net/qna75/1/.

PR #2802, opened 2014-07-21
Fixed bug with TileLayer.setUrl()

bennlich:
Fixes https://github.com/Leaflet/Leaflet/issues/2675.

After changing a `TileLayer` url with `setUrl()`, `_abortLoading()` needed to be called during `_reset()` to prevent lingering callbacks from decrementing the `_tilesToLoad` counter.

Also, the `complete` attribute of tiles is somehow getting set to `true` when they are given bad `src` urls (e.g. by passing an empty string into `setUrl()`). This prevented `_abortLoading()` from disabling the tiles' `onerror` and `onload` callbacks (now fixed).

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -143,11 +143,11 @@ L.TileLayer = L.GridLayer.extend({
 		for (i in this._tiles) {
 			tile = this._tiles[i];
 
-			if (!tile.complete) {
-				tile.onload = L.Util.falseFn;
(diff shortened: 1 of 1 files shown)


jacobp100:
I'm needing to use `_abortLoading` in my project---is this something that should be made part of the public API and documented?

perliedman:
@jacobp100 please don't just add to a very old and merged pull request, open a new issue if you believe you have a new feature request. If you do, also please state more clearly what you're trying to achieve and why you think this would be necessary.

PR #2936, opened 2014-10-08
#2662 - setStyle now takes className

### LINKED ISSUE #2662: Path `setStyle` ignores `className`
Class cannot be set after Path object is constructed. The only workaround is to pass in the options to the constructor, which isn't always feasible. In my use case, I want to change the class(es) on the Polygon based on live information.
http://jsfiddle.net/mattt2000/85gv4/1/

stuporglue:
http://leafletjs.com/reference.html#path-setstyle says that it accepts a  <Path options> object which includes a className parameter, however className is ignored. 

This pull request adds className support to setStyle and if accepted would resolve #2662 

If className is set in the <Path options> object, setStyle will set the object's class to the string provided. 

Notes: 
- Class names starting with 'leaflet-' are ignored when finding class names to remove so as not to break internal leaflet assumptions. This results in the case where a user could set a class name (eg. leaflet-customclass) which they can't remove. 
- I do not have a version of IE to test SVG.VML.js on. 
- [jmaxxz suggested](https://github.com/Leaflet/Leaflet/issues/2662#issuecomment-42316958) that addClass and removeClass methods would make more sense. I think those would be useful additions too, but I think there's also value in className working like the rest of setStyle. If there's interest in addClass/removeClass I'd be happy to implement those. 

Here's a simple demo: http://stuporglue.org/downloads/leaflet-2662/

The geojson features have an on('click') handler that sets a new className. CSS changes the style from blue to red.

### DIFF
--- a/src/layer/vector/SVG.VML.js
+++ b/src/layer/vector/SVG.VML.js
@@ -103,6 +103,15 @@ L.SVG.include(!L.Browser.vml ? {} : {
 			container.removeChild(fill);
 			layer._fill = null;
 		}
+
+		if (options.className) {
+			var eClasses = L.Util.splitWords(L.DomUtil.getClass(container))
+			.filter(function(e){
+				return e.indexOf('leaflet-') === 0;
+			})
+			.join(' ');
+			L.DomUtil.setClass(container,eClasses + ' ' + options.className);
+		}
 	},
 
 	_updateCircle: function (layer) {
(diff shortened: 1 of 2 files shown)


jleppert:
Toggling classes to define styles is infinitely more useful than putting all your styles in your javascript, just sayin'. Is the addClass API available? For now we have to access the _path object. blah.

Issue #1853, opened 2013-07-10
Force popup not to close when others open


tmcw:
For abusing the popup api for our share control, other popups close the 'share popup' - it would be great if it was possible to set closable: false or something of that sort to make sure that `map.openPopup` calls do not hide a certain popup

PR #2986, opened 2014-10-31
option to make popup not closeable by other popups

snkashis:
Make popups which stay on screen, even when other popups are triggered. 

References #1853


catbadger:
Is this in RC3?

PR #2382, opened 2014-01-15
Animations 2.0 (fractional zoom, pyramid tiles & zoomPanTo)

mourner:
- **Pyramid tiles** make animations with tiles much more pleasant and open the road for things like `zoomPanTo`.
- **Optimal path animation** (`map.zoomPanTo`) animates smoothly between two locations no matter how distant they are through the shortest zoom-pan curve. #1413
- **Fractional zoom support** for GridLayer makes Leaflet work smoothly with non-rounded zooms like `13.57`, scaling tile layers from round zooms when necessary. #607 #1941 #1309 #1934

#### Pyramid tile loading and animation
- [x] implement basic pyramid loading & animation
- [x] discard zoom levels when not used
- [x] implement pyramid **tile management** so that unnecessary tiles are removed in time for performance
#### Easey-style zoomPanTo (optimal path zoom animations)
- [x] have a basic animation loop through center/zoom values working (without intermediary tiles loading yet)
- [x] implement curved Easey-style zoomPanTo
- [x] add easing other than linear
- [x] tweak zoomPanTo duration and add options
- [ ] make sure events (zoom/zoomstart/zoomend, move/movestart/moveend) are fired correctly during zoomPanTo
- [x] **make zooming animations stoppable** (just like panning)
- [x] fix buggy scroll wheel zooming while zoomPanTo
#### Fractional zoom support
- [x] make fractional zoom working in GridLayer so that it loads tiles from a round zoom and scales the layer with CSS Transform as it would do when animating
- [x] fix fractional zoom wrapping/bounding
- [x] clean up fractional zoom changes
- [x] fix interactions to zoom in/out to the nearest round zoom instead of +/-1
- [x] add an option to not animate to the nearest round zoom on pinch-zoom on mobile

PR #3030, opened 2014-11-15
Add fractional zoom support to fitBounds and getBoundsZoom

cmgiven:
Adds a 'step' option to fitBounds (and equivalent 'step' parameter to getBoundsZoom), defaulting to 1, enabling these functions to use fractional zoom levels (#2382). For example, set 'step' to 0.25 for nearest quarter zoom level, i.e. 10, 10.25, 10.5, 10.75, and so on.

Optimizes getBoundsZoom with a binary search; previous implementation counted up from minZoom using whole numbers.


IvanSanchez:
Superseded by #3523

ronaldcurtis:
Is it still possible to merge this in to master? Would be nice if fitBounds and getBoundsZoom can support fractional zoom deltas

PR #3049, opened 2014-11-21
Add option autoCollapse to close controls after delay on touch devices

### LINKED ISSUE #2713: Collapse layer control after tap on mobile phones
Is it possible to have the layer control collapse/close once a user selects a new layer?  On desktop the layer control collapses when the cursor moves out of the control. On a mobile phone, a user has to tap outside the control to close it. It would be nice if it would close after a layer has been selected.

sharph:
Fixes #2713

[Review on Reviewable](https://reviewable.io/reviews/leaflet/leaflet/3049)

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -12,9 +12,14 @@ L.Control.Layers = L.Control.extend({
 	initialize: function (baseLayers, overlays, options) {
 		L.setOptions(this, options);
 
+		if (this.options.autoCollapse && !this.options.autoCollapseTimeout) {
+			this.options.autoCollapseTimeout = 5000; // default to 5s
+		}
+
 		this._layers = {};
 		this._lastZIndex = 0;
 		this._handlingClick = false;
+		this._collapseTimer = null;
 
 		for (var i in baseLayers) {
 			this._addLayer(baseLayers[i], i);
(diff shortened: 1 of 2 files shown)


src/control/Control.Layers.js
```diff
@@ -12,9 +12,14 @@ L.Control.Layers = L.Control.extend({
 	initialize: function (baseLayers, overlays, options) {
 		L.setOptions(this, options);
 
+		if (this.options.autoCollapse && !this.options.autoCollapseTimeout) {
```
patrickarlt: I think this could be simplified by just using one option `autoCollapse` which could accept a boolean (`true`/`false`) or a millisecond timeout. This would eliminate lines 15-18 entirely. I like the idea of a long timeout like 5 seconds by default.

sharph: I'll have a look at this and revise it tonight. I agree that simplifying the options would be a good solution.

sharph:
I've added another commit that allows the user to set `autoCollapse` to `false`, `true`, or a length in milliseconds, instead of setting the timeout though `autoCollapseTimeout`.

yohanboniface:
We took the sprint opportunity to review this and decided not to go this way: this is a bit too much in UX behaviour, and we think Leaflet should not go this way. We'll still have people wanting the collapse to be in another timing, or with other conditions, etc. And we don't want to arbitrate those choices.
The mantra is that those controls are good for a quick start or very generic needs, but if someone needs more control they need to create their own control (maybe inheriting for the Leaflet generic ones).
Nevertheless, we took the opportunity to make `expand` and `collapse` public methods, so it's easy to listen to `baselayerchange` and expand/collapse with custom UX.

Thanks anyway for your PR on this, and sorry for taking so long to review it.

See #4370

Issue #3058, opened 2014-11-26
add `openOnClick` option for popups


maackle:
Popup options have a `closeOnClick` option to override the default behavior when using `bindPopup`. In my use case I need more control over marker popups and want all the benefits of `bindPopup` (move and remove events on the marker, `openPopup`, `closePopup`, etc) EXCEPT for the `click` event. I see no easy way to do this. I can `.removeEventHandler('click')` but in the way I'm using it, it's tricky to make it not remove my own click handler.

I propose there should also be a `openOnClick` option, defaulting to true, to disable the default click event. I'm going to attempt a pull request for this, but let me know if there's any other better way to handle this.

yohanboniface:
See https://github.com/Leaflet/Leaflet/pull/3059#issuecomment-204728004

PR #3059, opened 2014-11-26
adds openOnClick popup option

maackle:
For #3058

### DIFF
--- a/src/layer/Layer.Popup.js
+++ b/src/layer/Layer.Popup.js
@@ -5,6 +5,7 @@
 L.Layer.include({
 
 	bindPopup: function (content, options) {
+		var events;
 
 		if (content instanceof L.Popup) {
 			this._popup = content;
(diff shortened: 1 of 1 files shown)


src/layer/Layer.Popup.js
```diff
@@ -17,11 +18,14 @@ L.Layer.include({
...
 
 		if (!this._popupHandlersAdded) {
-			this.on({
-				click: this._openPopup,
+			events = {
 				remove: this.closePopup,
 				move: this._movePopup
-			});
+			};
+			if (options.openOnClick !== false) {
```
yohanboniface: options may be  `undefined` here

yohanboniface:
So, after discussing with the team, we think it's not worth adding such an option. Mainly, the mantra is that `bindPopup` is a helper for handling the common simple tasks, but if someone wants more control they'd better attach the popup by themselves with the needed events.
Also, `options` object is passed to the popup constructor, and it's not that clean at the end to use a custom option that is not consumed by the popup itself.
So I'm closing the PR. Thanks anyway for your work! :)
And if we are missing an important use case, feel free to reopen or argue in favor of revisiting this decision!
Thanks again :)

PR #3097, opened 2014-12-21
L.Layer and L.Evented docs

patrickarlt:
Document the new `L.Layer` and `L.Evented` classes for 1.0. This documents some of major changes for in https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md#layers-refactoring.

I wasn't quite sure where to put the docs for `L.Layer`, I've added them here under "Utility" but They might be better documented under "ILayer" but I'll leave it split up for now and we can discuss.

Can @mourner and/or @danzel review?

### DIFF
--- a/atom.xml
+++ b/atom.xml
@@ -1,5 +1,4 @@
 ---
-layout: nil
 title: Leaflet Developer Blog Atom Feed
 ---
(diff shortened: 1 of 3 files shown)


mourner:
Rebased and merged to `gh-pages-master`

Issue #3098, opened 2014-12-21
1.0 Documentation checklist


patrickarlt:
@mourner I'm going to have some more time over the next 2 weeks (holidays!) to tackle more documentation related tasks and thought I would start by pulling together the various change logs and major issues into a checklist so we can get an idea of what needs to be done to the docs.

I'll work on picking stuff off this list and making PRs into the `gh-pages-master` branch.

#### Vectors refactoring

**Currently in progress in https://github.com/Leaflet/Leaflet/pull/3101**

* [x] Added `Renderer` layer (inherited by SVG/Canvas implementations) that handles renderer bounds and zoom animation.
* [x] ~~Added `SVG` & `Canvas` classes that now contain all renderer-specific code for vector layers and can be added as layers to the map.~~ *Probably not worth documenting at this time*
* [x] All vector layer classes (`Path`, `Polyline`, etc.) now don't contain any renderer-specific code and instead can be passed a renderer layer to use as `renderer` in options;
* [x] Removed `MultiPolyline` and `MultiPolygon` classes since multiple rings are now handled by `Polyline` and `Polygon` classes respectively. Layers with multiple rings now perform much better (since each is now physically a single path object instead of being a `FeatureGroup` of layers).
* [x] ~~Added retina support for Canvas layers.~~ *Doesn't include API changes that need to be documented*
* [x] Improved default vector layer styles.
* [x] Added `Polyline` and `Polygon` `getCenter` for getting centroids.
* [x] Changed `Circle` to inherit from `CircleMarker` and not the other way.

#### Layers refactoring

**Some items in progress in https://github.com/Leaflet/Leaflet/pull/3097** and https://github.com/Leaflet/Leaflet/pull/3107

* [x] Added `Layer` class which all layers added to a map should inherit from.
* [x] Added `add` and `remove` events to all layers.
* [x] Added `remove` method to layers and controls (`marker.remove()` is now equivalent to `map.removeLayer(marker)`).
* [x] Added `pane` option to all layers that can be changed (e.g. you can set `pane: 'overlayPane'` to a tile layer).
* [x] Added `map` `createPane` method for custom panes. [#1742](https://github.com/Leaflet/Leaflet/issues/1742)
* [x] Simplified map panes &mdash; removed `objectsPane`, all panes are on the same level now.
* [x] Added `shadowPane` option to markers as well.
[...]

PR #3101, opened 2014-12-22
Vector docs

patrickarlt:
Documents relevant changes for the vectors refactor in https://github.com/Leaflet/Leaflet/issues/3098.

I'm making a conscious choice to not document `L.Renderer`. I doubt anyone will ever need to use it and very few people (probably no one) will extend it, so i dont think it needs to be documented, especially for a 1.0-beta1

I removed `L.MultiPolygon` and `L.MultiPolyline from the doc entirely in favor of more usage examples for `L.Polyline` and `L.Polygon` demonstrating the single and multi variants..

This should probably be reviewed by someone involved in the vector refactor, most likely @mourner @danzel or @jacobtoye.

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -2707,6 +2707,11 @@ map.fitBounds(polyline.getBounds());</code></pre>
 		<td><code><a href="#latlngbounds">LatLngBounds</a></code></td>
 		<td>Returns the LatLngBounds of the polyline.</td>
 	</tr>
+	<tr id="path-getcenter">
+		<td><code><b>getCenter</b>()</code></td>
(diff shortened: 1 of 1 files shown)


mourner:
Need renderer-related updates/examples after #3266, but going to merge this for now.

ValentinH:
Thanks for the current documentation. Just want to know if you are planning to write a documentation about creating a custom Path object with the new version (or if there is already something like this)? Indeed, I had written an ArcPolyline plugin to be able to draw curved arc instead of lines but it is broken with the latest version.

PR #3107, opened 2014-12-29
GridLayer and pane docs + misc layer docs

patrickarlt:
This covers some more remaining layer and GridLayer items in https://github.com/Leaflet/Leaflet/issues/3098 including

* Docs for GridLayer
* Panes and pane management
* Reorganize shared methods for events, popups and layers

This doesn't remove a lot of documentation. For example TileLayer still retains all its options and method references even though half of them are in grid layer. I'm fine with removing them but that can be worked out in review.

@mourner for review

### DIFF
--- a/reference-tpl.html
+++ b/reference-tpl.html
@@ -45,7 +45,7 @@
 		<th>Description</th>
 	</tr>
 	<tr>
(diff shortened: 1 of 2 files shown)


mourner:
Rebased/merged with some cleanup (removed duplicate methods in various layers that now live in shared methods). Needs some more duplicate cleanup probably but good for now.

PR #3109, opened 2014-12-29
Improve popups for FeatureGroup

patrickarlt:
This PR makes the 2 following improvements for `FeatureGroup`

1. `FeatureGroup` no longer calls `bindPopup` on each of its layers. It now uses the built-in popup methods and the single instance of `Popup` from `Layer`.
2.  `bindPopup` now can also accept a function in addition to HTML or string as the first argument. This function is passed the layer (or source of the popup). This mean that constructing a unique popup for each layer in a large `FeatureGroup` no longer requires iterating over each feature OR storing each features content in memory.

Both of these changes mean that doing this...

```js
tenThousandFeatures.eachLayer(function(layer){
  layer.bindPopup('A unique template for this layer');
});
```

Can now be done like this...

```js
tenThousandFeatures.bindPopup(function(layer){
  return 'A unique template for this layer';
});
```

And everything will be handled on the fly. This has the potential to lead to huge memory reductions depending on how big the popup content is and how many feature are in the collection.

In my own tests with 10k features and the following template I got about 200mb of memory reduction.

```js
layer._leaflet_id + 'Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec odio. Quisque volutpat mattis eros. Nullam malesuada erat ut turpis. Suspendisse urna nibh, viverra non, semper suscipit, posuere a, pede.</p><p>Donec nec justo eget felis facilisis fermentum. Aliquam porttitor mauris sit amet orci. Aenean dignissim pellentesque.'
```

leaflet_debug_page


patrickarlt:
@mourner All popup related code is now out of `FeatureLayer`! The conditional to figure of the params looks a bit messy to me but its better then having to override popup code in `FeatureGroup`.

 The interesting side effect of this is that `LayerGroup` now supports popups since it to inherits from `L.Layer`. Since the only difference between `L.LayerGroup` and `L.FeatureGroup` is event propagation does it make sense to just merge the 2 together into a single class?

patrickarlt:
@mourner don't merge this yet. I want to be able to make it so you can return DOM elements from the function as well a template string.

I'll also squash all the commits in this down to a single one.

src/layer/Layer.Popup.js
```diff
@@ -41,10 +41,39 @@ L.Layer.include({
...
+			}
+			layer = layer || this;
+			latlng = layer._latlng || layer.getCenter();
+		} else if (target instanceof L.Layer) {
+			layer = target;
+			latlng = layer._latlng || layer.getCenter();
+		} else {
+			layer = this;
+			latlng = target;
+		}
```
mourner: Maybe there's a bit shorter way to write this...

src/layer/Layer.Popup.js
```diff
@@ -41,10 +41,39 @@ L.Layer.include({
...
+
 		if (this._popup && this._map) {
-			this._map.openPopup(this._popup, latlng || this._latlng || this.getCenter());
+			this._popup.options.offset = this._popupAnchor(layer); // update the popup offset based on our layer
+			this._popup._source = layer; // update popup source
+			this._popup.update(); // update the popup (will update content if popup uses a function)
```
mourner: Any side effects to this for non-function popups compared to how it worked before?

src/layer/Layer.Popup.js
```diff
@@ -77,11 +106,26 @@ L.Layer.include({
...
+		if (this._popup && this._map && this._map.hasLayer(this._popup) && this._popup._source === e.layer) {
+			this.closePopup();
+		} else {
+			var popupTarget = e.layer || e.target;
+			this._popup.options.offset = this._popupAnchor(popupTarget);
+			this._popup._source = popupTarget;
+			if (typeof this._popup._content === 'function') {
+				this._popup.update();
+			}
+			this._map.openPopup(this._popup, e.latlng);
```
mourner: There's duplication with a chunk of `openPopup` code here.

patrickarlt:
@mourner 

* Cleaned the logic in `openPopup`
* Move some shared logic into `_setupPopup`
* Added some more checks to `debug/map/popup.html` for testing

src/layer/Layer.Popup.js
```diff
@@ -41,10 +41,28 @@ L.Layer.include({
...
+				layer = this._layers[id];
+				break;
+			}
+		}
+
+		if (target instanceof L.Layer) {
+			layer = target;
+		}
+
+		latlng = layer._latlng || layer.getCenter();
```
mourner: now target can't be a latlng

mourner:
Looks much better now! Also, lets fix the build.

patrickarlt:
@mourner passing `latlng` to `openPopup()` works again. Tests also pass by commenting out the check to see if `_togglePopup` is called. https://github.com/Leaflet/Leaflet/pull/3109/files#diff-2e78276da4ad76f0284cf939a0587d13R60

mourner:
Hmm, why did it stop passing if uncommented in the first place?

patrickarlt:
I don't have any idea why it stopped working. Its clear that the method IS firing because the next assertions passes since the popup is on the map AND if I put a `console.log` in `_togglePopup` I can see it log during that test but call count stubbornly remains 0.

I've never been a huge fan of testing if specific methods run during a test. I honestly feel like this simpler test might be better. It just checks for the open/closed popup state on click. This is the behavior we want tested not if `_togglePopup has run or not`.

```js
  it("toggles its visibility when marker is clicked", function () {
    var marker = new L.Marker(new L.LatLng(55.8, 37.6));
    map.addLayer(marker);

    marker.bindPopup('Popup1');
    map.options.closePopupOnClick = true;

    // toggle open popup
    marker.fire('click', {
      latlng: new L.LatLng(55.8, 37.6)
    });
    expect(map.hasLayer(marker._popup)).to.be(true);
    marker._togglePopup.restore();

    // toggle close popup
    marker.fire('click', {
      latlng: new L.LatLng(55.8, 37.6)
    });
    expect(map.hasLayer(marker._popup)).to.be(false);
  });
```

mourner:
@patrickarlt that makes sense. Looks good, going to merge.

mourner:
Last question, do you think it's better to toggle the popup than to open a new one when you click on a different point of a polygon or polyline? Toggling feels a bit weird, although maybe I'm just accustomed to the current behavior.

mourner:
And another issue: after this pull, clicking on another layer within a group moves the popup without fade out/in animations, while previously animations were there. You can see the effect on the `debug/map.html` page. Changing this is tricky, looking into it...

patrickarlt:
> Last question, do you think it's better to toggle the popup than to open a new one when you click on a different point of a polygon or polyline?

Personally I think this is fine. It means that Markers and vectors have the same behavior.

>  after this pull, clicking on another layer within a group moves the popup without fade out/in animations

I can look into this too if you want. I probably need to close the popup and then do the update then reopen it. Are thre any pitfalls here I should know about?

mourner:
The pitfall here is that after closing, you need to initialize popup container as a different element so that you can animate both the old one fading out and the new one fading in at the same time.

mourner:
The tricky thing here is that recreating the popup element is often undesirable — some apps may depend on `_content` being persistent between popups. Maybe we should leave it as it is... People who depend on this behavior would just iterate over layers and put popups individually.

You can see that simple changes like this can be tricky, and may break apps if not done very carefully... This why I try not to merge too soon, at least until I reviewed the changes thoroughly and considered all side effects.

mourner:
Here's a slight simplification of the changes (also removing the toggling behavior): https://github.com/Leaflet/Leaflet/commit/7650eaef520c8ed57e79830253e2182c504a5c16

patrickarlt:
@mourner Those changes look great! I was trying to avoid changing the signature of `openPopup()` but this looks way cleaner then my version.

2 questions...

1. Are we ok with changing the signature of open popup? I don't forsee too many breaking changes from this.
2. Are we ok with changing the toggling behavior? Clicking repeatably on a marker no longer toggles the popup.

If the answer to both of these is yes I'll merge your commits into this branch and flesh out the test suite for popups with functions and we can merge this in!

patrickarlt:
@mourner Also these changes from @maackle seems like a really good idea now https://github.com/Leaflet/Leaflet/pull/3059.

This means that I could now use `bindPopup` to get a unique popup on every feature in a feature collection but have manual control over opening them.

mourner:
@patrickarlt yes, making use of #3059 sounds good!

> Are we ok with changing the signature of open popup? I don't forsee too many breaking changes from this.

It's changed in a way that is fully compatible with previous signature. The new one is only used internally and I'm not even sure it should be documented. What do you think?

> Are we ok with changing the toggling behavior? Clicking repeatably on a marker no longer toggles the popup.

No, markers should be togglable. I thought they would keep this — will look into it.

patrickarlt:
> It's changed in a way that is fully compatible with previous signature. The new one is only used internally and I'm not even sure it should be documented. What do you think?

I think we should probably document it. Opening the popup on a specific layer would probably be a use case.

> No, markers should be togglable. I thought they would keep this — will look into it.

I'll take a look at this.

From my perspective here is what needs to happen for this to be merged.

* [ ] Merge in your changes from https://github.com/Leaflet/Leaflet/commit/7650eaef520c8ed57e79830253e2182c504a5c16
* [ ] Restore toggling of popups on markers
* [ ] New tests for functions as content
* [ ] Make sure all existing tests pass

On the togglablility of the popup, when i open the popup on a polygon and click on the polygon again should the popup open at the new location? or close? If it should open at the new location this means that popups for markers and vector layers have different behaviors.

mourner:
@patrickarlt yep, sounds right! Also, rebase on current master since it switched to ESLint and catches more code style errors now.

patrickarlt:
@mourner This is all ready to go now. I just want to write more tests for the new functionality. Existing tests all pass and its up to date with the rest of master.

I'm going to work on more tests for this and revisit the 1.0 docs tomorrow.

mourner:
@patrickarlt awesome, merged! Lets add tests separately.

PR #3117, opened 2015-01-07
Add leaflet-layer-overpass to plugins

k-nut:
I added a reference to the leaflet-layer-overpass library. If there is anything missing in the PR I will try to provide it as quick as possible.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -488,6 +488,15 @@ While Leaflet is meant to be as lightweight as possible, and focuses on a core s
             <a href="https://github.com/brianreavis">Brian Reavis</a>
         </td>
     </tr>
+	<tr>
+		<td>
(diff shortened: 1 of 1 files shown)


danzel:
This PR is conflicted, fixing that would be good thanks :)
Usually rebasing will fix this.

k-nut:
I did not figure out how to rebase. I merged the newly added commit instead. Are you okay with that or should I try again?

mourner:
It's OK, but I'd suggest learning to rebase as it's an invaluable tool! For your case, you can just do `git rebase origin/gh-pages`, then `git push --force`. It should remove "merge" commits automatically.

k-nut:
I figured learning about reabse would be a good idea. I think I more or less understand how it works now. Thank you for pushing me into learning something new! I tried to rebase and it seems to have worked now! :sparkles:

mourner:
Perfect!

PR #3129, opened 2015-01-13
Clean up tile loading

jfirebaugh:
Simplify some internals; no ticketed issues fixed with these changes.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -34,8 +34,6 @@ L.GridLayer = L.Layer.extend({
 		this._levels = {};
 
 		this._tiles = {};
-		this._loaded = {};
-		this._retain = {};
 		this._tilesToLoad = 0;
 
 		this._reset();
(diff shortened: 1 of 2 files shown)


mourner:
Makes sense, thanks! I'll also change `_tilesToLoad` to `_noTilesToLoad` that returns boolean, as we don't care about the amount, only whether everything loaded or not.

src/layer/tile/GridLayer.js
```diff
@@ -593,6 +589,14 @@ L.GridLayer = L.Layer.extend({
 
 	_animateZoom: function (e) {
 		this._setZoomTransforms(e.center, e.zoom);
+	},
+
+	_tilesToLoad: function () {
+		var result = 0;
+		for (var key in this._tiles) {
+			if (this._tiles[key].loading) { result++; }
```
mourner: typo, should have been `!this._tiles[key].loaded`. I'll fix.

PR #3141, opened 2015-01-14
Use appropriate loses instead of looses

kyletolle:


kyletolle:
Loose is for unscrewing jar lids.
Lose is for having something taken away. In this instance, the map has its focus taken away.

mourner:
Thanks :)

Issue #3053, opened 2014-11-24
Tiles don't always fade-in with the new tile loading logic


mourner:
After merging the new tile loading logic #2382, background panes that become the main panes don't fade in, so there's an impression of tiles not always fading in.

jfirebaugh:
The issue is that with [throttled pruning](https://github.com/Leaflet/Leaflet/blob/2ee1aa2c2365e0cf4404817b4d268d5ed788a653/src/layer/tile/GridLayer.js#L32) and [deferred removal](https://github.com/Leaflet/Leaflet/blob/2ee1aa2c2365e0cf4404817b4d268d5ed788a653/src/layer/tile/GridLayer.js#L234), tiles can remain in the DOM and get reused during quick zoom-in-zoom-out operations. When that happens, the [CSS transition](https://github.com/Leaflet/Leaflet/blob/2ee1aa2c2365e0cf4404817b4d268d5ed788a653/dist/leaflet.css#L142-L153) for loaded tiles is not trigged.

Even if pruning wasn't throttled, we don't want to remove the `leaflet-tile-loaded` class from background tiles immediately, because they need to be cross-faded with loading tiles from the current layer. The other thing that I tried was to remove and re-add the `leaflet-tile-loaded` class from existing tiles in the current layer at the end of `_updateLevels`, but it doesn't look like this triggers CSS transitions. You probably have to remove the class in one frame and then re-add it in another for the transition to animate, and that would add even more asynchronous complexity to tile loading, which I'd like to avoid.

My preference would be to switch to JS-based transitions, which would also allow us to fix [the _deferRemove hack](https://github.com/Leaflet/Leaflet/blob/2ee1aa2c2365e0cf4404817b4d268d5ed788a653/src/layer/tile/GridLayer.js#L234). But I know that you are resistant to that. So what do you suggest?

mourner:
Several years ago, I always assumed native transitions would be faster than animating manually frame by frame. For some cases, that is a valid assumption, but for others, it has proved me wrong. For example, I recently ditched transition-based panning animation. 

So if we try it out and measure, it may turn out that it works pretty good, especially considering that we can calculate each tile in the view in one running context, without having to keep lots of independent timers for each tile and try to handle hard to track race conditions. Lets make a simple naive implementation, just to evaluate performance, and if it works, implement it properly.

mourner:
[...]

PR #3142, opened 2015-01-14
Fade tiles with requestAnimationFrame rather than CSS

jfirebaugh:
See #3053

There is currently some flickering with this approach around the edges of the map during a quick in-then-out zoom. For example, with quick zoom in to z15 and out to z14 you see z14 tiles at the edges flicker. They are visible for an instant because they are retained while z15 tiles are transitioning in as the first zoom in completes, and then they get pruned while the zoom animation to z14 is running.

This makes me want to see whether GridLayer can be notified of the new zoom level at the _beginning_ of the zoom animation rather than the end, so that tiles at the new zoom level can potentially fade and zoom concurrently, rather than having a 250ms zoom animation followed by a 200ms fade animation. I think this would make zooming feel much smoother and snappier.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -139,7 +139,6 @@
 
 /* zoom and fade animations */
 
-.leaflet-fade-anim .leaflet-tile,
 .leaflet-fade-anim .leaflet-popup {
 	opacity: 0;
 	-webkit-transition: opacity 0.2s linear;
@@ -147,7 +146,6 @@
 	     -o-transition: opacity 0.2s linear;
 	        transition: opacity 0.2s linear;
 	}
-.leaflet-fade-anim .leaflet-tile-loaded,
 .leaflet-fade-anim .leaflet-map-pane .leaflet-popup {
 	opacity: 1;
 	}
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -100,27 +100,30 @@ L.DomUtil = {
 			el.style.opacity = value;
 
 		} else if ('filter' in el.style) {
+			L.DomUtil._setOpacityIE(el, value);
+		}
+	},
 
-			var filter = false,
-			    filterName = 'DXImageTransform.Microsoft.Alpha';
+	_setOpacityIE: function (el, value) {
+		var filter = false,
+		    filterName = 'DXImageTransform.Microsoft.Alpha';
 
-			// filters collection throws an error if we try to retrieve a filter that doesn't exist
-			try {
-				filter = el.filters.item(filterName);
-			} catch (e) {
-				// don't set opacity to 1 if we haven't already set an opacity,
-				// it isn't needed and breaks transparent pngs.
-				if (value === 1) { return; }
-			}
+		// filters collection throws an error if we try to retrieve a filter that doesn't exist
+		try {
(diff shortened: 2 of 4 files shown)


mourner:
Awesome work!!! 
Lets look into the GridLayer zoomstart thing. Currently I see lots of flickering when browsing casually, and not only at screen edges.
How do you feel about perfromance vs transitions? We should look how it compares with a bunch of tile layer overlays + fullscreen + on mobile.

jfirebaugh:
> Lets look into the GridLayer zoomstart thing.

Started looking at implementing doing that based on `zoomanim` event. Is there a way to get map bounds at a given zoom level and center (not necessarily current)?

jfirebaugh:
> Currently I see lots of flickering when browsing casually, and not only at screen edges.

Say more? So far that's the only case where I see flickering.

mourner:
@jfirebaugh reproducible when you zoom in to a new area (no cached tiles) and then zoom out immediately. The whole screen flashes with gray.

mourner:
Fiddled with this a bit more. Performance seems acceptable, and fading during zoom animation looks great. I'll try to see if we can make zoom out more pleasantly looking, and also optimize it a bit to run less frame requests.

jfirebaugh:
One issue with the current code is that when animated zoom starts, tiles are added based on the zoom and bounds of animation endpoint (correct) but then immediately pruned based on the current zoom (incorrect). The fact that _pruneTiles is throttled confounds this further. We need to not throttle _pruneTiles, and prune based on a calculated bounds and zoom (in the `_update` case), or specific parent/child tiles (when a tile finishes fading in).

A related issue is that when starting an animated zoom, say from z15 to z16, we should not immediately prune z15 tiles that will end up outside the viewport when the animation completes, because that will cause flickering at the margins. Those tiles should be pruned only _after_ the animation completes.

jfirebaugh:
Current issues:

* Over-eager pruning of margin tiles when zooming in (see previous comment)
* Over-eager pruning of parent tiles when a tile fade finishes (should prune a parent when _all_ children are finished, not when first one finishes)
* Flickering when zooming in and out. Say z15 ⇢ z16 ⇢ z15. On the zoom in, z16 tiles start fading in. On the zoom out, z15 tiles get reset and start fading in. But now both z16 and z15 are fading at the same time, so you see through to the background.

mourner:
@jfirebaugh much better now! I also optimized performance by not requesting redundant frames (so number of times setOpacity is called dropped dramatically).

The third point (flickering) is still present but not on z15 -> z16 -> z15 but now on z15 -> z16 -> z17 (doesn't flicker on first zoom but flickers on subsequent ones), also apparent on longer fade in.

mourner:
Also `flyTo` is now broken. :( Doesn't load any tiles until the animation finishes.

mourner:
Maybe I was overeager to thorttle pruning... There are not too many tiles to iterate over usually, and pruning should only happen on tile fade-in & moveend probably, so not too often.

NeoRaider:
This pull broke overscaling. I've bisected the issue to commit `bbf32ebfcdb0d24d187dbb54eecbea367fd7b72d`.

mourner:
@NeoRaider thanks for investigating! We'll look into it.

PR #3147, opened 2015-01-15
BUG: ._tryAnimatedPan() lies...

cmlaidlaw:
... to .setView() about whether or not an animation was initiated.

This causes downstream problems like ._tilesToLoad being NaN instead of a number since .setView() returns early before ._resetView() is never called.


mourner:
Thanks, makes sense!

PR #3180, opened 2015-01-29
Added L.Path.Drag

w8r:
It's a dragging interaction support(extension) for vector shapes. Consistent with Marker.Drag interaction and events. It also adds matrix transformation support to SVG/VML renderers, which can be used for other transformations - skew/scale/etc. As the matter of fact, I think this particular thing could make a good part of Leaflet core functionality.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -1495,6 +1495,17 @@ While Leaflet is meant to be as lightweight as possible, and focuses on a core s
 			<a href="https://github.com/tinuzz/">Martijn Grendelman</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/w8r/Leaflet.Path.Drag">Leaflet.Path.Drag</a>
+		</td>
+		<td>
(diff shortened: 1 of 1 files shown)


mourner:
Not necessary in the core. It's extremely rare to need vector shape dragging, and things like this are usually covered by sophisticated drawing/editing plugins like Leaflet.draw.

w8r:
No, not the dragging, I agree, I meant the [SVG/VML matrix transformations](https://github.com/w8r/Leaflet.Path.Drag/blob/master/src/Path.Transform.js), on top of which things like that could be built. But of course, it's up to you to decide. Thanks for the consideration

mourner:
@w8r oh wow, didn't notice that! I think this would work well as a separate plugin for future Leaflet 1.0 version , with Canvas support baked in too.

w8r:
Ok, thanks, I'll look into it and will try to have it ready then. Good point about canvas renderer. I've already proposed that as a part of Leaflet.draw as well.

Issue #2214, opened 2013-11-20
fill-pattern


fab1an:
I would like to be able to fill GeoJson-Elements with a pattern like seen here: http://www.carto.net/svg/samples/patterns.shtml

pyrog:
@fab1an :+1: 

There are some issues in [HOT OSM Task Manager](http://tasks.hotosm.org/about) (Humanitarian OpenStreetMap Team)
* hotosm/osm-tasking-manager2#432
* hotosm/osm-tasking-manager2#500 and [specific comment](https://github.com/hotosm/osm-tasking-manager2/issues/500#issuecomment-67761569) about **[color blindness](http://vischeck.com/daltonize/)**

Even with this trick, the tiles over the map could be more readable with fill pattern :wink: 

See http://www.etre.com/tools/colourblindsimulator/

Without color blindness
![without color blindness](https://cloud.githubusercontent.com/assets/2158081/5591899/5e7c2f3a-9199-11e4-88be-899b98410e46.jpeg)

Protanopia
![protanopia](https://cloud.githubusercontent.com/assets/2158081/5591900/65b9b38a-9199-11e4-9208-3c07c9b3daea.jpeg)

Deuteranopia
![deuteranopia](https://cloud.githubusercontent.com/assets/2158081/5591901/6b870b5a-9199-11e4-9d54-70bb4d0fd2b8.jpeg)

Tritanopia
![tritanopia](https://cloud.githubusercontent.com/assets/2158081/5591902/71800caa-9199-11e4-8027-ca95cf472e0c.jpeg)

teastman:
I found we needed this feature for our product as well.  I took a stab at implementing it and have made a pull request.  If anyone can take a look and let me know what you think.  I put some example code uses here http://jsfiddle.net/ehucL6v5/ but I have no idea how to link my fork of leaflet.js in jsfiddle to show the example working.

pyrog:
In external resource tab, remove leaflet.js and add a link to your fork ?
![external ressource](https://cloud.githubusercontent.com/assets/2158081/5964690/682acbd6-a7f6-11e4-8da8-59899324b130.png)

teastman:
Ok, I've uploaded the initial code for a plugin to handle this.  I'm still adding more shapes and default patterns.  But pretty well any pattern can be created using the PatternPath shape.  [Leaflet.pattern](https://github.com/teastman/Leaflet.pattern)

mourner:
[...]

PR #3181, opened 2015-01-29
Added support for fill patterns to be used in paths.

teastman:
I've attempted to add support for fill patterns [Issue #2214](https://github.com/Leaflet/Leaflet/issues/2214)

This is my first attempt at contributing to Leaflet core, so let me know if I did anything wrong, I haven't written any tests for the feature yet.

Usage example can be seen here http://jsfiddle.net/ehucL6v5/ I have no idea how to link my fork of leaflet.js in jsfiddle to show the example working though.

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -130,6 +130,23 @@ var deps = {
 		desc: 'Circle overlays (with radius in meters).'
 	},
 
+	Pattern: {
+		src: ['layer/vector/Pattern.js'],
+		desc: 'Fill Patterns for use in Paths.'
+	},
+
+	Stripes: {
+		src: ['layer/vector/Stripes.js'],
+		deps: ['Pattern'],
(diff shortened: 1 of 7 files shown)


mourner:
Hey, thanks for the pull request! While this is a very useful feature, it has just too many code to consider including it in Leaflet code given that the pattern use case isn't too common. But it would make for a wonderful plugin! What do you think?

teastman:
Yeah plugin should work just fine.  I'll recode it as one, I was about to code it to work with 0.7.3 as well anyways.  Thanks for taking a look.

mourner:
Thanks! Looking forward to it.

teastman:
I've created a repo for a plugin to handle patterns and put it up at https://github.com/teastman/Leaflet.pattern
It's currently designed to work with 0.7.3

Issue #2757, opened 2014-06-24
Jumpy scrolling with touchpad (vertical/horizontal double finger scroll).


ChALkeR:
When using a touchpad with double-finger scroll enabled for both vertical and horizontal directions, the scroll becomes jumpy in some cases.

This happens when the user moves his fingers not following a straight vertical line, but a bit in horizontal direction. For example, left and bottom cause opposite scrolling, so left-bottom movement becomes jumpy.

This could be solved simply by ignoring horizontal mouse scroll events (like google maps do).

mourner:
Hey, can you check if https://github.com/Leaflet/Leaflet/pull/3653 works?

ChALkeR:
Sorry, missed this somehow.
I can confirm that this is fixed in the recent 1.0 beta.

IvanSanchez:
@ChALkeR beta2 was released 2015-oct-14, and the [commit fixing horizontal scroll](https://github.com/Leaflet/Leaflet/commit/a969d204766a1275ac362f76e64358d5e920194b) was merged 2015-oct-20. Please use `master` instead of beta2 for this.

ChALkeR:
Huh? I was under an impression that I was checking beta2. Will recheck later today.

IvanSanchez:
@ChALkeR Exactly my point: the bug is present in beta2 but fixed in `master`.

ChALkeR:
@IvanSanchez What do you mean? I stated above that I didn't observe this bug anymore.

ChALkeR:
Ah, yes. I was actually checking `master`, not beta2.

Issue #2732, opened 2014-06-10
Mouse wheel scrolls too much in Firefox (custom delta)


moneytoo:
Scrolling using mouse wheel in Firefox isn't level by level but several zoom levels at once. It is caused by custom wheel delta which results in high values returned from ``getWheelDelta()`` (``e.detail`` was 18 in my case). It is because of:

1. Settings panel for my mouse (Logitech SetPoint) contains settings "Scrolling Size" which is set to "3 Lines by default" but I changed it to "6 lines".
2. To further increase scrolling I increased Firefox property ``mousewheel.default.delta_multiplier_y`` from 100 to 300.

I do not think this is something desirable. Both Google Maps and local maps (mapy.cz) work even with my settings. I have no idea why delta is even read but wouldn't moving from obsolete ``DOMMouseScroll`` to ``wheel`` solve this? https://developer.mozilla.org/en-US/docs/Web/Reference/Events/wheel

mourner:
Can you check if https://github.com/Leaflet/Leaflet/pull/3653 fixes the issue?

CramericaIndustries:
I can test it if you tell me how I do get that version. I've no idea how to check it out from Git and how to build it. And what tools I do need to for that.
If your fix is already applied in the master branch, then the bug is still there. (tested with this: http://leaflet-cdn.s3.amazonaws.com/build/leaflet-master.zip on FF 39.0)

mourner:
Try with this JS: https://www.dropbox.com/s/r0ftoe11q8yow2w/leaflet.js?dl=0

CramericaIndustries:
Thanks for the link. I tested it and it works fine now on FF 39.0. However the problem appears in Chrome now (tested with v43.0.2357.134 m).

IvanSanchez:
Closing in favour of #3653.

bluenote10:
I don't think #3653 has resolved the issue, reopen?

I'm using Firefox 68.0.2 and I have increased my `mousewheel.default.delta_multiplier_y`, because I prefer an overall faster scrolling speed:

![image](https://user-images.githubusercontent.com/3620703/64210021-6b326780-cea2-11e9-8490-72084f87054e.png)

This is how leaflet behaves on a single mousewheel up/down, it is basically unusable for me:

![Peek 2019-09-03 23-25](https://user-images.githubusercontent.com/3620703/64210139-c8c6b400-cea2-11e9-9e3f-85fb618451d5.gif)

[...]

Issue #3178, opened 2015-01-28
Map idle event


amnesia7:
I'm posting this partly for confirmation that I've done it right and partly in case anyone else may need it in future or even if it is something that could/should be added to leaflet.

I was having trouble with `moveend` or `dragend zoomend` calling the results function too soon which prevented the user from being able to do more than a single move (drag or zoom) at once (unless they were quick) before it would want to run the function so, taking inspiration from google maps,  I've tried to replicate the idea of the `idle` event. Rather than run the function directly on `moveend` or `dragend zoomend` it allows me to wait for user interaction to stop (500ms since last movement in this case) before running the function.

```js
// declare timeout function name
var timeoutHandler;
...
// initialize map as normal
var map = L.map('map_canvas', {
  ...
});
// add event listener to map for movement
map.on('zoomend dragend', mapMoveHandler);
map.on('idle', mapIdleHandler);
...
function mapMoveHandler() {
  // cancel any timeout currently running
  window.clearTimeout(timeoutHandler);
  // create new timeout to fire idle event after 500ms (or whatever you like)
  timeoutHandler = window.setTimeout(function() {
    map.fire('idle');
  }, 500);
}
...
function mapIdleHandler() {
  // run some function when map idle to get results or update markers or something
  ...
}
```

amnesia7:
The 500ms for the timeout should probably actually be a calculation based on:
```js
((inertiaMaxSpeed / inertiaDeceleration) * 1000) + inertiaThreshold + 1
```
this way the map idle will be based on inertia values for dragging (converted to ms as required) + 1 to make sure that it is definitely complete.
This would mean that by default it needs to be (using the touch device `inertiaThreshold`) `((1500/3000)*1000)+32+1 = 532`

mourner:
@amnesia7 unlike `dragend`, `moveend` shouldn't be fired when you release a panning move — it should be fired when the map actually stops. If there's a case when it's not, it's a bug. Did you try using `moveend` but with a smaller delay, say, 50ms?

amnesia7:
@mourner I think the issue I had with `moveend` was that it was fired for `resize` as well as `dragend` and `zoomend` and I didn't want the page to re-run the search if the user flipped their phone or tablet from portrait to landscape which I assumed would cause resize to be fired.

I can change my page to just run the search function based on `moveend` and applying `trackResize: false` to the map seems to prevent `moveend` being called when the screen size changes - tested by resizing desktop browser anyway. 
[...]

PR #3204, opened 2015-02-06
Linear zooming animation with overhauled scroll behavior

### LINKED ISSUE #2154: Zoom issues with Apple trackpad (Magic Mouse)
Zooming with the scroll gesture causes very jumpy behavior between zoom levels. The calls to `_performZoom` seem to add up, meaning the map keeps zooming after the user has stopped scrolling.

Basically the Apple trackpad creates A LOT of calls to `_onWheelScroll` and because of inertial scrolling, continues to call it after the user has actually stopped scrolling. For my uses, changing the value 

`var left = Math.max(40 - (+new Date() - this._startTime), 0);`

to something more like:

`var left = Math.max(200 - (+new Date() - this._startTime), 0);`

works decently. It makes the apple mouse a little too sensitive (harder to move between small numbers of zoom levels), but seems to get rid of the jumping effect. Any ideas of a better solution?

mourner:
This gives Leaflet zoom animation a very different feel, and should also solve all our problems with scroll wheel behavior. There was a similar attempt earlier, but this time, after recent tile loading and animation improvements, it feels much nicer. Tested both mouse and Macbook trackpad. Closes #2154, #2732, #2757.

@ChALkeR @moneytoo @CramericaIndustries @ekelleyv @robrecord @rossmeissl @mearsheimer @schovi @jfirebaugh @danzel I'd appreciate your feedback guys. How does the zoom feel in this branch?

Note: pretty broken in Safari (at least for me on OS X) atm.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -156,10 +156,10 @@
 	will-change: transform;
 	}
 .leaflet-zoom-anim .leaflet-zoom-animated {
-	-webkit-transition: -webkit-transform 0.25s cubic-bezier(0,0,0.25,1);
-	   -moz-transition:    -moz-transform 0.25s cubic-bezier(0,0,0.25,1);
-	     -o-transition:      -o-transform 0.25s cubic-bezier(0,0,0.25,1);
-	        transition:         transform 0.25s cubic-bezier(0,0,0.25,1);
+	-webkit-transition: -webkit-transform 0.2s linear;
+	   -moz-transition:    -moz-transform 0.2s linear;
+	     -o-transition:      -o-transform 0.2s linear;
+	        transition:         transform 0.2s linear;
 	}
 .leaflet-zoom-anim .leaflet-tile,
 .leaflet-pan-anim .leaflet-tile {
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -3,8 +3,7 @@
  */
 
 L.Map.mergeOptions({
-	scrollWheelZoom: true,
-	wheelDebounceTime: 40
+	scrollWheelZoom: true
 });
 
 L.Map.ScrollWheelZoom = L.Handler.extend({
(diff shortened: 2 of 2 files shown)


rossmeissl:
@wboykinm would you please take a look at this?

wboykinm:
Actually it seems like it works best on safari with OSX :smiley_cat: 

## Chrome
linearzoom2_chrome

## Firefox
linearzoom2_ff

## Safari
linearzoom2_safari

mourner:
@wboykinm so what do you think vs the old behavior?

wboykinm:
@mourner I think the feel of the linear zoom is great, but I can't get the trackpad to behave right on chrome. The zoom gesture always double-zooms, no matter how faintly I touch it. The scrollwheel on my external mouse single-zooms as expected.

linearzoom2_chrome2

ekelleyv:
I like the linear zoom a lot. However, I am having the double-zoom issue with apple trackpad as well. Also, it seems to break image overlays (I think it is the same issue @wboykinm pointed out). I am also on Chrome.

image

mourner:
@wboykinm @ekelleyv thanks for the feedback! I think I know where the double-zoom comes from so I'll fix this. ImageOverlay seems like an unrelated regression though.

amnesia7:
Would this help with the issue I've pointed out here https://github.com/Leaflet/Leaflet/issues/3178#issuecomment-72753838 with respect to events called when zooming (in my case I was using the scroll on touchpad) or have I already missed a fix for that that's already in master?

danzel:
Seems awesome in Chrome on Windows.
IE11 on Windows does one zoom and then stops a moment and then does another.

This is me doing a big zoom out, then a bit zoom in:
zoom

ChALkeR:
How could that solve #2757?
The problem in #2757 is that Leaflet reacts to both horizontal and vertical
zoom events, not only vertical, as it should.

2015-02-08 21:23 GMT+03:00 Dave Leaver <notifications@github.com>:

mourner:
I now see that it's pretty much unusable on trackpad, so needs some more work...

jonkoops:
@mourner Is this still something we'd want to introduce or can this PR be closed?

mourner:
Let's close for now — not sure if we can reintroduce it in a buggy-free, compatible way.

PR #3215, opened 2015-02-06
Followups for zoom animation changes

jfirebaugh:
* Stricter selectors for will-change properties
* Remove obsolete unloadInvisibleTiles option
* Ensure tiles have 0 opacity when first painted

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -139,7 +139,7 @@
 
 /* zoom and fade animations */
 
-.leaflet-tile {
+.leaflet-fade-anim .leaflet-tile {
 	will-change: opacity;
 	}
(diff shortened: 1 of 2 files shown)


dist/leaflet.css
```diff
@@ -152,7 +152,7 @@
 .leaflet-fade-anim .leaflet-map-pane .leaflet-popup {
 	opacity: 1;
 	}
-.leaflet-zoom-animated {
+.leaflet-zoom-anim .leaflet-zoom-animated {
```
mourner: leaflet-zoom-anim class is only added during animation, would this kinda defeat the purpose (according to that article that says you don't want to put the property immediately before running the effect)?

Issue #2076, opened 2013-10-07
Partial scroll wheel zoom disabling


Zverik:
It was suggested that zooming by mouse wheel could work when `scrollWheelZoom` is set to false, but a modifier key (e.g. shift) is pressed. This way page scrolling with mouse wheel is not affected, but users still have a way to quickly zoom out.

ValentinH:
Any update about this feature? It would be really cool to only start the zoom when the scroll starts on the map...

Uyara:
Any update? I really would like to use this.

Nycroth:
Not a patch, sorry. However, here's a no underscore version of a debounce for a leaflet map assuming `map = L.map`. I'm sure this can be added nearly verbatim to leaflet.js with an option to toggle it.

```javascript
// Only enable scrolling if the user hasn't scrolled the page in 1 second
document.addEventListener("scroll", function() {
        map.scrollWheelZoom.disable()
        var timeOut = null;
        if(timeOut != null){
                clearTimeout(timeOut);
        }
        timeOut = setTimeout("map.scrollWheelZoom.enable()", 1000);
});
}```

DanielHeath:
@Nycroth That's only on the map container - if the debounce applied to the whole document instead (and only zoomed if the event target was in the map container) it would solve the issue.

vincent3569:
hi
is there some news on that issue?
the googlemap behavior on this issue is OK for me: if mousse wheel zoom is disabled, so ctrl + mousse wheel in enabled and users can zoom in or zoom out when mouse is over a map.

mguinness:
Reading through the various issues what is the consensus on the modifier key, is shift preferred over ctrl?

Keeping similarity with Google Maps might not be a bad idea, i.e. use `scrollWheelZoom.disable()` which can be overridden with ctrl+scroll.

Erwane:
Hi,
google maps display a backdrop with a text "Control + mousewheel to zoom".
Maybe it's possible to do the same with leaflet ?
Thanks.

IvanSanchez:
Already implemented by the https://github.com/elmarquis/Leaflet.GestureHandling/ plugin. No plans so far to implement this in the core.

PR #3221, opened 2015-02-11
"onFocus" option to scrollWheelZoom

juliensoret:
I've been confronted to this problem. My website content is scrollable and I had to set scrollWheelZoom on false. However, when interacting to the map, I'm frustrated about not being able to use the scroll wheel to (de)zoom as I always do on other maps.
I found this issue : https://github.com/Leaflet/Leaflet/issues/2076 and wanted to show one idea.
I add a "onFocus" option to scrollWheelZoom parameter which triggers scrollWheelZoom only when the map is on focus.

Here is a demo in which I already use this modification : http://kara.62.free.fr/scrollonfocus.html

(Sorry if my code is bad, I'm very new to javascript & git.)

### DIFF
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -67,4 +67,12 @@ L.Map.ScrollWheelZoom = L.Handler.extend({
 	}
 });
 
+L.Map.addInitHook(function () {
+	if (this.options.scrollWheelZoom === 'onFocus') {
+		this.options.scrollWheelZoom = false;
+		this.on('blur', function() { this.scrollWheelZoom.disable(); });
+		this.on('focus', function() { this.scrollWheelZoom.enable(); });
+	}
+});
+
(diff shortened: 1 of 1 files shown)


juliensoret:
Can someone tell me if I do it bad?

confiks:
This looks like good code to me. I would also like to have this feature merged into Leaflet.

IvanSanchez:
This is going to cause problems in IE due to #3314.

Also, this doesn't cover the case where a user focuses an element inside the map (e.g. pressing the Tab key until a marker is focused).

This is a good idea, but I'm afraid there are too many edge cases to merge this code right now. :-(

There is also #3581, which covers a similar problem. I'm not really sure about the ideal UX interactions on scrolling for these cases.

yohanboniface:
cf #2076

IvanSanchez:
Closing - this can be now achieved by enabling/disabling the scrollwheel map handler on the blur/focus DOM events.

PR #3224, opened 2015-02-11
Update GeoJSON.js

MuellerMatthew:
I modified the function to explicitly return the geoJSON properties. When using geoJSON's with different CRS's this function used to return the original CRS name but with geometry converted to WGS84 lat/lng format.  If the data was reused for another projection the locations would be in the wrong spots as the coordinates no longer matched the CRS.  Now this function just returns the lat/lng but without the CRS attached.

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -165,7 +165,11 @@ L.extend(L.GeoJSON, {
 
 	asFeature: function (geoJSON) {
 		if (geoJSON.type === 'Feature') {
-			return geoJSON;
+			return {
+				type: geoJSON.type,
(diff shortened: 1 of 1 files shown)


perliedman:
This sounds a bit specific for your use case to be included in Leaflet. Making this change, Leaflet will scrap all the GeoJSON's properties that it doesn't know about, which I think is a bad thing: other apps might be storing valuable information there, and Leaflet shouldn't touch or modify it.

I think a better solution would be to make sure that the code that's responsible for reprojecting the GeoJSON removes or corrects the CRS information before passing it on to Leaflet.

mourner:
Yeah, I agree with Per.

PR #3230, opened 2015-02-13
Do not try to remove drag class if marker as no icon

yohanboniface:
Since 4c46abe7810272704cc1a7ce0ccfb84c45456791 it's possible to have a marker off the map with dragging still enabled. We want to be able to disable it in this situation too.

Before 4c46abe7810272704cc1a7ce0ccfb84c45456791 removing a marker from the map was disabling dragging, so any subsequent call to `marker.dragging.disable` was not executed.


yohanboniface:
Thanks Vlad' :)

Issue #3219, opened 2015-02-09
Drag cursor showing with dragging set to false


psnowak:
When creating a map with { dragging: false }, the map still displays -webkit-grab cursor.

mourner:
Accepted, although low priority.

PR #3233, opened 2015-02-15
Fix cursor when dragging is set to false.

juliensoret:
As said in issue https://github.com/Leaflet/Leaflet/issues/3219 , the cursor is still "grab" by default even when dragging is set to false.

This commit change the css by removing default grab-cursor and this style is added when dragging is activated in Handler.

Demo here : http://kara.62.free.fr/cursor_issue/

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -192,7 +192,8 @@
 	cursor: auto;
 	}
 .leaflet-dragging .leaflet-grab,
-.leaflet-dragging .leaflet-grab .leaflet-interactive {
+.leaflet-dragging .leaflet-grab .leaflet-interactive,
+.leaflet-dragging .leaflet-marker-draggable {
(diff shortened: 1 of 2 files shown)


mourner:
Good approach, but doesn't work in all cases too — e.g. if you have map dragging disabled but marker dragging enabled, marker won't have grabbing cursors.

juliensoret:
Thanks you for your answer.
I forgot about the draggable marker and added to my demo. I also checked the code Marker.Drag.js and figured out that the marker already got a class when getting draggable. (Class I didn't find defined in css)
So I added the cursor to this class. If it was the only draggable item with map, it seems to works fine now.
PS : Travis CI says the build failed. But I didn't change JS and the build worked on my computer.

juliensoret:
Can someone tell me if I still miss something ?

PR #3234, opened 2015-02-16
update the url of the leaflet-tilejson plugin

paulovieira:
The current url is pointing to what seems to be an outdated repo. The original repo has been forked and now lives in the kartena account.


mourner:
Thanks!

PR #3243, opened 2015-02-21
Bridging any SVG library with Leaflet

akauppi:
The `ActiveOverlay` class provides support for any SVG vector graphics libraries to be used together with Leaflet, to create interactive "images" that pan and scale on top of the map just as `ImageOverlay` does for images (in fact, `ImageOverlay` can already be used to show inanimate SVG files).

Before this, there were limitations to working with SVG and Leaflet, together.

1. Leaflet itself uses SVG as a rendering platform, but only for lines, circles, rectangles, polylines and polygons (maybe something else). It does not expose the full SVG possibilities (nor should it, since it's also supporting canvas as another rendering platform).

2. Particular plugins for integrating with certain SVG libraries are available, s.a. for D3. It's difficult to select the right one, and there seem to be issues s.a. reorienting the SVG elements when zooming instead of simply changing the SVG's transformation, leaving the elements untouched.

`ActiveOverlay` is SVG library agnostic since it does not touch in any way the internals of the SVG DOM, in any way. It just serves as a cradle for such DOM, tying it to Leaflet maps.

- SVG side units are in meters (so i.e. making a 10x10m rectangle is trivial)
- SVG bounds are expected not to be extended during runtime (following the modeling according to `ImageOverlay`)
- panning the map is possible via the "non-filled" areas of the SVG
- SVG elements can freely implement their own mouse / touch interactions or animations

There are two files to demo the use cases:

`debug/map/active-overlay.html` - this shows an SVG crocodile picture (the Snap.svg mascot) and a black rectangle. The croc observes hover events and is draggable to a new position over the map.

croc

`debug/map/swiss-cities.html` - this shows adding data by Lat,Lng coordinates and is an adaptation from the D3 Leaflet plugin's demo. It shows that `ActiveOverlay` makes particular SVG library specific plugins unnecessary (also I think the performance is better; the D3 plugin calculated new positions for the elements when zoomed, making it less scalable).

swiss-cities

There are some questions among the source code as to how things should be properly done in Leaflet - I'm not sure.

This code can be implemented as a plugin, but given its generic usability, nice fitting in the Leaflet overlays model (i.e. being modeled by the existing ImageOverlay), I would first ask if it can be considered for inclusion in the main code base.

Permission from the Snap.svg and D3 plugin's authors need to be asked, to use their code (I'm on it).


akauppi:
Found a bug with regards to the `latLngToSvgPoint` (and back) conversions. It's fixed now, and there's code in the https://github.com/akauppi/Leaflet/blob/activeOverlay/debug/map/active-overlay.html file to check that things are behaving right. Before release (if Leaflet were to approve this PR) such debugging code can probably be removed, or moved to test suites.

As a side note, I observed weird behaviour of the Leaflet latLng `distanceTo` function (see these lines: https://github.com/akauppi/Leaflet/blob/activeOverlay/debug/map/active-overlay.html#L166-L189 ). It is giving consistently ~19% too big results for my sample map. Maybe I'm missing something, maybe it's a bug in Leaflet? Would someone have the time to look.

mourner:
@akauppi `distanceTo` is geodesic distance (taking curvature of Earth into consideration). It's not the same as Euclidian distance.

akauppi:
@mourner I'm not expecting them to be quite equal. However, here instead of 716 meters, `distanceTo` gives 849.90 meters (west-east). And instead of 488 meters, 579.95 meters. Those are way bigger differences than the curvatures would be able to explain.

akauppi:
The d34d630 commit has fix to the scaling issue, i.e. SVG units now are pretty close to 1.0m. I scale them by x ratio only; it can be made even more precise by scaling both x and y separately, but I don't believe that really matters.

There's some hickup with running `jake`, recently. I tried my best.

akauppi:
If this is considered to be included, I would suggest replacing the Croc SVG with Leaflet's own logo. And I still need to ask permission to use the D3 plugin's swiss-cities code.

akauppi:
One more bug with SVG scaling: dragging the croc is out of proportion. Will fix tomorrow.

mourner:
This is awesome work Asko, but it seems to me that it would look better as a plugin rather than as a part of the Leaflet core. It's a lot of code for a relatively uncommon feature, and it's also pretty easy to refactor into a plugin wired up externally.

akauppi:
Sure - thanks for the comments. Much of the "lines" comes from the croc SVG file, but you probably figured that out, anyways. I'll make it a plugin - please say if you reconsider placing it in the core at some point. The use case of it is making full-blown "apps" in SVG, over maps. Or things s.a. interactive charts that may change their looks based on a map location pointed elsewhere. I like to think of it as the Harry Potter's image overlay for Leaflet.

akauppi:
Woa, my merging with upstream `master` seems to have brought a lot of unrelated stuff to this PR. But you were probably able to figure that out.

akauppi:
Changed the title to better describe the benefits in the PR list.

mourner:
Please release it as a plugin! I don't want to bring this kind of complexity to Leaflet core, but it still looks too useful to pass up, so I'd love to see this in a separate repo and easy to plug in.

PR #3250, opened 2015-02-24
Add touchPan option for disabling touch pan

ljbade:
Not quite ready since the static START was referenced outside of Draggable.js

The documentation also needs to be updated. Where is this kept?


IvanSanchez:
Not updated since a long time ago, and we changed lots of touch stuff due to the Chrome 55 `PointerEvent`s. Closing for now.

PR #3255, opened 2015-02-26
Use standard .eslintrc config file naming

yohanboniface:
This PR renames the `eslintrc.json` files to [standard](http://eslint.org/docs/configuring/) `.eslintrc`, plus the `.eslintrc` file from `build` is moved to the root of the repository, so it's active also for `src/`.

The aim is to have those configuration files automatically detected by ESLint, including by editor plugins.

Thanks for reviewing! :)

Yohan


mourner:
Generally I try to avoid putting too many files in the root, but I guess this is OK...

yohanboniface:
Thanks :)

PR #3264, opened 2015-02-27
Forward original event on drag

yohanboniface:
This PR adds forwarding of the original event when firing `drag` or `move` events.

Use case: handle ctrlKey/altKey, etc. while dragging for creating different UX scenarios (eg. do not snap if `ctrlKey` is pressed while dragging).

Thanks for reviewing :)

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -99,13 +99,12 @@ L.Draggable = L.Evented.extend({
 		this._moving = true;
 
 		L.Util.cancelAnimFrame(this._animRequest);
-		var animationCallback = function () {
-			this._updatePosition(e);
-		};
-		this._animRequest = L.Util.requestAnimFrame(animationCallback, this, true, this._dragStartTarget);
+		this._lastEvent = e;
+		this._animRequest = L.Util.requestAnimFrame(this._updatePosition, this, true, this._dragStartTarget);
 	},
 
-	_updatePosition: function (e) {
+	_updatePosition: function () {
+		var e = {originalEvent: this._lastEvent};
 		this.fire('predrag', e);
 		L.DomUtil.setPosition(this._element, this._newPos);
 		this.fire('drag', e);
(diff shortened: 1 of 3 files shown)


src/dom/Draggable.js
```diff
@@ -99,13 +99,16 @@ L.Draggable = L.Evented.extend({
 		this._moving = true;
 
 		L.Util.cancelAnimFrame(this._animRequest);
-		this._animRequest = L.Util.requestAnimFrame(this._updatePosition, this, true, this._dragStartTarget);
+		var animationCallback = function () {
+			this._updatePosition(e);
+		};
+		this._animRequest = L.Util.requestAnimFrame(animationCallback, this, true, this._dragStartTarget);
```
mourner: You can use `L.bind(animationCallback, null, e)` instead. Or `this._lastEvent` to avoid closure entirely.

yohanboniface: Good point!
Which one do you prefer?

src/layer/marker/Marker.Drag.js
```diff
@@ -58,10 +58,11 @@ L.Handler.MarkerDrag = L.Handler.extend({
 		}
 
 		marker._latlng = latlng;
+		e.latlng = latlng;
 
 		marker
-		    .fire('move', {latlng: latlng})
-		    .fire('drag');
+		    .fire('move', e)
```
mourner: I don't like modified DOM event being passed as event data... In a similar situation in other place of Leaflet code, original event is passed as `e.originalEvent`.

src/dom/Draggable.js
yohanboniface: Latest (`this._lastEvent`) may be more readable, given that we are already passing a buntch of argument to `requestAnimFrame`.

src/layer/marker/Marker.Drag.js
yohanboniface: I originally done that, but then changed my mind, finding this version lighter, but I'll do that then :)

yohanboniface: Also, as we are doing two forwards, I'm not sure at which step we should introduce the `originalEvent` key ;)

mourner: Probably at the start.

yohanboniface:
Anything more I need to clean? :)

Issue #3268, opened 2015-03-02
Severe regression on iOS after tile anim refactor


mourner:
@jfirebaugh commit bbf32ebfcdb0d24d187dbb54eecbea367fd7b72d introduced a severe performance regression on iOS (at least on my iPhone). Pinch zoom is pretty much unusable. Probably a bug — this is apparent while you pinch, even before releasing the fingers for an animation.

mourner:
Hmm, while this could be fixed by disabling `GridLayer` `_reset` routine while doing pinch-zoom (seems to work), there's a different issue: animations have weird shifts even when you use the zoom buttons and double tap only. Reproducible pretty much immediately when you start pushing the zoom buttons.

jfirebaugh:
Didn't get a chance to test on non-simulator, but should be better now.

mourner:
The latter is fixed for now by disabling fade-in-during-zoom animation in mobile WebKit, which works better than unstable reqAnimFrame hacks with markers going out of sync.

Issue #2478, opened 2014-02-21
Android 4.1 crash on zoom


JerryBels:
Hello guys !

So I'm starting a new app based on Cordova, and wanted to use Leaflet for my maps - either via Mapbox or directly with OSM. I encountered an issue when testing the most basic example, on Android 4.1 only, the app crashes when zooming in. I looked for many hours all over the internet, and found solutions that either didn't work or even sometimes I didn't understand ( someone said << android23 = ua.search('android [23]') !== -1 || ua.search("android 4.0") !== -1 || ua.search("android 4.1") >> - where am I supposed to put that ?!)

Note that the issue isn't even linked to the application itself, because when launching the stock Android browser and going to the leaflet website or mapbox website, when trying to zoom in on the available examples, the browser crash ! On Chrome for Android all is fine, though.

Sooo there it is, how to prevent that crash ?

Thanks !

ilvalle:
@mourner is it visible to add this issue to https://github.com/Leaflet/Leaflet/milestones/1.0-beta1 ?

mourner:
@ilvalle not sure if it's the same bug as #2693, can you verify any of the proposed fixes?

ilvalle:
I changed Browser.js as
```
android23 = ua.search('android [23]') !== -1 || ua.search('android 4.1') !== -1
```
and loaded leaflet with
```
window.L_DISABLE_3D = true;
```
still crashing after a very few zoom.
With the following, it seems that the crash happens when you reach max-zoom 
```
#map {
    transform: none;
}    
```
In general the app crashes when is loading tiles and you try to zoom at the same time.
Hope it helps

fab1an:
Please have a look at my comment [1]. I did a lot of debugging on this and all of the freezes could be traced back to browsers crashing when `css-transform` was set to it's previous value.

[1] https://github.com/Leaflet/Leaflet/issues/2693#issuecomment-46053713

ilvalle:
I followed @fab1an's suggestion
I placed in https://github.com/Leaflet/Leaflet/blob/4c0f02879470cb4c1781db1739138d286040f0b4/src/dom/DomUtil.js#L141
```
pos.x = pos.x + (Math.random() / 10000);
```
it isn't so pretty, but it works. For me, the last thing is to apply this trick only when an android41 is detected.

fab1an:
[...]

PR #3270, opened 2015-03-02
Fix zoom animation freeze race condition

### LINKED ISSUE #2693: Map freeze problem, but unlocks after marker click
Hello

I'm using leaflet 0.7.2 in a windows RT application, developed with the use of the WinJS library. I'm also using angularjs as the basis for the application. The map freezes 1/20 times when i zoom in or out. When i press a marker on the map, the map unfreezes and becomes responsive once more. When i use the leaflet-src and debug the problem i notice, that when the problem occurs the "_catchTransitionEnd"  isn't called. When i press a marker the "_catchTransitionEnd" is called and the map unfreezes. Does anyone know what is causing this problem, and better still how to fix it? a quick workaround would be great, because of the deadline i'm on.

Any help would be greatly appreciated.

Thanks in advance

Ian

mourner:
In theory, this closes #2693 and #2478. Should be tested after fixing #3268 though, since pinch-zoom on iOS is currently broken.

### DIFF
--- a/src/map/anim/Map.ZoomAnimation.js
+++ b/src/map/anim/Map.ZoomAnimation.js
@@ -34,7 +34,15 @@ L.Map.include(!zoomAnimated ? {} : {
 		this._panes.mapPane.appendChild(proxy);
 
 		this.on('zoomanim', function (e) {
+			var prop = L.DomUtil.TRANSFORM,
+				transform = proxy.style[prop];
+
 			L.DomUtil.setTransform(proxy, this.project(e.center, e.zoom), this.getZoomScale(e.zoom, 1));
+
+			// workaround for case when transform is the same and so transitionend event is not fired
+			if (transform === proxy.style[prop] & this._animatingZoom) {
+				this._onZoomTransitionEnd();
+			}
 		}, this);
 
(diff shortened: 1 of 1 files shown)


src/map/anim/Map.ZoomAnimation.js
```diff
@@ -34,7 +34,15 @@ L.Map.include(!zoomAnimated ? {} : {
 		this._panes.mapPane.appendChild(proxy);
 
 		this.on('zoomanim', function (e) {
+			var prop = L.DomUtil.TRANSFORM,
+				transform = proxy.style[prop];
+
 			L.DomUtil.setTransform(proxy, this.project(e.center, e.zoom), this.getZoomScale(e.zoom, 1));
+
+			// workaround for case when transform is the same and so transitionend event is not fired
+			if (transform === proxy.style[prop] & this._animatingZoom) {
```
uniphil: should that be an `&&` in the if?

danzel: @mourner ^^

mourner: Yep, good catch

PR #3274, opened 2015-03-04
Add TileLayer.Gigapan

namrehs:
TileLayer.Gigapan allows Leaflet to work with pyramidal quadkey based tiles created by Gigapan Stitch (locally) and images on the gigapan.com website.


mourner:
Thanks!

Issue #3159, opened 2015-01-21
Zoom In/Out loses keyboard focus


webstandards:
Accessibility Finding: On the map, when a keyboard user action the Zoom In or Zoom Out, focus is removed from the map. Recommendation: Don't lose the focus.

http://www.w3.org/TR/WCAG20/#consistent-behavior
3.2.2 On Input: Changing the setting of any user interface component does not automatically cause a change of context unless the user has been advised of the behavior before using the component. (Level A)

So far I've verified against the master branch using the quick start html example. You can tab to the zoom in/out controls, however when you select one of the controls (by hitting enter) it will zoom in/out one level (good), however the focus on the zoom in/out button is totally lost and your keyboard focus is taken back to the start of the map. 

One observation: In the quick start example there is a Marker you can tab to focus and enter to open the Marker popup info, hit enter again and the popup info is closed and focus remains on the Marker. Just starting to dig through the codebase to see what the difference is between the keyboard handling of a Marker vs a Zoom button.

mourner:
You're right, this should be fixed.

mlegoff:
Hello,

I found the way to fix this bug. I just delete this line (69) on ControlZoom.js : .on(link, 'click', this._refocusOnMap, this); 
But it seems to me that it'a little bit simple. Maybe this line is used for something else ?
I make a little demo here : 
http://kara.62.free.fr/zoomfocus/

mourner:
Yeah, originally this was for #2148. I think the right solution is to detect if it was actually clicked by a mouse or a keyboard in the `_refocusOnMap` handler.

juliensoret:
Agree on checking click origin.
User who clicks with mouse doesn't require keeping focus on the button. He's more likely to use arrows as said in https://github.com/Leaflet/Leaflet/issues/2148.
While people who got to zoom by keyboard will surely understand they have to focus back on the map to use arrows.
Or I don't know the difficulty to make the arrows pressing handled by the map while giving back focus to the map.

webstandards:
Had my accessibility consultant review @mlegoff demo site looks good, and he also says ok to make the change only for keyboard users and leave current mouse behaviour alone.

Jakobud:
@mourner I could probably handle this one. Is it possible to pass the event into `_refocusOnMap` and if so, how do you determine if it was a keyboard click or mouse click for the event?

mourner:
@Jacobud that's what you would have to figure out :)

[...]

PR #3275, opened 2015-03-04
Prevent map refocus if control click event is from the keyboard

Jakobud:
Prevent map refocus if control click event is from the keyboard, which is determined/assumed by the click X,Y being at 0,0. There is definitely an assumption there that it is extremely unlikely that you would be clicking a control with your mouse at 0,0 (at the very top left of the screen).

Not sure if this is the best solution, but it does work.

### DIFF
--- a/src/control/Control.js
+++ b/src/control/Control.js
@@ -72,7 +72,7 @@ L.Control = L.Class.extend({
 	},
 
 	_refocusOnMap: function (e) {
-		// if map exists and event is keyboard event
+		// if map exists and event is not a keyboard event
 		if (this._map && e && e.screenX > 0 && e.screenY > 0) {
 			this._map.getContainer().focus();
 		}
--- a/src/control/Control.js
+++ b/src/control/Control.js
@@ -73,7 +73,7 @@ L.Control = L.Class.extend({
 
 	_refocusOnMap: function (e) {
 		// if map exists and event is keyboard event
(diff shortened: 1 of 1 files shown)


Jakobud:
for #3159

Jakobud:
Errors galore in the build! Not sure how to fix those :-P

src/control/Control.js
```diff
@@ -71,8 +71,9 @@ L.Control = L.Class.extend({
 		return this;
 	},
 
-	_refocusOnMap: function () {
-		if (this._map) {
+	_refocusOnMap: function (e) {
+		// if map exists and event is keyboard event
```
mourner: you probably meant "is not a keyboard event" :)

src/control/Control.js
```diff
@@ -71,8 +71,9 @@ L.Control = L.Class.extend({
 		return this;
 	},
 
-	_refocusOnMap: function () {
-		if (this._map) {
+	_refocusOnMap: function (e) {
+		// if map exists and event is keyboard event
+		if (this._map && (e.x > 0 && e.y > 0 && e.screenX > 0 && e.screenY > 0)) {
```
mourner: Try adding a check for truthy `e`, maybe it'll fix tests.

mourner:
Yeah, I don't know whether there's a more reliable way to detect this. Bummer... I think this will be good for now.

mourner:
:+1:

Issue #2982, opened 2014-10-29
Fix vector layers projecting/simplifying/clipping on every flyTo frame


mourner:

mourner:
Done in #3278

Issue #3272, opened 2015-03-03
IE7-9 graceful degradation for animations & fractional zoom


mourner:
In IE7-9, all API methods should never break. Fractional zoom levels should fallback to round ones. We can skip animations too.

mourner:
This issue is relatively easy to grab — the basic task is to make sure that:

- make sure basic map functionality and examples still work in IE 7-9
- calls to `flyTo` don't animate in IE < 9 (that doesn't support CSS3 transforms) and instead just jump to final value
- calls to `setView` and the family with fractional zoom values silently sets it to round zooms in IE < 9

Jakobud:
How are you typically detecting if IE < 9? Your second point (making `flyTo` not animate) will be pretty easy using CSS3 transform detection. But what feature do we need to try to detect regarding the `setView` change?

mourner:
I try to detect based on capabilities whenever possible (e.g. `L.Browser.any3d`), otherwise there are other props in `L.Browser`.
For `setView`, we need to use round zooms in case transforms are not available.

Jakobud:
Do you want those zoom values to round up, down or to the nearest whole number?

mourner:
To the nearest whole number I think, like GridLayer does.

Jakobud:
I should be able to tackle this. I'll take a look tonight.

mourner:
Did you look into this by chance?

Jakobud:
I started to but got sidetracked. Sorry I am in the process of getting my house ready to sell and haven't had time to mess with lately. In a few weeks I will have more time.

mourner:
Sure, no worries!

IvanSanchez:
In a related note: `setView(somewhere, 9.25);` works but `flyTo(somewhere, 9.25);` crashes.

IvanSanchez:
This is literally a 2-liner. However, IE <= 9 doesn't run `L.Map._resetView()` properly, as it somehow fails to reset the position of the tiles relative to the pane. :confused:

mourner:
@IvanSanchez that's pretty much expected since IE9 didn't support transform transitions as far as I remember, so it can't do zoom animations properly.

IvanSanchez:
Actually it was GridLayer's fault - it was calling code for 3d transforms without checking `L.Browser.any3d`.

Issue #2602, opened 2014-03-31
Open popup from another world


yaroslavvoloshchuk-codaio:
Code:

```
var map = L.map('map').setView([60, -2900], 13);

L.circle([51.508, -0.11], 1500).addTo(map).bindPopup("I am a circle.").openPopup();
```

Will broke tiles, like this:
![rsz_screenshot_from_2014-03-31_165908](https://cloud.githubusercontent.com/assets/874548/2568183/efbdabfc-b8de-11e3-9fbd-f77f5a032381.jpg)
but if we change setView zoom to 12:

```
var map = L.map('map').setView([60, -2900], 12);
```

All ok:
![rsz_screenshot_from_2014-03-31_170626](https://cloud.githubusercontent.com/assets/874548/2568194/1460ea46-b8df-11e3-9367-eb134b522c10.jpg)

Leaflet 0.8
Live demo: http://codepen.io/Wenqer/pen/nyzlq?editors=001

PR #3278, opened 2015-03-04
Refactor move/zoom events

### LINKED ISSUE #3134: flyTo generates many moveend events
Reported [here](https://github.com/mlevans/leaflet-hash/issues/38#issuecomment-69810963) and replicated with `debug/map/zoompan.html` (uncommenting the logging there for move events).

Expected behavior would be to emit a single moveend event when the animation completes or is canceled.

### LINKED ISSUE #3216: Rethink map events like move/viewreset
The old way of depending on viewreset after changing zoom and move for panning doesn't work with the new paradigm of fractional zoom and methods like `flyTo`. We need to rethink what events we fire and how layers like TileLayer and vectors depend on them.

mourner:
A work in progress to make movement-related events in Leaflet less confusing, more consistent across different types of movements/animations and fitting the new paradigm with fractional zoom levels and flyTo animations better.

- [x] remove `viewreset` event and depend solely on `zoom` event in layers instead
- [x] refactor confusing `Map` `_resetView` code into small separate methods
- [x] refactor `Map.TouchZoom` code to rely on `zoom` events during pinch-zoom
- [x] decide how to reset view when doing long-distance panning
- [x] simplify transformation math to enable transforming vector layers from any state
- [x] make `Renderer` rely on `zoom` events to scale renderer
- [x] get rid of the ambiguity of relying on `zoomanim` in places when there's no actual animation
- [x] stop vector reprojection/clipping/simplification on every frame #2982 
- [x] bring `viewreset` event back but in a different meaning (not zoom but full reset)
- [x] fix tests

Closes #3216, closes #3134. Related: #2982

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -546,7 +546,7 @@ L.Map = L.Evented.extend({
 		return this.fire('movestart');
 	},
 
-	_move: function (center, zoom) {
+	_move: function (center, zoom, data) {
 		if (zoom === undefined) {
 			zoom = this._zoom;
 		}
@@ -558,9 +558,9 @@ L.Map = L.Evented.extend({
 		this._pixelOrigin = this._getNewPixelOrigin(center);
 
 		if (zoomChanged) {
-			this.fire('zoom');
+			this.fire('zoom', data);
 		}
-		return this.fire('move');
+		return this.fire('move', data);
 	},
 
 	_moveEnd: function (zoomChanged) {
@@ -715,7 +715,6 @@ L.Map = L.Evented.extend({
 
 	_getNewPixelOrigin: function (center, zoom) {
 		var viewHalf = this.getSize()._divideBy(2);
-		// TODO round on display, not calculation to increase precision?
 		return this.project(center, zoom)._subtract(viewHalf)._add(this._getMapPanePos())._round();
 	},
 
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -193,7 +193,7 @@ L.Map = L.Evented.extend({
 
 		var oldSize = this.getSize();
 		this._sizeChanged = true;
-		this._initialCenter = null;
+		this._lastCenter = null;
 
 		var newSize = this.getSize(),
 		    oldCenter = oldSize.divideBy(2).round(),
@@ -291,8 +291,8 @@ L.Map = L.Evented.extend({
 	getCenter: function () { // (Boolean) -> LatLng
 		this._checkIfLoaded();
 
-		if (this._initialCenter && !this._moved()) {
-			return this._initialCenter;
+		if (this._lastCenter && !this._moved()) {
+			return this._lastCenter;
 		}
 		return this.layerPointToLatLng(this._getCenterLayerPoint());
 	},
@@ -505,6 +505,7 @@ L.Map = L.Evented.extend({
 		this._paneRenderers = {};
 
 		this._mapPane = this.createPane('mapPane', this._container);
+		L.DomUtil.setPosition(this._mapPane, new L.Point(0, 0));
 
 		this.createPane('tilePane');
 		this.createPane('shadowPane');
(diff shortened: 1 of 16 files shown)


mourner:
@IvanSanchez just rebased on master.

IvanSanchez:
We need to make sure this doesn't introduce a regression of #3272

mourner:
Yes. Also, pretty weird that the only failing test here is something about popups.

src/map/Map.js
```diff
@@ -520,43 +521,52 @@ L.Map = L.Evented.extend({
...
+	_move: function (center, zoom, data) {
+		if (zoom === undefined) {
+			zoom = this._zoom;
+		}
 
-		var loading = !this._loaded;
-		this._loaded = true;
+		var zoomChanged = this._zoom !== zoom;
 
-		this.fire('viewreset', {hard: !preserveMapOffset});
```
IvanSanchez: Not firing `viewreset` means that the GridLayers' `_resetView()` method is not called, which means the grid origins are not updated, which means a grey map whenever a `setView()` call moves the map around without changing zoom level.

There must be a (clean) way to tell the grid layers to rebuild the level origin whenever the position of `L.Map._mapPane` is reset.

This is not exactly #3272, but the symptom is very, very similar.

mourner: My head is spinning from thinking about this... I wanted to simplify things but it's not as easy. Perhaps we could do away without an additional event by passing some param in `moveend`, which should also be called when the map pane position resets.

IvanSanchez: How big is the overhead of an extra event? I'd rather have the `viewreset` event back, this time with a very clear semantic: "Recalculate the position of your DOM elements as if we would have just changed zoom".

mourner: We could add it back, although the semantics would change, since now changing zoom doesn't always imply recalculating everything (e.g. with fractions, or frames like flyTo does).

IvanSanchez: Maybe we could call `panBy()` inside `_resetView` when the zoom level doesn't change, and see if #2602 / 7e40b07d is a problem again.

mourner:
OK fiddling with this, I think we should bring back `viewreset`, just change the semantics so that it's only expected when the view is completely reset, not on zoom. 

Now the vector layers are a special case in that they need expensive reprojection/clipping/simplification process, which we can't do on each frame because that would be too slow. So we have two options:

- only reproject on `moveend` (e.g. after flyTo animation ended)
- reproject on any move but heavily throttled (e.g. not more often than once per 100ms)

The first option can cause unwanted appearance in some cases, while the second can cause hiccups during a smooth flyTo animation, and are generally less reliable and can cause subtle/hard to track bugs (we've been burned by this in GridLayer refactor).

I think we should go with the first for now and see how it goes.

mourner:
Got unstuck and made tons of progress here today! Vectors are working with flyTo, all the logic is much cleaner, all events should work as expected. Now we need to fix viewreset and tests, then test it thoroughly on stuff and it'll be a huge relief when this is merged.

yohanboniface:
\o/
I'll do some testing asap to join the party :)

src/map/Map.js
```diff
@@ -705,8 +715,7 @@ L.Map = L.Evented.extend({
 
 	_getNewPixelOrigin: function (center, zoom) {
 		var viewHalf = this.getSize()._divideBy(2);
-		// TODO round on display, not calculation to increase precision?
-		return this.project(center, zoom)._subtract(viewHalf)._add(this._getMapPanePos())._round();
+		return this.project(center, zoom)._subtract(viewHalf)._add(this._getMapPanePos());
```
IvanSanchez: Removing `._round()` causes an off-by-one-pixel error in the unit tests for popups.

mourner: cool, brought it back for now

igorshubovych:
Where is this Like button, when we need it so much?

Starefossen:
👏🏼😄

joshdickerson92:
@mourner, when you say "Vectors are working with flyTo", does this mean that they now render/ reproject as the flyTo animation is happening? If so, is there a parameter/ option that needs to be used?

mourner:
@joshdickerson92 they now render as the flyTo animation is happening, and don't reproject every frame but CSS-transformed instead for performance. Do you have any issues with this?

joshdickerson92:
yes, as per [this Stack Exchange question](http://gis.stackexchange.com/questions/168781/leaflet-vector-layers-not-rendering-during-flyto-animation) the vectors layers don't appear to be rendering during the animation. To be more specific I am using a marker.cluster vector layer with [this setup](http://jsfiddle.net/3v7hd2vx/).

mourner:
@joshdickerson92 please set up a minimal JSFiddle test case and open an issue.

joshdickerson92:
It appears that my issue is actually when using a GeoJSON as [this JSFiddle](http://jsfiddle.net/joshuadickerson92/3v7hd2vx/4/) works as expected when purely using markers, however this [JSFiddle](http://jsfiddle.net/joshuadickerson92/3v7hd2vx/5/) uses a GeoJSON and you can clearly see the issue.

mourner:
@joshdickerson92 hmm, your both JSFiddles are identical, so I don't see the issue.

joshdickerson92:
@mourner Whoops, should have updated the latter one now. If not then here is [the link](http://jsfiddle.net/joshuadickerson92/3v7hd2vx/7/)

mourner:
Ah, I see what you mean. The vectors are CSS-transformed, so zooming in they look pixelated and things like circle size doesn't change. I think we could update the behavior so that reprojection happens at least once in a while (e.g. every 200ms)

IvanSanchez:
Maybe we could reuse the logic from the tile loading algorithm: update things once per zoom level change.

joshdickerson92:
@IvanSanchez that is how I would perhaps expect it to work. @mourner can you see any quick work around to this in the time being?

Issue #2618, opened 2014-04-09
Polygon rings API


danzel:
Currently in polyline.js we have a bunch of TODOs around rings
https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Polyline.js#L20

And we have a weirdness with polygon creation that causes issues with leaflet.draw:
Create a polygon with new L.Polygon -> _latlngs is flat ```[ latLng, latLng, latLng ]```
Create a polygon from GeoJson -> _latlngs is nested ```[ [ latLng, latLng, latLng ] ]```

What are your thoughts on the Polyline/Polygon API?

We could change it so internally polygons are always nested and polylines never.
Then expand the api like:
```addLatLng: function (latlng, index)```
With index defaulting to 0.
By putting the index parameter second, people not using rings won't need to worry.

refs #2290

yohanboniface:
> We could change it so internally polygons are always nested and polylines never.

+1, this would mirror the [GeoJSON specs](http://geojson.org/geojson-spec.html), and thus facilitate a lot of interactions with GeoJSON.

@mourner do you want me to work on a PR on this direction?

mourner:
@yohanboniface sure!

yohanboniface:
Should we close, now that #3279 is merged?

PR #3279, opened 2015-03-04
Ensure polygons latlngs are never flat (cf #2618)

yohanboniface:
This is a first attempt to have a consistent internal structure for polygons latlngs.

This is a naive approach, I've focused on creating unit tests, and making it work, to start getting feedback.

My goals are:
- make Polygon and Polyline latlngs internal structure consistent
- no API retrocompat breaking
- ensure methods modifying latlngs works for simple cases but also complex ones (multipolygons with holes for example)

I've spot two methods that needed a bit of work: `addLatLng` and `spliceLatLngs`. Basically the need is to make that those methods can target a latlngs ring of a simple polyline but also, for example, some hole of a complex multipolygon.
What I come out with is to add an optional `latlngs` to those methods that is a reference to the array of latlngs we want to deal with, and when it's not given (default case) we are looking for the `defaultShape` of the Polygon or Polyline instance we are dealing with.

`spliceLatLngs` is a bit harder to make evolve, because it uses the magical `arguments` variable, and then it accept an undefined number of latlngs as parameters (as the default javascript splice Array method). I think the only way to make this evolve without breaking the API is to do a bit of duck typing. I chosen to inspect the last element that is given, so the API would look like:

    polygon.spliceLatLngs(1, 2, somelatlng, anotherlatlng, maybeanother, targetLatLngs)

Another option can be to change the API signature of those methods, but still supporting the "old" one, still with duck typing, for example:

    polygon.spliceLatLngs([1, 2, latng, latng, latng], [optionalTargetLatLngs])

And if the first element is an Array, we guess we are dealing with the new signature, otherwise we consider we have the old one.

For sure, we need to have both methods to have a consistent way of dealing with the target latlngs array.

Another point to look at is the naming: `_defaultShape` introduces the word `shape` which may be not enough acute (but I thought `_defaultLatLngs` was too confusing, and I didn't come out with a better idea).

I've first tried to support also targeting the latlngs array by its index (as suggested in #2618), but I found out that this cannot really apply in an elegant way to complex cases, so I thought it was better not to support it at all instead, and only accepting array reference.

Thanks in advance for your feedback :)


mourner:
Nice work! Happy to see so many tests.

Not sure about the get/splice/add methods API — it's pretty ambiguous and complicated, although it's tough to come up with a clear API that's backwards compatible here... Here's an alternative, simpler idea: 

- keep the current `getLatLngs`, `addLatLng`, `spliceLatLngs` working only with the first ring (keeps compat)
-  add a `Polygon` `getRings` method that returns the full structure for more sophisticated manipulation; to edit it, you would use the method, manipulate it as array of arrays with standard JS means, and then just use `setLatLngs`.

mourner:
cc @danzel @rwwagner90 — what do you think?

yohanboniface:
> add a Polygon getRings method that returns the full structure for more sophisticated manipulation

Humm, but, now, `getLatLngs` already returns the rings too, nope? Or am I missing something? :)

yohanboniface:
And I'm +1 on keeping the API simple, not trying to handle complex cases that don't fit in a simple API. We should just warn about that on the doc maybe if we go that way.

But one question at this point: given that `setLatLngs` already redraw the path, what about totally removing `addLatLngs` and `spliceLatLngs` that then would not be consistent in all the Polygon/line structures supported by Leaflet? We would only support `getLatLngs` and `setLatLngs`, but those are consistent and unsurprising.
That would indeed breaks the API, but maybe the 1.0 can support that.

rwwagner90:
I think both ways have merit, I think having the same methods, rather than a new getRings method would be nice, just so all cases would use the same method, but I can see the need for compatibility. Is it hard to make it compatible without separating the methods?

danzel:
Yeah this is a hard one. I expect most users polygons are single polygons with no holes. But we need to consistently support multipolygons with holes.

Do people even use spliceLatLngs? It seems pretty hard.

What @mourner has suggested sounds good, an easy API for easy polygons and a more powerful one when you are using complicated polygons.

mourner:
On second thought, just removing `spliceLatLngs` and `addLatLng` now seems like a pretty good idea. Moreover, now is the best chance to break compatibility, before 1.0 hits. 

I think a combination of `getLatLngs`, `setLatLngs`, `redraw` and manipulating the latlngs array by JS means is simpler, more powerful and less confusing at the same time.

mourner:
> Humm, but, now, getLatLngs already returns the rings too, nope? Or am I missing something? :)

I meant 0.7.3 always returns the first ring, so it would be compatible with the previous stable. But maybe that's not such a good idea.

yohanboniface:
> I think a combination of getLatLngs, setLatLngs, redraw and manipulating the latlngs array by JS 
> means is simpler, more powerful and less confusing at the same time.

Totally agreed! Let's update my PR that way :)

yohanboniface:
One more question: should we keep `_defaultShape` in a way or another (I've removed it for now) ? Refs: https://github.com/Leaflet/Leaflet/commit/602cff077f9d8eb778c26aaeb53c5fc37b3185da#diff-fa53d5de90cf287f76b8c7c7ff4fe13bR50 and https://github.com/Leaflet/Leaflet/commit/602cff077f9d8eb778c26aaeb53c5fc37b3185da#diff-e2f740d62a0e30737c493285332f0a13R51.

This could be of good help for simple cases, and this time it's consistent as it's only do what it says.

Still not sure about the name if we keep it and make it public, though: `getDefaultShape`, `getFirstShape` (the name should fit for both Polygon and Polyline, this is why I'm not suggesting to use "ring")?

yohanboniface:
@mourner any blocking point I should work on? :)

yohanboniface:
Related: https://github.com/Leaflet/Leaflet/pull/3386

mourner:
@yohanboniface please update to reflect the `_flat` change. Also, I think we should readd the `addLatLng` method at least, because it provides performance benefits over `setLatLngs` (fast bounds recalculation, no reconversion of all points).

yohanboniface:
> please update to reflect the _flat change.

Done :)

> Also, I think we should readd the addLatLng method at least, because it provides performance 
> benefits over setLatLngs (fast bounds recalculation, no reconversion of all points).

Makes sense. But as it was, it will break with nested polygons/polylines, so I need to check the "first" ring. Then back to [my previuous question](https://github.com/Leaflet/Leaflet/pull/3279#issuecomment-77522753), does it make sense to readd a `defaultShape` or `defaultLatLngs` method (private or maybe better public) ?

mourner:
@yohanboniface I think we should keep it simple, making `addLatLng` only add to the first ring in case of polygon (it would then be overrided in `Polygon.js`), because that covers most cases where performance matters (e.g. animating "growing" polylines).

yohanboniface:
What about letting people do `mypolygon._latlngs.push` then, instead of having a method that is not consistent across all cases?

mourner:
@yohanboniface `_latlngs.push` is not enough. I.e. you need to care about converting the latlng, recalculating bounds, etc. I think it's fine having a method that's not designed for extreme cases (e.g. adding latlng to a hole), but has an easy workaround for that and works in 99% cases.

yohanboniface:
> _latlngs.push is not enough. I.e. you need to care about converting the latlng, recalculating 
> bounds, etc. 

Good point :)

> I think it's fine having a method that's not designed for extreme cases (e.g. adding latlng to a 
> hole), but has an easy workaround for that and works in 99% cases.

Just to be sure we are sync on this: it's not only extreme cases, a simple polygon and a multipolygon (same for a polyline and multipolyline) doesn't share the structure, so if I just add the `latlng` to `_latlngs[0]` it will break for multipolygons, even without any hole. Same for multipolylines if I only add to `_latlngs`. And when I mean `break` I mean ends with a very bad situation, where we have added a `latlng` at the same level of an array of latlngs. This is why I think having a `defaultShape` or whatever the name should make this stronger, working everytime, but indeed only on the first ring, not on other rings, not on holes.

/me also calls in a [Mourner from the past](https://github.com/Leaflet/Leaflet/pull/3279#issuecomment-77435024) ;)

But if you are still convinced we should not care about multis, I'll do that way :)

mourner:
@yohanboniface OK revisited now. :) Lets add `_defaultShape`, and readd `addLatLng` using that (second argument there may also be a good idea). I probably was against all that because `spliceLatLngs` was too scary. :) But with just `addLatLng`, it looks simple.

yohanboniface:
Done :)

mourner:
:+1:

mourner:
Now we need to update docs.

yohanboniface:
How are we in the process to have doc on `master` branch? Should I wait a bit or should I update `gh-branch-master`?

mourner:
Update `gh-pages-master`. There are lots of changes already there, see #3098

PR #3285, opened 2015-03-05
Added zoomFactor map option

Jakobud:
Added the ability to control the zoom factor via Control.Zoom.js and ScrollWheel zooming.

This is in regards to fractional zooming. It will zoom the map in increments of whatever the zoomFactor is whenever you use the Zoom Control or Scrollwheel.

Pass the `zoomFactor` in the map options. Default value is 1. You can change the map's zoom factor at any time by using `map.zoomFactor = 0.25`.

I added a `zoom-factor.html` debug file to experiment with.

Also @mourner if you want, we can change the name zoomFactor to something else. Maybe zoomStep or zoomIncrement? zoomFactor sounded most straightforward at the time. This is a minor feature but it really should be in there in order to make fractional zooming shine.

Signed-off-by: Jake Wilson <jake.e.wilson@gmail.com>

### DIFF
--- a/debug/map/zoom-factor.html
+++ b/debug/map/zoom-factor.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+</head>
+<body>
+
+	<div id="map"></div>
+
+	<script type="text/javascript">
+
+		var osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+
+		var map = L.map('map', {zoomFactor: 0.25})
+				.setView([50.5, 30.51], 15)
+				.addLayer(osm);
+
+		function logEvent(e) { console.log(e.type); }
+
+		// map.on('click', logEvent);
+
+		// map.on('movestart', logEvent);
+		// map.on('move', logEvent);
+		// map.on('moveend', logEvent);
+
+		// map.on('zoomstart', logEvent);
+		// map.on('zoomend', logEvent);
+
+	</script>
+</body>
+</html>
--- a/debug/map/zoom-factor.html
+++ b/debug/map/zoom-factor.html
@@ -36,8 +36,9 @@
 
 		// map.on('zoomstart', logEvent);
 		// map.on('zoomend', logEvent);
+
 		map.on('zoomend', function(e){
-			console.log(e.target._zoom);
+			console.log("Current Zoom: " + e.target._zoom);
 		});
 
 	</script>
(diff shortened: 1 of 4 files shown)


iH8:
I've got a question regarding this commit. Let's say i have a minZoom of 3 and a maxZoom of 12 and my only available levels are 3, 6, 9 and 12. I'de set my zoomFactor to 3 and it nicely step from 3 to 6 etc to my understanding. But will this also be respected when using functions like fitBounds etc? I was wondering because of this question regarding zoom on Stackoverflow just now:

http://stackoverflow.com/questions/28904160/how-to-use-chunky-coarse-grained-zoom-levels-with-leaflet-maps

(This still wouldn't work in his/her case because he/she is working with zoomlevels which aren't multiples of the same number. Something like an array with available zoomlevels would be a solution, but that's beside my question)

Jakobud:
That is a really good question @iH8. I just tested it and the results are.... kind of unexpected. First of all, it appears to work perfectly when zooming with the Zoom Control. But with MouseScrollWheel zooming, it's a little weird. It appears that the zooming will not let you go beyond your maxZoom value, which is good. However, as you approach your maxZoom value, your zoom delta for some reason becomes exponentially less and less.

I just did a test starting at zoom level 15 with a `maxZoom: 18` and a `zoomFactor: .66`. Here is a screen shot from a `zoomend` handler showing the resulting zoom level:

screen shot 2015-03-06 at 12 37 52 pm

Everything works as expected until your distance to maxZoom is < zoomFactor, and then it starts tapering off but never actually hitting maxZoom, like an asymptote. I guess this kind of makes sense since the zoom delta is multiplied by the zoomFactor.

I need to rethink this slightly and figure out how to make this work properly.

On a related note, what is the point of the delta when zooming using the scroll wheel? The Zoom Control zooming is really straightforward. But the scroll wheel zooming functions are much more complicated. Why is it more complicated?

As for your `fitBounds` question I'm really not sure. Doing things like `map.setZoom(n)` does respect your `maxZoom`. Not sure how this would play into it. Clone this branch and check out the `debug/zoom-factor.html` and play around with it and let us know what you figure out.

Also the idea of passing an array of acceptable zoom levels is an interesting one..... Essentially we would need to refactor the zooming functions to use the array. And if there is no array passed then upon map instantiation, an array of zoom levels would be generated based on the min/max zoom and zoomFactor. Interesting idea to think about.

nothingisnecessary:
Hi there, I posted the original stack overflow question as this is an issue I am looking at with some software that uses Leaflet.

I don't necessarily think Leaflet should be changed to support multiple zoom levels, because this can be (and is currently being) handled by me on `zoomEnd`.

What do you think about doing some kind of a Layer option that uses the scaling or 'tweening' instead?

For example: when I zoom in to a level where I do not yet have local browser cache of the tile images, I can see Leaflet scale (stretch) the tiles at the previous zoom level (so that they appear a bit blurry) until these tiles are eventually replaced with the images for the new zoom level.

When the tile request completes, if the response does not contain a tile image (gets a 404 not found, or empty response), instead of doing whatever it does now (seems to just not change CSS class `leaflet-tile` into `leaflet-tile-loaded`), would there be a way to - based on a new layer option - use the scaled (transition?) tile image from the previous zoom level?

Not that big a deal - I know that tile providers skipping zoom levels is not common - just not sure if my current method of skipping to the next available zoom level is a "good user experience," whatever that means.. Thanks!

Jakobud:
I updated this to fix the issue where ScrollWheel Zooming had an Asymptotic Behavior as it approached `maxZoom`.

@nothingisnecessary Have you tried what you are describing with the current master branch? There is a `maxNativeZoom` option that uses scaled tiles at zoom levels beyond the maximum zoom layer for tiles you have.

IvanSanchez:
Superseded by #3523

PR #3287, opened 2015-03-06
e.button === 0 is the left mouse button.

Szunti:
I think e.which and e.button is used in these places for compatibility, and left mouse button is meant for both places. I.e. box zoom and dragging is only for left mouse button not for the middle one.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -46,7 +46,7 @@ L.Draggable = L.Evented.extend({
 	_onDown: function (e) {
 		this._moved = false;
 
-		if (e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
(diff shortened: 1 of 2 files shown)


mourner:
Good catch!

Trufi:
Unfortunately, this pull request broke map drag in IE8.

`e.button` values explanation from [here](http://www.w3schools.com/jsref/event_button.asp):

>Possible values:
0 : Left mouse button
1 : Wheel button or middle button (if present)
2 : Right mouse button
>
Note: Internet Explorer 8 and earlier has different return values:
1 : Left mouse button
2 : Right mouse button
4 : Wheel button or middle button (if present)

mourner:
Ha, so that's why I did it in the first place. I assume IE9 has `e.which`, which is why it all worked before. Reverting.

PR #3288, opened 2015-03-07
Also listen to events on marker icon childs elements

yohanboniface:
When using a DivIcon, we often add DOM elements in it,
which can then be the actual targets of DOM events, so we need
to add them in the "listened" targets.

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -166,5 +166,46 @@ describe("Marker", function () {
 
 			expect(spy.called).to.be.ok();
 		});
+
+		it('fires click event when clicked with DivIcon', function () {
+			var spy = sinon.spy();
+
+			var marker = L.marker([0, 0], {icon: new L.DivIcon()}).addTo(map);
+
+			marker.on('click', spy);
+			happen.click(marker._icon);
+
+			expect(spy.called).to.be.ok();
+		});
+
+		it('fires click event when clicked on DivIcon child element', function () {
+			var spy = sinon.spy();
+
+			var marker = L.marker([0, 0], {icon: new L.DivIcon({html: '<img src="data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=" />'})}).addTo(map);
+
+			marker.on('click', spy);
+
+			happen.click(marker._icon);
+			expect(spy.called).to.be.ok();
+
+			happen.click(marker._icon.querySelector('img'));
+			expect(spy.calledTwice).to.be.ok();
+		});
+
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Shouldn't this be handled by event bubbling?

yohanboniface:
Yeah, that was my first thought too, but I didn't get to a valid solution with this scenario, i.e. integrated in the actual event pattern.

AFAIK, we are listening DOM events on the map container only (which is certainly a good option, instead of having many native events on many elements), so we need to "register" the elements that, when real targets of the native event, have a Leaflet event attached to them.

To take an example, we could have the `map div` containing a `marker divicon` containing itself an `img`. The user clicks on the `img`, the event bubbles and we catch it on the `map div`; we check if we have any Leaflet event registered for the event target, which is the `img`, not the `marker divicon`. At this point, I have not clue how to play with bubbling again to allow that only registering internally the `marker divicon` we can still make a reference with the `img`.

If you, or someone else, have a suggestion, I'm happy to revisit my PR :)

mourner:
So, one idea I have is to loop through the parents (up to the map container) until we have found an event target. In theory it shouldn't be a significant performance overhead and should solve the issue.

yohanboniface:
@mourner r? :)

mourner:
I'd prefer a non-recursive approach (e.g. a `while` loop), otherwise looks perfect! Squash into one commit and I'll merge.

yohanboniface:
OK. That was my first idea too, but I didn't get to an elegant, strong and DRY `while` (nor `do/while`) implementation, given that we need to set target at least once before any check, and we need to go up to the parentNode only if we are not yet on the map container. Now I'm falling asleep so I'll look again tomorrow :)

mourner:
Maybe something like this:

```js
var el = srcElement; 
while (el) {
	var target = this._targets[L.stamp(el)];
	if (target) return target;
	if (el === this._container) break;
	el = el.parentNode;
}
return null;
```

mourner:
Cool!

yohanboniface:
Note: I went to a similar version (but with a while(true), yours is as usual more elegant), and was really more convinced by the recursive approach. So, out of curiosity and for my personal knowledge (and also for my future PR :) ), why do you prefer a loop version over the recursive one (in two words): is this related to performance, readability, code style consistence, or something else? :)

mourner:
@yohanboniface mostly performance. While it may not be very significant, I tend to always write code in its fastest form if it doesn't hurt code readability much — it adds up. I'm also thinking of relaxing the curly braces rule for one-liners to make code like this terser.

yohanboniface:
> I'm also thinking of relaxing the curly braces rule for one-liners to make code like this terser.

I'm clearly +1 on this! :)

PR #3289, opened 2015-03-07
Attempt to fix random PhantomJS DISCONNECTED errors on Travis

yohanboniface:
Source: https://github.com/karma-runner/karma/issues/598

### DIFF
--- a/spec/karma.conf.js
+++ b/spec/karma.conf.js
@@ -66,7 +66,6 @@ module.exports = function (config) {
 		// Workaround for PhantomJS random DISCONNECTED error
 		browserDisconnectTimeout: 10000, // default 2000
 		browserDisconnectTolerance: 1, // default 0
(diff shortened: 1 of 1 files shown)


yohanboniface:
Humm, so now the build are green, but I'm not sure to understand what happens there: there are two messages, one saying that it has disconnected before running all tests, the other one saying that all tests were ran successfully:

screenshot from 2015-03-07 12 55 40

It seems to say that they have be ran twice (may be thanks to the `browserDisconnectTolerance` parameter), but I'm not sure.

Then I don't know if we want to merge that or not.

mourner:
I think we shouldn't up `browserNoActivityTimeout`, since our tests don't run that long and if something didn't respond in 10s, Travis will probably just waste time waiting another 50s. But the tolerance setting which causes tests to rerun seems useful.

PR #3293, opened 2015-03-08
Add optional callback to map.flyTo

yohanboniface:
I know that having callback arguments it's not a Leaflet pattern, but given that `requestAnimFrame` is asynchronous and that `flyTo` can be quite "long", I think having a callback in this case makes sense, as it allows smoother animation sequence (i.e. being able to run the step following `flyTo` only when its fully finished).

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -585,7 +585,7 @@ describe("Map", function () {
 					done();
 				};
 			map.setView([0, 0], 0);
-			map.flyTo(newCenter, newZoom, callback);
+			map.once('zoomend', callback).flyTo(newCenter, newZoom);
 		});
 
 	});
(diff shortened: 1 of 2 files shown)


mourner:
Can you accomplish the same thing with `map.flyTo(center, zoom).once('moveend', callback)`?

yohanboniface:
Good point, master :)

I've removed the callback, but kept the unittest, refactored. I'll let you decide if we want to merge the unittest, or just close.

yohanboniface:
Humm, I'm doing some more plays here around and it seems to me that the "moveend" event is sent at the beginning of the flyTo process, while I'm expecting it to be sent only at the end, am I smoking to much air or is it the expected behaviour? I.e. should I investigate or just take holidays? ;)

mourner:
moveend should indeed be only sent in the end, but I need to finish off #3278 before we can investigate.

yohanboniface:
Oh right, I forgot about the dragon! ;)
I sent you my last 23 life points so you can kill it ;)

PR #3307, opened 2015-03-13
Always fire DOM event on the map too

yohanboniface:
Since the new event delegation API, when we add a listener on the map, for example on "mouseover", this listener is not called when the event is targeted by a layer (Polygon, Marker…).

What I understand, is that with the old API, the event was listened both on each layer and on the map. With the new API, it's listened only on the map (which is nice) and then forwarded to the original target, but only to this target.

So here is a suggestion to fix that.

Thanks for reviewing :)

### DIFF
--- a/debug/tests/mousemove_on_polygons.html
+++ b/debug/tests/mousemove_on_polygons.html
@@ -1,0 +1,48 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Leaflet debug page</title>
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+
+    <link rel="stylesheet" href="../../dist/leaflet.css" />
+
+    <link rel="stylesheet" href="../css/mobile.css" />
+    <style>
+        .mybox {
+            background-color: red;
+        }
+
+    </style>
+    <script type="text/javascript" src="../../build/deps.js"></script>
+    <script src="../leaflet-include.js"></script>
+</head>
+<body>
+
+    <div id="map"></div>
+
+    <script type="text/javascript">
+
+        var osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+            osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+            osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib}),
+            latlng = new L.LatLng(39.05, 8.40);
+
+        var map = new L.Map('map', {center: latlng, zoom: 12, layers: [osm]});
+
+        var marker = new L.Marker(latlng);
+        map.addLayer(marker);
+        var polygon = (new L.Polygon([
+            [39, 8.40],
+            [39.10, 8.50],
+            [39.05, 8.30]
+        ])).addTo(map);
+        map.on('mousemove', function (e) {
+            marker.setLatLng(e.latlng);
+        });
+        // We should be able to move marker around in a fluid way,
+        // plus going over the polygon with no issue.
+
+    </script>
+</body>
+</html>
--- a/debug/tests/mousemove_on_polygons.html
+++ b/debug/tests/mousemove_on_polygons.html
@@ -1,48 +1,107 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <title>Leaflet debug page</title>
+	<title>Leaflet debug page</title>
 
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
 
-    <link rel="stylesheet" href="../../dist/leaflet.css" />
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
 
-    <link rel="stylesheet" href="../css/mobile.css" />
-    <style>
-        .mybox {
-            background-color: red;
-        }
-
-    </style>
-    <script type="text/javascript" src="../../build/deps.js"></script>
-    <script src="../leaflet-include.js"></script>
+	<link rel="stylesheet" href="../css/screen.css" />
+	<style>
+		.mybox {
+			background-color: red;
+		}
+		td {
+			border: transparent solid 2px;
+		}
+		td.red {
+			border: red solid 2px;
+		}
+		td.updated {
+			border: transparent solid 2px;
+			animation-name: borderfade;
+			animation-duration: 0.5s;
+		}
+		
+		@keyframes borderfade {
+			from {
+				border: red solid 2px;
+			}
+			to {
+				border: transparent solid 2px;
+			}
+		}
+	</style>
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+	<table>
+	<tr><td>           </td><td>Enter       </td><td>Move       </td><td>Exit       </td><td>Click       </td></tr>
+	<tr><td>Triangle 1:</td><td id='enter1'></td><td id='move1'></td><td id='exit1'></td><td id='click1'></td></tr>
+	<tr><td>Triangle 2:</td><td id='enter2'></td><td id='move2'></td><td id='exit2'></td><td id='click2'></td></tr>
+	<tr><td>Map:       </td><td id='enter3'></td><td id='move3'></td><td id='exit3'></td><td id='click3'></td></tr>
+	</table>
 </head>
 <body>
 
-    <div id="map"></div>
+	<div id="map"></div>
 
-    <script type="text/javascript">
+	<script type="text/javascript">
 
-        var osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
-            osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors',
-            osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib}),
-            latlng = new L.LatLng(39.05, 8.40);
+		var osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+		    osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+		    osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib}),
+		    latlng = new L.LatLng(39.05, 8.40);
 
-        var map = new L.Map('map', {center: latlng, zoom: 12, layers: [osm]});
+		var map = new L.Map('map', {center: latlng, zoom: 12, layers: [osm]});
 
-        var marker = new L.Marker(latlng);
-        map.addLayer(marker);
-        var polygon = (new L.Polygon([
-            [39, 8.40],
-            [39.10, 8.50],
-            [39.05, 8.30]
-        ])).addTo(map);
-        map.on('mousemove', function (e) {
-            marker.setLatLng(e.latlng);
-        });
-        // We should be able to move marker around in a fluid way,
-        // plus going over the polygon with no issue.
+		function update(domid) {
+			return function(){
+				document.getElementById(domid).innerHTML = Date.now();
+				document.getElementById(domid).className = 'red';
+				window.setTimeout(function(){
+					document.getElementById(domid).className = 'updated';
+				},1);
+			}
+		}
+		
+		var polygon = (new L.Polygon([
+			[39, 8.40],
+			[39.10, 8.50],
+			[39.05, 8.30]
+		])).addTo(map).on('mouseover',update('enter1'))
+		              .on('mousemove',update('move1'))
+		              .on('mouseout',update('exit1'))
+		              .on('click',update('click1'))
+		              .bindPopup('Triangle 1');
+		
+		var polygon2 = (new L.Polygon([
+			[39.03, 8.30],
+			[39.10, 8.40],
+			[39.00, 8.30]
+		])).addTo(map).on('mouseover',update('enter2'))
+		              .on('mousemove',update('move2'))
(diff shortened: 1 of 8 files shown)


yohanboniface:
I'm thinking about something else.

As far as I understand, with the old API, as we were listening *DOM events* both on the vectors or marker and on the map, we were able to use the `stopPropagation` when catching the event on the vector or marker, which then would prevent the map to catch the event.
I think this will not work anymore with my current patch proposal, as we are dealing with *Leaflet events*.

Also, I wonder if a better approach is not to extend the propagate API to always also propagate to the map.

(Breakfast time. After that I'll try to add more unit tests to cover the `stopPropagation` use cases.)

yohanboniface:
@mourner I think I need your thoughts before going forward :)

mourner:
Haven't thought about this closely yet, will do :)

yohanboniface:
@mourner this last commit makes all tests happy, but I feel like workarounding the issue, so I think a more global (expert) view is still needed :)

MelnikVasya:
In this branch don't work popup

yohanboniface:
Yes! Because `click` event is sent both to the layer and to the map, so it opens the popup and close it just in the same stack. So the layer should stop it for it not to be propagated to the map.

But clearly this means that this branch is work in progress. ;)

mourner:
I'm kinda lost here, need to revisit and rethink how events and propagation work to have some input in this...

mourner:
@yohanboniface as an alternative, can you please check https://github.com/Leaflet/Leaflet/pull/3450 and see if it's a good approach for this case? This way we explicitly bubble up all layer events to map. There should be the same problem with popups though.

IvanSanchez:
Note to self: create a webpage inside `/debug/test` with overlapping polygons, draggable markers, and mouseover/mouseout event handlers on everything.

yohanboniface:
@IvanSanchez added :)
So it works smoothly in the `map-domevent` branch, but it doesn't in the master branch:

- when the mouse is over the marker itself there is no `mousemove` event fired so the marker stop, then the mouse goes out the marker, and the map fire again the `mousemove` so the marker move above the mouse again, and so on
- when the mouse is over the polygon, nothing happens for the same reason

yohanboniface:
@mourner I'm travelling in Italy right now, so not really connected and not really behind my computer, but I will have a look for sure. As first thoughts, this option sounds really interesting, but I do think we still should support bubbling events from features (polygon, marker...) to map, and being able to stop them when we do not want this bubbling.

IvanSanchez:
So the problem with popups is that the order of the events is:

1. polygon preclick
2. polygon click (popup opens here)
3. map preclick (popup closes here)
4. map click

I'll spend some time trying to figure out how to fix this order into preclick-preclick-click-click.

yohanboniface:
Good catch :)
Seems to me that, out of what you found, the popup issue is a generic problem introduced by this PR: events listened both on a feature AND the map will often need to be stopped not to trigger twice the listener.
I feel like we need to spend a bit of time more to clean/rationalize the way Leaflet events can be stopped if we want to dig more on this PR.
I'm at work today, but on IRC if you are on this and need a plastic duck to talk to :)

mourner:
What I don't like in the code changes is that there's now `handleDOMEvent` method that used to actually handle DOM events, but now also handles Leaflet events in special cases, and there's Leaflet `DomEvent.stopPropagation` method that used to only deal with DOM events and now has a special case for Leaflet events. It's confusing, and will probably lead us to more traps later.

We should look for a way to handle the issue more gracefully.

yohanboniface:
> We should look for a way to handle the issue more gracefully.

This needs a sprint in Trondheim for sure! ;) cc @IvanSanchez 

(/me at SotM France for three days right now)

yohanboniface:
@mourner, tested with addEventParent/removeEventParent: seems much more cleaner. Not sure about the perfs, though, as we are adding the map in the hash of every single Layer instance. What do you think?

> there's Leaflet DomEvent.stopPropagation method that used to only deal with DOM events and 
> now has a special case for Leaflet events. It's confusing, and will probably lead us to more traps 
> later.

I agree that, as it is, it's hacky. However, I think that if we take care of propagating events (which I think we should) we should also have of a way to stop them, nope?
I thought that the lightest way was to rely on the `cancelBubble` property even for Leaflet events, so I've made a quick POC. But maybe we should instead have an internal property, which would be less confusing (like `_stopped` or whatever). What do you think?

mourner:
You're probably right about performance, but I need to recheck this.

> But maybe we should instead have an internal property, which would be less confusing (like _stopped or whatever). What do you think?

Sounds good. `cancelBubble` here is indeed a bit confusing.

yohanboniface:
@mourner Done :)
Should we document that somehow? Or we wait a bit to see if we are sure we go this direction? :)

mourner:
@yohanboniface lets wait, I'm still not sure about the approach and have a feeling that we may regret merging this too hastily.

yohanboniface:
OK :)

jfirebaugh:
Seems to me the right approach to this problem is to combine event delegation with simulated event bubbling. So:

* Map receives an event bubbled from a child element. Now it's responsible for performing delegation. It traverses the DOM hierarchy upwards from the source element, looking for listening targets.
* Right now the code assumes there is only one listening target. It should be changed to fire an event to _each listening target_ in the hierarchy, as it traverses it. This is **simulating event bubbling** for delegated events.
* After firing the event at a particular target, check [`e.isPropagationStopped()`](http://www.w3.org/TR/2003/WD-DOM-Level-3-Events-20030331/events.html#Events-Event-isPropagationStopped). If so, stop the simulated bubble. (Will need to check browser compatibility of `isPropagationStopped` and possibly implement a polyfill.)

With this strategy we shouldn't need any special handling for firing events on the map container additionally -- it just needs to be registered with `addInteractiveTarget` -- and we get a clean way to cancel propagation.

yohanboniface:
Thanks @jfirebaugh, as usual your explanation is accurate and synthetic. :)

Some comments though:

- seems that you are strictly talking about DOM events, while at the moment Leaflet is catching the DOM event then firing a custom event; which then doesn't know about `isPropagationStopped` for example. So we need to decide if we switch to using only original events (but adding some properties to it, like latlng?, which sounds a bit hackish to me), or to add a `isPropagationStopped` like method on it (which is what is done in this PR BTW)

- when referring to "hierarchy", I'm not sure if we are talking about DOM hierarchy or Leaflet hierarchy; i.e. are we traversing DOM parent of the original target (say svg => overlay div => … => map div), or the Leaflet hierarchy (say marker => layer => map)? As we are listening events on Leaflet object instances, I guess we want to traverse the Leaflet hierarchy, but then this concept is a bit fuzzy to me: when an event is initiated on a marker, for example, how do we define its parents hierarchy? Everything we have called `addTo` with it as parameter? In this case I should make that `addEventParent` is called on the right place

- does this only works for DOM events, or should we expect that every fired event should "bubble" through its parents hierarchy?

I really like the idea of firing and bubbling Leaflet events and being able to stop them just like "normal" events (so for example catching every marker events on a FeatureLayer), but we should reach a simple, strong, and efficient implementation.

@mourner @IvanSanchez thoughts? I'm on IRC if needed.

I'll hack a bit on this today, so I may post further comments :)

*(1.0-beta1, watch out, we are coming! :) )*

jfirebaugh:
I'm talking about DOM hierarchy traversal; the existing traversal is [here](https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L627-636). I'm not suggesting extending the existing `addEventParent` scheme.

This is just for events that go through the `map._handleDOMEvent` logic. I'm not sure we should simulate bubbling for custom events. Perhaps there's a way to unify this with the `addEventParent` stuff but it's not immediately obvious to me.

yohanboniface:
> I'm talking about DOM hierarchy traversal; the existing traversal is here. I'm not suggesting 
> extending the existing addEventParent scheme.

I think I'm not fully following you. I don't get how to connect DOM events with Leaflet listeners is this case.
Let's take an example. If I do `mymap.on('mousemove', myhandler)`, I'm attaching the listener to the `map` object, but not as a normal DOM listener (I mean, it's just a entry in a JavaScript hash). So when processing the DOM event, caught from a polygon for example, even if I traverse all the DOM hierarchy, I'll never fire `myhandler` (which is not attached to the map container).

Are you suggesting to specifically attach map DOM event listeners to the map container, or am I missing? :)

jfirebaugh:
@yohanboniface There is already a map between DOM elements and `L.Evented`s, it's  implemented by `map._findEventTarget` and `layer.addInteractiveTarget` (which could easily be moved to `map`). Given a DOM element, we get its `L.Evented` via [`this._targets[L.stamp(src)]`](https://github.com/Leaflet/Leaflet/blob/04a0a4e4d189a792e74c8f561742e5946c58622a/src/map/Map.js#L628-628).

yohanboniface:
@mourner one more try ;)
A first shot implementing @jfirebaugh recommendations.
Some remarks:
- `isPropagationStopped` doesn't seem supported at all neither by FF nor Webkit at this time (this is why I'm still using `_stopped` instead)
- while catching DOM events, we are firing Leaflet events (because we add some properties, like the `latlng` the event occurred at)
- when firing an event, we first clone it, so we can't rely on the event itself while bubbling by hand (this is why I'm setting `_stopped` on `e.originalEvent` instead of `e`)

mourner:
Whoa, looks pretty nice! I'll check it out. So we avoid the performance implications of `addInteractiveTarget`, and popups also work as expected, right?

yohanboniface:
>  So we avoid the performance implications of addInteractiveTarget

Yep :)

> and popups also work as expected, right?

As far as I can test, yes, and unit tests are green. :)

mourner:
Just got back from a 2-week US trip, now looking into it again. :) After merging this we're only left with docs updates for a proper beta release. @jfirebaugh can you help me review this?

jfirebaugh:
* `_fireDOMEvent` is used by `Canvas.js`, it will need changes too.
* This will change what properties the `preclick` event object has -- before it has `containerPoint`, `layerPoint`, `latlng`, after it does not. I don't know whether this is important or not.
* I would do some refactoring here, essentially inlining `_findEventTargets` into `_fireDOMEvent` so that it isn't necessary to construct a temporary `targets` array on each event.

yohanboniface:
>  _fireDOMEvent is used by Canvas.js, it will need changes too.

Gasp! Good catch :) I'm surprised this doesn't break any unittest :(
I'll have a look, it seems a bit more work than just changing how `_fireDOMEvent` is called.

>  This will change what properties the preclick event object has -- before it has containerPoint, layerPoint, latlng, after it does not. I don't know whether this is important or not.

Not sure about that. I've added a unittest that covers this, tell me if I'm missing something. :)

> I would do some refactoring here, essentially inlining _findEventTargets into _fireDOMEvent so that it isn't necessary to construct a temporary targets array on each event.

I've moved the call of `_findEventTargets` in to `_fireDOMEvent`. Did you mean actually *moving* the code of `_findEventTargets` inside instead?

jfirebaugh:
> Did you mean actually moving the code of _findEventTargets inside instead?

Yes, I mean collapsing the two loops ([1](https://github.com/Leaflet/Leaflet/blob/ababbe46485bfea1a6ae02c07aa9b031a30511c5/src/map/Map.js#L629), [2](https://github.com/Leaflet/Leaflet/blob/ababbe46485bfea1a6ae02c07aa9b031a30511c5/src/map/Map.js#L693)) into one.

mourner:
@yohanboniface any further updates, or should I attempt to address the comments above myself? That's the only issue to finish and merge for 1.0-beta1. :)

yohanboniface:
@mourner as always I'm travelling while working, so not that much extra time. My plan is to work on this one morning this week. My main issue is to unit test the Canvas scenario.
But if you feel to work on it, I don't want to refrain you :)

mourner:
Looks like popups still don't open with the latest commits. Checked on `debug/map/map.html`.

yohanboniface:
Ahem, good catch @mourner :/
I'll look on this too.
While sailing today I had a quick shot on the canvas issue. I've pushed a working version, but not really happy with it.
Basically, I think we can't use the same manual tree traversing while using canvas (i.e. we can't add the `canvas` as `addInteractiveTarget`) given that the canvas takes all the map area, and thus listen to events also where nothing is drawn.
On a train tomorrow, hopefully I'll continue hacking.

yohanboniface:
@mourner fixed the popup issue.

What I don't really like in the current solution, is that we have two ways of bubbling events: one using our internal propagate system (to reach for example the LayerGroup from a marker), the other one traversing the DOM tree. That doesn't seem really ideal.
Should we try to merge those two patterns? I.e. something like while in `_fireDOMEvent`, looping through both DOM tree to find registered listeners AND looking if there are some manually registered ones (through `addEventParent`) ?

mourner:
Thanks for the fix!
Yeah, the double propagation mechanic is pretty complicated, but if we try to refactor this now, we may get into a trap and stuck for a while more, so I'd rather merge now and release, and think about refactoring later. We can't make it perfect after all, — only good enough.

mourner:
@yohanboniface wanna do a final rebase and clean up before we merge? E.g. marking fixup commits, etc.

yohanboniface:
> wanna do a final rebase and clean up before we merge?

Sure!

> E.g. marking fixup commits, etc.

Not sure what you mean by "marking fixup commit". Shouldn't I just rebase everything into one clean commit?

mourner:
I mean, when you do an interactive rebase, you can reorder commits, mark some of the commits as "fixup" so that they're merged with a commit above (often useful when you're doing small fixes such as whitespace or typo).

yohanboniface:
OK, I'm doing that in a moment :)

mourner:
> I would do some refactoring here, essentially inlining _findEventTargets into _fireDOMEvent so that it isn't necessary to construct a temporary targets array on each event.

BTW, do we want to refactor at least this bit before merge? It would be nice, but I didn't find a clean way to do that yet, and performance gain is probably not significant here because it only happens on DOM events.

yohanboniface:
I've not reached a nice solution when tried (but maybe it was just too much waves ;) ), after fixing the canvas issue. I'll make a new shot just after nomnom :)

mourner:
Ah what the hell!

Issue #2558, opened 2014-03-19
Leaflet 19 Mar sprint


mourner:
cc @tmcw @jfirebaugh @danzel 

The biggest task today is pushing #2382 (`easey` branch) to a state where it is usable and can be merged to master. The things remaining:
- [x] currently buffer tiles from other zooms are not managed when browsing the map — we should implement removing lower/higher-res tiles in places where current-res tiles were loaded or probably off the screen to make things fast
- [x] fix the other-res tiles to fade in correctly when zooming back to corresponding zoom
- [x] make zoom animations (and generally all animations) stoppable — this means parsing the transform matrix for zoom animations as it happens for panning, adding a `stop` method that stops any animation and stopping any old animations when starting new ones
- [ ] look into improving the scroll-zooming behavior so that it's not as erratic, adjusting the easing when zooming more in the middle of an animation (we will be able to do that after the previous point)
- [x] make sure all events are consistent across all animation methods (move/zoom/drag/etc. start/end/etc.)
- [ ] test things heavily on different browsers and devices

Issue #426, opened 2012-01-10
drift of coordinates after zooming


przemekl:
var mapCenter = new L.LatLng(10.000000, 10.000000);
map.setView(mapCenter, 1)
for (i=2;i<19;i++){
map.zoomIn(); }
map.getCenter() - returns (9.96885, 9.84375)
Second place is shifted ca. 20km
Whether it is a rounding error, or screen coordinates conversion?

PR #3325, opened 2015-03-25
0.73: Allow fractional zoom stopping with Map.TouchZoom and fix CRS.scale calc with fractional zoom

adriankb:
Hi.

I believe that I have found a fix to allow touch/pinch fractional zoom support (when using plain Leaflet object and or GeoJson objects). But I could not comment as to if this will break any map with tiles. We currently don't use tiles, so I don't have expertise.

This Pull is for stable (sorry if this is the wrong destination), I will create a second for master https://github.com/Leaflet/Leaflet/pull/3326 (as the code is different). I believe work has been done in master for version 1.0 to support better zooming so if this Pull is ignored in favour of the master pull then at least this code might serve as a reference for others using 0.7.

The underlying issue being solved is the "jump" in the map when stopping a touch/pinch zoom whilst the map snaps to a whole number zoom level. Our users on iPad's and Android were finding this very obvious and it was very hard for them to zoom in/out to the level of detail that they wanted. 

Changes to Map.TouchZoom support stopping the zoom level exactly where the user has stopped the pinch.

The secondary issue uncovered is that the built in CRS.scale calculation also does not support fractional zoom. The side-affect of this is that when touch/pinch zooming, the percent with which the map is grown or shrunk does not reflect the final size that the map will be drawn at. Thus this means that there is still a "jump" in the map when stopping a touch/pinch zoom, even with the first fix in place.

The commit to CRS.js supports Scale calculation for fractional zoom levels.

Please refer to these related issues:
https://github.com/kartena/Proj4Leaflet/issues/57
https://github.com/kartena/Proj4Leaflet/pull/73
https://github.com/Leaflet/Leaflet/issues/426
https://github.com/Leaflet/Leaflet/pull/1309
http://stackoverflow.com/questions/24151177/fractional-zoom-levels
https://github.com/Leaflet/Leaflet/pull/2382
https://github.com/Leaflet/Leaflet/issues/2558

### DIFF
--- a/src/geo/crs/CRS.js
+++ b/src/geo/crs/CRS.js
@@ -22,7 +22,13 @@ L.CRS = {
 	},
 
 	scale: function (zoom) {
-		return 256 * Math.pow(2, zoom);
+		var iZoom = Math.floor(zoom);
+        var baseScale = 256 * Math.pow(2, iZoom);
+        var nextScale = 256 * Math.pow(2, iZoom + 1);
+        var scaleDiff = nextScale - baseScale;
+        var zDiff = (zoom - iZoom);
+
+        return baseScale + scaleDiff * zDiff;
 	},
 
 	getSize: function (zoom) {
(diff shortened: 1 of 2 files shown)


mourner:
This will break a lot of apps, especially ones that use tiles (which is 99% of Leaflet apps). So I don't want to merge it into stable — lets continue discussion on the master changes.

PR #3331, opened 2015-03-26
Add L.Sleep plugin

atstp:
Sate Leaflet's scroll-hunger with a small plugin.


mourner:
Can you add a plain description (maybe in parentheses, a couple words) so that people could find your plugin e.g. by "scroll" keyword?

atstp:
it's fixed and squashed

mourner:
Great, thanks!

PR #3340, opened 2015-03-30
Fix #3333 by catching the onScroll event on the map container

### LINKED ISSUE #3333: Keyboard-tabbing to an offscreen marker breaks pane positioning
* Create a map with at least one L.Marker (ensure the "keyboard" option of the marker is set to "true", should be the default) - http://leafletjs.com/examples/quick-start-example.html will work.
* Pan the map so that the L.Marker is partially off-screen, like so:

out-of-screen-tabbing-1

* Press "Tab" on your keyboard until the L.Marker is focused. The positioning of all map panes breaks up: see how all the L.Controls (attribution, zoom, etc) move around:

out-of-screen-tabbing-2

I can reproduce this in Firefox 35, Chrome 41, and MSIE11.

IvanSanchez:


mourner:
Fix the build and then squash the commit please :)

IvanSanchez:
`git commit --amend` is my friend. How did I manage to f*ck up a 6-line commit anyway? :-)

mourner:
Thanks :)

PR #3341, opened 2015-03-30
Leaflet.BoatMarker

tomaszbrue:
Added my leaflet plugin https://github.com/tomaszbrue/leaflet.boatmarker to the list. Thanks!


mourner:
Cool!

Issue #3332, opened 2015-03-27
IE8 tile loading bug


Trufi:
Can you look at leaflet (master branch) in IE8? Tiles sometimes freeze and stop loading after several drags and zooms. It looks like:

![screenshot from 2015-03-27 14 30 50](https://cloud.githubusercontent.com/assets/3996552/6864717/06798cc0-d487-11e4-8d7b-afa4916b6cdd.png)
Windows XP, Service Pack 3, IE8

perliedman:
I can reproduce this with latest master on IE8/Win7 (virtual machine from MS in Virtual Box) with the `/debug/map/map.html` example.

An interesting question is if Leaflet still aims to support IE8, which starts to get _really_ old, and it seems some libraries are dropping support for it.

I'll try to have a look, but given the state of IE8 debugger, I can't promise anything :)

mourner:
Yeah, Leaflet still has to support IE8 since A LOT of government/enterprise apps depend on this. Thanks for looking into this.

perliedman:
Follow up: I ran the test specs on IE8, and there are quite a few errors. Some of them appear in `GridLayer`, which I guess might be the issue we're seeing here.

perliedman:
So far, I've found out that the tiles load, but for some reason the `leaflet-tile-loaded` class is never applied to the element. I'll try to find some more time to dig into why.

As a side note, this issue can be debugged in IE11 if you set the debugger's document mode to IE8, which makes things a lot easier.

RLRR:
I think I've managed to pinpoint the bug, here is a possible solution: #3346

mourner:
@RLRR thanks a lot! Merged.

PR #3346, opened 2015-04-01
Fix tile loading in IE8

RLRR:
See #3332. Minimal steps to reproduce the bug:

1. Drag the map a couple of screens away (zooming is not necessary)
2. Drag back to original position

The problem is related to browser caching.  IE8 seems to call `onload` handlers of `` elements synchronously (immediately after assignment to `tile.src`) if it gets the image from cache.

This caused the `_tileReady` handler (in `GridLayer`) to be called before a tile was added to `this._tiles` object, so the handler returned without applying `leaflet-tile-loaded` class to the element.

This pull request fixes the bug with a simple `setTimeout` hack. I wasn't able to find a more elegant solution without changing a lot of code.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -85,7 +85,12 @@ L.TileLayer = L.GridLayer.extend({
 	},
 
 	_tileOnLoad: function (done, tile) {
-		done(null, tile);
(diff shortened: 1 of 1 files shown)


mourner:
Wow, nice! Can't say if there's a better fix, lets go with this one for now.

PR #3347, opened 2015-04-01
Update reference.html

Lambdac0re:
L.version is not a function, so it doesn't return anything


mourner:
Thanks! Wow, it still says 0.5, I should update this :)

Issue #3343, opened 2015-03-31
Add an option CRS to L.TileLayer


molokovskikh:
Example code:
```js
var map = new L.map("map").setView([55,37], 10); 
L.control.layers(
{
    'Yandex': L.tileLayer('http://{s}.pvec.maps.yandex.net/?l=pmap&x={x}&y={y}&z={z}',{
        crs:L.CRS.EPSG3395,//that's what I would like to
        subdomains:['00','01','02','03','04'],
        attribution: '&copy; <a href="http://yandex.ru/copyright">Яндекс</a>'
    }), 

    'OpenStreetMap':L.tileLayer('http://{s}.tile.osm.org/{z}/{x}/{y}.png', {
        attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a>'
    }).addTo(map)
}).addTo(map);
```
There is a difference between the base layers when switching. 
You need to add functionality presumably L.GridLayer (_reset, _update), so to have a tile layer was binding to the CRS option TileLayer.

danzel:
Have formatted your code.
Sort of a dupe of #2553
I have a feeling this was going to be hard to support but can't find the bug I was thinking about. @perliedman should know :)

molokovskikh:
I have a <a href='http://jsfiddle.net/yk6nkwqu/5/'>solution</a> to this problem, but I would like to make this functionality was already in the engine without hacks. Preparing a pull-request.

perliedman:
Also see #2927 and #3002, which are also discussion of this issue.

While there might be a few use cases for this, in general it's not a good idea to switch CRS on a map, and I have a hard to visualizing how it would work with a single tilelayer having a different CRS.

PR #3351, opened 2015-04-02
Add support option CRS

molokovskikh:
Added support for the option of the coordinate system #3343

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -37,6 +37,10 @@ L.TileLayer = L.GridLayer.extend({
 		if (typeof options.subdomains === 'string') {
 			options.subdomains = options.subdomains.split('');
 		}
+		
+		if(this.options.crs) {
+			this._supportCrs();
+		}			
 
 		// for https://github.com/Leaflet/Leaflet/issues/137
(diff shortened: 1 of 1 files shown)


perliedman:
Can you elaborate a bit on what functionality this PR adds?

I've read #3343 and see that it's about adding a `crs` option to `L.TileLayer`. Some questions about this:

1) What happens when you add two (or more) layers with different `crs` properties?
2) What happens when you add one layer with `crs` and one without (my guess is same as (1), but that might not be obvious to users without experience working with CRSs)

My feeling is that this creates a more complex and less intuitive API, a bit like the problems with OpenLayers, where you can (or at least could, in v2.x) set projection on each layer, setting you up to shoot yourself in the foot.

I would be more positive to altering the CRS on the map, that has been suggested a couple of times before, since it at least keeps the CRS consistent for all layers, with a single setting.

mourner:
> My feeling is that this creates a more complex and less intuitive API

Yes. So I don't think this needs to be in the Leaflet core. You could extract that into a useful plugin though.

johnd0e:
@molokovskikh 
It seems that this patch is not compatible with recent leaflet versions (

PR #3356, opened 2015-04-04
Added Leaflet.Weather plugin

oskosk:
Added row in control plugins table with reference to https://github.com/oskosk/Leaflet.Weather


jieter:
Nice!

PR #3358, opened 2015-04-06
Change translate3d detection

Trufi:
https://github.com/Leaflet/Leaflet/pull/2574 created a year ago. Now, no webkit browsers like firefox works fine with `translate3d` property.

I added `L.Browser.opera12` to avoid the logical confusion with `is3d = !L.Browser.opera3d`.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -35,6 +35,7 @@
 		android23: android23,
 		chrome: chrome,
 		safari: !chrome && ua.indexOf('safari') !== -1,
+		opera12: opera3d,
 
 		ie3d: ie3d,
 		webkit3d: webkit3d,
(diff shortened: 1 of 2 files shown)


mourner:
Good catch. Actually I think `any3d` should be good here. No need to add a separate `opera12` property.

mourner:
I mean, could we just ditch `opera3d` property and add `!opera` to `any3d` definition?

Trufi:
I hope I understood you right.

We can't add `!opera` to `any3d` definition, because last versions of Opera work good with `translate3d`.

mourner:
Yes, that looks good. No need to expose `opera12` property publicly though I think, but that's a minor thing.

PR #3360, opened 2015-04-08
Add isLoading to GridLayer

stephank:
We currently monkey-patch this in a setup where we do preloading.

We have several instances of Map, and one instance of TileLayer shared between the maps. Before creating a new Map instance, we create an invisible instance of the same size, add the TileLayer, and either check `isLoading()` or wait for the next `load` event.

I guess this also saves on calls to `_noTilesToLoad()` :-)


mourner:
That looks pretty reasonable, I'll merge probably.

PR #3371, opened 2015-04-12
Fix Polyline.getCenter returning undefined with very small line

yohanboniface:
As the center is computed from pixels (which is needed not to be out when computing near the poles), when the line is so tiny in the current map view that all points are on the same pixel, it wasn't returning a LatLng because if never entered the loop.

This may need to be checked on the Polygon side too, but I prefer to have your feedback before going further :)


mourner:
Thanks! The polygon should be good in theory.

PR #3372, opened 2015-04-12
Fix Polygon.getCenter returning invalid LatLng when all points in same pixel

yohanboniface:
;)

By the way, given that the center depends on the current view but returns a LatLng (which can then be expected as invariable), I wonder if we shouldn't rename `getCenter` in something like `getCurrentCenter`, `getViewCenter`, `getFlattenCenter`… what do you think?


mourner:
Thanks! No need to rename I think. Simple names are better.

Issue #3075, opened 2014-12-03
touchstart on markers


guyisra:
I have many markers on the map, each with its own popup. As I don't need all popups to be setup initially, I change the popup content when `mouseover`, which then request the data for the popup, and when `click` it will open the popup

```
marker.on('click', function (e) {
              this.openPopup();
    });
    marker.on('mouseover', function (e) {
          // get popup content
        }
    });
```

This works fine on non-mobile, but on mobile, there is no `mouseover` or  support for `touchstart`. How can I create this behavior in mobile using the existing marker events? (unless there is a way to extend the marker class to support touch events)

Issue #1542, opened 2013-03-21
binding to "touchstart" events on map?


gr2m:
I try to make this work on iOS / Safari:

``` js
var map = L.map('map').setView([51.505, -0.09], 13);
map.on('touchstart', function() {
  // does not get here when touching the map
  alert('touchstart')
})
```

But couldn't figure out how. It's not listed in the [API reference](http://leafletjs.com/reference.html#map-events) but I wonder if there is any way to support this event?

Or maybe there is even a better way to achieve the following: 
I want to add a marker to a map, by touch and holding for a second. It was easy to set this up for mouse events, but I can't find a way to make the same for touch events.

Any idea?

levijackson:
Expanding on the post above... for touch events the `touch.clientX` and `touch.clientY` may not be accurate if the map is not full screen. When I tried the above code the coordinates were off by a bit. I took a look at the way Leaflet figures out the coordinates from event data in `L.DomEvent. getMousePosition()` and replicated the logic with touch events.

Instead of doing
```
touch = e.touches[0]
containerPoint = L.point(touch.clientX, touch.clientY)
```
I did
```
var touch = e.touches[0],
    rect = this._container.getBoundingClientRect(),
    lat = touch.clientX - rect.left - this._container.clientLeft,
    lng = touch.clientY - rect.top - this._container.clientTop,
    containerPoint = L.point(lat, lng),
```

PR #3375, opened 2015-04-14
Add mobile touch events

lee101:
Adds support for adding touch events "touchstart touchmove touchend".

Events can be added with `L.DomEvent.addListener(map, "touchstart", touchStartHandlerFunction);`

I found the need for these events in leaflet while developing a drag and drop "orange man" style leaflet control so people could select a point on the map.

Its interesting to note that that touchend isn't the same as mouseup, as i understand it mouseup fires on the element where the mouse was released and touchend fires on the element where touchstart occurred.

Something strange i noticed was that touchend isn't propagated if touchstart's on leaflet controls so i needed to attach a touchend event to the leaflet control and figure out the lat,lng using `map.mouseEventToLatLng(event)` (as oppose to attaching touchend to the map and doing `event.latlng`)

a shim to add them is described in #1542  
also related to #3075

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -172,15 +172,19 @@ L.DomEvent = {
 	},
 
 	getMousePosition: function (e, container) {
+		var evt = e;
+		if (evt.touches) {
+			evt = evt.touches[0] || evt.changedTouches[0];
+		}
 		if (!container) {
-			return new L.Point(e.clientX, e.clientY);
+			return new L.Point(evt.clientX, evt.clientY);
 		}
 
 		var rect = container.getBoundingClientRect();
 
 		return new L.Point(
-			e.clientX - rect.left - container.clientLeft,
-			e.clientY - rect.top - container.clientTop);
+			evt.clientX - rect.left - container.clientLeft,
+			evt.clientY - rect.top - container.clientTop);
 	},
 
 	getWheelDelta: function (e) {
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -607,7 +607,7 @@ L.Map = L.Evented.extend({
 		var onOff = remove ? 'off' : 'on';
 
 		L.DomEvent[onOff](this._container, 'click dblclick mousedown mouseup ' +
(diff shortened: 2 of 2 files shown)


mourner:
Sorry for a late response! I've been meaning to add something like this a long time ago. Need some time to think about this though.

src/dom/DomEvent.js
```diff
@@ -172,15 +172,19 @@ L.DomEvent = {
 	},
 
 	getMousePosition: function (e, container) {
+		var evt = e;
+		if (evt.touches) {
+			evt = evt.touches[0] || evt.changedTouches[0];
```
IvanSanchez: I don't think this is a good idea. After all, the function is called get**Mouse**Position and not getEventPosition.

I presume there will be edge cases where the first touch originates outside the map container, then a second one inside, but `getMousePosition` uses only the first one.

I would argue that a new getTouchPosition**s** function is needed, to translate all the touch points into map coordinates.

Until then, it should be fine to expose just `ev.originalEvents` (as with `keypress` events, see `L.Map._fireDomEvent()`) and let the upper functionality decide which touches (if any) should get more attention and if whether to do some `containerPointToLatLng` calls or whatever.

IvanSanchez: Or maybe make sub-cases of `_fireDOMEvent` named `_fireMouseEvent`, `_fireTouchEvent` and so on.

alexkavon:
Perhaps utilize something like hammerjs?

IvanSanchez:
> Perhaps utilize something like hammerjs?

Not sure that's a good idea either. HammerJS might interfere with `Map.Tap.js` and alternative libraries used by developers including Leaflet in their webpages.

---

I'm not convinced by this PR - I think managing coordinate transformations of multi-touch events has to be more complete, and AFAIK one-touch events are emulated properly as mouse drags.

IvanSanchez:
After the `PointerEvent` updates in chrome 55, I don't think this is worth anymore. A better idea would be to switch everything to `PointerEvent` anyway.

alexkavon:
That wouldn't really be friendly to iOS devices...though perhaps I'm misunderstanding?

http://caniuse.com/#feat=pointer

PR #3386, opened 2015-04-18
Fix L.Polyline._flat returning true for a nested empty array

yohanboniface:
`L.Polyline._flat` was returning true for `[[]]`, which seems to me a valid use case of an empty nested latlngs array (case of a Polygon).

While I'm on it, I've some use cases where I would like to call `L.Polyline._flat` without being in a context of a given `Polyline` instance, what do you think about moving `_flat` as a class method instead of an instance method (i.e. `L.Polyline._flat` instead of `this._flat` to use it)?

### DIFF
--- a/spec/suites/layer/vector/PolylineSpec.js
+++ b/spec/suites/layer/vector/PolylineSpec.js
@@ -88,4 +88,29 @@ describe('Polyline', function () {
 
 	});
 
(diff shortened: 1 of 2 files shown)


mourner:
Hmm yeah, I think `L.Polyline._flat` static method makes sense.

PR #3390, opened 2015-04-21
Add LatLng.clone

yohanboniface:
From time to time, I need to clone a LatLng instance, so I'm thinking it can make sense to have an util in the core. What do you think?


mourner:
Yes, no harm from that. :)

PR #3392, opened 2015-04-21
Implement L.Map.flyToBounds(), #3391

### LINKED ISSUE #3391: Implement map.flyToBounds(bounds, options)
I'd like the feature to fly to bounds instead of a target center and zoom. I think this will be pretty easy to implement because all helper methods are already there and we could even use the flyTo function in the end to get the flying effect.

This is how my code looks now with out the flyToBounds. It would be nice to abstract away all the calculations.

```js
var bounds = geojson.getBounds();
var zoom = map.getBoundsZoom(bounds);
var swPoint = map.project(bounds.getSouthWest(), zoom);
var nePoint = map.project(bounds.getNorthEast(), zoom);
var center = map.unproject(swPoint.add(nePoint).divideBy(2), zoom);
map.flyTo(center, zoom);
```

Note that this does not account for padding which the flyToBounds method should in order to be consistent with the map.fitBounds(bounds, options) method.

IvanSanchez:
Abstracted some of the `Map.fitBounds()` code so that `Map.flyToBounds()` can reuse that bit. Should close #3391.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -88,7 +88,7 @@ L.Map = L.Evented.extend({
 		return this.setView(newCenter, zoom, {zoom: options});
 	},
 
-	fitBounds: function (bounds, options) {
+	_getBoundsCenterZoom: function (bounds, options) {
 
 		options = options || {};
 		bounds = bounds.getBounds ? bounds.getBounds() : L.latLngBounds(bounds);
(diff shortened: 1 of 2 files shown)


mourner:
Looks good, thanks!

IvanSanchez:
Should I add `flyToBounds()` to the docs?

ErikOrjehag:
thanks!

mourner:
@IvanSanchez yep, `gh-pages-master` branch

PR #3394, opened 2015-04-22
Add flyTo() and flyToBounds() to the API reference

IvanSanchez:
As discussed in #3392.


mourner:
> The map will be zoomed out until both the current map center and the target are visible

Not sure this is actually the case. I think the algorithm doesn't set constraints like this; it's just a smooth zoom-pan curve, and how much it zooms out is controlled by another param inside the method code.

IvanSanchez:
Changed the wording and squashed the commit. I will not overthink the text.

PR #3409, opened 2015-04-29
Move functionality from point, latlng, bounds factory to constructor

### LINKED ISSUE #3408: "L.latLngBounds" behavior differs from "new L.LatLngBounds"
If I want empty bounds, I can write

```
var bounds = new L.LatLngBounds(); //returns empty bounds
```
but can't write
```
var bounds = L.latLngBounds(); //returns undefined
```

Is it intended behavior or a bug?

IvanSanchez:
Constructors for point, latlng, bounds, latlngbounds behave just like the factories.

Fixes #3408.

Note that `new L.LatLng(50);` will raise an error and `L.latLng(50)` will return null. This is as per the current specs.

### DIFF
--- a/src/geo/LatLng.js
+++ b/src/geo/LatLng.js
@@ -2,7 +2,30 @@
  * L.LatLng represents a geographical point with latitude and longitude coordinates.
  */
 
-L.LatLng = function (lat, lng, alt) {
+// constructs LatLng with different signatures
+// (LatLng) or ([Number, Number]) or (Number, Number) or (Object)
+L.LatLng = function (a, b, c) {
+	var lat, lng, alt;
+	if (L.Util.isArray(a) && typeof a[0] !== 'object') {
+		if (a.length === 3) {
+			lat = a[0];
+			lng = a[1];
+			alt = a[2];
+		}
+		if (a.length === 2) {
+			lat = a[0];
+			lng = a[1];
+		}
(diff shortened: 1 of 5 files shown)


src/geo/LatLng.js
```diff
@@ -53,30 +76,13 @@ L.LatLng.prototype = {
...
-	if (a === undefined || a === null) {
-		return a;
-	}
-	if (typeof a === 'object' && 'lat' in a) {
-		return new L.LatLng(a.lat, 'lng' in a ? a.lng : a.lon, a.alt);
-	}
-	if (b === undefined) {
+	try {
+		return new L.LatLng(a, b, c);
+	} catch(err) {
```
mourner: Again, try/catch here is very harmful. To see the effect of this, try doing benchmarks of old vs new code on jsperf.com and you'll see why I am concerned. It's also generally a good idea to benchmark sensitive code like LatLng construction when doing any changes.

mourner:
I'm actually not sure we should move out the coercion logic from factories — it makes sense there, keeps constructors simple and focused and we can just close issues like #3408 with "Sorry, it was intentional API decision. Please use factories instead of new Class signature". This becomes more of a docs issue then.

src/geo/LatLng.js
IvanSanchez: Note to self: learn to use jsperf.com, then give this another shot.

PR #3411, opened 2015-04-29
Prevent MSIE from focusing a tilelayer container.

### LINKED ISSUE #3314: IE loses focus after panning twice
Works fine in Firefox Safari and Chrome.

If you click and drag to pan, the map gets focus, if you do it again the map loses focus in IE.

IvanSanchez:
Fixes #3314. For some weird reason MSIE allows `leaflet-tile-container`s to grab focus.

I don't know if this might have any side effects.


mourner:
Not a very nice hack... It'd be better to figure out why tile layer container gets focus in the first place. Is the behavior the same for 0.7.3 (e.g. can we bisect that)?

mourner:
Closing until we find a better way around this (tracking in #3314).

PR #3415, opened 2015-04-30
TileLayer - Use correct tileSize when checking tile bounds

jlaxson:
When using options.bounds in combination with options.maxNativezoom,
the tileSize used in _tileShouldBeLoaded can be incorrect, causing
tiles to not get loaded or not get displayed.


mourner:
Thanks!

mourner:
Does master need a similar change or is it fixed there already?

jlaxson:
From reading TileLayer and GridLayer, it looks to me like master should behave correctly, but I haven't actually run that version.  Thanks for the quick merge!

PR #3419, opened 2015-04-30
Detect Firefox for Android as L.Browser.mobile

IvanSanchez:
Firefox for Android doesn't implement `window.orientation` because of https://bugzilla.mozilla.org/show_bug.cgi?id=920734 (so the current `L.Browser.mobile` code ignores it) , but it adds "mobile" to the user-agent string.


IvanSanchez:
Should we backport this to stable?

mourner:
Probably not critical to backport.

Issue #3424, opened 2015-04-30
Empty coordinate-array in LineString is not handled


hs06:
When importing a kml (via omnivore) of which a segment with empty coordinate set (<coordinates></coordinates> exists, leafletjs will crash (coordinates NaN).

the problem exits in 	coordsToLatLngs: function (coords, levelsDeep, coordsToLatLng) { // (Array[, Number, Function]) -> Array

I fixed the problem by throwing an own exception and catching it here:
L.extend(L.GeoJSON, 
in 		case 'GeometryCollection':
			for (i = 0, len = geometry.geometries.length; i < len; i++) {

with this:

                        try {
				newGeo = this.geometryToLayer({
					geometry: geometry.geometries[i],
					type: 'Feature',
					properties: geojson.properties
				}, pointToLayer, coordsToLatLng, vectorOptions);
				layers.push(newGeo);
                        }
                        catch (e)
                        {
                                console.log("Coordinate error: "+e);

                        }

Hope that helps. I'm using stable / 0.7.3

hs06:
PS: I'm not doing a fork/pull request as 1) i'm not really into git yet and 2) i'm not sure how the throw/catch should be best handled. my fix works for me, but it should be up to those who know and maintain leatletJS to judge and apply the patch. cheers! -hendrik

IvanSanchez:
Can you please provide a sample KML file, or a GeoJSON object, which triggers the bug?

hs06:
sure. here it is.


````
[...]

PR #3426, opened 2015-05-04
Verbosier GeoJSON errors

IvanSanchez:


mourner:
I'd rather avoid try/catch blocks because they harm performance (functions with try/catch are never optimized).

IvanSanchez:
I cannot think of a cleaner way to provide meaningful messages as per #3424, specially when users are able to run stuff like `L.geoJSON(JSON.parse(fileContents))` bypassing Omnivore. 

Otherwise, it's duplicating the sanity checks in the `L.LatLng`, `L.Polyline` and `L.Polygon` constructors to avoid the `try`-`catch`. Do you think that'll be better?

mourner:
Hard to say, maybe we should go on completely without sanity checks. To completely guard off GeoJSON problems, the checking code should be more complicated and there are already tools like http://geojsonlint.com/. We can't do full-blown checks everywhere, otherwise it will bloat Leaflet and make it slower. It's also sometimes easier to leave an issue to app developers than do a half-solution.

Issue #3432, opened 2015-05-05
Map does not return to max bounds on minor/slow panning


rikvanmechelen:
On 0.8-master branch, when you pan slowly, slightly diagonally the map does not snap back to its max bounds.

cduzan:
I'm getting a similar issue. I'm not sure if it's diagonal, but the map only snaps back if you drag it quickly and release. If I make sure I stop moving before releasing the mouse then the map doesn't snap back.

PR #3433, opened 2015-05-05
fix for #3432

rikvanmechelen:
- we removed the .round(), since it did not handle small numbers very well
- we made the test a conjunction, since it should snap back even when only moving on one axis


mourner:
Thanks, will review this in a moment.

mourner:
Question, why do you remove `round`? Can you describe a use case where rounded offset fails?

rikvanmechelen:
If i remember correctly, we did not want it to round down to 0, and it worked fine without it.

mourner:
Commited as https://github.com/Leaflet/Leaflet/commit/c4b6e79fa4ab8f7b20e71ed89a4e7f2917f38cca (without the round change), thanks

PR #3435, opened 2015-05-07
 Grouping layers into categories.

WilliamTokarev:
Added ability to mark a category layers. Name the first layer (specified category) contains the name of the category (before the symbol #).
Example: ... 'SAT#Google': g01, ...

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -498,3 +498,11 @@
 	background: #fff;
 	border: 1px solid #666;
 	}
+
+.leaflet-base-layers-label {
+	margin-top: 8px;
+	margin-bottom: 4px;
+	font-size: 10px;
+	text-align: center;
+	background: #e8e8e8;
+	}
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -178,6 +178,20 @@ L.Control.Layers = L.Control.extend({
 	},
 
 	_addItem: function (obj) {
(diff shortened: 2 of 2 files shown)


mourner:
Thanks for the pull! I'm currently not merging any features before Leaflet 1.0 release to allow it to stabilize, but will review this afterwards.

IvanSanchez:
I'd rather see this in a plugin that would subclass L.Control.Layers into something like L.Control.Layers.Categorized.

WilliamTokarev:
Edit small. Whether it is necessary to allocate its plugin. Symbol # is questionable from a layer name (e.g., # to | or @ can be replaced).

IvanSanchez:
Given that users can use the [Leaflet.groupedlayercontrol](https://github.com/ismyrnow/Leaflet.groupedlayercontrol) plugin, I see no reason for this merge request.

WilliamTokarev:
In the Leaflet.groupedlayercontrol plugin "Grouping base layers is not currently supported"...

IvanSanchez:
Then try https://github.com/davicustodio/Leaflet.StyledLayerControl, https://github.com/robbiet480/leaflet-categorized-layers or https://github.com/stefanocudini/leaflet-panel-layers, fix Leaflet.groupedlayercontrol, report a bug/feature request to Leaflet.groupedlayercontrol, or create your own plugin.

PR #3438, opened 2015-05-07
Reorganized plugin categories

### LINKED ISSUE #3414: Plugins page: split in categories
As is pointed out in #3240, the plugins page is getting out of hand. An idea to fix this would be to split plugins into more categories, e.g.:

- Composition/transformation layers (works with other layers)
- Data Viz layers (works with data)
- Markers & UI layers
- Tile layer sources
- Vector layer sources
- Geocoding
- Routing
- Visual controls
- Non-visual controls
- Controls for 3rd party services

An index to jump to the different plugin categories would also be great.

IvanSanchez:
This should fix #3414. I don't know how I ended up with 35 (gasp!) categories. Also, TWO HUNDRED plugins?!?!!

An ugly preview can be easily seen at https://github.com/IvanSanchez/Leaflet/blob/gh-pages/plugins.md

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -5,36 +5,594 @@ title: Plugins
 
 ## Notable Leaflet Plugins
 
-While Leaflet is meant to be as lightweight as possible, and focuses on a core set of features, an easy way to extend its functionality is to use third-party plugins. Thanks to the awesome community behind Leaflet, there are lots of nice plugins to choose from.
+While Leaflet is meant to be as lightweight as possible, and focuses on a core set of features, an easy way to extend its functionality is to use third-party plugins. Thanks to the awesome community behind Leaflet, there are literally hundreds of nice plugins to choose from.
 
 ---
 
-### Layers and Overlays
+## <a name='index'></a> Index
 
-<table class="plugins">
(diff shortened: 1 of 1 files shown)


mourner:
Holy shit!

mourner:
Let me clone this and take a look live... This looks great!
The only thought I have now is that Leaflet users are not used to complicated terminology like "symbolizers", and also markers and icons are probably not perceived as "vector layers". Maybe change "Vector display" category wording to "Overlays", "Custom overlays"?

mourner:
Maybe we should also simplify this a bit, since too many categories also make this harder to skim — too much visual noise. Also, the table of contents needs some UI treatment so that it takes several columns and looks nice like the API reference ToC.

IvanSanchez:
The ToC looks sooooo much nicer with a 4-column layout. IMHO the amount of categories doesn't clutter the ToC - *personally* I see no reason to fuse categories.

mourner:
That looks awesome. Merging, further improvements may come as separate PRs.

IvanSanchez:
Cool! :+1:

Issue #3430, opened 2015-05-05
Handle 'focus' and 'blur' in L.Evented


IvanSanchez:
If we want to attach a handler to the `focus` event (or the `blur` event) to a marker, the way to do it right now is `L.DomEvent.on(marker._icon, 'focus', doSomething)`. Even though it's pretty much syntactic sugar, it'd be nice to be able to write `marker.on('focus', doSomething)` instead. (see #3429)

I've seen that `L.Marker` delegates all events to its icon via `L.Layer.addInteractiveTarget`, but right now I cannot understand all the interlocking parts of the code that make events work. It seems that the `L.Map` captures all (possible) events within and then delegates the handler, but the `focus` event doesn't bubble so it's not captured by the code in `L.Map._findEventTarget` and so on.

Should `L.Evented._on` handle `focus` and `blur` (and/or other non-bubbling events) differently to attach these directly to their target instead of adding it to the targets list? I'm unaware of the performance implications.

Also: 
[ ] add `focus` and `blur` as possible events triggered from a `L.Map`, at `L.Map._initEvents`. 
[ ] skip the `data.latlng` calculations in `L.Map._fireDOMEvent`.

mourner:
This is tricky. I implemented delegated events so that we don't actually attach DOM listeners to every marker, which slows things down considerably when dealing with lots of markers. If we have that in place already, I wouldn't want to add attaching listeners directly again. On the other hand making a special treatment in L.Evented to add listeners lazily adds complexity to an already comlex system, adding headache and potential source of bugs. So I'd punt on this for now.

IvanSanchez:
I wonder if using `useCapture` set to true on `addEventListener` will solve this **and** not cause a performance hit. I might give it a shot.

mourner:
See https://github.com/Leaflet/Leaflet/pull/3444#issuecomment-109070911, will reopen in case there will be a strong need for this.

Taekjoon:
ㅗㅕㅍ
숏다리?ㄷ

PR #3444, opened 2015-05-08
Add 'focus' and 'blur' event handling

IvanSanchez:
My shot at implementing #3430, using Leaflet's event delegation to make Vladimir happy and keep the code simple.

@mourner: Do you think that forcing `useCapture` to true in AddEventListener is OK? If not, I could try to turn this into a plugin.

### DIFF
--- a/debug/map/marker-focus.html
+++ b/debug/map/marker-focus.html
@@ -1,0 +1,50 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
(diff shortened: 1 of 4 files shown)


mourner:
On the surface, that looks like an excellent solution (minimal code intrusion is always good). But I assume it won't work in IE8 and less, right? Is there a painless fallback? Also are there any caveats to using `useCapture`?

IvanSanchez:
> Also are there any caveats to using useCapture?

Apparently this solution runs `addEventListener(document,'keydown')` every time I tab through the markers (thus spiking CPU usage). I need to see what's going on.

> I assume it won't work in IE8 and less, right? Is there a painless fallback?

Yes and no (respectively). I know of no way to capture focus events on IE8 and no easy callback (short of going with my original idea of attaching an event handler to every marker's icon)

mourner:
Keeping things simple for now. Given that we now have `layer.getElement()`, it's easy to workaround.

TomazicM:
Since this was not implemented, what is the workaround using `layer.getElement()`? I attached `focus` event function to marker element, but it did not get fired.

PR #3445, opened 2015-05-08
Make L.Polyline._flat a static method

yohanboniface:
As discussed in https://github.com/Leaflet/Leaflet/pull/3386#issuecomment-99764041 :)

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -115,7 +115,7 @@ L.Polyline = L.Path.extend({
 	// recursively convert latlngs input into actual LatLng instances; calculate bounds along the way
 	_convertLatLngs: function (latlngs) {
 		var result = [],
-		    flat = this._flat(latlngs);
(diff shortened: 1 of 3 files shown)


src/layer/vector/Polyline.js
```diff
@@ -228,3 +223,12 @@ L.Polyline = L.Path.extend({
...
 	return new L.Polyline(latlngs, options);
 };
+
+L.extend(L.Polyline, {
+
+	_flat: function (latlngs) {
+		// true if it's a flat array of latlngs; false if nested
+		return !L.Util.isArray(latlngs[0]) || (typeof latlngs[0][0] !== 'object' && typeof latlngs[0][0] !== 'undefined');
+	}
+
```
mourner: No need to use `extend` if we're only adding one function. `L.Polyline._flat = ...`.

yohanboniface:
Done :)

PR #3446, opened 2015-05-08
Fix vector in ie8

ilfa:
In ie8 vector layers don't work.
Problem was in legacy code in SVG.VML.js file.

For test:
debug/map/popup.html
debug/tests/svg_clicks.html


mourner:
Awesome!

PR #3449, opened 2015-05-08
Update L.Polyline.addLatLng doc

yohanboniface:
Thoughts? :)

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -2615,11 +2615,12 @@ var latlngs = [
 	</tr>
 	<tr>
 		<td><code><b>addLatLng</b>(
-			<nobr>&lt;<a href="#latlng">LatLng</a>&gt; <i>latlng</i> )</nobr>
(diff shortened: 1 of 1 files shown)


mourner:
> but can be overriden passing another LatLngs group

I think we need to say this more clearly, e.g. "can be overridden by passing a specific ring as a `LatLng` array (that you can earlier access with `getLatLngs`)".

Also forgot to remove `spliceLatLngs` :)

mourner:
Also, do we need to adjust `getLatLngs` description? Is it always clear what gets returned?

yohanboniface:
> I think we need to say this more clearly, e.g. "can be overridden by passing a specific ring as a 
> LatLng array (that you can earlier access with getLatLngs)".

Thanks! My English is terrible…

> Also forgot to remove spliceLatLngs

Meeeeeeh :p

> do we need to adjust getLatLngs description? Is it always clear what gets returned?

Made a suggestion :)

PR #3450, opened 2015-05-08
Add missing event parent calls for L.Layer - fixes #3428

### LINKED ISSUE #3428: vector layers stop double click from propagating to map
Currently doubleClicking for zooming in and doubleClicking + shift for zooming out does not work when double clicking above a geoJSON layer.  This issue is visible in the second map in the Choropleth example - http://leafletjs.com/examples/choropleth.html . I am not yet sure how to fix it, so the default actions occur.

nathancahill:
I think this is the right way to fix it.

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -54,6 +54,8 @@ L.Layer = L.Evented.extend({
 			map.on(this.getEvents(), this);
 		}
 
+		this.addEventParent(map);
+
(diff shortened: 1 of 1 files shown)


mourner:
Thanks for the PR! This makes it propagate Leaflet events, which may be a correct solution, but first I'd want to check why the underlying DOM events don't propagate in the first place. They should in theory.

PR #3454, opened 2015-05-09
Fix polygon.addTo(map) failing when latlngs was empty.

yohanboniface:
Since #3279 an empty polygon as a nested array [[]] as latlngs,
so testing _latlngs.length returns true.

I'm thinking that is a breaking change: I guess many users rely on `this._latlngs.length` to check if a polygon has some latlngs. So we need to make this very clear in the change log.

What about maybe a `hasLatLngs` method, to have a clean API instead of this kind of widespread usage based on a private property?


mourner:
Yeah, a public method sounds good. Maybe a better name would be `isEmpty()`.

Issue #2609, opened 2014-04-03
panto().panby() not chained correctly


Guillaumez:
Hi,
i'm on Mapbox 1.6.2 and when i when i try to chain `map.panTo({...}).panBy([...]);` it works well on firefox, chrome, safari but not on IE8.

IE8 does directly the panBy without taking care of the panTo coordinates.

I've tried to solve by converting latLng to Point etc... but it seems to still not work.

mourner:
Can someone confirm this with a JSFiddle?

MuellerMatthew:
I am having a similar issue with IE9. I have 
`map.panTo(e.latlng, map.zoom, {animate:true,duration:.25}).panBy([LeftOffset/-2,0],{animate:true, duration:.25});`
Which I am using to shift the marker right because I have an overlay over the left side of the map, and I want to move the marker to the visible center of the map, and in most browsers it seamlessly pans to the final location.  What is happening is that in IE9, the panBy animation and the panTo animation don't chain, so the entire map just keeps shifting to the right.
I tested it with the setView function chained also and it is the same issue:  
`map.setView(e.latlng, map.zoom, {animate:true,duration:.25}).panBy([LeftOffset/-2,0],{animate:true, duration:.25});`

The only way that IE positions the marker in the correct position is when I turn off the panTo or setView animation which runs first, but this means that users don't know which way the map moved.  I think the issue is with how leaflet is blocking the animations because they seem to both be running at the same time in IE9.
I even tried calling the second functions using `setTimeout()` to delay the panBy function from running until the first one is completed, but that just causes the map to pan back and fourth instead of straight to the final position.

I have been poking around the Leaflet code trying to figure out if adding callbacks in a few of the sub function might fix the issue, but so far I don't have a fix for the bug.

MuellerMatthew:
here is a fiddle: http://jsfiddle.net/ofs291vp/ with the two functions chained. 

I have the map animating to reset the view to the same center point several times, and then shift over the same distance with animation turned on and with it turned off.  Both options should have the map positioned in the same spot when they are done, but they don't when the animation is turned on.

johnd0e:
@Falke-Design 
So you have fix at least for some of mentioned issues.
Why not open PR and then deal with the rest of issues?

Falke-Design:
Because I thought that we should first discuss if it is good the create a loop with `setTimeout`:
```
			const runAnimation = ()=>{
				if (!this._panAnim._inProgress) {
[...]

PR #3460, opened 2015-05-11
Graceful degradation of zoom-pan animations, fixes #3272.

### LINKED ISSUE #3272: IE7-9 graceful degradation for animations & fractional zoom

IvanSanchez:


mourner:
Great! Do you think that covers everything or any broken issues left? What IEs did you test?

IvanSanchez:
I tested only IE8 in a VM, and IE11 on IE8 mode.

mourner:
@IvanSanchez lets also make sure IE9 is good.

IvanSanchez:
I just confirmed that the code at `debug/map/zoompan.html` works as expected in IE7, IE8 and IE9 :smile: 

Obviously there's still #2609, but that's a completely different issue, I guess.

mourner:
:+1: awesome!

PR #3462, opened 2015-05-12
flyTo() events now firing only when appropiate.

### LINKED ISSUE #3134: flyTo generates many moveend events

IvanSanchez:
Should fix #3134, even if the code is a tiny bit hackish IMHO.

### DIFF
--- a/debug/map/zoompan.html
+++ b/debug/map/zoompan.html
@@ -33,7 +33,12 @@
 		<div><button id="lnd">LND</button>(fract. zoom)</div>
 		<div><button id="kyiv">KIEV</button>(setView, fract. zoom)</div>
 		<div><button id="stop">stop</button></div>
-		<div id='pos'></div>
+		<table>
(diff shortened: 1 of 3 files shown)


mourner:
Thanks! I rewrote the confusing logic in `_resetView` in #3278, I guess we need to coordinate the changes before I can merge this. Can you look at that PR and see if those changes make sense?

IvanSanchez:
Let me rebase this into `events-refactor` and see how much stuff breaks down.

mourner:
@IvanSanchez it will probably be easier to review the changes before rebasing, the latter would get confusing without background.

IvanSanchez:
Not needed once #3278 gets merged, events are fired as expected in the `events-refactor` branch.

PR #3464, opened 2015-05-12
Getter/setter compatibility

akauppi:
Current Leaflet `L.Class.extend` does not work well if the derived class uses ECMAscript 5 get/set methods. This PR fixed that, based on [this blog from 2007](http://ejohn.org/blog/javascript-getters-and-setters/).

Why does this matter?

Getters/setters seem like a nice addition to JavaScript and the Leaflet class extension mechanism can be made to be compatible with them.

There are ways to bypass the current limitation (see [SO](http://stackoverflow.com/questions/30197434/javascript-getter-why-does-it-get-called-at-construction)) but those are kind of clumsy, causing get/set code to end up separated from normal methods.

Are there downsides?

- the suggested implementation does not check for browser capabilities. This should be added. Older Leaflet-supported browsers would not know of getters/setters. But this is doable.

- `jake test` now expects support for the get/set syntax from the underlying V8 engine. It works for mine (node 0.12.2) but I'm not sure about older node versions. Don't think this would be a major concern, though.

To sum up. This PR allows the library user to use `L.Class.extend` mechanism, together with his/her own code utilizing ECMAscript 5 getters/setters in his/her own code. It does not suggest Leaflet should use getters anywhere. I'd like to hear comments on the value of such a change - i.e. have others come across this current limitation.

### DIFF
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -169,6 +169,30 @@ describe("Class", function () {
 
 			expect(spy1.called).to.be.ok();
 		});
+
+		it("allows class(es) to have getter methods", function () {
+			var Klass2 = Klass.extend({
+				initialize: function (v) {
+					this._goo = v;
+				},
+				get goo () {
+					if (!this._goo) { throw "'initialize' should have run first!"; }   // using assert would need 'require("assert")';
+					return this._goo * 2;
+				}
+			});
+			var a = new Klass2(9);
+
+			expect(a.goo).to.eql(9 * 2);
+
+			var Klass3 = Klass2.extend({
+				initialize: function() {
+					Klass2.prototype.initialize.call(this, 11);
+				}
+			});
+			var b = new Klass3();
+
+			expect(b.goo).to.eql(11 * 2);
+		});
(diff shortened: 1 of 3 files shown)


mourner:
Thanks for the PR! While it is certainly useful in some cases, my main concern about this change is performance. `L.extend` is very critical function, it's used throughout the whole Leaflet code in many places, and takes up considerable time when dealing with many layers according to profiling results. This change will hit performance a lot, while satisfying a very rare use case, so it's something I can't afford in the core.

However, this can easily be released as a plugin! It would override `L.Util.extend`, accomplishing the same.

akauppi:
Sure, @mourner. 

I mostly just wanted to "report" the surprise that getters/setters didn't work.

Also I was worried about the performance implications - and I tremendously value you keeping a keen eye on those.

However, if you ever wish to have support for them, there's another way (came to think of it after filing the PR). One could have `.props` similar to how `.statics`, `.includes` etc. currently are - i.e. have the properties cut down to their own area, which should help with the performance aspect, yet fit nicely (I think) to the overall Leaflet way.

For now, nothing needs to be done.

PR #3465, opened 2015-05-12
Added debug page for infinite horizontal bounds

IvanSanchez:


mourner:
LOL :) btw why `Number.POSITIVE_INFINITY` and not just `Infinity`?

IvanSanchez:
> why `Number.POSITIVE_INFINITY` and not just `Infinity`?

Because deep down inside my head there was a time where I had problems using plain `Infinity` in some code somewhere. :smile:

mourner:
forget about that time, it was probably a nightmare :)

PR #3476, opened 2015-05-18
Add plugin to manage oriented markers dynamically.

alexandreDavid:


IvanSanchez:
Nice!

Suggestions, though:
* Use github pages to provide a live example (cloning to see the examples working is cumbersome)
* Demo how to create a non-interactive marker with a predefined angle
* Check UTF-8 in your examples - the © is in ISO-8859-1 instead.

IvanSanchez:
Also, following Leaflet's convention on naming subclasses, this should be named `L.Marker.Oriented` instead.

PR #3488, opened 2015-05-22
Persist marker dragging when removing/readding it.

IvanSanchez:
This *should* persist runtime changes to the draggable state of a marker, in the case of removing it from the map and adding it later, but I haven't really tested it.


txp314:
works like a charm! Thanks

mourner:
A bit weird, but apparently works so merging.

PR #3489, opened 2015-05-25
Make panInsideBounds method accept bounds in array form

RLRR:
See https://github.com/2gis/mapsapi/issues/200. `panInsideBounds` method failed if bounds were passed as a simple array.


mourner:
:+1:

Issue #3490, opened 2015-05-26
Problem with long panTo in IE8


Trufi:
Short demo:
```html
<div id="map" style="width:1000px; height:400px"></div>
<script>
    var map = L.map('map', {
        center: [54.9802, 82.8980],
        zoom: 18
    });

    L.tileLayer('http://{s}.tile.osm.org/{z}/{x}/{y}.png').addTo(map);

    setTimeout(function() {
        map.panTo([55.75253,37.62298]);
    }, 1500);
</script>
```

After `panTo` tiles are grouped into one in ie8:
![screenshot from 2015-05-26 16 30 22](https://cloud.githubusercontent.com/assets/3996552/7810864/2f5c2212-03bf-11e5-8e9a-07f273f7d187.png)
Leaflet applies `style.left = '8439431px'` attribute to tile image, but `style.left` is limited to 1342177 pixels in ie8.

IvanSanchez:
This same problem is giving us some headaches on https://github.com/Leaflet/Leaflet/pull/3278#discussion-diff-30127405 - 

In other words, I doubt this bug can be approached without finishing work on #3278 first.

Have you tried to run `map.setView()` instead of `map.panTo()`? That should trigger a view reset in current master.

> style.left is limited to 1342177

I'm curious - where does this limit come from? This doesn't look like 2^n or anything that would make sense.

@mourner Stupid crazy idea: maybe apply a modulo to the tile pixel coordinates?

Trufi:
Yes, I tried `map.setView` and `map._resetView` directly, but same situation.

mourner:
For a workaround, try setting `reset: true` option http://leafletjs.com/reference.html#map-zoompanoptions

Trufi:
Unfortunately, it didn't help.

mourner:
Should be fixed by #3278

PR #3491, opened 2015-05-26
Fix for long panTo in IE8

Trufi:
Issue https://github.com/Leaflet/Leaflet/issues/3490

Not sure that this is a correct solution, but it solves the problem.


mourner:
I'll review this again after merging #3278 and will merge if all's ok.

mourner:
In theory, this should be unnecessary after I merged #3278 — can you check this?

Trufi:
You're right, the bug was gone.

PR #3495, opened 2015-05-27
Document L.Path.options.fillRule.  Fixes #3478.

### LINKED ISSUE #3478: Document L.Path.options.fillRule
L.Path has an undocumented `fillRule` option with two possible values, added in #2834.

This needs to be briefly documented in the `gh-pages-master` branch and optionally in `gh-pages` also.

garnertb:


garnertb:
You can view the change here: http://garnertb.io/Leaflet/reference.html#path before merging.

mourner:
@IvanSanchez hmm, this rule was added in master, not stable, right?

IvanSanchez:
No, the rule exists on stable, but only for the Canvas renderer:
```
~/devel/Leaflet/src\ [stable]$ grep -ri fillRule
layer/vector/canvas/Path.Canvas.js:                     ctx.fill(options.fillRule || 'evenodd');
```

Issue #3281, opened 2015-03-05
Layer isPopupOpen method


piotreksenel:
Hi,

I wanted to open popup for `circleMarker` on `mouseover`, it was working correct for the `marker`, but with the `circleMarker` whenever we move mouse inside of it the popup is closed and reopened. I tried to use `mouseenter` and I found that there is a code responsible for handling that event but it's not fired, so are we suppose to use this event?

Cheers

pufanalexandru:
If you want to prevent the popup from opening again and again, you can use the following code:
    
    //create the circleMarker
    var circle = L.circleMarker([latlng].addTo(map);

    //Bind the desired popup to the circleMarker
    circle.bindPopup('Desired content');

    //Use mouseover event to open the popup, if it's not already opened
    circle.on('mouseover', function () {
        if (!circle._popup._isOpen) {             //this is important, as it checks to see if the popup is already opened
            this.openPopup();
        }
    });

piotrdonicz:
I used `circle._popup._isOpen` as a temporary solution, I would prefer not to use private properties and wondering if there is an option to solve that in a different way

PR #3499, opened 2015-05-27
add isPopupOpen method to layer

vingiarrusso:
Addresses #3281 by adding isPopupOpen method.

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -36,7 +36,7 @@ L.Layer = L.Evented.extend({
 	},
 
 	isPopupOpen: function() {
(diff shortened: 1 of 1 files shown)


src/layer/Layer.js
```diff
@@ -35,6 +35,10 @@ L.Layer = L.Evented.extend({
 		return this;
 	},
 
+	isPopupOpen: function() {
+		return !!(this._map._popup && this._map._popup.isOpen);
```
yohanboniface: `popup.isOpen()` is a function, IIRC. ;)

vingiarrusso: You're absolutely right, thank you.  :)

Issue #2981, opened 2014-10-28
Leaflet tries to get negative value images or images that don't exist


donaldallen:
Here is my Leaflet config:

```javascript
function init() {
    var mapMinZoom = 2;
    var mapMaxZoom = 5;
    var map = L.map('map', {
        maxZoom: mapMaxZoom,
        minZoom: mapMinZoom,
        crs: L.CRS.Simple
    }).setView([0, 0], mapMaxZoom);

    var mapBounds = new L.LatLngBounds(
        map.unproject([0, 4352], mapMaxZoom),
        map.unproject([2048, 0], mapMaxZoom));

    map.fitBounds(mapBounds);
    L.tileLayer('/assets/themes/sec/img/tiles/{z}/{x}/{y}.png', {
        minZoom: mapMinZoom,
        maxZoom: mapMaxZoom,
        bounds: mapBounds,
        noWrap: true,
        continuousWorld: true
    }).addTo(map);

    var store_marker = L.marker([0, 0]).addTo(map);
}
```

Everything works, but I get a lot of "resource not found" in my console whenever I zoom or pan:

```console
[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (0.png, line 0)
http://localhost:3000/assets/themes/sec/img/tiles/2/-1/0.png

[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (0.png, line 0)
http://localhost:3000/assets/themes/sec/img/tiles/2/1/0.png

[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (1.png, line 0)
http://localhost:3000/assets/themes/sec/img/tiles/2/-1/1.png

[Error] Failed to load resource: the server responded with a status of 404 (Not Found) (1.png, line 0)
http://localhost:3000/assets/themes/sec/img/tiles/2/1/1.png
```

I have no negative numbered tiles. The files referenced here don't exist. Yet everything looks fine on the map. I used MapTiler to build the map. Anything I can do to fix this?

SundialServices:
[...]

PR #3508, opened 2015-06-02
Implement 'overlaps' spatial operator on bounds and use it on gridlayers

IvanSanchez:
Implement an 'overlap' spatial operator in both `L.Bounds` and `L.LatLngBounds`, with a semantic of "the intersection of the given rectangles is a rectangle".

Note that this is technically not the same as the [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) 'overlaps' operator, as will return `true` if both rectangles are the same.

This operator is used in `L.GridLayer` to prevent tiles just touching the later bounds from being loaded.


mourner:
Looks like a good solution to me. Lets wait for #2981 commenters to try this out and then merge.

mkoehler2911:
#2981 is that bug fixed in version 0.7.7?

PR #2224, opened 2013-11-21
Add new maxBoundsResistance option

kikuchan:
Obey maxBounds during dragging a map.

New option "maxBoundsResistance" (0.0 - 1.0) controls its behavior;
0.0: Classic (No effect during dragging)
0.8: Modern reaction
1.0: Strictly forbid to beyond the maxBounds

PR #3510, opened 2015-06-02
Add an option to make the map's maxBounds non-bouncy. Fixes #1673.

### LINKED ISSUE #1673: Set map boundings without animation
I would like an option whether the map smoothly bounces back into allowed bounds or rigidly just can't exceed them. At the moment there is no easy way for the latter option, is it?

IvanSanchez:


mourner:
Nice! Can you also take a look at #2224 and see if it's a good approach as well?

src/map/handler/Map.Drag.js
```diff
@@ -28,8 +28,9 @@ L.Map.Drag = L.Handler.extend({
 				dragend: this._onDragEnd
 			}, this);
 
+			this._draggable.on('predrag', this._onPreDragLimit, this);
```
IvanSanchez: The map's maxBounds can be set dynamically, and it would be a real hassle to intercept changes to the maxBounds and attach/detach this event handler :-/

IvanSanchez:
Even though it is possible to rewrite the code in `Map.Drag.js` to make use of `L.Map._limitOffset` (as #2224 does), that would involve calling several coordinate projects on every frame of the drag. Right now I prefer to write three more lines of code (and calculate latLng to pixels just once) than needlessly increasing the complexity of the map dragging.

mourner:
A couple projection calculations is probably negligible even as a part of a move event, but the current approach looks good too.

mourner:
Lets handle any further improvements and subsequent PRs if needed.

PR #3511, opened 2015-06-02
Add third argument to pointToLayer call for MultiPoint

robholmes:
The third argument is the index of the point in the GeoJSON geometry. This enhances the ability to style the returned Marker based upon data stored in GeoJSON properties.

Without this change there is simply no way to utilise any properties you have, as you cannot tell what data you have been given, other than the `latlng`.


mourner:
Can you give a sample of a GeoJSON that would benefit from this? Usually MultiPoint is considered one feature, with one set of properties shared between all points, — otherwise it wouldn't be encoded as MultiPoint (but rather a set of distinctive Point features).

lonre:
Could you please reconsider this pr @mourner

Issue #2788, opened 2014-07-11
0.8-dev IE10 map overflows outside of map container


DeadBurger:
On Internet Explorer 10 the map overflows outside the container when zooming and then resets back to normal.

Here is a JSFiddle to reproduce the issues: http://jsfiddle.net/D3vC8/5/

danzel:
From @surkova
A member of our team has a laptop with touch support (Lenovo Yoga Pro 2, Windows 10 Technical preview, touch screen with 3200x1800 resolution) and he uses IE11 11.0.9800.0. He experiences a weird zoom bug that you can reproduce even on Leaflet's homepage (currently using version 0.7.3) on Overview Tab. When you try to zoom in / out in the map, it overflows the container and the overflow may take up to whole screen.

IvanSanchez:
This looks like a IE10 rendering heisenbug to me. I can reproduce by disabling the animated zoom end transition so it stays on the screen... but as soon as you alt-tab to the web developer tools, or just  move the cursor on top of the faulty tiles, IE10 redraws the screen. And triggering while having a tile selected in the DOM explorer causes it to be non-reproducible: the map just glitches between zoom levels without overflowing.

I see no way of solving this glitch, short of disabling animations for IE10 altogether :-(

IvanSanchez:
Interestingly enough, setting `z-index: -1` (or lower) in the map `<div>` container seems to alleviate the bug :-D

mourner:
Wow, WTF.

IvanSanchez:
Unfortunately, setting `z-index: -1` on the map container will break all touch interactions on WinPhones (and make it glitch extra hard).

Which begs the question of how the f*** to solve this for the edge case of IE10 on a Win8 tablet.

I have been trying a few alternative methods to try and make IE10 isolate the map div so it will clip the contents (nested container, alternate ways of creating a z-index stacking context) with no luck at all.

PR #3512, opened 2015-06-03
Work around IE10 CSS transform glitch by tweaking map container z-index.

IvanSanchez:
Should work around #2788 for most users. I do guess that the bug has to do with the z-index of the tile pane relative to the rest of the elements of the page, and ideally the z-index of the map container should be lower than its DOM parent for this to work reliably, but meh.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -470,6 +470,10 @@ L.Map = L.Evented.extend({
 			throw new Error('Map container is already initialized.');
 		}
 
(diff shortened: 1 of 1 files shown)


mourner:
I'm worrying if this might trigger different bugs in other situations... Doesn't feel very good.

IvanSanchez:
Maybe changing the z-level of the panes during a transition (instead of the map container) can also work around #2788, if you feel that'll be cleaner.

Issue #2718, opened 2014-06-02
Destroy map trouble


sergeantklin:
I use several maps at a time. They can be created and deleted dynamically. But when I destroy a map using map.remove() there is a memory leak and "Detached DOM trees" appears in Chrome Dev Tools.

Working example here http://84.53.192.170/patyi/index.html

How to destroy maps properly?

IvanSanchez:
First, `map.remove()` should remove all layers. If not, these layers keep a dangling reference to `this._map` and cannot be added to any other map instance afterwards.

Second, I've found a non-completely-clean way of triggering `map.remove()` when the map container is removed from the DOM (#3514). See https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver , which seems to be the only way to do it.

mourner:
@IvanSanchez lets not use mutation observers since we want Leaflet to work reliably in older browser (it's not supported in IE up to 10)

IvanSanchez:
@mourner I'm surprised you still have hopes for *anything* to work reliably on IE8 :-P

mourner:
I just don't like _adding_ code that's not strongly necessary and not working everywhere. :)

mourner:
Merged.

foodaka:
Hi there. Im also having trouble removing the map. Im using react with a bootstrap modal to render the map. see the removeMap function below.

```
class NeabySpot extends React.Component {

    constructor(props) {
        super(props);
        this.state = { showModal: false };
    }


    componentDidUpdate(prevProps, prevState){
        if(prevProps.selectedSpot !== this.props.selectedSpot) {
            const lat = this.props.selectedSpot.getIn(['gpsLocation', 'lat']);
            const lng = this.props.selectedSpot.getIn(['gpsLocation', 'lng']);

            this.drawMap(lat, lng)
        }
    }


    drawMap(lat, lng) {


        if (!this.map) {
            this.map = L.map("map-nearby");
            L.tileLayer("http://{s}.tile.osm.org/{z}/{x}/{y}.png", {
                attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
            }).addTo(this.map);
            this.markers = new L.FeatureGroup();
            this.map.addLayer(this.markers);
        }
        this.map.setView([lat, lng], 3);
        this.mapIcon = L.icon({
            iconUrl,
            iconRetinaUrl,
            shadowUrl,
            iconSize: [25, 41],
[...]

PR #3514, opened 2015-06-04
Cleaning up map layers when map.remove() is run                             

IvanSanchez:
Should fix memleaks from #2718


mourner:
Let's ditch mutation observer part and leave that up to the user. Also a build to fix.

PR #3518, opened 2015-06-08
Reworked pixelBounds calculation in GridTile to deal with fractional zoom

### LINKED ISSUE #3273: Incorrect calculation of visible tiles on decimal value zooms
When I set zoom with decimal values the map doesn't load properly. When I drag over the map, in the edges of map, the tiles aren't load if the new margin is minor than the tile width. Only when the margin is enough, is when the load occurs. In version 0.7.3 worked decimal values but not in version 0.8. The link http://jsfiddle.net/alecabdia/dzdz59ce/ is an example of what happens.

IvanSanchez:
Fixes #3273

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -413,15 +413,19 @@ L.GridLayer = L.Layer.extend({
 		if (!map) { return; }
 
 		if (center === undefined) { center = map.getCenter(); }
-		if (zoom === undefined) { zoom = Math.round(map.getZoom()); }
+		if (zoom === undefined) { zoom = map.getZoom(); }
+		var tileZoom = Math.round(zoom);
 
-		if (zoom > this.options.maxZoom ||
-			zoom < this.options.minZoom) { return; }
+		if (tileZoom > this.options.maxZoom ||
+			tileZoom < this.options.minZoom) { return; }
 
-		var pixelBounds = map.getPixelBounds(center, zoom),
-			tileRange = this._pxBoundsToTileRange(pixelBounds),
-			tileCenter = tileRange.getCenter(),
-			queue = [];
+		var scale = this._map.getZoomScale(zoom, tileZoom),
+		    pixelCenter = map.project(center, tileZoom).floor(),
+		    halfSize = map.getSize().divideBy(scale * 2),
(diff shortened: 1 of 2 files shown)


mourner:
I wonder, could we change `map.getPixelBounds` method to work here so that the code is clean and nice? Why does it fail here? (also: fix build)

IvanSanchez:
I did try reusing `map.getPixelBounds` to account for the fractional zoom level, but that involved passing a fake scale around in the rest of the calls, and it just didn't look clean enough to me :-/

Will fix the build ASAP, though.

src/layer/tile/GridLayer.js
```diff
@@ -413,15 +413,19 @@ L.GridLayer = L.Layer.extend({
 		if (!map) { return; }
 
 		if (center === undefined) { center = map.getCenter(); }
-		if (zoom === undefined) { zoom = Math.round(map.getZoom()); }
+		if (zoom === undefined) { zoom = map.getZoom(); }
 
 		if (zoom > this.options.maxZoom ||
 			zoom < this.options.minZoom) { return; }
```
mourner: is there no side effect to changing the min/maxZoom logic so that you compare unrounded zoom?

IvanSanchez: The zoom level is rounded a few lines down; the original zoom level is needed for fetching the tile scale; and I didn't want to run more than one call to `map.getZoom()`.

mourner: yeah, I mean, e.g. with `zoom = 14.2` and `maxZoom = 14`, now it'll return early, whereas before the PR it wouldn't because it compares the rounded value.

IvanSanchez: I'll test that tomorrow.

IvanSanchez: I'm not 100% sure this is the right way to go, because semantics are different for the Map and the GridLayers. In the map you can specify a fractional maxZoom/minZoom, but maxZoom/minZoom on GridLayers will only have a meaning on integer values.

I'll make the change; we can think about semantics (and fractional GridLayer zoom levels) later.

mourner:
Merged! Lets handle any code improvements in subsequent commits (if any).

Issue #1794, opened 2013-06-26
Mouse wheel zooms too fast in FF


klimashkin:
One mousewheel zooms page 3 times.
Leaflet 0.6, FF 22

ganeshx:
Has this issue been fixed? The zoom in/out is too sensitive

mourner:
@ganeshx can you try the master version?

CramericaIndustries:
Sorry but that bug is back. Tested with Fire Fox 54.0.1 and Leaflet 1.1.0. Chrome works fine (v59.0.3071.115).
Example map: [http://leafletjs.com/](http://leafletjs.com/)

perliedman:
@CramericaIndustries please open a new issue if you believe this is a bug in the current Leaflet release, it's highly unlikely it's related to this four year old issue.

CramericaIndustries:
The `mousewheel.with_alt.delta_multiplier_y` parameter was set to `500` (default 100) in `about:config`

PR #3523, opened 2015-06-09
Fractional zoom controls

IvanSanchez:
Allows developers to specify the zoom delta for user interactions (scrollwheel, keyboard, zoom control buttons), and whether the zoom level will snap back to place to integers (or fractions of an integer) on user interaction (box zoom, flyTo stop).

Behaviour based on the discussion in #3273. Supersedes #3326, #3030 and #3285 because they were written before #3278 (big refactor) was merged.

[Review on Reviewable](https://reviewable.io/reviews/leaflet/leaflet/3523)

A few things pointed out by @hyperknot that need a bit of review/rethinking/rework:
* [x] «Are you sure you want to send `this._map.options.zoomSnap` as `_animateZoom` / `noUpdate`?»
* [x] «ideally, `scrollZoom` should benefit from the smoothness of the fractional zoom levels, but wouldn't change the zooming speed slower or faster.»

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -331,41 +331,25 @@ L.Map = L.Evented.extend({
 
 	getBoundsZoom: function (bounds, inside, padding) { // (LatLngBounds[, Boolean, Point]) -> Number
 		bounds = L.latLngBounds(bounds);
-
-		var zoom,
-		    maxZoom = this.getMaxZoom(),
-		    size = this.getSize(),
-		    snap = L.Browser.any3d ? this.options.zoomSnap : 1,
-
-		    nw = bounds.getNorthWest(),
-		    se = bounds.getSouthEast(),
-
-		    zoomNotFound = true,
-		    boundsSize;
-
 		padding = L.point(padding || [0, 0]);
 
-		if (snap <= 0) {
-			zoom = this.getZoom();
-			boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)).add(padding);
-			var scale = Math.min(size.x / boundsSize.x , size.y / boundsSize.y);
-			return this.getScaleZoom(scale, zoom);
+		var zoom = this.getZoom(),
+		    min = this.getMinZoom(),
+		    max = this.getMaxZoom(),
+		    nw = bounds.getNorthWest(),
+		    se = bounds.getSouthEast(),
+		    size = this.getSize(),
+		    boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)).add(padding),
+		    snap = L.Browser.any3d ? this.options.zoomSnap : 1;
+
+		var scale = Math.min(size.x / boundsSize.x, size.y / boundsSize.y);
+		zoom = this.getScaleZoom(scale, zoom);
+
+		if (snap) {
+			zoom = inside ? Math.ceil(zoom / snap) * snap : Math.floor(zoom / snap) * snap;
 		}
 
-		zoom = this.getMinZoom() - (inside ? snap : 0);
-
-		do {
-			zoom += snap;
-			boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)).add(padding).floor();
-			zoomNotFound = !inside ? size.contains(boundsSize) : boundsSize.x < size.x || boundsSize.y < size.y;
-
-		} while (zoomNotFound && zoom <= maxZoom);
-
-		if (zoomNotFound && inside) {
-			return null;
-		}
-
-		return inside ? zoom : zoom - snap;
+		return Math.max(min, Math.min(max, zoom));
 	},
 
 	getSize: function () {
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -17,7 +17,9 @@ L.Map = L.Evented.extend({
 		trackResize: true,
 		markerZoomAnimation: true,
 		maxBoundsViscosity: 0.0,
-		transform3DLimit: 8388608 // Precision limit of a 32-bit float
+		transform3DLimit: 8388608, // Precision limit of a 32-bit float
+		zoomSnap: 1,
+		zoomDelta: 1
 	},
 
 	initialize: function (id, options) { // (HTMLElement or String, Object)
@@ -72,11 +74,13 @@ L.Map = L.Evented.extend({
 	},
 
 	zoomIn: function (delta, options) {
-		return this.setZoom(this._zoom + (delta || 1), options);
+		delta = delta || (L.Browser.any3d ? this.options.zoomDelta : 1);
+		return this.setZoom(this._zoom + delta, options);
 	},
 
 	zoomOut: function (delta, options) {
-		return this.setZoom(this._zoom - (delta || 1), options);
+		delta = delta || (L.Browser.any3d ? this.options.zoomDelta : 1);
+		return this.setZoom(this._zoom - delta, options);
 	},
 
 	setZoomAround: function (latlng, zoom, options) {
@@ -232,11 +236,8 @@ L.Map = L.Evented.extend({
 	},
 
 	stop: function () {
-		L.Util.cancelAnimFrame(this._flyToFrame);
-		if (this._panAnim) {
-			this._panAnim.stop();
-		}
-		return this;
+		this.setZoom(this._limitZoom(this._zoom));
+		return this._stop();
 	},
 
 	// TODO handler.addTo
@@ -331,9 +332,10 @@ L.Map = L.Evented.extend({
 	getBoundsZoom: function (bounds, inside, padding) { // (LatLngBounds[, Boolean, Point]) -> Number
 		bounds = L.latLngBounds(bounds);
 
-		var zoom = this.getMinZoom() - (inside ? 1 : 0),
+		var zoom,
 		    maxZoom = this.getMaxZoom(),
 		    size = this.getSize(),
+		    snap = L.Browser.any3d ? this.options.zoomSnap : 1,
 
 		    nw = bounds.getNorthWest(),
 		    se = bounds.getSouthEast(),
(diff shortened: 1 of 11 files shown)


src/map/Map.js
```diff
@@ -16,7 +16,9 @@ L.Map = L.Evented.extend({
 		fadeAnimation: true,
 		trackResize: true,
 		markerZoomAnimation: true,
-		maxBoundsViscosity: 0.0
+		maxBoundsViscosity: 0.0,
+		zoomSnap: 1,
```
IvanSanchez: Any operation that changes zoom level (including fitBounds and stopping a flyTo) will snap the zoom level to be a multiple of `map.options.zoomSnap`.

src/map/Map.js
```diff
@@ -16,7 +16,9 @@ L.Map = L.Evented.extend({
 		fadeAnimation: true,
 		trackResize: true,
 		markerZoomAnimation: true,
-		maxBoundsViscosity: 0.0
+		maxBoundsViscosity: 0.0,
+		zoomSnap: 1,
+		zoomDelta: 1
```
IvanSanchez: Any user interaction for the zoom level (keyboard, scrollwheel, zoom buttons) will change zoom by this amount.

src/map/Map.js
```diff
@@ -71,11 +73,13 @@ L.Map = L.Evented.extend({
 	},
 
 	zoomIn: function (delta, options) {
-		return this.setZoom(this._zoom + (delta || 1), options);
+		delta = delta || (L.Browser.any3d ? this.options.zoomDelta : 1);
```
IvanSanchez: There are a few checks for `L.Browser.any3d` scattered around. This is a bit dirty, but I don't think it's better to have `map.setZoomDelta()` and `map.setZoomSnap()` functions to account for old browsers with no CSS transforms.

src/map/Map.js
```diff
@@ -227,11 +231,8 @@ L.Map = L.Evented.extend({
 	},
 
 	stop: function () {
-		L.Util.cancelAnimFrame(this._flyToFrame);
-		if (this._panAnim) {
-			this._panAnim.stop();
-		}
-		return this;
+		this.setZoom(this._limitZoom(this._zoom));
+		return this._stop();
```
IvanSanchez: `map.stop()` snaps zoom. `map._stop()` does not.

src/map/Map.js
```diff
@@ -338,8 +340,21 @@ L.Map = L.Evented.extend({
 
 		padding = L.point(padding || [0, 0]);
 
+		if (snap <= 0) {
```
IvanSanchez: I know adding code is not nice, but when the zoom level is not snapped, bounding boxes should fit exactly to the map bounds (or at least I'd expect them to).

src/map/handler/Map.Keyboard.js
```diff
@@ -4,8 +4,7 @@
 
 L.Map.mergeOptions({
 	keyboard: true,
-	keyboardPanOffset: 80,
-	keyboardZoomOffset: 1
+	keyboardPanDelta: 80
```
IvanSanchez: I know this is an API change. But hey, it's Leaflet 1.0 and we can make braking changes if just for the sake of variable naming consisency! :-)

src/map/Map.js
```diff
@@ -338,8 +340,21 @@ L.Map = L.Evented.extend({
...
 
+		if (snap <= 0) {
+			zoom = this.getZoom();
+			boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)).add(padding);
+			var scale = Math.max(boundsSize.x / size.x, boundsSize.y / size.y);
+			if (Math.log2) {
+				return zoom + Math.log2(1 / scale);	// ES6 only
+			} else {
+				return zoom + (Math.log(1 / scale) / Math.log(2));
+			}
```
mourner: No need for branching here. Just use `Math.log` & `Math.LN2`.

src/map/Map.js
```diff
@@ -349,7 +364,7 @@ L.Map = L.Evented.extend({
 			return null;
 		}
 
-		return inside ? zoom : zoom - 1;
+		return inside ? zoom : zoom - snap;
```
mourner: Could we apply some type of rounding instead of looping through zoom values here? It would be more efficient and take less code.

IvanSanchez: I've given it a bit of thought and maybe we should do a bisection algorithm. I just don't think it's the right time to do that bit of refactoring.
Also, there is the issue of zoom ramps with irregular steps (e.g. 1, 1.1, 1.5, 2, 2.1, 2.5, 3...). I was giving it some thought, and at some point implemented partial support for a user-provided function to snap a zoom to the nearest step. But for a bisection algorithm to work properly, the user would need to provide `round()`, `floor()` and `ceil()`, so I decided to let the bisection stuff rest for a while.

src/map/Map.js
IvanSanchez: Speaking of which: the clean way to do this is have the inverse of `L.CRS.scale()`. I am assuming a scale of 2^zoom for this calculation, and I'm aware this might not be always true.

mourner: `map.getScaleZoom`

IvanSanchez:
Please note that #3273/#3518 is reproducible again, specially on mobile while pinch-zooming. Setting `zoomSnap` to a very low value (e.g. 0.00001) alleviates the problem.

I suspect this is some kind of floating-point rounding error on (or around) `GridLayer._update` or `GridLayer._pxBoundsToTileRange()`.

Jakobud:
Hey guys, FYI take a look at my commit from March #3285 that I think accomplishes this same thing. We should combine efforts on this and make sure neither of us missed anything and combine the commits somehow.

IvanSanchez:
@Jakobud Yeah, #3326, #3030 and #3285 do more or less the same thing, but cannot be applied to the 1.0-beta1 codebase as they would create a huge merge conflict. I did review all of these and decided to go with `zoomSnap` and `zoomDelta`, which I think cover all use cases.

Jakobud:
Cool. What does zoomSnap do exactly?

Jakobud:
Nevermind I just read up on it. Looks good!

IvanSanchez:
@Jakobud I would appreciate if you could check out the `fractional-zoom-controls` branch, play a bit with the `debug/map/zoom-delta.html` page and see if everything makes sense :-)

@mourner What do you think about the API-breaking changes introduced by this? They're worth it IMHO.

Jakobud:
on `zoom-delta.html` the `zoomDelta` value is not working for me when using the zoom controls. Scroll wheel zooming seems to work fine.

Also on the right map somehow I made the zoom value go to `8.893519760186656`. Not sure how I did that but I try to recreate it. 

image

IvanSanchez:
@Jakobud You're right, I need to double check the code for the zoom buttons.

The second map has `zoomSnap` set to zero, which means that any pinch-zoom (in touchscreens) or box-zoom (shift-dragging) will set the zoom level to an arbitrary value. So, "it's not a bug, it's a feature".

Jakobud:
Ah that makes sense now. I stopped the Flyover and the zoom value was arbitrary because no snap. This seems to work very nice except for the buttons not working. I know I had the buttons working good in my PR so take a look at that.

IvanSanchez:
@Jakobud The zoom buttons issue should be resolved by 3497577

skalb:
This works great in our app! Any reason it's not merged yet?

liebrand:
Any update on this? I'm interested to know if this is planned to be merged in or not?

mourner:
It is going to be merged for beta3, as you can see in the milestone for the issue.

mourner:
Time to rebase this probably! @IvanSanchez

hyperknot:
Did a commit with my idea for a smooth scrool zoom:
https://github.com/hyperknot/Leaflet/commit/37e382b8d70bbe8d41d0fcf736bbf7d8b44a791e

Can be tested here (try to not concentrate at the tile-pruning bug):
http://jsbin.com/junake/3/edit?html,output

Just some of my ideas:
- While scroll zooming, there shouldn't be any kind of zoom snapping. Floating point zoom levels should provide a really pleasant experience.
- Scroll zoom is a much better experience without zoomAnimation, as there is no added jumps. Without zoomAnimation the zoom can simply be linked to the user's scroll, just like on MapboxGL.
- _limitZoom, setZoom, setZoomAround would need optional parameters to disable snapping and zoomAnimation conditionally.
- The hard case is the end of the zoom, if zoomAnimation == false and zoomSnap is enabled: the JS code would need to handle some kind of smoothing to the nearest snap level.

hyperknot:
I updated the leaflet build used by http://jsbin.com/junake/3/edit?html,output to disable GridLayer / _resetAll, now the tiles are not flickering with zoomAnimation false.

kentongray:
just tried out the demo @hyperknot super smooth on my machine definitely feels more "natural" to me. Hoping that this may hit the main repo some day!

hyperknot:
@kentongray great! Can you share some details about your machine and the kind of scrolling device you've used? I've been experimenting with the smooth zooming option and I implemented mouse / trackpad detection on that build, nice to know that it worked well.

@mourner, @IvanSanchez have a look at the scrolling detection and the reworked sigmoid function diffs: https://github.com/Leaflet/Leaflet/compare/fractional-zoom-controls...hyperknot:fractional-zoom-controls

IvanSanchez:
@hyperknot What's the benefit of having three different settings for the sigmoid function? IMHO telling apart between wheel and touchpad shouldn't be needed :-/

hyperknot:
@IvanSanchez at the moment, zoomAnimate provides some kind of processing to the input with the transitions (settings 1). 

If we disable zoomAnimate for smooth zooming we need to have a different sigmoid, as it works differently now. In the new case, we have to differentiate between trackpad (settings 2) and mouse wheel (settings 3). 

With this we could also incorporate FF specific input processing if needed for: https://github.com/Leaflet/Leaflet/issues/1794#issuecomment-173874314

What do you think about leaving scroll zooming intact in this PR and doing a smooth zooming one? I'd be happy to start working on that one.

IvanSanchez:
> What do you think about leaving scroll zooming intact in this PR and doing a smooth zooming one? I'd be happy to start working on that one.

@hyperknot I'd like to merge this ASAP (and get the fractional zoom snapping all across the board), and worry about smoother sigmoids later. I'm making a small change to keep the scrolling speed roughly the same as it is now, though.

hyperknot:
@IvanSanchez I agree! The smooth zoom will require a lot of testing to make sure it's working well, and it might actually require implementing some non-trivial easing at zoom end, and the change on gridLayer to avoid flicker, so that's a bigger task.

kentongray:
@hyperknot missed your comment but i actually tested on both mouse and trackpad first gen Macbook Pro Retina

hyperknot:
@IvanSanchez checked the PR and for me it doesn't break anything existing. Behaviour with default settings is unchanged on MacBook touchpad (and should be on anything else).

It is slower when using fractional zoom, but that is a different issue and can be fixed in a future PR.

PR #3524, opened 2015-06-09
Fix GridLayer move event handler, fixes #3522

### LINKED ISSUE #3522: _pruneTiles throws exception when swapping TileLayers
I get an exception on the following line if I switch the TileLayer/GridLayer before it's finished loading all tiles: https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L238

Currently, I'm removing the current TileLayer/GridLayer before I add the new one, which means the layer loses reference to underlying map object.

Shouldn't there be a null reference check for map before getting the current zoom?

RLRR:
`getEvents` function in `GridLayer` created a brand new listener for `move` event each time it was called. So, if a tile layer was removed from map, this listener didn't get cleared and continued to be called on every map move, which resulted in an exception.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -104,7 +104,11 @@ L.GridLayer = L.Layer.extend({
 
 		if (!this.options.updateWhenIdle) {
 			// update tiles on move, but not more often than once per given interval
-			events.move = L.Util.throttle(this._onMoveEnd, this.options.updateInterval, this);
+			if (!this._onMove) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
You know when you're about to commit some code and some other person commits **exactly the same, letter for letter** ten seconds before you? :-)

mourner:
Haha! @IvanSanchez you had a fair share of commits recently so lets get this one in :)
@RLRR great catch!

PR #3526, opened 2015-06-09
Fix setting of offset for popups.

### LINKED ISSUE #3525: layer.bindPopup does not use offset option
Some of the popup options are used, like closeButton, but not offset.

Example here: http://jsfiddle.net/skalb/7mjgLokx/1/

snkashis:
Fixes #3525

### DIFF
--- a/src/layer/Layer.Popup.js
+++ b/src/layer/Layer.Popup.js
@@ -127,7 +127,8 @@ L.Layer.include({
 
 	_popupAnchor: function (layer) {
 		var anchor = layer._getPopupAnchor ? layer._getPopupAnchor() : [0, 0];
-		return L.point(anchor).add(L.Popup.prototype.options.offset);
+		var offsetToAdd = layer._popup ? layer._popup.options.offset : L.Popup.prototype.options.offset;
+		return L.point(anchor).add(offsetToAdd);
 	},
 
 	_movePopup: function (e) {
--- a/debug/tests/popup_offset.html
+++ b/debug/tests/popup_offset.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+</head>
+<body>
+
+	<div id="map"></div>
+
+	<script type="text/javascript">
+
+		var osmUrl = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="http://openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+
+		var map = L.map('map')
+				.setView([50.5, 30.51], 15)
+				.addLayer(osm);
+
+		L.marker(map.getCenter())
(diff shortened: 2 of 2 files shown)


snkashis:
Yep, my mistake.

src/layer/Layer.Popup.js
```diff
@@ -127,7 +127,8 @@ L.Layer.include({
 
 	_popupAnchor: function (layer) {
 		var anchor = layer._getPopupAnchor ? layer._getPopupAnchor() : [0, 0];
-		return L.point(anchor).add(L.Popup.prototype.options.offset);
+		var offsetToAdd = layer._popup ? layer._popup.options.offset : L.Popup.prototype.options.offset
```
yohanboniface: Naive question: why not `this._popup`, as it's the popup that will be open then (or pass it as parameter on the initial call to `_popupAnchor`) ?

mourner:
Cool! Please squash into one commit.

snkashis:
@mourner I don't think we can merge yet,I believe this code has introduced a slight horizontal offset that is not intended.

patrickarlt:
This will sound really really weird but I think @snkashis solution is totally correct and the offset is happening because of browser rounding in the tip of the popup. Currently the popup tip is 17x17 pixels compare this to a  16x16 pixel tip.

A radius 1 circle marker is used for reference on my retina MBP.

screen_shot_2015-07-01_at_9_40_31_am_png___1850___rgb_8__


screen_shot_2015-07-01_at_9_38_02_am_png___2460___rgb_8__

snkashis:
interesting @patrickarlt . what do you think @mourner , is this mergeable in current state?

patrickarlt:
Actually I took the default popup as it sits not on the this branch and measured it all.

Left edge of popup content to center of popup tip: 163px
center of popup tip to right edge of popup content: 161px

It looks like something is off by 1px somewhere I'll see if I can figure it out.

patrickarlt:
@mourner I think this is good to merge this like it is now. It addresses the problem of not using offset with my latest changes. The slight horizontal offset is still an issue in https://github.com/Leaflet/Leaflet/commit/dfbd0fa30902c5c44a3f1f92bbaaf0ea410dff17 which is the commit just before you merged my changes in. Check that commit and look at `debug/map/markers.html`

The popup is placed correctly on the market labeled "Static" and incorrectly on the marker labeled "Draggable"

screen_shot_2015-07-01_at_10_34_55_am_png___618___rgb_8__

screen_shot_2015-07-01_at_10_34_59_am_png___700___rgb_8__

mourner:
@snkashis please squash into one commit and I'll merge.

patrickarlt:
I've filed an issue for the horizontal popup offset in https://github.com/Leaflet/Leaflet/issues/3586

snkashis:
Okay @mourner , squash done.

snkashis:
And...I think we need to revert this because of an issue I just noticed on `/debug/tests/reuse_popups.html`.Offset keeps getting added - this is not correct and I guess a test for this in the future would be a good idea if we can create one
screen shot 2015-07-01 at 2 14 31 pm
 - @patrickarlt got an idea?

mourner:
Reverted.

patrickarlt:
Yup this line https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.Popup.js#L63.

`_popupAnchor` needs to be called every time a popup opens to handle things like a Marker and a Polygon in the same feature group where they need to have different offsets.

patrickarlt:
The fix for this is pretty simple actually. You just have to save the originally passed offset somewhere like `this._originalPopupOffset` and use to calculate the popup anchor every time instead of `this._popup.options.offset`.

I'll have a PR for this soon.

si-the-pie:
An offset keeps getting added if the second (optional) argument to `bindPopup(content, options)` is left unspecified. A workaround is to use `bindPopup(content, {})`.

PR #3527, opened 2015-06-10
Add map presence check to onMoveEnd handler, fixes #3522

### LINKED ISSUE #3522: _pruneTiles throws exception when swapping TileLayers

RLRR:
This will prevent `_onMoveEnd` handler from execution if it is called by `L.Util.throttle` after layer removal.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -407,6 +407,8 @@ L.GridLayer = L.Layer.extend({
 	},
 
 	_onMoveEnd: function () {
+		if (!this._map) { return; }
+
 		this._update();
 		this._pruneTiles();
 	},


IvanSanchez:
Is this needed at all after #3524?

RLRR:
@IvanSanchez, yes, since `L.Util.throttle` works by scheduling `_onMoveEnd` to execute after a certain delay. So if we remove a layer **during** that delay, the layer will be already removed when the handler is called. See https://github.com/Leaflet/Leaflet/issues/3522 for more details.

PR #3528, opened 2015-06-10
Add L.DomEvent.once

yohanboniface:
@mourner not sure you'll merge this one ;)
But it's very useful, and also consistent with `L.Evented` API, so my opinion is that it's worth it. :)
Sadly I can't see a good way to factorize with `L.Evented.once`, but I see that none of `on` or `off` has been factorized.
Of course, I'll add a bit of doc if you think it's mergeable :)

### DIFF
--- a/spec/suites/dom/DomEventSpec.js
+++ b/spec/suites/dom/DomEventSpec.js
@@ -99,4 +99,46 @@ describe('DomEvent', function () {
 			expect(simulateClick(el)).to.be(false);
 		});
 	});
+
+	describe('#once', function () {
+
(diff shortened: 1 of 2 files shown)


mourner:
Lets leave that out of core for now.

yohanboniface:
Found today one more case where I needed it ;)  #lobbying ;)

IvanSanchez:
> Found today one more case where I needed it ;) #lobbying ;)

+1 :-)

yohanboniface:
One more use case today :) #lobbying

PR #3533, opened 2015-06-11
Synthetic doubletap events respect stopPropagation, fixes #3532

### LINKED ISSUE #3532: Events fired from DoubleTap ignore stopPropagation
This is better explained with an example:

* Using anything with a touchscreen, go to any page with a map that has a `L.Control.Zoom`. http://leafletjs.com/examples/mobile-example.html works OK for this.
* Zoom all the way out, so that the "zoom out" button is disabled and grayed out.
* Double-tap on the "zoom out" button.
* See how the map is zoomed in because of a `dblclick` event, even though `L.Control.Zoom._createButton` explicitly stops propagation of dblclick events.

This is reproducible both on 0.7.3 and master.

Maybe our resident expert in events @yohanboniface can shed some light on the issue. And BTW, the bug also happens in the `map-domevent` branch, in case you were wondering.

IvanSanchez:


src/dom/DomEvent.DoubleTap.js
```diff
@@ -33,7 +33,7 @@ L.extend(L.DomEvent, {
 		}
 
 		function onTouchEnd() {
-			if (doubleTap) {
+			if (doubleTap && !touch.cancelBubble) {
```
yohanboniface: `touch` can't be `undefined`?

IvanSanchez: Apparently not, because onTouchStart sets touch to an event. There might be a degenerate case where the user starts the touch outside the map, then finishes it inside the map. Anyway the current code (a few lines down) assumes that `touch` exists and is an object, so no regression here.

Maybe the right thing to do is to follow the pattern of the `preclick` event in the `map-domevent` branch - fire a synthetic event which resembles a DOM event, instead of a leaflet event (which includes coordinates).

Right now I'd rather patch the bug and not think of any more refactoring.

yohanboniface: LGTM :)
(I'll let the master click the green button ;) )

PR #3534, opened 2015-06-12
Check bounds before clipping

wpf500:
Hi, I'm working with a map that has a lot of polygons outside of the map view at any one time. I'd like to suggest a change that checks if the polygon is visible within the renderer before clipping, which I have found is more performant with no discernible impact on clipping visible polygons.

Performance test:
- Out-of-bounds polygon: http://jsperf.com/leaflet-bounds-test
- Visible polygon: http://jsperf.com/leaflet-in-bounds-test

### DIFF
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -65,11 +65,6 @@ L.Polygon = L.Polyline.extend({
 	},
 
 	_clipPoints: function () {
-		if (this.options.noClip) {
-			this._parts = this._rings;
-			return;
-		}
-
 		// polygons need a different clipping algorithm so we redefine that
 
 		var bounds = this._renderer._bounds,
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Why is the check done only for polygons and not for lines/circles? (or paths, for that matter)

wpf500:
I had assumed that most shapes used native SVG elements (`circle`, `rect`, etc.) but if that's not the case I guess this could apply to all of them. I wouldn't have thought simpler shapes would show much of a performance gain though?

IvanSanchez:
> I wouldn't have thought simpler shapes would show much of a performance gain though?

You haven't worked with displaying road networks, I see :-)

> I had assumed that most shapes used native SVG elements ...

Does this code play well with Canvas? Also, will this break `flyTo()` animations? Keep in mind that @mourner very recently modified the code (in master) for animating zoom on vector layers, and redraws might be tricky.

Issue #333, opened 2011-10-15
Drawing vectors and saving (i.e creating var ) into GeoJSON


ns-1m:
Can we draw vectors in next release of Leaflet?

I hope we can draw and edit vectors and save it in GeoJSON (i.e. create a GeoJSON file / variable from it.

Then, we can save the GeoJSON in spatialite. Spatialite 3.0.0 can produce Geojson and save Geojson as well.

Spatialite has javascript bindings thru QuickConnect hybrid  for Android and iOS devices.

PR #3536, opened 2015-06-13
No Border on a.leaflet-popup-close-button

katosh:
I had a problem with a strange line crossing the close-button of leaflet popups. It turned out it was a `border-bottom: 1px solid #333;` set for links by the css-stylesheet of the page. This underlining is not erased by `text-decoration: none;`.


mourner:
Thanks!

PR #3539, opened 2015-06-15
Add WFST to plugins page

kuzkok:


IvanSanchez:
That plugin should go into the 'overlay formats' section, and not in the '3rd party' section. 3rd party is for plugins that *only* work with some specific vendor.

Please fix that, amend the commit and we'll merge it. Plugin demo looks OK for me :-)

PR #3547, opened 2015-06-18
Simple canvas optimizations

philippelatulippe:
I'm trying to use Leaflet with a large amount of polylines, and I've found a few low-hanging fruits which make things snappier when using the canvas renderer.

* throttle mouseover using L.Util.throttle
* don't check for hover when the map is being panned (since the mouse isn't moving in relation to the map)
* only check visible layers when checking for hover (by storing unculled layers in an array and only checking these layers)

Thank you for reviewing. I'm new to Leaflet, so I hope this doesn't break anything that I don't know about!

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -17,7 +17,7 @@ L.Canvas = L.Renderer.extend({
 		var container = this._container = document.createElement('canvas');
 
 		L.DomEvent
-			.on(container, 'mousemove', this._onMouseMove, this)
+			.on(container, 'mousemove', L.Util.throttle(this._onMouseMove, 66, this), this)
 			.on(container, 'click dblclick mousedown mouseup contextmenu', this._onClick, this)
 			.on(container, 'mouseout', this._handleMouseOut, this);
 
@@ -26,6 +26,8 @@ L.Canvas = L.Renderer.extend({
 
 	_update: function () {
 		if (this._map._animatingZoom && this._bounds) { return; }
+
+		this._drawnLayers = [];
 
 		L.Renderer.prototype._update.call(this);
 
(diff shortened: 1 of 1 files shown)


philippelatulippe:
I fat-fingered the enter key wayyy before this was ready, sorry about that!

mourner:
Wow, nice optimizations! I'll merge when it's ready :)

wpf500:
Thanks a lot for this, I've cherry-picked these changes onto my fork.

One small problem I found is that if you change the polygon styles but don't move the map (thus triggering `_updatePoly` but not `_update`), `_drawnLayers` just grows perpetually.

A quick change I would suggest is to make `_drawnLayers` an object and use the layer's `_leaflet_id` as an index, avoiding duplicate entries for the same layer. I've made the changes on my fork here: https://github.com/wpf500/Leaflet/commit/00c5e46a2daff196d1fe1e73f092fa12c085b2f9

mourner:
@philippelatulippe @wpf500 hey, would love to see this ready for merging! Any updates?

mourner:
@philippelatulippe @wpf500 bump :)

philippelatulippe:
The best thing would be that layers don't get added to _drawnLayers when their style changes. Let's see if I can do that elegantly.

src/layer/vector/Canvas.js
```diff
@@ -219,8 +223,8 @@ L.Canvas = L.Renderer.extend({
 	_handleMouseHover: function (e, point) {
 		var id, layer;
 		if (!this._hoveredLayer) {
-			for (id in this._layers) {
-				layer = this._layers[id];
+			for (id in this._drawnLayers) {
+				layer = this._drawnLayers[id];
```
mourner: If `drawLayers` is an array, it should be iterated with a usual `for` loop since it's much faster than `for in`. Or you might want to consider @wpf500's suggestion to turn it into an id-keyed object.

philippelatulippe:
Alright, I cherry-picked @wpf500's commit.

mourner:
:+1: thanks a lot!

Issue #3087, opened 2014-12-09
Allow finer control over how GridLayer chooses which tiles to load and retain


irskep:
When panning around in a map with a `GridLayer`, tiles are only created once their rect enters the map's visible bounds. This means that as you pan around, you see tiles pop in constantly.

I'd like to have the ability to load additional tiles outside the view (for example, within a 200px margin) so that tiles are loaded before they enter the visible rect.

Possible fixes:

* Instead of calling `this._map.getBounds()` to determine the working set of tiles, write a simple wrapper `this._getBounds()` that can be overridden
* Take a `preloadMargin` argument (or something). Might be hard to document.

If @mourner has a preferred implementation, I'd be happy to submit a PR to implement this feature.  `GridLayer` has solved a lot of my problems and it would be cool to have it solve even more. :heart:

irskep:
Ping for guidance

IvanSanchez:
Possibly a duplicate of #512

irskep:
Yeah, this is more a request for a plugin hook than an actual implementation of #512. Hence the wording of the title. I'd like to be able to plug in some kind of function that takes the current map bounds and returns a list of tiles to keep loaded.

@mourner said it could be implemented as a plugin, but from my experience with GridLayer's internals, I don't think it's very practical.

mourner:
Yeah, I think we should make this easier in GridLayer for plugins. We'll be able to do that once we're done with events refactorings.

mourner:
I think we're good for now after #3551

PR #3551, opened 2015-06-20
edgeBuffer property for pre-loading tiles outside the current viewport

TolonUK:
Reduces the likelihood of seeing blank map areas at the edge of the viewport when panning. Default value for the new property is 0, which disables the new feature.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -17,8 +17,11 @@ L.GridLayer = L.Layer.extend({
 		zIndex: null,
 		bounds: null,
 
-		minZoom: 0
+		minZoom: 0,
 		// maxZoom: <Number>
+
+		// edgeBuffer: the number of tiles to load beyond those required for the current viewport
+		edgeBuffer: 0
 	},
 
 	initialize: function (options) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Given that this feature is not really needed in the core leaflet, personally I'd rather have it as a plugin.

TolonUK:
@IvanSanchez Can you recommend a convenient way of doing this as a plugin? I'd rather not extend TileLayer as I would like to use this feature alongside other plugins and having multiple plugins all extending the same the same base could get messy. FYI - The way it works at the moment is by extending the pixel bounds area in the _update method before it is converted into the tile bounds area.

IvanSanchez:
Save a reference to the TileLayer's original function, then use `L.Class.Include()` to overwrite the function but call the previously defined one, like this: 

https://github.com/MazeMap/Leaflet.ControlledBounds/blob/master/src/Leaflet.ControlledBounds.js#L344

Note how that plugin is really one big function call, to wrap everything in a closure to maintain function references in memory.

TolonUK:
Thanks, that's really helpful. Would you have any objections to still adding the _getTiledPixelBounds function to the core (as in https://github.com/TolonUK/Leaflet/commit/1f8d2a80723f9e0303bc9621dab69df03d65cc64 )? The only change here is to move some of the pixel bounds calculation out of the _update function, which makes the code more readable IMHO, but this would also make writing the plugin much easier as I think I could implement it by just dealing with _getTiledPixelBounds and avoiding the _update function.

mourner:
@TolonUK hey, that sounds good. Actually there's a ticket about exactly this — see #3087.

TolonUK:
@mourner I knew I couldn't have been the first one to want this feature. :)

TolonUK:
Ok, I've created a plugin for this that relies on there being a new _getTiledPixelBounds method in GridLayer. I've created a new core Leaflet branch with only this change in; should I close this pull request and open a new one?

FYI - My new plugin is at https://github.com/TolonUK/Leaflet.EdgeBuffer with a link to a demo.

danzel:
Yep send a PR to add this plugin to the list.

I think ```gh-pages-master``` is the branch you'll be after :)
https://github.com/Leaflet/Leaflet/blob/gh-pages-master/plugins.md

PR #3552, opened 2015-06-20
Added option to avoid removing all tiles in GridLayer.redraw()

TolonUK:
Enable external calls to GridLayer.redraw to trigger the tile update mechanism without first removing the tiles. Used for tile layers that can change over a small temporal period.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -87,9 +87,13 @@ L.GridLayer = L.Layer.extend({
 		return this._loading;
 	},
 
-	redraw: function () {
+	redraw: function (removeTiles) {
+		if (typeof removeTiles === 'undefined') { removeTiles = true; }
+
 		if (this._map) {
-			this._removeAllTiles();
+			if (removeTiles) {
+				this._removeAllTiles();
+			}
 			this._update();
 		}
 		return this;


colllin:
It looks like this commit changes the default `redraw` behavior.  Previously, it removed all tiles by default.  Now it only removes all tiles if you pass in `true`.  Was that the intention?

IvanSanchez:
See #3642 - there might be similar use cases where redrawing all the tiles too fast might be problematic.

TolonUK:
@colllin The intention was to leave the default behaviour the same. If removeTiles is undefined then it is set to true to keep the default behaviour. The new behaviour should only take effect when you change a call from removeTiles() to removeTiles(false).

IvanSanchez:
@TolonUK OK, but what problem does this solve?

TolonUK:
@IvanSanchez It's useful for tile layers that change over a short period of time and need to be frequently refreshed.

Without this change all the tiles in the layer are removed when the new tiles are requested; to the user this looks like the entire layer disappears for a fraction of a second until the new tiles are received.

With the change the old tiles stay in place until the new ones are actually received.

IvanSanchez:
@TolonUK If that's the case, then your rapidly-changing custom layer can call `GridLayer._update()` instead. IMHO it's way cleaner than changing the default methods.

TolonUK:
@IvanSanchez Ah, now I see. It seems a little bit obvious now looking at the implementation of redraw()... :) I'll try calling _update() instead of redraw() and see if I get the desired behaviour. Thanks.

IvanSanchez:
It seems there's no need for this after all :-)

TolonUK:
Yep, calling _update() does the trick for me. Thanks again!

Issue #3191, opened 2015-02-03
L.geoJson accepts an array, which is invalid


tmcw:
This line: https://github.com/Leaflet/Leaflet/blob/master/src/layer/GeoJSON.js#L18

Lets you pass an array of GeoJSON objects to L.geoJson. This shouldn't be the case: arrays are not [valid GeoJSON root objects](http://geojson.org/geojson-spec.html#geojson-objects)

iH8:
When learning people how `L.GeoJSON` works i've also ran into this discrepancy more than once. I always recommend them to use valid GeoJSON objects but that an array of feature objects will work too. But when i refer them to the API, that isn't quite clear and so it seems like it's something hacky. Updating the API docs would be a good thing.

Updating the tutorial on `L.GeoJSON` also would be even better. In there you'de have more space to explain that it does also take an array for ease of use. What's confusing at the moment in the tutorial is that it talks about features and featurecollections, but in the code examples there is no example of a proper collection. It's just using arrays without explanation.

A novice/beginner (or someone who doesn't read the spec that's linked) could easily come to the conclusion that an array of features is in fact a valid collection. This could easily be prevented. I bet that for most people reading that tutorial, it's their first encounter with GeoJSON, and making some minor changes would really help in their understanding of the specification/concept.

PR #3556, opened 2015-06-23
Document passing an array of GeoJSON objects

justinwp:
There is no explicit documentation of passing an array of GeoJSON objects.

See #3191 and https://github.com/Leaflet/Leaflet/blob/master/src/layer/GeoJSON.js#L18


mourner:
Looks good. Can you squash the commits into one?

mourner:
Merged manually, thanks.

PR #3557, opened 2015-06-24
Move some logic from GridLayer._update to _getTilePixelBounds

TolonUK:
This move makes the implementation of the Leaflet.EdgeBuffer plugin easy, as previously discussed in https://github.com/Leaflet/Leaflet/pull/3551


mourner:
Cool!

Issue #2603, opened 2014-03-31
Custom Leaflet image finder regex


tmcw:
[this method for finding images](https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Icon.Default.js#L32-33) fails when leaflet.js is not called leaflet.js. In mapbox.js land, we've settled on [hardcoding this](https://github.com/mapbox/mapbox.js/blob/repackage/mapbox.js#L3-5), but that approach breaks as soon as we want to distribute non-hosted, like as a zip download, bower, or etc.

IvanSanchez:
We might want to consider using the error stack as the source for the filename of the leaflet library:

https://gist.github.com/plepe/3891980
https://gist.github.com/BR0kEN-/8620537

mourner:
@IvanSanchez does it work on old IE?

sheppard:
See #3559 for an alternate approach using CSS properties.

ChristopheBougere:
Hi,
I have the same issue than @branonbarrett when minifying leaflet.js : "Couldn't autodetect L.Icon.Default.imagePath, set it manually.".
The only way is to define L.Icon.Default.imagePath manually, or do you think that it's going to be fixed on next release ?
Thanks

mourner:
@ChristopheBougere there's no way to fix it for everyone. Leaflet code can't figure out where the images are if they're not present where it expects by default. I don't think it's a big hassle to add a 1 simple line of code.

ChristopheBougere:
Sure @mourner, thank you !

tmcw:
Closing; not seeing any actions out of this issue.

PR #3559, opened 2015-06-24
use CSS property to compute icon path

sheppard:
Here's an alternative to #2603.  Instead of determining the default `imagePath` based on `<script>` urls, this approach reads from the CSS to figure out where the layer icon is and assumes other images are in the same folder.  The one downside is that it requires creating a DOM element.  Another approach might be to scan through the stylesheet directly but that would require more code (I think).

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -21,6 +21,10 @@ L.Icon.Default = L.Icon.extend({
 		var path = L.Icon.Default.imagePath;
 
 		if (!path) {
+			path = L.Icon.Default.imagePath = computeImagePath();
+		}
+
+		if (!path) {
 			throw new Error('Couldn\'t autodetect L.Icon.Default.imagePath, set it manually.');
 		}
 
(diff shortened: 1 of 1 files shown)


sheppard:
Hmm, looks like this can break if `document.body` is not available when it executes.  I tried doing it without attaching to `document.body`, but then the trick didn't work (some browsers apparently only apply the CSS rule if the node is attached).

sheppard:
Ok, I resolved this by wrapping the `document.body` code in a try/catch block and double-checking the `imagePath` the first time a default icon is created.

mourner:
This approach actually makes a lot of sense! @tmcw @IvanSanchez what do you think?
How about removing the detection call from the top level and only keeping it in icon creation, making try/catch unnecessary?

sheppard:
Sure, that makes sense.

IvanSanchez:
Sounds like a sane approach. I kinda like it.

There's always the risk of a user redefining the CSS for the layer icon, but then again those kinds of users know how to set icon images and look for missing paths.

Maybe it can be even better to have a separate CSS class just for detecting the image path?

sheppard:
Something like this?

```css
.leaflet-default-icon {
	background-image: url(images/marker-icon.png);
}
.leaflet-retina .leaflet-default-icon {
	background-image: url(images/marker-icon-2x.png);
}
```

I suppose the JavaScript would then need to append the element to the map container to get the `.leaflet-retina` class:

```javascript
el = L.DomUtil.create('div', 'leaflet-default-icon', this._map._container); //?
path = L.DomUtil.getStyle(el, 'background-image');
```

mourner:
@sheppard no, we wouldn't switch the markers to use background images because they're not printed and have other accessibility problems.

sheppard:
The markers could still use ``, we would just use the CSS rule to detect which image to use.  I'm fine with re-using the existing rule for `layers.png` and just replacing the text (as it is now).  However, I can see @IvanSanchez's case for a dedicated rule, in which case we would want it to resolve to the image that we are intending to load (which would require leveraging the existing `.leaflet-retina` class).

sheppard:
Or I suppose we could do this:

```css
.leaflet-default-icon {
	background-image: url(images/marker-icon.png);
}
.leaflet-default-icon-2x {
	background-image: url(images/marker-icon-2x.png);
}
```

```javascript
el = L.DomUtil.create('div',  'leaflet-default-icon' + L.Browser.retina ? '-2x' : '', document.body);
path = L.DomUtil.getStyle(el, 'background-image');
```

sheppard:
Fixed with #4605.

PR #3561, opened 2015-06-25
Adds Leaflet.hotline plugin

klaftertief:
The plugin only works with the latest Leaflet from the master branch.
Should this go into the gh-pages or into the gh-pages-master branch?


mourner:
It's fine to have it live on the website, just make sure that it has a prominent compatibility notice in the readme.

mourner:
Wow, really awesome plugin by the way!

klaftertief:
Thanks.
I had a kind of working version for 0.7.3, but switching to the master branch made it so much easier. Nice work.

PR #3563, opened 2015-06-26
Extend L.Class.include to allow references to previously defined methods

IvanSanchez:
When writing plugins, I've run into the problem of modifying the behaviour of a class by having a method run the previously defined method.

So far I've been using the pattern from [leaflet.activearea.js](https://github.com/Mappy/Leaflet-active-area/blob/master/src/leaflet.activearea.js), but that involves creating global variables instead of anonymous closures.

The usual way to work with `L.Class.extend` is to call `previousclassname.prototype.methodname.call(params)`. I would like to use a similar pattern when using `L.Class.include` (as shown in `debug/tests/include-previous.html` in this PR).

### DIFF
--- a/debug/tests/include-previous.html
+++ b/debug/tests/include-previous.html
@@ -1,0 +1,37 @@
+<html>
+	<head>
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+	<link rel="stylesheet" href="../css/screen.css" />
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+	</head>
+	<body>
+
+		<div id="map"></div>
+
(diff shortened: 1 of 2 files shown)


mourner:
Hmm, why not just save variables in a local closure? I don't think this is much more complicated than the example in the PR:

```js
(function () {
    var oldGetTileUrl = L.TileLayer.prototype.getTileUrl;

    L.TileLayer.include({
        getTileUrl: function (coords) {
            var url = oldGetTileUrl.call(this, coords);
            ...
        }
    });
})();
```

IvanSanchez:
Hm, I didn't think about that. The good news is that I can cancel this PR, the bad news is that I need to update several of my plugins :-)

Issue #3167, opened 2015-01-24
Layers control does not support scrolling


lookuptable:
I've noticed that the layers control does not support scrolling such that, when having a large number of layers, the items at the bottom won't be displayed and there is thus no way to select/unselected these layers.

I semi-repro of this problem can be seen here (I've used a very narrow map): http://jsfiddle.net/lookuptable/ke3gevug/1/

jieter:
I fixed this for L.Control.Minimap with this [`_onResize()` method](https://github.com/leaflet-extras/leaflet-providers/blob/master/preview/L.Control.Layers.Minimap.js#L143-L150) ([called on map resize](https://github.com/leaflet-extras/leaflet-providers/blob/master/preview/L.Control.Layers.Minimap.js#L104-L105)) and some extra [CSS](https://github.com/jieter/Leaflet.layerscontrol-minimap/blob/gh-pages/control.layers.minimap.css)

mourner:
I think this was fixed on master, can you verify?

Starefossen:
[Here](http://jsfiddle.net/starefossen/ke3gevug/2/) is an updated fiddle with latest master.

lookuptable:
Looks like it hasn't be fixed on master based on @Starefossen 's fiddle.

mourner:
Ah, I now remember there was a related issue #2882

rowanwins:
@mourner can be closed now

PR #3571, opened 2015-06-30
Scrollbar for layers control with too many layers

rowanwins:
Hi again,

This fix addresses #3167 
On the _expand function I check the map height and the legend height, and if the map is shorter than the legend I add some styling to the legend to give it an overflowY scrollbar.

Hope it helps!
Cheers
Rowan

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -248,15 +248,11 @@ L.Control.Layers = L.Control.extend({
 
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
-		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
-		var legendHeight = getLegendDiv.clientHeight;
-		var getMap = document.getElementById('map');
-		var controlHeight = getMap.clientHeight - 40;
-		if (controlHeight < legendHeight)
+		var acceptableHeight = this._map._size.y - (this._container.offsetTop * 4);
+		if (acceptableHeight < this._form.clientHeight)
 		{
-			getLegendDiv.style.height = controlHeight + 'px';
-			getLegendDiv.style.overflowY = 'scroll';
-			getLegendDiv.style.paddingRight = '10px';
+			L.DomUtil.addClass(this._form, 'leaflet-control-layers-scrollbar');
+			this._form.style.height = acceptableHeight + 'px';
 		}
 	},
 
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -248,6 +248,16 @@ L.Control.Layers = L.Control.extend({
 
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
+		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
+		var legendHeight = getLegendDiv.clientHeight;
+		var getMap = document.getElementById('map');
+		var controlHeight = getMap.clientHeight - 40;
+		if (controlHeight < legendHeight)
+		{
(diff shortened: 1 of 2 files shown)


src/control/Control.Layers.js
```diff
@@ -248,6 +248,16 @@ L.Control.Layers = L.Control.extend({
 
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
+		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
```
mourner: no need to use a query when it can be accessed with `this._form`.

src/control/Control.Layers.js
```diff
@@ -248,6 +248,16 @@ L.Control.Layers = L.Control.extend({
 
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
+		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
+		var legendHeight = getLegendDiv.clientHeight;
+		var getMap = document.getElementById('map');
```
mourner: Map height can be accessed with `this._map.getSize().y`.
Also `getSomething` names are misleading — usually you would call a function like this, not a map.

src/control/Control.Layers.js
```diff
@@ -248,6 +248,16 @@ L.Control.Layers = L.Control.extend({
 
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
+		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
+		var legendHeight = getLegendDiv.clientHeight;
+		var getMap = document.getElementById('map');
+		var controlHeight = getMap.clientHeight - 40;
```
mourner: Magic number. Can we get rid of it somehow so that it works with any styling?

src/control/Control.Layers.js
```diff
@@ -248,6 +248,16 @@ L.Control.Layers = L.Control.extend({
...
+		var getLegendDiv = document.getElementsByClassName('leaflet-control-layers-list')[0];
+		var legendHeight = getLegendDiv.clientHeight;
+		var getMap = document.getElementById('map');
+		var controlHeight = getMap.clientHeight - 40;
+		if (controlHeight < legendHeight)
+		{
+			getLegendDiv.style.height = controlHeight + 'px';
+			getLegendDiv.style.overflowY = 'scroll';
+			getLegendDiv.style.paddingRight = '10px';
+		}
```
mourner: what happens when you remove layers from the control to the point when the control shouldn't be scrolled?

rowanwins:
Thanks for the feedback @mourner , I'll make those tweaks and resubmit.

As it stands the scrollbar only gets added to the layer control when the layer control is greater than the height of the map.

A slightly alternate approach for the styling would be to run the height calculations and if required add an additional class to the form which we could then manage the styling for in leaflet.css rather than adding it inline on the element. Does this sound like a more appealing approach?

mourner:
Yes, that sounds good. Also see popup scrolling code for a similar approach. 
No need to resubmit, just do `git commit --amend` and then `git push --force` to update the current PR.

rowanwins:
Hi @mourner ,
Have amended based on your advice and it is much neater now I think. There is an extra class in the css although we do have to define a px height for the legend to enable to have a scrollbar so I'm having to add that as inline css, having said that I think this is fairly inline with the approach taken by the popup.
The magic number is slightly less magic now, its now calculated based on the container offset.
Fingers crossed this time. 

PS No idea about this git commit --amend business, you're lucky I haven't blown things up yet!

mourner:
Looks good!

paullc:
How can I implement this on a leaflet map made with R?

PR #3573, opened 2015-06-30
Implement non-square tiles for L.GridLayer

### LINKED ISSUE #3570: non-square tiles?
To faithfully represent non-square images one needs to use non-square tiles.

Unfortunately leaflet does not support this.

Last year I did a nasty patch so that it does support this (options.tileSize became a Point) and went on to create a zoomable image gallery.

Would you like me to clean up that patch and do a pull request, or is this feature out of the scope of the project?

IvanSanchez:
Fixes #3570. See discussion there.


mourner:
:+1:

PR #3576, opened 2015-06-30
Experiments using a big background photo of leaves for the pages.

IvanSanchez:
TODO: 
* [x] Retouch the image, make the branch more prominent
* [ ] Cut into tiles
* [x] Set maxBounds dinamically with getBoundsZoom so some image will always be seen.


mourner:
WTF, I thought that was a joke! 
btw the page doesn't work for me :) no image visible and scrolling is broken.

IvanSanchez:
@mourner It works better if I remember to upload the background image :-D

leaflet-redesign

mourner:
It's fun, but I think it's too much :)

IvanSanchez:
Let's see if either @rowanhogan or @jieter have any comments on this.

jieter:
Hmm, I think I agree with @mourner. Fun for once and then only distracting every time you need the API docs.

rowanhogan:
I like the idea, but I agree with @mourner that the site should be kept clean and minimal.

IvanSanchez:
> I think it's too much :) 

> I agree with @mourner

> I agree with @mourner

okay2_kindledx

PR #3577, opened 2015-06-30
Batch canvas context switches

wpf500:
I'm using the canvas renderer with a large number of polygons and have found the following optimization helps. It is basically just removing one of your TODOs (hopefully!).
```
// TODO optimization: 1 fill/stroke for all features with equal style instead of 1 for each feature
```

It groups together features with the same styles by creating a hash of the options that the canvas renderer uses, then draws the features in batches in an animation frame. The hashing algorithm looks a bit ugly, but is the quickest I've tried. One potential optimization would be to cache the hash for each feature, but I'm not sure how that would interact with changing styles.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -4,8 +4,14 @@
 
 var contextProps = ['fill', 'fillOpacity', 'fillColor', 'fillRule', 'color', 'stroke', 'weight',
 	'lineCap', 'lineJoin'];
+var contextPropsLen = contextProps.length;
+
 function getContextHash(layer, closed) {
-	return contextProps.map(function (prop) { return layer.options[prop]; }).join('#') + '#' + closed;
+	var hash = closed;
+	for (var i = 0; i < contextPropsLen; i++) {
+		hash += '|' + layer.options[contextProps[i]];
+	}
+	return hash;
 }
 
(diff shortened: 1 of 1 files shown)


mourner:
Wow, so awesome that you attempted that todo! It was pretty intimidating and I thought it would be too much work and ridden with bugs... Will happily review that! 

How much does the performance improve in your tests? Also, did you try that on examples like the choropleth map?

wpf500:
The project I'm working on is a choropleth map, is there a particular example you want me to run it against?

In terms of performance, I ran some profiling yesterday but forgot to save the results! I'll run them again today or tomorrow and post the results here

mourner:
Any updates on performance?

mourner:
@wpf500 bump :)

wpf500:
Yeah sorry for the delay, been away and project deadline is looming. Will try to get it done ASAP.

mourner:
Merged #3547, so I guess this needs a rebase now.

mourner:
> I'm not sure there's much to be done to improve rendering, batching together canvas contexts (#3577) doesn't actually seem to have the impact I thought it would.

Does this mean this PR should be closed (added complexity not worth the benefit)?

wpf500:
Yeah I think close for now, I might take another stab at it at some point when I have time to have a proper look!

PR #3580, opened 2015-07-01
Check bounds before clipping

wpf500:
This is just a resubmission of #3534, which I stupidly submitted as a merge from master rather than a branch!

=========
Hi, I'm working with a map that has a lot of polygons outside of the map view at any one time. I'd like to suggest a change that checks if the polygon is visible within the renderer before clipping, which I have found is more performant with no discernible impact on clipping visible polygons.

Performance test:
- Out-of-bounds polygon: http://jsperf.com/leaflet-bounds-test
- Visible polygon: http://jsperf.com/leaflet-in-bounds-test

=========

Addressing the concerns raised in the original pull request:

This change has no impact on the result of the `_clipPoints` method, its just a shortcut for out-of-view polygons (either way `_parts == []`). I can't really see how it could impact any existing functionality, but I'm probably wrong!

### DIFF
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -65,11 +65,6 @@ L.Polygon = L.Polyline.extend({
 	},
 
 	_clipPoints: function () {
-		if (this.options.noClip) {
-			this._parts = this._rings;
-			return;
-		}
-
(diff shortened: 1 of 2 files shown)


mourner:
Looks good! Can you squash this into one commit?

wpf500:
Yup, done!

PR #3578, opened 2015-06-30
Website Redesign

mourner:
Master ticket for the redesign. Looks like this: http://mourner.github.io/Leaflet
Remaining stuff:

- [x] add actual links to logos
- [x] revise visual hierarchy (e.g. too much color variation between headings)
- [x] update docs table styling for mobile
- [x] other mobile adjustments (e.g. tagline too big)
- [x] more prominent section headings on the docs page

Dynamic ToC:

- [x] dynamic ToC on the plugins page
- [ ] make docs categories clickable as well
- [ ] fix not being able to click docs permalink icons with certain window width
- [ ] add "to the top" link?

Revising logo list:

- [x] remove Hipmunk (uses GMaps now)
- [x] add Facebook
- [x] add Flickr
- [x] add European Commission
- [x] add Data.gov
- [x] add National Park Service
- [x] add Etsy
- [x] add IGN.com?
- [x] anything else?

Copy edits:

- [x] simplify features section wording
- [x] blog post about redesign

cc @rowanhogan

PR #3581, opened 2015-07-01
Do not stop scroll events if already at min/max zoom.

IvanSanchez:
@jieter's comments on #3576 and the scrollbars in nested code snippets in #3578 made me think.

Whenever a user uses the scrollwheel in a scrollable element, it scrolls that element until it reaches either the top of the bottom and, if so, passes the scroll event to the parent scrollable element.

I guess it's not an insane assumption that Leaflet should behave in a similar way - bubble the mousewheel event upwards when the zoom is at min/max already. This should be useful in any cases a non-fullscreen map is shown in a big page in a small screen.

I'm not 100% of the UX, and I'm not sure if this should become an option somehow.

### DIFF
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -4,7 +4,8 @@
 
 L.Map.mergeOptions({
 	scrollWheelZoom: true,
-	wheelDebounceTime: 40
+	wheelDebounceTime: 40,
+	scrollAtZoomLimits: false
 });
 
 L.Map.ScrollWheelZoom = L.Handler.extend({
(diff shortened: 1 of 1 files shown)


mourner:
Not sure about this. Imagine you're browsing a map near the max zoom — every time you zoom in a bit (and you often can't precisely control how much, e.g. with trackpad), it will zoom in AND scroll the page, which can be very annoying. I prefer the current behavior.

IvanSanchez:
On the other hand, the scrollwheel interaction in a map with only one zoom level feels weird right now.

mourner:
That's a less likely use case.

IvanSanchez:
I've added a `scrollAtZoomLimits` map option which should cover both cases.

src/map/handler/Map.ScrollWheelZoom.js
```diff
@@ -20,14 +21,13 @@ L.Map.ScrollWheelZoom = L.Handler.extend({
 	removeHooks: function () {
 		L.DomEvent.off(this._map._container, {
 			mousewheel: this._onWheelScroll,
-			MozMousePixelScroll: L.DomEvent.preventDefault
+			MozMousePixelScroll: this._preventScroll
```
mourner: Does `MozMousePixelScroll` pass delta as a second argument well?

IvanSanchez: No, that's why I wrote `if (!delta) { delta = L.DomEvent.getWheelDelta(e); }` a bit further down.

On second thought, handling the (deprecated) `MozMousePixelScroll` event fit better in `L.DomEvent` instead.

IvanSanchez:
I don't think there is a lot of value (or interest) in getting this merged.

Issue #3568, opened 2015-06-29
Mouse events on canvas rendering have possible race condition 


shawel:
Hovering over one layer to another (say two different adjacent states ) triggers mouseover and mouseout events in different order.  This is not appear to be the case in svg rendering. 
["mouse over"]
["mouse out"]
["mouse over"]
 ["mouse over"]
["mouse out"]
A hack i currently have in my code is to execute code on mouseover to be about 100ms delay using setTimeout

danzel:
Could you make a jsfiddle to reproduce this? Thanks

shawel:
Here is a code pen showing the mouse event triggers.  It shows the tooltip on some states disappearing b/c of mouse event sequence. It happens only on canvas rendering.  

http://codepen.io/anon/pen/Jdpdav   <= canvas version 
http://codepen.io/anon/pen/rVJVqG  <= svg version

Another unrelated issue I found is, overlapping layers like circles drawn on each state, randomly disappear when you change styles of state layers on mouse over event. The problem can be seen on the same code pen. I can create a new issue. This also happens on the canvas drawing only.  

-- Thanks for a great library. Saved me a lot of time on many projects. The performance on the new version of leaflet is impressive. I was looking to upgrade.

mourner:
Yay fixed!

shawel:
Just tested the problem still exists. 
new build
http://codepen.io/anon/pen/eNVBJz

mourner:
cc @wpf500 :(

wpf500:
I think this is because the canvas hit detection is calculating that the layers are "overlapping". Given that under normal circumstances (i.e. real SVG) you would never get `mouseover` for more than one element at time, the fix is probably just to only look for one hit then stop. What do you think @mourner?

mourner:
@wpf500 yes, sounds right.

mourner:
@shawel how about now?

shawel:
@mourner @wpf500  it works now :+1:

PR #3582, opened 2015-07-01
Fix ordering of mouse events in Canvas renderer

wpf500:
Another quick one (sorry for all the PRs!). I think the ordering of the mouse events in the canvas renderer doesn't quite reflect how the browser would do it if using SVGs. As far as I can tell, the events are usually occur in the following order: `mouseout` -> `mouseover` -> `mousemove`.

Context behind the problem: I'm using the `mouseover` and `mouseout` events to show/hide a tooltip. Depending on the "order" of `_layers`, the tooltip sometimes hid on hovering over a new feature because the `mouseout` event was triggering after the `mouseover`.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -200,17 +200,29 @@ L.Canvas = L.Renderer.extend({
 		if (!this._map || this._map._animatingZoom) { return; }
 
 		var point = this._map.mouseEventToLayerPoint(e);
+		var id;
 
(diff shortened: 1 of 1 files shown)


mourner:
Great! This fixes #3568, right?

wpf500:
Oh yeah, looks like it!

mourner:
> (sorry for all the PRs!)

Not at all — they're great, keep'em coming!

PR #3585, opened 2015-07-01
accept geojson with null geometry

pglotov:
When creating a layer with  `L.geoJson(geojson.data, geojson.options)` where `geojson.data.geometry` is `null` (a valid value according to spec) an error message appears. This PR fixes that.


mourner:
Please fix the whitespace and amend the commit.

pglotov:
fixed

mourner:
Thanks.

Issue #3586, opened 2015-07-01
Slight horizontal offset on popups


patrickarlt:
In some cases there is a slight horizontal offset to popups relative to the center of their markers. This is present both [before](https://github.com/Leaflet/Leaflet/commit/dfbd0fa30902c5c44a3f1f92bbaaf0ea410dff17) and after the popup changes in #3109 . In `debug/map/markers.html` The popup is placed correctly on the market labeled "Static" and incorrectly on the marker labeled "Draggable".

![screen_shot_2015-07-01_at_10_34_55_am_png___618___rgb_8__](https://cloud.githubusercontent.com/assets/378557/8460887/379fde0c-1fdd-11e5-89e4-514b30771b86.png)

![screen_shot_2015-07-01_at_10_34_59_am_png___700___rgb_8__](https://cloud.githubusercontent.com/assets/378557/8460900/42c0db24-1fdd-11e5-9a02-dffdb29b3876.png)

PR #3587, opened 2015-07-01
Fix popup offsets

patrickarlt:
@mourner fixes https://github.com/Leaflet/Leaflet/issues/3525. Incorporates work from  https://github.com/Leaflet/Leaflet/pull/3526 by @snkashis with a  fix for continually added offsets.

Leaving separate for now since there are 2 contributors. But can squash into 1.

### DIFF
--- a/debug/map/markers.html
+++ b/debug/map/markers.html
@@ -41,6 +41,7 @@
 
         var poly = new L.Polygon([[0, 10], [0, 15.5], [0, 50], [20, 20.5]]);
         map.addLayer(poly);
+        poly.bindPopup("Polygon");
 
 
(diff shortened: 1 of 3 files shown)


patrickarlt:
@mourner @snkashis this also solves the horizontal offset issue. Since the default offset is `[1, 34]` that was added to the popup offset `[0, 7]` which created a SLIGHT horizontal shift.

I was having trouble debugging this because I was using `L.CircleMarker` which because it inherits from `L.Path` opens the popup wherever you click not at the center. So this also fixes https://github.com/Leaflet/Leaflet/issues/3586.

mourner:
Rebased into 2 commits. Thanks!

PR #3588, opened 2015-07-02
Updates API docs heading styles

rowanhogan:
This updates the design of the heading on the API page to somewhere between my initial design for the headings and changes from @mourner  which are live on `gh-pages`.

I feel that currently the bottom border is too thick and doesn't fit with the reset of the design. To keep the high visual contrast I've used darker tones (and increased font-weight), but the weight is now consistent between the type and the border.

See screenshots for variations:

_Original_
original

_Current / Live_
current

_This PR_
new

This is just my opinion, I won't be offended if everyone else disagrees!

### DIFF
--- a/docs/css/main.css
+++ b/docs/css/main.css
@@ -91,7 +91,8 @@ h2 {
 	font-size: 2em;
 	font-weight: 300;
 	color: black;
-	margin-top: 1.8em;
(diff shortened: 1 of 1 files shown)


mourner:
Yes, that looks better!

PR #3590, opened 2015-07-02
Only mouse over one feature at a time

### LINKED ISSUE #3568: Mouse events on canvas rendering have possible race condition

wpf500:
This should fix #3568. It stores the currently hovered layer in `_hoveredLayer` and only checks that for mouse events until `mouseout`.


mourner:
Nice! Looks good to me.

PR #3591, opened 2015-07-02
Allow for wide use of pane z-index for ordering.

patrickarlt:
@mourner I expect one of the big uses of the new `createPane` API will be something like this.

```js
var map = new L.Map(...);

// these polygons should always appear OVER other overlays
var polygonPane =  map.createPane('polygonPane');
polygonPane.style.zIndex = 5;

// these polylines should always appear OVER the polygons
var polylinePane = map.createPane('polylinePane');
polygonPane.style.zIndex = 6;

var polygon = new L.Polygon(..., {
 pane: 'polygonPane'
}).addTo(map);

var polyline = new L.Polyline(..., {
 pane: 'polylinePane'
}).addTo(map);
```

This creates a problem because our custom panes will now interfere with the marker and shadow panes.

This PR simply multiplies all the panes `z-index` values by 100 giving people more flexibility when using the API this way.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -70,16 +70,16 @@
 	-moz-user-select: none;
 	}
 
-.leaflet-pane         { z-index: 4; }
+.leaflet-pane         { z-index: 400; }
 
-.leaflet-tile-pane    { z-index: 2; }
-.leaflet-overlay-pane { z-index: 4; }
-.leaflet-shadow-pane  { z-index: 5; }
-.leaflet-marker-pane  { z-index: 6; }
-.leaflet-popup-pane   { z-index: 7; }
+.leaflet-tile-pane    { z-index: 200; }
+.leaflet-overlay-pane { z-index: 400; }
+.leaflet-shadow-pane  { z-index: 500; }
+.leaflet-marker-pane  { z-index: 600; }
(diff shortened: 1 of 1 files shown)


mourner:
Would a multiple of only 10 suffice as well? I think there was some caveat to using big z-index values but I don't remember exactly.

patrickarlt:
I thought that too but I couldn't find anything searching for high z-indexes causing issues. There are stacking order issues in IE though see the bottom of https://css-tricks.com/almanac/properties/z/z-index/. But that occurs at any zi-ndex.

patrickarlt:
@mourner maybe this? http://www.quirksmode.org/bugreports/archives/2006/01/Explorer_z_index_bug.html doesn't affect this PR though.

patrickarlt:
More then happy to agree to 10x ordering if you want though.

mourner:
Ah, nevermind, let's do it :)

vdh:
Does it really need that many zeros? Wouldn't increments of 10 be enough?

z-index wars are always the worst thing to debug, I just had to debug an issue because the zoom controls [skyrocketed to 1000](https://github.com/Leaflet/Leaflet/blob/master/dist/leaflet.css#L114)…

PR #3592, opened 2015-07-02
Update plugin description

ismyrnow:
Updated description of Leaflet.FunctionalTileLayer plugin to better represent use case.


ismyrnow:
Thanks. Love the new design, btw.

Issue #3338, opened 2015-03-29
Missing basic "installation" instructions for non-CDN use


teo1978:
At http://leafletjs.com/download.html
it says:
> Using a Hosted Version of Leaflet
> The latest stable Leaflet release is hosted on a CDN — to start using it straight away, place this in the head of your HTML code:

```
<link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.css" />
<script src="http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.js"></script>
```

But it doesn't say a word about where you are supposed to place the downloaded files and how to link them if I want to use the downloaded version hosted on my own server. Specifically, whether the "images" folder should be placed in the same path where the .css and .js are, or where else.

Note that CDN cannot be used on a page served over https as per #3337, which is equivalent to saying that the CDN version is half useless.

IvanSanchez:
Done in #3595.

PR #3593, opened 2015-07-02
instructions for using downloaded leaflet

jseppi:
ref #3338


mourner:
Thanks! I think the section should go after the hosted section.

jseppi:
squashing into a single commit in another pr

PR #3598, opened 2015-07-03
New test for roundtripping complex geojson

patrickarlt:
To confirm my own curiosity about round-tripping complex GeoJSON with the new MultiPolygon removal.


mourner:
I think that should be added as a proper spec test run on `npm test`.

PR #3601, opened 2015-07-06
Make L.TileLayer.getTileSize aware of non-square tiles

### LINKED ISSUE #3600: Non-square tiles and overscaling probably broken
cf511130 introduced a change in the behaviour of `L.GridLayer._getTileSize`, but we forgot to modify `L.TileLayer._getTileSize`. As a result, overscaling non-square tiles is probably broken.

IvanSanchez:
Should fix #3600, but I have ***not*** tested it out.


mourner:
Still looks good to me. Let's test master for that super-rare case.

PR #3605, opened 2015-07-08
Stop click on _openPopup

### LINKED ISSUE #3604: Click bubbling from marker to map
I have:
* marker with popup
* map click handler (for geocoding)

When I click on marker I've got 'click' event on both map and marker.
So popup opened (ok) and geocoding start working (wrong!).

Why event from marker propagates to underlying  map?
How I can prevent this?

yohanboniface:
Fix #3604.
I had a quick break.
Bubbling from marker is supposed to be the default, now, so I just stopped the event on `_openPopup`.
I could also just add it on `bindPopup`, if you prefer this pattern.

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -147,6 +147,13 @@ L.Util = {
 		return (Object.prototype.toString.call(obj) === '[object Array]');
 	},
 
+	indexOf: function (array, el) {
+		for (var i = 0; i < array.length; i++) {
+			if (array[i] === el) { return i; }
+		}
+		return -1;
+	},
+
 	// minimal image URI, set to an image when disposing to flush memory
 	emptyImageUrl: 'data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs='
 };
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -6,6 +6,7 @@ L.Marker = L.Layer.extend({
 
 	options: {
 		pane: 'markerPane',
+		nonBubblingEvents: ['click', 'dblclick', 'mouseover', 'mouseout', 'contextmenu'],
 
 		icon: new L.Icon.Default(),
 		// title: '',
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -696,7 +696,8 @@ L.Map = L.Evented.extend({
 		for (var i = 0; i < targets.length; i++) {
 			if (targets[i].listens(type, true)) {
 				targets[i].fire(type, data, true);
-				if (data.originalEvent._stopped) { return; }
+				if (data.originalEvent._stopped
+					|| (targets[i].options.nonBubblingEvents && L.Util.indexOf(targets[i].options.nonBubblingEvents, type) !== -1)) { return; }
 			}
 		}
 	},
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -207,5 +207,35 @@ describe("Marker", function () {
 			expect(spy.calledTwice).to.be.ok();
 		});
 
+		it("do not propagate click event", function () {
+			var spy = sinon.spy();
+			var spy2 = sinon.spy();
+			var mapSpy = sinon.spy();
+			var marker = new L.Marker(new L.LatLng(55.8, 37.6));
+			map.addLayer(marker);
+			marker.on('click', spy);
+			marker.on('click', spy2);
+			map.on('click', mapSpy);
+			happen.click(marker._icon);
+			expect(spy.called).to.be.ok();
+			expect(spy2.called).to.be.ok();
+			expect(mapSpy.called).not.to.be.ok();
+		});
+
+		it("do not propagate dblclick event", function () {
+			var spy = sinon.spy();
+			var spy2 = sinon.spy();
+			var mapSpy = sinon.spy();
+			var marker = new L.Marker(new L.LatLng(55.8, 37.6));
+			map.addLayer(marker);
+			marker.on('dblclick', spy);
+			marker.on('dblclick', spy2);
+			map.on('dblclick', mapSpy);
+			happen.dblclick(marker._icon);
+			expect(spy.called).to.be.ok();
+			expect(spy2.called).to.be.ok();
+			expect(mapSpy.called).not.to.be.ok();
+		});
+
 	});
 });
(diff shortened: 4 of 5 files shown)


mourner:
> Bubbling from marker is supposed to be the default, now

Hmm, but why? I think I'd prefer to keep not bubbling up from markers the default.

yohanboniface:
> Hmm, but why?

The map is an event target, so it's considered while traversing the DOM, and given that it listens for `click` the event is fired.

> I think I'd prefer to keep not bubbling up from markers the default.

Are we sure it's a problem (out of retrocompat) ? It's closer to DOM behaviour, no?

ktoto:
@yohanboniface problem not in `openPopup`. problem happen when i have 'click' handler in marker and in map.
`openPopup` just is special case with this binding.

yohanboniface:
@ktoto (take this with caution, as it's not fully decided): with the current code when you listen to a click event and you don't want the map to also receive it, you are responsible of stopping it.
In the case of `bindPopup`, given that the event listener has been added by Leaflet internally, it's its responsibility to then stop it. Makes sense?

ktoto:
@yohanboniface but before #3307 all be fine. I just want to understand why logic with map is changed?
Why `map` want `click` if the first interaction element was `marker`? Why just not stop on `marker` at all?

yohanboniface:
>  Why map want click if the first interaction element was marker? Why just not stop on marker at all?

Just like in normal DOM, when an event is fired by an element, all its ancestors get it, unless `stopPropagation` is called at some point.
You may have a use case where you want to catch all `click` events, both on markers (or polygons) and map.

mourner:
I'd rather not follow DOM conventions and do an exception here, since I can't think of a use case where you'd want to trigger both map and marker on marker click, but people will have a very hard time adjusting to this breaking change.

yohanboniface:
> I'd rather not follow DOM conventions and do an exception here

You mean exception for marker or for `click` ? i.e. never bubble `click` or never bubble `click` on marker but still bubble it on polygons?
I can think of many use cases where I'd love to receive `click` events that where initially fired on a polygon. And by the way, being able to receive `click` on a map whatever the source was would be a killer simplification for all "drawing" application on the map.
But you decide :)

mourner:
Exception for marker specifically, as opposed to vector layers and image overlays (which can be perceived as "click through" in some interfaces).

yohanboniface:
OK, I'll do that asap :)

yohanboniface:
@mourner only for `click` or we can think about other events?

mourner:
@yohanboniface let's follow the 0.7.3 behavior here.

src/core/Events.js
```diff
@@ -4,6 +4,10 @@
 
 L.Evented = L.Class.extend({
 
+	options: {
+		unpropagateEvents: ''  // Space separated events that should not be propagated
+	},
+
```
mourner: This should be in `Layer`.

src/layer/marker/Marker.js
```diff
@@ -6,6 +6,7 @@ L.Marker = L.Layer.extend({
 
 	options: {
 		pane: 'markerPane',
+		unpropagateEvents: 'click dblclick mouseover mouseout contextmenu',
```
mourner: Is there a better option name? Maybe `stopEvents` or `nonBubblingEvents`.

src/map/Map.js
```diff
@@ -696,7 +696,8 @@ L.Map = L.Evented.extend({
 		for (var i = 0; i < targets.length; i++) {
 			if (targets[i].listens(type, true)) {
 				targets[i].fire(type, data, true);
-				if (data.originalEvent._stopped) { return; }
+				if (data.originalEvent._stopped
+					|| (targets[i].options.unpropagateEvents && targets[i].options.unpropagateEvents.indexOf(type) !== -1)) { return; }
```
mourner: `click` will be a false positive if you have `dblclick` non-bubbling.

yohanboniface: Huuuu, good catch! Should we switch to using an array instead? Not sure about perfs :/

src/layer/marker/Marker.js
yohanboniface: `nonBubblingEvents` better to me too, I'll update

src/core/Events.js
yohanboniface: I hesitated ;)

src/map/Map.js
```diff
@@ -696,7 +696,8 @@ L.Map = L.Evented.extend({
 		for (var i = 0; i < targets.length; i++) {
 			if (targets[i].listens(type, true)) {
 				targets[i].fire(type, data, true);
-				if (data.originalEvent._stopped) { return; }
+				if (data.originalEvent._stopped
+					|| (targets[i].options.nonBubblingEvents && targets[i].options.nonBubblingEvents.indexOf(type) !== -1)) { return; }
```
mourner: Unfortunately IE only supports array indexOf starting from 9. We'll probably have to iterate over the array in a loop.

yohanboniface: Gasp :/
Should I add a polyfill `L.Util.inArray` for that? To make the code shorter here and more readable.

mourner: Yeah, maybe something like `L.Util.indexOf`.

yohanboniface: `L.Util.indexOf` seems meaning it also will work on strings, is that what we want?

mourner: It's fine to keep only for arrays since string indexOf doesn't need an util method. It's an internal utility after all, we don't have to make it generic.

yohanboniface: OK done :)

src/core/Util.js
```diff
@@ -147,6 +147,13 @@ L.Util = {
 		return (Object.prototype.toString.call(obj) === '[object Array]');
 	},
 
+	indexOf: Array.indexOf || function (array, el) {
```
mourner: There's no `Array.indexOf` :)

yohanboniface:
Meeeeehhhh, how the console is playing with me!?
screenshot from 2015-07-09 13 50 10

`Array.prototype.indexOf`?

mourner:
Mine is like this :)
image
I think we're fine just not using native indexOf. It's not that much of a difference.

yohanboniface:
Yeah, Chrome vs FF :/
OK, I make the loop for everybody :)

mourner:
Actually I think it's fine at this moment :)

mourner:
OK, let's rebase now.

yohanboniface:
> Actually I think it's fine at this moment :)
Humm, not sure which moment :p Seems I've commit one second before to remove the use of Array.indexOf

mourner:
Nevermind :)

yohanboniface:
Rebased and green :)

PR #3612, opened 2015-07-09
Only consider a target if it listens for event type (fix #3607)

### LINKED ISSUE #3607: Mouse events on markers only return the marker's single latlng point
We use DivIcon markers to display text on our maps, but it looks like https://github.com/Leaflet/Leaflet/pull/3032/files changed marker events to always fire with the marker's single LatLng ignoring the mouse event information.

Not sure if this is a bug, or our use case is just not handled.

Hopefully this simplified fiddle explains better: http://jsfiddle.net/skalb/3sc6nk6b/1/

yohanboniface:
Fix #3607

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -624,11 +624,11 @@ L.Map = L.Evented.extend({
 		this._container.scrollLeft = 0;
 	},
 
-	_findEventTargets: function (src, bubble) {
+	_findEventTargets: function (src, type, bubble) {
 		var targets = [], target;
 		while (src) {
 			target = this._targets[L.stamp(src)];
-			if (target) {
+			if (target && target.listens(type, true)) {
 				targets.push(target);
 				if (!bubble) { break; }
 			}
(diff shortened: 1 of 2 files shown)


src/map/Map.js
```diff
@@ -687,18 +687,17 @@ L.Map = L.Evented.extend({
 		};
 
 		if (e.type !== 'keypress') {
-			data.containerPoint = target instanceof L.Marker ?
+			var isMarker = target instanceof L.Marker;
+			data.containerPoint = isMarker ?
 					this.latLngToContainerPoint(target.getLatLng()) : this.mouseEventToContainerPoint(e);
 			data.layerPoint = this.containerPointToLayerPoint(data.containerPoint);
-			data.latlng = this.layerPointToLatLng(data.layerPoint);
+			data.latlng = isMarker ? target.getLatLng() : this.layerPointToLatLng(data.layerPoint);
```
yohanboniface: @mourner this line is not needed for this patch actually, but `latLngToContainerPoint -> containerPointToLayerPoint -> layerPointToLatLng` does not roundtrip, so I was not able to make a unittest strong enough.
Plus I thought the event latlng in case of a marker should be the exact same as the marker one. Not sure though this is the best way to fix this.

mourner:
Looks good to me!

PR #3614, opened 2015-07-12
Hardcoded dependency L.CRS.Earth in L.Control.Scale

cmulders:
Currently,  L.Control.Scale uses hard-coded the L.CRS.Earth.distance function, while this CRS could have been changed by the user during map creation. 
Changing this to map.distance will remove this hard-coded dependency adding flexibility, while maintaining the same functionality.


mourner:
Thanks!

PR #3618, opened 2015-07-13
Check canvas mouseout

wpf500:
The canvas renderer was not checking for its own `mouseout` event, so a feature wouldn't get a `mouseout` event even if there was something above it (e.g. an absolutely positioned `<div>`). I think if the browser was handling mouse events (i.e. for SVGs), the `mouseout` event would trigger, so this PR brings the behaviour inline with that.


mourner:
Looks great!

PR #3621, opened 2015-07-13
Resuscitate L_PREFER_CANVAS

yohanboniface:
I vote for adding it instead in the breaking changes but in case we change our mind, here is a proposal to take it back to life ;)


IvanSanchez:
Works for me.

PR #3623, opened 2015-07-14
Switch from L_PREFER_CANVAS to map.options.preferCanvas

yohanboniface:
There is a bit of noise in the diff because I also switched debug pages to new syntax.


mourner:
:+1:

PR #3634, opened 2015-07-16
Do not fire preclick when dragging (fix #3632)

### LINKED ISSUE #3632: preclick fires on dragend
`debug/map/markers.html`
- click on marker (got popup)
- drag map
- popup closed when drag ended (wrong)

Before new event model popup is still opened after drag.

It's happed because `preclick` fired on drag end. `preclick` fired because we have `click` event from browser.
`_fireDOMEvent` checked current state in https://github.com/Leaflet/Leaflet/blob/cdb9ce0beb70705fb31b79fdc9376f72c42dc596/src/map/Map.js#L683

May be same check must be in `_handleDOMEvent` or move preclick firing to `_fireDOMEvent`?

yohanboniface:
Should fix #3632.
Don't merge yet, I'll see if I can add a unittest to cover this.

### DIFF
--- a/spec/suites/SpecHelper.js
+++ b/spec/suites/SpecHelper.js
@@ -40,3 +40,38 @@ expect.Assertion.prototype.nearLatLng = function (expected, delta) {
 	expect(this.obj.lng).to
 		.be.within(expected.lng - delta, expected.lng + delta);
 };
+
+happen.at = function (what, x, y, props) {
+	this.once(document.elementFromPoint(x, y), L.Util.extend({
+        type: what,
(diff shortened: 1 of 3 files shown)


yohanboniface:
Added a unittest.
This also adds two utils (from Leaflet.Editable) that allows to simulate a user dragging on the screen.
So this needs double checking before merge :)
cc @tmcw in quality of `happen` master, you may have a better pattern to suggest for simulating drag :)

mourner:
Drag util and test look great to me.

Issue #3026, opened 2014-11-15
BoxZoom remains in bad state after initial use


brendan-ward:
```this._moved``` is set to true on first capture of mouse move event (https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.BoxZoom.js#L49)

however, it is not reset until the next ```mousedown``` event, which results in it being in a bad state.  Outside calls to its ```moved()``` function (e.g., [this fix](https://github.com/Leaflet/Leaflet/commit/aa8be5451361db6c5044116fe275e05c16336cce)) result in other click events on the map being blocked.

The fix should be as simple as adding 
```this._moved = false;``` toward the end of ```BoxZoom::_onMouseUp()``` function.

Happens on current version of master branch.

balrog-kun:
Is anything preventing merging of this fix?  The problems is still present in master on firefox.

yohanboniface:
> However it won't work for IE <9.

I think it's the "anything" your are asking for ;)

mourner:
> Trying to track _moved state has been the source of a lot of bugs.

It's true that an alternative solution would be perfect, but since we didn't find a cross-browser approach to it, I'd try to fix the current boxZoom issue in a more trivial way.

mourner:
Should be fixed by #3637

PR #3637, opened 2015-07-17
Fix boxZoom._moved remaining in bad state after use (cf #3026)

yohanboniface:
This very small fix seems to fix it for me, both on Firefox and Chromium.
I can't see a reason for this not to work on other browser, but I may have missed some edge case to test.

### DIFF
--- a/src/map/handler/Map.BoxZoom.js
+++ b/src/map/handler/Map.BoxZoom.js
@@ -26,10 +26,14 @@ L.Map.BoxZoom = L.Handler.extend({
 		return this._moved;
 	},
 
+	_resetState: function () {
+		this._moved = false;
+	},
+
 	_onMouseDown: function (e) {
 		if (!e.shiftKey || ((e.which !== 1) && (e.button !== 1))) { return false; }
 
-		this._moved = false;
+		this._resetState();
 
 		L.DomUtil.disableTextSelection();
 		L.DomUtil.disableImageDrag();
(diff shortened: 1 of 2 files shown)


yohanboniface:
Found an edge case: hitting `escape` while drawing the box to cancel the action doesn't reset `_moved` state.

mourner:
Also, as far as I remember, the whole thing about remaining `_moved` state was about not firing the click event after something like box zoom. Needs checking.

yohanboniface:
>  the whole thing about remaining _moved state was about not firing the click event after something like box zoom

Good point, seems less trivial then!

mourner:
Maybe resetting it on something like setTimeout 0 would suffice, since click/mouseup/dragend should all happen in the same JS tick.

yohanboniface:
> Maybe resetting it on something like setTimeout 0 would suffice

Works here both in FF and Chromium, in the normal boxZoom scenario plus hitting ESC while drawing the box.

For the record, FF do not issue a `click` event at all after a drag, while Chromium do.

Anyone to test it on other browsers/OS? On the master, just open the `debug/map/map.html` page in the browser, and check that `click` isn't logged into the console after a boxZoom nor a boxZoom + ESC (while doing a basic click on the map should log `click`.

mourner:
> For the record, FF do not issue a click event at all after a drag, while Chromium do.

Even after the setTimeout fix?

src/map/handler/Map.BoxZoom.js
```diff
@@ -26,6 +26,17 @@ L.Map.BoxZoom = L.Handler.extend({
...
 
+	abort: function () {
+		this._resetState();
+		this._finish();
+	},
+
+	_resetState: function () {
+		// Postpone to next JS tick so internal click event handling
+		// still see it as "moved".
+		setTimeout(L.bind(this._resetState, this), 0);
```
mourner: this looks weird, isn't it going to an infinite loop? where is `this._moved = false` part?

yohanboniface: Meeeeh, last time refactor + without unittests => BAM ;)

yohanboniface:
> Even after the setTimeout fix?

No, I meant before the fix, sorry for not being clear.

PR #3642, opened 2015-07-18
Fix #3355 by making GridLayer._resetAll less agressive at removing tiles

### LINKED ISSUE #3355: Tiles not loading when panning the map on requestAnimationFrame()
Hi,

Here is a jsfiddle documenting the issue:
https://jsfiddle.net/ft4qr9r2/3/
It could be related to https://github.com/Leaflet/Leaflet/issues/3196
Worth noting: this was working in a previous version of leaflet (not sure which one, I can look into it!)

There might be a smarter way to incrementally pan the map, or to force the redrawing of the tiles... But I could not find it.

Thanks for your great work!

IvanSanchez:
In #3355, GridLayer._resetAll() is called on every frame - this removes the tiles, and creates new tiles exactly on the same place, but tiles take more than one frame to be requested from cache and rendered.

I wonder if calling `GridLayer._pruneTiles()` instead of `GridLayer._removeAllTiles()` will trigger some dirty tiles in some edge case :-/

### DIFF
--- a/debug/map/panto.html
+++ b/debug/map/panto.html
@@ -1,0 +1,46 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Superseded by #3815. The algorithm there fixes some other problems as well.

Will keep this open for now until a decision about that PR is taken.

Issue #3572, opened 2015-06-30
Move map at the edge of map on marker dragging


marslan390:
when you drag a marker to the edge of the map, the map would pan along with you, making it easy to move your marker long distances.  But, as far as I can tell, Leaflet doesn't do this.  In  order to move your marker to a location that isn't currently visible, you have to zoom out, move your marker, and then zoom back in.

Am I missing something?

mourner:
Yes, that's a feature that Leaflet is missing currently.

IvanSanchez:
I gave a shot at this a couple days ago. Triggering a map `panBy()` and then hacking some offset in the `Marker._draggable` in four different ways only proved to be incredibly glitchy.

I'm afraid this doesn't look easy to implement :-(

PR #3643, opened 2015-07-18
Pan map on marker drag at the edges

cmulders:
Hi,
As discussed in #3572 auto panning at the edges by markers is not supported.

I also gave it a shot and found a solution by giving the `L.Draggable` object a context when a marker is initialized (the map). It works pretty good and the marker stays at the cursor while panning. I am not aware of the current implementation details of `L.Draggable` and probably the code that I added should be in a plugin or extended class.

I tried to minimize the impact on the core code and therefore it uses the event handlers to check when to pan etc. The code could be implemented in `L.Marker.Drag` or an extension of that class.

An example is available in the folder `debug/map/markers-autopan.html`.

### DIFF
--- a/debug/map/markers-autopan.html
+++ b/debug/map/markers-autopan.html
@@ -7,7 +7,7 @@
 
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
 
-	<link rel="stylesheet" href="../css/screen.css" />
+	<link rel="stylesheet" href="../css/mobile.css" />
 
 	<script type="text/javascript" src="../../build/deps.js"></script>
 	<script src="../leaflet-include.js"></script>
@@ -16,6 +16,8 @@
 	<div id="map"></div>
 
 <script type="text/javascript">
+//Function to calculate the 'map'-div position relative to the users window
+// TODO: change to a more elegant way to detect when to pan
 function getOffsetSum(elem) {
     var top=0, left=0;
     while(elem) {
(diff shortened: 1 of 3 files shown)


mourner:
Thanks for the PR. I'll follow up after a review.

kaynz:
@mourner, is this ready to be merged?

TimMcCauley:
@cmulders I got your jsfiddle to work with Firefox, however I'm experiencing difficulties with webkit as ``` e.originalEvent``` is undefined:

``` e.originalEvent``` in Chrome:
``` Object {type: "drag", target: o.C…s.e…d.e} ``` 

``` e.originalEvent``` in Firefox:
``` Object { originalEvent: mousemove, type: "drag", target: Object, latlng: Object } ``` 

Have you got any ideas how to fix this ? Thanks a bunch ,Tim

cmulders:
@mourner Any thoughts for possible implementation of making L.Draggable map or context aware? 

In my opinion is the current implementation hacky. L.Draggable should be unaware of any L.map and only focus on pure DOM manipulation. 
However if a callback or response to the 'predrag' and 'drag' events of L.Draggable could be made,  updating the position information, the map panning handling could be done in L.Marker.Drag instead of in L.Draggable, keeping it clean and tidy without any reference to a 'map' of some sort.

In base Leaflet this functionally is not really required (map panning while dragging a marker), but in drawing plugins this feature or callback point could be very handy to have.

TimMcCauley:
@cmulders we are using this in http://openls.geog.uni-heidelberg.de/openrouteservice-3.0/ and it's working stable, thanks for this.
@mourner I was actually wondering when and if this will be merged into the live branch?

m314:
+1

perliedman:
Closing this in favor of the recently merged #5651.

PR #3650, opened 2015-07-20
Fix mouseout bubbling while initial target was not listening (fix #3648)

### LINKED ISSUE #3648: Map mouseout event fired by TileLayer change of DOM element (in master branch)
Please see http://jsfiddle.net/v1so6m56/
If you move the mouse inside the map, every time you change the img DOM element of tilelayer the map "mouseout" event is triggered.
Reproducible in every browser.
Works fine in LF 0.7.3

yohanboniface:
This simple change should fix #3648.
Sadly, I can't add a unittest because happen doesn't seem to allow configuration of `relatedTarget` [when firing an event](https://github.com/tmcw/happen/blob/cf9e4fc8271b99dbd42ac13ae69b93de2b0445a0/happen.js#L95) (not sure why, cc @tmcw), so while in unittests [this check](https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L673) is always `true`.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -630,8 +630,8 @@ L.Map = L.Evented.extend({
 			target = this._targets[L.stamp(src)];
 			if (target && target.listens(type, true)) {
 				targets.push(target);
-				if (!bubble) { break; }
 			}
+			if (!bubble) { break; }
 			if (src === this._container) {
 				break;
 			}


mourner:
Awesome!

janpaul123:
I think this fix has regressed. (Too bad a unit test wasn't possible…) If a `target` doesn't listen in `_findEventTargets` it will skip it but continue to the parent, thus implicitly bubbling, even if the event is in `nonBubblingEvents`. I can write up an example to reproduce, if you want.

yohanboniface:
> I can write up an example to reproduce, if you want.

Would be very nice, thanks! You can use http://playground-leaflet.rhcloud.com/ for that :)

yohanboniface:
For the record, it's now possible to set `relatedTarget in happen, since https://github.com/tmcw/happen/pull/17

IvanSanchez:
Related: https://github.com/Leaflet/prosthetic-hand/issues/9

janpaul123:
Here you go: http://playground-leaflet.rhcloud.com/damu/1/edit?html,output

There is a click event bound to the map which prints to the console. Clicking on either marker should not trigger that event, because the default options of `Marker` include `nonBubblingEvents` to include `click` (among other events). However, if you click on the top marker (which does not have a popup bound), it still fires. This is exactly the bug that this PR originally fixed.

yohanboniface:
Thanks!

I'm not sure this is a bug actually. Some details: 
- this PR was about preventing `mouseout` and `mouseover` to be targeted on parent elements of the original target
- `nonBubblingEvents` means not firing events on parents of an element *if this element actually caught the event*
- in your example, `topMarker` is not listening for `click`, thus the event is sent to the map
- if you don't want the map to received the event, you need to catch it on the marker and stop it: `topMarker.on('click', L.DomEvent.stop);`

Hope this makes sense! :)

janpaul123:
Ah, if this is intentional then that's fine, but it still doesn't seem great to me that just setting an event handler has the side effect of preventing bubbling.

I would expect people to be confused by this, especially because the browser's bubbling does not work like this. For example, you don't have to call `topMarker.on('click', L.DomEvent.stop);`, it also prevents bubbling when you do `topMarker.on('click', function() {});`. I did not expect that.

If you do want to keep this behaviour, I'd suggest adding some comments in the code to clarify that this is how it's supposed to work (right now [this](https://github.com/Leaflet/Leaflet/blob/63870f1c5b4bc9459d27247a6c82c868d9482da7/src/layer/Layer.js#L34) is the only comment as far as I can tell). And perhaps something about this in the documentation, too.

yohanboniface:
> For example, you don't have to call topMarker.on('click', L.DomEvent.stop);, it also prevents bubbling when you do topMarker.on('click', function() {});. I did not expect that.

See https://github.com/Leaflet/Leaflet/pull/3605#issuecomment-119589147 for more context about this design decision. I agree that this needs to be somewhere in the doc.

janpaul123:
Yeah, that makes sense, this would be a significant breaking change. I suppose both solutions have some element of surprise in them. Perhaps it would also help if we wouldn't call this stuff "bubbling", since that implies things work similar to DOM bubbling. I don't have good alternatives though.

Anyway, thanks for your quick responses and providing context!

PR #3651, opened 2015-07-20
Fix undefined error when removing control layer

davidlukerice:
When removing control layers during an `baselayerchange` event, there's the possibility of layers being removed during the `_onInputClick` loop, causing `input` to be undefined. This PR simply loops backwards to handle removals.

The use case is that I have differing layers depending on which base layer is selected. When the base layer is changed, I remove layers that are no longer applicable and add new ones, throwing the error on [Control.Layers.js:225](https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js#L225).

This is also an issue in current stable, but was unsure what branch to branch/merge from/to for the fix.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -220,7 +220,7 @@ L.Control.Layers = L.Control.extend({
 
 		this._handlingClick = true;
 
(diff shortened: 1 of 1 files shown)


mourner:
Cool, thanks!

PR #3653, opened 2015-07-20
Switch to wheel event where available

### LINKED ISSUE #2757: Jumpy scrolling with touchpad (vertical/horizontal double finger scroll).

mourner:
May fix #2757 #2732. We don't care about FF17 and less where `MozMousePixelScroll` was needed anymore.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -185,6 +185,7 @@ L.DomEvent = {
 		return (e.deltaY && e.deltaMode === 0) ? -e.deltaY :        // Pixels
 		       (e.deltaY && e.deltaMode === 1) ? -e.deltaY * 18 :   // Lines
 		       (e.deltaY && e.deltaMode === 2) ? -e.deltaY * 52 :   // Pages
+		       (e.deltaX || e.deltaZ) ? 0 :	// Skip horizontal/depth wheel events
 		       e.wheelDelta ? (e.wheelDeltaY || e.wheelDelta) / 2 : // Legacy IE pixels
 		       (e.detail && Math.abs(e.detail) < 32765) ? -e.detail * 18 : // Legacy Moz lines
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Now looks good on FFX 38 (which scrolls by lines) and Chromium 46 (which scrolls by pixels) on both a trackpad and a mouse.

Need to check:
- [ ] IE8 
- [ ] IE 10/11 w/ mouse
- [ ] IE 10/11 w/ trackpad
- [ ] Safari w/ mouse
- [ ] Safari w/ magic mouse, trackpad

mourner:
That's cool, although delta in IE that was `/ 40` is now `/ 2`, which looks kinda weird. Did we make it 20 times more sensitive?

IvanSanchez:
@mourner Now we normalize into pixels instead of into lines. In order to get a smooth result in Firefox with default settings, 3 lines has to be mapped to (at least) 52 pixels.

The legacy IE deltas are always 120 pixels. It seems logical to convert them into 60 pixels now (instead of 3 lines), although it needs testing.

IvanSanchez:
> Now we normalize into pixels instead of into lines.

We should mention this in the 1.0.0 docs, BTW.

mourner:
This looks pretty good now, should we merge now and track any concerns in new issues? Do we need any more testing before merge?

IvanSanchez:
I've tried with as many devices I could (which is quite a few), so yeah.

mourner:
:+1: awesome!

Issue #3652, opened 2015-07-20
Opacity flickr when zooming WMS Layers


cosme-benito:
Hello, I have noticed a small flickr issue on layers which have an opacity lower than 1. 

If a layer has an opacity of 0.5 and you zoom in, it seems like the tiles are first loaded with opacity=1 and then are updated to opacity=0.5. This causes a weird flickr. This issue does not occur in leaflet 0.7.

Fiddle with leaflet 1.0-dev (zoom in and notice the flickr when the tiles are refreshing): http://jsfiddle.net/qf9huvk0/1/
Fiddle with leaflet 0.7.3 (no issues): http://jsfiddle.net/p6k4ejuq/3/

Is there any way I can solve this issue? Many thanks!

mourner:
Maybe some race condition after refactoring fade animations cc @jfirebaugh

w8r:
That's trickier than that. Race condition apparently, the flickering is caused by overlapping tiles during the zoom transition.

mourner:
What is surprising to me is that tiles overlapped during zoom transition in 0.7.3 as well, so I don't know why it's not reproducible on the stable version.

w8r:
because they don't have individual opacity settings, so what user sees is a blend of the blurred and scaled and the original tile at `opacity: 1` and container has `opacity: 0.5`

here we got 2 tiles with `opacity: 0.5` on top of each other

mourner:
Oh, good catch! Hopefully we can find a fix to account for that.

w8r:
So how many attempts were made before the decision was made to set the opacity on tile elements individually? Is it worth trying to get IE `filter` inheritance working? Cause it was working for 0.7.3, and the solution seems like an overkill to satisfy the old and broken browser.

w8r:
@mourner what's your idea? cause I'm willing to spend some time on that, if you don't mind

mourner:
Opacity is now set on tile element individually because tile fade animation is now programmatic (requestAnimationFrame-based as opposed to CSS transitions) — it was not possible to implement the new multi-zoom tile loading logic reliably without that change.

Global tile layer opacity being accounted for in individual tile calculations was an afterthought — we probably decided it would be easier on the performance. But if it introduces side effects like this, we could reintroduce opacity on the container. Not sure — feel free to poke with this and see what's the best compromise here.

w8r:
Btw there are also problems with `z-index` in IE9 in the above example
[...]

PR #3658, opened 2015-07-22
GridLayer opacity flicker fix

w8r:
See #3652 

The effect is seen in `debug/map/wms.html`

This seems to fix it, I don't know why 1 for opacity was hardcoded there.


w8r:
No, wrong guess, that doesn't solve the issue. Somehow `_levels` are getting filled with the tiles with `opacity:1`

Issue #2809, opened 2014-07-23
Vector layers loose styling in IE8 when removing and adding back to the map


XellossT:
Hey,

Just found a weird bug on leaflet 0.8-dev on the CircleMaker on IE8.

When i create a GeoJSON layer of circleMarker they loose color when i remove/add them from the map.

The problem is appearing also on cluster (when he display the circleMarker the second time).

Second problem : The circleMarker on IE8 is not so circle. 

Here some screen 
![circlemarker](https://cloud.githubusercontent.com/assets/6204071/3672875/f40b6100-1265-11e4-9d42-b8056beb4930.PNG)
![circlemarkerblanc](https://cloud.githubusercontent.com/assets/6204071/3672877/f64ae83c-1265-11e4-984f-85cbbc1e4812.PNG)

Here a jsFiddle to test 
http://jsfiddle.net/Xelloss/s6VNm/

To test on IE8 you have to run the JSFiddle and access the following url on IE8
http://jsfiddle.net/draft/

danzel:
Closing #3035 as a duplicate of this. There are a few more details in there (It affects Polyline too, probably all vml shapes).
We probably just need to re-add the stroke/fill properties when the line is added back to the map?

danzel:
@kerryarts ^^

PR #3662, opened 2015-07-23
Fix for #2809

fminuti:
Fix for "Vector layers loose styling in IE8 when removing and adding back to the map" #2809


mourner:
Awesome, thanks!

PR #3663, opened 2015-07-23
FlyTo with targetCenter == currentCenter not animated

fminuti:
Please see http://jsfiddle.net/v1so6m56/4/

If targetCenter is the same of currenteCenter, the parameter u1 (the distance between "to" and "from") is 0.

The proposed patch resolve this by forcing the distance to 1.

### DIFF
--- a/src/map/anim/Map.FlyTo.js
+++ b/src/map/anim/Map.FlyTo.js
@@ -19,7 +19,7 @@ L.Map.include({
 
 		var w0 = Math.max(size.x, size.y),
 		    w1 = w0 * this.getZoomScale(startZoom, targetZoom),
-		    u1 = to.distanceTo(from),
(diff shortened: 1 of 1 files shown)


mourner:
Thanks for the PR! The proposed change looks like a hack though, I wonder if I can find a more elegant solution...

IvanSanchez:
After fiddling with this for quite a while, I don't think it's easy to come up with a version of the `r(i)` function that works OK for a distance of zero. Using a minimum distance of one pixel (as this PR does) doesn't sound like a bad idea after all.

mourner:
I'll take a look.

mourner:
So, I was going to port over the special case [from here](https://github.com/mapbox/easey-DEAD/blob/mb-pages/src/easey.js#L270-L282) but it's complicated and this one seems to produce exactly the same results. :)

Issue #2006, opened 2013-09-02
Scale control is inaccurate


mataamad:
The scale control is inaccurate, for example the square areas south west of Palm Springs / Palm Desert should be exactly 1 mile by 1 mile:
http://www.openstreetmap.org/#map=12/33.8195/-116.3038

e.g. compare with Google Maps: https://www.google.com/maps/preview#!data=!1m4!1m3!1d70454!2d-116.610454!3d33.7160376

At zoom level 12 the squares show up as around 1.3 miles wide, ~0.9 miles wide at zoom level 11, and 0.8 miles wide at zoom level 10.

I'm pretty sure this isn't just an open street map issue because we get the same problem with our own custom tile server.
![invalid map scales](https://f.cloud.github.com/assets/2156335/1069314/4811060e-1412-11e3-9e29-85dcfa04b1da.png)

MarZab:
Sorry to open up an old bug.

![untitled 4](https://cloud.githubusercontent.com/assets/1311249/8720109/fff12f44-2bb0-11e5-9a2d-3b7dd26a0b18.png)

Here ( [gist](https://gist.github.com/MarZab/7e3b1522ebc383f448a0), [preview](http://zabreznik.net/wp-content/uploads/2015/05/lf.html) ) is a example of this issue on leaflet-1.0.0-b1 using a linear CRC and no custom CSS with this fix:
```
.leaflet-control-scale-line {
  box-sizing: border-box;
}
```

mourner:
Hmm, I wonder what's up with the wonky zoom animation in your example... 
Can you clarify what's exactly wrong, and what's the expected output?

MarZab:
Yeah the wonky zoom is something I've been wondering about for ages as well :)

I would expect leaflet to have a somewhat accurate scale on a generic page by default. 
The example I gave has a linear projection so we can rule out projection distortion.
Image 1 is without the fix and image 2 is with the fix - try it with subtracting the coordinates at the beginning and end of the scale.

PR #3668, opened 2015-07-24
Fix Control Scale padding

fminuti:
The Control Scale is too big: the horizontal padding (5px) makes the line too wide (and incorrect).

This PR adds a <span> element around the scale text with the appropriate padding.
The difference:

Before:
before

After:
after

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -380,7 +380,7 @@
 	border: 2px solid #777;
 	border-top: none;
 	line-height: 1.1;
-	padding: 2px 5px 1px;
+	padding: 2px 0 1px;
 	font-size: 11px;
 	white-space: nowrap;
 	overflow: hidden;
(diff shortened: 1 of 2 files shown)


pedroetb:
I also detected this issue, while working with a map with a grid layer (where each box is 5km x 5km).

Here, you can see the control scale with the padding (incorrect):
withpadding

And here, without the padding (correct):
withoutpadding

mourner:
Yes, going to merge this.

mourner:
The purpose of box-sizing in CSS was to make the width setting not dependent on padding, I wonder why it doesn't work here. Needs looking into.

yohanboniface:
For reference: https://github.com/Leaflet/Leaflet/issues/2006 and https://github.com/fminuti/Leaflet/commit/fd86dea4c252cf25f2708f7a572bf505dac6684e

yohanboniface:
See also #3906 and http://playground-leaflet.rhcloud.com/gor/edit?html,output

PR #3671, opened 2015-07-24
Gridlayer opacity flicker fix

### LINKED ISSUE #3652: Opacity flickr when zooming WMS Layers

w8r:
#3652 

* Layer opacity applied on container
* Tile fade animation in 0...1 values

* IE<=9:
  * opacity set on tiles individually, no animation, as before

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -156,13 +156,13 @@ L.GridLayer = L.Layer.extend({
 
 	_updateOpacity: function () {
 		if (!this._map) { return; }
-		var opacity = this.options.opacity;
 
 		// IE doesn't inherit filter opacity properly, so we're forced to set it on tiles
-		if (!L.Browser.ielt9 && !this._map._fadeAnimated) {
-			L.DomUtil.setOpacity(this._container, opacity);
+		if (L.Browser.ielt9 || !this._map._fadeAnimated) {
 			return;
 		}
+
+		L.DomUtil.setOpacity(this._container, this.options.opacity);
 
 		var now = +new Date(),
 			nextFrame = false,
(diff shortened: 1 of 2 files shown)


src/layer/tile/GridLayer.js
```diff
@@ -156,14 +156,14 @@ L.GridLayer = L.Layer.extend({
 
 	_updateOpacity: function () {
 		if (!this._map) { return; }
-		var opacity = this.options.opacity;
 
 		// IE doesn't inherit filter opacity properly, so we're forced to set it on tiles
-		if (!L.Browser.ielt9 && !this._map._fadeAnimated) {
-			L.DomUtil.setOpacity(this._container, opacity);
+		if (L.Browser.ielt9 && !this._map._fadeAnimated) {
```
mourner: opacity in <IE9 is never animated, so the second condition is superfluous, right?

w8r: exactly, wanted to leave a note here as well

w8r: that was in the original code, I would change `&&` to `||` to make it work

mourner: Yeah, `||` sounds good.

w8r:
Pushed that change as well

mourner:
Squash into one commit and force-push please.

w8r:
Done

mourner:
Thanks!

PR #3675, opened 2015-07-27
Set popup source to marker when marker.bindPopup gets an instance of L.Popup

Eschon:
If you use an instance of L.Popup to bind a popup to a marker in the current stable version (it's fixed in master) the source of the popup is not set. So if you close a popup the popupclose event is only fired on the map and not on the marker. See [fiddle](http://jsfiddle.net/Eschon/a87woskf/)

This probably won't be a big deal for most people but it broke a few things in my code.


mourner:
Thanks!

PR #3678, opened 2015-07-27
Fix baselayerchange event argument (fix #3677)

### LINKED ISSUE #3677: map.on('baselayerchange', {}) returns layer instead of event, unclear docs
From what I can see in the 1.0 beta docs, the `baselayerchange` event handler should receive a [LayerEvent](https://mourner.github.io/Leaflet/reference.html#layer-event) or [LayersControlEvent ](https://mourner.github.io/Leaflet/reference.html#layers-control-event) depending on whether you believe the [Map API doc](https://mourner.github.io/Leaflet/reference.html#map-events) or the [Control.Layers API doc](https://mourner.github.io/Leaflet/reference.html#control-layers). They both say that the event is triggered when you change the baselayer via the layer control, but say that different events will be fired. 

In any case, both types of events have a `layer` property, but it appears that instead of the event object, the actual layer being changed to is returned instead.

JSBin demonstrating: 
http://jsbin.com/wevezidofo/5/edit?js,console,output

If you use the layer control to switch layers, you will see from the console output that the event properties are undefined, but the object returned has all the properties of a TileLayer.

yohanboniface:
Should fix #3677 
Unittest wasn't actually run, because of missing `map.setView`.

### DIFF
--- a/spec/suites/control/Control.LayersSpec.js
+++ b/spec/suites/control/Control.LayersSpec.js
@@ -6,18 +6,24 @@ describe("Control.Layers", function () {
 	});
 
 	describe("baselayerchange event", function () {
+		beforeEach(function () {
+			map.setView([0, 0], 14);
+		});
+
 		it("is fired on input that changes the base layer", function () {
-			var baseLayers = {"Layer 1": L.tileLayer(), "Layer 2": L.tileLayer()},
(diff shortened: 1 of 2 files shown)


mourner:
:+1:

PR #3681, opened 2015-07-28
fix imageoverlay setBound when not added to map

fminuti:
If imageoverlay#setBounds is called when the layer is not added to a map, the _reset method gets errors on execution (sorry for former incomplete PR)


mourner:
Good catch!

PR #3682, opened 2015-07-28
Do not stop mousedown event propagation on L.Draggable

yohanboniface:
Rationale is to let Leaflet plugins use the mousedown event.

@mourner feedback? :)

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -46,9 +46,8 @@ L.Draggable = L.Evented.extend({
 	_onDown: function (e) {
 		this._moved = false;
 
-		if (e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
-
-		L.DomEvent.stopPropagation(e);
+		if (L.Draggable._dragging || e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
+		L.Draggable._dragging = true;  // Prevent dragging multiple objects at once.
 
 		if (this._preventOutline) {
 			L.DomUtil.preventOutline(this._element);
@@ -142,5 +141,6 @@ L.Draggable = L.Evented.extend({
 		}
 
 		this._moving = false;
+		L.Draggable._dragging = false;
 	}
 });


mourner:
I remember removing the global `dragging` switch hack some time ago because of a different issue. Generally global switches seem like a bad practice and may cause problems in unexpected places. Can you look through the history and git blame and the code and see why it was removed?

yohanboniface:
Humm, seems something like that: https://github.com/Leaflet/Leaflet/commit/c97c48325afee8973311f9a7d54f4392da43d922

yohanboniface:
Problem seems to be when there are more than one map on a same page (so sharing L.Draggable).

yohanboniface:
So, after thinking a bit more about it, I'm not sure we are falling in the same trap: the issue fixed by c97c483 was because ZoomAnimation also was relying on `L.Draggable._disabled`, which is not the case in the current PR. And I don't see how we would be *actually dragging* two maps in the same time.

I understand having a global switch is not the best pattern ever, but in the same time I'm sure eating the `mousedown` is a bad practice I'd love to fix. So, as for now, I'm in favour of going this way.
But I'd understand that you prefer not taking any risk, and I can look for another option. :)

src/dom/Draggable.js
```diff
@@ -46,9 +46,8 @@ L.Draggable = L.Evented.extend({
 	_onDown: function (e) {
 		this._moved = false;
 
-		if (e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
-
-		L.DomEvent.stopPropagation(e);
+		if (L.Draggable._dragging || e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
+		L.Draggable._dragging = true;  // Prevent from dragging the map when dragging a vector.
```
mourner: I think "Prevent dragging multiple objects at once" would be a better comment.

mourner:
OK sounds good, let's proceed with this :+1:

yohanboniface:
Updated comment :)

mourner:
Can you verify that it works properly across browsers, including mobile ones, and doesn't cause any side effects?

yohanboniface:
I'll do my best :)

yohanboniface:
Tested on the browsers around me:
- Ubuntu FF 29
- Ubuntu Chromium / Chrome
- Ubuntu Touch Browser
- Android Chrome
- iPhone Chrome
- iPhone Safari
- Mac Chrome
- Mac Safari

As you can see, I'm missing colleagues with Windows :/

mourner:
@yohanboniface solid testing, thanks a lot! I think it's enough for now :)

PR #3687, opened 2015-07-29
Fix Marker dragging with changing Icon

fminuti:
Fix Maker dragging and changing the Icon inside a 'drag[start]' event handler.
Try http://jsfiddle.net/v1so6m56/6/ If you try to drag the green leaf you'll see that the shadow doesn't move, and if you try to drag again the marker starts over from original position.

In http://jsfiddle.net/v1so6m56/7/ is implemented the proposed PR: try to drag again the marker: the shadow moves and if you drag again the marker starts from last position.

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -130,7 +130,10 @@ L.Marker = L.Layer.extend({
 		}
 
 		this._icon = icon;
-		this._initInteraction();
+
+		if (addIcon) {
+			this._initInteraction();
+		}
 
 		if (options.riseOnHover) {
 			this.on({


IvanSanchez:
I've cleaned up the code a tiny bit and committed as 39f03304.

I fail to see any harmful side effects from this. Thanks!

yohanboniface:
> I fail to see any harmful side effects from this.

Found one, but really an edge case: in Leaflet.Editable, there is a situation where I need to remove an icon from the map then readd it while transferring it from a Marker instance to another (start dragging of a middle marker; needs is to workaround the fact that browsers will not fire touchend when event source is no more on DOM).
See https://github.com/yohanboniface/Leaflet.Editable/issues/35 for a bit of context.
Not sure this needs any action though, I can call `_initInteraction` myself (even if this piece of code starts to become a very bad hack: https://github.com/yohanboniface/Leaflet.Editable/blob/679502e7dd7e426b81d86fc869f170148f024635/src/Leaflet.Editable.js#L507-L515 :/ )

gorshkov-leonid:
this fix is broken in https://github.com/Leaflet/Leaflet/issues/3976 ?

PR #3688, opened 2015-07-29
Fix Internet Explorer 11 map overflow outside container

### LINKED ISSUE #2788: 0.8-dev IE10 map overflows outside of map container

fminuti:
Please see http://jsfiddle.net/v1so6m56/8/ : try to zoom (by double click, o wheel) with IE 11 (and IE 10 emulated in Dev Tools). The map overflows outside the container.

The proposed PR is tested in  http://jsfiddle.net/v1so6m56/9/ 

Basically this change disable the usage of css transform3d for ie3d. 

Microsoft Edge on Windows 10 is not affected (L.Browser.ie3d is false, because L.Browser.ie is false )

Should Fix #2788

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -143,7 +143,7 @@ L.DomUtil = {
 		var pos = offset || new L.Point(0, 0);
 
 		el.style[L.DomUtil.TRANSFORM] =
-			'translate3d(' + pos.x + 'px,' + pos.y + 'px' + ',0)' + (scale ? ' scale(' + scale + ')' : '');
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Wow, nice catch! :+1:

I wonder if it would be possible to simplify the code by using `translate()` instead of `translate3d()` for every browser instead?

mourner:
@IvanSanchez no, it would disable hardware acceleration in some browsers, so let's go as is for now.

fminuti:
I don't think. translate3d should be hw accelerated (at least in webkit).

PR #3692, opened 2015-07-30
Do not set draggable flags during a zoom animation

### LINKED ISSUE #3685: Dragging while zoom animation is active breaks dragging
When the map is locked while the tiles are loading, the user experience becomes not as smooth as it could be. Some map tile providers load map tiles slower than others. I have seen map tile requests that take more than a second to load. That is one second where the map cannot be dragged and moved around. The map becomes clunky to use with speed and efficiency.

IvanSanchez:
Fixes #3685.

Has the side effect of making draggable markers jump around if dragged during a zoom animation, but I don't expect anyone to drag a marker within 0.25 secs of starting a zoom animation.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -46,14 +46,14 @@ L.Draggable = L.Evented.extend({
 	_onDown: function (e) {
 		this._moved = false;
 
+		if (L.DomUtil.hasClass(this._element, 'leaflet-zoom-anim')) { return; }
+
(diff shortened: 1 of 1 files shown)


bsudekum:
@mourner this is a pretty big bug that is affecting a lot of people. Should there be release before 1.0.0 that includes this fix?

/cc @jfirebaugh

mourner:
@bsudekum yes, I was thinking of a 0.7.4 release.

bsudekum:
@mourner :+1:

PR #3693, opened 2015-07-30
Prevent division by zero in FlyTo when currentCenter == targetCenter

IvanSanchez:
An alternative to #3663. Skip division by the distance from start to end when that distance is zero.

### DIFF
--- a/src/map/anim/Map.FlyTo.js
+++ b/src/map/anim/Map.FlyTo.js
@@ -51,7 +51,7 @@ L.Map.include({
 				this._flyToFrame = L.Util.requestAnimFrame(frame, this);
 
 				this._move(
-					this.unproject(from.add(to.subtract(from).multiplyBy(u(s) / u1)), startZoom),
+					this.unproject(u1 ? from.add(to.subtract(from).multiplyBy(u(s) / u1)) : from, startZoom),
 					this.getScaleZoom(w0 / w(s), startZoom));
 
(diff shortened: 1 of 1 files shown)


src/map/anim/Map.FlyTo.js
```diff
@@ -24,7 +24,7 @@ L.Map.include({
 		    rho2 = rho * rho;
 
 		function r(i) {
-			var b = (w1 * w1 - w0 * w0 + (i ? -1 : 1) * rho2 * rho2 * u1 * u1) / (2 * (i ? w1 : w0) * rho2 * u1);
+			var b = (w1 * w1 - w0 * w0 + (i ? -1 : 1) * rho2 * rho2 * u1) / (2 * (i ? w1 : w0) * rho2);
```
mourner: what's going on here? it can't be removed like that because of the left part of the formula

IvanSanchez: Oh, wait, I've misread the left part. I was assuming `u1` multiplied everything.

fminuti:
I don't think this solution is working:
try
http://jsfiddle.net/v1so6m56/10/
where I copied your implementation.

fminuti:
Sorry, you changed your implementation while I was writing.
Here http://jsfiddle.net/v1so6m56/12/
the new implementation still not working correctly

PR #3694, opened 2015-07-30
L.Control.Layers shows disabled inputs for non-visible tilelayers

### LINKED ISSUE #252: Don't display or disable layers in Layers Control when not in a supported zoom
I am using the new layers control to show two layers. One layer has a minZoom of 13. When zoomed out, you can still change to the layer in the layers control. The layer should either be disabled or hidden in the control, so a user can't get the layer into a bad zoom level.

IvanSanchez:
Fixes #252

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -26,11 +26,18 @@ L.Control.Layers = L.Control.extend({
 		}
 	},
 
-	onAdd: function () {
+	onAdd: function (map) {
 		this._initLayout();
 		this._update();
 
+		this._map = map;
+		map.on('zoomend', this._checkDisabledLayers, this);
+
 		return this._container;
+	},
+
+	onRemove: function() {
+		this._map.off('zoomend', this._checkDisabledLayers, this);
 	},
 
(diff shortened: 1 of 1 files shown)


src/control/Control.Layers.js
```diff
@@ -254,10 +262,26 @@ L.Control.Layers = L.Control.extend({
...
+		var inputs = this._form.getElementsByTagName('input'),
+		    input,
+		    layer,
+		    zoom = this._map.getZoom();
+
+		for (var i = inputs.length - 1; i >= 0; i--) {
+			input = inputs[i];
+			layer = this._layers[input.layerId].layer;
+			input.disabled = (layer.options.minZoom !== undefined && zoom < layer.options.minZoom) ||
+			                 (layer.options.maxZoom !== undefined && zoom > layer.options.maxZoom);
```
mourner: I wonder if we should add `layer.supportZoom(z)` for things like this...

IvanSanchez: `L.GridLayer` does the zoom check twice already, so it *might* be worth to refactor. But I don't have a strong opinion about it right now.

PR #3695, opened 2015-07-30
add missing class leaflet-zoom-anim in flyTo

fminuti:
If you have any layer the uses the ‘leaflet-zoom-hide’ to be hidden while zooming, without the ‘leaflet-zoom-anim’ class in flyTo, that layer twill be shown while animating the fly.

### DIFF
--- a/src/map/anim/Map.FlyTo.js
+++ b/src/map/anim/Map.FlyTo.js
@@ -55,11 +55,15 @@ L.Map.include({
 					this.getScaleZoom(w0 / w(s), startZoom));
 
 			} else {
+
+				L.DomUtil.removeClass(this._mapPane, 'leaflet-zoom-anim');
 				this
 					._move(targetCenter, targetZoom)
 					._moveEnd(true);
 			}
 		}
+
+		L.DomUtil.addClass(this._mapPane, 'leaflet-zoom-anim');
 
 		this._moveStart(true);
 


mourner:
That doesn't look right. Adding this class will add CSS transitions to layer transforms, making `flyTo` extremely slow because it will try to animate between every frame.

fminuti:
Ok, I've seen what you are saying in Chrome Dev Console by slowing animation.

But I think that the bug (not hiding the leaflet-zoom-hide layers) should be resolved anyway, but I have no idea how by now.

mourner:
@fminuti I think it's fine to show the layer because the same logic that sets the layer position on round zooms will work on each frame on flyTo. What's your use case?

fminuti:
I don't have a real use case. 
If I need, I can animate the layer or hide it via javascript. 
I opened this for consistency: I've seen you use the leaflet-zoom-hide class, i.e. in Marker#_initIcon (https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L106) or in Leaflet.heat (https://github.com/Leaflet/Leaflet.heat/blob/gh-pages/src/HeatLayer.js#L88).

mourner:
The zoom-hide functionality was done for those layers where you don't want or need to implement complicated calculations for CSS3 transform transition to animate them on zoom, but you don't have to do this for flyTo because it's not transition-based, so I think it's fine.

PR #3701, opened 2015-07-31
Experiment: Use preprojected points and reduce object allocations

wpf500:
This pull request isn't intended for merging (it almost certainly breaks things), its more of a proof of concept around more performant zooming with vector layers.

This is what currently happens when you zoom:
image

It seems that Leaflet spends about half the time projecting, the other half rendering. The screenshot above shows three problems I think: 1) projecting takes a while, 2) there are too many GC events and 3) rendering takes a while.

I'm not sure there's much to be done to improve rendering, batching together canvas contexts (#3577) doesn't actually seem to have the impact I thought it would. However for 1) and 2), we could pre-project the coordinates and remove a lot of the object allocation (mainly new `Point` instances)

Zooming then looks like this:
image
Object allocations pre/post optimization:
image

This seems pretty good. It works well for me but then I have a very specific use case. Obviously not everyone is going to want to pre-project their features (for reference my pre-projection code is here: https://gist.github.com/wpf500/37d288c3991527a21017), but it could probably be done in `_convertLatLngs`? The code changes I've made are very specific to my project (I'm not suggesting this gets merged), but before trying to write a general implementation I thought it was worth asking what things might break as a result that I haven't thought about?

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -103,8 +103,13 @@ L.Polyline = L.Path.extend({
 	},
 
 	_setLatLngs: function (latlngs) {
-		this._bounds = new L.LatLngBounds();
-		this._latlngs = this._convertLatLngs(latlngs);
+		this._bounds = new L.Bounds();
+		this._pxBounds = new L.Bounds(new L.Point(), new L.Point());
+		this._latlngs = latlngs;
+		this._points = [];
+		this._rings = [];
+		var unitScale = L.CRS.EPSG3857.scale(0);
+		this._convertLatLngs(latlngs, this._points, this._rings, this._bounds, unitScale);
 	},
 
 	_defaultShape: function () {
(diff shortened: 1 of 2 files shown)


src/layer/vector/Polyline.js
```diff
@@ -103,8 +103,13 @@ L.Polyline = L.Path.extend({
...
 
 	_setLatLngs: function (latlngs) {
-		this._bounds = new L.LatLngBounds();
-		this._latlngs = this._convertLatLngs(latlngs);
+		this._bounds = new L.Bounds();
+		this._pxBounds = new L.Bounds(new L.Point(), new L.Point());
+		this._latlngs = latlngs;
+		this._points = [];
+		this._rings = [];
+		var unitScale = L.CRS.EPSG3857.scale(0);
```
IvanSanchez: You most probably want to use `this._map.CRS` instead of 3857.

wpf500: I tried but unfortunately `this._map` is `undefined` when `_setLatLngs` gets called. Like I said this is just a proof of concept, code isn't ready.

mourner:
Really cool! Now I wonder if the same idea can be done in a clean way that doesn't break things — this may be a challenge...

PR #3711, opened 2015-08-05
Update TileLayer.WMS.js

granci:
added WMS parameters 'time' and 'elevation'


IvanSanchez:
Not needed. By specifying a `time` and/or `elevation` option into the WMS layer, that will be added as a WMS param. See https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/TileLayer.WMS.js#L28

The WMS 1.1.1 spec clearly state that time and elevation are optional. Thus, there is no reason for Leaflet to send them by default with empty values.

Issue #3721, opened 2015-08-09
maxZoom option messes with tile  layer ordering


patrickarlt:
I've found a regression in 1.0.0-beta.1 it looks like `L.tileLayer` gets a z-index based on its maxZoom option. This has some undesirable effects on layer ordering. That are clearly seem in these 2 fiddles. I think this can be traced to https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L214.

It looks like this is really trivial to fix with zIndex but if this is the expected behavior we should note it in the changelog.

* 0.7.3 http://jsfiddle.net/858asofb/2/ second tile layer is over the OSM tiles
* 1.0.0-beta.1 http://jsfiddle.net/858asofb/1/ the second tile layer is under the OSM tiles
* 1.0.0-beta.1 with zIndex http://jsfiddle.net/ywy9b3kp/1/

![edit_fiddle_-_jsfiddle](https://cloud.githubusercontent.com/assets/378557/9157239/aed92ece-3eaa-11e5-9930-ea5f249e6794.png)
![edit_fiddle_-_jsfiddle](https://cloud.githubusercontent.com/assets/378557/9157241/dd63f404-3eaa-11e5-9497-c35c1ba1c93c.png)

PR #3722, opened 2015-08-09
Always force GridLayer to have a zIndex

patrickarlt:
This is a super naive fix for https://github.com/Leaflet/Leaflet/issues/3721. If all grid containers have a z index of 1 they will default to DOM ordering which is usually the order they were added in which is the expected behavior. Also since the container now has a z index all their children (levels) get ordered relative to their parent.

@mourner let me know if my reading of the source code was correct and Ill squash this down and merge it.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -9,12 +9,12 @@ L.GridLayer = L.Layer.extend({
 
 		tileSize: 256,
 		opacity: 1,
+		zIndex: 1,
 
 		updateWhenIdle: L.Browser.mobile,
 		updateInterval: 200,
-
+		
 		attribution: null,
-		zIndex: null,
 		bounds: null,
 
 		minZoom: 0
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -13,7 +13,7 @@ L.GridLayer = L.Layer.extend({
 
 		updateWhenIdle: L.Browser.mobile,
 		updateInterval: 200,
-		
+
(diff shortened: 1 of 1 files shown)


mourner:
> Also since the container now has a z index all their children (levels) get ordered relative to their parent.

Hmm the container should have its own stacking context because of `position: absolute` already... I'll look into it.

patrickarlt:
According to [MDN on when a new stacking context is created](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Understanding_z_index/The_stacking_context) 

> positioned (absolutely or relatively) with a z-index value other than "auto",

right now each container has `position: absolute` but no z-index. In http://jsfiddle.net/858asofb/1/ the html structure looks like this:

```html
<div class="leaflet-pane leaflet-tile-pane">
  <div class="leaflet-layer">
    <div class="leaflet-tile-container leaflet-zoom-animated" style=
    "z-index: 19; transform: translate3d(0px, 0px, 0px) scale(1);">
      <!-- OSM Tiles -->
    </div>
  </div>

  <div class="leaflet-layer">
    <div class="leaflet-tile-container leaflet-zoom-animated" style=
    "z-index: 18; transform: translate3d(0px, 0px, 0px) scale(1);">
      <!-- Other Tiles -->
    </div>
  </div>
</div>
```

We have a stacking context inside `leaflet-tile-pane` (`position:absolute` + `z-index: 200`) but each `leaflet-layer` only has `position: absolute` and thus it's children `leaflet-tile-container` participate in the stacking context provided by `leaflet-tile-pane`.

PR #3723, opened 2015-08-10
Build sourcemaps via magic-string

IvanSanchez:
Because having sourcemaps is cool. Pretty much based on https://www.npmjs.com/package/magic-string#bundling

Build system works just the same, just run `npm install` and `jake`, and the sourcemap for `leaflet-src.js` will be automatically generated.

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -1,6 +1,7 @@
 var fs = require('fs'),
     UglifyJS = require('uglify-js'),
     zlib = require('zlib'),
+    MagicString = require('magic-string'),
 
     deps = require('./deps.js').deps;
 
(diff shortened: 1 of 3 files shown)


build/build.js
```diff
@@ -69,12 +70,24 @@ function loadSilently(path) {
...
 
-function combineFiles(files) {
-	var content = '';
+function bundleFiles(files, copy) {
+	var bundle = new MagicString.Bundle();
+
 	for (var i = 0, len = files.length; i < len; i++) {
-		content += fs.readFileSync(files[i], 'utf8') + '\n\n';
+
+		console.log(files[i]);
```
mourner: leftover?

IvanSanchez: ooooops.

mourner:
Nice work! 
I wonder if it's better to have embedded map sources or external ones (e.g. `leaflet-src.map`).

IvanSanchez:
Not sure, I'm not an expert at sourcemaps.

Also, `magic-string` is not perfect. While a better, more compact sourcemap which doesn't include the full source code again might be possible, I don't think `magic-string` is able to generate it (or I haven't been able to discern how).

patrickarlt:
if you want to distribute the sourcemap via CDN (like I do with Esri Leaflet) it is WAY better to have external sourcemaps that way the file size doesn't get bloated. Embedded sourcemaps are HUGE because they contain all the source files that went into them so they usually end up doubling file size or more.

With external sourcemaps they dont get requested unless you turn on sourcemaps in your dev tools so they never affect download speeds.

mourner:
Cool! I see that Ivan changed the PR to have external maps.

PR #3726, opened 2015-08-10
added new method toBBoxArray()

stefanocudini:
This could be very useful for compatibility with 3rd party software


mourner:
Can you give us examples of that 3rd-party software?
I don't see a reason to include that in the core since formats 3rd-party software expects are very different, and it's easy enough to just use `getWest()`/etc methods.

PR #3729, opened 2015-08-11
GridLayer async mode

caiiiycuk:
In function GridLayer.createTile we should return tile immediately even if we draw in async mode. Otherwise undefined exception will be thrown.

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -3387,8 +3387,10 @@ map.fitBounds(bounds);</code></pre>
 		tile.width = size.x;
 		tile.height = size.y;
 
+		return tile;
+		
 		// draw something and pass the tile to the done() callback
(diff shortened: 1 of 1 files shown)


IvanSanchez:
You're right about returning the tile, but you're wrong about not calling the `done` callback (it's needed to pull the tile from the queue of pending tiles and fire the `load` event).

I can make some changes to the docs to reflect this.

caiiiycuk:
@IvanSanchez 
Callback should be called after ```return tile;```, if we call it before nothing happens. Adding timeout to this function looks complex. So i don`t have good idea how to fix this. Can i close PR?

IvanSanchez:
Superseded by #3730. Thanks for noticing the error!

PR #3755, opened 2015-08-18
add removeFrom function for marker

moklick:


reference.html
```diff
@@ -1484,6 +1484,14 @@ <h3 id="marker-options">Options</h3>
 		<td>Adds the marker to the map.</td>
 	</tr>
 	<tr>
+		<td><code><b>removeFrom</b>(
+			<nobr>&lt;<a href="#map-class">Map</a>&gt; <i>map</i> )</nobr>
+		</code></td>
+
+		<td><code><span class="hljs-keyword">this</span></code></td>
+		<td>Removes the marker to the map.</td>
```
IvanSanchez: Removes the marker *from* the map

IvanSanchez:
`removeFrom` is inherited from `L.Layer` anyway, so the right thing to do would be to remove `addTo`. I should be able to switch everything to Leafdoc anyway, so this won't be an issue.

PR #3759, opened 2015-08-18
Add `buffer` to GridLayer options.

colllin:
It pads the grid by the specified number of tiles beyond the map bounds.

Usage:

    L.tileLayer(theTilestackUrl, {
        buffer: 1
    });

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -17,8 +17,10 @@ L.GridLayer = L.Layer.extend({
 		zIndex: null,
 		bounds: null,
 
-		minZoom: 0
-		// maxZoom: <Number>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Given that there is a plugin for that (https://github.com/TolonUK/Leaflet.EdgeBuffer), I see no big reason to merge this into the core... :-/

colllin:
@IvanSanchez Thanks for pointing me to the plugin.  Up to you what to do with the PR.  I'll submit this same PR to that plugin, since I think it's a cleaner approach.  Either way it looks like I'm forced to use a plugin for now.

IvanSanchez:
Actually this is a duplicate of #3551, so closing.

PR #3776, opened 2015-08-26
Use radius if radiusY is not set

knabar:
`_radiusY` may not be set, use `_radius` instead in that case as done in other places.


mourner:
Thanks!

PR #3778, opened 2015-08-26
Clear tilelayer's BG buffer on zoomanim end if out of zoom limits

### LINKED ISSUE #3777: Tiles still shown after zooming out of minZoom/maxZoom
This is best explained with a fiddle: http://jsfiddle.net/ryy8abzt/2/

In there, the Toner layer is visible only at zoom levels 17 and 18. When zooming out from 17 to 16, is is still seen.

I don't really care about the visibility of the tile layer during the zoom animation. But (obviously) the layer has to be hidden at some point.

<s>Reproducible on 0.7.x with a variety of browers. Gotta check with 1.0-beta1.</s>Reproducible **only** on 0.7.x with a variety of browsers.

IvanSanchez:
Fixes #3777

### DIFF
--- a/src/layer/tile/TileLayer.Anim.js
+++ b/src/layer/tile/TileLayer.Anim.js
@@ -29,6 +29,11 @@ L.TileLayer.include({
 		// force reflow
 		L.Util.falseFn(bg.offsetWidth);
 
+		var zoom = this._map.getZoom();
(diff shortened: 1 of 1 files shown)


dagjomar:
omg that was quick ^_^

PR #3795, opened 2015-09-01
quick start guide should reference the Leaflet maintained CDN

mejackreed:
Rather than the cdnjs, reference the one maintained by Leaflet


IvanSanchez:
Thanks! It's always difficult to *not* miss a reference to the CDN urls

mejackreed:
:smile_cat:

Issue #3708, opened 2015-08-04
mouseover and mouseout events do not behave like mouseenter and mouseleave on markers


Eschon:
According to #1144 Leaflets `mouseover` and `mouseout` should behave like `mouseenter` and `mouseleave`. However [this fiddle](http://jsfiddle.net/Eschon/fevwbrjq/) shows that this is not the case.

If you move the mouse over the outer ring of the marker the popup shows up but as soon as you enter the inner circle it is closed because `mouseout` is fired. If you move back to the outer ring the popup shows up again.

mourner:
Thanks for ticketing, this is a pretty interesting use case. I guess it'll be the same if you use `divIcon` with a nested HTML too.

mourner:
cc @yohanboniface

Eschon:
> this is a pretty interesting use case

I have to make a bunch of different icons that are all the same except for the background colour.
So if I do it this way I can load one svg file and change the background.

IvanSanchez:
>  So if I do it this way I can load one svg file and change the background.

Maybe you can try to use `:hover` CSS selectors and do some trickery there. Using .pngs and changing the background position on `:hover` is a well-known technique.

mourner:
@Eschon yeah, you can achieve the same look by using `divIcon` and `border-radius` for the circle look plus borders.

Eschon:
@IvanSanchez CSS would work if what I described above was the only problem. But the icons are used on multiple maps and the colors can be completely different on two separate maps. This would still be possible but probably it would also upset the guy who has to prepare those icons for me.

@mourner divIcons could work but as you said above they have the same problem:
http://jsfiddle.net/Eschon/spkj72te/

yohanboniface:
I think we have two issues here:

- firstly, effectively mouseout/mouseover are fired when the mouse is moving from a marker div element to a child of this element and vice-versa
- secondly, when the popup opens, it triggers a mouseout because the mouse is now on the popup and no more on the div itself

So event if we fix the first issue, the second will remain and of course it 0is not a bug that we can fix ;)

About the first issue, I'm not sure what we want to do. This behaviour is the natural DOM behaviour, but indeed I've read that in Leaflet 'mouseout' and 'mouseover' should act as 'mouseleave' and 'mouseenter'.

@mourner do you confirm this? I.e. why don't we let people use 'mouseleave' and 'mouseenter' when they need this behaviour?
[...]

PR #3803, opened 2015-09-03
Make sure mouse has an external relatedTarget on mouseout/over 

### LINKED ISSUE #3797: mouseover is not fired on nested HTML inside marker
The example mentioned in this [issue](https://github.com/Leaflet/Leaflet/issues/3708) doesn't work properly in master for hovering markers with nested HTML. Expected behaviour is:
1. mouseover event on marker when mouse is entering it from the map
2. mouseout event as mouse moves from marker itself onto the nested HTML
3. another mouseover event fired by marker, which bubbles from the nested HTML

The second mouseover is not fired because of [this commit](https://github.com/Leaflet/Leaflet/commit/c1f21eb38f62b73977a8a7654297c4863ca439d3), preventing bubbling for hover events. 

That makes hovering markers with nested HTML unusable. Any ideas?

yohanboniface:
Fix #3797 #3708 

This commit controls that, on `mouseout` or `mouseover`, the relatedTarget was an external DOM element (and thus not a child).

This fixes #3797, but not reproducing the exact behaviour of 0.7.5, which conflicts with other choices on event management: basically, in 0.7.5 `mouseover` and `mouseout` was still bubbling (so fired twice when moving mouse from a div icon to its child), while now we are not firing those events when relatedTarget is a descendant element of the listener.

The frustrating point is that `happen` does not seem to be able to set the `relatedTarget` property, which prevents us from testing the scenarios. I'll have a look if this can be fixed on `happen`.

I've also renamed `_checkMouse` to `_isExternalTarget` because I found myself always needing to check the method content to remember what it is actually doing. I'm open to better names, though ;)

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -693,7 +693,7 @@ describe("Map", function () {
 			expect(mapSpy.called).not.to.be.ok();
 		});
 
-		it("mouseout is only forwared if fired on the original target", function () {
+		it("mouseout is forwarded if fired on the original target", function () {
 			var mapSpy = sinon.spy(),
 			    layerSpy = sinon.spy(),
 			    otherSpy = sinon.spy();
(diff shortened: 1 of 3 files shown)


yohanboniface:
https://github.com/tmcw/happen/pull/17

yohanboniface:
Added unittests (now that happen accepts relatedTarget option).

mourner:
On my plate. Can you rebase so that it doesn't conflict with master?

yohanboniface:
Rebased on master, master :)

ilfa:
Branch need rebase. Again.

yohanboniface:
Done, thanks :)

Issue #3785, opened 2015-08-28
Consistency between Circle and CircleMarker radiuses


IvanSanchez:
The factory for `L.Circle` takes in a `radius` parameter, whereas `L.CircleMarker` takes in the options parameter with a `radius` option inside.

For consistency, it would be very nice to have both classes have the radius as an option (instead of a constructor parameter).

yohanboniface:
I agree. With retrocompatibility in mind (`if (typeof options === "string")…`), this can be done within the 1.0 jump I think.

yohanboniface:
Committed with 3924d49bcdd4d88099b0d9a3196ebcd29479a4f7

PR #3805, opened 2015-09-04
Consistency between L.circle and L.circleMarker

IvanSanchez:
Implements #3785

### DIFF
--- a/src/layer/vector/Circle.js
+++ b/src/layer/vector/Circle.js
@@ -5,10 +5,10 @@
 
 L.Circle = L.CircleMarker.extend({
 
-	initialize: function (latlng, radius, options) {
+	initialize: function (latlng, options) {
 		L.setOptions(this, options);
 		this._latlng = L.latLng(latlng);
-		this._mRadius = radius;
+		this._mRadius = options.radius;
 	},
 
 	setRadius: function (radius) {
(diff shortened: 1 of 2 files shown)


src/layer/vector/Circle.js
```diff
@@ -62,6 +62,10 @@ L.Circle = L.CircleMarker.extend({
 	}
 });
 
-L.circle = function (latlng, radius, options) {
-	return new L.Circle(latlng, radius, options);
+L.circle = function (latlng, options, legacyOptions) {
+	if (typeof options === 'number') {
+		// Backwards compatibility with 0.7.x factory (latlng, radius, options?)
+		options = L.extend(L.extend({}, legacyOptions), {radius: options});
```
yohanboniface: When `options` is already an object, `radius` will be wrong, no?

yohanboniface: `if (typeof options === 'number')` sorry :)

yohanboniface:
I think it deserves one unittest for new API, and one unittest for backward compat, what do you think?

mourner:
Committed with https://github.com/Leaflet/Leaflet/commit/3924d49bcdd4d88099b0d9a3196ebcd29479a4f7

PR #3806, opened 2015-09-04
Call layer.getEvents before calling layer.onAdd (fix #3745)

### LINKED ISSUE #3745: openPopup() broken in 1.0
There are two serious bugs with openPopup():

1. It doesn't open popup, when it needs to pan first.
screen shot 2015-08-14 at 12 19 07

2. If the user pans the map, and it's still in the inertia animation (including the time where it only moves invisibly small distances), the map arrives at a totally wrong position. 
screen shot 2015-08-14 at 12 18 05

Wanted result:
screen shot 2015-08-14 at 12 19 20

JSFiddle: http://jsfiddle.net/858asofb/3/

Presents both on latest Chrome and Firefox under OS X 10.9.5.

yohanboniface:
Fix #3745

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -45,14 +45,14 @@ L.Layer = L.Evented.extend({
 		this._map = map;
 		this._zoomAnimated = map._zoomAnimated;
 
+		if (this.getEvents) {
+			map.on(this.getEvents(), this);
+		}
+
 		this.onAdd(map);
 
 		if (this.getAttribution && this._map.attributionControl) {
 			this._map.attributionControl.addAttribution(this.getAttribution());
-		}
-
-		if (this.getEvents) {
-			map.on(this.getEvents(), this);
 		}
 
 		this.fire('add');


mourner:
I remember that changing the order in `_layerAdd` could have unexpected consequences, so watch out for possible regressions...

yohanboniface:
/me protects itself behind the passing unittests ;)

But yeah, something will arise for sure! :(
Though, while thinking about it, I did not find any good reason to register events after adding the layers to the map only. So hopefully that's a good move in per se ;)

IvanSanchez:
The events are added after the element has been created and the reference to `this._map` has been set... I cannot think of any use case of an event that would break with those set up.

mourner:
Looking at this closer, maybe there's indeed no visible bad effects, but there could be duplicate calls of layer initializing functions when adding a layer, although this may be OK.

yohanboniface:
> I remember that changing the order in _layerAdd could have unexpected consequences, so watch 
> out for possible regressions...

polygon-bug

You meant *that*? :p

yohanboniface:
#3836

PR #3812, opened 2015-09-06
Allow subscribing to standalone popup events

lightoze:


mourner:
Can you describe the rationale behind this? What's the use case?

lightoze:
I am creating a standalone (i.e. not attached to a marker) popup layer and I need to know when it's closed.

IvanSanchez:
@lightoze There are already events fired from the popup https://mourner.github.io/Leaflet/reference.html#popup-removed

lightoze:
My bad, I was consulting 0.7 documentation at http://leafletjs.com/ and it's not mentioned there.
Also https://mourner.github.io/Leaflet/reference.html#popup-removed says that event name is `removed`, but it is in fact `remove`. Thank you for pointing out where to look at, anyway.

Issue #3515, opened 2015-06-04
Fix mobile not loading tiles during flyTo animation


mourner:
We don't load tiles during animation on mobile, but there really should be an exception for flyTo, since it doesn't make much sense without tiles during long animation. Perhaps there should be a bigger throttling interval though.

PR #3815, opened 2015-09-07
Fix TileLayer._update so it doesn't break drag-then-pinch interactions.

### LINKED ISSUE #3814: One-finger pan + two-finger zoom in = broken tiles on master
To reproduce, follow the instructions on #3798, and zoom **in** a couple of times. Tiles at the appropriate zoom level do not load (former tiles or grey space is seen instead).

Reproducible on Chrome on Android:
img_20150907_160642

And on Safari on iPad:

img_20150907_162016

And obviously we love windows phone:

img_20150907_162503

Obvious first place to look is the pinch-zoom interaction, about how the first two finger points are loaded.

Not a duplicate of #3798 because this affects loading of tile layers, not display of vectors.

Might or might not be related to #3765, #3659, #3683.

IvanSanchez:
Fixes #3814.

So `_update` was using a rounded zoom level, but internally there were calls to `createTile`, which depended on `this._tileZoom`. The result was that, after a drag-then-pinch interaction, if a tile to be loaded had a index `x:y:z`, the URL would be `z+1/y/x` (or z-1), but the key stayed the same - which was breaking havoc on the tile loading/unloading algorithm.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -354,7 +354,7 @@ L.GridLayer = L.Layer.extend({
 		if (!noUpdate || tileZoomChanged) {
 
 			this._tileZoom = tileZoom;
-			
+
 			if (this._abortLoading) {
 				this._abortLoading();
 			}
@@ -433,13 +433,14 @@ L.GridLayer = L.Layer.extend({
 		return new L.Bounds(pixelCenter.subtract(halfSize), pixelCenter.add(halfSize));
 	},
 
+	// Private method to load tiles in the grid's active zoom level according to map bounds
 	_update: function (center, zoom) {
 		var map = this._map;
 		if (!map) { return; }
 
 		if (center === undefined) { center = map.getCenter(); }
 		if (zoom === undefined) { zoom = map.getZoom(); }
-		if (this._tileZoom === undefined) { return; }
+		if (this._tileZoom === undefined) { return; }	// if out of minzoom/maxzoom
 
 		var pixelBounds = this._getTiledPixelBounds(center, zoom, this._tileZoom);
 
@@ -451,7 +452,9 @@ L.GridLayer = L.Layer.extend({
 			this._tiles[key].current = false;
 		}
 
-		if (Math.abs(zoom - this._tileZoom) > 1) { this._setView(center, zoom); }
+		// _update just loads more tiles. If the tile zoom level differs too much
+		// from the map's, let _setView reset levels and prune old tiles.
+		if (Math.abs(zoom - this._tileZoom) > 1) { return this._setView(center, zoom); }
 
 		// create a queue of coordinates to load tiles from
 		for (var j = tileRange.min.y; j <= tileRange.max.y; j++) {
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -454,7 +454,7 @@ L.GridLayer = L.Layer.extend({
 
 		// _update just loads more tiles. If the tile zoom level differs too much
 		// from the map's, let _setView reset levels and prune old tiles.
-		if (Math.abs(zoom - this._tileZoom) > 1) { return this._setView(center, zoom); }
+		if (Math.abs(zoom - this._tileZoom) > 1) { this._setView(center, zoom); return; }
 
 		// create a queue of coordinates to load tiles from
 		for (var j = tileRange.min.y; j <= tileRange.max.y; j++) {
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -343,8 +343,15 @@ L.GridLayer = L.Layer.extend({
 	},
 
 	_setView: function (center, zoom, noPrune, noUpdate) {
-		var tileZoom = Math.round(zoom),
-			tileZoomChanged = this._tileZoom !== tileZoom;
+		var tileZoom = Math.round(zoom);
+		if (this.options.maxZoom !== undefined) {
+			tileZoom = Math.min(tileZoom, this.options.maxZoom);
+		}
+		if (this.options.minZoom !== undefined) {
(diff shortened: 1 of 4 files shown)


src/layer/tile/GridLayer.js
```diff
@@ -431,12 +431,8 @@ L.GridLayer = L.Layer.extend({
 
 		if (center === undefined) { center = map.getCenter(); }
 		if (zoom === undefined) { zoom = map.getZoom(); }
-		var tileZoom = Math.round(zoom);
 
-		if (tileZoom > this.options.maxZoom ||
-			tileZoom < this.options.minZoom) { return; }
```
yohanboniface: I don't get why you remove this check

IvanSanchez: Somehow I assumed that check was already wherever `this._tileZoom` was being calculated. Apparently it is not.

IvanSanchez:
Note to self: check out examples with `maxZoom`, `minZoom` and `maxNativeZoom`, and see why the unit tests fail at gridlayer unused zoom levels.

src/layer/tile/GridLayer.js
```diff
@@ -446,11 +449,13 @@ L.GridLayer = L.Layer.extend({
 			this._tiles[key].current = false;
 		}
 
+		if (zoom < this.options.minZoom) { return; }
```
IvanSanchez: I changed `_update` so it works on the gridlayer's "active zoom pyramid level" only, which is good. But if a tile layer has e.g. `minZoom=16` and the map is at zoom e.g. 2, then the code would generate a few million ``s, then try to delete them later.

No check is needed for `maxZoom` because I really don't care about creating *fewer* tiles than normal, then deleting them.

mourner: > then the code would generate a few million s, then try to delete them later.

That sounds horrible.

IvanSanchez: Yeah, that's why I added that check :-)

IvanSanchez: Yeah, that's why I added that check :-)

IvanSanchez:
There is a worrying edge case. In a big enough tablet it's possible to pinch-zoom out 5 or 6 levels at once. This means that the first `_update` will run with `this._tileZoom` 5 or 6 times greater than `map._zoom`, and that will create 4^5 to 4^6 temporary tiles, that will be removed by `_abortLoading` when the second `_update` is run (when the zoom animation ends).

It works, except for the fact that it an crash an iPad.

I'm guessing it might be a good idea to prevent tiles from loading if the tilelayer's `_tileZoom` differs from the map's zoom level by more than a threshold - maybe 1, maybe `map.options.zoomAnimationThreshold`. This would replace the `minZoom` check in `_update`.

IvanSanchez:
A threshold of more than 1 zoom levels to cancel a `GridLayer._update` does the trick, so I've amended the commit.

IvanSanchez:
I think this fixes #3515 as well :-D

src/layer/tile/TileLayer.js
```diff
@@ -56,8 +56,8 @@ L.TileLayer = L.GridLayer.extend({
 	createTile: function (coords, done) {
 		var tile = document.createElement('img');
 
-		tile.onload = L.bind(this._tileOnLoad, this, done, tile);
-		tile.onerror = L.bind(this._tileOnError, this, done, tile);
+		L.DomEvent.on(tile, 'load', L.bind(this._tileOnLoad, this, done, tile));
+		L.DomEvent.on(tile, 'error', L.bind(this._tileOnError, this, done, tile));
```
IvanSanchez: Firefox has real issues with `.onload`. It just doesn't fire during a `flyTo` animation, but somehow `L.DomEvent.on('load',...)` works flawlessly.

src/layer/tile/GridLayer.js
```diff
@@ -356,13 +361,17 @@ L.GridLayer = L.Layer.extend({
...
 				this._update(center, tileZoom);
 			}
 
 			if (!noPrune) {
 				this._pruneTiles();
 			}
+
+			// Flag to prevent _updateOpacity from pruning tiles during
+			// a zoom anim or a pinch gesture
+			this._noPrune = !!noPrune;
```
IvanSanchez: During a pinch gesture, if the user zooms in, the tile the event started on might be destroyed because `_onTileLoad` will call `_updateOpacity` which will call `_pruneTiles`, which checks parent/child tiles and prunes them if needed.

src/layer/tile/GridLayer.js
```diff
@@ -343,10 +343,15 @@ L.GridLayer = L.Layer.extend({
...
+		var tileZoom = Math.round(zoom);
+		if ((this.options.maxZoom !== undefined && tileZoom > this.options.maxZoom) ||
+		    (this.options.minZoom !== undefined && tileZoom < this.options.minZoom)) {
+			tileZoom = undefined;
+		}
+
+		var tileZoomChanged = (tileZoom !== this._tileZoom);
 
-		if (!noUpdate && tileZoomChanged) {
+		if (!noUpdate || tileZoomChanged) {
```
IvanSanchez: If animating, ignore the `noUpdate` flag and reload tiles anyway if we're switching the grid layer to another level. This way there's no need to throttle!

src/layer/tile/GridLayer.js
```diff
@@ -446,11 +451,13 @@ L.GridLayer = L.Layer.extend({
 			this._tiles[key].current = false;
 		}
 
+		if (Math.abs(zoom - this._tileZoom) > 1) { this._setView(center, zoom); }
```
yohanboniface: Maybe a comment here would help future maintainers (with mantra "say why not what" in mind, of course) ?

yohanboniface:
Is #3811 also taken into account?

IvanSanchez:
@yohanboniface No, and I've just noticed that I've somehow broken `_getTiledPixelBounds` for fractional zoom. :-(

PR #3822, opened 2015-09-09
add leaflet-icon-pulse

filipzava:


danzel:
image
This change has literal + characters in it ;)
I've removed them now.

PR #3828, opened 2015-09-10
Release for Beta .1 not .2 exists

wurde:
So I check the package.json for which release version to download and v1.0.0-beta.2.zip is not found but v1.0.0-beta.1.zip is. Is there any particular reason why the .2 was referenced here and no corresponding release exists?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "leaflet",
-  "version": "1.0.0-beta.2",
+  "version": "1.0.0-beta.1",
   "description": "JavaScript library for mobile-friendly interactive maps",
   "devDependencies": {
     "copyfiles": "0.2.1",


yohanboniface:
> Is there any particular reason why the .2 was referenced here and no corresponding release exists?

Yes, because this is the Leaflet dev version, so it describes the next release.

wurde:
So let me explain my use case. I create a rails template which will grab the current target version of a piece of vendor software such as Bootstrap. In that case I can look at the twbs/bootstrap package.json and then point to an existing https://github.com/twbs/bootstrap/releases to download.

yohanboniface:
Uses releases, not dev versions, either via npm, Github tags, bower…

wurde:
Ok, developers choice I guess.

PR #3830, opened 2015-09-11
Prevent to use extreme values in transform3d

### LINKED ISSUE #3608: Firefox bug when using setView with animate and zoom > 19
JSFiddle to reproduce it (only Firefox): http://jsfiddle.net/heu9hsrL/4/
Removing `animate: true` makes it work.
Changing zoom in `setView` from 19 to 18 makes it work.
Works on Chrome.
According to the network panel, tiles are loaded, and inspecting the img elements I can't see any obvious reason they are not shown. Seems they are out of map view.
Removing the `transform3d` CSS property from both the img element and the `leaflet-map-pane` element make the img appear in view.
I've not seen any difference with Chrome values, while comparing the situation with FF.
I had a quick look at the code some days ago, but I didn't find anything yet. I may have another look in the incoming days, but please anyone feel free to check it anyway :)

yohanboniface:
Prevent to use extreme values in transform3d:

- FF bugs in positioning the element
- Chrom(e|ium) have a one pixel offset

Fix #3608 

Just took the suggestion from https://github.com/Leaflet/Leaflet/issues/3608#issuecomment-121403977
Not really proud of this patch actually, I'm sure there is something lighter to do, without asking for full resetView, but cannot put my brain on it :/

«Lost in translate3d» would be the name of my yesterday movie! ;)

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -16,7 +16,8 @@ L.Map = L.Evented.extend({
 		fadeAnimation: true,
 		trackResize: true,
 		markerZoomAnimation: true,
-		maxBoundsViscosity: 0.0
+		maxBoundsViscosity: 0.0,
+		transform3DLimit: 9999990
 	},
 
 	initialize: function (id, options) { // (HTMLElement or String, Object)
@@ -611,6 +612,8 @@ L.Map = L.Evented.extend({
 		if (this.options.trackResize) {
 			L.DomEvent[onOff](window, 'resize', this._onResize, this);
 		}
+
+		this[onOff]('moveend', this._onMoveEnd);
 	},
 
 	_onResize: function () {
@@ -622,6 +625,16 @@ L.Map = L.Evented.extend({
 	_onScroll: function () {
 		this._container.scrollTop  = 0;
 		this._container.scrollLeft = 0;
+	},
+
+	_onMoveEnd: function () {
+		if (L.Browser.any3d && this.options.transform3DLimit &&
+			(Math.abs(this._mapPane._leaflet_pos.x) >= this.options.transform3DLimit
+			 || Math.abs(this._mapPane._leaflet_pos.y) >= this.options.transform3DLimit)) {
+			// https://bugzilla.mozilla.org/show_bug.cgi?id=1203873 but Webkit also have
+			// a pixel offset on very high values, see: http://jsfiddle.net/dg6r5hhb/
+			this._resetView(this.getCenter(), this.getZoom());
+		}
 	},
 
 	_findEventTargets: function (src, type, bubble) {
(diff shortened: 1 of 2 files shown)


yohanboniface:
Upstream bug opened by @IvanSanchez https://bugzilla.mozilla.org/show_bug.cgi?id=1203873

src/map/Map.js
```diff
@@ -611,6 +612,8 @@ L.Map = L.Evented.extend({
 		if (this.options.trackResize) {
 			L.DomEvent[onOff](window, 'resize', this._onResize, this);
 		}
+
+		this[onOff]('moveend', this._onMoveEnd);
```
mourner: why not move the if from `_onMoveEnd` here?

yohanboniface: humm, but the `_leaflet_pos` value changes at every call, no?
And my first implementation was keeping the `if (L.Browser.any3d && this.options.translate3dLimit)` here, but then I thought that this was a trap for people wanting to override `L.Map` to have `_onMoveEnd` sometimes registered sometimes not. One can argue that `_xxx` means private method and we don't care about people overriding but…. ;)

Thoughts?

mourner: I don't know who would want to override this method — it's easier to just set false in options and then put your own listener for custom behavior. Not sure what you mean by `_leaflet_pos` changes.

src/map/Map.js
```diff
@@ -624,6 +627,16 @@ L.Map = L.Evented.extend({
 		this._container.scrollLeft = 0;
 	},
 
+	_onMoveEnd: function () {
+		if (L.Browser.any3d && this.options.transform3DLimit &&
+			(Math.abs(this._mapPane._leaflet_pos.x) >= this.options.transform3DLimit
+			 || Math.abs(this._mapPane._leaflet_pos.y) >= this.options.transform3DLimit)) {
```
mourner: There's a private method `_getMapPanePos`, let's use it. Also, you could make this shorter if you use `Math.max`.

src/map/Map.js
yohanboniface: > Not sure what you mean by _leaflet_pos changes.

`initEvents` is called only once at map init, right? But `map._mapPane._leaflet_pos` value is changed at every `_resetView` call, so we cannot rely on its value at init time. Correct?

src/map/Map.js
IvanSanchez: I'm thinking about making a `L.Bounds` with the max values for 3D transforms and just use something like `max3dTransforms.contains(this._getMapPanePos())`.

src/map/Map.js
mourner: Oh, no, I mean only move the condition, not the filter itself. Don't put a moveEnd listener if there's no point.

yohanboniface: So like this https://github.com/Leaflet/Leaflet/pull/3831/files#diff-607b516b5121104eef3007cb5727e69cR616 ? :)

mourner: Yes, exactly!

PR #3833, opened 2015-09-11
upgrade uglify-js

fscherwi:
See: https://nodesecurity.io/advisories/uglifyjs_incorrectly_handles_non-boolean_comparisons

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "magic-string": "^0.7.0",
     "mocha": "~2.3.0",
     "tin": "^0.5.0",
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Apparently this only affects https://github.com/Leaflet/Leaflet/blob/master/src/dom/Draggable.js#L51

I don't think we have any bugs related to dragging behaviour, so might as well let NPM run its course and decide the version at build time.

mourner:
Yeah, this won't affect anyone since NPM will install the latest version anyway.

PR #3837, opened 2015-09-12
Always call renderer.getEvents before Path.getEvents (fix #3836)

### LINKED ISSUE #3836: Paths off screen are not updated when moved into screen
polygon-bug

According to `git bisect`, the guilty commit is 61c07bd3e3d9a5a2bb4855ba81e1f368cc80090b

And indeed when I revert it it works back as expected. Swimming pool time, but I'll try to have a shot at investigating today. :)

yohanboniface:
Fix #3836 

61c07bd3e3d9a5a2bb4855ba81e1f368cc80090b inverted order in which Path._update and Rendered._update were called.

I think that we want to keep 61c07bd3e3d9a5a2bb4855ba81e1f368cc80090b changes, so here is a possible fix.

### DIFF
--- a/src/layer/vector/Path.js
+++ b/src/layer/vector/Path.js
@@ -23,8 +23,13 @@ L.Path = L.Layer.extend({
 		interactive: true
 	},
 
+	beforeAdd: function (map) {
+		// Renderer is set here because we need to call renderer.getEvents
+		// before this.getEvents.
+		this._renderer = map.getRenderer(this);
+	},
+
 	onAdd: function () {
-		this._renderer = this._map.getRenderer(this);
 		this._renderer._initPath(this);
 		this._reset();
 		this._renderer._addPath(this);


yohanboniface:
Erf, that can't work as it is: relation path<=>renderer is not one to one.

Issue here is that Path.getEvents is called before Renderer.getEvents since 61c07bd. The current pattern is that the renderer is instantiated and added to the map only *during* the `onAdd` call for the first Path added to the map. But this Path instance, being in the `_layerAdd` process, yet has called its `getEvents` method, and thus its event listeners will come before the renderer ones.

Again, I think we want to keep 61c07bd, and thus we need to find a way to call Renderer.getEvents before Path.getEvents. Maybe just instantiating the renderer at map init?

yohanboniface:
Now setting renderer in `beforeAdd` method. Should be good. :)

mourner:
Included in #3840 so I'll merge that one instead.

mourner:
Actually #3840 is still in the works so I'll merge this one.

PR #3839, opened 2015-09-14
Redo some (ms)pointer work to be consistent and fix IE10

### LINKED ISSUE #3804: Text in popups not selectable in Leaflet 0.7.4 and .5 in IE10
I'm sorry, I haven't got a reproducible case yet, but I'll add it soon.

The issue occurred when I upgraded an internal app from Leaflet 0.7.3 to 0.7.4, where selecting text wouldn't work. I tried upgrading again to 0.7.5, but that failed too. Downgrading back to 0.7.3 fixed the issue.

danzel:
Currently we preventDefault all of the pointer events, which means they don't create ```mousedown```/```mouseup``` events.
This stops mouse events on things on top of the map like popups. It also breaks leaflet.draw in IE10.

I've removed the maxTouchPoints detection as it just makes testing pointer events harder.

This is worth backporting as we broke IE10 a bit in the 0.7.4/5 updates.
Fixes #3804

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -17,7 +17,7 @@
 
 	    mobile = typeof orientation !== 'undefined' || ua.indexOf('mobile') !== -1,
 	    msPointer = !window.PointerEvent && window.MSPointerEvent,
-	    pointer = (window.PointerEvent && navigator.pointerEnabled && navigator.maxTouchPoints) || msPointer,
+	    pointer = (window.PointerEvent && navigator.pointerEnabled) || msPointer,
 
 	    ie3d = ie && ('transition' in doc.style),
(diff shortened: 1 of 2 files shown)


mourner:
Thanks a lot Dave! Looks good to me, leaving it to @IvanSanchez to do final verification.
Could you submit a stable backport PR as well?

danzel:
Stable one sent in #3842

IvanSanchez:
I don't have my windows VMs at hand, so testing will have to wait a week.

mourner:
Let's merge and then wait a week, it's fun :)

danzel:
We're running this on our live system, pretty certain this hasn't caused any issues... none I've heard of yet anyway :)

PR #3840, opened 2015-09-14
Comprehensive bounds on canvas layer update (fix #3583)

### LINKED ISSUE #3583: Layer setStyle masks/dissappears top layers sporadically on canvas drawing.
Circles drawn on top of state layers disappear sporadically when setStyle is applied on mouse over events on state layers. Happens on canvas drawing. 

using current build of leaflet. 
http://codepen.io/anon/pen/VLQmZy

yohanboniface:
Fix #3583 

The issue is that when updating a layer, we update all others layers that intersect with this first layer bounds, but leaving the edge case of other layers that may intersect with those updated layers.

For example, in this situation:

When mouse over the Maryland, only the Maryland center marker and the West Virginia polygon intersect with its polygon and thus are updated. See the Maryland bounds:

Though I'm not big fan of the proposed patch: it computes recursively the comprehensive bounds of all impacted layers. So I'm not sure how this would scale, and if the optimization of not updating all layers still worth it at this cost.

Thoughts?

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -9,7 +9,8 @@ L.Canvas = L.Renderer.extend({
 
 		this._layers = this._layers || {};
 
-		// redraw vectors since canvas is cleared upon removal
+		// Redraw vectors since canvas is cleared upon removal,
+		// in case of removing the renderer itself from the map.
 		this._draw();
 	},
 
@@ -96,11 +97,17 @@ L.Canvas = L.Renderer.extend({
 
 	_draw: function (clear) {
 		this._clear = clear;
-		var layer;
+		var layer, bounds = this._redrawBounds;
+		this._ctx.save();
+		if (bounds) {
+			this._ctx.beginPath();
+			this._ctx.rect(bounds.min.x, bounds.min.y, bounds.max.x - bounds.min.x, bounds.max.y - bounds.min.y);
+			this._ctx.clip();
+		}
 
 		for (var id in this._layers) {
 			layer = this._layers[id];
-			if (!this._redrawBounds || layer._pxBounds.intersects(this._redrawBounds)) {
+			if (!bounds || layer._pxBounds.intersects(bounds)) {
 				layer._updatePath();
 			}
 			if (clear && layer._removed) {
@@ -108,6 +115,7 @@ L.Canvas = L.Renderer.extend({
 				delete this._layers[id];
 			}
 		}
+		this._ctx.restore();  // Restore state before clipping.
 	},
 
 	_updatePoly: function (layer, closed) {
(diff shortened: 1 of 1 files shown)


yohanboniface:
Depends on #3837

mourner:
Looks great to me! I'll verify with the debug page and merge.

src/layer/vector/Canvas.js
```diff
@@ -8,9 +8,6 @@ L.Canvas = L.Renderer.extend({
 		L.Renderer.prototype.onAdd.call(this);
 
 		this._layers = this._layers || {};
-
-		// redraw vectors since canvas is cleared upon removal
-		this._draw();
```
mourner: is this not necessary anymore?

yohanboniface: Humm, now that you ask me, I see there is a scenario where it's still needed: when removing and readding the rendered itself from the map.
I've removed this line because when adding a Path, it calls itself `renderer._updatePath`. And, when only dealing with adding or removing layers (not the renderer itself) from the map,  `renderer._layers` is empty at `Rendered.onAdd` time, and thus the call to `_draw` is flailing at the air.
Anyway, I'm reverting this change! :)

src/layer/vector/Canvas.js
```diff
@@ -96,11 +97,17 @@ L.Canvas = L.Renderer.extend({
 
 	_draw: function (clear) {
 		this._clear = clear;
-		var layer;
+		var layer, bounds = this._redrawBounds;
+		this._ctx.save();
+		if (bounds) {
+			this._ctx.beginPath();
+			this._ctx.rect(bounds.min.x, bounds.min.y, bounds.max.x - bounds.min.x, bounds.max.y - bounds.min.y);
```
IvanSanchez: The bounds will create artifacts when changing the style of features (e.g. line width on hover). Somehow some padding must be added to these bounds.

mourner:
```
this._ctx.rect(bounds.min.x - 10, bounds.min.y - 10, bounds.max.x - bounds.min.x + 20, bounds.max.y - bounds.min.y + 20);
```

This gets rid of artifacts in geojson.html, but obviously we don't want to have it hardcoded. Ideally we'd use half of stroke width + 1 for padding (or something like that).

yohanboniface:
Not sure about which artifacts you are referring to in geojson.html. If you are talking about the white states border being drawn above the grey border on mouseover, this bug is also reproducible in stable when in canvas: http://playground-leaflet.rhcloud.com/top/edit?html,output

screenshot from 2015-09-20 11 42 15

mourner:
I'm referring to a different bug. If you look at canvas-powered geojson example and move your mouse a bit, you'll see leftovers of strokes like this:

image

Issue #3829, opened 2015-09-10
Removed Path with mouseout throws error


timwis:
I've put together a test case that demonstrates what I'm attempting to do and the error that it causes: http://codepen.io/anon/pen/RWPyej?editors=001

To demonstrate, open the javascript console in Chrome then click on one of the polygons. It will remove the geojson layer entirely (desired) but when you move your mouse outside of where the polygon you clicked used to be, you'll get a console error: `Uncaught TypeError: Cannot read property 'mouseEventToContainerPoint' of null`

```javascript
    layer = L.geoJson(geojson, {
        onEachFeature: function(feature, layer) {
          layer.on({
            click: onClick
          });
        }
      }).addTo(map)

    function onClick(e) {
      map.removeLayer(layer);
    }
```

When a user clicks on a polygon, I'd like to replace the geojson layer with other geojson data. I can either do this by removing the layer from the map or clearing the layer and adding new data; it seems either route produces the same error.

I assume this has to do with the listeners for each polygon remaining after I've removed the layer, so I've tried to use `clearAllEventListeners()` to no avail.

@yohanboniface pointed out that it only seems to be occurring in Chrome and not Firefox or Safari. He also suggested I use leaflet.js from the master branch -- doing so stops the error and replaces it with a warning, `'webkitMovementX' is deprecated. Please use 'movementX' instead.`

Thanks for your help!

IvanSanchez:
I cannot reproduce with Chromium 44.0.2403.107-1 (no error messages at all after clicking a polygon then moving the mouse around).

Also, please note you're using the `master` branch javascript but the 0.7.5 CSS files in your codepen, which is causing graphical glitches when zooming.

timwis:
My apologies - I didn't realize swapping out the .js file to test master would save it on the codepen. I've switched it back to the stable 0.7.5 .js file. Can you give it another try? (Remember to move your mouse away from the area after clicking)

IvanSanchez:
Still cannot reproduce. :-/ Tried zomming and out and moving away out of the area or out of the map container, and still no errors in my Chromium console.

secularkid:
Hi. I just ran to the similar error today, and glady found this post. My workflow is not really the same but I'm able to reproduce error on @timwis codepen.

I've made a small capture from my screen, available there : http://www.filedropper.com/leafleterrorchrome (it's a .swf and I don't know how to host it directly here).

[...]

PR #3845, opened 2015-09-15
Issue #3829 - Removed Path with mouseout throws error

sambernet:
- add check to avoid triggering mouseout events on paths that have already been removed from the map (e.g. when path is removed in onclick event)


yohanboniface:
Please rebase without dist/ files.

sambernet:
Replaced with new Pull-Request containing only the fix:
https://github.com/Leaflet/Leaflet/pull/3849

(Sorry wasn't aware new commits get added to the PR automatically)

PR #3849, opened 2015-09-16
Issue #3829 - Removed Path with mouseout throws error

sambernet:
- add check to avoid triggering mouseout events on paths that have already been removed from the map (e.g. when path is removed in onclick event)


mourner:
Thanks!

PR #3853, opened 2015-09-18
Fix touch drag in Edge browser

Neorth:
Should make it possible to panorate the map on touch devices using the Edge browser. This was fixed on master in e44179da12dd3aaa7b1a0684ec193de61d4ec207

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -21,6 +21,7 @@
 .leaflet-container {
 	overflow: hidden;
 	-ms-touch-action: none;
+	touch-action: none;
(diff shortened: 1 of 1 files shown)


mourner:
Thanks!

r0b-:
This disables page scrolling on Chrome and Edge if I disable dragging & tap.
If I set the touch-action back to 'auto' everything seems to work fine... So is this setting still needed?

IvanSanchez:
@r0b- Do you experience that with 1.0.0-beta also?

PR #3859, opened 2015-09-21
Fix bug with maxNativeZoom == 0

olee:
Setting `maxNativeZoom` to `0` does not work as expected (only use zoom level 0) but instead disables the function.
This PR fixes this problem.
Same fix applies for (afaik current stable) 0.7.5


mourner:
Nice catch, thanks!

olee:
Can you apply the same patch to stable (0.7.5), too?
Or should I make a PR for this one as well?

PR #3860, opened 2015-09-22
Add new plugin - Leaflet.curve

elfalem:


mourner:
Cool, thanks!

jieter:
@elfalem note that your demo doesn't work for me.
You are loading het `leaflet.curve.js` file from raw.githubusercontent.com domain which sets MIME type `text/plain`:
2015-09-22-131002_705x653_scrot

You should add the script to the gh-pages branch and use a relative path.

elfalem:
@jieter, thanks - that should be fixed now

jieter:
Yes, works like charm now.

You can also ditch the master branch and use gh-pages as the main branch. This pattern is used by [some](https://github.com/mapbox/leaflet-image) [other](https://github.com/calvinmetcalf/leaflet-ajax) [plugins](https://github.com/jieter/Leaflet.layerscontrol-minimap), and makes sure your github pages is never out of sync.

elfalem:
Thanks for the tip! I will try that.

PR #3868, opened 2015-09-24
Added zoom for the custom scale in L.CRS.Simple

### LINKED ISSUE #3867: L.CRS.Simple zoom implementation
Hi,

I am trying to use L.CRS.Simple for a custom (image) map. However, on touch devices the map does not zoom properly, it flips out basically... I am using the current dev branch of Leaflet (3ff1b65). 

http://playground-leaflet.rhcloud.com/dig/edit?html,console,output
Try to zoom with touch device (so it uses TouchZoom handler). It seems that the issue is in the L.CRS.Simple, where only a new scale method has been defined: 

```
	scale: function (zoom) {
		return Math.pow(2, zoom);
	},
```

Where in the standard L.CRS:
```
	// defines how the world scales with zoom
	scale: function (zoom) {
		return 256 * Math.pow(2, zoom);
	},
	
	zoom: function (scale) {
		return Math.log(scale / 256) / Math.LN2;
	},
```

Adding the following to L.CRS.Simple should fix this little bug:
```
        zoom: function (scale) {
		    return Math.log(scale) / Math.LN2;
    	},  
```

cmulders:
Fixes #3867

### DIFF
--- a/src/geo/crs/CRS.Simple.js
+++ b/src/geo/crs/CRS.Simple.js
@@ -11,9 +11,9 @@ L.CRS.Simple = L.extend({}, L.CRS, {
 	},
 
 	zoom: function (scale) {
-            return Math.log(scale) / Math.LN2;
-        }, 
-	
+		return Math.log(scale) / Math.LN2;
+	}, 
+
 	distance: function (latlng1, latlng2) {
 		var dx = latlng2.lng - latlng1.lng,
 		    dy = latlng2.lat - latlng1.lat;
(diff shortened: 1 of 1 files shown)


src/geo/crs/CRS.Simple.js
```diff
@@ -10,6 +10,10 @@ L.CRS.Simple = L.extend({}, L.CRS, {
 		return Math.pow(2, zoom);
 	},
 
+	zoom: function (scale) {
+            return Math.log(scale) / Math.LN2;
+        },
```
mourner: please fix the whitespace (`}` doesn't line up and we use tabs rather than spaces)

mourner:
Thanks! Committed as https://github.com/Leaflet/Leaflet/commit/bcde047f0991b922ea845d3786d9a887f0b02f7e

PR #3869, opened 2015-09-25
Fixed FeatureGroup crash

Lambdac0re:
Now a LayerGroup can be removed from a FeatureGroup without crashing. FeatureGroup's addLayer had the check in place, but removeLayer did not!


Lambdac0re:
Dirty workaround until this is merged is to let LayerGroup have a fake 'off' fucntion:
    L.LayerGroup.prototype.off = L.Util.falseFn;
That way calls to a FeatureGroup's removeLayer and clearLayers don't end up with:
    TypeError: layer.off is not a function

mourner:
Thanks, nice catch!

PR #3872, opened 2015-09-28
Simulate mousemove on touchmove (fix #3825)

### LINKED ISSUE #3825: Can't re-enable dragging on mobile browsers
Using leaflet master, I'm unable to re-enable map dragging after it's disabled.

Simple example: https://jsfiddle.net/skalb/otwgyutn/7/
Complex example: https://jsfiddle.net/skalb/otwgyutn/5/

Works:
- Desktop browsers

Doesn't work:
- Chrome/Safari on iPhone 6
- Safari on iPhone 6 simulator

yohanboniface:
Fix #3825

Well, sounds like a bit audacious (like hey, I have a joker in my sleeve), but it does fix the issue and tests passe ;)

Thoughts?

### DIFF
--- a/src/map/handler/Map.Tap.js
+++ b/src/map/handler/Map.Tap.js
@@ -90,6 +90,7 @@ L.Map.Tap = L.Handler.extend({
 	_onMove: function (e) {
 		var first = e.touches[0];
 		this._newPos = new L.Point(first.clientX, first.clientY);
+		this._simulateEvent('mousemove', first);
 	},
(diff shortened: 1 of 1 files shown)


mourner:
That fix looks really weird, I should try to figure out what's going on here...

mourner:
You just blew my mind so I'll just merge and hope it won't break anything.

yohanboniface:
:p

PR #3882, opened 2015-10-01
Prevent Popup._adjustPan while already panning (fix #3744)

### LINKED ISSUE #3744: openPopup() breaks setView() with animation
Scenario is the following very common setup: clicking in sidebar to center map on marker and open marker's popup. What happens instead: map pans to wrong place and/or moves only till marker is in the map, but not in the center. It also happened that the popup got displayed in a totally different place than the marker on the screen.

Abstract code for trigger:
```
item_title_link.click(function() {
    map.panTo(layer.getLatLng(), {animate: true});
    layer.openPopup();
});
```

What happens: 
1. [bug]: When clicking for the **first time**, the map doesn't center on the marker, but pans until it is in the corner/edge of map. In this case, `panBy` is called twice.
2. [good]: When clicking for the **second time**, the map centers on the marker. In this case, `panBy` is only called once.

The reason as far I was able to debug is the following. The following paths are all fighting to move the screen and end up in broken results many times.
1. `panTo` -> `setView` -> `_tryAnimatedPan` -> `panBy` is always called. This is good and it is the required behaviour.
2. `Layer.openPopup` -> `Map.openPopup` -> `setLatLng` -> `update` ->  `_adjustPan` -> `panBy`. This should not happen, as it destroys the real panBy movement. This path is not happening once the marker is visible on the screen (at the edges/corners), thus the second time the activation happens this path doesn't run.

The only "fix" I found is to call setView with `{reset: true}` to completely disable animation.

Leaflet v1.0.0-beta.1

OS X 10.9.5, latest Chrome

yohanboniface:
Fix #3744 

Bisect point at a33eff73f0d46b326d6fe59a7072f31624527c23 

Basically, when calling openPopup *while* the map is animating, the `Popup._adjustPan` maths are wrong, and I think we just don't want to do them in this situation.

Thoughts?

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -248,7 +248,7 @@ L.Popup = L.Layer.extend({
 	},
 
 	_adjustPan: function () {
-		if (!this.options.autoPan) { return; }
+		if (!this.options.autoPan || (this._map._panAnim && this._map._panAnim._inProgress)) { return; }
 
 		var map = this._map,
 		    containerHeight = this._container.offsetHeight,


src/map/anim/Map.PanAnimation.js
```diff
@@ -40,6 +40,10 @@ L.Map.include({
 	panBy: function (offset, options) {
 		offset = L.point(offset).round();
 		options = options || {};
+		this._animatingPan = true;
+		this.once('moveend', function () {
```
yohanboniface: Not sure about this way of resetting the flag actually, it may ends with race condition in some edge case. Should I just replace with resetting before everything `return this.fire.('moveend')` instead?

mourner: I think we can track animating state easier with `_panAnim._inProgress` or something like this.

yohanboniface: Done :)

PR #3886, opened 2015-10-02
Replace links to google groups with ones to GIS StackExchange

IvanSanchez:
I hereby propose getting rid of the "user forum" and using gis.stackexchange.com instead. The volume and quality on the current "forum" is quite low, and there is quite a high number of people using stackexchange nowadays, anyway.

This would need a bit of follow-up: posting a notice in the google group, and pushing the changes to CONTRIBUTING.md to the master branch.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ Third-party patches are absolutely essential on our quest to create the best map
 However, they're not the only way to get involved with the development of Leaflet.
 You can help the project tremendously by discovering and [reporting bugs](#reporting-bugs),
 [improving documentation](#improving-documentation),
-helping others on the [Leaflet forum](https://groups.google.com/forum/#!forum/leaflet-js)
+helping others on [GIS StackExchange](https://gis.stackexchange.com/questions/tagged/leaflet)
 and [GitHub issues](https://github.com/Leaflet/Leaflet/issues),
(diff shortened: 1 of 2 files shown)


ghybs:
Hi,

I agree with Ivan about the forum.
Many beginners post there but receive little feedback.

On the other hand, the stackexchange platform offers more reactivity, alongside with much more features (comments, links to other / related threads, not even talking about the reputation system), which makes it more appropriate for problem solving, whereas a forum would be more oriented to open discussions.

mourner:
Another question is, should we link to GIS Stackexchange, or Stackoverflow? The latter has 3x as many question tagged `leaflet`.

ghybs:
Good (and tricky) point!

I think the situation is similar for Open Layers (3.3k in GIS StackExchange for v2, 700 for v3, vs. 7k in StackOverflow with "open layers") and somehow for Ask Ubuntu (200k vs. 150k in SO with "ubuntu").
The former chose to point to SO (sorry if it is forbidden talking about "competitors"…).
Of course, Ubuntu links to its dedicated SE site.

Community fragmentation is caused by the StackExchange platform architecture. It also fails at categorizing questions that involve mixed libraries, for which case people usually fall back to SO…

No solution is ideal, but IMHO it would be rude to omit a part of a community. Leaflet can be used for more than GIS, but on the other hand SO is a big mess.

Maybe we could recommend one (SO ?), while still mentioning the other (GIS SE)?

mourner:
Done in #4055

Issue #3832, opened 2015-09-11
Zooming (Gesture or Buttons) on LG G4 makes map unresponsive


pakro:
Hello, 

we experience a problem with the leaflet version 1.0 beta on a LG G4 stock browser (com.android.browser (38.0.2125.102)). After the map is loaded every zooming interaction leads to a unresponsive map and the tiles are not loaded anymore. However just panning the map around works fine. There are no errors on the console. On the stable release (0.7.5) it works just fine. Any clue what might cause this issue on this specific browser? 

Reproduction is simple. Just use the appropriate leaflet version, let the map take up all available space and add zooming controls. We used OSM as tile provider, but it also didn't work for Here.

IvanSanchez:
Can you please make sure this is not a duplicate of #3814?

pakro:
As far as I'm understanding the pictures of the issue #3814 right, it also happens in Chrome on Android and Safari on iOS. This issue here does only appear in the native LG G4 Browser (which should be Chrome 38 but maybe has some modifications from LG). It works flawlessly on the current available Chrome Browser for Android. However the symptoms sound similar, although this issue here is very simple to reproduce (it does exactly take one gesture or one zoom button press). To be honest I'm not absolutely sure whether this is a duplicate or not. But we had never observed such issues on Chrome and Safari. The G4 and its native browser is the first device.

I will try to investigate the issue and will post it here, if I will find something.

IvanSanchez:
I've got no trouble with the default stock Android browsers in a Samsung S3 and in an HTC One. I have no way to try with a LG stock browser, so I cannot know what's wrong here.

pakro:
We also do not experience any issues with the S3 stock browser and the HTC One stock browser. But the LG does not work. Maybe someone else can confirm that issue. I can provide a minimal code example, which works every time.

IvanSanchez:
> I can provide a minimal code example, which works every time.

That's always useful! :-)

But your description of the bug makes me think that https://github.com/Leaflet/Leaflet/blob/master/debug/map/map-mobile.html might work for reproducing this one.

pakro:
Yes, I can confirm that your link does also produces the misbehavior.

pakro:
I'm not exactly sure, but I have a possible solution... at least for the LG G4. But I am absolutely not sure what else is affected. 

[...]

PR #3892, opened 2015-10-05
Fix g4 issues

pakro:
That's my solution for the stock browser issue on the G4. By listening to all possible "transitionend" types you never miss one, and since only the one that is appropriate for the device/browser is fired, it's just the initial overhead in binding the events. But I'm honest. At this point I did only basic testing. The other part of the fix removes the proxy element that is created for listening on the "transitionend" events. The proxy element somehow doesn't get the "transitionend" event. However the mapPane does. I didn't change too much at this place. Hope that helps you somehow. These changes seem to work quite well in our project.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -182,11 +182,12 @@ L.DomUtil = {
 	// webkitTransition comes first because some browser versions that drop vendor prefix don't do
 	// the same for the transitionend event, in particular the Android 4.1 stock browser
 
-	var transition = L.DomUtil.TRANSITION = L.DomUtil.testProp(
+	L.DomUtil.TRANSITION = L.DomUtil.testProp(
 			['webkitTransition', 'transition', 'OTransition', 'MozTransition', 'msTransition']);
 
-	L.DomUtil.TRANSITION_END =
-			transition === 'webkitTransition' || transition === 'OTransition' ? transition + 'End' : 'transitionend';
+	// trying to decide which type is appropriate is not as robust as adding all available types
+	// although this adds overhead (e.g LG G4 stock browser uses webkitTransition but fires transitionend)
+	L.DomUtil.TRANSITION_END = 'webkitTransitionEnd OTransitionEnd transitionend';
 
 
 	if ('onselectstart' in document) {
--- a/src/map/anim/Map.ZoomAnimation.js
+++ b/src/map/anim/Map.ZoomAnimation.js
@@ -21,7 +21,7 @@ if (zoomAnimated) {
 
 			this._createAnimProxy();
 
-			L.DomEvent.on(this._proxy, L.DomUtil.TRANSITION_END, this._catchTransitionEnd, this);
+			L.DomEvent.on(this._panes.mapPane, L.DomUtil.TRANSITION_END, this._catchTransitionEnd, this);
 		}
 	});
 }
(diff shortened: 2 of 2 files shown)


IvanSanchez:
@pakro When you make a pull request, please mention the original bug (#3832) so magical links can happen :-)

src/dom/DomUtil.js
```diff
@@ -182,11 +182,12 @@ L.DomUtil = {
...
 	// the same for the transitionend event, in particular the Android 4.1 stock browser
 
-	var transition = L.DomUtil.TRANSITION = L.DomUtil.testProp(
+	L.DomUtil.TRANSITION = L.DomUtil.testProp(
 			['webkitTransition', 'transition', 'OTransition', 'MozTransition', 'msTransition']);
 
-	L.DomUtil.TRANSITION_END =
-			transition === 'webkitTransition' || transition === 'OTransition' ? transition + 'End' : 'transitionend';
+	// trying to decide which type is appropriate is not as robust as adding all available types
+	// although this adds overhead (e.g LG G4 stock browser uses webkitTransition but fires transitionend)
```
IvanSanchez: What. The. Hell.
I wonder if it's worth doing a `L.Browser` hack to detect this particular case, or we can live with the extra bit of overhead.

IvanSanchez:
@pakro Nice work noticing the browser weirdness! I know debugging these things can be hard.

Regarding the `transitionend` events... some browsers might fire one `transitionend` on the map **and** on the proxy, or just a `transitionend` on **either** the map or the proxy, or (worst case) none. So I'm undecided about that.

pakro:
@IvanSanchez Sorry for not linking the bug. This is actually my first pull request. Hope you can excuse this mistake. Won't happen again. :) Regarding the detection of the appropriate event. I'm not sure whether it is possible to detect the LG G4 stock browser with the user agent or not. It would be also interesting whether other LG devices (e.g G2 G3) are also affected and share the same user agent or not. At our office we only have the G4. 

Concerning the map and proxy topic. Here I honestly just made it work and tested whether it affects other devices and browser lying around our office or not. It seemed to work well for this particular selection in our office. However if you say that they are cases where it may fire either on the map or the proxy, the fix cripples down the devices/browsers where the proxy is the receiver and not the map, as I removed it. So this may be not the best solution. I'm not mad on you if you are undecided on this. I also found it very weird and didn't understand the problem in its full complexity.

I'll try finding some time to investigate the map and proxy issue.

Eschon:
> It would be also interesting whether other LG devices (e.g G2 G3) are also affected

I have a G3 right here if you could provide a way to reproduce the bug this should fix I can test it.

pakro:
@Eschon It's actually quite simple, check out the the current master load the debug/map/map-mobile.html on your stock browser and zoom the map with either the zoom controls or just a pinch to zoom gesture.  If the map stops reacting after that action you have the bug.

Eschon:
Just tested it and seems to work without problems on my G3.

pakro:
Can you please check whether your stock browser version matches that one of the LG G4. Under Settings/Apps/All/Internet what version is displayed? The LG G4 has the version 5.2.38F27. If the versions are the same it would be very weird, but if they are different that makes sense.

Eschon:
@pakro My version is 5.2.34B13

pakro:
OK the version is different and therefore it is very likely that the used engine has also a different version. That may a justification for a L.Browser hack, as this problem is LG G4 Android 5.1 exclusive. I assume the G2 does also use a different stock browser version.

IvanSanchez:
This has merge conflicts, so I'll close it down for now.

I'm not very optimistic about making changes in the `animationend` event detection, because browsers must handle some (documented) strange behaviour. In other words, things can get messy.

See:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1236979
* https://bugzilla.mozilla.org/show_bug.cgi?id=1236930
* https://github.com/whatwg/compat/issues/24

Updated web browsers *should* handle the compatibility issue in the `whatwg` compat list. Please reopen the bug if it's still reproducible with up-to-date browsers.

PR #3894, opened 2015-10-06
Hack around missing 'ontransitionend' event on zoom anim

### LINKED ISSUE #2693: Map freeze problem, but unlocks after marker click

IvanSanchez:
Should fix #2693

### DIFF
--- a/src/map/anim/Map.ZoomAnimation.js
+++ b/src/map/anim/Map.ZoomAnimation.js
@@ -104,6 +104,14 @@ L.Map.include(!zoomAnimated ? {} : {
 			zoom: zoom,
 			noUpdate: noUpdate
 		});
+
+
+		// This shouldn't be needed if browsers fired ontransitionend consistently
+		setTimeout(L.bind(function () {
+			if (this._animatingZoom) {
+				this._onZoomTransitionEnd();
+			}
+		}, this), 630);
 	},
 
 	_onZoomTransitionEnd: function () {
--- a/debug/hacks/ontransitionend.html
+++ b/debug/hacks/ontransitionend.html
@@ -1,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width,initial-scale=1 maximum-scale=1.0 user-scalable=0">
+	<link rel="stylesheet" href="../css/mobile.css" />
+
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+</head>
+<body>
+
+	<div id="map"></div>
+	<div >
+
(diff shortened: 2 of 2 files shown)


src/map/anim/Map.ZoomAnimation.js
```diff
@@ -104,6 +104,14 @@ L.Map.include(!zoomAnimated ? {} : {
...
 			noUpdate: noUpdate
 		});
+
+
+		// This shouldn't be needed if browsers fired ontransitionend consistently
+		setTimeout(L.bind(function(){
+			if (this._animatingZoom) {
+				this._onZoomTransitionEnd();
+			}
+		}, this), 630);
```
yohanboniface: Where is 630 from? :)

src/map/anim/Map.ZoomAnimation.js
```diff
@@ -104,6 +104,14 @@ L.Map.include(!zoomAnimated ? {} : {
 			zoom: zoom,
 			noUpdate: noUpdate
 		});
+
+
+		// This shouldn't be needed if browsers fired ontransitionend consistently
+		setTimeout(L.bind(function(){
```
yohanboniface: eslint does not complain about the missing space around `()`?

jieter: It does in travis:
2015-10-06-111923_699x60_scrot

IvanSanchez:
@yohanboniface The decision to use `630` msec comes from checking times with `performance.now()` between the `touchend` and `transitionend` events - it was around 620-625.

yohanboniface:
Shouldn't we take a margin? I guess this time will be different for different CPUs or not?

IvanSanchez:
@yohanboniface I'm doing stuff with an iPad 2 (2011!) so it should be slow enough. Do feel free to experiment with `debug/hacks/ontransitionend.html`, check out the timings in the debug console, and see if you can come up with a different number.

mourner:
This looks very similar to the hack I did for 0.7.5, but firing `setTimeout` at the same time as launching the transition to clean up after transitionend. This seems less fishy than having a magic 630ms to fire after touchend. Maybe we should just port over the fixes from stable then?

IvanSanchez:
@mourner Sure, sounds like a good idea. Is there a PR or commit with the stable fix I can have a look at?

mourner:
@IvanSanchez https://github.com/Leaflet/Leaflet/commit/91477588faa44e2b266d86a5a44cb98490244d39, https://github.com/Leaflet/Leaflet/commit/19c295ea7cc27f6aaaabed7c14d4affc8113279f, https://github.com/Leaflet/Leaflet/commit/3e200bed22bdb1f59e54f37c93b4fce034fd7ee4

Issue #3700, opened 2015-07-31
Major disparity in L.Projection.SphericalMercator.project function (between 1.0b and 0.73)


vas85:
It seems as part of change 49af3fda5c7811d3a023f2737a364e2bf0d87c31 a bunch of refactoring was done to the L.Projection.SphericalMercator class. 

The project() and unproject() functions are returning different values before and after, causing weird issues in bounds checking and a few other things for me in master.

Have a look at this JSFiddle: http://jsfiddle.net/6Lbm9970/3/

Note: this is not the same as https://github.com/Leaflet/Leaflet/pull/2062. My before version has the fix in that PR.

Thanks!

mourner:
cc @perliedman

Ariki:
The Mercator projection formulae produce infinite ordinate at ±90°. For that reason it's always truncated at some latitude, and one shouldn't rely on values returned at higher latitudes. The old version had a limit at 85.0511287798° and truncated all that was beyond that. The new one has much larger limit but projected ordinates at high latitudes are accordingly high. I believe it's user's responsibility to check whether the data is inside the domain of a particular projection.

vas85:
Thanks @Ariki that makes perfect sense. Seeing how the latitude projection is calculated using Math.tan makes that even more evident.

I do think that silent truncation / inaccuracies at higher latitudes could be surfaced better (maybe a warning, console.log, or some such)? I'll let someone else weigh in, but I think we can close this.

mourner:
Yeah, I probably removed the truncation because I thought it should be handled by CRS, not projection itself.

perliedman:
Keeping the truncation in the projection seems like the right thing: looking at the definition of EPSG:3857, it's valid in -85.06 < lat < 85.06 (http://epsg.io/3857).

I can try to fix this later tonight, if it sounds like a good idea and noone beats me to it.

perliedman:
The other option is of course to refuse projecting such latitudes, but IMO that would be against the spirit of Leaflet - if you try to put a marker on the north pole, most users would prefer a marker at the top of the map, rather than an error that would be cryptic to most, although the latter is technically more correct handling.

Ariki:
I think the old version with truncation at MAX_LATITUDE=arctan(sinh(π))=85.0511287798 was perfectly correct because it produced a square map consistent with popular tiling schemes. Some applications may need wider bounds but such cases are not what EPSG:3857 was intended for.

Ariki:
[...]

PR #3895, opened 2015-10-06
L.Renderer shall use center and not topleft as reference

### LINKED ISSUE #3769: Zoom anim breaks vector layers at low zooms
How to reproduce:

* Load up `/debug/vector/vector.html`
* Repeteadly zoom in and out between zoom levels 0-1 or 1-2.
* See how the vector doesn't line up with the markers.

leaflet-bug-1
leaflet-bug-2

Apparently vector layers are not positioned vertically in the right position during zoom anims, but only when the vector layer fits within the entire map container.

Affects both canvas and svg (see `/debug/vector/vector2.html`)

Using latest master. Reproducible on (at least) FFX and Chrome.

IvanSanchez:
Fixes #3769. Not the simplest of calculations, but they will have to do if we don't want to break #3700 again or do some crazy edge cases in the projections code.

### DIFF
--- a/src/layer/vector/Renderer.js
+++ b/src/layer/vector/Renderer.js
@@ -47,18 +47,22 @@ L.Renderer = L.Layer.extend({
 
 	_animateZoom: function (e) {
 		var scale = this._map.getZoomScale(e.zoom, this._zoom),
-		    offset = this._map._latLngToNewLayerPoint(this._topLeft, e.zoom, e.center);
+		    position = L.DomUtil.getPosition(this._container),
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This obsoletes `L.Map._latLngToNewLayerPoint`, which is not needed anymore. Is there any problem removing that as well?

mourner:
Yes, let's remove that. I was trying to come up with an easy animation math that could be reused by plugin authors for animating custom layers, but not a good progress so far (although much better than 0.7). Let's merge now and try to simplify later.

PR #3897, opened 2015-10-06
Zoom anim fires 'transitionend' event handler on a timer

IvanSanchez:
Like #3894, but porting @mourner's code from https://github.com/Leaflet/Leaflet/commit/91477588faa44e2b266d86a5a44cb98490244d39

### DIFF
--- a/src/map/anim/Map.ZoomAnimation.js
+++ b/src/map/anim/Map.ZoomAnimation.js
@@ -112,12 +112,15 @@ L.Map.include(!zoomAnimated ? {} : {
 	_onZoomTransitionEnd: function () {
 		if (!this._animatingZoom) { return; }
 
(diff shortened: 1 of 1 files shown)


mourner:
This caused a tile loading race condition on iOS that I had to hack-fix with https://github.com/Leaflet/Leaflet/commit/3e200bed22bdb1f59e54f37c93b4fce034fd7ee4. Should we port that here as well?

IvanSanchez:
@mourner Are you sure that iOS race condition is still present in `master`?

mourner:
Not sure, it's hard to reproduce. We can risk it and not port and then see if anyone complains, but that's not a perfect approach. :(

IvanSanchez:
More hackish than I would like, but OK.

PR #3899, opened 2015-10-06
On map stop(), also stop zoom animation (if any).

### LINKED ISSUE #3754: Calling setView during a zoom animation freezes the map.
Present in 1.0.0-beta.1...

Similar problem to what is described in #2021.

For now I can avoid it by disabling animations.  I'll look into it more when I have time and try to provide a jsfiddle.

colllin:
Fixes #3754 

Demo here:
http://playground-leaflet.rhcloud.com/lox/1/edit?html,console,output

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -236,7 +236,7 @@ L.Map = L.Evented.extend({
 			this._panAnim.stop();
 		}
 		if (this._animatingZoom) {
-			this._onZoomTransitionEnd();
+			this._stopZoomTransition();
 		}
 		return this;
 	},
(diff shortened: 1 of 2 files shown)


colllin:
Although this PR does fix the issue described in #3754 , there seems to be a collateral effect on zoom aesthetics and usability.  See this video:

http://cl.ly/3j0f0G1B3K1q

What else might be dependent on zoom transitions working a particular way?  I can look into smoothing those out.

mourner:
Unfortunately it's not that simple. I tried really hard implementing proper stop for CSS transitions-powered zoom animations, spent a few days, melted my brain and failed. I don't see a way to do this unless we decide to switch to manual frame-based zoom animations.

So you'll probably have to find a simpler fix for the issue.

PR #3908, opened 2015-10-08
bringToFront / bringToBack for popups.

### LINKED ISSUE #3813: Cannot bring overlapped popups opened with Map#addLayer on top
When you have multiple popups opened with Map#addLayer and some of them overlapped with others, it is not possible to bring the one behind others front.
So we have part of popup A on top of popup B. It would be good when you click on B, it comes froward and popup A goes behind it.
Any thought?

danzel:
Will need a docs update too, will send over in a moment.

Fixes #3813


yohanboniface:
And an entry in the CHANGELOG ;)

danzel:
Sorted :+1:

mourner:
@danzel this goes into beta2, not beta1 :)

yohanboniface:
Well, my fault, my comment was not clear about beta2 being in a branch still :s

danzel:
I'm bad at read :)

danzel:
ok, that should do it!

PR #3910, opened 2015-10-08
Fix a bunch of unbalanced tags

danzel:


danzel:
@mourner this will get out of date easy, maybe merge plz :)

PR #3916, opened 2015-10-09
🍂doc

IvanSanchez:
🍂doc coverage is complete. Running `jake docs` will fetch everything and build a static docs file (in sync with the code! Yay!) almost ready to be copied to `gh-pages`.

- [x] replace leaf with `@` symbol
- [x] add ability to write prop/method definition and description on the same line (maybe with `;` separator?)
- [x] add ability to write several commands in one line (e.g. `@class Foo @aka FooBar`)
- [x] Change the table subheaders' grey background into something else
- [ ] See if the accordion animation can be made smoother
- [ ] Come up with a way to deploy docs on new releases

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ tmp/**/*
 _site
 dist/*.js
 dist/*.map
+dist/reference.html
 coverage/
 *.js.html
 index.html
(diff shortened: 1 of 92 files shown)


mourner:
Very cool! First things I would like to see before I take a pass at the docs:

- [ ] replace leaf with `@` symbol
- [ ] add ability to write prop/method definition and description on the same line (maybe with `;` separator?)
- [ ] add ability to write several commands in one line (e.g. `@class Foo @aka FooBar`)

IvanSanchez:
[ ] See why the `GridLayer` unit tests are mysteriously failing

PR #3917, opened 2015-10-09
Canvas events fixes (fix #3915)

### LINKED ISSUE #3915: L.DomEvent.stop doesn't work with canvas
http://playground-leaflet.rhcloud.com/mak/edit?html,output

DomEvent.stop() doesn't prevent map click.

yohanboniface:
Fix #3915 

- fix fireEvent called for each layer containing event and then firing event on map once per layer (not sure we don't squarely want to only consider one of the layers, like we'd do for SVG layers)
- fix layer not removed from this._layer at remove
- fix L.DomEvent.stop(e) not honoured for canvas because events are both listenned on canvas container and map container


mourner:
Looks good to me!

mourner:
Looks good to me!

mourner:
Oh wow, I reviewed the PR twice and wrote the same comment. Time to sleep apparently.

yohanboniface:
:p

PR #3924, opened 2015-10-12
Fix duration not passed through from setView to panBy (fix #3300)

### LINKED ISSUE #3300: SetView 'duration' option does not work.
Looking in the SetView declaration the 'duration' option is not merged in the call to '_tryAnimatedPan'.

Suggest:

options.pan = L.extend({animate: options.animate, **duration: options.duration**}, options.pan);

yohanboniface:
Fix #3300


IvanSanchez:
Looks good to me.

Issue #3901, opened 2015-10-07
Streamline release workflow


mourner:
Currently making a release involves a lot of manual steps, introducing lots of friction, so we tag releases less often than we could. This is a big problem and needs to be addressed urgently, with the following steps:

- [x] Describe the current release workflow in detail (specific list of steps done on each release)
- [x] Investigate whether it's possible to grant @yohanboniface @IvanSanchez access to the S3 bucket Leaflet CDN is on, and do it
- [x] Look into how we can make the process simpler, preferably just a few steps with as much as possible automated

yohanboniface:
Would http://docs.travis-ci.com/user/deployment/ help?

mourner:
Probably. Travis is already set up to upload builds on each commit (see https://github.com/Leaflet/Leaflet/blob/master/.travis.yml). We should improve the set up to do the same for release tags, plus upload zip downloads too.

yohanboniface:
OK, so I let you do this task "Describe the current release workflow in detail (specific list of steps done on each release)" and I'll try to work on the `.travis.yml` to match it then :)

mourner:
It would probably be easier for me to do since I have access to Leaflet S3 and can test everything out.

jfirebaugh:
+1 for automating this so that Travis uploads to S3 any time it builds a tag. For mapbox-gl-js, the way this works (modulo that it's on Circle rather than Travis) is:

* AWS credentials in [encrypted environment variables](http://docs.travis-ci.com/user/environment-variables/#Encrypted-Variables)
* [deploy.sh script](https://github.com/mapbox/mapbox-gl-js/blob/master/deploy.sh)
* call the script if `$TRAVIS_TAG` is non-empty

mourner:
Now the current release workflow (since v1.0.0-beta.2) is:

- [x] update version in `src/Leaflet.js`
- [x] update version in package.json https://github.com/Leaflet/Leaflet/commit/55fe46250828d54d95cac0b8ca8b7a21dd97dbeb
- [x] run `npm run release` (provided you have `npm publish` rights)
- [x] wait for the Travis to build and upload artifacts (the build, leaflet.zip, etc.)
- [ ] Update `download.md` page in `gh-pages` branch to point to the new release (`http://cdn.leafletjs.com/leaflet/v1.0.0-beta.2/leaflet.js`, etc.)
- [ ] post a blog post and tweet about the new release

[...]

PR #3931, opened 2015-10-14
New Travis configuration with automatic releases

mourner:
Work in progress, just trying to force travis to build on force pushes.


danzel:
Awesome

mourner:
@danzel ended up reverting some of the stuff in the PR, but the workflow still got much simpler: https://github.com/Leaflet/Leaflet/issues/3901#issuecomment-148107538

PR #3937, opened 2015-10-15
Make non-interactive markers not fire pointer events

### LINKED ISSUE #2396: Set pointer-events: none to TileLayer and all non-iteractive layers

### LINKED ISSUE #3936: Non-interactive markers prevent event propagation to overlapping vectors
If a marker is on top of a vector layer (for example, a polygon or a circle), clicking on the marker will prevent any mouse events from being fired in said vector layer.

Explained better with a bit of code ([live example at the playground](http://playground-leaflet.rhcloud.com/wiy/edit?html,output)):
```js
var somewhere = L.latLng([51.505, -0.09]);
var circle = L.circle(somewhere, 1000).addTo(map);
var marker = L.marker(somewhere, { interactive: false}).addTo(map);

map.toString = function(){ return "map"; };
circle.toString = function(){ return "circle"; };

function logEvent(ev) {
	console.log(ev.type + ' fired on ' + ev.target, performance.now());
}

map.on('click', logEvent);
circle.on('click', logEvent);
```

The expected behaviour is: whenever the user clicks anywhere inside the circle, the `click` event fires. That includes pixels covered by the non-interactive marker.

@yohanboniface I'm looking into the problem right now, I'll see if this can be actually worked around with some CSS.

IvanSanchez:
Fixes #3936. Also fixes #2396.

Please note that I haven't tested this (yet) with IE10, which doesn't implement the CSS property `pointer-events:none`.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -203,6 +203,15 @@
 	cursor:    -moz-grabbing;
 	}
 
+/* marker interactivity */
+.leaflet-marker-icon,
+.leaflet-marker-shadow {
+	pointer-events: none;
+	}
+
+.leaflet-marker-icon.leaflet-interactive {
+	pointer-events: auto;
+	}
 
 /* visual tweaks */
 
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -203,13 +203,17 @@
 	cursor:    -moz-grabbing;
 	}
 
-/* marker interactivity */
+/* marker & overlays interactivity */
 .leaflet-marker-icon,
-.leaflet-marker-shadow {
+.leaflet-marker-shadow,
+.leaflet-image-layer,
+.leaflet-pane > svg path {
 	pointer-events: none;
 	}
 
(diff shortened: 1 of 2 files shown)


yohanboniface:
Sounds nice. Not how we could unittests that sadly :/

mourner:
We should probably add similar handling to all other layers with `interactive` option, e.g. ImageOverlay. And also change [this line](https://github.com/Leaflet/Leaflet/blob/55fe46250828d54d95cac0b8ca8b7a21dd97dbeb/src/layer/vector/SVG.js#L105) to do it from CSS side for consistency (and only override when `pointerEvents` option is set).

IvanSanchez:
Tested with SVG and image overlays. Untested with VML.

IvanSanchez:
The `L.Canvas` renderer is still problematic (if a `<canvas>` in the overlay pane, it will forcefully capture clicks meant for other overlays, such a `L.ImageOverlay`). However, the original bug (non-interactive markers and polygons) is fixed.

Issue #3941, opened 2015-10-15
Error when checking for moving draggable in canvas mouseMove handler


klaftertief:
Moving over a map with a canvas layer and the `dragging` option of a map is set to `false` throws an error. See http://jsbin.com/xoyiqifoki/edit?js,console,output

This is because the test in https://github.com/Leaflet/Leaflet/pull/3547/files#diff-5b3512e2ba3d374197128a67b330a0efR205 (introduced in #3547) doesn't check if the map is draggable.

yohanboniface:
Fixed by #3942

PR #3942, opened 2015-10-15
Only check for moving draggable in canvas renderer when map is draggable

klaftertief:
This is a fix for #3941

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -215,7 +215,7 @@ L.Canvas = L.Renderer.extend({
 	},
 
 	_onMouseMove: function (e) {
-		if (!this._map || (this._map.dragging._enabled && this._map.dragging._draggable._moving) || this._map._animatingZoom) { return; }
+		if (!this._map || this._map.dragging.moving() || this._map._animatingZoom) { return; }
 
 		var point = this._map.mouseEventToLayerPoint(e);
 		this._handleMouseOut(e, point);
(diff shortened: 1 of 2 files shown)


src/layer/vector/Canvas.js
```diff
@@ -215,7 +215,7 @@ L.Canvas = L.Renderer.extend({
 	},
 
 	_onMouseMove: function (e) {
-		if (!this._map || this._map.dragging._draggable._moving || this._map._animatingZoom) { return; }
+		if (!this._map || (this._map.dragging._enabled && this._map.dragging._draggable._moving) || this._map._animatingZoom) { return; }
```
yohanboniface: This becomes a bit hard to follow, I wonder if we should not have a method on the L.Draggable object to make this more readable.
Something like `this._map.dragging.moving()` that would do both checks.

yohanboniface: Also, `this._map._dragging`, inheriting from `L.Handler` has a `enabled()` method that should be used here instead of private property.

yohanboniface:
Good catch! See code comments :)

src/layer/vector/Canvas.js
mourner: I agree. Should be a simple change, but let's look if we can update similar places to use the new method.

klaftertief: Yep. I'm happy to update the pull request later.

klaftertief:
Ok, I just added a moving() method to the MapDrag handler, just like the moved() method there.
I didn't find any other occurrences of the `_moving` property, except in the `Draggable` class itself.

Issue #3947, opened 2015-10-16
Canvas paths disappear on animated zoom start and only appear again when zoom ends


klaftertief:
See http://jsbin.com/xoyiqifoki/edit?html,js,output or `debug/vector/vector-canvas.html` and zoom in/out. This was probably introduced in #3923 
There might be a race condition in writing and reading `this._map._animatingZoom` (the important read for this issue is here https://github.com/Leaflet/Leaflet/blob/v1.0.0-beta.2/src/layer/vector/Canvas.js#L29, it's always `undefined` or `false` there), but an easy fix might be to just not call `update()` in `_onZoomStart`.

Malvoz:
Fixed in https://github.com/Leaflet/Leaflet/pull/3950.

PR #3948, opened 2015-10-16
Disable `_onZoomStart` in canvas renderer

klaftertief:
This fixes disappearing paths when starting an animated zoom. See #3947

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -25,6 +25,8 @@ L.Canvas = L.Renderer.extend({
 		this._ctx = container.getContext('2d');
 	},
 
+	_onZoomStart: L.Util.falseFn,
+
 	_update: function () {
 		if (this._map._animatingZoom && this._bounds) { return; }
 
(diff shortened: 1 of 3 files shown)


IvanSanchez:
Good catch! Seems like I forgot to check the canvas renderer.

Maybe it would be a better idea to call `update()` only for the SVG renderer instead?

klaftertief:
That would mean to

- either split up the list of events in `getEvents()` between the base renderer and the SVG renderer
- or have the `_onZoomStart()` be a noOp in the base renderer and override it in the SVG renderer.

I'm not sure what would be the cleaner way. I can update the pull request either way.

IvanSanchez:
I think splitting `getEvents` would be cleaner, but I'm not 100% sure (no inheritance, etc). If you could try that and see if it looks better, that'd be great.

klaftertief:
The additional commit looks a bit nicer to me then the splitter `getEvents` in #3950.

Though I'm not sure if this is really the best fix in general. Panning and then zooming on mobile in the `debug/vector/vector-canvas.html` test looks works way better in the beta 1 version. There is a slight offset movement on zoom start now.

IvanSanchez:
Superseded by #3950

PR #3949, opened 2015-10-16
Add Jakefile to bower ignore list

IvanSanchez:
I noticed that, when installing beta2 from bower, the Jakefile is still there.

I wonder if `component.json` and the plugin guide need to be there too, though.

bower-files


IvanSanchez:
We're probably going to deprecate bower support sometime in the future, so let's dump this.

PR #3950, opened 2015-10-16
Alternative fix for #3947

klaftertief:
This should not get merged when #3948 gets merged. See https://github.com/Leaflet/Leaflet/pull/3948#issuecomment-148707860

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -25,6 +25,8 @@ L.Canvas = L.Renderer.extend({
 		this._ctx = container.getContext('2d');
 	},
 
+	_onZoomStart: L.Util.falseFn,
+
 	_update: function () {
 		if (this._map._animatingZoom && this._bounds) { return; }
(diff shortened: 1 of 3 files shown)


klaftertief:
Just mentioning #3947 here in the comment again to have a reference in the issue itself.

yohanboniface:
This seems a regression since 0.7.x, a `git bisect` would be helpful to be sure we are not breaking something else or missing the right fix, imho (idem for #3948).

yohanboniface:
Bisecting point at 2191cfa8b035280215ab7dda0895e1abaa68e490 so I'm happy to leave it to @IvanSanchez :)

klaftertief:
Thanks for the bisect. That was my guess in #3947 as well. '_animatingZoom' might get set too late to 'true' internally for a correct '_onZoomStart' behaviour.

IvanSanchez:
Looks good to me :-)

PR #3952, opened 2015-10-17
Implement L.Label

yohanboniface:
So this is a first naive approach, just hit the "it works" point.

Basically, I've extracted common methods of L.Popup to L.PopupBase, and made L.Label inherit from it.

The API is quite similar to Leaflet.Label plugin, but options are not exactly the same.

Here are the main supported option at this stage:
- **permanent** should the label appear on mouseover or be permanent
- **sticky** should the label follow the mouse or only appear at the layer center
- **direction** `left`, `right` or `auto`, like Leaflet.Label, plus `top` and `bottom`
- **interactive** should the label be clickable (in this case, it's a event target for the source it is attached to)

There is for sure a lot more work to be done (more factorisation, custom events…), but I wanted to push a first version to start discussion and feedback.

There is a debug page in `debug/map/label.html`.

## TODO

- [x] unittests
- [x] custom events
- [x] check with `FeatureGroup`
- [x] add support for `top`/`bottom` position
- [x] make the label styling more integrated with Leaflet styling (and make it easily overridable)
- [ ] factorize enough so one can create a new class inheriting from `L.PopupBase`, customize `initLayout` and it works™ (for a FancyLabel or WhizzyPopup plugin for example)
- [x] add documentation

## Preview

leaflet-label

@mourner @IvanSanchez feedback more than welcome :)


.gitignore
```diff
@@ -13,4 +13,5 @@ coverage/
 index.html
 .mailmap
 bower.json
-component.json
\ No newline at end of file
+component.json
+debug/local/
```
yohanboniface: I've a bunch of debug pages that I don't want to push, just added them here to prevent committing them by mistake.

src/layer/Label.js
```diff
@@ -0,0 +1,114 @@
...
+		offset: [12, -15],
+		direction: 'right',
+		static: false,  // Reserved word, use "permanent" instead?
+		followMouse: false,
+		clickable: false,
+		// className: '',
+		zoomAnimation: true
+	},
+
+	openOn: function (map) {
```
yohanboniface: This kind of method would better be factorized.

mourner:
Wow, you're fast! Taking a look on Monday. Looks pretty cool.

src/map/Map.js
```diff
@@ -519,6 +519,7 @@ L.Map = L.Evented.extend({
 		this.createPane('shadowPane');
 		this.createPane('overlayPane');
 		this.createPane('markerPane');
+		this.createPane('labelPane');
```
yohanboniface: Or we add labels in the popupPane?

mourner: Yeah, I think no need for a separate pane.

yohanboniface: This is the situation I wanted to prevent:

screenshot from 2015-10-18 09 32 41

I think we want the labels always above the markers and vectors BUT always below the popups, right?

Dedicated pane seemed to me the strongest way to achieve that, other option is to play with z-index inside the popup pane.

src/layer/Label.js
yohanboniface: A big part of this still duplicated code between `L.Label` and `L.Popup` comes from the need of having `this._label` or `this._popup` on the source. That's important because one can have both a label and a popup and we don't want that opening one would close the other.

Nevertheless this code duplication is under-optimal. It's theoretically possible to factorize and have a sort dynamic property naming (`this[prop]`) but at this stage I don't see any elegant (simple, readable and maintainable) way to achieve this. Any suggestion here is welcome :)

almet:
Neat. +1 Anything else needed?

src/layer/Layer.Label.js
```diff
@@ -0,0 +1,164 @@
+/*
+ * Adds label-related methods to all layers.
+ */
+
+L.Layer.include({
+
+	bindLabel: function (content, options) {
```
yohanboniface: @mourner any good idea to factorize with L.Layer.bindPopup (same with openLabel/Popup, toggleLabel/Popup…)?

Mainly, we would need to:
- make `this._popup`/`this._label` dynamic
- make `instanceof L.Popup` dynamic also
- deal with some specific behaviour here and there (like calling `this._map._openLabel` or `this._map._openPopup` or label `clickable` option)
Plus we want to keep `marker.bindPopup` and `marker.bindLabel` with API just as simple as now.

I've thought about something like having `_bindPopupBase(content, options, type, class)` that would then be called by both `bindLabel` and `bindPopup` (eg. `this._bindPopupBase(content, options, 'label', L.Label)`). But that turns out to be quite ugly code, not very elegant nor easily readable.
Another option would be to have a proxy model (like we have `L.Marker.Drag`), but I feel this would be overcomplicating the internal API.

Thoughts? :)

yohanboniface:
Hey, Santa Claus, can you send me a review for XMas? :)

computermacgyver:
Was just working on something very similar. This is PR is great and I would love to see it merged. Nice work!

yohanboniface:
ping @hyperknot for a review :)

yohanboniface:
Well, it needs a huge rebase, thanks to @IvanSanchez :p

yohanboniface:
Merged with master! So ready for review again :)

Note: I've not rebase, because (due to merge with leafdoc branch) the diff was huge, and I want to be sure I've not made a mistake during the merge. We'll rebase when ready for merging.

yohanboniface:
As discussed in IRC:

- added `top` and `bottom` directions support
- made the styling a bit more neutral and integrated with other Leaflet styling

Comments welcome :)

yohanboniface:
Added documentation.

dist/leaflet.css
```diff
@@ -521,3 +522,109 @@
...
+.leaflet-label:after {
+	border-left: 6px solid black;
+	border-left-color: inherit;
+	right: -16px;
+}
+
+.leaflet-label-top:after,
+.leaflet-label-top:before {
+	/* Calc include 4 px of border - minus half of the 6px of tip size  */
+	left: calc(-50% + 1px);
```
perliedman: `calc` isn't supported in IE8, and I still think we support IE8, right?

dist/leaflet.css
```diff
@@ -521,3 +522,109 @@
...
+	border-right-color: transparent;
+	border-bottom-color: transparent;
+	border-left-color: transparent;
+	border-top-color: transparent;
+}
+
+.leaflet-label-bottom:after,
+.leaflet-label-bottom:before {
+	/* Calc include 4 px of border - minus half of the 6px of tip size  */
+	left: calc(-50% + 1px);
```
perliedman: See comment above about `calc`.

hyperknot:
I've looked at Label, my biggest questions are:
- at the moment I believe it mostly is a port of Leaflet.Label, but I am not sure that defaults for a specific plugin are good defaults for core Leaflet. For example I am surprised that the default Label is a hover/popup style label. Is that really what most Leaflet users would think about when they want a Label? Do they need a direction by default?
- in my idea the default use case of a label would be a static text on a map, with no interactivity and no direction by default. but this is only my idea, we should discuss what is it for others and possibly ask users (if only here).
- i'm thinking about PopupBase, if we believe this is the right solution we have to be quite future proof here, to make sure that we are not artificially limiting either popup or label by having this hybrid parent class. In my opinion many of the popupbase options are very specific to popup right now.

yohanboniface:
Those are valid points, and I think mainly what L.Label should be by default is a matter of taste, and thus I can't argue.
So I'm gonna close here and extract back this work to a separate module (thus more opinionated).
Let's see after that if we can reintegrate it as new version of Leaflet.Label.

hyperknot:
One plus point for the separate module is that you can target more modern browsers, independent from Leaflet core. For example you can make it IE9+ or IE10+ compatible and just state it in the readme.

mourner:
@yohanboniface don't give up. :) I don't think we're ready to dismiss the idea of a common base for both label and popup, it just needs more review/discussion.

The Popup-specific options @hyperknot mentioned could be moved down to Popup — things like `autoPan`, width/height, `closeButton` etc. We could rename PopupBase to something like `DivOverlay` that has no label/popup specifics. And make Label simpler — e.g. assume simple single-line labels that don't need autopanning and buttons that open to the left or to the right.

yohanboniface:
> it just needs more review/discussion.

The thing is that this PR has been opened last October, and there have been no really interest in reviewing/discussing.
On the other hand it's now two years since a new version of uMap because I've based my devs on the Leaflet master trunk which was not stable enough to have a release. Label is the last thing I needed to publish that new uMap (rc) version.
Next week-end we have the French SotM, and I want to talk about the new uMap version, thus I want to push it live. So I need to move forward, so I think I'll dedicate my energy to a separate module for now. Two years is way too much. Including for my sanity. And I know you know what I mean ;)

Anyway, thanks for kindly reopening, and thanks @perliedman for your review :)

yohanboniface:
Moved options. Rebased.
Some tests are skipped in Travis but will be fixed when #4705 will be merged (by loading CSS).

src/map/Map.js
```diff
@@ -776,6 +776,9 @@ L.Map = L.Evented.extend({
 		// @pane markerPane: HTMLElement = 600
 		// Pane for `Icon`s of `Marker`s
 		this.createPane('markerPane');
+		// @pane labelPane: HTMLElement = 7
```
mourner: typo? is this 700?

src/layer/marker/Marker.Label.js
```diff
@@ -0,0 +1,9 @@
+/*
+ * Label extension to L.Marker, adding label-related methods.
+ */
+
+L.Marker.include({
+	_getLabelAnchor: function () {
+		return this.options.icon.options.labelAnchor || [0, 0];
+	}
+});
```
mourner: Maybe let's just move this to `Marker.js` — this is too small to be a separate file.

src/layer/PopupBase.js
```diff
@@ -0,0 +1,211 @@
+/*
+ * @class PopupBase
+ * @inherits Layer
+ * @aka L.PopupBase
+ * Base model for L.Popup and L.Label. Inherit from it for very custom popup like plugins.
```
mourner: Are we happy with the `PopupBase` name for the class?

src/layer/marker/Marker.Label.js
yohanboniface: Or merge with `Marker.Popup.js` in someting like `Marker.PopupBase.js` (whatever the name we'll chose? Just for consistency between the two.

debug/map/label.html
```diff
@@ -0,0 +1,69 @@
...
+		L.polygon([[41.20, 9.41], [41.20, 9.39], [41.21, 9.40]]).addTo(map).bindLabel('Polygon label following mouse', {sticky: true});
+		L.polygon([[41.18, 9.42], [41.17, 9.40], [41.19, 9.38]]).addTo(map).bindLabel('Permanent polygon label', {permanent: true});
+		L.marker([41.20, 9.4307]).addTo(map).bindLabel('label on the left', {direction: 'left'});
+		L.marker([41.206, 9.44]).addTo(map).bindLabel('click me, I have a popup', {permanent: true, interactive: true}).bindPopup('See?');
+		L.circleMarker([41.206, 9.48], {color: 'Chocolate', radius: 12}).addTo(map).bindLabel('Hello Left World', {direction: 'left'});
+		L.circleMarker([41.20, 9.50], {color: 'Chocolate', radius: 12}).addTo(map).bindLabel('Hello top World', {direction: 'top', permanent: true});
+        L.circleMarker([41.20, 9.47], {color: 'Tomato', radius: 10}).addTo(map).bindLabel('Seems I am centered', {direction: 'center', permanent: true, interactive: true}).bindPopup('Yeah');
+        L.circleMarker([41.195, 9.47], {color: 'Tomato', radius: 10}).addTo(map).bindLabel('Me too', {direction: 'center'}).bindPopup('Yeah');
```
jieter: Some inconsistent whitespace here

src/map/Map.js
```diff
@@ -776,6 +776,9 @@ L.Map = L.Evented.extend({
 		// @pane markerPane: HTMLElement = 600
 		// Pane for `Icon`s of `Marker`s
 		this.createPane('markerPane');
+		// @pane labelPane: HTMLElement = 700
```
jieter: Should be [`650` I guess](https://github.com/Leaflet/Leaflet/blob/a1e43ed3bb3032e6d43c76e53161cec8b03e4e6f/dist/leaflet.css#L87)

PR #3954, opened 2015-10-18
leaflet-layer-tree-plugin: Leaflet plugin allows to switch layers on …

bambrikii:
…and off, display then in a tree-like way.

leaflet-layer-tree-plugin: Leaflet plugin allows to switch layers on and off, display then in a tree-like way


mourner:
Thanks!

Issue #3521, opened 2015-06-08
Shift + scroll wheel zoom


IvanSanchez:
It would be nice if holding the Shift key while scrollwheel-zooming would multiply the zoom delta by 3.

Rationale is that I'm very used to shift-scrolling in text editors, and I expect that shift-scrolling would do the same as scrolling, but faster.

mourner:
Hmm, never seen that behaviour in any editors I've used — is this common? I don't mind adding this to Leaflet though, in addition to the already added shift-click on zoom buttons or shift+keypress +/-.

IvanSanchez:
I'd wait until #3523 gets merged in order to tackle this down. No need to worry about a minor feature such as this.

PR #3955, opened 2015-10-19
Add shiftkey option to scroll wheel zoom

neorth:
This PR adds the possibility to set the map option "scrollWheelZoom" to enable scroll wheel zoom only when the shift key is pressed. Useful when you want to disable scroll wheel zoom but still make it possible to use the scroll wheel in some way. 

I'm not sure if this breaks any other mouse/keyboard events so it would be great if someone with better Leaflet knowledge could take a look at it.

### DIFF
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -25,6 +25,7 @@ L.Map.ScrollWheelZoom = L.Handler.extend({
 	},
 
 	_onWheelScroll: function (e) {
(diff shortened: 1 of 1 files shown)


yohanboniface:
Ref https://github.com/Leaflet/Leaflet/issues/3521

yohanboniface:
Also https://github.com/Leaflet/Leaflet/issues/2076

Neorth:
Good references Yohan. Seems like me and Ivan are after different things.

IvanSanchez:
It kinda makes sense. #3521 could be implemented as a different `scrollWheelZoom` option (or as the default, or whatever).

Issue #3964, opened 2015-10-26
Simplify Animation calculations


mourner:
This looks bad: https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Renderer.js#L56

We need a really simple way to do animation calculations so that plugin developers could add animation easily.

cmulders:
The calculations become much simpler when the transform-origin is set to the middle, then the scaling does not influence the translation during the transform. This reduces the ```viewhalf``` calculation and the complicated ```topLeftOffset```

Made pull request with the changes

mourner:
Thanks! The problem is though that we also want the animation code between different layers (vectors, tile layers, image overlays, etc.) consistent. It was calculated with 50% origin earlier, but I switched it to `0 0` for every layer because having it different made things even crazier.

cmulders:
Ah I see your change (1e9c864). Hmm, then it becomes indeed more complicated, possibly some utility function to calculate the correct transform-offset for a given offset + scale? Reducing the math for plugin authors?

mourner:
@cmulders yes, exactly. Preferably we'd come up with a universal public method for things like this. I'm going to try rethinking this today with a fresh mind. This is a really brainmelting task — I remember my previous attempts.

cmulders:
I think that using the existing L.Transformation (possibly implementing Homogenous coordinates) consistently is the best option. Keeping track of a 'origin point' in every layer and multiplying this with the transformation matrix when a zoom animation is started. 

I tried to implement something, but I could not get setZoomAround working properly, because the origin of scaling is shifted (and must be done by the transformation). Seriously brainmelting..

mourner:
@cmulders thanks for looking into this! Now you see what I mean by the `brainmelt` label...

PR #3967, opened 2015-10-26
Simplify animation calculation

cmulders:
As suggested in #3964.

It changes the transform origin to the center of the Renderer pane, making it easier to translate and scale at the same time, without calculation overhead.


cmulders:
Center transform-origin is not preferred and basically reverts 1e9c864. Closing for now

Issue #3976, opened 2015-10-29
L.marker.setIcon() removes .leaflet-interactive


chronixlo:
Calling setIcon on a marker strips the leaflet-interactive -class.

Demo: http://playground-leaflet.rhcloud.com/mal/edit?html,output

IvanSanchez:
So when `L.Marker.setIcon` is called, `_initIcon` sets a flag and skips the logic for adding interactivity to the new icon. The solution is to review the behaviour of the `addIcon` flag in `_initIcon`, and/or make an explicit call to `L.Marker._initInteraction` from within `L.Marker.setIcon`.

This looks like a pretty easy task, so I'll ask newbies to stand up for the challenge and submit a pull request for this.

gorshkov-leonid:
Hi, we have problem with marker dragging:
https://jsfiddle.net/goleon/qo7a9no4/ -beta2 with hack (same as fix above)
https://jsfiddle.net/goleon/qo7a9no4/

If during handle of dragstart we changes icon then dragend will not be occurred and marker will jump during next drag because new position was not saved.

gorshkov-leonid:
this fix brokes https://github.com/Leaflet/Leaflet/pull/3687?

PR #3978, opened 2015-10-30
init interaction regardless of new icon or not

celadevra:
[issue #3976]

previously interaction only get inited when not reusing icon, i.e. when `addIcon`
flag is set to `true`. by taking the statement out of the if condition,
interaction is inited whenever `_initIcon` is called.


IvanSanchez:
Seems to work fine. Thanks, @celadevra !

PR #3981, opened 2015-11-01
Fixes #3753: Updates Custom Layer Example

### LINKED ISSUE #3753: Custom layer elements moving around when zooming off-center
Using the example from http://mourner.github.io/Leaflet/reference.html#layer

http://jsfiddle.net/stz8fm3y/

oliverroick:
I updated the docs using the example given by @IvanSanchez. Tested it using the jsfiddle from the issue #3753.


IvanSanchez:
Since the Leafdoc merge, `gh-pages-master` is used no longer. Edit the docstrings and send a new PR if you still feel this is needed.

PR #3984, opened 2015-11-02
Detach click handlers on layer removal properly

whatersoer:


mourner:
Closing because we just released 1.0 and minor issues in 0.7.x are unlikely to be fixed.

Issue #3979, opened 2015-10-30
once() can cause infinite loops


foundryspatial-duncan:
I noticed the one-time event listener function `once` [runs the callback, and _then_ unbinds the event listener](https://github.com/Leaflet/Leaflet/blob/v1.0.0-beta.2/src/core/Events.js#L186). If the callback function fires more of the same event, this can cause infinite loops, and I don't think that would ever be the desired behaviour.

jQuery's similar `one` function [unbinds the event listener before it runs the callback](https://github.com/jquery/jquery/blob/master/src/event.js#L83), so it can't ever get stuck in a loop.

A simple example:

```
/* this code causes an infinite loop */
myMap.once('fooEvent', function(){
    console.log('fooEvent fired!'); // this should only run once, but it goes infinitely
    myMap.fire('fooEvent'); // fire the event again inside the listener callback
});
myMap.fire('fooEvent'); // fire the event initially
```

jQuery equivalent:

```
/* this code works as expected */
$(myMap).one('fooEvent', function(){
    console.log('fooEvent fired!'); // this will only print once
    $(myMap).trigger('fooEvent'); // fire the event again inside the listener callback
});
$(myMap).trigger('fooEvent'); // fire the event initially
```

I was using Leaflet's `once` to listen for an initial `filtersReady` event, and then the callback would run a bunch of methods that could sometimes result in another `filtersReady` being fired, which froze the app in a loop described above.
Switching the order of the callback and the unbind would make `once` more flexible, and probably not break any existing code.

IvanSanchez:
@foundryspatial-duncan Seems you've got the problem rounded up. Care to submit a pull request?

foundryspatial-duncan:
I took a look, but I wasn't able to get it working with the existing code. I thought initially that it would be as simple as switching the order of the `.on` or `.off` lines, but that's not the case.

I also tried to mimic the flow jQuery.one uses, where it explicitly unbinds the listener and then calls `.apply` on the callback. Any thoughts?

radnor:
@foundryspatial-duncan I've successfully mimicked the flow of jQuery.one in #4003 - this should fix your issue!

mourner:
Closing in favor of #5376.

PR #4003, opened 2015-11-10
Prevent infinite loops inside once()

radnor:
I changed `once()` to act like `one()` from jQuery (see [here](https://github.com/jquery/jquery/blob/master/src/event.js#L83)) per the discussion in #3979. 

This resolves the issue where an event can enter an infinite loop, as `once()` currently unbinds the event listener _after_ triggering it.

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -362,7 +362,7 @@ describe('Events', function () {
 			    spy = sinon.spy();
 
 			obj.once('test', spy, obj);
-			obj.off('test', spy, obj);
+			obj.off('test');
 
 			obj.fire('test');
 
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -183,16 +183,17 @@ L.Evented = L.Class.extend({
 			return this;
 		}
 
-		var handler = L.bind(function () {
-			this
-			    .off(types, fn, context)
-			    .off(types, handler, context);
-		}, this);
+		// remove the listener first to avoid potential recursion and then execute the function
 
-		// add a listener that's executed once and removed after that
+		var origFn = fn;
+		var _this = this;
+		fn = function () {
+			_this.off(types, fn, context);
+			return origFn.apply(_this, arguments);
+		};
+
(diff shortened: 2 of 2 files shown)


spec/suites/core/EventsSpec.js
```diff
@@ -362,7 +362,7 @@ describe('Events', function () {
 			    spy = sinon.spy();
 
 			obj.once('test', spy, obj);
-			obj.off('test', spy, obj);
+			obj.off('test');
```
yohanboniface: This change seems dubious to me. Why do we need it?

src/core/Events.js
```diff
@@ -183,16 +183,17 @@ L.Evented = L.Class.extend({
...
-		var handler = L.bind(function () {
-			this
-			    .off(types, fn, context)
-			    .off(types, handler, context);
-		}, this);
+		// remove the listener first to avoid potential recursion and then execute the function
+
+		var origFn = fn;
+		var _this = this;
+		fn = function () {
```
yohanboniface: Minssing `var`.

src/core/Events.js
```diff
@@ -183,16 +183,17 @@ L.Evented = L.Class.extend({
...
 
-		var handler = L.bind(function () {
-			this
-			    .off(types, fn, context)
-			    .off(types, handler, context);
-		}, this);
+		// remove the listener first to avoid potential recursion and then execute the function
+
+		var origFn = fn;
+		var _this = this;
```
yohanboniface: Why not using `L.bind` instead?

src/core/Events.js
yohanboniface: ah no, OK, it comes from the arguments. But why override it instead of naming it `handler` like before?

perliedman:
Long time reviewing this, _but_... I've tried this approach myself while trying to solve #5376, and it has the unfortunate side effect that you can no longer do `off(event, fn)` after `once`, since the event handler for `fn` isn't explicitly registered any longer.

I'm going to close this, but we still need a solution for #5376 and #3979.

Issue #4001, opened 2015-11-10
Official support for adding data to markers


ccolgrove:
Markers and CircleMarkers both have an "options" parameter which supports a certain set of customizations, but there's no explicit support for putting any additional data on the markers. However, due to the way options are implemented, passing additional options to the Marker and CircleMarker constructors will actually attach these options to the marker for later use. 

We are currently using this functionality to attach ids to each of the markers so we can identify them later.  It's important that we be able to supply these ids ourselves so they match the ids elsewhere in the code. While our current solution works, we'd prefer not to rely on undocumented functionality and would like to see a supported way to achieve this.

The simplest way to support this is just to document that all options, even those that aren't explicitly defined on PathOptions or MarkerOptions, get added to the markers.

IvanSanchez:
> The simplest way to support this is just to document that all options [...] get added to the markers.

That actually happens with all classes. I suggest you have a look at the `reference.md` file in the `gh-pages` branch, add some info about this in the "options" bit of the `L.Class` documentation, and submit a pull request.

mourner:
@ccolgrove you could also just do `marker.id = id;`.

ccolgrove:
@mourner Yes, definitely possible, but since we're using typescript rather than plain javascript, it's preferable for us to use the existing types when possible.

HarelM:
So this means I should use `marker.options.anyDataIWant`?

perliedman:
@HarelM not necessarily _should_, but you _could_. :smile: 

As we've discussed before, I wouldn't use this approach myself, I find for example GeoJSON features with properties a way better alternative for things like this. Another way is to create for example markers in a JavaScript scope where you have access to the instance specific data you need in for example listeners.

The only thing is that we now document this possibility, since apparently at least some people want to work this way, and we certainly will not stop them.

PR #4004, opened 2015-11-10
Add documentation about unrecognized options

ccolgrove:
https://github.com/Leaflet/Leaflet/issues/4001: Specifically document that options that are otherwise not recognized by particular classes will still get merged with the options object.


IvanSanchez:
This is already documented in `master` at https://github.com/Leaflet/Leaflet/blob/master/src/core/Class.leafdoc#L86 , and we want to phase out the 0.7.x docs as 1.0.0 approaches, anyway.

PR #4010, opened 2015-11-11
Fix event target fallbacking to map after marker drag (fix #3971)

### LINKED ISSUE #3971: Event bubbling issue for draggable markers in 1.0b1
We have discovered an issue in our platform that we didn't notice before updating to 1.0b1..
Looks like draggable markers are having their dragend event fall into click events on the underlying map(if the marker doesn't have a popup bound to it)
I have built a simple test page (https://github.com/snkashis/Leaflet/blob/bd84dc7c428bccf77b2a1bb39d1effeeea15bcca/debug/tests/click_after_drag.html) in my branch https://github.com/snkashis/Leaflet/tree/drag_bubbling to demonstrate.

To see this behavior, try dragging the east-most marker of the two on this map. 
In 1.0b1, a click event(which logs 'click' to the console) on the map is fired after drag end.
Bisection indicates ee3f2f8b1e259833badd590a646e2523a5eaff7b is the problem commit.

If you copy this new debug file into your repo, and then switch back to 7ca19b59eae4e98abf0fc0fbd46e0ce335b9d18b(the previous), you will notice that the click event is not fired. Is there some other issue in play here causing this?

yohanboniface:
Fix #3971 

Given that the dragged marker was not listening to click events, `_findEventTargets` was using current map instance as default (which is the normal behaviour) target, but then in `_fireDOMEvent` we were not checking `draggableMoved` with the real event target.

@mourner `e._simulated` comes from f9ea4758f8082c68ada8c05c9374e59460cc6fdc, but I don't get why we needed and still need it when checking for `draggableMoved`, do you recall? :)

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -647,10 +647,16 @@ L.Map = L.Evented.extend({
 		var targets = [],
 		    target,
 		    isHover = type === 'mouseout' || type === 'mouseover',
-		    src = e.target || e.srcElement;
+		    src = e.target || e.srcElement,
+		    dragging = false;
 
 		while (src) {
 			target = this._targets[L.stamp(src)];
+			if (target && (type === 'click' || type === 'preclick') && !e._simulated && this._draggableMoved(target)) {
+				// Prevent firing click after you just dragged an object.
+				dragging = true;
+				break;
+			}
 			if (target && target.listens(type, true)) {
 				if (isHover && !L.DomEvent._isExternalTarget(src, e)) { break; }
 				targets.push(target);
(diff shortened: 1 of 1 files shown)


mourner:
I don't recall :(

IvanSanchez:
@yohanboniface `ev._simulated` is about synthetic `click` events coming from very small touch-drags. In other words, a simulated click is a drag small enough to fire events.

IvanSanchez:
@yohanboniface Not the cleanest PR ever (I wouldn't expect less from a 'brainmelt' tag ;-) ) but looks good for me.

PR #4019, opened 2015-11-16
Add new plugin: Leaflet.GeoJSONAutocomplete

utahemre:
This plug-in works for us successfully in several projects. I hope this will help to others.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -1469,6 +1469,15 @@ Plugins that search for overlays and enhance how to display information about th
 			<a href="http://yohanboniface.me">Yohan Boniface</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Looks good! Two comments, though:

- Is it necessary to write your name in capital letters (i.e. `Özkaya` instead of `ÖZKAYA`)? I would prefer to contributors' names to be consistent in that regard.
- Your documentation is a bit unclear on the format of the GeoJSON search service. Maybe you could add something like "Every time the user types in the search box, a query to the service is made". Also, an example of a request-response cycle would be nice IMHO.

utahemre:
Thank you very much for your interest and fast reply. I changed my name and add more detailed documentation. Please let me know if you have another comments.

IvanSanchez:
@utahemre ;-)

PR #4022, opened 2015-11-17
Load more tiles than the actual bounds.

artdent:
This makes for smoother panning, because adjacent tiles are already loaded.


IvanSanchez:
This has been discussed (and rejected) in the past, see #3551 and #3087

artdent:
Cool, the plugin mentioned there (https://github.com/TolonUK/Leaflet.EdgeBuffer) should solve this. Thanks for the pointer!

Issue #3888, opened 2015-10-02
Overlayed circles does not respond to mouse events (CANVAS ONLY)


shawel:
Mouseover events on circles added as a layer on top of state layers do not respond to mouse events. 

Here is an example. Hovering over circle should update the text  top right legend. This feature existed on 0.7 version of leaflet. It works on svg version. 

http://playground-leaflet.rhcloud.com/hag/edit?html,output

IvanSanchez:
OK, the culprit is around https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L314

That is never called because `L.Canvas._updateCircle` doesn't add the circle marker to `this._drawnLayers`, so the loops in `_onClick` and `_handleMouseHover` skip it.

So in order to fix this:
- Change the code in `L.Canvas._updateCircle` to make sure that circle markers are added/removed to/from `this._drawnLayers`
- Check that `L.CircleMarker._containsPoint` runs on every pointer interaction on a canvas renderer
- Check that this works for both `L.CircleMarker` and `L.Circle`

IvanSanchez:
I think this is easy enough for any newbie dev to fix, so I'm calling the good folks at @yourfirstpr to see if somebody wants to pick this one up. Remember to read `CONTRIBUTING.md`.

Charlotteis:
@IvanSanchez thanks Ivan! I'll be tweeting about this at 1PM GMT today. Great instructions :sunny:

brylie:
Hi, just following up with a couple of questions. 

Are the circle markers in the `ctx`object? Where are the pointer interaction event hooks?

IvanSanchez:
> Are the circle markers in the ctx object? 

@brylie No, `ctx` is a [2D Canvas rendering context](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D).

The instances of `L.CircleMarker` (and any other subclass of `L.Path`, such as lines and polygons) are in `L.Canvas._layers`. Paths add themselves to a `L.Renderer` (by default `L.Map` instantiates one). Paths ask the renderer to re-render on certain conditions (and not the other way around).

> Where are the pointer interaction event hooks? The following seem relevant:

Yeah, events are set up in [`L.Canvas._initContainer()`](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L20).

[...]

PR #4033, opened 2015-11-19
circle markers not being added to drawn layers

Deftwun:
Seems to fix issue #3888 

Basically a newbie so let me know if I'm doing this wrong ;)


mourner:
Merged as a part of #4072, thanks!

PR #4040, opened 2015-11-22
Fix GridLayer pixelBounds computation

yohanboniface:
Map size does not change from one zoom to another, so we don't need
to scale it.

Without this change, we are creating more or less 4x tiles when
zooming in.
There are still cases where we are requesting a bit more tiles than
what I would expect, but I see it also happens on 0.7.3.

I've added a test, it passes on Firefox or Chromium, but it fails on phantomJS, not clue why, joy of headless browser… Not sure if we want to keep it or not, then.

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -106,6 +106,48 @@ describe('GridLayer', function () {
 		});
 	});
 
+	describe('#createTile', function () {
+
+		beforeEach(function () {
+			// Simpler sizes to test.
+			div.style.width = '512px';
+			div.style.height = '512px';
+		});
+
+		afterEach(function () {
+			div.style.width = '800px';
+			div.style.height = '600px';
+		});
+
+		// Passes on Firefox, but fails on phantomJS: done is never called.
+		xit('only creates tiles for visible area on zoom in', function (done) {
+			map.remove();
+			map = L.map(div);
+			map.setView([0, 0], 10);
+
+			var grid = L.gridLayer(),
+			    count = 0,
+			    loadCount = 0;
+			grid.createTile = function (coords) {
+				count++;
+				return document.createElement('div');
+			};
+			var onLoad = function (e) {
+				expect(count).to.eql(4);
+				count = 0;
+				loadCount++;
+				if (loadCount === 1) {  // On layer add.
+					map.zoomIn();
+				} else {  // On zoom in.
+					done();
+				}
+			};
+			grid.on('load', onLoad);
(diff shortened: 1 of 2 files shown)


mourner:
Interesting! Did you test that on mobile devices with pinch-zoom?

johnbendi:
Original Message                                                                                                                                                                                                                                                                                                                                                                                                                                      From: Vladimir AgafonkinSent: Monday, November 23, 2015 5:05 PMTo: Leaflet/LeafletReply To: Leaflet/LeafletSubject: Re: [Leaflet] Fix GridLayer pixelBounds computation (#4040)Interesting! Did you test that on mobile devices with pinch-zoom?

—Reply to this email directly or view it on GitHub.

yohanboniface:
I tested with non round zooms on my laptop, and it works as expected. I'll do a test with mobile :)

But it you think a bit more about the logic of the algo, I think this patch makes sense. Note that I've been like two hours trying to understand why the pixelBound was not the one expected, and I was really finding the pattern "compute size" + "divide by 2 to get the half" + "scale it with the zoom difference" so logical that I was looking every where around but there :p

What we want is the pixel bounds (in pixels, so…) at the tile zoom (the final displayed zoom), and I think this pixel bounds will always be the width of the map container at the end, no?

I'm not even sure we need to compute the new center before actually.

yohanboniface:
Just tested on mobile, I see no issue.

mourner:
:+1:

hyperknot:
@yohanboniface can you check if https://github.com/Leaflet/Leaflet/pull/4156 left the tile usage the same? Is there an example for trying it?

yohanboniface:
> can you check if #4156 left the tile usage the same?

No, it does not… cc @IvanSanchez 

> Is there an example for trying it?

You can run this test on Firefox (any browser but PhantomJS actually): https://github.com/Leaflet/Leaflet/pull/4040/files#diff-19a7d53df5d545ee8cde69a53256057cR123

hyperknot:
I tried to make a playground example from @yohanboniface 's code.

Results:
- 0.7.5 (only using browser tools): 8 tiles at load, +6 tiles at zoom: http://playground-leaflet.rhcloud.com/puh/1/edit?js,output
- 1.0 beta 2: 8 tiles at load, +24 tiles at zoom: http://playground-leaflet.rhcloud.com/pux/2/edit?js,console,output
- 1.0 just before #4156: 8, +6, http://playground-leaflet.rhcloud.com/meq/1/edit?js,console,output
- 1.0 after #4156: 8, +24 http://playground-leaflet.rhcloud.com/pul/1/edit?js,console,output

Issue #3913, opened 2015-10-09
Calling panBy() while already panning results in blurry tiles (Chrome)


RLRR:
JSFiddle: http://jsfiddle.net/0x66297q/

Tiles become blurry after panning animation and stay like that even if you drag/zoom the map.

This reproduces in Chrome with current master.

IvanSanchez:
Duplicate of #3697

PR #4046, opened 2015-11-24
Round new map position before animating pan

RLRR:
This prevents map from panning into a fractional position (and becoming blurry) if `panBy` is called while already panning.

Demo: http://jsfiddle.net/0x66297q/

Fixes https://github.com/Leaflet/Leaflet/issues/3913 which was closed as a duplicate, however, I think this particular case is worth fixing :)

### DIFF
--- a/src/map/anim/Map.PanAnimation.js
+++ b/src/map/anim/Map.PanAnimation.js
@@ -69,7 +69,7 @@ L.Map.include({
 		if (options.animate !== false) {
 			L.DomUtil.addClass(this._mapPane, 'leaflet-pan-anim');
 
-			var newPos = this._getMapPanePos().subtract(offset);
+			var newPos = this._getMapPanePos().subtract(offset).round();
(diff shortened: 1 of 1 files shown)


mourner:
Nice, thanks! Just wondering, the fractional position is coming from `map._getMapPanePos()`, right? Would it be worth rounding there to potentially fix similar issues in other places?

RLRR:
> Would it be worth rounding there to potentially fix similar issues in other places?

1. I have looked through usages of `_getMapPanePos`, and `panBy` looks like the only place where rounding is needed.

2. Modifying `_getMapPanePos` is going to change behavior of `map._moved()` function—it will return `false` instead of `true` if offset is less than 0.5 pixels (although it is unlikely to affect anything).

3. It also seems to be more reasonable generally to leave private low-level functions as dump as possible.

So I'd rather leave this rounding where it is :) What do you think?

mourner:
OK, sounds reasonable. Thanks!

PR #4055, opened 2015-11-27
Replace links to google groups with ones to Stack Overflow

ghybs:
Following discussion in PR #3886, proposition to switch to Stack Overflow instead of Google Groups.

GIS is still mentioned though (besides SO) on bottom of Overview page, as the community is also active there (but less material than in SO as of today).

As mentioned in #3886, if this PR is accepted, next step would be to update CONTRIBUTING in master ([here](https://github.com/Leaflet/Leaflet/blame/master/CONTRIBUTING.md#L15) and [there](https://github.com/Leaflet/Leaflet/blame/master/CONTRIBUTING.md#L41)), and to post an "official" notice in the google group.


IvanSanchez:
I like this, but I wonder what @mourner and @yohanboniface think.

mourner:
Let's do it.

PR #4063, opened 2015-12-01
call getSize() instead of _size to avoid render error

### LINKED ISSUE #4062: L.control.layers missing call to L.Map.getSize()
In version 1.0.0-b1 I am noticing a very weird bug.

I am using this code to create a control layer with options

```javascript
function generateMap(map) {
    var mapTiles = getMapTiles();
    var overlayMaps = getMaps();
    var options = {
        collapsed: false,
    };
    L.control.layers(mapTiles, overlayMaps, options).addTo(map);
}
```

When I include the `options` variable, rendering fails with the error `ncaught TypeError: Cannot read property 'y' of undefined.

the script works perfectly when I omit the options.

```javascript
function generateMap(map) {
    var mapTiles = getMapTiles();
    var overlayMaps = getMaps();
    L.control.layers(mapTiles, overlayMaps).addTo(map);
}
```

Brobin:
Resolves #4062 

Changed the call of `_map._size` in `L.Control.Layers._expand` to `_map.getSize()` so the value can be retrieved before rendering or user interaction.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -254,7 +254,7 @@ L.Control.Layers = L.Control.extend({
 	_expand: function () {
 		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
 		this._form.style.height = null;
-		var acceptableHeight = this._map._size.y - (this._container.offsetTop + 50);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Nice!

PR #4064, opened 2015-12-02
Add Bing Maps tile layer plugin

gmaclennan:


IvanSanchez:
The class name should be `L.TileLayer.Bing` and **not** `L.BingLayer`, as per the [plugin guide](https://github.com/Leaflet/Leaflet/blob/master/PLUGIN-GUIDE.md#plugin-api).

Also, you should note in the documentation that `options` inherits from the available L.TileLayer options (e.g. you can use `minZoom` and `opacity` and etc).

Please fix that and squash your commits into one.

gmaclennan:
Sorry missed that piece of the plugin guide. All fixed now.

IvanSanchez:
@gmaclennan No problem :-)

PR #4065, opened 2015-12-02
insert add/removeEventParent in documentation

jgravois:
* added `addEventParent()` and `removeEventParent()` to Event Methods documentation
* added information about `fire()`s optional third parameter which allows events to [propogate](https://github.com/Leaflet/Leaflet/blob/master/src/core/Events.js#L198)

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -4661,6 +4661,14 @@ map.off('click', onClick);</code></pre>
 		<td>Adds a set of type/listener pairs, e.g. <code>{click: onClick, mousemove: onMouseMove}</code></td>
 	</tr>
 	<tr>
+		<td><code><b>addEventParent</b>(
+			<nobr>&lt;Object&gt; <i>parent</i> )</nobr>
+		</code></td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
`addEventParent` and `removeEventParent` do not exist in 0.7.x, which is what `reference.html` documents right now. These have been taken care of in the 1.0.0 docs already.

jgravois:
> These have been taken care of in the 1.0.0 docs already.

ok.  it took some digging but i was able to find the changes in the `gh-pages-leafdoc` branch in the `reference-1.0.0.html` file.  

it looks like you're working on something new and cool to auto-generate the API reference (and maintain legacy doc at the same time).  once you iron out the procedure it'd be helpful if you updated the [Improving Documentation](https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#improving-documentation) section of your `CONTRIBUTING.md` to give people a clue about whats going on.

PR #4072, opened 2015-12-05
fix bug for hover event between circle overlapping polygon on canvas

fimietta:
@codebar, @ladieswhocode
 * Added layer to drawnLayers in updateCircle
 * removed check to see if the was a hoveredLayer in handleMouseHover
 * removed break statement in loop in handleMouseHover. This should find last rather than first layer in drawnLayers array. Uncertain if this properly represents z-level!

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -157,6 +157,8 @@ L.Canvas = L.Renderer.extend({
 		    r = layer._radius,
 		    s = (layer._radiusY || r) / r;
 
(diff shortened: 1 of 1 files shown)


fimietta:
bugfix to issue #3888

IvanSanchez:
Nice work, thanks!

This seems to get rid of #3888 but, unfortunately, creates an edge case when hovering from a feature to another doesn't fire a `mouseout` event. We have similar problems with overlaid SVG vector layers, so I guess a new issue will be created if this becomes a real problem.

Issue #2427, opened 2014-01-31
panInsideBounds becomes unstable with custom projection


perliedman:
In an issue reported on Proj4Leaflet (https://github.com/kartena/Proj4Leaflet/issues/59#issuecomment-33789825), @kvolleberg has figured out that ```panInsideBounds``` can become unstable for some situations, causing the map to continue panning very slowly in some direction.

An easy fix appears to be to adjust ```LatLng.MAX_MARGIN```, to make it less sensitive to small coordinate differences, but that might of course also affect other code.

The calculations in ```Map._limitCenter``` could also be adjusted to make it less sensitive to small rounding errors, but I'm not sure how the best accomplish that.

perliedman:
Here's a fiddle to reproduce the issue: http://jsfiddle.net/v5ej8rv4/

As mentioned in #3200, it seems that the bug is most easily triggered by switching to fullscreen (F11) and zooming in a bit. Don't know if it's related, but zooming near to northern bounds (about Den Helder in the fiddle), seems to easily trigger it.

Have been able to reproduce it on both Chrome 40 and Firefox 35 on Ubuntu

perliedman:
Ok, made an update to the fiddle that monkey patches `Map.panInsideBounds` to output the difference between `center` and `newCenter` when there is a difference.

When the bug triggers, the output looks like this on my machine:

```
"dlat=4.429224986779445e-9, dlng=1.2464296261782692e-9" script.js:50
"dlat=4.429288935625664e-9, dlng=1.2464358434272071e-9" script.js:50
"dlat=4.42935998989924e-9, dlng=1.2464411724977253e-9" script.js:50
"dlat=4.429445255027531e-9, dlng=1.2464447252114041e-9" script.js:50
"dlat=4.4294878875916766e-9, dlng=1.2464500542819223e-9" script.js:50
"dlat=4.429551836437895e-9, dlng=1.2464536069956011e-9"
```

Currently the tolerance for difference between two `LatLng`s is set by `L.LatLng.MAX_MARGIN` to 1e-9. Obviously, changing this to 1e-8 makes the problem go away on my machine. From https://github.com/Leaflet/Leaflet/issues/2427#issuecomment-33874223, it seem that this might not be a general fix, though, although I can preproduce such a huge error (0.5 degrees difference sounds a bit odd, maybe indicating some other problem).

I'll think some more about this.

perliedman:
[...]

PR #4074, opened 2015-12-07
Move L.LatLng.equals to L.CRS.equals.

perliedman:
This makes it possible to set a default margin (tolerance)
depending on CRS implementation, for example to address
problems with custom projections as mentioned in #2427.

`L.LatLng.equals` still exists, but uses `L.CRS.Earth.equals`, much like `L.LatLng.distanceTo` falls back to using `L.CRS.Earth.distance`.

I'm a bit unsure of the name - perhaps `equals` should have a more expressive name, to make it clear that it's two latlngs that are being compared?

Currently not covered by this PR: `L.LatLngBounds.equals` will still use `L.LatLng.equals`, and not be dependant on CRS, so there's no way to compare to bounds to each other with CRS dependant equals. I can see two alternatives to address this:

1. Create yet another method in `CRS` that compares two bounds instances
2. Make it possible to pass a CRS to `L.LatLngBounds.equals` as an option, which is used to compare the latlngs to each other, defaulting to `L.CRS.Earth`.

I prefer 2, since it's not really the bounds comparison that is CRS dependant, just the LatLng comparison that it uses. However, alternative 1 is more in line with `equals` and `distanceTo`.

### DIFF
--- a/src/geo/LatLng.js
+++ b/src/geo/LatLng.js
@@ -17,15 +17,7 @@ L.LatLng = function (lat, lng, alt) {
 
 L.LatLng.prototype = {
 	equals: function (obj, maxMargin) {
-		if (!obj) { return false; }
-
-		obj = L.latLng(obj);
-
-		var margin = Math.max(
-		        Math.abs(this.lat - obj.lat),
-		        Math.abs(this.lng - obj.lng));
-
-		return margin <= (maxMargin === undefined ? 1.0E-9 : maxMargin);
+		return L.CRS.Earth.equals(this, obj, maxMargin);
 	},
 
 	toString: function (precision) {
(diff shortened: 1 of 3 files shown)


mourner:
This looks great! Two random thoughts on this:

1. Do we want to introduce max margin as a CRS property so you can set a default one for each CRS? 
2. Would it make sense to compare projected or screen coordinates rather source values?

perliedman:
1. Yes, I think that's a good idea, will add that
2. For #2427, I think [I already suggested that](https://github.com/Leaflet/Leaflet/issues/2427#issuecomment-33868545), if I understand what you mean correctly; unfortunately it was reported that this, surprisingly, did not fix the problem - it could be worth giving it a shot again though, since it sounds like a simpler solution

perliedman:
I vote for closing this in favor of #4077!

Issue #2703, opened 2014-05-23
Map.setMaxBounds does not work with non-inverted CRS


Expyron:
I'm using a CRS with all positive transformation coefficients (e.g. "L.Transformation(1,0,1,0)" ). The coefficient c especially is positive, meaning that the y axis is not reverted.

When using Map.setMaxBounds, the map does get correctly bound horizontally. However, if I try to move the map vertically, it always goes back to the vertical center of the bounds, even if I didn't pan it out of bounds.

Expected result: Map should be panned correctly inside the bounds.
Observed result: Map cannot be moved vertically and always goes back to the center of the bounds.

I tested it with several browsers and always got the same error. I'm using the master branch

OleLaursen:
I have a fix for this in #4030.

I'm currently monkey-patching _getBoundsOffset from the outside, and so far everything else seems to be working with the non-inverted y axis.

PR #4078, opened 2015-12-09
Fix bug with max bounds and custom projections

OleLaursen:
_getBoundsOffset was assuming that SE/NW were the min/max corners, but
this isn't necessarily true after projection.

Fixes issue #2703.


OleLaursen:
I ended up not adding a test case for this - it looks to me as if the right approach would be to run a bunch of the tests twice, with a default projection and a custom inverted one. But I'm afraid that's a bit more involved than I can manage right now. :)

perliedman:
I've done some quick testing with the pages in `debug`, and this looks :+1:. Leaving the final decision to merge to you, @mourner.

mourner:
Looks great to me. Thanks for taking on this!

PR #4082, opened 2015-12-11
add leaflet-echarts plugin

wandergis:
I have written a plugin for leaflet,this plugin can show bigdata visualization more cool. hope you  accept it.


IvanSanchez:
Looks very good! (but a bit slow)

PR #4089, opened 2015-12-14
Make sure to always reset _enforcingBounds.

perliedman:
In some cases, the flag `_enforcingBounds` was not reset, especially now that we exit early when `center` and `newCenter` is equal.

This PR wraps the logic in a try-finally-block to make sure the flag is always reset.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -177,17 +177,15 @@ L.Map = L.Evented.extend({
 
 	panInsideBounds: function (bounds, options) {
 		this._enforcingBounds = true;
-		try {
-			var center = this.getCenter(),
-			    newCenter = this._limitCenter(center, this._zoom, L.latLngBounds(bounds));
+		var center = this.getCenter(),
+		    newCenter = this._limitCenter(center, this._zoom, L.latLngBounds(bounds));
 
-			if (center.equals(newCenter)) { return this; }
-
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Wouldn't it be easier to just ... ?
```js
if (center.equals(newCenter)) { 
  this._enforcingBounds = false;
  return this;
}
```

perliedman:
Sure, it's a matter of taste I guess.

Personally, I like try-finally to avoid the duplicate flag reset (it must be reset for normal exit as well) and also always making sure the flag is reset no matter what , even if something throws an exception in there, but I guess the advantage of that can be debated.

It might be that try-finally is more common in C#/Java. It appears this is the first time `finally` is used in Leaflet, so perhaps better to do it your way?

yohanboniface:
Even if the try/finally is more elegant, afaik it's not optimizable (see https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax for example), so I'd duplicate the reset as for now.

perliedman:
Ok, removed the try-finally, hope this looks alright now!

@yohanboniface interesting point regarding optimization, will keep that in mind (although this code is not on any performance critical path, as I understand it).

yohanboniface:
Nice turnaround :)

PR #4094, opened 2015-12-15
Use _tileCoordsToKey everywhere

### LINKED ISSUE #3986: GridLayer tile cache key generation inconsistent
In GridLayer.js there is a method:

```javascript
    // converts tile coordinates to key for the tile cache
    _tileCoordsToKey: function (coords) {
        return coords.x + ':' + coords.y + ':' + coords.z;
    },
```

which provides a key to be used for the tile cache given tile coordinates (x,y,z). There are however two places in GridLayer.js where this method is not used for generating the key, namely in _retainParent and _retainChildren where we have somthing like:

```javascript
    var key = i + ':' + j + ':' + (z + 1),
```

Since developers may override the tileCoordsToKey method this can cause problems and is somewhat inconsistent. I have for example tiles which also have a time dimension (and vertical dimension).

My suggestion is to use the _tileCoordsToKey method in those two places.

Kind regards,
Jesper

jblarsen:
Fixes #3986

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -293,9 +293,11 @@ L.GridLayer = L.Layer.extend({
 	_retainParent: function (x, y, z, minZoom) {
 		var x2 = Math.floor(x / 2),
 		    y2 = Math.floor(y / 2),
-		    z2 = z - 1;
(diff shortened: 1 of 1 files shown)


yohanboniface:
Thanks :)

PR #4100, opened 2015-12-17
Added FeatureCollection to geometryToLayer function

franbenz:
A GeoJSON object with the type "FeatureCollection" can now be transformed to a layer.

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -123,6 +123,15 @@ L.extend(L.GeoJSON, {
 			}
 			return new L.FeatureGroup(layers);
 
+		case 'FeatureCollection':
+			for (i = 0, len = geometry.features.length; i < len; i++) {
+				layers.push(this.geometryToLayer(geometry.features[i],
+					pointToLayer,
+					coordsToLatLng,
+					vectorOptions));
+			}
+			return new L.FeatureGroup(layers);
+
 		default:
 			throw new Error('Invalid GeoJSON object.');
 		}


IvanSanchez:
This functionality is already in 1.0.0, and I'm not really sure it's really needed to be backported into stable at this point. :-/

franbenz:
I needed this on the stable branch since I'm using leaflet as an angular directive with a project that depends on 0.7.x (https://github.com/tombatossals/angular-leaflet-directive/blob/master/bower.json#L18)

mourner:
This only affects nested feature collections, right? Because `addData` handles the top level feature collection case already.

franbenz:
No, in my application I got an "Invalid GeoJSON object" error when passing a Feature Collection object to geometryToLayer(in the stable branch), that's why I needed the patch I'm adding.

franbenz:
Any suggestions with this issue?

kentr:
@franbenz Are you passing a GeoJSON string for `data` in the layer definition?  If so, that may be the cause of the error.  I'm also using ui-leaflet, and it caused this error for me.

I question whether `geometryToLayer()` is supposed to receive `Feature` or `FeatureCollection` objects, rather than exclusively [`Geometry` objects](http://geojson.org/geojson-spec.html#geometry-objects).

For me, the solution was to create the layer definition similar to this:

```
var newLayerDef = {
  name: 'POI',
  type: 'geoJSONShape',
  visible: true,
  data: JSON.parse('{"type": "FeatureCollection", "features": [ ... ]}'),
  layerOptions: {
    pointToLayer: function (feature, latlng) {
      return L.marker(latlng);
    }
  },
};
```

mourner:
Closing because we just released 1.0 and minor issues in 0.7.x are unlikely to be fixed.

Issue #4103, opened 2015-12-19
Fix retina in 1.0


hyperknot:
There are two ways of displaying retina resolution titles.

**Real retina tile rendered on server**
In this mode, instead of loading `1/2/3.png` leaflet of size 256x256 pixels, the client is downloading `1/2/3@2.png` of size 512x512 pixels and displaying in as higher resolution version of the original. Mapbox JS does this.

**Fake retina / underzoom on client**
In this mode, the client is composing 4 tiles from the next zoom level to create 1 tile. This is a hack and while it works for layers like satellite without lables, it produces unreadably small text for anything else. Leaflet calls this option `detectRetina`.

In Leaflet 0.7 the only way to enable real retina was to use something like: `{z}/{x}/{y}' + (L.Browser.retina ? '@2x' : '') + '.png'` in the tileLayer url.

The problem is that in Leaflet 1.0, leaflet started supporting the `{r}` part, and at the moment these two concepts are mixed, and it's only possible to enable both of them or none-of them. When both are enabled, you get 4 times the tile downloads, as well as the downsides of fake retina's tiny texts.

The problematic lines are:
https://github.com/Leaflet/Leaflet/blob/92e449a0ce7cacedec169c9f24fcddbc985ef707/src/layer/tile/TileLayer.js#L79
combined with
https://github.com/Leaflet/Leaflet/blob/92e449a0ce7cacedec169c9f24fcddbc985ef707/src/layer/tile/TileLayer.js#L28

This is because `r: this.options.detectRetina && L.Browser.retina && this.options.maxZoom > 0 ? '@2x' : ''`

I think there are two options here:
- rename `detectRetina` to `fakeRetina` and use `detectRetina` to actually detect real retina tiles (that is use the `{r}` part if present).
- just use real retina tiles whenever `{r}` is present. That would mean changing the line to simply:
`r: L.Browser.retina ? '@2x' : ''` Probably the better choice, since it'd be backwards compatible with Leaflet 0.7. Users who would like not to enable retina, would just remove the `{r}` part from the url.

mourner:
Hmm you're right! The second approach looks better to me.

yohanboniface:
@hyperknot do you mind opening a pull request with option 2?

hyperknot:
No,  I'm happy to do it!

IvanSanchez:
Fixed in #4110.

PR #4110, opened 2015-12-28
retina url fix

hyperknot:
Fixes the `{r}` part in the tile URL, as in: https://github.com/Leaflet/Leaflet/issues/4103

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -76,7 +76,7 @@ L.TileLayer = L.GridLayer.extend({
 
 	getTileUrl: function (coords) {
 		return L.Util.template(this._url, L.extend({
-			r: this.options.detectRetina && L.Browser.retina && this.options.maxZoom > 0 ? '@2x' : '',
+			r: L.Browser.retina ? '@2x' : '',
 			s: this._getSubdomain(coords),
 			x: coords.x,
 			y: this.options.tms ? this._globalTileRange.max.y - coords.y : coords.y,


pke:
This does not seem to work in v1.0.2. Text is way to small using a mapbox street layer:

```
L.tileLayer("https://api.mapbox.com/styles/v1/mapbox/streets-v10/tiles/{z}/{x}/{y}{r}?access_token={accessToken}", {
      accessToken,
      keepBuffer: 4,
      updateWhenIdle: false,
      updateWhenZooming: false,
    })
```
image

perliedman:
@pke it looks like this is related to that particular style. For example, look at this example that uses some older version of Mapbox's API to get tiles for their outdoors style: http://playground-leaflet.rhcloud.com/dorag/edit?html,output

For me, that text is perfectly readable on a retina screen, while the URL you provided renders tiny, tiny labels. So maybe bring this up with Mapbox?

ghybs:
Hi @pke,

Actually this is reproducible on non-Retina screens as well.
The root cause is the tile size of 512px, but mapping the world as if they were still 256px.

See for example this tile:
`https://api.mapbox.com/styles/v1/mapbox/streets-v10/tiles/15/19160/11042?access_token={accessToken}`

The proper fix is to have Mapbox re-generate those tiles, as @perliedman suggested.

But there is also an easy workaround, see this post on Stack Ovelflow:
[Street labels in Mapbox Tile Layer too small](https://stackoverflow.com/questions/37040494/street-labels-in-mapbox-tile-layer-too-small/37043490#37043490)

The idea is to correctly specify the Tile Size of 512px, but forcing a Tile Layer [`zoomOffset`](http://leafletjs.com/reference-1.0.3.html#tilelayer-zoomoffset) option of `-1` to account for the fact that they still map the world as if they were 256px size:
```javascript
L.tileLayer(mapBoxUrl, {
  attribution: attributionText,
  tileSize: 512,
  zoomOffset: -1
});
```

PR #4113, opened 2015-12-28
prevent undefined property from removing all click events when removed from map

robertleeplummerjr:
If a class does not have a click event, and it is removed, it causes all click events to be removed from map.  This prevents that from happening.

### DIFF
--- a/src/layer/vector/canvas/Path.Canvas.js
+++ b/src/layer/vector/canvas/Path.Canvas.js
@@ -37,7 +37,7 @@ L.Path = (L.Path.SVG && !window.L_PREFER_CANVAS) || !L.Browser.canvas ? L.Path :
 		    .off('moveend', this._updatePath, this);
 
 		if (this.options.clickable) {
-			this._map.off('click', this._onClick, this);
+			this._map.off('click', this._onClick || function() {}, this);
 			this._map.off('mousemove', this._onMouseMove, this);
 		}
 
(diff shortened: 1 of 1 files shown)


robertleeplummerjr:
@mourner That was the fastest I've ever seen someone interact with a pull request.  Are you sure you are a human? :+1:

robertleeplummerjr:
The other way to solve this is to properly set the option, or: `L.Class.prototype._onClick = function () { };`.  The most logical for stability seemed like what was included in the pr.

mourner:
Thanks! Actually that label was added by a bot (waffle.io). I'm mostly off for holidays. :)

robertleeplummerjr:
:+1:

robertleeplummerjr:
Also, waffle.io rocks.

z0mt3c:
Actually i think this change/merge is not correct. According to https://github.com/robertleeplummerjr/Leaflet/commit/3cec7fc1d43f978e1152d939b1f091272472880b i would suggest it should just be renamed to _fireMouseEvent? Your proposed change will allow you to register handlers but will not de-register the listener on removal.

robertleeplummerjr:
The origin: https://github.com/Leaflet/Leaflet/commit/3cec7fc1d43f978e1152d939b1f091272472880b of the above.  @z0mt3c You use "it" a few times here, can you clarify which forks/revisions/libraries you are referring to when you mention them above?

Issue #4020, opened 2015-11-16
map.fitBounds issue with custom projection


MAD-GooZe:
Hello!

I am not quite sure if this is even a bug (or a bug of proj4leaflet library), but anyway.
I am trying to use Leaflet to display GeoJSON of Russia using EPSG:3576 projection. 

I have a problem with fitBounds method - when I try to fit the bounds of GeoJSON layer, map moves to some strange position. 
See the example at http://jsfiddle.net/MAD_GooZe/eomyf5qp/2/
The same also happens with maxBounds option - it limits map to some strange position.
Any ideas how this can be fixed?

IvanSanchez:
@MAD-GooZe Please specify the Leaflet version you're using.

There is a similar issue, #3968. They might (or might not) be related. My guess is that there is a similar bug (a `LatLngBounds` object being calculated with just two and not four projected corners) somewhere in the code.

MAD-GooZe:
I am using v1.0.0-beta.2

perliedman:
@IvanSanchez correct me if I'm wrong, but even using all four corners might not necessarily solve this.

An axis-aligned box in lat-lng space, can end up with sides which are curved when reprojected, which in turn means that the max and min coordinates might not be at the corners, but rather somewhere along one of the sides.

I have no experience with EPSG:3576, but looking at the projected map in @MAD-GooZe's example, this might be the case here: the point with the minimum projected y-coordinate might not be the same as the point with the minimum latitude.

Using four corners is no doubt better than just two, but I don't think it will solve this in the general case.

perliedman:
To get some insight to the complexities involved when transforming an evelope from one projection to another, have a look at GeoTool's code to do this: https://github.com/geotools/geotools/blob/master/modules/library/referencing/src/main/java/org/geotools/referencing/CRS.java#L1392

If this kind of complexity should be handled at all, I think Proj4Leaflet is probably a better place than in Leaflet's core.

MAD-GooZe:
It looks like all bounds related stuff in Leaflet assumes Mercator projection (correct me if I`m wrong).

I made a dirty hack to fit the map to the territory of Russia (based on Leaflet code):
```javascript
function getPointsZoom(map, point1, point2) {
    point1 = L.latLng(point1);
    point2 = L.latLng(point2);
    var zoom = map.getMinZoom(),
        maxZoom = map.getMaxZoom(),
        size = map.getSize(),
        zoomNotFound = true,
        boundsSize;
[...]

Issue #3968, opened 2015-10-27
Custom projection causes visible geojson shape to be clipped in 1.0.0 beta


cgoessen:
Hi,

I've recently moved a leaflet application from 0.7.x to 1.0.0-beta2 (nice performance improvements by the way).

I'm using various custom projections (epsg3035 mainly), in this projection the shape clipping behaves strangely, some countries are clipped while they should still be visible.

I've built a minimal exemple here:
http://plnkr.co/edit/84JBRgw5ZzqzjG5dETE0

The most noticeable glitchy country is Russia.

My guess is that it has to do with the bound checking being done using unprojected coordinates, but that's a wild guess.

I'm not totally sure if it's related to leaflet or proj4leaflet.

IvanSanchez:
Thanks for the reproducible example! It makes debugging these kinds of things easier.

IvanSanchez:
I found this is even easier to reproduce by using just a bounding box for Brazil: 

```js
	var rectangle = L.rectangle([[5.244486, -34.729993],[-33.768378, -73.987235]]).addTo(map);
	map.options.crs = epsg3035;
        map.fitBounds(rectangle.getBounds());
```

In this case, the rectangle disappears when panning vertically. Live demo at http://playground-leaflet.rhcloud.com/pos/edit?html,output

I'm not sure whether that bbox should show as an actual rectangle in the map's CRS.

IvanSanchez:
I've tracked the bug down to `L.Polyline._project`, [here](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Polyline.js#L140).

That code assumes a rectangular projection, and uses only two corners from the bounding box. For non-rectangular projections, all four corners must be projected into pixel coordinates, and *then* expand the pixel bounds by the tolerance.

perliedman:
Some testing reveals that using four points doesn't really solve this, at least not with this naïve approach:

```js
if (this._bounds.isValid()) {
	var map = this._map,
	    bounds = this._bounds,
	    pxBounds = new L.bounds([
	    	map.latLngToLayerPoint(bounds.getSouthWest()),
	    	map.latLngToLayerPoint(bounds.getNorthEast()),
	    	map.latLngToLayerPoint(bounds.getSouthEast()),
	    	map.latLngToLayerPoint(bounds.getNorthWest())
	    ]);

	pxBounds.min._subtract(p);
	pxBounds.max._add(p);

	this._pxBounds = pxBounds;
}
[...]

PR #4114, opened 2015-12-28
Use all projected coords when calculating Polyline's pixel bounds.

perliedman:
Instead of using the `LatLngBounds`' two corners to calculate the pixel bounds of a `Polyline` or `Polygon`, this PR uses all projected vertices to calculate the bounds.

This fixes problems with projections where the x and y axis aren't (mostly) aligned with longitudes and latitudes, and in particular where longitudal lines are projected into curves, like in #3968 and #4020.

I have made a very small performance check by timing the handling of the `zoomend` event both with and without this fix together with the code in #3968, and couldn't measure any significant difference. It might show for polys with more vertices, though.

An alternative approach would be to let the CRS implement this logic, so that the default could still be the simplified handling, but I had a hard time coming up with an API for that that felt natural.

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -129,23 +129,22 @@ L.Polyline = L.Path.extend({
 	},
 
 	_project: function () {
+		var pxBounds = new L.Bounds();
 		this._rings = [];
-		this._projectLatlngs(this._latlngs, this._rings);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Looks good to me. I guess the performance hit can be addressed later if needed, by caching bounds.

mourner:
:+1: looks good to me, thanks Per!

PR #4116, opened 2015-12-29
update reference for layerGroup methods

loorinm:
Added documentation for an additional use for these methods which was not listed.

### DIFF
--- a/reference.html
+++ b/reference.html
@@ -3101,7 +3101,7 @@ map.fitBounds(bounds);</code></pre>
 		</code></td>
 
 		<td><code><span class="hljs-keyword">this</span></code></td>
-		<td>Adds a given layer to the group.</td>
+		<td>Adds a given layer to the group, or adds a given group to the map.</td>
 	</tr>
 	<tr>
 		<td><code><b>removeLayer</b>(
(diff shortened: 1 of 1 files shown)


mourner:
Hmm, I'm not sure — does it really do that? `group.addLayer(layer)` only adds a layer to the group. It only adds that layer to the map if the group was added to the map before.

loorinm:
I'm using this code to repeatedly add and remove layerGroups from the map:

    //adjust the markers on the page accordingly
    if (tone.status) {
      //show all the layers in the group
      $scope.map.addLayer($scope.layerGroups[tone.name])
    } else {
      //hide all the layers in the group
      $scope.map.removeLayer($scope.layerGroups[tone.name])
    }

here's my app:
tweetmoodmap.herokuapp.com

hyperknot:
@loorinm I think you are mistaking `map.addLayer` with `layerGroup.addLayer`. They are both called the same, but their documentation is in different part of the page.

Look:
http://leafletjs.com/reference.html#map-addlayer
and
http://leafletjs.com/reference.html#layergroup-addlayer

loorinm:
Ahh, I see now. Thanks for the clarification!

PR #4122, opened 2015-12-31
fix for safari css syntax warning

jgravois:
fix for a warning flagged by Safari `9.0.2`
unexpected token

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -487,7 +487,7 @@
 	margin: 0 auto;
 
 	-ms-filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678)";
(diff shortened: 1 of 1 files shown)


mourner:
Are you sure this doesn't have any effect on IE compatibility?

hyperknot:
Just looked it up, this page says it's not valid:
https://msdn.microsoft.com/en-us/library/ms532847(v=vs.85).aspx

If dropping IE 7 support is an option though and you can remove that line altogether.

mourner:
Dropping IE7 support is a bigger issue to discuss, but while we still support it, I think it's fine to keep Safari warnings, they do no harm.

jgravois:
thanks for the history lesson @hyperknot.  i had no idea that `ie7` expects/parses [invalid css](https://github.com/less/less.js/issues/734).

PR #4124, opened 2016-01-03
Optimize PNG images with ZopfliPNG and pngquant

vtduncan:
Leaflet is already very lightweight, but I noticed that the PNG images could be made smaller. Using a combination of ZopfliPNG and pngquant (for the two marker images), I was able to reduce the total size from 11.0 KB to 6.6 KB.

ZopfliPNG is lossless, but pngquant does alter the images slightly. I wasn't able to tell the difference with my eyes, but I guess you can use GitHub's fancy image comparison tool to check it yourself!

Thanks for working on Leaflet!


mourner:
Great, thanks! I can't see the difference as well.

danzel:
Awesome! Thanks.

Issue #4127, opened 2016-01-06
Map 'dblclick' event not firing on Chrome


delner:
Using *v1.0.0-beta.2* from the CDN, I'm having trouble getting `dblclick` events from the map to fire in Chrome.

Setup a map as follows:

    this.activeLayer = L.tileLayer(tileHost + '/tiles/' + mapName + '/{z}/{x}/{y}.png');
    this.map = L.map('map', {
      maxZoom: 7,
      doubleClickZoom: false,
      bounceAtZoomLimits: false,
      layers: [this.activeLayer]
    });

Then binding 3 events to test with:

    this.map.on('click', function (eventData) { console.log('map click!'); });
    this.map.on('dblclick', function (eventData) { console.log('map double click!'); });
    $("body").on('dblclick', function (eventData) { console.log('body double click!'); });

Double clicking on the map in:

 - *Firefox 39.0* yields `map double click`, `body double click!` and `map click!`x2.
 - *Chrome 47.0.2526.106 64-bit* yields only `body double click!` and `map click!`x2.

Switching back to 0.7.7 circumvents the issue.

1.0.0beta2 seemed to work ok in Chrome before, not sure why it isn't working now. Is anyone else seeing similar behavior?

yohanboniface:
Works for me on *Chromium Version 47.0.2526.73 Ubuntu 15.10 (64-bit)* and *Chrome Version 47.0.2526.106 (64-bit)* (Ubuntu).

Can you check here to be sure (latest master): http://playground-leaflet.rhcloud.com/coh/edit?html,console,output

IvanSanchez:
Also works as expected in my Chromium 47.0.2526.80 on Debian.

delner:
@yohanboniface Still no luck in *Chrome 47.0.2526.106 64-bit on Ubuntu* using the playground you linked: here are [the results](http://i.imgur.com/Wjcat1q.png) from my double click. Works fine in *Firefox 39.0 on Ubuntu*.

yohanboniface:
Very weird. Any Chrome plugin that could explain?

delner:
Not that I can think of: I run vanilla web browsers (no ad blockers, JS stuff, etc.) The only extensions I have installed are for Google Docs, and those come default I think.

yohanboniface:
Just for investigation, can you try with Chromium?

delner:
[...]

Issue #2820, opened 2014-07-30
addDoubleTapListener slowdown in IE Touch


nkovacs:
`L.Marker._initInteraction` and `L.Path._initEvents` register a dblclick event. This causes addDoubleTapListener to add an event listener **to document.documentElement**. Because the event listener is bound to the document element, it's not cleaned up when the element is removed. This is a memory leak, and it causes massive slowdowns on IE Touch (Windows Phone 8.1 and Windows 8.1 on a touchscreen tablet with desktop IE 11, L.Browser.touch and L.Browser.pointer are both true).

Here's an example demonstrating the problem:
https://gist.github.com/nkovacs/543e2cc4cf4ac90e1236
Every time you click add, 100 markers are added to the map, and 100 events to the document. It doesn't matter if you remove the markers or not, the adds will be slower and slower (several seconds after a few thousand markers).

[This patch for 0.7.3](https://github.com/nkovacs/Leaflet/commit/56cccf5a7cd04f55a3787abd894238ccd6cbe900) removes the listener when you remove a marker or SVG path. However, addDoubleTapListener is still very slow in itself, so it's not a solution if you want to have several hundred markers or paths on the map.

PR #4131, opened 2016-01-07
Possible fix for #4127, assume dblclick DOM event exists in touch-capable browsers

IvanSanchez:


mourner:
The only issue with this is that it may make the performance problems on IE+touch even worse by adding even more event handlers to everything...

IvanSanchez:
@mourner Skipping the native `dblclick` on modern IEs/edges should be easy enough, gimme a few minutes.

IvanSanchez:
Interesting. I cannot reproduce #2820 anymore, so I guess skipping the native `dblclick` is not needed after all...

kohenkatz:
Would it be possible to rebase this PR on top of the changes from #4143, to make testing easier?

yohanboniface:
Not very comfortable with the workardound, but I've no plan B, it's and edge (haha) case, and we need to move on, so merging.

PR #4134, opened 2016-01-10
add meteor-leaflet to frameworks

bevanhunt:
Adds meteor-leaflet to leaflet frameworks. Ref: https://github.com/bevanhunt/meteor-leaflet/issues/62

I think Atom removed some trailing spaces as well.


IvanSanchez:
Removing some trailing whitespace shouldn't be a problem :-)

PR #4148, opened 2016-01-15
zoom to integer levels

hyperknot:
Round target zoom levels for zoom controls, like it's done at DoubleClickZoom already:

https://github.com/Leaflet/Leaflet/blob/161172bbb92e60d2d7fc21173da7e58fc725dd2b/src/map/handler/Map.DoubleClickZoom.js#L21

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -72,11 +72,11 @@ L.Map = L.Evented.extend({
 	},
 
 	zoomIn: function (delta, options) {
-		return this.setZoom(this._zoom + (delta || 1), options);
+		return this.setZoom(Math.floor(this._zoom) + (delta || 1), options);
 	},
 
 	zoomOut: function (delta, options) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
No. See #3523 - with fractional zoom, it's not about rounding anymore, it's about snapping to a number defined as a map option (default 1).

@hyperknot I would appreciate if you could review that PR, though.

hyperknot:
It'll take some time to read #3523 and the linked issues in that one, but as of now, there is inconsistent behaviour between DoubleClickZoom (13.5 -> 14 -> 15) and zoom controls (13.5 -> 14.5 -> 15.5).

IvanSanchez:
@hyperknot I see your point, but I'd rather work on #3523 instead of patching this up and creating a merge conflict :-/

hyperknot:
@IvanSanchez I agree, I'll review that one.

PR #4150, opened 2016-01-15
different crs for base layer and overlay

alekzonder:
here live example: http://jsfiddle.net/alekzonder/qxdxqsm3/

In leaflet Map class add crs param for project, unproject, and _getNewPixelOrigin methods

### in fiddle

gis layer in EPSG3857 crs
yandex layer in EPSG3395 crs
yandex traffic layer in EPSG3395 crs

by default shown gis layer (EPSG3857) and overlay traffic layer in EPSG3395

create custom L.TileLayerCustomCrs extends from L.TileLayer and use modified project, unproject, and _getNewPixelOrigin methods

change layers from gis to yandex, toggle traffic layer

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -411,14 +411,16 @@ L.Map = L.Evented.extend({
 
 	// conversion methods
 
-	project: function (latlng, zoom) { // (LatLng[, Number]) -> Point
+	project: function (latlng, zoom, crs) { // (LatLng[, Number][, L.CRS]) -> Point
+		crs = !crs ? this.options.crs : crs;
 		zoom = zoom === undefined ? this._zoom : zoom;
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm going to say "no".

Having EPSG:3395 and EPSG:3857 in the same map looks like a good idea, but mixing CRSs doesn't always match (specially in raster tiles). See your fiddle near California, for example. And I don't want to imagine mixing two coordinate systems which don't share the same map projection (mercator, in this case).

I suggest you create a `L.TileLayer.CustomCRS` plugin, and **redefine `L.Map.project` there** (just overwrite the definition of `L.Map.prototype.project` *et al*). Redefining `L.Map.project` in the Leaflet core just for the edge case of two spherical projections is more trouble than it's worth, IMHO.

alekzonder:
ok, thanks

PR #4155, opened 2016-01-18
GeoJSON set default style to Point and MultiPoint

karru:
Now you can use MultiPoint and Point with Layers GeoJSON with default styles like customize icon


mourner:
Not sure about this. You can do the same by passing a `pointToLayer` function, do we need to extend the API like this?

jieter:
Yeah, tomorrow someone wants to add popups automatically. I think the current default for Point is a marker is just right.

karru:
This is not mandatory as it is only optional. Of course if you want you can use 'pointToLayer' but in the others types of GeoJson you have the posibility to use defaults options. Perhaps is better modify https://github.com/tombatossals/angular-leaflet-directive...

karru:
Ok! Is true! I can use a pointToLayer function. Thanks a lot.

PR #4156, opened 2016-01-18
Partial revert of e2fbe196, fixes #4145

### LINKED ISSUE #4145: Fractional zoom tiles not loading at map edges
Just made an example for the 1px fraction tiles bug and realised in master, the tiles near the map edges do not load:
http://jsbin.com/ratoli/5/edit?html,js,output

In beta 2 they load OK.

IvanSanchez:


src/layer/tile/GridLayer.js
```diff
@@ -444,8 +444,9 @@ L.GridLayer = L.Layer.extend({
 
 	_getTiledPixelBounds: function (center) {
 		var map = this._map,
+		    scale = map.getZoomScale(map.getZoom(), this._tileZoom),
 		    pixelCenter = map.project(center, this._tileZoom).floor(),
-		    halfSize = map.getSize().divideBy(2);
+		    halfSize = map.getSize().divideBy(scale * 2);
```
yohanboniface: Scaling the size does not make sense when you think about it: the map *size* does not change from one zoom to another. :)

So what this revert do is make Leaflet request too much tiles (like 4 times too much in some round cases), so it seems to be fixing the issue but it's not ;)

PR #4158, opened 2016-01-18
Rant about transitionend

IvanSanchez:


mourner:
Classic Ivan :trollface: Per chat, this fun post would look better on Medium. We'll keep the Leaflet blog free of rants.

PR #4160, opened 2016-01-19
Disable event defaults when disabling scroll propagation

### LINKED ISSUE #4159: Scroll-zooming in Safari can move page position
Both in 0.7 and in 1.0, scrolling can move the page scroll in Safari. It's even visible on http://leafletjs.com, you can scroll down to the "Trusted by the best" part, by just doing a big scroll inside the Leaflet map. It happens 1 out of 5 times, but when a big jump happens it's quite shocking.

Made an example for trying the different versions: 
http://playground-leaflet.rhcloud.com/wey/1/edit?html,output

Usually it's just 10-20 px, but can be a big jump as well. 

Safari 9.0.2, OS X 10.9.5

IvanSanchez:
Fixes #4159

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -139,7 +139,7 @@ L.DomEvent = {
 	},
 
 	disableScrollPropagation: function (el) {
-		return L.DomEvent.on(el, 'mousewheel', L.DomEvent.stopPropagation);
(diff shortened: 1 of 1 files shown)


mourner:
Stop. Will this change make it impossible to scroll a scrollable popup with maxHeight?

IvanSanchez:
Yes, yes it does :-(

PR #4167, opened 2016-01-21
Update plugin guide to prevent some new common pitfalls

IvanSanchez:
I've seen a few developers screwing up the camelCase of the factories and committing their built/minified files to their repos, and I want to prevent that from happening in the future.

### DIFF
--- a/PLUGIN-GUIDE.md
+++ b/PLUGIN-GUIDE.md
@@ -17,6 +17,7 @@ This guide lists a number of best practices for publishing a Leaflet plugin that
 	- [Plugin API](#plugin-api)
 3. [Publishing on NPM](#publishing-on-npm)
 4. [Module Loaders](#module-loaders)
+5. [Adding to the plugins list](#adding-to-the-plugins-list)
 
 ## Presentation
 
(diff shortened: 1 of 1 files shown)


PLUGIN-GUIDE.md
```diff
@@ -152,6 +154,21 @@ Here is an example of a `package.json` file for a Leaflet plugin.
...
+If possible, do not commit your minified files (e.g. `dist`) to a repo; this can
+lead to confussion when trying to debug the wrong file. Instead, let the user 
+run the build/minify scripts automatically when they run `npm install`, for example:
+
+```json
+{
+  "name": "my-leaflet-plugin",
+  ...
+  "scripts": {
+    "postinstall": "grunt build"
```
mourner: Actually the current best practice is to build in `prepublish` rather than run build on user install, so that the files get published to NPM and are available on install without devDependencies immediately.

mourner:
Otherwise, :+1:

PLUGIN-GUIDE.md
IvanSanchez: Hah, didn't know this (gotta update a couple of my plugins). Let me amend the commit.

PR #4174, opened 2016-01-27
update mapbox token

lyzidiamond:
This PR updates the Mapbox access token across the Leaflet site. Once this PR is merged, the previous token will be deleted. cc @mourner


mourner:
Thanks :)

PR #4175, opened 2016-01-27
Layer early events hanging up fix

ilyapro:
onAdd handler can changes '_zoomAnimated' property and 'getEvents' 'zoomanim' depends on this property. So it makes sense to call 'onAdd' first, then 'getEvents'. I got error in my project based on this wrong order and this PR is fixing it.


IvanSanchez:
This is basically a revert of #3806.

@ilyapro I must ask you to check if #3745 is reproducible after your changes. Otherwise, can you provide a better explanation of the problem your code has with the current event ordering? It might be a good idea to see if it has a greater impact than #3806.

Issue #3431, opened 2015-05-05
tilerLayer.setOpacity not changing active tiles on master


jplante:
Is anybody else having a problem setting the opacity on a tileLayer with the latest master build?
Calling `.setOpacity(x)` doesn't seem to do anything.  Maybe there is an issue with the `._updateOpacity` method check perhaps?  Sample page and code reference are listed below.

**Simple Sample Page**
The page has the latest zip from master set locally.  It simply creates a map, a tile layer and then attempts to change its opacity. This happens even on layers higher in the stack.  No tile layers seem to change their opacity.
```
<html><head><title>Sample Page</title>

<-- Pulled from latest master download -->
<link rel="stylesheet" href="/stylesheets/leaflet-0.8.css" />
<script type="text/javascript" src="/javascripts/leaflet-0.8.debug.js"></script>

<script type="text/javascript">
	var map = L.map('map').setView([51.508, -0.11], 3);
	var earth = L.tileLayer('http://otile{s}.mqcdn.com/tiles/1.0.0/osm/{z}/{x}/{y}.png', {
		subdomains: ["1", "2", "3", "4"]
	}).addTo(map);
	
	// ** This does not seem to have ANY effect  **
	earth.setOpacity(0.5);  
</script>

</head>
<body>
	<div id="map"></div>
</body>
</html>
```

**Code Reference**
I'm just poking around, so I'm not really sure what the issue is, or if anyone else can reproduce it.  Stepping through however, it looks like `tile.active` is always true and then just continues instead of changing the opacity.
```
_updateOpacity: function () {
	...
	for (var key in this._tiles) {
		var tile = this._tiles[key];

		// ** THIS ALWAYS SAYS tile.active = TRUE AND CONTINUES **
		if (!tile.current || !tile.loaded || tile.active) { continue; }  

		var fade = Math.min(1, (now - tile.loaded) / 200);
		if (fade < 1) {
			L.DomUtil.setOpacity(tile.el, opacity * fade);
					nextFrame = true;
				} else {
					L.DomUtil.setOpacity(tile.el, opacity);
					tile.active = true;
					this._pruneTiles();
				}
			}
```

jplante:
Further investigation seems to indicate that it is in fact tied to tiles that are currently showing in the viewport.  panning away reveals tiles whose opacity has changed, but they are not consistent with the tiles that were visible when the call was made, and those visible tiles should have changed too.

jplante:
[...]

PR #4192, opened 2016-02-04
Prune active tiles, fixes #3765

### LINKED ISSUE #3765: Tiles are sometimes retained when they shouldn't
Using Leaflet 1.0-dev (46d2d6a)

When I zoom in, the old (more zoomed out) layers do not disappear.   You have to drag or cause some other even for the layer to properly refresh.  See the attached image.  My hack to fix this is to add map.panBy([1, 1]);  to zoomend - not ideal.  Are there work arounds, or a fix in the works?   

Not sure if it matters, but the map div is downstream/inside a reactjs object - there might be some dom manipulation that's screwing things up...

screen shot 2015-08-21 at 4 13 27 pm

jblarsen:
Please see the discussion here:

https://github.com/Leaflet/Leaflet/issues/3765

This fixes the issue. But please also see:

https://github.com/Leaflet/Leaflet/issues/3431

https://github.com/Leaflet/Leaflet/pull/3513/

where the if statement was introduced. I am not sure whether my pull request will reintroduce that bug or if it will have performance implications for the use case with an opacity slider.

### DIFF
--- a/debug/map/zoom-delta.html
+++ b/debug/map/zoom-delta.html
@@ -1,0 +1,109 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<meta name="viewport" content="initial-scale=1.0" />
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<link rel="stylesheet" href="../css/mobile.css" />
+
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+	<style>
+	.container {
+		float:left; width: 600px; height: 600px;
+		position: relative;
+		border: 1px solid gray;
+	}
+	#map1, #map2 {
+		position:absolute;
+		top:2em;
+		bottom:2em;
+		left:0;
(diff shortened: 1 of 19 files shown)


mourner:
>  I am not sure whether my pull request will reintroduce that bug or if it will have performance implications for the use case with an opacity slider.

Can you check this?

jblarsen:
> Can you check this?

I have used the sample page provided by jplante in issue #3431 to test that the bug is not reintroduced:

Leaflet master:
http://playground-leaflet.rhcloud.com/cuyi/edit?html,console,output

Code with fix:
http://playground-leaflet.rhcloud.com/bod/edit?html,console,output

Please note that the links will not work forever since the js and css files are not uploaded to a unique and persistent URL.

I will try to make two examples with an opacity slider to see what happens with the two versions.

jblarsen:
I have made a version in which you can check out how many calls are made to the _pruneTiles method when you change the opacity:

http://playground-leaflet.rhcloud.com/kom/1/edit?html,console,output

It is only called once per opacity change (which it should). This is the same as without the change I suggest.

PR #4193, opened 2016-02-04
Fix extra tile usage

### LINKED ISSUE #4180: Fixing 4x tile usage
Leaflet 1.0 beta has 4x tile usage compared to 0.7, when zooming in (and other scenarios involving fractional zoom levels).

@yohanboniface fixed it in: https://github.com/Leaflet/Leaflet/pull/4040
however this broke fractional zoom levels at the edge of the map: https://github.com/Leaflet/Leaflet/issues/4145

@IvanSanchez fixed it in https://github.com/Leaflet/Leaflet/pull/4156
however it raised the tile usage again to 4x.

I made a test for it (based on @yohanboniface 's test):

- 0.7.5 (only using browser tools): 8 tiles at load, +6 tiles at zoom: http://playground-leaflet.rhcloud.com/puh/1/edit?js,output
- 1.0 beta 2: 8 tiles at load, +24 tiles at zoom: http://playground-leaflet.rhcloud.com/jet/1/edit?js,console,output
- 1.0 just before #4156: 8, +6, http://playground-leaflet.rhcloud.com/ciz/1/edit?js,console,output
- 1.0 after #4156: 8, +24 http://playground-leaflet.rhcloud.com/biq/1/edit?js,console,output

IvanSanchez:
Should fix #4180

Added a new `viewprereset` map event, run before the `viewreset` event.

So the sequence for when `fadeAnimation` is false is: unload tiles on `viewprereset`, then move the map to the proper viewport, then load tiles on `viewreset`. This makes PhantomJS a viable option for unit-testing tile loading counts.

Unit tests in graphical browers do fail as of now, though.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -667,7 +667,14 @@ L.GridLayer = L.Layer.extend({
 		if (this._noTilesToLoad()) {
 			this._loading = false;
 			this.fire('load');
-			this._pruneTiles();
+
+			if (L.Browser.ielt9 || !this._map._fadeAnimated) {
+				L.Util.requestAnimFrame(this._pruneTiles, this);
+			} else {
+				// Wait a bit more than 0.2 secs (the duration of the tile fade-in)
+				// to trigger a pruning.
+				setTimeout(L.bind(this._pruneTiles, this), 250);
+			}
 		}
 	},
 
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -654,7 +654,7 @@ L.GridLayer = L.Layer.extend({
 			this._fadeFrame = L.Util.requestAnimFrame(this._updateOpacity, this);
 		} else {
 			tile.active = true;
-// 			this._pruneTiles();
+			this._pruneTiles();
 		}
 
 		L.DomUtil.addClass(tile.el, 'leaflet-tile-loaded');
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -445,7 +445,8 @@ L.GridLayer = L.Layer.extend({
 
 	_getTiledPixelBounds: function (center) {
 		var map = this._map,
-		    scale = map.getZoomScale(map.getZoom(), this._tileZoom),
+		    mapZoom = map._animatingZoom ? Math.max(map._animateToZoom, map.getZoom()) : map.getZoom(),
+		    scale = map.getZoomScale(mapZoom, this._tileZoom),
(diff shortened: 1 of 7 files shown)


src/layer/tile/GridLayer.js
```diff
@@ -444,7 +445,8 @@ L.GridLayer = L.Layer.extend({
 
 	_getTiledPixelBounds: function (center) {
 		var map = this._map,
-		    scale = map.getZoomScale(map.getZoom(), this._tileZoom),
+		    mapZoom = map._animatingZoom ? Math.max(map._animateToZoom, map.getZoom()) : map.getZoom(),
+		    scale = map.getZoomScale(mapZoom, this._tileZoom),
```
yohanboniface: Sorry to insist a bit, but as per my understanding we don't need scale: if the map size is 600px at zoom 10 it will still be 600px at zoom 11 or 17 or 17.358, isn't it? Or am I missing something?

IvanSanchez: @yohanboniface You're missing fractional zoom. In those cases, 0.5 < scale < 2. See the `debug/tests/tile-bounds.html` file.

mourner:
> Added a new viewprereset map event, run before the viewreset event.

That's just asking for double brainmelt in future. :)

IvanSanchez:
> That's just asking for double brainmelt in future. :)

@mourner Actually, I don't think so. Layers should clear everything up on `viewprereset`, then add everything back again in `viewreset`. Makes sense.

God, I want leafdoc merged in so I can document these kind of things inline right away.

yohanboniface:
> I want leafdoc merged in so I can document these kind of things inline right away.

Why are we not there yet?

IvanSanchez:
I think this PR introduces some kind of race condition on `flyTo` animations that makes tiles to not be pruned when switching levels, likely due to https://github.com/Leaflet/Leaflet/pull/4193/files#diff-f1e6be67599c594731fff6191c710420L655 - must find a way to trigger a pruning once in a while.

hyperknot:
Some debugging on tile-events.html:

1. The load count for TileLayers is not reflecting the actual Img requests from the browser tools. In my case: grid = 20, postiron = 40, img requests = 20
2. It is happening when grid is added to map. If `map.addLayer(grid);` is commented out the values are almost correct (see point 3.)
3. tileloadstart does not fire for initial loading of tileLayer. Adding setTimeout to gridLayer fixes tileLayer.
```
   setTimeout(function(){
            this.fire('tileloadstart', {
                tile: tile,
                coords: coords
            });
        }.bind(this));
```

hyperknot:
#4145 also OK with this one:
http://jsbin.com/ratoli/9/edit?html,js,output

PR #4197, opened 2016-02-05
Run tests in SlimerJS also.

IvanSanchez:
This PR changes what `npm test` does. Instead of using just PhantomJS (headless webkit-based browser), it will _try_ to use SlimerJS (almost-headless gecko-based browser).

SlimerJS is available as a NPM package, so running `npm install` should install all the SlimerJS stuff (including karma connectors and whatnot).

The main advantage is being able to run tests involving CSS animations (and even more complex stuff like WebGL) in a quasi-headless, automated way.

The main disadvantage is that SlimerJS might not be readily available or usable in Win platforms, or when SSHing into some machine due to the need for a graphical environment :-/

I would like some MacOSX user to try this, and see if a simple `npm install` then `npm test` runs the tests in both (quasi-)headless browsers.


mourner:
I can't install `karma-slimerjs-launcher` for some reason, it says:

```bash
Error extracting /Users/mourner/.npm/slimerjs/0.9.5/package.tgz archive: ENOENT: no such file or directory, open '/Users/mourner/.npm/slimerjs/0.9.5/package.tgz'
```

Although the Slimer version installed is 0.9.6.

mourner:
OK got it — the solution was to install with NPM 2 instead of NPM 3, lol. :) Tests run just fine locally. Should we check if they're fine on Windows too?

IvanSanchez:
I cannot check on windows (my Win VMs only have IE & Edge, no node tools), and I'm afraid that my workaround for https://github.com/karma-runner/karma-slimerjs-launcher/issues/1 might not kill the dangling slimerjs process afterwards.

As long as the tests don't create more problem in Win, I'll be happy.

danzel:
Trying to test on windows, ran in to the same npm error as above. Bit dumb that we can't install with npm 3 now :(
Will try get npm 2 going

build/build.js
```diff
@@ -198,6 +211,19 @@ exports.test = function(complete, fail) {
 	console.log('Running tests...');
 
 	var server = new karma.Server(testConfig, function(exitCode) {
+
+		// Work around https://github.com/karma-runner/karma-slimerjs-launcher/issues/1
+		if (autoSlimer) {	// Only if we were able to run slimerjs and killall
+			var slimerjsPids = require('child_process').execSync('ps -Af | grep slimerjs | grep xulrunner | awk \'{print $2}\'').toString();
```
danzel: This won't work on windows.

build/build.js
```diff
@@ -198,6 +211,19 @@ exports.test = function(complete, fail) {
...
+			slimerjsPids = slimerjsPids.trim().split('\n');
+			for (var i=0; i<slimerjsPids.length; i++) {
+				try {
+					var pid = Number(slimerjsPids[i]);
+					require('child_process').execSync('ps -A | grep ' + pid + ' && kill ' + pid);
```
danzel: This won't work on windows.

danzel:
Here is the console output.
The slimerjs window appears and disappears eventually, maybe disable it on windows, or rewrite the lines  that use the command line bits.

image

IvanSanchez:
@danzel What's the value of [`os.platform()`](https://nodejs.org/api/os.html#os_os_platform) in your nodejs environment (`win32`? `linux` because of cygwin?). It should be easy to add an extra check for those.

build/build.js
IvanSanchez: I was kinda expecting that the

```js
   var child = require('child_process').execFileSync(   require('slimerjs').path  );
```

would throw an error in win32 and skip PhantomJS together.

danzel:
```
console.log(require('os').platform())
win32
```

IvanSanchez:
@danzel https://github.com/Leaflet/Leaflet/commit/e8e6dae652107510146e5fc5e61ebb9342cba8ba should help, but I'm not sure if the change is too naïve.

danzel:
That stopped that error, but the slimerjs tests still don't work.
The window appears, we run 0 tests and eventually it times out and continues (and the window closes)

image

We could disable slimer on windows for now? Or maybe I could take a look later in the week if we're lucky.

mourner:
Should we revisit this or close?

cherniavskii:
#5845 #5831 will run tests in Firefox and Chrome in Travis CI. So it looks like supporting SlimmerJS doesn't make a lot of sense now.
Feel free to reopen it if I'm missing something.

PR #4203, opened 2016-02-08
Replace 1.0 reference with Leafdoc-generated file

IvanSanchez:
Related to #3916 AKA "🍂doc"

Replaces the `reference-1.0.0.html` file with a 🍂doc-generated one. The process is not fully automatic: a dev has to manually change the header (and `</html>`) of the 🍂doc-generated page to fit the jekyll includes.

### DIFF
--- a/reference-1.0.0.html
+++ b/reference-1.0.0.html
@@ -9,2725 +9,6081 @@ bodyclass: api-page
 <p>This reference reflects <strong>Leaflet 1.0</strong>.</p>
 
 <p>Docs for 0.7 are <a href='reference.html'>available here</a>.
+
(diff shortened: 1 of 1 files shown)


mourner:
Any chance we could solve the 3x size increase before doing the replace?

IvanSanchez:
@mourner I don't think there's a *real* chance for that. I'm fine with a bigger size for the docs and waiting less for the release.

perliedman:
Since this is not the _stable_ docs yet, I think it's fine even if you could argue they're unnecessarily large.

Hopefully we'll have time to look into it before 1.0.0, but even if we haven't, they're not unacceptably large.

PR #4212, opened 2016-02-09
Branch L.Renderer.updateTransform when not L.Browser.any3d, fixes #4211

### LINKED ISSUE #4211: Doubled-up transforms on SVG layers in non-3D environments (e.g. PhantomJS)
In the 1.0 beta, SVG layers get both a transform and a top/left position applied when running in a non-3D browser, effectively doubling their offset and rendering them visually out-of-sync with the rest of the map. I ran into this while trying to do some snapshotting in PhantomJS (where 3D is turned off):

output
*GeoJSON/SVG layer of bus routes overlaid on SF city streets*

This is happening because SVG layers update their viewbox and call `DomUtil.setPosition()` in their [`_update()` method](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/SVG.js#L47-L48), but also inherit an [`_updateTransform()` method from `L.Renderer`](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Renderer.js#L66) that explicitly calls `DomUtil.setTransform()`. When 3D is on, they both wind up setting a transform on the SVG (which is fine; one overrides the other), but when 3D is off, as in PhantomJS, one sets a transform and other sets the top/left properties, resulting in a doubled-up offset.

I don't know the exact history of how this situation came to be or how these two parts are intended to work (or not intended!) together, so I'm not sure on the best approach for a fix. Either way, though, these two methods seem to be in conflict.

IvanSanchez:
I'm not 100% sure this is the right fix - maybe a new `L.DomUtil.setPositionScale()` function could be added, which then can be statically assigned to either `L.DomUtil.setTransform()` or `L.DomUtil.setPosition()`, but maybe that's premature optimization.

### DIFF
--- a/src/layer/vector/Renderer.js
+++ b/src/layer/vector/Renderer.js
@@ -63,7 +63,11 @@ L.Renderer = L.Layer.extend({
 
 		    topLeftOffset = viewHalf.multiplyBy(-scale).add(position).add(viewHalf).subtract(centerOffset);
 
-		L.DomUtil.setTransform(this._container, topLeftOffset, scale);
+		if (L.Browser.any3d) {
+			L.DomUtil.setTransform(this._container, topLeftOffset, scale);
+		} else {
+			L.DomUtil.setPosition(this._container, topLeftOffset);
+		}
 	},
 
 	_reset: function () {


Mr0grog:
Yes, this works for me. FWIW, my hacky temporary solution was:

```js
L.DomUtil.setPosition(this._container, topLeftOffset);
L.DomUtil.setTransform(this._container, null, scale);
```

That way, the scale is still getting set, since I presume that could be pretty essential in some cases (haven't dug around nearly enough to be sure about that, though).

PR #4214, opened 2016-02-09
Better cleanup of L.Control.Layers, fixes #4213 (plus unit tests)

### LINKED ISSUE #4213: L.Control.Layers does not cleanup event listeners when removed
When a layer control is removed from a map using `removeControl`, it does not remove the event listeners that it has attached to its layers.

This causes an exception if one these layers are later added or removed from the map, since it will trigger the layer control's listener, which assumes `this._map` to be defined, but it isn't.

I think the control should remove all listeners when it is removed from the map.

Example to reproduce: http://playground-leaflet.rhcloud.com/gic/edit?html,output

Tested with 1.0.0-beta.2.

IvanSanchez:


perliedman:
:+1: Works as expected, removing and readding the same control multiple times works without problems.

PR #4219, opened 2016-02-10
Added leaflet-tracksymbol to plugin page.

lethexa:
I would be very happy if you add my leaflet plugin link to the plugin page.


IvanSanchez:
Looks good! A couple of comments, though:

- If your plugin creates a new kind of `L.Marker`, it should define a `L.Marker.TrackMarker` class, as per the Leaflet coding conventions
- It would help a lot if you could provide a live example of the plugin in action, or at least a screenshot

Issue #4215, opened 2016-02-09
Right click touchpad action zooms out map


hyperknot:
Since a1ff60bf, a right click on the touchpad (Macbook, OS X), triggers a zoom out. 

I believe it is because of ths `Math.max(d3, this._map.options.zoomSnap || 0)` line. I think it should be a round to the nearest zoomSnap, not a max.

Can be tested in every sample / example.

IvanSanchez:
Does this mean that right-clicking the touchpad triggers a `wheel` event??

I've seen problems with horizontal scroll touchpads, but I'm not aware this was possible also :-S

hyperknot:
Yes, I've just tested it, it triggers a 0.

hyperknot:
Here is a comparison between the old function (zigzag) and the new one (smooth) on -1000...1000 delta range. I *think* what you wanted originally is a smoother zig-zag, isn't it? Also the old one returned 0 -> 0, the new one returns 0 -> 1, hence the bug.

![figure_1](https://cloud.githubusercontent.com/assets/494223/12993193/0561a4fa-d119-11e5-98ee-f1c460aaeb2e.png)

hyperknot:
Fixed in #4224, gives smaller zig-zag with zoomSnap and identical to old version when zoomSnap = 1.

![figure_1](https://cloud.githubusercontent.com/assets/494223/12995257/70f1ac30-d126-11e5-9c71-8cbb4293ab76.png)

Gives smooth curve with zoomSnap = 0.
![figure_1](https://cloud.githubusercontent.com/assets/494223/12995278/a1f8ca66-d126-11e5-81bc-da8662ae28b4.png)

(green is old version)

hyperknot:
Fixed as of https://github.com/Leaflet/Leaflet/pull/4224

PR #4224, opened 2016-02-11
fix fractional zoom calculation

hyperknot:
Uses correct fractional ceiling calculation, #4215.


IvanSanchez:
I don't think it's perfect (not 100% sure about the `ceil()` and the steepness of the sigmoid and such), but this will do until we have time to properly think about it :-)

PR #4227, opened 2016-02-12
Use array in L.Control.Layers internally, fixes #2086

### LINKED ISSUE #2086: Layer order in `L.Control.Layers` is not preserved with GeoJson/FeatureGroup
When I do

```JavaScript
L.control.layers({
    'OSM': L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(map)
}, {
    'Circle': L.circle([53, 4], 111111),
    'Polygon': L.polygon([[48, -3], [50, -4], [52, 4]]),
    'GeoJSON': L.geoJson(geojson)
}, {
    collapsed: false
}).addTo(map);
```

I expect the order of the overlays to be as I supplied them, however, GeoJSON and FeatureGroup layers are moved up top. This is because the FeatureGroup gets its `_leaflet_id` while it's created, while the other layers get it while they're added to Control.Layers.
FeatureGroup is stamped because it's passed as the context to `layer.on` in FeatureGroup.js:19.

Not sure how to fix this:
 - stamping each layer while initializing? Seems dirty. 
 - Different sorting mechanism in Control.Layers?

http://jsfiddle.net/km5H9/1/

jieter:
Quick implementation of using an array to keep the layers in order in `L.Control.Layers`.

Fixes #2086

When calling the constructor with numerical keys is still not preserved, so this:
```JavaScript
L.control.layers({
    2: layer,
    1: other_layer
}).addTo(map);
```
will result in:
2016-02-12-164628_140x49_scrot

To fix that, we'll need to accept arrays of [name, layer]-arrays in stead of objects in case order is important.

### DIFF
--- a/debug/map/control-layers.html
+++ b/debug/map/control-layers.html
@@ -1,0 +1,51 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+</head>
+<body>
+	<div id="map"></div>
+
+	<script type="text/javascript">
+		var geojson = {
+			"type": "Polygon",
+			"coordinates": [[
+				[5.4931640625, 51.781435604431195],
+				[0.9008789062499999, 53.35710874569601],
+				[-2.30712890625, 51.795027225829145],
+				[2.8125, 49.109837790524416],
+				[5.4931640625, 51.781435604431195]
+			]]
+		};
+
+		var map = L.map('map').setView([50.5, 0], 5);
+
+		var OSM_Mapnik = L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
+			maxZoom: 19,
+			attribution: '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>'
+		}).addTo(map);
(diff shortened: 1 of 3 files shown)


jieter:
hang on, just noticed I missed a spot it assumes an object.

jieter:
This should be better, I forgot to update `removeLayer`. Also added two tests, the second covers `removeLayer`.

jieter:
Just rebased this on current master, there was a merge conflict.

jieter:
@yohanboniface This is the PR for @2086

perliedman:
:tada: 

Thanks, great to finally have this longstanding issue fixed!

theashyster:
Just wondering if this is a regression or it is supposed to be like this from now on, but when I try to do `this.LayerControl.removeLayer(layer)` and the `layer` is not added to the LayerControl before I get an error in the console saying `TypeError: this._layers[i] is undefined`. It worked fine before on Leaflet Beta 2, but I get this error now using Leaflet RC 1.

IvanSanchez:
@theashyster Can you let us see a reproducible example of this (e.g. in http://playground-leaflet.rhcloud.com )?

jieter:
This currently fails:

		it('silently returns when trying to remove a non-existing layer from the control', function () {
			var baseLayer = L.tileLayer(),
			    overlay = L.marker([0, 0]),
			    layers = L.control.layers({'base': baseLayer}, {}).addTo(map);

			layers.removeLayer(overlay);
		});

message:
```
undefined is not an object (evaluating 'this._layers[i].layer')
_getLayer@/home/jieter/workspace/Leaflet/src/control/Control.Layers.js:194:31
removeLayer@/home/jieter/workspace/Leaflet/src/control/Control.Layers.js:112:27
/home/jieter/workspace/Leaflet/spec/suites/control/Control.LayersSpec.js:109:22
```

theashyster:
@IvanSanchez @jieter Yep, that is basically the same situation when it fails for me.

PR #4228, opened 2016-02-12
rm spurious fadeAnimated check, fixes #4068

### LINKED ISSUE #4068: On 1.0beta2, Map.fadeAnimation option inteacts with GridLayer.opacity option
In fact, a false Map.fadeAnimation prevents any opacity option on GridLayer taken into account.

I think I've found the problem on src/layer/tile/GridLayer.js:

The blame view on the _updateOpacity() function gives us the result:

```javascript
 		// IE doesn't inherit filter opacity properly, so we're forced to set it on tiles
-		if (!L.Browser.ielt9 && !this._map._fadeAnimated) {
-			L.DomUtil.setOpacity(this._container, opacity);
+		if (L.Browser.ielt9 || !this._map._fadeAnimated) {
 			return;
 		}
 
+		L.DomUtil.setOpacity(this._container, this.options.opacity);
```

You can see those two sentences groups are not equivalent: **!( !a && !b)** is not **a || !b**, it should be **a || b**.

This could make sense on the logic side, but we will have another interaction: GridLayer.setOpacity() will never update the opacity if Map.fadeAnimation is true.

The Map.fadeAnimation should be only taken into considetation if called from the GridLayer._initContainer() function, not from GridLayer.setOpacity() or GridLayer._tileReady() functions.

Perhaps a new boolean parameter on _updateOpacity() could do the job.

IvanSanchez:
Turns out that `this._map._fadeAnimated` is not needed at all in `_updateOpacity`, as `tileload` will not call `_updateOpacity` at all when that option is set to `false`.

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -30,6 +30,17 @@ describe('GridLayer', function () {
 			var grid = L.gridLayer().addTo(map);
 			expect(grid.setOpacity(0.5)).to.equal(grid);
 		});
+
+		it('works when map has fadeAnimated=false (IE8 is exempt)', function (done) {
+			map.remove();
+			map = L.map(div, {fadeAnimation: false}).setView([0, 0], 0);
+
+			var grid = L.gridLayer().setOpacity(0.5).addTo(map);
+			grid.on('load', function () {
+				expect(grid._container.style.opacity).to.equal('0.5');
+				done();
+			});
+		});
 	});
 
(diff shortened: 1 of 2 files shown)


yohanboniface:
I don't see why one Travis build passed but not the other...
Tried to rerun it to see if it was a casual timeout issue, but it still fails, so it may reveal something else.

IvanSanchez:
@yohanboniface I'll rebase and see if having PhantomJS 2.1.1 will get rid of it

PR #4234, opened 2016-02-14
Added an example on map panes

IvanSanchez:
Based on https://stackoverflow.com/questions/35377829/leaflet-js-tilelayer-visible-above-geojson-layer-geojson-interactivity-issue/35378935#35378935


IvanSanchez:
The question is: do we want to show an example which needs 1.0.0-beta?

examples.md
```diff
@@ -50,6 +50,13 @@ A case study of creating a colorful interactive [choropleth map](http://en.wikip
 
 A tutorial on how to manage groups of layers and use the layer switching control.
 
+***
+[<img src="docs/images/labels-pane.png" class="example-img bordered-img" />][8]
+
+### [Working with map panes][8]
```
yohanboniface: I'd add here a small label like `(v1.0.0)`, what do you think?

yohanboniface:
Excellent! Push it up! :)

> The question is: do we want to show an example which needs 1.0.0-beta?

Yes, but let's add a small label in examples list to warn about this?

IvanSanchez:
> let's add a small label in examples list to warn about this?

@yohanboniface I've added a *line of text* instead. I'm such a rebel. :stuck_out_tongue:

yohanboniface:
>I've added a line of text instead. I'm such a rebel.

I'm so merciful that I've merged anyway ;)

PR #4241, opened 2016-02-15
Add test for throwing error if circle radius is NaN

k-nut:
This is a test for the new fix in #4236


yohanboniface:
Nice, thanks! :)

IvanSanchez:
Yeah, tests are always great to have! :-)

PR #4245, opened 2016-02-15
MapSpec #addLayer fix test

scherebedov:
Hi. We use this tests for our code and have one problem. If layers is a lot of then .removeLayer 
hasn't time to remove the layer and spy is called. We can use event 'layerremove' for this. Thanks.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -309,8 +309,10 @@ describe("Map", function () {
 			map.on('layeradd', spy);
 			map.addLayer(layer);
 			map.removeLayer(layer);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Attaching the event should be done *before* calling `removeLayer`. As it is now, removing the layer will fire an event that won't trigger any event handler.

I'm afraid to say I don't think your changes add any functionality to Leaflet :disappointed: - that unit test is about `layeradd`, not about `layerremove`.

scherebedov:
@IvanSanchez , I'm understood. Sorry ))

IvanSanchez:
@scherebedov No problem :smile:

PR #4250, opened 2016-02-16
Stop drag propagation

### LINKED ISSUE #4249: Need to stop map drag propagation
The drag start event in L.Draggable propagates causing trouble if a draggable map is placed within a draggable container. The dragging of the container is overriding map dragging, which is not what I would expect. I'm not able to access the event outside the Leaflet API to stop the propagation.

Leaflet drag bug
The fix is to add "L.DomEvent.stopPropagation(e);" to the _onDown-method: https://github.com/Leaflet/Leaflet/blob/master/src/dom/Draggable.js#L46

Leaflet drag fix
Leaflet version 1.0 beta 2

turban:
This commit fixes #4249

### DIFF
--- a/debug/map/tile-debug.html
+++ b/debug/map/tile-debug.html
@@ -1,0 +1,249 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Leaflet tile debug</title>
+
+    <link rel="stylesheet" href="../../dist/leaflet.css" />
+
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+    <link rel="stylesheet" href="../css/screen.css" />
+    <style>
+    * {
+        margin: 0;
+        padding: 0;
+    }
+    #outer {
+        position: absolute;
+        width: calc(2*256px + 600px);
+        height: calc(2*256px + 400px);
+        border: 10px solid rgba(0, 0, 255, 0.4);
+        background-color: transparent;
+    }
+    #middle, #map {
+        position: absolute;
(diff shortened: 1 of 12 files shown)


IvanSanchez:
Gratulerer i ditt første pull request :-)

I see a couple of issues, though: 
* The unit tests fail (check the travis details, run `npm test` locally next time)
* You have several whitespace-only commits, which you should probably squash into one
* Stopping propagation of the event should be done only if the drag is going to start - in other words, do not stop propagation if I'm holding the `shift` key or dragging with the right mouse button

turban:
Thanks for your guidance @IvanSanchez :-) I'll try to get this right. 

> Stopping propagation of the event should be done only if the drag is going to start - in other words, do not stop propagation if I'm holding the shift key or dragging with the right mouse button

How would you do this check?

IvanSanchez:
> How would you do this check? 

It's right there on `_onDown`, a few lines down. Look closer ;-)

mourner:
This may affect dragging behavior in some browsers so let's test extensively before merging.

turban:
The event now stops to propagate after the three return statements, and I've checked that it still solves the issue.

mourner:
@turban please squash into one commit.

turban:
I have some troubles squashing the commits of this pull requests, so I'm creating a new one.

turban:
New pull request with one commit: https://github.com/Leaflet/Leaflet/pull/4306

PR #4256, opened 2016-02-17
Check for e.originalEvent in DomUtil._filterclick

### LINKED ISSUE #3724: e.originalEvent is undefined for leaflet b1
Hi there,

I'm getting an error: "e.originalEvent is undefined" in the _filterClick function of L.DomEvent, when using Firefox (39) on Android (5).

adding a simple:

    if (e.originalEvent === undefined) {
        handler(e);
        return;
    }

at the top of the function seems to fix the problem without any obvious adverse effects. I'm not sure if i should be reporting issues for the beta version to this GitHub repo ?

cheers

-i

IvanSanchez:
Should fix #3724

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -225,7 +225,7 @@ L.DomEvent = {
 
 	// this is a horrible workaround for a bug in Android where a single touch triggers two click events
 	_filterClick: function (e, handler) {
-		var timeStamp = (e.timeStamp || e.originalEvent.timeStamp),
+		var timeStamp = (e.timeStamp || (e.originalEvent && e.originalEvent.timeStamp)),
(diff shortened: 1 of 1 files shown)


frbehrens:
Are you sure about the fix or am I wrong?
Look at the two lines
```javascript
    var timeStamp = (e.timeStamp || (e.originalEvent && e.originalEvent.timeStamp)),
 		    elapsed = L.DomEvent._lastClick && (timeStamp - L.DomEvent._lastClick);
```
The right part of the OR condition is only evaluated if `e.timeStamp` is not available. But when also `e.originalEvent` is not set, the variable `timeStamp` gets no value assigned. In that case the following expression will use an undefined value.

IvanSanchez:
@frbehrens The point of the fix is to prevent exceptions in the code. If this is causing you glitches (doubled `click` events), then it's worth to have a deeper look.

PR #4261, opened 2016-02-18
Better looking zoomOutText

zakjan:
Replace default zoomOutText: a minus sign instead of a dash, because it looks much better.

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -7,7 +7,7 @@ L.Control.Zoom = L.Control.extend({
 		position: 'topleft',
 		zoomInText: '+',
 		zoomInTitle: 'Zoom in',
-		zoomOutText: '-',
+		zoomOutText: '&minus;',
 		zoomOutTitle: 'Zoom out'
 	},
 


mourner:
What browser/OS do you use? Here's how the current one looks for me in Chrome/OSX:

image

zakjan:
Chrome / Linux.

Font family is Lucida Console, Monaco, monospace. My system is missing all these fonts, so it defaults to monospace, which is DejaVu Sans Mono in my case.

Anyway, a minus sign is still better choice, because it should have the same width as a plus sign.

mourner:
I think it's better to keep it as it is — it was tested extensively across most browsers and OS, and any changes like this that optimize for one browser/OS will cause unwanted changes in others.

zakjan:
> tested extensively across most browsers and OS

Both Lucida Console are Monaco fonts are not present on Linux systems by default. It fallbacks to an uncontrolled font.

In a typographic sense, it should be safer to use a plus/minus pair, because they are guaranteed to have similar width. http://oleb.net/blog/2015/02/minus-sign/

Issue #3944, opened 2015-10-16
IE 10 identified as touch device in v0.7.5


trallarn:
After upgrading from v0.7.3 to v0.7.5 we have noticed that L.Browser.touch is an object on IE 10. We have had reports that the same is true on IE 9 but I have only been able to verify it in document mode 10 on a IE 11 installation running on a mac desktop.

IvanSanchez:
Well, IE10 is, in fact, supporting touch events. So my first response to this would be "it's a feature, not a bug".

What specific issue do you have?

trallarn:
We use L.Browser.touch as an identifier for touch-devices since some of our map controls aren't supported on these. After the upgrade from v0.7.3 to v0.7.5 these controls suddenly wasn't shown on IE 10 because of this change. 

The change appeared as a bug to me, since IE 10 running on a desktop is not a touch-device, even though it might support touch interaction.

IvanSanchez:
> since IE 10 running on a desktop is not a touch-device, even though it might support touch interaction. 

You are using a wrong assumption. IE10 running on a desktop *might* be using a touchscreen. And there's no (good, known) way of detecting that until the first touch event happens, and Leaflet needs to set up stuff *in case* a touch event happens.

I suggest you watch https://www.youtube.com/watch?v=wwffqMAS8K8 (also http://www.slideshare.net/redux/getting-touchy-an-introduction-to-touch-and-pointer-events-web-rebels-oslo-22-may-2014) Interesting bits around minute 15 (or slide 46).

Welcome to the pointer/touch/click jungle. ;-)

IvanSanchez:
> some of our map controls aren't supported on [touch devices]

@trallarn I suggest you do the complete opposite - instead of disabling map controls on touch devices, enable map controls on mouse-capable devices. Feature detection FTW.

trallarn:
That seams reasonable. I will look into inverting the checks. Thank you for your help Ivan. And thanks for the links.

PR #4262, opened 2016-02-18
fix ie touch

scherebedov:
Hi. In desktop Internet Explorer 10,11 we have _L.Brwoser.touch_ is _true_. It is not ok. Can we have use _navigator.MaxTouchPoints_ and _navigator.msMaxTouchPoints_. Thanks.


IvanSanchez:
No. See #3839 and #3944. 

Furthermore, your fix would all iOS devices, as [Safari doesn't implement `maxTouchPoints`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints#Browser_compatibility).

PR #4263, opened 2016-02-18
centralize setting of attributionControl on map from onAdd

### LINKED ISSUE #4257: Attributions not changing when tilelayer changes
Hi!

Problem: Modified Attribution Control text not changing when tilelayer is changed.
Version: Not working in 1.0, but works in 0.7.7

My code:
```
var map = L.map('map', {
	center: [58.81, 23.08],
	minZoom: 8,
	maxZoom: 19,
	attributionControl: false,
	zoomControl: !L.Browser.mobile,
	layers: [navionics_kaart, weatherstations]
});
```

```
L.control.attribution({
	prefix: false,
	position: 'bottomleft'
}).addTo(map);
```

When I delete second block and use default position for Attribution control, then everything works fine.

Am I using L.control.attribution incorrectly in 1.0? Because same code in 0.7.7 works as I want.

[Link to my website where bug is present](http://siim.planet.ee/meri/)

snkashis:
Set attributionControl during onAdd instead of during initHook

Fixes #4257

### DIFF
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -15,6 +15,7 @@ L.Control.Attribution = L.Control.extend({
 	},
 
 	onAdd: function (map) {
+		map.attributionControl = this;
 		this._container = L.DomUtil.create('div', 'leaflet-control-attribution');
 		if (L.DomEvent) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Wow, nice catch!

PR #4273, opened 2016-02-24
whenReady callback fails

Siliconrob:
Experimenting with multiple vector layers adding/removing dynamically the callback is undefined for the previous GeoJson layer and will cause an exception to be triggered on a dead layer.  When an undefined callback is handled by a layer it falls back to a no op function as last resort.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -742,7 +742,7 @@ L.Map = L.Evented.extend({
 	},
 
 	whenReady: function (callback, context) {
-		callback = callback || function(e) {};
+		callback = callback ? callback : function() {};
 		if (this._loaded) {
 			callback.call(context || this, {target: this});
 		} else {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
How are you removing the callback?

Also, note that the unit tests have failed - I suggest you run `npm locally` before pushing a commit.

Siliconrob:
Adding a layer like 

var dynLayer = L.geoJson(feature);
dynLayer.addTo(theMap);

and then removing later seems to orphan the geojson layer.  Exception kept popping up in whenReady because the callback is undefined.

Currently trying to determine how to make the tests pass as this is a minor change

Siliconrob:
Not sure how to fix the failing tests from PhantomJS on code I didn't touch at all.  I ran it locally several times and got the same thing as the Travis CI is saying

My local machine is a Windows 10 running node x64 and npm 3.7.3

When I went and tried the browser tests under the /spec folder it didn't have these test failures, but had a single different error about icon changing 

-- suites/layer/marker/MarkerSpec.js:35:4

PR #4275, opened 2016-02-25
Added Leaflet.CountrySelect plugin

ahalota:


IvanSanchez:
@ahalota I fing the plugin a bit buggy - I cannot scroll the drop-down list, or drag the scrollbar (events get passed to the map). Can you fix that?

jieter:
I can do both in the drop-down without a problem...

http://ahalota.github.io/Leaflet.CountrySelect/demo.html

IvanSanchez:
Apparently it works in chrome but fails in firefox.

ahalota:
@IvanSanchez What do you mean by drag the scrollbar?

As for scrolling the drop down list - When I run it in firefox, I can scroll the list with the key up/down buttons. The change event is only fired once you hit 'enter' however. I believe this is just part of how firefox handles change events on `select` elements

> When the element loses focus after its value was changed, but not commited (e.g. after editing the value of &lt;textarea&gt; or &lt;input type="text"&gt;

[MDN change - event reference](https://developer.mozilla.org/en-US/docs/Web/Events/change)

ahalota:
I fixed the clicking, it shouldn't act as strange in Firefox or IE anymore.

I am still working on getting the scrollwheel to play nice. I am having trouble stopping leaflet from 'stealing' the scroll event from my dropdown in Firefox and IE. I will see what I can do about that. Thank you for the feedback.

PR #4278, opened 2016-02-26
Pass GeoJSON options object to Marker constructor, fixes #4277

### LINKED ISSUE #4277: GeoJSON markers being placed on a default pane instead of expected specified pane / 1.0.0-beta.2
Not sure if it's a bug or not but it looks strange as lines and polygons can be placed to separate pane and markers - can not.
Could it be intentionally as markers use two panes actually?

Fix is pretty simple, pr follows.

Probably we can have panes created automatically by `L.Map` instead of creating them ahead, what do you think about it?

Actually, I was looking for an easy way of turning layers on/off and I would vote for a something in api like this:

```javascript
var layer = new L.GeoJSON(data);

// somewhere later
layer.hide();
layer.show();
``` 
instead of

```javascript
var pane = map.createPane('green'),
    layer = new L.GeoJSON(data, {pane: 'green'});

// somewhere later
pane.style.display = 'none';
pane.style.display = 'block';
```

as it looks better in code and has pretty obvious semantic.
I've looked through #4, #1742 and #3266 and I feel I'm not alone searching for this. :)

Thank you for a great library!

ysegorov:
This pr fixes #4277

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -89,12 +89,12 @@ L.extend(L.GeoJSON, {
 		switch (geometry.type) {
 		case 'Point':
 			latlng = coordsToLatLng(coords);
-			return pointToLayer ? pointToLayer(geojson, latlng) : new L.Marker(latlng);
+			return pointToLayer ? pointToLayer(geojson, latlng) : new L.Marker(latlng, options);
 
 		case 'MultiPoint':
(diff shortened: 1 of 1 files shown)


FadilEldin:
Closed.
Should be permamnet close

IvanSanchez:
Closing (merge conflict, already implemented some other way)

johnd0e:
> already implemented some other way

But #4277 is still opened.

Issue #4274, opened 2016-02-25
Adding or removing layers from layers control which has been removed from the map causes null reference


perliedman:
If you add a layers control to the map, and later remove it from the map, you can't add or remove layers from it, since it causes a null reference.

The problem occurs when `_addItem` is called, and `this._map` is not set: https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js#L190

We could probably just skip calling `_update` when `this._map` is not set, since `_update` will be called again when (if) the control is added to the map again.

This issue also exists in 0.7.x.

Fiddle to reproduce: http://playground-leaflet.rhcloud.com/cal/edit?html,output

IvanSanchez:
Looks like skipping `this._update()` when calling `addBaseLayer` or `addOverlay` or `removeLayer` if `this._map` is not set should do the trick.

This seems very simple. How about we tell @Charlotteis/@YourFirstPR so a newbie can fix it?

perliedman:
@IvanSanchez :+1: sounds like a plan!

Charlotteis:
Thank you!

errebenito:
I can probably do this one. Just so we're all on the same page, I suppose the _modus operandi_ would be as follows, feel free to correct me if I overlooked something:

1. Claim the issue (done, I guess)
2. Fork the repo
3. Modify _addBaseLayer_, _addOverlay_ and _removeLayer_ to only return _this._update()_ if _this._map_ is not null. Point of interest: what should be returned if it is null?
4. Commit fix and make pull request.

If that is correct and nobody else claimed it while I typed this, I'll get to it.

IvanSanchez:
@errebenito If you read the [docs for `L.Control.Layers`](http://leafletjs.com/reference-1.0.0.html#control-layers-addbaselayer), you'll see that the three methods return `this`. In other words, they return a reference to the control so you can chain methods.

If you look at the code, then the three public methods do `return this._update()` and then `_update()` does `return this;`.

(Por cierto, Raúl: si rodeas texto con *accent greu*, entonces tiene `apariencia de código`. Si lo rodeas con asteriscos, *entonces es cursiva* ;-) )

errebenito:
Ah, I see. So if `this._map` is null I merely return `this` without calling `_update()`. Alright, I'll comment again once it's done. Thanks for the clarification (y el consejo sobre Markdown).

errebenito:
Pull request created. For reference, the return statement now looks like this:

[...]

PR #4280, opened 2016-02-26
Skip call to this._update() if this._map is null

errebenito:
Changed `addBaseLayer()`, `addOverlay()`y and `removeLayer()` to only return `this._update()` if `this._map` is not null, and return `this`otherwise. Hopefully solves [issue 4274](https://github.com/Leaflet/Leaflet/issues/4274).


mourner:
The build passed in master :+1:

PR #4287, opened 2016-02-29
Clean up navigator.pointerEnabled

IvanSanchez:
Quoting from https://docs.webplatform.org/wiki/dom/Navigator/pointerEnabled :

> In late 2013, pointerEnabled was removed from the specification as checking PointerEvent in Window object is sufficient for feature detection. Do not use this property and use PointerEvent instead. 

See also https://msdn.microsoft.com/en-us/library/windows/apps/hh972607.aspx , https://github.com/jquery/PEP/commit/30493f1c490200645294f516134a515bc2f4ccf7. This should be safe to remove.


mourner:
Looks like the tests are fine except these timeouts, so :+1: on merge.

PR #4290, opened 2016-02-29
Increase depth of legacy support

JrFolk:
A lot of plugins use the capitalized internal constructors rather than
the API.  This should broaden legacy support for plugins that use the
L.Circle constructor.

### DIFF
--- a/src/layer/vector/Circle.js
+++ b/src/layer/vector/Circle.js
@@ -5,7 +5,11 @@
 
 L.Circle = L.CircleMarker.extend({
 
-	initialize: function (latlng, options) {
+	initialize: function (latlng, options, legacyOptions) {
+		if (typeof options === 'number') {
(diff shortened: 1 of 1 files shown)


JrFolk:
I discovered this issue with [Leaflet.draw](https://github.com/Leaflet/Leaflet.draw/blob/master/dist/leaflet.draw-src.js#L922-L925).  Rather insidious without the [thrown error](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Circle.js#L12), as `setRadius()` is invoked on `mousemove` so it only manifested in the layer attached to the event.

mourner:
I agree. Please fix the build and I'll merge.

mourner:
Oh, the build failure is unrelated to the change. Good to merge then.

PR #4300, opened 2016-03-04
Add plugin Leaflet Underneath

perliedman:


perliedman:
@IvanSanchez wow, quick! Just went to get a glass of water before merging it myself!

Thanks.

IvanSanchez:
BTW, it's a bit difficult to see the highlighted points when they are north of the clicked position - the popup obscures them ;-)

perliedman:
Yes, agreed. That's really just the demo, but I should probably fix that somehow.

PR #4303, opened 2016-03-05
Added plugin leaflet.marker.pin

wwwouaiebe:


wwwouaiebe:
Added plugin leaflet.marker.pin

yohanboniface:
Thanks!

PR #4306, opened 2016-03-07
Stop drag propagation, fixes #4249

### LINKED ISSUE #4249: Need to stop map drag propagation

turban:
New pull request with one commit.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -60,6 +60,8 @@ L.Draggable = L.Evented.extend({
 
 		if (this._moving) { return; }
 
+		L.DomEvent.stopPropagation(e);
+
 		this.fire('down');
 
 		var first = e.touches ? e.touches[0] : e;


yohanboniface:
Humm, this is the opposite of what have been done in #3682.
This will break Leaflet.Editable. I'll try to have a look with a bit of calm to make a counter-proposal OR fix on Leaflet.Editable side.

mourner:
Damn. Maybe make stopPropagation optional, and enable it just for the map container?

turban:
Would it be possible to expose the original event so it can be stopped outside the API.

yohanboniface:
@turban, you should have it in `e.originalEvent`

turban:
@yohanboniface, I've not yet succeeded in doing that. Which event on the map should I listen to, to access the original event?

yohanboniface:
I'd say something like:
```javascript
map.on('mousedown', function (e) {
    e.originalEvent.stopPropagation();
});
```
See http://playground-leaflet.rhcloud.com/keri/edit?html,console,output for a playground ;)

turban:
I was sure I tested that @yohanboniface, but on a second try this seems to work. I'm happy to withdraw my pull request @mourner.

mourner:
Reverted.

PR #4173, opened 2016-01-27
Possible change for #200 - Adding support for dashArray for line canv…

gommo:
Wanting some review of this. Would this affect performance.

I don't think I should store my temp variable lineDash on the canvas ctx. Wanting advise there as well

PR #4308, opened 2016-03-08
Port #4173 to master: dashArray in Canvas

IvanSanchez:
Allows using `dashArray` in `<canvas>` renderers.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -74,6 +74,11 @@ L.Canvas = L.Renderer.extend({
 	},
 
 	_updateStyle: function (layer) {
+
+		if (layer.options.dashArray) {
+			layer.options._dashArray = layer.options.dashArray.split(',').map(Number);
+		}
+
 		this._requestRedraw(layer);
 	},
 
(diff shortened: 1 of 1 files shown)


yohanboniface:
Worth a unittest?

IvanSanchez:
@yohanboniface This is a graphical thing - in order to do unit tests for this we would need to do like Mapnik does, and compare screenshots to reference screenshots. Bit of an undertaking right now?

yohanboniface:
Was just thinking about testing that the attributes are on the DOM element. But maybe that would be like testing browser Canvas implementation at the end…

mourner:
Oh, I didn't even notice it was for stable previously...

gommo:
We just found out that setLineDash isn't supported on IE10 and below so the code needs to check for that

IvanSanchez:
@gommo Is there any workaround for IE10? According to https://msdn.microsoft.com/en-us/library/dn265063%28v=vs.85%29.aspx , it doesn't seem like it.

We might just need to note this down in the API docs for the `dashArray` option.

gommo:
I'll need to add a change so we don't actually call setLineDash

IvanSanchez:
@gommo Wait - does IE10 throw an error or something like that?

gommo:
Yeah it does!

IvanSanchez:
@gommo Then we need to fix this again!

PR #4314, opened 2016-03-10
Fix #4313 : correct simulated click handling in L.Path

### LINKED ISSUE #4313: Path click event is ignored when click is simulated from L.Map.Tap
In L.Map.Tap a click event is simulated "if the touch didn't move too much" (according to the source comments).

And this behavior is well implemented in `L.Map._onMouseClick` function :
```javascript
if (!this._loaded || (!e._simulated &&
		        ((this.dragging && this.dragging.moved()) ||
		         (this.boxZoom  && this.boxZoom.moved()))) ||
		            L.DomEvent._skipped(e)) { return; }
```
(note the    `!e._simulated &&`   )

But in `L.Path._onMouseClick` (SVG implementation), the `e._simulated` is not tested.
As a consequence, when a click is simulated, the click event on the Path is abnormally ignored but not the Map one.
I think that 
```javascript
	_onMouseClick: function (e) {
		if (!e._simulated && this._map.dragging && this._map.dragging.moved()) { return; }

		this._fireMouseEvent(e);
	},
```
for `L.Path._onMouseClick` would do the trick.

I'll try to make a Pull request.
Sorry for my english and thanks for this great library!

elkami12:
Related to #4313 : add a test of e._simulated in `L.Path._onMouseClick` in SVG implementation

### DIFF
--- a/src/layer/vector/Path.SVG.js
+++ b/src/layer/vector/Path.SVG.js
@@ -129,7 +129,7 @@ L.Path = L.Path.extend({
 	},
 
 	_onMouseClick: function (e) {
-		if (this._map.dragging && this._map.dragging.moved()) { return; }
+		if (!e._simulated && this._map.dragging && this._map.dragging.moved()) { return; }
 
 		this._fireMouseEvent(e);
 	},


yohanboniface:
LGTM

PR #4317, opened 2016-03-11
Update quick-start.md (small correction)

Kayoku:
When you need to make sure that the map container has a defined height with :

`#mapclass { height: 180px; }`

that's not working. Because we did this before :

`<div id="mapid"></div>`

Rename **"mapclass"** to **"mapid"**. It's more clear for beginner who can just copy/paste code.


mourner:
Thanks!

PR #4318, opened 2016-03-11
prettier image-overlay.html

BaptisteOrieux:
We made a prettier version of debug/map/image-overlay.html
It's now easier to see what this function does.

imageoverlay


mourner:
The original example is better IMHO since it shows an actual georeferenced image rather than some abstract illustration that could be added a simple marker instead.

PR #4324, opened 2016-03-15
fixing this._times.length undefined in Map.Drag.js, fixes #4311

### LINKED ISSUE #4311: `this._times.length` is `undefined` in `Map.Drag.js`
I have just got report from my client that `this._times.length` is `undefined` in [Map.Drag.js](https://github.com/Leaflet/Leaflet/blob/09c322121197ae8ecb6110ac61bffdafd9d72d25/src/map/handler/Map.Drag.js) in `_onDragEnd` on line `noInertia = !options.inertia || this._times.length < 2;`. I could not reproduce the error so far. The error was triggered on IPad, iOS 9.2.1 in WKWebview. `options.inertia` is kept at default, i.e. `true`.

My only explanation to this error is that in rare situations the `_onDragEnd` is called without `_onDragStart` which then means that this

    if (map.options.inertia) {
        this._positions = [];
        this._times = [];
    }

is never called and `this._times` remains `undefined`.

What do you think?

LucasMouraDeOliveira:


src/map/handler/Map.Drag.js
```diff
@@ -77,10 +79,6 @@ L.Map.Drag = L.Handler.extend({
 		    .fire('movestart')
 		    .fire('dragstart');
 
-		if (map.options.inertia) {
-			this._positions = [];
-			this._times = [];
```
IvanSanchez: Why remove the initialization in `_onDragStart`? If inertia is enabled, we want to forget the speed of the previous drag(s) whenever a new drag is started; so we want to keep this bit of code.

LucasMouraDeOliveira: Sorry, my bad, i'm updating it right now.

IvanSanchez:
Looks good now, but please [squash your commits](https://davidwalsh.name/squash-commits-git) into one. For very small features/bugfixes such as this one we prefer to have just one commit.

IvanSanchez:
:+1:

PR #4337, opened 2016-03-21
Support {-y} in tile layer urls (Fixes #4284)

### LINKED ISSUE #4284: L.tileLayer: Support tms {-y} format
When using tms format with a -y value in the tileserver specification, currently Leaflet.js requires this to be specified as ```{y}``` but with a flag ```tms: true``` as documented at:
http://leafletjs.com/reference.html#tilelayer-tms

Various documentation contains references like this:

```
http://geo.nls.uk/mapdata2/bartholomew/great_britain/{zoom}/{x}/{-y}.png
```

e.g. many examples here:
http://wiki.openstreetmap.org/wiki/National_Library_of_Scotland
http://osmlab.github.io/historic-imagery-index/

It would be useful if Leaflet.js could natively support ```{-y}``` so these can be cut-and-pasted and then people spend a few hours working out that they need to remove the - and add the ```tms: true``` flag.

jieter:
Support `{-y}` in tile layer urls.

Fixes #4284

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -75,11 +75,14 @@ L.TileLayer = L.GridLayer.extend({
 	},
 
 	getTileUrl: function (coords) {
+		var invertedY = this._globalTileRange.max.y - coords.y;
+
 		return L.Util.template(this._url, L.extend({
 			r: L.Browser.retina ? '@2x' : '',
 			s: this._getSubdomain(coords),
 			x: coords.x,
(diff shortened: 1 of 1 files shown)


jieter:
Should I add docs for this to `gh-pages`?

balciseri:
are you sure it's working when also using  crs: L.CRS.Simple?
I'm not sure but  `var invertedY = this._globalTileRange.max.y - coords.y;`
will cause `Uncaught TypeError: Cannot read property 'max' of undefined`
Can you check?

jieter:
I'm sure it's not working with CRS.simple, there is another issue for that: #4338.

PR #4346, opened 2016-03-22
add Leaflet.BootstrapZoom plugin

MAD-GooZe:
This plugin adds a drop-in Twitter Bootstrap styling for Leaflet.js zoom control buttons.


perliedman:
Thanks!

Two things:

* There's mixed tabs and spaces in the indentation for the HTML you added; please only use tabs
* could you squash these commits into one before we merge? It helps keep our history somewhat clean.

PR #4352, opened 2016-03-26
GeoJSON layer - this context for option callbacks

tugrul:
Following option callbacks can access GeoJSON layer
instance through this keyword.

- pointToLayer
- onEachFeature
- filter
- coordsToLatLng

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -31,7 +31,7 @@ L.GeoJSON = L.FeatureGroup.extend({
 
 		var options = this.options;
 
-		if (options.filter && !options.filter(geojson)) { return this; }
+		if (options.filter && !options.filter.call(this, geojson)) { return this; }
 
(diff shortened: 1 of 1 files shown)


tugrul:
There are some lack about this pull request. I didn't see `geometryToLayer` is static function and i'm going to change this. Please ignore first commit of this pull request.

IvanSanchez:
I cannot see why this is needed. Can you illustrate with a use case?

tugrul:
@IvanSanchez use case https://gist.github.com/tugrul/851ed5fa3e435cf9b940

mourner:
I don't think this is necessary, since you can just use `bind` to set the context whatever you need.

tugrul:
@mourner using `bind` can be workaround but not solution. Also i can set option in initializer and access in this scope. Basic principle: Every callback should have context own instance.

Issue #4345, opened 2016-03-22
L.Map.fitBounds() with invalid bounds error message could be better


Jonas-WMDE:
`Uncaught TypeError: Cannot read property 'lat' of undefinedo.LatLngBounds.getNorth @ leaflet.js:6o.LatLngBounds.getNorthWest @ leaflet.js:6o.Map.o.Class.extend.getBoundsZoom @ leaflet.js:6o.Map.o.Class.extend.fitBounds @ leaflet.js:6wikibase.queryService.ui.resultBrowser.CoordinateResultBrowser.SELF.draw @ CoordinateResultBrowser.js:64(anonymous function) @ App.js:411`

http://tinyurl.com/jbxrwlv

IvanSanchez:
Maybe a check for `bounds.isValid()` inside `fitBounds()` is a good idea (throw a descriptive error if not). Otherwise, `isValid()` can be called from the user's code.

IvanSanchez:
Maybe this bug report is a good candidate for @yourfirstpr ?

Code-wise, the changes are simple (inside the `fitBounds` method in `L.Map`, check if the bounds are valid with a call to `bounds.isValid()`, then throw an error in the same way `_initContainer` does. 

Other than that, it's the usual "claim the issue, fork, clone, branch, edit, commit, pull request" workflow.

theotow:
i will give it a try

IvanSanchez:
Closed via #4353

renandegrandi:
Formerly the fitBounds () function could receive an array of markers.
With this the validation:
`bounds =  L.latLngBounds(bounds);`

Will always be invalid.

But within the call:
`_getBoundsCenterZoom: function (bounds, options) {`

Is validated as follows the bounds.
`bounds = bounds.getBounds ? bounds.getBounds() : L.latLngBounds(bounds);`

I made some tests here and to stop the errors in the new version the validation should be as follows:

```
	fitBounds: function (bounds, options) {

	    bounds = bounds.getBounds ? bounds.getBounds() : L.latLngBounds(bounds);

		if (!bounds.isValid()) {
			throw new Error('Bounds are not valid.');
		}

		var target = this._getBoundsCenterZoom(bounds, options);
		return this.setView(target.center, target.zoom, options);
	},
```

ghybs:
Hi,

[...]

PR #4353, opened 2016-03-28
feat: added better error for fitBounds

theotow:
related to #4345

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -119,6 +119,11 @@ L.Map = L.Evented.extend({
 	},
 
 	fitBounds: function (bounds, options) {
+
+		if (!bounds.isValid()) {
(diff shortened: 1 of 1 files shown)


yohanboniface:
Thanks for this PR @theotow, however one test is failing on Travis, have you seen that and been able to investigate?
Just ask if you need some hand (you can also join a chat either in [Gitter](https://gitter.im/Leaflet/Leaflet) or [IRC](https://webchat.freenode.net/?channels=leaflet)). :)

jieter:
Seems to be an prosthetic hand timing issue again, let's get those callbacks into the hand

IvanSanchez:
Yeah, it looks like a prosthetic-hand issue. I'll merge the code in and will take the blame if I break `master`.

Thanks for the PR, @theotow !

PR #4355, opened 2016-03-30
Give popups an id

tylercubell:
Useful in the case where having a class is too general and finer-grain control is necessary.

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -25,6 +25,7 @@ L.Popup = L.Layer.extend({
 		autoClose: true,
 		// keepInView: false,
 		// className: '',
+		// id: '',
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I don't think this is very useful for the general case - if you really need that, you can replace it with a three-liner in your own code:

    L.Popup.addInitHook(function(){
        this._container.id = this.options.id;
    });

tylercubell:
`this._container` is undefined. Any ideas?

IvanSanchez:
Oh yeah, `this._container` is created in `onAdd`, not in `initialize`.

    L.Popup.addInitHook(function(){
        this.on('add', function() {
            this._container.id = this.options.id;
        });
    });

PR #4358, opened 2016-03-31
Adds a workaround for Leaflet/Leaflet#3575

href:
As requested in #3575. See discussion there.


mourner:
I don't think this workaround is acceptable, since as far as I remember, it disables hardware acceleration of tiles, making Leaflet slower on all platform. Perhaps the best option would be to just wait until the bug is fixed in browsers, and ask developers that need this fixed to use the workaround in the application code.

PR #4364, opened 2016-04-01
Initialize canvas dash on init. Check that canvas supports setLineDash.

### LINKED ISSUE #4351: dashArray on canvas crashes IE9, IE10
See #4171 and specially the comments at the end of #4308: the method for setting the `dashArray` on a `<canvas>` renderer triggers a crash in IE10.

A bit of debugging in an actual IE10 is needed, but the fix might be something like

```
if ('setLineDash' in ctx) { ctx.setLineDash(....) }
```

### LINKED ISSUE #4363: dashArray on canvas is only used if set by setStyle
The initlization of `this.options._dashArray` is only done in `_updateStyle`, which is only called when `setStyle` is used: https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L76

perliedman:
Close #4363. Close #4351.

### DIFF
--- a/debug/vector/vector-canvas.html
+++ b/debug/vector/vector-canvas.html
@@ -83,6 +83,11 @@
 		});
 		group.addLayer(polygon);
 
+		var line = new L.Polyline([h1, h4, h5], {
+			dashArray: '5, 5'
+		});
+		group.addLayer(line);
+
 		polygon.bindPopup('I am a polygon');
 
 		map.addLayer(group);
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Note to self: test in IE8/IE9/IE10. Maybe the canvas context implements `setLineDash` as a stub or something that might break this code.

perliedman:
Still waiting for my IE10 VM to download, but there's definitely other code feature detecting this way: https://github.com/CreateJS/EaselJS/blob/master/src/easeljs/display/Graphics.js#L2147

perliedman:
Confirmed, this fix/feature detect works in IE10.

PR #4367, opened 2016-04-02
Just keep _moveEnd inside requestAnimFrame. Close #4023.

### LINKED ISSUE #4023: zoomEnd event fired twice in master with FF & Chrome
Hi,

It looks like Leaflet master (and 1.0.0-beta.2) sometimes fires **twice** `zoomEnd` event after a zoom in our out, whether using the mouse wheel, zoom control buttons, or programmatical `map.zoomIn()` / `map.zoomOut()`.
- With Leaflet CDN 1.0.0-beta.2: http://jsfiddle.net/3v7hd2vx/11/
- With CloudFlare CDN 1.0.0-beta.2.rc.2: http://jsfiddle.net/3v7hd2vx/9/

However, Leaflet 1.0.0-b1 does not seem to have this issue (nor stable).
- With Leaflet CDN 1.0.0-b1: http://jsfiddle.net/3v7hd2vx/10/

I am using FF and Chrome on Ubuntu.

perliedman:
Modifications for avoiding duplicate zoom events from firing, as I suggested in https://github.com/Leaflet/Leaflet/issues/4023#issuecomment-204461090.

Note that this change changes behaviour: previously then `moveend` event fired _immediately_ after the last `move` event, but this change will make other code run between the two events.

### DIFF
--- a/src/map/anim/Map.ZoomAnimation.js
+++ b/src/map/anim/Map.ZoomAnimation.js
@@ -114,13 +114,13 @@ L.Map.include(!zoomAnimated ? {} : {
 
 		L.DomUtil.removeClass(this._mapPane, 'leaflet-zoom-anim');
 
+		this._animatingZoom = false;
+
+		this._move(this._animateToCenter, this._animateToZoom);
+
 		// This anim frame should prevent an obscure iOS webkit tile loading race condition.
 		L.Util.requestAnimFrame(function () {
-			this._animatingZoom = false;
-
-			this
-				._move(this._animateToCenter, this._animateToZoom)
-				._moveEnd(true);
+			this._moveEnd(true);
 		}, this);
 	}
 });


yohanboniface:
We still need an iOS device to check that, cc @IvanSanchez ;)

IvanSanchez:
I made a playground using the build for this pull request to check if the `zoomstart ànd `zoomend` event counts are the same:

http://playground-leaflet.rhcloud.com/duw/edit?html,output

We need somebody with iOS to test it (in that case, the URL to check is http://playground-leaflet.rhcloud.com/duw )

hyperknot:
The bug was not reproducible in iOS 8 simulator under XCode (it was correct
under beta2 and master).

If someone from an actual iOS device could test it and report back it would
be great.

cmulders:
I tested the url on my iPhone 5s iOS 9.3. The counts stay the same, if I do not interrupt while animating. 
When I pinch zoom/double tap zoom, the zoom end sometimes does not fire (double zoomstart I guess) and the counts go out of sync.

IvanSanchez:
@cmulder That sounds nice, actually! The problem was *duplicated zoomend events* due to CSS, and what you're experiencing sounds your touch-pinches interfere with the programmed zooms, which is not a problem for the Leaflet core.

PR #4368, opened 2016-04-02
Make latLngBounds factory return an empty bounds with not argument

### LINKED ISSUE #3408: "L.latLngBounds" behavior differs from "new L.LatLngBounds"

yohanboniface:
Just like the constructor.

Fix #3408


perliedman:
:+1:

PR #4370, opened 2016-04-02
Make L.Control.Layers.collapse/expand public methods

### LINKED ISSUE #2713: Collapse layer control after tap on mobile phones

yohanboniface:
The aim is to allow people to call them programmatically.
So they must be public, documented and thus reliable.

Closes #3049
Closes #2713

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -113,6 +113,27 @@ L.Control.Layers = L.Control.extend({
 		return (this._map) ? this._update() : this;
 	},
 
+	// @method expand(): undefined
+	// Expand the control container if collapsed.
+	expand: function () {
+		L.DomUtil.addClass(this._container, 'leaflet-control-layers-expanded');
+		this._form.style.height = null;
+		var acceptableHeight = this._map.getSize().y - (this._container.offsetTop + 50);
+		if (acceptableHeight < this._form.clientHeight) {
+			L.DomUtil.addClass(this._form, 'leaflet-control-layers-scrollbar');
+			this._form.style.height = acceptableHeight + 'px';
+		} else {
+			L.DomUtil.removeClass(this._form, 'leaflet-control-layers-scrollbar');
+		}
+		this._checkDisabledLayers();
+	},
+
+	// @method collapse(): undefined
+	// Collapse the control container if expanded.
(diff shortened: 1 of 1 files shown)


src/control/Control.Layers.js
```diff
@@ -113,6 +113,27 @@ L.Control.Layers = L.Control.extend({
 		return (this._map) ? this._update() : this;
 	},
 
+	// @method expand(): undefined
```
perliedman: Maybe return `this`, like most of our public methods which would otherwise be `void` do?

src/control/Control.Layers.js
```diff
@@ -113,6 +113,27 @@ L.Control.Layers = L.Control.extend({
...
+		if (acceptableHeight < this._form.clientHeight) {
+			L.DomUtil.addClass(this._form, 'leaflet-control-layers-scrollbar');
+			this._form.style.height = acceptableHeight + 'px';
+		} else {
+			L.DomUtil.removeClass(this._form, 'leaflet-control-layers-scrollbar');
+		}
+		this._checkDisabledLayers();
+	},
+
+	// @method collapse(): undefined
```
perliedman: Maybe return `this`, like most of our public methods which would otherwise be `void` do?

src/control/Control.Layers.js
```diff
@@ -346,7 +350,18 @@ L.Control.Layers = L.Control.extend({
 			                 (layer.options.maxZoom !== undefined && zoom > layer.options.maxZoom);
 
 		}
+	},
+
+	_expand: function () {
+		// Backward compatibility, remove me in 1.1.
```
perliedman: Don't know if we've deprecated something before, and how we did it, but perhaps put a `console.warn` here and in `_collapse`, telling we're removing this in 1.1 and that you should use `collapse`/`expand`?

yohanboniface: As just seen IRL, but for the record: given it was private method, I think we are kind already to keep backward compatibility and there is not need to go the formal process. :)

Issue #2699, opened 2014-05-21
Input elements in controls on top of the map won't respond to touch on IE11


fnicollet:
Hi,

This issue only happens on recent version of IE11 with touch enabled. If you have a select tag in a leaflet-control, it won't open when you tap on it.
Debugging the code, the function addPointerListenerStart is called when initializing the map and adds a "pointerdown" listener on the map (and all his descendants).
In the callback function, the first instruction is:

L.DomEvent.preventDefault(e);

Which prevent the default select behaviour (opening the options). From what I see, it should also affect inputs in other components such as marker popups.

I am not sure how to fix this

Thanks,
Fabien

yohanboniface:
@perliedman can you just have a quick look with IE11 to see if the issue is still present, and if then we should close, fixed during today sprint, keep as 1.0 fix or move to 1.1? :)

perliedman:
Example to test this out using latest Leaflet-dev: https://playground-leaflet.rhcloud.com/rul/edit?html,output

perliedman:
This is still a problem on IE11. I'm going for implementing that changes suggested by @danzel in https://github.com/Leaflet/Leaflet/issues/2699#issuecomment-44219069 and https://github.com/Leaflet/Leaflet/issues/2699#issuecomment-44222633.

perliedman:
`L.DomEvent.disableClickPropagation` stops propagation of `touchstart` for touch devices; for the `select` element, this unfortunately prevents it from opening on touch (at least in IE11, still works in Chrome). So it looks like modifying and using `disableClickPropagation´ is not an option, since it must disable `touchstart`.

perliedman:
Closed by #4371, we `preventDefault` on some form controls in IE.

PR #4371, opened 2016-04-02
Don't prevent pointerdown if target is a form control

perliedman:
This is WIP to attempt fixing #2699.

The current approach is to white-list some tag types and not call `preventDefault` if they're the target, more or less what @fnicollet [suggests](https://github.com/Leaflet/Leaflet/issues/2699#issuecomment-61363343). This seems to fix the problem, but I can't say it's elegant.

The other alternative I've considered involves actually calling `stopPropagation` on every form element (like suggested by @danzel: https://github.com/Leaflet/Leaflet/issues/2699#issuecomment-44219069) which is messy and something people are likely to not understand that they have to do.

Any better solution? @danzel, do you have any suggestions?

### DIFF
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -8,6 +8,7 @@ L.extend(L.DomEvent, {
 	POINTER_MOVE:   L.Browser.msPointer ? 'MSPointerMove'   : 'pointermove',
 	POINTER_UP:     L.Browser.msPointer ? 'MSPointerUp'     : 'pointerup',
 	POINTER_CANCEL: L.Browser.msPointer ? 'MSPointerCancel' : 'pointercancel',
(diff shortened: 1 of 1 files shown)


danzel:
If the only time we need this is for pointer events, then this is probably an ok place to do it.
Definitely want a note explaining why though, cause it's a weird thing to do!

Could we do a prosthetic hand test?


I like that this is something that we handle vs forcing users to do it.

perliedman:
I added a comment.

Not entirely sure if prosthetic hand can do this, I tried just very quickly to hack up something that touches a control, but couldn't make it actually open the control. @IvanSanchez what do you think, is this something we could prosthetic for?

perliedman:
So, I removed the _"DON'T MERGE"_ warning from the title. Given current feedback it seems like an ok approach, and fixes a real issue for IE11 users.

A test would be nice, but I haven't been able to get prosthetic hand to click on on a `select` control so far; don't know if I can expect it to do that.

PR #4372, opened 2016-04-02
Switch prosthetic-hand to 1.3.0 and use onStop callback

IvanSanchez:
This should fix our troubles with some tests failing randomly


yohanboniface:
prosthetic :+1: :)

PR #4373, opened 2016-04-02
Remove use of ES5 function (map). Close #4362.

### LINKED ISSUE #4362: dashArray on canvas uses ES5 functions (map)
See https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L79

`map` and other ES5 functions (`forEach`, `reduce`, `filter`, etc)  are not supported by IE8, which Leaflet still supports.

perliedman:


IvanSanchez:
LGTM

Issue #4226, opened 2016-02-12
flyTo broken when start latlng == end latlng and zoom diff >= 10


hyperknot:
flyTo goes in an infinite loop when start latlng == end latlng and zoom level differences are >= 10.

Reproducing:
In debug/map/zoompan.html, change the initial setView to level 14, and dc onclick level to 4. Click DC.

Leaflet goes in an infinite loop of console logging: 
`LatLng.js:7 Uncaught Error: Invalid LatLng object: (NaN, NaN)`

It is interesting that a given zoom level difference is required, at least 10. So 4-14 triggers, 5-15 triggers, 5-14 doesn't.

IvanSanchez:
@hyperknot Can you make a unit test for this? You can base it on the "tile count during flyto" tests

hyperknot:
@IvanSanchez yes

mourner:
Let's try a better workaround that would not limit `flyTo` range: https://github.com/Leaflet/Leaflet/pull/4376#issuecomment-204845988

hyperknot:
I just wanted to note here, that I was unclear before: the fix doesn't limit flyTo in general, it only makes a "jump" in the case when the browser would have gone to an infinite loop. Any other case is unchanged, including the 10+ zoom level animations.

I'll be looking into the u1 check, as linked by @mourner:
https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/camera.js#L636-L645

ZumelzuR:
> I just wanted to note here, that I was unclear before: the fix doesn't limit flyTo in general, it only makes a "jump" in the case when the browser would have gone to an infinite loop. Any other case is unchanged, including the 10+ zoom level animations.
> 
> I'll be looking into the u1 check, as linked by @mourner:
> https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/camera.js#L636-L645

404 for your link. Can you put a solution here pls?

johnd0e:
> 404 for your link. Can you put a solution here pls?

https://github.com/mapbox/mapbox-gl-js/blob/9846a4f6d54bd09dfba5c655c56e11d4484458f9/js/ui/camera.js#L635-L645

PR #4376, opened 2016-04-02
Workaround for flyTo bug

hyperknot:
1. Made a test for checking the flyTo bug
2. Changed the old flyTo test to use a visible div, as it is required for flyTo animation.
3. Split the long line of r() inside flyTo to multiple lines.
4. Added workaround for floating point precision error in flyTo.
5. Changed the zoompan example to trigger this bug when clicking on DC.

This workaround now effectively limits the animation of flyTo to maximum 9 zoom levels, after that it jumps at the end. Still this is much better then making the browser go into an infinite loop.

Proper fix would be to change the r() function not to use a tiny number and then take it's logarithm.
cc: @mourner

### DIFF
--- a/debug/map/zoompan.html
+++ b/debug/map/zoompan.html
@@ -56,7 +56,7 @@
 
 		var map = L.map('map', {
             zoomSnap: 0.25
-        }).setView(dc, 10);
+        }).setView(dc, 14);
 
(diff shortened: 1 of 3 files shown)


hyperknot:
#4226

mourner:
I think there is a better workaround for this — we addressed this in GL JS some time ago. See this code https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/camera.js#L636-L645

PR #4377, opened 2016-04-02
Round bounds size.

### LINKED ISSUE #4255: getBounds() + fitBounds() zooms out map
Running a getBounds() and then a fitBounds() might zoom out the map.

It depends on zoom level, map element size, probably some rounding error.

http://playground-leaflet.rhcloud.com/juq/2/edit?html,output

perliedman:
Bounds size should be in pixels, so round it to avoid
any floating point inaccuracies from projection.

Closes #4255.


hyperknot:
Looks good!

PR #4387, opened 2016-04-03
Make drag handler insensitive to order of handlers events

### LINKED ISSUE #4315: Markers move erratically while zooming on mobile
Given a Leaflet map, if you initialize the map, add a circle marker, and invoke the following:

```
map.dragging.disable();
map.dragging.enable();
```

Then when you "pinch-to-zoom" on mobile via touch, the markers on the map move erratically while zoom is in progress, snapping back to the correct position after the zoom is complete. This makes for an ugly user experience.

http://i.imgur.com/O7ZlME5.gif
markers moving on zoom

Replicated on Leaflet latest (1.0.0), on Android Chrome, and Chrome on a MS Surface Pro 3 running Windows 10 (using the touch screen.) Cannot replicate on desktop using a scroll wheel.

See [playground here](http://playground-leaflet.rhcloud.com/nuh/edit?html,console,output).

perliedman:
Draggable handles touch events, and does not rely on
simulated mouse events; under some circumstances, it
even breaks on simulated events (see #4315).

This PR ignores any simulated events in the event handlers,
to just deal with the real events. This makes it insensitive
to if the tap handler has run before, and fired simulated
events.

Close #4315.

### DIFF
--- a/spec/suites/map/handler/Map.DragSpec.js
+++ b/spec/suites/map/handler/Map.DragSpec.js
@@ -192,6 +192,64 @@ describe("Map.Drag", function () {
 			toucher.wait(100).moveTo(200, 200, 0)
 				.down().moveBy(1, 0, 20).moveBy(1, 0, 200).up();
 		});
+
+		it.skipIfNotTouch('reset itself after touchend', function (done) {
+
+			var container = document.createElement('div');
+			container.style.width = container.style.height = '600px';
+			container.style.top = container.style.left = 0;
+			container.style.position = 'absolute';
+			document.body.appendChild(container);
+			var map = new L.Map(container, {
+				dragging: true,
+				inertia: false,
+				zoomAnimation: false	// If true, the test has to wait extra 250msec
+			});
+			map.setView([0, 0], 1);
+
+			// Change default events order to make the tap comming before the touchzoom.
+			// See #4315
+			map.dragging.disable();
+			map.dragging.enable();
+
+			var center = map.getCenter(),
+			zoom = map.getZoom();
+
+
+			var mouseHand = new Hand({
+				timing: 'fastframe',
+				onStop: function () {
+					document.body.removeChild(container);
+					expect(map.getCenter()).to.eql(center);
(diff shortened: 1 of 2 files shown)


yohanboniface:
Related: https://github.com/Leaflet/Leaflet/pull/3872#issuecomment-146169261

yohanboniface:
The more we go, the more I feel uncomfortable with having both a native event and a simulated one sent in the same time. I'm wondering then if the "proper" fix is not a bit before in the chain, trying to send either the native event or the simulated one.
Cf also https://github.com/Leaflet/Leaflet/pull/3872#issuecomment-146169261 that I just pasted above.

Also, what about making a *prostetic* unit test for this, do you feel it's doable and can be isolated enough without being to much coupled (given some events are simulated, if we simulate the simulation, what are we testing? :p) ?

perliedman:
Wow, yeah your comment was indeed very spot on, it more or less exactly describes this issue:

> when disabling/reenabling dragging, mousedown comes first, because this time L.Map.Tap has precedence on event callback order

As it is now, only the first touch from then touch event is passed on to the mouse event (https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.Tap.js#L101-103). As far as I could understand, that might be the core of at least this particular issue. How about passing along all touches in the simulated event? I think (but have not tested) this would fix this issue. What other consequences would it have, is there good reasoning behind just passing one touch?

yohanboniface:
@perliedman any chance we can remote pair on this during the week-end so we can proceed on rc1? :)

perliedman:
@yohanboniface sounds like a good idea! I think I will be able to set aside time for this on friday (15th of april).

perliedman:
@yohanboniface sorry, I obviously didn't read that sentence fully :) I think I'll have a hard time scheduling time tomorrow, unfortunately :(

yohanboniface:
@perliedman that's ok, let's plan a session on Friday then :)
@IvanSanchez @hyperknot @danzel @mourner Leaflet day on Friday April 15th, book your agenda! :)

Issue #3939, opened 2015-10-15
Make playground official


yohanboniface:
IMHO, [playground](http://playground-leaflet.rhcloud.com/) has proven to be a very helpful tool for Leaflet maintenance, and I do think we should make it stable and official.

Here are the actions I think that need to be done:
- [ ] have an official server
- [ ] have a domain easy to remember (I suggest play.leafletjs.org)
- [ ] add major plugins to the list of libraries
- [ ] move the code to Leaflet Github organisation
- [ ] have a deploy process so that any of us can push update of the code

@IvanSanchez please amend as you are the playground master. :)

@IvanSanchez @mourner thoughts?

mourner:
I agree. Before we move forward on this, is JSBin the best platform for this? Are there better/simpler alternatives? JSBin has some weird quirks... Just asking to make sure.

IvanSanchez:
- [x] Decipher a way to persist JSBin's «session secret» so that bins are kept across server resets and app re-deployments in rhcloud.
- [ ] Clean up the code from IvanSanchez/jsbin into a PR that @jsbin can have a look at, but just the bits about deployment to OpenShift (npm scripts, app startup, env variables), not the bits about Leaflet customization (logo, libraries, html stuff).
- [ ] Write up `docs/rhcloud.md`.
- [ ] Send even more kudos to @jsbin.
- [ ] Check if the app can be accessed by a DNS alias (maybe openshift does routing magic), or if JSBin needs some persistent config for that (more env vars???), or openshift wants DNS alias itself.

IvanSanchez:
@mourner I did a bit of research, and JSBin seemed to be the best option - there is no public code for jsfiddle, for example. Even if it needed some hacks to get it up and running "in da cloud".

iH8:
As far as i know is http://plnkr.co/ completely opensource and imo a much better option then JSbin.  I could be wrong tho because i'm not aware of your requirements. It does everything JSbin does and much, much more. They're currently working on Plunker.NEXT which has even more awesome functionality like running repo's and gists from GIT and much more.

PR #4392, opened 2016-04-05
Change simple test case link to the Leaflet JS Bin

danzel:
This is a much easier way to make a reproducible test case rather than remembering all the things you need to do.


yohanboniface:
See also #3939

Issue #2356, opened 2014-01-08
Glitchy marker clicks / taps in Windows.


springlab:
I'm building an app using Mapbox / Leaflet in Windows. 

When you mousedown on a marker and move slightly (even 1px) it will not activate a marker click on mouseup. Same happens with emulated touch events and real touch input.

I need some sort of click tolerance similar to the tap tolerance. But I'm not sure how to implement that.

To reproduce the issue, open any Mapbox / Leaflet map in Chrome on Windows. Mouse down on a marker, move the mouse just a hair, then release.
Also, trying to tap on a marker using a Windows touch screen rarely results in marker activation. You usually have to double tap to fire the popup.

Tested with Chrome / Chromium in Windows 8. Also testing with Mouse, Emulated Touch and real Touch Events.

Any help is greatly appreciated.

mourner:
Can anyone confirm this being fixed on master?

HarelM:
I'm not sure this is related to this issue or not, but I can open a new one if needed.
I have an issue with firefox on android device and the code throws an error on the following line:
```js
// this is a horrible workaround for a bug in Android where a single touch triggers two click events
	_filterClick: function (e, handler) {
		var timeStamp = (e.timeStamp || e.originalEvent.timeStamp),
```
the error is: ```TypeError: e.originalEvent is undefined```
The steps to reproduce are:
surf to: http://israelhiking.osm.org.il on android with firefox, click on the marker on the right pane.
click on the map to add a marker.
click the "V" to close it and click on the marker again to open popup.
runs well on any device/browser except this combination (popup doesn't open).
uses leaflet 7.7.
Any thoughts?
PS, great library.

HarelM:
OK, Managed to create a fully reproducible scenario:
Click doesn't work for divIcon in firefox for android:
steps to reproduce: click once to add the icon, click on the icon.
The code below works on firefox on desktop but doesn't work on firefox on android.
```html
<!DOCTYPE>
<html>
<head>
	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
	<link rel="stylesheet" href="https://npmcdn.com/leaflet@0.7.7/dist/leaflet.css" />
	<script src="https://npmcdn.com/leaflet@0.7.7/dist/leaflet.js"></script>
</head>
<body>
	<div id="map" style="position: absolute;top: 0;bottom: 0;right: 0;left: 0;overflow: hidden;"></div>
[...]

PR #4396, opened 2016-04-06
Added click tolerance also for non-touch devices

DavidUv:
There used to be a default click tolerance value of 3 (hard coded) and it was only applicable for touch devices. In my experience the problem of an intended click event being registered as a drag event also exists for non-touch devices.
This PR removes the test for touch devices and adds a new option for the tolerance.

### DIFF
--- a/spec/suites/map/handler/Map.DragSpec.js
+++ b/spec/suites/map/handler/Map.DragSpec.js
@@ -22,7 +22,7 @@ describe("Map.Drag", function () {
 			expect(map.dragging.enabled()).to.be(true);
 		});
 		it("calls the map with dragging disabled and worldCopyJump enabled; " +
-		"enables dragging after setting center and zoom", function () {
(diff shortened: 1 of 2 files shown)


danzel:
I haven't tested this, but this should fix #2356 I think maybe?

yohanboniface:
@DavidUv thanks for the PR! Any motivation to add one touch/one non-touch unittest to cover that? Using prostetic-hand, it should be quite easy (see examples in the other specs).

DavidUv:
@danzel Yep. I had the same problem.
@yohanboniface Sure,I'll have a go at it.

yohanboniface:
Thanks!
For the record, this PR removed the check on `L.Browser.touch` for click tolerance control, which in any case does not make really sense: we were checking about the browser touch support, but not if the current event was an actual *touch* one.

PR #4397, opened 2016-04-06
Make fitBounds less sensitive to floating point rounding errors

### LINKED ISSUE #4395: Initial fitBounds regression
#4377 introduced a regression to the initial fitBounds method. By initial I mean when run on a map which has not yet received a zoom level + center, but is using fitBounds to set the view.

The result is that the view is sometimes off, sometimes totally broken with NaN bounds exception.

The reason is that in getBoundsZoom, for zoom level 0, boundsSize has values like:
```
boundsSize = L.Point {x: 0.08415031433105469, y: 0.029537200927705953},
```

Now with the added _rounds() this gets rounded to 0, 0. 

I recommend the following:
1. revert b511c7bc
2. investigate if this affects non-0 zoom levels, or only 0.
3. find a way to solve this
4. come up with test cases for fitBounds on initial map load.

cc: @perliedman

perliedman:
Reverts _round() fix from b511c7bcc8aae025fcad6169a06043c99da0cf75.
Close #4395.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -822,6 +822,36 @@ describe("Map", function () {
 			});
 			map.fitBounds(bounds, {animate: false});
 		});
+
+		it('Fits to small bounds from small zoom', function (done) {
+			map.once('zoomend', function () {
+				map.once('zoomend', function () {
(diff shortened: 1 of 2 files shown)


yohanboniface:
Sounds good to me :)

PR #4402, opened 2016-04-07
fitBounds round fix

hyperknot:
I think this is the right fix for the fitBounds jumping bug.

1. The jump actually happens only inside the snap part (with zoomSnap: 0, there is no jumping).
2. It is because of the Math.floor function: zoom level 14.9999 becomes 14.
3. This fix rounds to the nearest snap level, if within 1% of it. So 14.9999 becomes 15 and then floor and ceil doesn't make a jump.


perliedman:
Yes, this is a much better approach! :+1:

PR #4404, opened 2016-04-08
Fix regression when setting style of Path with canvas renderer

### LINKED ISSUE #4401: Exception in L.Canvas _updateStyle function
I am getting an exception in the _updateStyle function of the Canvas Renderer
I have traced the exception to the call _updatedDashArray().  The _updateStyle is calling _updateDashArray with out any parameters, when the function expects the layer to be passed in.
This causes the "TypeError: Cannot read property 'options' of undefined" to be thrown.

```javascript
    _updateStyle: function (layer) {
		this._updateDashArray(); //this expects a parameter of layer
		this._requestRedraw(layer);
	},

   _updateDashArray: function (layer) {
		if (layer.options.dashArray) {
			var parts = layer.options.dashArray.split(','),
			    dashArray = [],
			    i;
			for (i = 0; i < parts.length; i++) {
				dashArray.push(Number(parts[i]));
			}
			layer.options._dashArray = dashArray;
		}
	},
```
I am using Leaflet 1.0.0-beta.2 ("139340e")

perliedman:
Pass layer on when calling _updateDashArray
Add specs to verify intialization and setStyle with dashArray.
Close #4401.

### DIFF
--- a/spec/suites/layer/vector/CanvasSpec.js
+++ b/spec/suites/layer/vector/CanvasSpec.js
@@ -98,4 +98,18 @@ describe('Canvas', function () {
 
 	});
 
+	describe('#dashArray', function () {
+		it('can add polyline with dashArray', function () {
+			var layer = L.polygon(latLngs, {
+				dashArray: "5,5"
+			}).addTo(map);
+		});
+		it('can setStyle with dashArray', function () {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
LGTM. Nice to have tests for this.

PR #4405, opened 2016-04-08
update default Control position to match leafdoc

sheppard:
It looks like after merging the :fallen_leaf:doc changes, the Zoom control is appearing on the `topright`.  I fixed this by adjusting the default position on `L.Control` to match what :fallen_leaf: doc says is the default.  However, I wonder if it might be best to also restore an explicit `position: "topleft"` in `L.Control.Zoom`.

### DIFF
--- a/src/control/Control.js
+++ b/src/control/Control.js
@@ -13,7 +13,7 @@ L.Control = L.Class.extend({
 		// @option position: String = 'topleft'
 		// The position of the control (one of the map corners). Possible values are `'topleft'`,
 		// `'topright'`, `'bottomleft'` or `'bottomright'`
-		position: 'topright'
+		position: 'topleft'
 	},
 
 	initialize: function (options) {


IvanSanchez:
Shouldn't we change the documentation instead? Just wondering.

sheppard:
I'm fine either way as long as we get the zoom control back to the top left (in master it's on the top right).  This PR is essentially a followup to 560e73bac54624df89d01be45cf0a77bc09f94fe - which had two relevant changes:

 - Removed `position: "topleft"` from `Control.Zoom.js`, making it fall back to the default in `Control.js`
 - Added :fallen_leaf:doc to `Control.js` indicating the default was `topleft` (but left the actual default as `topright`)

Taken together, I assumed that you had intended for `topleft` to become the new default, which makes sense to me.  For backwards compatibility, however, it might make sense to keep `topright` as the default in `Control.js` (changing the docs) and restore the `topleft` override in `Control.Zoom.js`.

mourner:
I think we should keep the default behavior. The Leafdoc commit wasn't intended to change the code I think.

hyperknot:
Just to clarify it (wasn't clear to me till I checked it out):
- 0.7, beta1, beta2 had zoom control on top left by default
- master has zoom control on top right by default now

So task is to move master back to top left, is that right?

perliedman:
Yes, default for `L.Control.Zoom` should be `topleft`, and default for `L.Control` should be `topright`. Changing the defaults would likely cause trouble or at least annoyance when upgrading.

yohanboniface:
If we want to avoid any breaking change, my suggestion is to:
- keep `L.Control` defaulting to `topright`, and update the doc accordingly
- add back `position: topleft` for `L.Control.Zoom`

@sheppard are you around to make that change or should I go for it? We indeed need to fix that for rc-1 release.

yohanboniface:
@perliedman race condition :p

Issue #3471, opened 2015-05-14
composer.phar support


fmiguelroliveira:
Hello,

I was wondering if the leafletjs team have in mind to support leafletjs by the composer.phar .

This would be great for lots of developers who already have 3rd party components managed by composer.

IvanSanchez:
I don't think any of the core devs use composer.phar. I, for one, don't want to get involved in a flamewar about NPM vs Bower vs Composer.

I suggest you make a pull request adding the composer.phar files to Leaflet's build system - that will get this done far quicker.

PR #4409, opened 2016-04-11
install composer

Preovaleo:
HI,

i make a composer.json file
and this solve #3471

what to do next:
- [ ]  enable the package on packagist (i can do that too)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,18 +21,18 @@
     "issues": "https://github.com/Leaflet/Leaflet/issues",
     "forum": "https://stackoverflow.com/questions/tagged/leaflet",
     "source": "https://github.com/Leaflet/Leaflet",
-    "docs": "http://leafletjs.com/reference.html",
-    "extra": {
-      "component": {
-        "scripts": [
-          "dist/leaflet-src.js"
-        ],
-        "files": [
-          "dist/leaflet.js",
-          "dist/leaflet.css",
-          "dist/images/*.png"
-        ]
-      }
+    "docs": "http://leafletjs.com/reference.html"
+  },
+  "extra": {
+    "component": {
+      "scripts": [
+        "dist/leaflet-src.js"
+      ],
+      "files": [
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -0,0 +1,38 @@
...
+  "author": {
+    "url": "http://agafonkin.com/en/",
+    "name": "Vladimir Agafonkin"
+  },
+  "support": {
+    "issues": "https://github.com/Leaflet/Leaflet/issues",
+    "forum": "https://stackoverflow.com/questions/tagged/leaflet",
+    "source": "https://github.com/Leaflet/Leaflet",
+    "docs": "http://leafletjs.com/reference.html",
+    "extra": {
```
Preovaleo: not in support

mourner:
Yeah, we don't want to support every possible package manager out there and have hundreds of files like this in the repo root.

pvhee:
@mourner I'd argue against this saying `composer` isn't just every other package manager, but widely used within the PHP community, and other PHP-dependant communities like Drupal / Symfony / etc. are relying increasingly on composer.

Leaflet is widely used within the PHP community, so I don't see why this couldn't be added to this repo? You already do support `bower` and `npm`, so you do support package managers.

If anyone else is looking for Composer support for Leaflet, you might want to take a look at https://github.com/mrohnstock/leaflet

blackice2999:
> we don't want to support every possible package manager

but you should support the most relevant package managers... the consequence are badly manages forked repositories this couldnt be your interest.

tormi:
As a workaround, you can declare leaflet package by yourself in your project's `composer.json`, but that's not the simplest way:

```json
  "repositories": [
  {
    "type": "package",
    "package": {
      "name": "Leaflet/Leaflet",
      "version": "1.0.2",
      "dist": {
        "url": "https://github.com/Leaflet/Leaflet/archive/v1.0.2.zip",
        "type": "zip"
      }
    }
  }
]
```

PR #4418, opened 2016-04-14
Consistent return type with aliased method

nathancahill:
Evented.listens() returns a boolean, so hasEventListeners() will as well


yohanboniface:
Good catch, thanks :)

PR #4424, opened 2016-04-14
Uncomment maxZoom option in L.GridLayer

nathancahill:
Not sure why this was commented out in @IvanSanchez's doc fixes. It will be undefined regardless.


perliedman:
@IvanSanchez any particular reason you commented this out? Is this how we want to normally handle options where falsy is the default?

IvanSanchez:
IIRC, the option  was commented out in the first place, as it is initialized somewhere else in the code (or can be left uninitialized).

IvanSanchez:
See https://github.com/Leaflet/Leaflet/commit/560e73bac54624df89d01be45cf0a77bc09f94fe#diff-f1e6be67599c594731fff6191c710420L21

PR #4426, opened 2016-04-14
Uncomment crossOrigin option in L.ImageOverlay

nathancahill:
The check that's used later is `if (this.options.crossOrigin)`, which currently evaluates to undefined. Setting the default option to false keeps the typing consistent.

### DIFF
--- a/src/layer/ImageOverlay.js
+++ b/src/layer/ImageOverlay.js
@@ -31,12 +31,11 @@ L.ImageOverlay = L.Layer.extend({
 
 		// @option attribution: String = null
 		// An optional string containing HTML to be shown on the `Attribution control`
-		attribution: null
+		attribution: null,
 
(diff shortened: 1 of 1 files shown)


perliedman:
I would expect this to be a pretty common pattern in the code, but explicitly stating options in the code might be a good idea, especially now that we have docs inside the code.

Out of curiosity, how did you spot this?

nathancahill:
Explicit is definitely a good idea. I didn't see this pattern anywhere else, and went through the options pretty thoroughly. I spotted this and #4424 by automatically generating a Typescript file from the docstrings. Caught a lot of inconsistent types and values. These two were the only ones that were actually commented out.

The only reason I can think for not explicitly state undefined or falsy default values is file size, which I think is negligible here?

PR #4433, opened 2016-04-14
Change parameter type in docstring from |null to optional

nathancahill:
createIcon and createShadow both take an optional oldIcon parameter that should probably not be null for consistent types. It's checked with `||`, so undefined will work here.


perliedman:
Great! :sunny:

PR #4437, opened 2016-04-14
Add LineUtil.clipSegment docstring

nathancahill:


perliedman:
Hi and thanks for making the docs more complete!

For `clipSegment`, it looks like the return type shouldn't be `void`, it returns the clipped segment points, and the return value is important, see https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Polyline.js#L239.

nathancahill:
Opps, totally missed the one-line return values. What do you think about opening these up for readability as well?

```javascript
if (!(codeA | codeB)) { return [a, b]; }

if (codeA & codeB) { return false; }
```

would become:

```javascript
if (!(codeA | codeB)) {
    return [a, b];
}

if (codeA & codeB) {
    return false;
}
```

nathancahill:
Additionally, consistent return types would be good. Returning an empty list `[]` instead of `false` would still be falsy in the `if (!segment)` check. Then the return type would be `Point[]` instead of `Point[]|boolean`

mourner:
FYI empty arrays are truthy.

```js
!![] // true
```

nathancahill:
Ahh, of course, you're right. In that case, ignore that second thought.

nathancahill:
Updated with correct return types and expanded return statements.

PR #4442, opened 2016-04-15
Fixing hover/mouseover for circle/circleMarker with Cavas renderer

computermacgyver:
I ran into a problem using the canvas render. With SVG, it is mouseover on ``L.CircleMarker`` or ``L.Circle`` works just fine. However, with ``L.Canvas`` the mouseover event will not fire. 

Example:

```
var oxf = [51.751944, -1.257778];
var map = L.map('mapid',{preferCanvas:true}).setView(oxf, 5);
L.CircleMarker(oxf, 
	{"radius":10, fill: true, fillColor: "#00FF00", fillOpacity: 1, stroke: false})
	.on("mouseover click",function(e){console.log(e)})
	.addTo(map);
```

I traced this to the fact that in ``L.Canvas._updateCircle`` does not add the circle/circleMarkers to the ``_drawnLayers`` object.

This pull requeset adds that code so that mouseover behaves consistently between SVG and Canvas. I have not seen an unexpected behaviour or ill-effects from this change, but I defer to you if there is a good reason not to do this.

Thank you for a great library!

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -198,6 +198,8 @@ L.Canvas = L.Renderer.extend({
 
 		if (layer._empty()) { return; }
 
+		this._drawnLayers[layer._leaflet_id] = layer;
+		
 		var p = layer._point,
 		    ctx = this._ctx,
 		    r = layer._radius,


yohanboniface:
@computermacgyver seems there is a linting issue, can you fix it?

Also, it seems to be working for me with canvas: http://playground-leaflet.rhcloud.com/moh/edit?html,console,output I see the mouseover event event to much, actually! :/

computermacgyver:
Thanks @yohanboniface  
The Travis error is just a trailing space, which can be easily fixed.

Very strangely, the [CDN hosted file](https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/master/leaflet-src.js) (``Leaflet 1.0.0-beta.2 (ce9b9f5)``) you used in the example has this change in it already:

<pre>
_updateCircle: function (layer) {

		if (layer._empty()) { return; }

		var p = layer._point,
		    ctx = this._ctx,
		    r = layer._radius,
		    s = (layer._radiusY || r) / r;

		<b>this._drawnLayers[layer._leaflet_id] = layer;</b>

		if (s !== 1) {
			ctx.save();
			ctx.scale(1, s);
		}

		ctx.beginPath();
		ctx.arc(p.x, p.y / s, r, 0, Math.PI * 2, false);

		if (s !== 1) {
			ctx.restore();
		}

		this._fillStroke(ctx, layer);
	},
</pre>

The [file I download](http://cdn.leafletjs.com/leaflet/v1.0.0-beta.2/leaflet.zip) from [leaflet's download page](http://leafletjs.com/download.html) is a slightly different version (```Leaflet 1.0.0-beta.2 (954e83d)```) and **does not** have this change.

computermacgyver:
Actually looking back at the source code, in the master branch, I see that this change is already in the repository. So, this pull request should be closed.

The remaining question is why the download on the leaflet website doesn't have this? Using the CDN would have saved me time tracking down why this didn't work, but all OK, I guess. Perhaps the version on the homepage should be update to save anyone else form repeating this.

yohanboniface:
> The remaining question is why the download on the leaflet website doesn't have this?

It has it if you download the "dev" version instead of the "beta2" ;)

computermacgyver:
Ahh, I was thrown off because the header in both files reads ``Leaflet 1.0.0-beta.2``, but I see now they are actually very different! Thanks.

PR #4444, opened 2016-04-15
Switch options type from path to polyline for polygon and rectangle

### LINKED ISSUE #4427: Path options vs Polyline options
Polyline options adds `smoothFactor` and `noClip` to Path options. Since Polygon and Rectangle all follow the code paths in Polyline that use those options (`_simplifyPoints()` and `_clipPoints()`), it would make sense to switch the documentation from `<Path options> options?` to `<Polyline options> options?`. Otherwise it would seem like Polygon and Rectangle don't have those options.

The downside is that the link in the docs wouldn't go to [Path options](http://leafletjs.com/reference-1.0.0.html#path-option), and the Polyline inherited options could be missed.

I can submit a pull request for this.

nathancahill:
Fixes #4427

### DIFF
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -140,7 +140,7 @@ L.Polygon = L.Polyline.extend({
 });
 
 
-// @factory L.polygon(latlngs: LatLng[], options?: Path options)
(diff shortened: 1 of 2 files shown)


yohanboniface:
LGTM, thanks :)

PR #4446, opened 2016-04-15
Fix typo in flyTo docstring options type

nathancahill:


yohanboniface:
You may want to just read the doc and issue a whole pull request after then, at some point ;)

nathancahill:
You're right. I keep thinking I've found them all, and then another pops up 🕵

PR #4455, opened 2016-04-18
Build scripts: stop using magic-string and use git rev as semver metadata

IvanSanchez:
As @perliedman pointed out in the gitter chat, sourcemaps for -rc1 are broken and point to line 1 of all files (besides, they refuse to be loaded with sourcemap debugging tools like https://github.com/sokra/source-map-visualization).

I replaced the code for generating sourcemaps from the simplistic `magic-string` library to the more raw `source-map` library, by copy-pasting some code from my modifications to `gobble-concat`.

---

After pondering with the release scripts, and after seeing a lot of bug reports of leaflet between beta2 and rc1 with the same value of `L.version`, I thought it would be a good idea to add "build metadata" as per the semver specs.

These build scripts will look in `package.json` for the version, then append the git rev automatically on `jake build`. e.g. `L.version = "1.0.0-rc.1+75dda76"`

This can be disabled by running `jake build[,,true]` (which is exactly what `publish.sh` does).

Hopefully this would help track "unofficial" builds, and tell the git rev of those.

### DIFF
--- a/Jakefile.js
+++ b/Jakefile.js
@@ -14,7 +14,7 @@ For a custom build, open build/build.html in the browser and follow the instruct
 
 var build = require('./build/build.js'),
     buildDocs = require('./build/docs'),
-    version = require('./src/Leaflet.js').version;
+    git = require('git-rev');
 
 function hint(msg, args) {
 	return function () {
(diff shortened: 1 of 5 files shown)


perliedman:
Both very nice changes! Checked them out locally and verified version looks correct, and source maps work as expected.

:+1:

PR #4459, opened 2016-04-19
Sourcemaps: replace magic-string with source-map.SourceNode

IvanSanchez:
As @perliedman pointed out in the gitter chat, sourcemaps for -rc1 are broken and point to line 1 of all files (besides, they refuse to be loaded with sourcemap debugging tools like https://github.com/sokra/source-map-visualization).

I replaced the code for generating sourcemaps from the simplistic `magic-string` library to the more raw `source-map` library, by copy-pasting some code from my modifications to `gobble-concat`.

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -1,7 +1,7 @@
 var fs = require('fs'),
     UglifyJS = require('uglify-js'),
     zlib = require('zlib'),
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Superseded by #4455

PR #4460, opened 2016-04-19
Better handling of markerZoomAnimation event hooks

### LINKED ISSUE #4453: Remove marker Bug if map.markerZoomAnimation = false
Only tested on Master.
If you initialize the map with markerZoomAnimation = false the map will fire the event zoomanim even for the removed markers when you zoom in and out.

https://playground-leaflet.rhcloud.com/juzu/1/edit?html,js,output

Zoom in and out on that map for 10 seconds, no problems.
After 10 seconds the marker is removed.
Now if you try to zoom you get an error in the console: `error Script error. (line 0)`
I don't know how to link proper errors from that website, but on my website this is the error i get in the console:
`'Uncaught TypeError: Cannot read property '_latLngToNewLayerPoint' of null` @ https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L290`

I think this is happening because when we add a marker this is happening:
https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L86 is True so
https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L124 gets set.
so it's registering zoomanim  https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L89
Then https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L101 is False
But when we remove the marker this is what is going on:
https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L171 is going to:
https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L123 which is false
so it's not removing the handler for the  zoomanim event.

IvanSanchez:
Fixes #4453. This is two-fold:

* `L.Layer` shall remove the same layer events that it attached. No longer assume that two calls to `L.Layer.getEvents()` will return the same values. (not 100% needed for this bug, but it is a sane change)
* `L.Marker` needs to know a map option before deciding whether to attach a `zoomanim` event handler or not. Manual handling on `onAdd` and `onRemove` is thus needed.

Despite my efforts, I couldn't make a failing test case before the bugfixes, so no new unit tests :-(

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -86,7 +86,11 @@ L.Layer = L.Evented.extend({
 		this._zoomAnimated = map._zoomAnimated;
 
 		if (this.getEvents) {
-			map.on(this.getEvents(), this);
+			var events = this.getEvents();
+			map.on(events, this);
+			this.on('remove', function () {
+				map.off(events, this);
+			}, this);
 		}
 
 		this.onAdd(map);
@@ -165,10 +169,6 @@ L.Map.include({
 
 		if (layer.getAttribution && this.attributionControl) {
 			this.attributionControl.removeAttribution(layer.getAttribution());
-		}
-
-		if (layer.getEvents) {
-			this.off(layer.getEvents(), layer);
 		}
 
 		delete this._layers[id];
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -88,7 +88,7 @@ L.Layer = L.Evented.extend({
 		if (this.getEvents) {
 			var events = this.getEvents();
 			map.on(events, this);
-			this.on('remove', function () {
+			this.once('remove', function () {
 				map.off(events, this);
 			}, this);
 		}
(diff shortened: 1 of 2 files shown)


src/layer/Layer.js
```diff
@@ -86,7 +86,11 @@ L.Layer = L.Evented.extend({
 		this._zoomAnimated = map._zoomAnimated;
 
 		if (this.getEvents) {
-			map.on(this.getEvents(), this);
+			var events = this.getEvents();
+			map.on(events, this);
```
yohanboniface: `once`?

src/layer/Layer.js
```diff
@@ -86,7 +86,11 @@ L.Layer = L.Evented.extend({
...
+			this.on('remove', function () {
```
yohanboniface: I don't get this change, actually.

mourner:
Any update on this?

balciseri:
looks good to me

PR #4462, opened 2016-04-19
Add 'updateWhenZooming' option to L.GridLayer

IvanSanchez:
Normally, we update grid layers on every integer zoom level to make the `flyTo()` animations smooth and nice, and also applies to the touch-zoom interaction.

However, this is a problem on gridlayers which take time to render - I'm talking a [`Leaflet.VectorGrid`](https://github.com/IvanSanchez/Leaflet.VectorGrid) with a few megabytes of `geojson-vt`ed data. These can take a second or two to render on a mobile device, and updating them during a touch-zoom makes the UX very sluggish.

I hereby propose a new `L.GridLayer` option, dubbed `updateWhenZooming`.

(Bonus: leafdoc documentation means *one* pull request instead of two!)


mourner:
Sounds good to me.

PR #4464, opened 2016-04-19
Allow point and latlng expressions in all methods

nathancahill:
Found a couple methods that only accept L.Point and L.LatLng instances, instead of the expressions that all other methods accept ([number, number], {lat: number, lng: number}, etc). Only exception is LineUtil, so added a note in the docstring that only real L.Points are accepted.


IvanSanchez:
Not sure we want to do this - some of the methods (`setPosition`, `project`) get called thousand of times, so this might imply a performance decrease.

IvanSanchez:
Merge conflict (from the ES6 modules syntax) is too big, closing for now.

PR #4465, opened 2016-04-19
Create points from objects with x and y properties

nathancahill:
Standardizes L.point() with L.latLng(). L.latLng accepts `{lat: Number, lng: Number}`, now L.point accepts `{x: Number, y: Number}`


perliedman:
Thanks, this is a small but nice improvement!

PR #4466, opened 2016-04-19
Add CRS.distance docstring

nathancahill:
Undocumented at the moment.

### DIFF
--- a/src/geo/crs/CRS.js
+++ b/src/geo/crs/CRS.js
@@ -73,6 +73,9 @@ L.CRS = {
 		return L.bounds(min, max);
 	},
 
+	// @method distance(latlng1: LatLng, latlng1: LatLng): Number
+	// Returns the distance between two geographical coordinates.
+
 	// @property code: String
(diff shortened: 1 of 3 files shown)


src/geo/crs/CRS.js
```diff
@@ -73,6 +73,9 @@ L.CRS = {
 		return L.bounds(min, max);
 	},
 
+	// @method distance(latlng1: LatLng, latlng1: LatLng): Number
+	// Returns the distance between two geographical coordinates.
```
perliedman: Would be great to add that the returned value is in meters!

nathancahill: distance is not necessarily in meters, only for L.CRS.Earth and it's inheritants. CRS.Simple is just euclidean distance.

nathancahill: Added notes about distance() return values to each of those descriptions.

perliedman: Right! I should have known that! :smile:

PR #4467, opened 2016-04-19
className option for TileLayers

jayvarner:
It would be nice to be able to add an arbitrary class to a TileLayer like you can on vector layers.


IvanSanchez:
This is a nice feature to have, but:

* The comment is superfluous and should be removed
* The new option should be documented alongside the rest of the options (see https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L65 )

Can you fix those and amend the commit?

jayvarner:
@IvanSanchez, done. Thanks for the feedback.

src/layer/tile/GridLayer.js
```diff
@@ -311,7 +315,8 @@ L.GridLayer = L.Layer.extend({
 	_initContainer: function () {
 		if (this._container) { return; }
 
-		this._container = L.DomUtil.create('div', 'leaflet-layer');
+		this._container = L.DomUtil.create('div', 'leaflet-layer ' + this.options.className);
```
IvanSanchez: This might evaluate to the string `"leaflet-layer undefined"` or `"leaflet-layer false"` in some edge cases. See https://github.com/Leaflet/Leaflet/blob/master/src/layer/Popup.js#L288 and https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Icon.js#L108

src/layer/tile/GridLayer.js
```diff
@@ -101,7 +101,7 @@ L.GridLayer = L.Layer.extend({
 
 		// @option maxZoom: Number = undefined
 		// The maximum zoom level that tiles will be loaded at.
-//		maxZoom: undefined,
+		// maxZoom: undefined,
```
IvanSanchez: We might want to remove the comment here, or else it will show up in the documentation as a continuation of the previous line.

jayvarner:
@IvanSanchez thanks again for your feedback. I made the adjustments you suggested. I added the `.trim()` to clean up the extra space if someone doesn't supply a `className`. I realize it's not the prettiest thing. I'm open to suggestions and/or willing to find a better way. Also, I rebased my local fork and it added a lot of noise and I'm not sure how to fix that now. Maybe close this one and I'll do a fresh PR so it's clear what's actually changing here?

IvanSanchez:
@jayvarner There's no need to `trim()` the `className` - browsers will do that for you, see [the spec for class names](https://developer.mozilla.org/en-US/docs/Web/API/Element/className). Adding a extra space is fine, as the `L.Popup` code shows.

And yes, you should rebase your local fork/branch against the Leaflet master, that should clear things up.

jayvarner:
@IvanSanchez, I removed the trim() and fought through the rebase but it's just gotten really messy. Since I'm only changes 3 lines in one file, I'm happy to make a new fork and PR and reference this thread in the new PR. Whatever you think is best. Sorry about all that :(

IvanSanchez:
I think it'll be best to make a new PR. There's the possibility of doing an interactive rebase to make the commits look pretty, but that can be time-consuming if you never have done it before.

So I'll close this and wait for a second PR.

PR #4469, opened 2016-04-20
Leafdoc: fix internal links & inconsistent return types

IvanSanchez:
I did a read-through of the docs, and found quite a few uppercase/lowercase inconsistencies and minor typos.

Also, due to a bug in Leafdoc, some of the internal links were not working. After bumping to Leafdoc 1.2.2 (which adds a call to `trim()` these should work fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "karma-mocha": "~0.2.2",
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-safari-launcher": "~0.1.1",
-    "leafdoc": "^1.2.1",
+    "leafdoc": "^1.2.2",
     "mocha": "~2.4.5",
     "phantomjs-prebuilt": "^2.1.7",
     "prosthetic-hand": "^1.3.0",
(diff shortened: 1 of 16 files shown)


nathancahill:
@IvanSanchez, wondering if the return type `void` would be better than `undefined`? It's more widely used. Omitting a return type is also acceptable (as Google's Closure Compiler recommends). I believe that's what most function signatures use now.

IvanSanchez:
@nathancahill No. `void` only exists in C (and C-like) languages. This is javascript, which means `undefined` is a valid [primitive type](https://developer.mozilla.org/en-US/docs/Glossary/Primitive). Furthermore, [if you're `return`ing nothing, you're really `return`ing `undefined`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/return).

nathancahill:
Right, but `undefined` could be left off in the return type, since it's implied?

```
@function cancelAnimFrame(id: Number): undefined
```
vs

```
@function cancelAnimFrame(id: Number)
```

For consistency, most functions don't have return types when they return undefined: https://github.com/Leaflet/Leaflet/blob/master/src/dom/DomUtil.js#L50

PR #4472, opened 2016-04-20
Fix broken page scroll on touch devices

lazopm:
The css rule `.leaflet-container { touch-action: none }` breaks page scrolling for touch devices.
This line was added to fix an Edge zoom issue a year ago.
e44179d

I tested the current version of Edge and I can't reproduce any zoom bugs without the touch-action css rule.

I tested here http://jsfiddle.net/kzfv4jjt/2/

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -17,7 +17,6 @@
 .leaflet-container {
 	overflow: hidden;
 	-ms-touch-action: none;
-	touch-action: none;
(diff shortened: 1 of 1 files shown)


IvanSanchez:
We will probably use #4552 instead anyway, as it seems to make more sense.

mourner:
Yeah, closing in favor of #4552.

PR #4473, opened 2016-04-21
Add missing parameters to L.Util.indexOf docstring

nathancahill:


IvanSanchez:
I'd guess that users should refer to the linked docs, but LGTM

PR #4476, opened 2016-04-21
Fix L.Evented.listens() on removed event handlers, #4474

### LINKED ISSUE #4474: L.Evented: when a listener is removed with it's function, "listens" still returns true
Testcase:
```javascript
var m = new L.Marker();
var handler = function() {console.log("event fired")};

m.on("test", handler);     // register
m.fire("test");            // fires

console.log(m.listens("test"));    // returns true

m.off("test", x);                  // remove listener 
console.log(m.listens("test"));    // still returns true!
m.fire("test");                    // but doesn't fire anymore!
```

https://playground-leaflet.rhcloud.com/rog/edit?html,js,output

IvanSanchez:
Fixes #4474. cc @fab1an

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -237,17 +237,19 @@ describe("Marker", function () {
 			expect(mapSpy.called).not.to.be.ok();
 		});
 
-		it("do not catch event if it does not listen to it", function () {
+		it("do not catch event if it does not listen to it", function (done) {
 			var marker = new L.Marker([55, 37]);
 			map.addLayer(marker);
 			marker.once('mousemove', function (e) {
 				// It should be the marker coordinates
-				expect(e.latlng).to.be.nearLatLng(marker.getLatLng());
+				expect(e.latlng.equals(marker.getLatLng())).to.be.equal(true);
 			});
 			happen.mousemove(marker._icon);
+
 			map.once('mousemove', function (e) {
 				// It should be the mouse coordinates, not the marker ones
-				expect(e.latlng).not.to.be.nearLatLng(marker.getLatLng());
+				expect(e.latlng.equals(marker.getLatLng())).to.be.equal(false);
+				done();
 			});
 			happen.mousemove(marker._icon);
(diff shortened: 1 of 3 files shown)


IvanSanchez:
Hmm. Apparently this breaks an existing unit test.

IvanSanchez:
0836527 was weird. Apparently the `expect` helpers for `nearLatLng` were **never** working, and a race condition in the unit test itself (async calls to `expect()` with no `done()` callback) were skipping the checks and running them out of order.

This looks good to me, but definitely needs another pair of eyes.

src/core/Events.js
```diff
@@ -159,6 +159,9 @@ L.Evented = L.Class.extend({
 					}
 				}
 			}
+			if (listeners.length === 0) {
```
yohanboniface: I'd move this in the previous `if`, because it seems to me that we only want to check `length` if `listeners` is not `undefined`/`null`.

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -237,17 +237,19 @@ describe("Marker", function () {
...
 		});
 
-		it("do not catch event if it does not listen to it", function () {
+		it("do not catch event if it does not listen to it", function (done) {
 			var marker = new L.Marker([55, 37]);
 			map.addLayer(marker);
 			marker.once('mousemove', function (e) {
 				// It should be the marker coordinates
-				expect(e.latlng).to.be.nearLatLng(marker.getLatLng());
+				expect(e.latlng.equals(marker.getLatLng())).to.be.equal(true);
```
yohanboniface: Why `to.be.equal(true)` and not simply `to.be(true)`?
(And why not just `assert e.latlng == marker.getLagLng()`? Ah, we are not in python, sorry ;))

IvanSanchez: Not really sure. We're using a custom version of `expect.js`, and the behaviour of `to.be` and `to.be.equal` and `to.eql` is *funny*. We could see if changing to a NPM package for this would make this any better.

Issue #3666, opened 2015-07-23
Can't disable map drag (Windows IE11 only)


skalb:
I'm trying to disable the map drag when a user clicks a layer. It works fine on Mac/Chrome, but on Windows/IE the map drag is not disabled.

It seems like the mousedown event handler on the map object is never actually removed.

See: http://jsfiddle.net/skalb/otwgyutn/3/

Any ideas?

Edit: using leaflet master

mourner:
also cc @danzel

IvanSanchez:
Variable scope problem. `this` is not set correctly when running `map.dragging.disable()` asynchronously.

IvanSanchez:
Hm. Apparently disabling the map dragging works, but the order IE11 fires the events means it's disabled *after* starting the drag (in that fiddle's use case).

yohanboniface:
>  but the order IE11 fires the events

I'm curious about the details

skalb:
Upon further testing, I'm seeing this issue on both iOS Chrome and Safari as well (using iPhone 6 and simulator).

Same fiddle: https://jsfiddle.net/skalb/otwgyutn/5/embedded/result/

skalb:
This looks like a different issue so I opened #3825 instead.

IvanSanchez:
Apparently this is reproducible in beta2 (as reported in #4157) - anyone up to `git bisect` the code?

amir-infsoft:
I did some further testing and found that is works correctly in one (sub)version of 1.0beta2 and not correct in another.

There are also fiddles where you can check.

In current version from [www.leaflet.js] (http://leafletjs.com/2015/10/14/leaflet-1.0-beta2-released.html) it does not work:
Leaflet: Leaflet 1.0.0-beta.2 (954e83d)
http://cdn.leafletjs.com/leaflet/v1.0.0-beta.2/leaflet.js
http://jsfiddle.net/xu72htcy/7/embedded/result/

In version I found in one example fiddle it works correctly in IE:
Leaflet: Leaflet 1.0.0-beta.2 (08d655f)
https://s3.amazonaws.com/fieldwire-dev/leaflet-master/leaflet-src.js
http://jsfiddle.net/xu72htcy/4/embedded/result/

IvanSanchez:
@amir-infsoft Does it work in the current `master` branch? (See https://playground-leaflet.rhcloud.com → "add library" → "Leaflet development version")

amir-infsoft:
[...]

PR #4479, opened 2016-04-21
Ignore events if Draggable is disabled, fixing disabling drag on click in IE11

perliedman:
This is a fix for #3666, that simply prevents the draggable's event handlers to run if the draggable is disabled.


mourner:
What's the next action here? Should we just merge, or does it need some test updates?

perliedman:
@mourner I talked to @IvanSanchez about adding a unit test, but was halted by some issue with prosthetic hand in IE11. I'll bring it up with Ivan again and try to fix the test.

It's otherwise good to merge.

perliedman:
The added test can only be run in IE11 after https://github.com/Leaflet/prosthetic-hand/pull/11 has been merged, since the current version of prosthetic hand does not work in IE11.

IvanSanchez:
@perliedman That's been merged, and a new version of prosthetic-hand should be available in NPM.

PR #4483, opened 2016-04-21
Add L.Draggable options and fix docstring

nathancahill:
L.Draggable options are ignored right now. Also, a required argument can't follow an optional one.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -39,9 +39,11 @@ L.Draggable = L.Evented.extend({
 		}
 	},
 
-	// @constructor L.Draggable(el: HTMLElement, dragHandle?: HTMLElement, preventOutline: Boolean)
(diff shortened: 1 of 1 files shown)


src/dom/Draggable.js
```diff
@@ -39,9 +39,11 @@ L.Draggable = L.Evented.extend({
 		}
 	},
 
-	// @constructor L.Draggable(el: HTMLElement, dragHandle?: HTMLElement, preventOutline: Boolean)
+	// @constructor L.Draggable(el: HTMLElement, dragHandle?: HTMLElement, preventOutline?: Boolean, options?: Draggable options)
```
IvanSanchez: @nathancahill Please add a `@aka` directive so that `Draggable options` will actually become a link to the class' options section, see https://github.com/Leaflet/Leaflet/blob/master/src/layer/Popup.js#L41

IvanSanchez:
Superseded by #5301

Issue #4486, opened 2016-04-22
L.Control throws error when removing non-added layer


IvanSanchez:
From #4227:

This currently fails:

		it('silently returns when trying to remove a non-existing layer from the control', function () {
			var baseLayer = L.tileLayer(),
			    overlay = L.marker([0, 0]),
			    layers = L.control.layers({'base': baseLayer}, {}).addTo(map);

			layers.removeLayer(overlay);
		});

message:
```
undefined is not an object (evaluating 'this._layers[i].layer')
_getLayer@/home/jieter/workspace/Leaflet/src/control/Control.Layers.js:194:31
removeLayer@/home/jieter/workspace/Leaflet/src/control/Control.Layers.js:112:27
/home/jieter/workspace/Leaflet/spec/suites/control/Control.LayersSpec.js:109:22
```

jieter:
This one can be closed.

yohanboniface:
:)

PR #4487, opened 2016-04-22
Make sure L.Control.removeLayer() doesn't raise an error if a layer n…

jieter:
…ot in the control is removed from it

Fixes issue #4486 reported by @theashyster in #4227


theashyster:
I also tried this locally before posting the issue and this does not throw the error anymore, but has another side effect - it removes other layers from the LayerControl.

jieter:
doh

jieter:
Fixed now. I already added this check before, but since the test also succeeded without it, I removed it again. Time for coffee again.

theashyster:
Thanks @jieter 
Damn tests :smiley:

PR #4493, opened 2016-04-23
Fix canvas not filtering click event after drag

### LINKED ISSUE #4458: Canvas circleMarker fires 'click' after being dragged
https://playground-leaflet.rhcloud.com/yaq/edit

Scenario:
Map initialized with  `preferCanvas: true` .
Added a circleMarker 
```
var markerC = L.circleMarker(myCenter2,{radius:30}).bindPopup("popUp");
    map.addLayer(markerC);
```
Bug: if you start the drag of the map while having the cursor ontop of the circlemarker and you stop the drag with the cursor still on the circleMarker the click event will fire ( i'm just assuming it's a click event) and the popup will open.
This is not the default behaviour, if you set `preferCanvas: false` this will not happen.
I've found an intresting piece of code https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L969 but i don't know how to fix this for canvas markers.

EDit: the bug occurs also if you don't actually start the drag with the cursor ontop of the marker, it has just to be ontop of the marker when you end the drag

yohanboniface:
Fix #4458 

The issue come from the fact that we compute targets in a different way while in canvas or vector DOM, but we were now checking for drag only in the DOM scenario.
I've tried to refactor a bit to get the check only once, but I think this is not doable given that we are computing the targets in two totally distinct ways.
Note to @IvanSanchez I've added a `happen.click` on the test to simulate the `click` event after down/up, I'm open to a better suggestion :)

This PR also fixes `preclick` not being fired by canvas layers. It's in a separate commit and I think it should stay separate and not be rebased.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -253,7 +253,7 @@ L.Canvas = L.Renderer.extend({
 
 		for (var id in this._layers) {
 			layer = this._layers[id];
-			if (layer.options.interactive && layer._containsPoint(point)) {
+			if (layer.options.interactive && layer._containsPoint(point) && !this._map._draggableMoved(layer)) {
 				L.DomEvent._fakeStop(e);
 				layers.push(layer);
 			}
--- a/spec/suites/layer/vector/CanvasSpec.js
+++ b/spec/suites/layer/vector/CanvasSpec.js
@@ -90,6 +90,34 @@ describe('Canvas', function () {
 			layer.off();
 		});
 
+		it("should not fire click when dragging the map on top of it", function (done) {
+			var downSpy = sinon.spy();
+			var clickSpy = sinon.spy();
+			var preclickSpy = sinon.spy();
+			layer.on('click', clickSpy);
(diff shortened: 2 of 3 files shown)


balciseri:
greate work, i tried this fix and it works like a charm

src/layer/vector/Canvas.js
```diff
@@ -253,7 +253,7 @@ L.Canvas = L.Renderer.extend({
 
 		for (var id in this._layers) {
 			layer = this._layers[id];
-			if (layer.options.interactive && layer._containsPoint(point)) {
+			if (layer.options.interactive && layer._containsPoint(point) && !this._map._draggableMoved(layer)) {
```
perliedman: Might it be possible to just remove the click listener while the map is dragged? I guess that would make it a bit clearer what's going on, and make it less coupled with `Map`'s internals.

mourner:
Damn, squashed accidentally — sorry! The commit has a link to PR and a summary though.

@perliedman this may be a better approach — let's follow up if necessary.

PR #4494, opened 2016-04-23
Fix setMaxBounds not resetting maxBounds when passing null argument

yohanboniface:


src/map/Map.js
```diff
@@ -254,7 +254,8 @@ L.Map = L.Evented.extend({
 	setMaxBounds: function (bounds) {
 		bounds = L.latLngBounds(bounds);
 
-		if (!bounds) {
+		if (!bounds.isValid()) {
```
IvanSanchez: Shouldn't this be `if (!bounds || !bounds.isValid())` ??

yohanboniface: L.latLngBounds cannot return a falsy thing anymore https://github.com/Leaflet/Leaflet/blob/master/src/geo/LatLngBounds.js#L236

PR #4498, opened 2016-04-24
Correct TileLayer constructors in debug

nathancahill:
Was `new L.tileLayer` should be `new L.TileLayer`


yohanboniface:
Better to remove `new` and use the factory at this point ;)

nathancahill:
Fixed :)

PR #4499, opened 2016-04-25
Remove spec.hintrc.js (#4439)

### LINKED ISSUE #4439: the extension of spec.hintrc.js should NOT be .js
since spec.hintrc.js is a JSON file, it is not a valid JS file
using .js extension will cause some parsers failure

akkatracker:
Fixes #4439


yohanboniface:
Thanks :)

PR #4504, opened 2016-04-25
Consistent returns for Map.addLayer

nathancahill:
If layer already exists, return `this` instead of `layer`


yohanboniface:
LGTM :)

Issue #941, opened 2012-08-26
Auto-detect changes to map size (and invalidateSize)


broofa:
I have a map that's styled like this:

```
#map {position: absolute; top: 10px; left: 10px; right: 10px; bottom: 90px;}
```

... i.e. it has no explicit height.  So, of course, leaflet isn't rendering all of the tiles.  However I do `setTimeout(map.invalidateSize.bind(map))`, after creating the map it renders properly.

This has me thinking that it's not actually necessary for the map to have an explicit height - it's just a question of insuring Leaflet is smart about detecting the height at the appropriate time.

HarelM:
My two cents:
I think a minor fix would help in order to prevent a bug I had: keep trying to get the size if it is 0,0.
```js
getSize: function () {
		if (!this._size || this._sizeChanged || (this._size && this._size.x == 0 && this._size.y == 0) {
			this._size = new Point(
				this._container.clientWidth || 0,
				this._container.clientHeight || 0);

			this._sizeChanged = false;
		}
		return this._size.clone();
	},
```
The element could be initialized with zero size but later change without calling resize (I think, at least this is what I think happened in the scenario I had to fix).

CourtneyJane2904:
I appreciate that this issue has been closed; I had this issue today and reading this discussion inspired my fix (bare in mind that my case is VERY simple- just a map with a view and a marker)

It might help a fellow noob like myself:

`$('#tab').click(function()
    {setTimeout(function(){setView|tileLayer|add to map|add markers},200);}`

This prevents the default none display of tab content interfering with the tiling process (I'm sure this has probably been mentioned though!) :)

PR #4507, opened 2016-04-26
Added L.Handler.ContainerMutation to call invalidateSize() automatically

IvanSanchez:
Detecting size changes in the map container through `MutationObserver` is doable, even though [@mourner disagreed with the idea a few years back](https://github.com/Leaflet/Leaflet/issues/941#issuecomment-8456510).

I'm not 100% convinced this will cover all use cases, as a mutation observer only checks changes in the CSS attributes, not in the actual rendered values. Relative CSS units like `50%` and a parent size change might break this :-/

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -194,6 +194,11 @@ var deps = {
 		desc: 'Enables keyboard pan/zoom when the map is focused.'
 	},
 
+	ContainerMutation: {
+		src: ['map/handler/Map.ContainerMutation.js'],
+		desc: 'Enables tracking DOM mutations to the map container size.'
+	},
+
 	MarkerDrag: {
 		src: ['layer/marker/Marker.Drag.js'],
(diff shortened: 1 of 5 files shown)


mourner:
How often does the mutation event get triggered? Is it only on the direct change of the container element, or any children too? I wonder if this has any performance implications.

IvanSanchez:
AFAIK, the mutation event is triggered whenever the *attributes of the DOM element* change. This includes the style attributes (`HTMLElement.style.width`, `.height`). That happens consistently in my debug page.

I am not 100% sure the mutation is triggered when the element size changes by some other obscure means, such as:
* Responsive CSS rules
* Browser scaling
* Letting the map container have a CSS rule of `width: 50%`, then changing the size of a parent element
* Flexboxes
* Visibility clip with jquery accordions
* CSS `width: calc( ... )` rules

PR #4509, opened 2016-04-26
Replace for..in with standard for in L.Control.Layers

### LINKED ISSUE #4508: Problem with for...in loop on layers array in L.Control.Layers on IE
So I know that this issue is basically caused by the library I am using - JSTS version 1.1.1, but it also shows a small flaw in the way the layers array is looped through.

In the `L.Control.Layers` function `_update()` the iteration through layers array happens using `for...in` which is okay to use on an associative array as I read here http://stackoverflow.com/a/242888, but in the same thread in another comment http://stackoverflow.com/a/243778 it is suggested to also do this check `if (obj.hasOwnProperty(name))`. The same thing should be checked in `onRemove` method for `L.Control.Layers`.

So my suggestion would be to add this line so the iteration looks like this
```
for (i in this._layers) {
    if (this._layers.hasOwnProperty(i)) {
        obj = this._layers[i];
        this._addItem(obj);
        overlaysPresent = overlaysPresent || obj.overlay;
        baseLayersPresent = baseLayersPresent || !obj.overlay;
        baseLayersCount += !obj.overlay ? 1 : 0;
    }
}
```

Example using Leaflet 1.0.0 RC 1 and JSTS 1.1.1 http://playground-leaflet.rhcloud.com/win/1/edit?html,output
This results in an error in the console `Unable to get property '_leaflet_id' of undefined or null reference` and `L.control.layers` not being added to the map.

jieter:
Fixes #4508 
Leftover from #4227

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -85,8 +85,8 @@ L.Control.Layers = L.Control.extend({
 	onRemove: function () {
 		this._map.off('zoomend', this._checkDisabledLayers, this);
 
-		for (var id in this._layers) {
-			this._layers[id].layer.off('add remove', this._onLayerChange, this);
+		for (var i = 0; i < this._layers.length; i++) {
+			this._layers[i].layer.off('add remove', this._onLayerChange, this);
 		}
 	},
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
LGTM.

theashyster:
Retested with the fix, works fine now :+1:

PR #4512, opened 2016-04-27
Correct docs for GeoJSON.onEachFeature

nathancahill:
onEachFeature is called as `(feature, layer)` instead of `(layer)`


perliedman:
That's an ugly error, good catch!

PR #4515, opened 2016-04-27
Various debug fixes

nathancahill:
Type checked all of the debug/ code and found a couple minor fixes:

- fix missing variables
- fix trailing commas
- remove useless jquery
- use public apis instead of _private apis
- correct constructor options

### DIFF
--- a/debug/map/image-overlay.html
+++ b/debug/map/image-overlay.html
@@ -28,8 +28,8 @@
 		map.addLayer(osm);
 
 		var	bounds = new L.LatLngBounds(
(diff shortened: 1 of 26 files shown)


mourner:
Looks great apart from some whitespace problems (spaces mixed with tabs etc.)

nathancahill:
Good call, fixed.

jieter:
We can setup some eslint checks for the debug/ folder using [eslint-plugin-html](https://www.npmjs.com/package/eslint-plugin-html), to try to keep them cleaner in the future. I'm willing to open a PR for that.

PR #4516, opened 2016-04-28
Typescript definition file and tests

nathancahill:
I believe that having type definitions for Leaflet's public interface would be very helpful. Writing the definition over the past 2 weeks surfaced a number of inconsistencies in interfaces, docs, tests and debug scripts leading to 45 pull requests. So not only is it useful for people using Typescript, but it adds another check for correctness in the project. The definition file is based off of https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/leaflet, but fully re-written for 1.0.

The tests file, besides testing every single interface, also includes all of the code in debug/ which I think is a good sampling of edge cases and code in the wild.

Since most of the interfaces and types are defined in the docstrings, I worked on a tool to auto-generate the definition file from the docs. This worked ok, but not well enough to be automated, so that's not included.

The reason I think this should live in the Leaflet repo (besies the DefinitelyTyped repo), is to keep it in sync with the current master version and each release. DefinitelyTyped would always have the latest stable. Immutable.js takes this approach: https://github.com/facebook/immutable-js/tree/master/type-definitions

Tests are run strictly with --noImplicitAny. Added a command for running the tests in the Jakefile (`jake typescript`), not sure if `typescript` should be added to the devDependencies.

### DIFF
--- a/type-definitions/leaflet-tests.ts
+++ b/type-definitions/leaflet-tests.ts
@@ -1,0 +1,1252 @@
+/// <reference path='leaflet.d.ts' />
+
+let mapOptions: L.MapOptions = {
+    preferCanvas: false,
+
+    attributionControl: true,
+    zoomControl: true,
+
+    closePopupOnClick: true,
+    zoomSnap: 1,
+    zoomDelta: 1,
+    trackResize: true,
+    boxZoom: true,
+    doubleClickZoom: true,
+    dragging: true,
+
+    crs: L.CRS.EPSG3857,
+    center: L.latLng([51.505, -0.09]),
+    zoom: 13,
+    minZoom: 3,
+    maxZoom: 8,
+    layers: [],
+    maxBounds: L.latLngBounds([L.latLng(-60, -60), L.latLng(60, 60)]),
+    renderer: L.canvas(),
+
+    fadeAnimation: true,
+    markerZoomAnimation: true,
+    transform3DLimit: 2^23,
+    zoomAnimation: true,
+    zoomAnimationThreshold: 4,
+
+    inertia: true,
+    inertiaDeceleration: 3000,
+    inertiaMaxSpeed: Infinity,
+
+    easeLinearity: 0.2,
+    worldCopyJump: false,
+    maxBoundsViscosity: 0.0,
+
+    keyboard: true,
+    keyboardPanDelta: 80,
+
+    scrollWheelZoom: true,
+    wheelDebounceTime: 40,
+    wheelPxPerZoomLevel: 50,
+
+    tap: true,
+    tapTolerance: 15,
+    touchZoom: true,
+    bounceAtZoomLimits: true,
+}
+
+let map = L.map(document.getElementById('map'), mapOptions);
+map = L.map('map', mapOptions);
+map = new L.Map('map', mapOptions);
+map = new L.Map('map', {
+    doubleClickZoom: 'center',
+    scrollWheelZoom: 'center',
+    touchZoom: 'center'
+})
+
+map.setView(L.latLng([51.505, -0.09]), 13, { animate: false, duration: 0.25, easeLinearity: 0.25, noMoveStart: false })
+   .setView([51.505, -0.09])
+   .setView({ lat: 51.505, lng: -0.09 })
+   .setZoom(12)
+   .zoomIn(2)
+   .zoomOut(2)
+   .setZoomAround(L.latLng([51.505, -0.09]), 12)
+   .setZoomAround([51.505, -0.09], 12)
+   .setZoomAround({ lat: 51.505, lng: -0.09 }, 12)
+   .fitBounds(L.latLngBounds([L.latLng(-60, -60), L.latLng(60, 60)]))
+   .fitBounds([L.latLng(-60, -60), L.latLng(60, 60)])
+   .fitBounds([[-60, -60], [60, 60]])
+   .fitBounds([{ lat: -60, lng: -60 }, { lat: 60, lng: 60 }])
+   .fitWorld()
+   .panTo(L.latLng([51.505, -0.09]))
+   .panTo([51.505, -0.09])
+   .panTo({ lat: 51.505, lng: -0.09 })
+   .panBy(L.point([3, 5]))
+   .panBy([3, 5])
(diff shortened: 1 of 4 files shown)


danzel:
Who is going to maintain the typescript definition?
Are we expecting everyone to keep it up to date when they make changes?

(I use typescript and use the definition from DT, so want to see this succeed somehow :+1: )

nathancahill:
Valid point. I tried to find a way to generate some part of the definition file automatically with Leafdoc, since most of the type information is there already. However, that's quite a way off from being feasible.

Changes to public interfaces would have to be made in the definition file as well. Implementation details would not. And I imagine there will be less (ideally no) interface changes between major releases.

Beyond that, I'd volunteer to take ownership of it.

jmlopez-rod:
Hi Nathan, thank you for taking the time to create the typings for the new version of leaflet. I'm currently trying out it and it seems very easy to follow.

A few suggestions though, from the leaflet library I have been using some of the most simple objects like `Evented` and `Layer`. I can see from the definitions that I can now use `Layer` but I have no luck with `Evented`.

It would be nice to be able to get all the definitions, including the private members (those with the underscore prefix) to see what we have under the hood. We may want to at least declare them as protected (now that typescript supports it https://github.com/Microsoft/TypeScript/pull/688). In any case,
I exposed `Evented` as follows based on your definition for `Layer`.

```
  export interface EventedStatic extends ClassStatic {
    new(): Evented;
  }
  export var Evented: EventedStatic;
```

Is this the right way to do it? Will you consider adding it?

nathancahill:
I don't think the _underscore members should be defined in Typescript since they are implementation details and could change. Evented should be though, I'll update the PR.

jmlopez-rod:
Another one that should be added is the constructor signature for `L.LatLngBounds`, currently doing something like:

```javascript
const bounds = new L.LatLngBounds();
bounds.extend(latlng1);
bounds.extend(latlng2);
...
```

complains saying that the supplied parameters do not match any signature of call target. The only way to stop this error for now is to do `const bounds = new L.LatLngBounds(null)` but that looks ugly. Looking forward to these type definitions making it to master.

type-definitions/leaflet-tests.ts
```diff
@@ -0,0 +1,1252 @@
...
+control = new L.Control({ position: 'topright' });
+
+control.setPosition('topright')
+       .addTo(map)
+       .remove();
+
+control.getPosition();
+control.getContainer();
+
+let handler = new L.Handler();
```
jmlopez-rod: needs a argument

nathancahill:
Nice catch, I'll update those things too.

As far as splitting up the definition, we could mirror the structure of `src/ `, or have the definitions along side each class in `src/` or have one big long file. Thoughts on that?

jmlopez-rod:
I would prefer to mirror the file structure. It would make it easier to make updates on the typings with every PR since we could later identify the location where we need to update. A similar thing should be done with the testing, I have a feeling your testing file is going to get a bit too long for Github to display.

jmlopez-rod:
Another thought is to split the files to mirror exactly all the `src` directory and then make one big file as part of the build (this means that the big file would then only be commited on a release). Then in `package.json` we would add `typings: "path/to/leaflet-typings.d.ts`" so that projects that import leaflet can pick those definitions up.

nathancahill:
Right, it would still compile to `leaflet.d.ts`

jmlopez-rod:
Hi Nathan, I need an idea of how to use access the `dragging` handler extended functions. As defined in the typings it only has 5 methods: `enable`, `disable`, `enabled`, `addHooks` and `removeHooks`. But `map.dragging` is a derived Handler and adds the methods: `moved` and `moving`: <https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.Drag.js>. Any suggestions on how to modify the typings to have access to them?

type-definitions/leaflet-tests.ts
```diff
@@ -0,0 +1,1252 @@
+/// <reference path='leaflet.d.ts' />
```
darcyparker: @nathancahill - for when you get back to this pull request: 

See `red flags` section of https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

It would be nice to have the updated `d.ts` included in the `package.json`

jmlopez-rod:
Any updates on this PR? I wanted to point out that with the newer versions of typescript (2.1) we could make all the option objects non-optional. For instance:

```
  export interface TileLayerOptions extends GridLayerOptions {
    minZoom?: number;
    maxZoom?: number;
    maxNativeZoom?: number;
    subdomains?: string | string[];
    errorTileUrl?: string;
    zoomOffset?: number;
    tms?: boolean;
    zoomReverse?: boolean;
    detectRetina?: boolean;
    crossOrigin?: boolean;
  }
```

We can remove the `?` so that the type is just the intended type (when we turn on the strictNull option things get pretty annoying). So how can we pass options to the constructor without having to define each and every one of the properties? For that we can use declare the options object as `Partial<TileLayerOptions>`. See: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html

When we turn on the strictNull flag typescript keeps reminding us that a particual option could be undefined. This is not case however we always give it default values. For this reason I think it may be best to use Partials when passing the options in the constructor.

Either way, I'm currently using the types you provided but I'm afraid they may have changed now. Is there a plan to keep the types up to date?

IvanSanchez:
@jmlopez-rod None of the current maintainers use TypeScript, so the priorities are elsewhere. Please read #5003, particularly the bits about generating TS defs from https://github.com/Leaflet/Leafdoc.

jmlopez-rod:
@IvanSanchez thank you for the response and making it very clear several times already that nobody in leaflet use typescript. The question was mainly aimed at @nathancahill since he's the one that kindly provided the type definitions that I currently use.

danzel:
Closing this, see the comments here:
https://github.com/Leaflet/Leaflet/issues/5003#issuecomment-321225045
https://github.com/Leaflet/Leaflet/issues/5003#issuecomment-321369241

Please contribute these changes to the DefinitelyTyped repository if they have not been already.

PR #4531, opened 2016-05-03
Docstrings: non-standard WMS parameters as L.TileLayer.WMS options

IvanSanchez:
In order to decrease user confusion, see https://gis.stackexchange.com/questions/192118/adding-wms-layer-with-cql-filter/192132#192132


perliedman:
:+1:

Issue #4528, opened 2016-05-03
fitBounds with padding zooms the map really far out


theashyster:
I am not sure if this is a regression or not, but I have experienced that since the RC1 release the map `fitBounds` function with `padding` option does not work the same way as it worked before.

Some examples:
Leaflet 0.7.5 http://playground-leaflet.rhcloud.com/niv/edit?html,output
Leaflet 1.0.0-beta.2 http://playground-leaflet.rhcloud.com/jasi/edit?html,output
Leaflet 1.0.0-rc.1 http://playground-leaflet.rhcloud.com/cipe/edit?html,output

The padding is just really big and the map is zoomed far out. We have used this function with the padding on layer containing SVG polygon also and since RC1 that also started to act strange by zooming the map really far out.

I can get the same result on RC1 by setting the `padding` option on a point that is really small `L.point(0.001, 0.001)`.

jieter:
Sounds like a perfect chance for a simple unit test and some git bisecting...

IvanSanchez:
The culprit seems to be [this line](https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L509):

```
boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)).add(padding),
```

That's getting the size (in pixels) of the bounds to fit, **at the current zoom level**, *then* adding the padding.

What we really want is to have the padding at the *destination* zoom level, not at the *original* zoom level.

IvanSanchez:
A possible solution would be to subtract `padding` from `size` - but that needs testing.

I think this is a rather self-contained, explained bug, so I think any newbie can try fixing this. (Hello, @yourfirstpr!).

Things to be done:

* [ ] Let people know you're interested
* [ ] Fork the leaflet repo and clone it locally
* [ ] Modify `src/map/Map.js` around line 509 so that `padding` is subtracted from `size` (instead of being added to `boundsSize`)
* [ ] Add some unit tests to `spec/map/MapSpec.js`, next to the tests for `getBounds()` - ensure that calls to `getBoundsZoom()` with small areas and big padding returns high levels of zoom (currently it returns low levels like 0 and 1)
* [ ] Run `npm install`, `npm test`
* [ ] `git commit`, `git push`
* [ ] Create merge request

Charlotteis:
:wave: Thanks so much for another wonderful guide, @IvanSanchez! Will post this out to @yourfirstpr shortly :)

theashyster:
[...]

PR #4532, opened 2016-05-03
getBoundsZoom with small size and padding

dianjin:
https://github.com/Leaflet/Leaflet/issues/4528

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -162,6 +162,31 @@ describe("Map", function () {
 		});
 	});
 
+	describe("#getBoundsZoom", function () {
+		var halfLength = 0.0001;
+		var bounds = [[-halfLength, -halfLength], [halfLength, halfLength]];
+		var padding = [100, 100];
+		var height = '400px';
+
+		it("returns high levels of zoom with small areas and big padding", function () {
+			var container = map.getContainer();
+			container.style.height = height;
+			document.body.appendChild(container);
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.equal(21);
+		});
+
+		it("returns multiples of zoomSnap when zoomSnap > 0 on any3d browsers", function () {
+			var container = map.getContainer();
+			container.style.height = height;
+			document.body.appendChild(container);
+			L.Browser.any3d = true;
+			map.options.zoomSnap = 5;
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.equal(20);
+			map.options.zoomSnap = 0.25;
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.equal(21);
+		});
+	});
+
 	describe('#setMaxBounds', function () {
 		it("aligns pixel-wise map view center with maxBounds center if it cannot move view bounds inside maxBounds (#1908)", function () {
 			var container = map.getContainer();
@@ -198,27 +223,6 @@ describe("Map", function () {
 			// the view is inside the bounds
 			expect(bounds.contains(map.getBounds())).to.be(true);
 			document.body.removeChild(container);
-		});
-	});
-
-	describe("#getMinZoom and #getMaxZoom", function () {
-		describe('#getMinZoom', function () {
-			it('returns 0 if not set by Map options or TileLayer options', function () {
-				var map = L.map(document.createElement('div'));
-				expect(map.getMinZoom()).to.be(0);
-			});
-		});
-
-		it("minZoom and maxZoom options overrides any minZoom and maxZoom set on layers", function () {
-
-			var map = L.map(document.createElement('div'), {minZoom: 2, maxZoom: 20});
-
-			L.tileLayer("{z}{x}{y}", {minZoom: 4, maxZoom: 10}).addTo(map);
-			L.tileLayer("{z}{x}{y}", {minZoom: 6, maxZoom: 17}).addTo(map);
-			L.tileLayer("{z}{x}{y}", {minZoom: 0, maxZoom: 22}).addTo(map);
-
-			expect(map.getMinZoom()).to.be(2);
-			expect(map.getMaxZoom()).to.be(20);
 		});
 	});
 
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -505,8 +505,8 @@ L.Map = L.Evented.extend({
 		    max = this.getMaxZoom(),
 		    nw = bounds.getNorthWest(),
 		    se = bounds.getSouthEast(),
-		    size = this.getSize(),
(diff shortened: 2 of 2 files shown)


dianjin:
@hyperknot: I tested with different zoomSnaps (with L.browser.any3d = true) and the returned value is always a multiple of the zoomSnap. Setting zoomSnap to zero looks like it's returning the most precise zoom level, in decimal. Do we want tests for this?

spec/suites/map/MapSpec.js
```diff
@@ -222,27 +236,6 @@ describe("Map", function () {
...
-
-			L.tileLayer("{z}{x}{y}", {minZoom: 4, maxZoom: 10}).addTo(map);
-			L.tileLayer("{z}{x}{y}", {minZoom: 6, maxZoom: 17}).addTo(map);
-			L.tileLayer("{z}{x}{y}", {minZoom: 0, maxZoom: 22}).addTo(map);
-
-			expect(map.getMinZoom()).to.be(2);
-			expect(map.getMaxZoom()).to.be(20);
-		});
-	});
-
```
jieter: Seems you accidentally removed this test?

dianjin: I removed it on purpose, this test is an exact copy of the one right above it. (Is there a reason to include the same test twice?)

jieter: Ah, I see 👍

hyperknot:
@dianjin nice work! I only wrote testing different zoomSnap values as it is important to understand the logic and to visually make sure that it works correctly. I believe there is no need to write a new test for it, zoomSnap has it's own tests.

spec/suites/map/MapSpec.js
```diff
@@ -162,6 +162,20 @@ describe("Map", function () {
...
+		it("returns high levels of zoom with small areas and big padding", function () {
+			var container = map.getContainer();
+			container.style.height = '400px';
+			document.body.appendChild(container);
+
+			var halfLength = 0.0001;
+			var bounds = [[-halfLength, -halfLength], [halfLength, halfLength]];
+			var padding = [100, 100];
+
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.greaterThan(10);
```
hyperknot: Maybe it's better to go with a specific value, not just `greaterThan(10)`?

IvanSanchez: I agree - this should be deterministic.

IvanSanchez:
> I believe there is no need to write a new test for it, `zoomSnap` has it's own tests.

Personally I'd feel a bit safer if we had a test for just `getBoundZoom` with the defaults, and one for `getBoundsZoom`+`zoomSnap`.

dianjin:
Changed the tests in response to your feedback.

spec/suites/map/MapSpec.js
```diff
@@ -162,6 +162,31 @@ describe("Map", function () {
...
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.equal(21);
+		});
+
+		it("returns multiples of zoomSnap when zoomSnap > 0 on any3d browsers", function () {
+			var container = map.getContainer();
+			container.style.height = height;
+			document.body.appendChild(container);
+			L.Browser.any3d = true;
+			map.options.zoomSnap = 5;
+			expect(map.getBoundsZoom(bounds, false, padding) % map.options.zoomSnap).to.be.equal(0);
```
IvanSanchez: Why not `to.be.equal(20.5)`?

spec/suites/map/MapSpec.js
```diff
@@ -162,6 +162,31 @@ describe("Map", function () {
...
+		var height = '400px';
+
+		it("returns high levels of zoom with small areas and big padding", function () {
+			var container = map.getContainer();
+			container.style.height = height;
+			document.body.appendChild(container);
+			expect(map.getBoundsZoom(bounds, false, padding)).to.be.equal(21);
+		});
+
+		it("returns multiples of zoomSnap when zoomSnap > 0 on any3d browsers", function () {
```
IvanSanchez: Should be `it.skipInPhantom("returns ...")`. See `spec/suites/map/anim/Map.ZoomAnimationSpec.js` and the end of `spec/SpecHelper.js`

jieter: Hmm, I don't see why that should be, this test manually sets `any3d` to `true` and doesn't rely on css animations, or does it?

IvanSanchez: @jieter Ha, you're right, didn't see that at first glance.

But won't this change the value of `L.Browser.any3d` for future tests? We might just be lucky about the order of the tests.

jieter: Yeah, might want to set it to the original value after the test is done.

dianjin: Some of the other tests are also setting/resetting L.Browser.any3d (looks like this is the only property in L.Browser being manipulated though). Is there a way to reset L.Browser to default values in beforeEach?

spec/suites/map/MapSpec.js
dianjin: I compared the modulus to zero so the test would match the test description more, but I can see that this is basically accepting an infinite number of results so I've changed it to just compare the result to a number.

IvanSanchez:
@dianjin Everything looks good!

I'll tweak the values for the unit tests afterwards, and we'll worry about resetting `L.Browser.any3d` afterwards if needed (after all, the `it.skipInPhantom()` trick is working perfectly so far).

And sorry for making so many comments on your code - I know this makes us seem unfriendly from the eyes of a newcomer, but that's how we usually work :-)

Again, thanks for jumping in and fixing this bug! I hope it was an instructive experience, and hope you will keep an eye on other leaflet bugs :-)

dianjin:
Thanks @IvanSanchez for the constructive feedback! I'm glad you guys pointed out where I could improve, thank you Leaflet and @Charlotteis for encouraging beginners to contribute! 😄

Moriphete:
> @dianjin Everything looks good!
> 
> I'll tweak the values for the unit tests afterwards, and we'll worry about resetting `L.Browser.any3d` afterwards if needed (after all, the `it.skipInPhantom()` trick is working perfectly so far).
> 
> And sorry for making so many comments on your code - I know this makes us seem unfriendly from the eyes of a newcomer, but that's how we usually work :-)
> 
> Again, thanks for jumping in and fixing this bug! I hope it was an instructive experience, and hope you will keep an eye on other leaflet bugs :-)

Bugs Bats and unanswered queries. Now jow much time did it take to calculate the total depth of pie?

Moriphete:
The total depth of pi

edwaka (commented):
#4532

PR #4535, opened 2016-05-04
Added Leaflet.FeatureGroup.LoadEvents (for v0.7.*)

glathoud:
`FeatureGroup` that supports the `"loading"` and `"load"` events by aggregating the events fired by its sublayers. This can be useful to show a loading indicator for groups of tilelayers.

See also:
 * a [detailed discussion](https://github.com/Leaflet/Leaflet/pull/4530) that led to the present pull request.
 * a corresponding [issue](https://github.com/ebrelsford/Leaflet.loading/issues/33) opened in [`Leaflet.loading`](https://github.com/ebrelsford/Leaflet.loading/).

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -499,6 +499,15 @@ The following plugins change the way that tile layers are loaded into the map.
 			<a href="https://github.com/ghybs">ghybs</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/Outdooractive/Leaflet.FeatureGroup.LoadEvents">Leaflet.FeatureGroup.LoadEvents</a>
+		</td><td>
+			`FeatureGroup` that supports the `"loading"` and `"load"` events (for v0.7.*).
+		</td><td>
+			<a href="http://glat.info">G. Lathoud</a>, <a href="http://www.outdooractive.com">Outdooractive</a>.
+		</td>
+	</tr>
 </table>
 
 


IvanSanchez:
Fair enough. Please remember to send another PR to remove this if/when the issue in Leaflet.loading is resolved.

glathoud:
Thanks @IvanSanchez 

> Please remember to send another PR to remove this if/when the issue in Leaflet.loading is resolved.

`Control.Loading` was [fixed](https://github.com/ebrelsford/Leaflet.loading/commit/21a3e85c708b92569b0bd694d2c63c50ed38629d#diff-d6944cbccec35befbc93665d1cf75f87).

Now, as API provider we still need the plugin to prevent the issue
from happening *ever* again, by making sure all async layers fire
`"loading"` and `"load"`, regardless of their nature (single, group, ...).

If we don't, that is an API *design* decision that can "break", by triggering
 another issue as soon as a client uses another loading
indicator. In my experience, if an API design decision can break, it
*will*, sooner or later.

Since (1) the plugin is helping us, and (2) it costed your time and
mine, and (3) someone else may encounter a similar issue, I am not
going to submit a "delete" PR. I hope you unterstand my position.

Many thanks for the detailed feedback in issue #4530. It led to a
better plugin and a good fix.

IvanSanchez:
> Since (1) the plugin is helping us, and (2) it costed your time and mine, and (3) someone else may encounter a similar issue, I am not going to submit a "delete" PR. I hope you understand my position.

Sure. I hope you also understand my position: with over 200 plugins, I want to not have duplicated features all over the place :-)

Issue #4051, opened 2015-11-25
Regression: embedded maps on Android devices will prevent drag-scroll in v0.7.7 (and in 1.0.0-beta)


louh:
The trick specified in #2031 to add `tap: false` to the map options in order to prevent Leaflet from swallowing `touchmove` events is broken in Leaflet 0.7.7, as well as in 1.0.0-beta.1 and 1.0.0-beta.2. So, for maps that are embedded in part of a page, Leaflet 0.7.5 is the last version I could recommend using if you want to prevent that map from not hijacking touch-scroll (edit: on Android devices).

I have a very simple page you can load in a mobile browser or mobile simulator to test this effect on each of the four versions I've specified: http://louh.github.io/nyc-community-boards

The code for this lives here: https://github.com/louh/leaflet-embedded

louh:
Addendum: I just tried this on iOS Simulator (iOS 9.1, iPhone 6) and it does not appear to affect iOS. My normal devices are on Android Chrome and also the mobile simulator of Chrome on OSX.

littleiffel:
in 0.7.7 in the CSS file .leaflet-container received the property

```
touch-action:none
```

Removing this line made the page scroll again even while touching/dragging on the map container.

Currently this line is still present in the CSS.

perliedman:
@louh can you clarify exactly what the problem is with this example on mobile: http://louhuang.com/nyc-community-boards/#11/40.6051/-74.1055

When loading it in mobile mode, it shows a header and search field on top, the map in the middle, and an information text below. I can pan the map with touch, but there's nothing I can scroll on this page. Or is that example running a working version?

Also, I would suggest you put the examples, with different versions of Leaflet, on http://playground-leaflet.rhcloud.com/ , where it is easy to switch library version, and also easier for Leaflet devs to work with.

louh:
Hi @perliedman, thanks for responding! The http://louhuang.com/nyc-community-boards/#11/40.6051/-74.1055 link is probably not the best to look at since I did a bunch of work on it since then. 

I've done the leaflet playground as you suggest:

with Leaflet v1, drag-scrolling the page fails on Android or Chrome mobile simulator when the user touches the map:
http://playground-leaflet.rhcloud.com/fuzo/edit?html,output

with Leaflet v0.7.5, drag-scrolling the page using the map works:
http://playground-leaflet.rhcloud.com/los/1/edit?html,output

IvanSanchez:
[...]

PR #4552, opened 2016-05-10
Don't prevent browser's touch scroll and/or zoom unless handlers are enabled

perliedman:
Currently, we set `touch-action: none` for the whole map container, preventing the browsers native scroll and zoom functionality, even when the touch zoom and map drag handlers are disabled. This is a problem, see #4051.

This adds to new CSS classes that are only added to the container when the drag and touch zoom handlers are enabled.

Currently, only IE supports enabling _just_ pinch zoom (touch zoom). This means that for Chrome/FF, etc. you still can't do browser zoom, even if the touch zoom handler is disabled.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -16,8 +16,6 @@
 	}
 .leaflet-container {
 	overflow: hidden;
-	-ms-touch-action: none;
-	touch-action: none;
 	}
 .leaflet-tile,
 .leaflet-marker-icon,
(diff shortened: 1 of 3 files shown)


IvanSanchez:
@perliedman Do you think it would be possible to have prosthetic-hand tests for this?

perliedman:
@IvanSanchez I was thinking about it. I'm not sure if the events fired by prosthetic hand actually trigger things like browser scrolling and zooming, or if those things happen on another level in the browser, outside JavaScript events. Do you know?

PR #4555, opened 2016-05-10
Prevent listeners from being called when all listeners are removed

perliedman:
In the special case where all listeners for an event are removed by a call to `off`, we didn't clear the existing listeners in the same way we do when removing a single listener, which made it possible for listeners that had already been removed to be called.

This PR adds a fix and a test to verify.


IvanSanchez:
Looks good. I guess there is a small chance we might hit some confused developers on some edge case wondering why their event handlers disappear.

PR #4557, opened 2016-05-11
Scale down getWheelDelta() in win+chrome and ffx+retina, should fix #4538

### LINKED ISSUE #4538: Scrollwheel Zoom issue in Chrome
I have noticed since updating to the newest leaflet version that all of my scroll wheel zooms are now jumping 2 zoom levels in chrome instead of one every time I use my scroll wheel.  

The issue appears to be the result of using a default value of '50' for wheelPxPerZoomLevel to control zoom speed, while using a normalized getWheelDelta function which converts your scroll wheel data to a standardized value. In Chrome my `getWheelDelta` function is returning `100` for a single scroll on the scroll wheel while firefox only returns `50`

IvanSanchez:
I feel dirty by making these kind of browser hacks.

Should fix #4538.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -26,6 +26,8 @@
 	    chrome    = ua.indexOf('chrome') !== -1,
 	    gecko     = ua.indexOf('gecko') !== -1  && !webkit && !window.opera && !ie,
 
+	    win = navigator.platform.indexOf('Win') === 0,
+
 	    mobile = typeof orientation !== 'undefined' || ua.indexOf('mobile') !== -1,
 	    msPointer = !window.PointerEvent && window.MSPointerEvent,
 	    pointer = window.PointerEvent || msPointer,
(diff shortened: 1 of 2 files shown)


hyperknot:
Dirty but good :-)

src/core/Browser.js
```diff
@@ -26,6 +26,8 @@
 	    chrome    = ua.indexOf('chrome') !== -1,
 	    gecko     = ua.indexOf('gecko') !== -1  && !webkit && !window.opera && !ie,
 
+	    win = navigator.platform.indexOf('Win') === 0,
```
hyperknot: why not "Windows NT"?

IvanSanchez: Because https://stackoverflow.com/questions/19877924/what-is-the-list-of-possible-values-for-navigator-platform-as-of-today . My code just checks the three first characters.

hyperknot: Ahh, I get it, it's navigator.platform not ua, right!

IvanSanchez: Yeah, detecting the OS is much more reliable if `navigator.platform` is used. I just really really hope there will be no weird combinations of Chrome in random WinPhones or WinTablets plus USB mice.

hyperknot:
@IvanSanchez can we merge this? (I think so, anyone else on other opinion?)

Issue #4556, opened 2016-05-11
Missing Leafdoc template files


bbecquet:
When trying to build the 1.0.0 API doc according to https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#api-documentation, the build fails with a missing `constructor.hbs` message.

Comparing with https://github.com/Leaflet/Leafdoc/tree/gh-pages/templates/leaflet, `constructor.hbs` & `destructor.hbs` are actually missing from the list of files.

IvanSanchez:
Yeah, I forgot to provide templates for Leaflet when I added the constructor and destructor directives to Leafdoc. It should be simple enough for anybody to copy the `factory.hbs` template into `constructor.hbs` and `destructor.hbs`, modify the text slightly and send a PR. Any takers?

bbecquet:
Ok, thanks for the quick reply. I'll propose the fix tonight if nobody took it in the meantime.

PR #4558, opened 2016-05-11
Add missing Leafdoc template files

bbecquet:
Fix https://github.com/Leaflet/Leaflet/issues/4556

The `destructor.hbs` file was listed in Leafdoc so I've also added it, but it's actually not used by any Leaflet comment yet. I can remove it if we decide it's really useless.


IvanSanchez:
Let's leave it for now. The real fix is to make Leafdoc not load templates if they are not needed.

PR #4559, opened 2016-05-11
docstrings: fix typos and mistakes in L.GridLayer

bbecquet:
Various typos, duplicated word, wrong variable in code example, and added an anchor link to options section like other classes.


IvanSanchez:
Thanks! We keep catching typos in the documentation, but there seems there are still some!

Issue #4519, opened 2016-04-28
geolocalization stop working on cellphones


rossanag:
Hi, locate() isn't working on smart devices, it works fine on desktop. The message error is: _"Geolocation error: Only secur origins are allowed"_Any ideas?

hyperknot:
Geolocation is disabled in most modern browsers if you are not on a https:// website. It is not related to smart devices or desktop, it's only the question of browsers.

If you are looking for a free https solution, I recommend you to simply use CloudFlare (easy) or set up Let's Encrypt (intermediate).

rossanag:
Well, the weird thing is that if I use the google API it works fine, and that it was working before using locate as well.
I don't have a clue how worked under those constraints in cellphones then.
Best

hyperknot:
It's just been rolled out in Chrome 50, I believe we'll be seeing a lot more of these tickets in the next weeks.
https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only

PR #4565, opened 2016-05-12
Some browsers require HTTPS for geolocation

simon04:
See #4519 and also https://goo.gl/rStTGz

The linked example http://leafletjs.com/examples/mobile-example.html does not work since it (is not|cannot be) served via HTTPS.

### DIFF
--- a/examples/mobile.md
+++ b/examples/mobile.md
@@ -39,6 +39,8 @@ L.tileLayer('https://api.tiles.mapbox.com/v4/<a href="https://mapbox.com">MapID<
 
 ### Geolocation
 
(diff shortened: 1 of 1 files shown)


myshov:
I believe that in the future there will be not access to geolocation API for non-secure connection at all. Chrome have already done some steps in that direction https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only?hl=en

therefromhere:
Has there been any movement in getting SSL support for http://leafletjs.com/ ?

theel0ja:
https://blog.github.com/2018-05-01-github-pages-custom-domains-https/

HTTPS has been released for Github Pages custom domains

PR #4574, opened 2016-05-16
Added plugin entry: L.Control.AutoLayers

aebadirad:
A new plugin to dynamically pull from multiple mapservers who publish a
service dictionary and automagically add/organize/search and z-index
order them.

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -437,6 +437,15 @@ The following plugins change the way that tile layers are loaded into the map.
 <table class="plugins"><tr><th>Plugin</th><th>Description</th><th>Maintainer</th></tr>
 	<tr>
 		<td>
+			<a href="https://github.com/aebadirad/Leaflet.AutoLayers">Leaflet.AutoLayers</a>
+		</td><td>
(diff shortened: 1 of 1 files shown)


perliedman:
It's not clear cut, but it might be that this plugin fits better under the section [Layer Switching Controls](http://leafletjs.com/plugins.html#layer-switching-controls]. What do you think?

aebadirad:
I was mixed on it as well. On one hand, it does automatically load up layers for you, but on the other, it has built in layer management. I ultimately went with [Tile Load](http://leafletjs.com/plugins.html#tile-load) because it involves loading layers. Though the plugin can be used for layer management without the automatic part. So perhaps you are correct that this would be better suited for the Layer Switching Controls (and it is basically a L.Control.Layer substitute in spirit).

Should I close this pull request and issue a new one under the Layer Switching Controls category then?

perliedman:
No, you can just add another commit to your branch, and it will be part of this pull request automatically.

Just ping me when you're done and I'll merge!

aebadirad:
Moved to the Layer Switching Control section and placed alphabetically. Thank you @perliedman

perliedman:
:+1:

Issue #4583, opened 2016-05-20
doc: split off marker's mouse events


IvanSanchez:
The mouse events as described by the `L.Marker` documentation are also received by `L.Path`s and `L.GridLayer`s (and in the future, `L.Label`s), but they are not documented.

These should show in a collapsed sub-section, something like "Mouse events inherited from *interactive layer*" or something.

Maybe create a new (abstract) class, and explicitly inherit it where appropiate.

PR #4585, opened 2016-05-20
docstrings: Refactor interactive layers, implements #4583

IvanSanchez:


build/leafdoc-templates/html.hbs
```diff
@@ -116,6 +116,7 @@
 				<li><a href="#class">Class</a></li>
 				<li><a href="#evented">Evented</a></li>
 				<li><a href="#layer">Layer</a></li>
+				<li><a href="#interactive-later">Interactive layer</a></li>
```
yohanboniface: Is `interactive-later` a sort of promise? :)

IvanSanchez: It's just the least ugly artifact I could think of - this way, classes show a section called "events inherited from interactive layer" and looks nice :-)

yohanboniface: @IvanSanchez Take a cup of coffee and read *that* line again ;)

PR #4590, opened 2016-05-24
Add Leaflet.smoothzoom to Plugins

zzarcon:
Just a tiny plugin to fix the default zooming on OSX devices - https://github.com/zzarcon/Leaflet.smoothzoom 

Hope you like it, thanks!

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -1993,6 +1993,17 @@ Change the way the user can interactively move around the map.
 			<a href="https://github.com/unbam">Masashi Takeshita</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/zzarcon/Leaflet.smoothzoom">Leaflet.smoothzoom</a>
+		</td>
+		<td>
+			Fixes the default leaflet zoom on OSX.
+		</td>
+		<td>
+			<a href="https://github.com/zzarcon">Hector Zarco</a>
+		</td>
+	</tr>
 </table>
 
 


yohanboniface:
Hi @zzarcon thanks for submitting your plugin!
I feel like the description is not really descriptive: it's not immediately clear what `Fixes` is referring to. Maybe something like "Make zooming smoother on OSX" would be a bit more descriptive?
BTW, I wonder why would that only target OSX?
Also, ote that the demo is broken for me, because http://zzarcon.github.io/Leaflet.smoothzoom/bower_components/Leaflet.smoothzoom/L.Map.SmoothZoom.js is 404.

zzarcon:
Hi @yohanboniface thanks for feedback! I already fixed the demo, sorry for it :)

Regarding the description, I think you are right, the current one is not very descriptive... I like your "Make zooming smoother on OSX" suggestion. I put OSX just because I thought is what the people might search for? But is actually not just OSX as you said, is in any device which uses trackpads and things like that (because internally the **mousewheel** gets fired more than one) ^^.

So, personally I like the "Make zooming smoother on OSX" one but I'm really up to you if you think we should change the name for a better one, any ideas? 

Thanks!! 💪

IvanSanchez:
In reality, scroll-zooming is quite a bit more difficult than that. The amount of pixels scrolled per "wheel click" (or per trackpad frame) varies wildly between browsers - see #4538.

The ""problem"" with OSX is that scrolling is quite slow - a "wheel click" will scroll down by about 4 pixels, and touchpads have some magic time-dependent smoothing curve going on.

The core of the issue is that Leaflet (right now) can only know about how many pixels have been scrolled, and change zoom levels every `wheelPxPerZoomLevel`.

If there was a **very** consistent way of knowing which browsers/input device combinations produce such low values, and adjust `wheelPxPerZoomLevel` by a factor, I'd be in for that. But by "fixing" things for OSX you might break them for some other browser combo.

I'm not against plugins for changing this kind of behaviour - just be very sure of describing exactly what it is supposed to do, and test it in as many browsers as possible.

zzarcon:
@IvanSanchez I had no idea at all about you just wrote, but I found it very interesting and difficult as you said, so thanks for the info. 

I totally see your point of being careful with the description, actually what the `smoothzoom plugin` does is just throttle the zoom levels you make based on a delay interval.

So, when I said "fix" I didn't meant that the actual behaviour is wrong :) but I think i should get rid of the current description and go with the "Make zooming smoother on OSX" maybe? I'm totally open on that and don't want to create false expectations for the users.

PR #4591, opened 2016-05-24
Remove extra init of drag start

jwoyame:
In Draggable.js, the `_startPos` for the drag operation gets initialized twice.

Once in  `_onDown`:

```javascript
this._startPos = this._newPos = L.DomUtil.getPosition(this._element);
```

And again when the movement first starts in `_onMove`:

```javascript
this._startPos = L.DomUtil.getPosition(this._element).subtract(offset);
```

This small edit removes the extra, unused initialization in `_onDown`. This also removes the unused initialization of `_newPos`, which is set on every call to `_onMove`.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -97,7 +97,6 @@ L.Draggable = L.Evented.extend({
 		var first = e.touches ? e.touches[0] : e;
 
 		this._startPoint = new L.Point(first.clientX, first.clientY);
-		this._startPos = this._newPos = L.DomUtil.getPosition(this._element);
 
 		L.DomEvent
 			.on(document, L.Draggable.MOVE[e.type], this._onMove, this)


yohanboniface:
Thanks! This commit looks good, but I've a small doubt on why those two lines survived until now, and it may need a bit more investigation to understand if there isn't an edge case, and thus adapt the fix:

- the line has already been removed in https://github.com/Leaflet/Leaflet/commit/2d2fc74110f2fe7fbc06d7f6e0a2e90528ce46c6
- but then been readded in a "fix various master regressions" commit in https://github.com/Leaflet/Leaflet/commit/4ed3c97bcdb3f6745edf705efb2d18e7b7812939 (@mourner, guess who I'm looking at? :) )

jwoyame:
It makes a lot of sense why @mourner originally moved the initialization from `_onDown` to `_onMove`. It leaves me to wonder if the later reappearance of that line in `_onDown` could have been a mistake. Any code that would be using the `_startPos` of a drag before `_onMove` happens, which is where an actual drag operation is recognized, would be incorrect anyway!

There are only two other places where `_startPos` is used:

1. In `_onUp` when a movement completes, to measure the total distance of the drag. This code only calculates the distance if `_onMove` was called, so the extra init in `_onDown` won't affect it.

2. In the `_onPreDragLimit` handler in `Map.Drag.js`, which is only called whenever an actual reposition occurs in response to `_onMove`.

mourner:
That commit was in 2012 and I didn't have a good commit message discipline back then. :) I think this is good to go since I can't think of any reason why it should be there.

Issue #4592, opened 2016-05-25
Scrollwheel zoom too fast in MS Edge


IvanSanchez:
As reported in https://github.com/Leaflet/Leaflet/issues/4538#issuecomment-221545760 , the `wheel` event in MS Edge returns a normalized `delta` of *around* 120 pixels per wheel click. The exact number seems to depend on the window geometry, or its aspect ratio, or something like that.

![edge-wheel-event](https://cloud.githubusercontent.com/assets/1125786/15538939/a6d4c10c-227f-11e6-884f-e64b64b5e495.png)

Most of the discussion in #4538 applies here. This calls for *another* browser hack for Edge (setting `_wheelPxFactor` to about 2.5 in Edge), and once again considering using the legacy `wheelDelta` properties of the event, as they seem to always be multiples of 120px across browsers.

PR #4594, opened 2016-05-25
TouchZoom and MouseWheelScroll

SveinLoken:


src/dom/DomEvent.js
```diff
@@ -227,9 +227,10 @@ L.DomEvent = {
 
 	// Chrome on Win scrolls double the pixels as in other platforms (see #4538),
 	// and Firefox scrolls device pixels, not CSS pixels
-	_wheelPxFactor: (L.Browser.win && L.Browser.chrome) ? 2 :
-	                (L.Browser.gecko) ? window.devicePixelRatio :
-	                1,
+	_wheelPxFactor: ((L.Browser.win && L.Browser.chrome) || L.Browser.safari) ? 2 :
+					L.Browser.edge ? 3 :
```
IvanSanchez: OK, so I think this is trying to address #4592

src/map/handler/Map.TouchZoom.js
```diff
@@ -109,6 +109,15 @@ L.Map.TouchZoom = L.Handler.extend({
 		    .off(document, 'touchmove', this._onTouchMove)
 		    .off(document, 'touchend', this._onTouchEnd);
 
+		var snap = L.Browser.any3d ? this._map.options.zoomSnap : 1;
+		if (snap) {
```
IvanSanchez: Why are you changing the touch zoom behaviour in the same pull request with a scrollwheel change? If there is an issue about touch-zoom snapping, do provide a bug number for it, and split the change into another commit & pull req.

src/dom/DomEvent.js
SveinLoken: Yes:
Chrome: 50
Firefox: 54
Safari: 120
Edge: 122.7
IE11: 60

src/map/handler/Map.TouchZoom.js
SveinLoken: Sorry! I am a newbie! Should I remove everything and create new pull requests?

IvanSanchez: Yes. One pull request per bug or feature. This pull request has **3** different things:
* A change in `L.Browser.chrome` that is not really needed (Edge having 'chrome' in the UA string doesn't break anything - if it breaks, send a bug report explaining what is broken) - and BTW this happens only in Edge 14, not in Edge 12
* A possible fix for #4592
* A change of behaviour in touch-zoom snapping (which has had no discussion at all, open up a bug about why it's needed)

As a general rule, don't make several changes in several pieces of the software in one commit, unless the changes are very tightly coupled (this is not such a case).

SveinLoken: In my test above edge will return a value if 2 because it match (L.Browser.win && L.Browser.chrome). So I think they are tightly coupled (agree?).
What to do now? Remove my pull request completely before I follow your general rule description?

IvanSanchez: There's still no need to change `L.Browser.chrome`. Just check for edge first, then chrome, then ffx, then everything else.

Either remove this PR and create a new one, or amend it to remove the unneeded changes. That's up to you.

SveinLoken: @IvanSanchez I think #4592 will solve the problem with edge. Sorry to disturbed you. Next time I know how to create a correct pull request.

Issue #4575, opened 2016-05-17
svg and canvas styles leak through to markers


janpaul123:
[These lines](https://github.com/Leaflet/Leaflet/blob/fbc6265fb7a99b442c6997643e4b12d8c43ba346/dist/leaflet.css#L8-L9) make it harder to use SVG and Canvas elements in markers, as you have to explicitly override these styles. Would it be possible to use something more specific?

perliedman:
Closed by #4597.

janpaul123:
Thank you!

PR #4597, opened 2016-05-25
Limit CSS selectors for SVG/Canvas

jwoyame:
Addresses #4575.

The CSS uses 3 different selectors to reference the SVG/Canvas elements for vectors:

`.leaflet-map-pane svg`
`.leaflet-overlay-pane svg`
`.leaflet-pane > svg`

This commit removes the first form in favor of the second, making the selectors a little more standardized and more specific. The `.leaflet-pane > svg` form was left as-is.

### DIFF
--- a/debug/tests/custom-panes.html
+++ b/debug/tests/custom-panes.html
@@ -1,0 +1,70 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+	
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+	
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+	
+	<link rel="stylesheet" href="../css/screen.css" />
+	
+	<script type="text/javascript" src="../../build/deps.js"></script>
+	<script src="../leaflet-include.js"></script>
+	
+	<style>
+		body { max-width: 1200px }
+		#mapSvg { width: 500px; height: 500px; margin: 20px }
+		#mapCanvas { width: 500px; height: 500px; margin: 20px }
+		.left { float: left; text-align: center }
+		.right { float: right; text-align: center }
+	</style>
+</head>
+<body>
+
+	<div class="right">
+		<h1>Canvas</h1>
+		<div id="mapCanvas"></div>
+	</div>
+	<div class="left">
+		<h1>SVG</h1>
+		<div id="mapSvg"></div>
+	</div>
+
+	<script type="text/javascript">
+
+		function makeMap(container, options) {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Won't this break vector layers in custom panes?

jwoyame:
So maybe all of the references should be the third form, `.leaflet-pane > svg`, instead?

IvanSanchez:
I guess `.leaflet-pane > svg` would be the best. If you could make a debug webpage using panes, and check that nothing breaks, that'd be great.

jwoyame:
Yeah I can do that. It makes me wonder if there would also be the same issue with the two other instances of `.leaflet-overlay-pane svg`:

```css
/* .leaflet-container svg: reset svg max-width decleration shipped in Joomla! (joomla.org) 3.x */
/* .leaflet-container img: map is broken in FF if you have max-width: 100% on tiles */
.leaflet-container .leaflet-overlay-pane svg,
.leaflet-container .leaflet-marker-pane img,
.leaflet-container .leaflet-tile-pane img,
.leaflet-container img.leaflet-image-layer {
	max-width: none !important;
	}
```

```css
/* workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=888319 */
.leaflet-overlay-pane svg {
	-moz-user-select: none;
	}
```

jwoyame:
I changed to `.leaflet-pane > svg` and confirmed that it worked for custom panes with the attached debug page, which has both an SVG and Canvas map that use custom panes.

mourner:
Does IE8 support `>` selectors? I don't remember.

IvanSanchez:
@mourner According to https://developer.mozilla.org/en-US/docs/Web/CSS/Child_selectors , it's supported since IE7.

jwoyame:
@mourner A fix that could avoid `>`, if that is a concern (I see it is used very rarely in Leaflet), would be to put a special class on SVG and Canvas renderer targets.

nathancahill:
This looks good to me.

perliedman:
I agree with @nathancahill, this looks good.

My only concern would be situations where users have custom panes with some element on top (say a `div`), and then put a `canvas` or `svg` inside - these would not get the Leaflet styling with this patch. Whether that is something that people actually do or we should care... I'm not so sure.

jahed:
So that it's more clear on what's being selected, shouldn't Leaflet elements use namespaced classes instead of tag selectors? It would avoid conflicts with nested custom elements.

Also as a suggestion, [BEM](http://getbem.com/) has pretty straight forward guidelines which would prevent a lot of conflicts and specificity issues like this.

Issue #4596, opened 2016-05-25
Click on leaflet-popup-tip-container propagates on map and not on underlying object


eliboni:
I've an app that creates a circleMarker at the very point where the user clicks on.
A popup is bind to the circleMarker
When clicking on the circleMarker the popup opens (with the tail pointing in the circle, exactly at where the user clicked). And that's good.

Now, with the popup opened, by clicking again **within the very same circleMarker area** there is an unpredictable behaviour:
1. if the pointer is on on popup tail or at its left/right (i.e. on  leaflet-popup-tip-container) the popup closes but  the event is then propagated to the map: another circleMarker is created.
2. if the pointer is just below  leaflet-popup-tip-container, the popup closes and reopens .

No.2 would be the expected behaviour since the click is anyway within the original circlemarker. 
I tried to add an extra on click event on the circleMarker but I still get behaviour no.1, i.e. the click goes to the map and not on the object. 

Using leaflet 0.7.7

eliboni:
note, what I described looks like what is mentioned in the "a side note" section of #2880

jwoyame:
Looks like this issue can be closed now?

johnd0e:
It looks that the issue is still the case in 1.70..

Issue #2880, opened 2014-08-28
Tapping a Vector Layer doesn't register (sometimes)


atstp:
#### Overview

Tapping on polygons or other vector layers with attached popups seems to only occasionally act as a click on the map instead of the layer. I haven't had much luck in distinguishing what makes a click successful.

It originally came up on the homepage of www.cliffcloud.com which (somewhat) relies upon tapping vector layers. 

But, for the sake of the bug it should be easier to work with Leaflet's [example](http://leafletjs.com/examples/quick-start-example.html) because it demonstrates the same issue, and has an event attached to the map which shows a little more.
#### To reproduce

at http://leafletjs.com/examples/quick-start-example.html
Here's the code that creates the circle:

```
L.circle([51.508, -0.11], 500, {
  color: 'red',
  fillColor: '#f03',
  fillOpacity: 0.5
}).addTo(map).bindPopup("I am a circle.");
```

But sometimes there's this:

> ![oh no](https://gist.githubusercontent.com/atstp/02b49cc241f24a84d37f/raw/7bab4867acb852cb4776b5bccc3cbed3d087837f/ChromeOnAndroid.png)  
> 
> which is the expected behavior for clicking on the map
#### Versions

> Leaflet: V0.7.2
> Android: V4.4.2
> Chrome: 36.0.1985.135 ( where it first came up ) 
> 
> > unable to reproduce on Firefox: V31.0 

---
### Sidenote:

I also came across the following when I was trying debug the main issue:

if you click in the area highlighted in this image, i.e. inside the `leaflet-popup-tip-container`, where it overlaps a polygon, you should get the same undesired result: your click acts as a click upon the map instead of the vector layer.  

This part can be reproduced on a desktop (i'm running Chrome 36.0.1985.143 on Ubuntu 14.04).  IT's not really a big problem and almost all the time will just close the popup, which is fine, but It's mentioned here to distinguish between the bug above (touch only) and the `leaflet-popup-tip-container` 'bug' which affects clicks and touches.

> ![ clicking this is no good ](https://gist.githubusercontent.com/atstp/02b49cc241f24a84d37f/raw/2952e4d2f3d0243100a4a4282a3923eb494ef13f/ChromeOnUbuntu.png)

PR #4599, opened 2016-05-25
Disable pointer events on popup tip

jwoyame:
This is one possible solution to #4596 and the sidenote of #2880.

Please see the [comparison demo](http://jsbin.com/fovolugoco/1/edit?html,css,js,output). The new tip style is on the bottom map. Click to the sides of the tip to see how it differs. Pay attention to the cursor as well.

The `.leaftlet-popup-tip-container` is a fairly large 40x20 div under the popup content (see screenshot below). It needs extra space below the tip because of the shadow. This entire invisible region will block pointer events from reaching vectors around the popup tip, which can be a frustration when trying to do any sort of fine clicking. It can also be confusing as any clicks on this 40x20 rectangle register as map clicks.

This modification will make pointer events pass right through the invisible region and the tip element, on browsers that support `pointer-events`. Even for browsers that don't support `pointer-events`, it improves the situation on the sides of the tip. For IE8, the tip is still styled through `.leaflet-oldie` without issue.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -430,6 +430,7 @@
 .leaflet-popup {
 	position: absolute;
 	text-align: center;
+	margin-bottom: 20px;
 	}
 .leaflet-popup-content-wrapper {
 	padding: 1px;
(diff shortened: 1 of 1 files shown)


jwoyame:
The demo has been updated.

mourner:
This is great, thank you!

eliboni:
Works as expected for me as well,  thanks!

Issue #4602, opened 2016-05-26
Leaflet 0.7.7 quick start not working in 1.0.0-rc.1


jwoyame:
The problem
---
When there is a click event registered for the map itself to show a popup, popups no longer work on child layers.

The cause
---
Clicks are being propagated to the map even when handled by the layer's popup. Although the layer's handler runs, the popup is instantly replaced by the map's own popup.

How to reproduce
---
Compare the quick start:
http://leafletjs.com/examples/quick-start-example.html

With this `v1.0.0-rc.1` demo:
http://jsbin.com/novoxeturi/edit?html,output

This occurs in master as well, I submitted a demo for it (#4601).

af7:
Did something change on click events? It appears that a click event on a "rectangle" does not fire at all. No popup exists on the map. We just want the rectangle to be clicked upon and fire an event, but this does not work anymore. Did anything change and created a regression?

PR #4603, opened 2016-05-26
Prevent map click when layer has popup

jwoyame:
This is a fix for #4602.

The code can be tested with `map-popup.html` proposed in #4601.

### DIFF
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -229,6 +229,18 @@ describe('Popup', function () {
 		L.Icon.Default.prototype.options.popupAnchor = popupAnchorBefore;
 	});
 
+	it("prevents an underlying map click for Layer", function () {
+		var layer = new L.Polygon([[55.8, 37.6], [55.9, 37.7], [56.0, 37.8]]).addTo(map);
+		layer.bindPopup("layer popup");
+
+		var spy = sinon.spy();
+		map.on('click', spy);
+
+		expect(spy.called).to.be(false);
+		happen.click(layer._path);
+		expect(spy.called).to.be(false);
+	});
+
 });
 
 describe("L.Map#openPopup", function () {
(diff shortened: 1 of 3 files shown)


yohanboniface:
LGTM.
What happens here is that because there is a popup attached both to the vector AND to the map, when clicking on the vector:
- the event is fired on the vector and the popup is opened
- the event is fired on the map, and the first popup is closed before opening the map one

Stopping the event in `_openPopup` is the right thing to do IMHO.

@jwoyame can you add a unittest coverring that case?

jwoyame:
I added unit tests for the Path popup and for the Path itself, to ensure clicks propagate through the Path when there is no popup.

@yohanboniface: That's a correct description. And the reason we don't notice this on Markers is because of the default option set on every marker:

`nonBubblingEvents: ['click', 'dblclick', 'mouseover', 'mouseout', 'contextmenu']`

Unlike Markers, we wouldn't want to always turn off bubbling for Paths (or Layers in general). This fix for layer popups will only stop the propagation for a Layer click when a Popup is being opened, and the new tests make sure of it.

spec/suites/layer/PopupSpec.js
```diff
@@ -229,6 +229,18 @@ describe('Popup', function () {
 		L.Icon.Default.prototype.options.popupAnchor = popupAnchorBefore;
 	});
 
+	it("prevents an underlying map click for Layer", function () {
```
yohanboniface: Those tests are testing what the code is actually doing, while I think it would be better to test what a user should expect. So instead I would make sure the path popup is open, no matter how.
Something like `map.hasLayer(path._popup)`.
What do you think?

jwoyame: I started out with the intent to check for a popup, but that is one symptom of an underlying issue: the click is getting through to the map. So maybe check for both? The tests would be:

1. Make sure popups work on Paths
2. Make sure Paths with popups don't propagate clicks to map (what I added)
3. Make sure that when the Path and map both have a popup, the popup matches the path._popup (to add)

Sound good?

yohanboniface: > Sound good?

Yup :)

jwoyame: Okay, please check. I was able to avoid adding another test in a way that still turned out pretty clean.

yohanboniface:
Thanks :)

Issue #4604, opened 2016-05-27
Allow full replacement of default marker icons


denydias:
This is a feature request, not a bug.

As per #3696, Leaflet's `L.Icon.Default` only allows changing `iconUrl` and `shadowUrl`. It would be VERY nice if Leaflet also support full `L.Icon.Default` replacement. This would allow, for instance:

``` js
L.Icon.Default = L.Icon.Default.extend({
  options: {
    iconUrl: 'my-new-marker-icon.png',
    iconRetinaUrl: 'my-new-marker-icon@2x.png',
    shadowUrl: 'my-new-marker-shadow.png',
    shadowRetinaUrl: 'my-new-marker-shadow@2x.png',
    iconSize: [ 35, 45 ],
    iconAnchor: [ 17, 42 ],
    popupAnchor: [ 1, -32 ],
    shadowAnchor: [ 10, 12 ],
    shadowSize: [ 36, 16 ],
    className: "my-new-css-class"
  }
});
```

Or even better, although requiring deeper changes:

``` js
L.Icon.Default = L.ExtraMarkers.icon({
  icon: 'fa-home',
  markerColor: 'blue',
});
```

`L.ExtraMarkers.icon()` comes from [Leaflet.extra-markers plugin](https://github.com/coryasilva/Leaflet.ExtraMarkers) (but it could be [Leaflet.awesome-markers](https://github.com/lvoogdt/Leaflet.awesome-markers) or any other handling marker icons).

This is mostly desired for projects sourcing Leaflet from `https://cdnjs.cloudflare.com`, as there is no way to change the source code.

denydias:
It's so nice to see how this issue evolved! Nice work, @IvanSanchez et all.

ghybs:
Hi,

It seems to me that it is actually quite easy to change the default icon already, but there _is_ a trap in that `L.Icon.Default` is referenced in `L.Marker` options at script initialization.
Therefore you cannot just re-assign `L.Icon.Default`  (the `L.Marker` option will keep the original reference)

You just have to override that `L.Marker` option with your desired icon:
```javascript
L.Marker.mergeOptions({
  icon: L.ExtraMarkers.icon({ // whatever you like
    icon: 'glyphicon glyphicon-heart',
    markerColor: 'blue',
  })
});
```

Demo: http://playground-leaflet.rhcloud.com/boda/1/edit?html,output

[...]

Issue #4579, opened 2016-05-18
IE8: L.Icon.Default.imagePath detected incorrectly


perliedman:
In the debug pages, IE8 picks up the path to the default marker icon incorrectly.

Looking at the result marker `img` tag, it har `src="http://192.168.1.53:8081/debug/images/marker-icon.png`", where it should be the equivalent of `http://192.168.1.53:8081/dist/images/marker-icon.png`.

Chrome shows the URL `http://localhost:8081/debug/../src/../dist/images/marker-icon-2x.png` on the same page, so I guess IE8 is thrown off by multiple `..` in the path.

IvanSanchez:
Shall we just inline the image in a `data:` or `blob:` URI and get that over with?

perliedman:
Sounds like that would avoid some problems.

But it would also force everyone to load the default marker images and shadows, even if they're not using them... I guess that will not be really popular.

johnd0e:
Should be resolved by #4605

Issue #3696, opened 2015-07-30
L.Icon.Default(options) icon retina override


jonenst:
Hi,
With leaflet 0.7.3 and current master, the Default icon allows to override the iconUrl and shadowUrl but not the iconRetinaUrl. See http://jsfiddle.net/vcbvzr55/2/ (note: you can test the retina mode by zooming in firefox)

The attached patch fixes the behavior for icon2 while maintaining the "degradation to normal icon" behavior of icon3, feel free to use it in any way you like. I noticed @mourner and @danzel worked on this part of the code.
- jsfiddle top line: current behavior:
  - icon1 (left): OK: uses retina icons
  - icon2 (middle): NOK: doesn't use the supplied retina icon
  - icon3 (right): OK: correctly uses the default icon even in retina mode because iconRetinaUrl is not defined but iconUrl is defined
- jsfiddle bottom line: after patch:
  - icon5 (middle): OK: uses retina icons

It might be useful to also update the docs:
Change

> L.Icon.Default extends L.Icon and is the blue icon Leaflet uses for markers by default.

To

> L.Icon.Default extends L.Icon and is the blue icon Leaflet uses for markers by default. The constructor accepts the same options as L.Icon except it ignores 'shadowRetinaUrl'. You can use the L.Icon.Default.imagePath property to use your additional images from leaflet's default folder, for example:
> 
> ``` javascript
> new L.Icon.Default({
>     iconUrl: L.Icon.Default.imagePath + "/marker-icon-green.png",
>     iconRetinaUrl: L.Icon.Default.imagePath + "/marker-icon-green-2x.png"
> });
> ```

patch:

``` diff
--- /tmp/Icon.Default.js        2015-07-30 17:38:54.504167213 +0200
+++ /tmp/Icon.Default-retina.js 2015-07-30 17:44:33.844183175 +0200
@@ -13,6 +13,14 @@

        _getIconUrl: function (name) {
                var key = name + 'Url';
+               var isRetinaIcon = L.Browser.retina && name === 'icon';
+
+               if (isRetinaIcon) {
+                       var Retinakey = name + 'RetinaUrl';
+                       if (this.options[Retinakey]) {
+                               return this.options[Retinakey];
+                       }
+               }

                if (this.options[key]) {
                        return this.options[key];
@@ -24,7 +32,7 @@
                        throw new Error('Couldn\'t autodetect L.Icon.Default.imagePath, set it manually.');
                }

[...]

PR #4605, opened 2016-05-27
CSS-based paths for L.Icon.Default. #4604, #3696, #4579

IvanSanchez:
Takes some inspiration from #3559 (thanks,@sheppard!). This:

* Allows overriding the retina URLs for `L.Icon.Default` (#3696)
* Allows replacing the autodetected path for `L.Icon.Default` images (#4604)
* The CSS method from #3559 *should* make IE8 compute paths properly (#4579)
* Adds unit tests for the default images (just checking image size, which means default icon images are loaded)
* Removes the phantomJS hack in the unit tests (by adding `leaflet.css` to the Karma conf), as the CSS method enables phantomJS to autodetect the path!!

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -32,11 +32,11 @@ L.Icon.Default = L.Icon.extend({
 		// way, set this option to point to the right absolute path.
 		if (!('imagePath' in this.options)) {
 			var el = L.DomUtil.create('div',  'leaflet-default-icon-path', document.body);
-			this.options.imagePath = L.DomUtil.getStyle(el, 'background-image')
-				.replace(/^url\([\"\']?/, '')
-				.replace(/[\"\']?\)$/, '');
+			var path = L.DomUtil.getStyle(el, 'background-image');
 
-			console.log(this.options.imagePath);
+			this.options.imagePath = path.indexOf('url') === 0 ?
+				path.replace(/^url\([\"\']?/, '').replace(/[\"\']?\)$/, '') : '';
+
 			document.body.removeChild(el);
 		}
 
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -32,7 +32,7 @@ L.Icon.Default = L.Icon.extend({
 		// way, set this option to point to the right absolute path.
 		if (!('imagePath' in this.options)) {
 			var el = L.DomUtil.create('div',  'leaflet-default-icon-path', document.body);
-			var path = L.DomUtil.getStyle(el, 'background-image');
+			var path = L.DomUtil.getStyle(el, 'background-image') || L.DomUtil.getStyle(el, 'backgroundImage');
 
 			this.options.imagePath = path.indexOf('url') === 0 ?
 				path.replace(/^url\([\"\']?/, '').replace(/[\"\']?\)$/, '') : '';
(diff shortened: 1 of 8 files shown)


jwoyame:
This is very clean, which makes me happy. Really awesome work. The only potential issue I could see: @sheppard used an actual image for the test, would `background-image: url(images/);` be too dependent on browser behavior? Are there any browsers that would choke on this or generate a warning in the console?

mourner:
I like this! If all support browsers are fine with this, me too. The only minor thing I'd change is to still support the global `L.Icon.Default.imagePath` for compatibility — e.g. if it is set, use it instead of autocomputing.

IvanSanchez:
@mourner Just tried it in IE8, it just needed a minor tweak. Will have a look at backwards compatibility of `L.Icon.Default.imagePath`.

@jwoyame In the browsers I've tested this, the CSS rule doesn't even make a network request for the image, it just sets the string in the style property. The exception is PhantomJS, which sets the style, makes a (failing) network request, and still works as expected.

jwoyame:
@IvanSanchez I will give your code a spin. I just tried creating a simple test page (on IIS) where a DIV had the `.leaflet-default-icon-path` class, and I got a 403 error when the `images\` folder exists and a 404 when it doesn't. I tried it by adding the DIV dynamically as well. But like I said, I will need to try it just as you have it.

IvanSanchez:
It seems no errors are shown when this is used in a map, but shows errors when the unit tests are run. If this is an issue, we could replace `images/` with `images/marker-icon.png` in the CSS, and make the regexp a tiny bit more complicated.

src/layer/marker/Icon.Default.js
```diff
@@ -1,45 +1,51 @@
...
+		// `L.Icon.Default` will try to auto-detect the absolute location of the
+		// blue icon images. If you are placing these images in a non-standard
+		// way, set this option to point to the right absolute path.
+		if (!('imagePath' in this.options)) {
 
-		if (!path) {
-			throw new Error('Couldn\'t autodetect L.Icon.Default.imagePath, set it manually.');
-		}
+			if (L.Icon.Default.imagePath) {	// Deprecated, backwards-compatibility only
+				this.options.imagePath = L.Icon.Default.imagePath;
```
mourner: There's no need to reset the original `imagePath` property in options. I think this should be just a variable, used throughout the function. Otherwise I think this is ready to merge!

IvanSanchez: This doesn't reset the variable, just sets it once. Or do you mean something else?

mourner: I mean just do `var path = ...`.

sheppard: It makes sense not to overwrite `options`, though the computed value needs to be persisted somewhere, otherwise we'll end up re-computing it whenever `_getIconUrl` is called.  As discussed in #3559, the reason to do the detection during icon creation rather than during startup is to avoid issues with the DOM not being available during startup.  I was using `L.Icon.Default.imagePath` to store the initial result.

hyperknot: I think writing to options was considered a bad practice in some other tickets. If so, what about just using L.Icon.Default.imagePath to store this value?

perliedman:
I made an attempt to address the comments from @mourner, @sheppard and @hyperknot so that `options` are not overwritten, but otherwise functionally equivalent.

Feel free to revert this if you feel there's some issue with this apprach.

IvanSanchez:
@perliedman Looks good to me. I'll try to rebase to get rid of the merge conflict.

PR #4606, opened 2016-05-27
Gh pages

marslan390:
Leaflet.BeautifyMarker, a lightweight plugin that adds colorful iconic markers without images for Leaflet by giving full control of style to end user ( i.e. unlimited colors and many more...). It has also ability to adjust font awesome and glyphicon icons


yohanboniface:
@marslan390 you PR is targeting the wrong branch, you should use `gh-pages`.

PR #4608, opened 2016-05-30
Icon size x can now be initialised with a number

### LINKED ISSUE #3185: Accept Icon iconSize as number
Consider the following code snippet

```Javascript
myLayer.setIcon(new L.Icon({
    'iconUrl': myIconUrl,
    'iconSize': 96
});
```

This will reach the function called _setIconStyles in src/layer/marker/Icon.js 

Notice that the anchor variable will be instantiated effectively by going
```Javascript
var anchor = L.point(96/2);
```
At this stage, anchor.x is defined but anchor.y is undefined. That anchor.y is undefined is the key fact here.

Further down in _setIconStyles, the function says
```Javascript
img.style.marginTop  = (-anchor.y) + 'px';
// Which is the same as 
// img.style.marginTop = "NaNpx"
```

On Chrome, this works fine however on IE8, this will cause an "Invalid argument." error.

 You can work around this issue by doing the following
```Javascript
layer.setIcon(new L.Icon({
    'iconUrl': myIconUrl,
    'iconSize': [myIconSize, myIconSize],
    'iconAnchor': [myIconAnchor, myIconAnchor]
}));
```

Arguably I should have been doing this from the start however it would be a fairly simple fix to instantiate L.point with both an x and a y if the assumption that the icon is a square is acceptable.

rheh:
Icon size can now be passed as a single numeric as well as the [x, y].  Additional tests added to cover new parameter option.

Fixes #3185

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -56,21 +56,6 @@ describe("Marker", function () {
 			expect(icon.style.width).to.be(expectedXY + 'px');
 			expect(icon.style.height).to.be(expectedXY + 'px');
 		});
-	});
-
-	describe("#setIcon", function () {
-
-		it("changes the icon to another image", function () {
-			var marker = new L.Marker([0, 0], {icon: icon1});
-			map.addLayer(marker);
-
-			var beforeIcon = marker._icon;
-			marker.setIcon(icon2);
-			var afterIcon = marker._icon;
-
-			expect(beforeIcon).to.be(afterIcon);
-			expect(afterIcon.src).to.contain(icon2._getIconUrl('icon'));
-		});
 
 		it("changes the icon to another image", function () {
 			var marker = new L.Marker([0, 0], {icon: icon1});
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -23,6 +23,55 @@ describe("Marker", function () {
 	});
 
 	describe("#setIcon", function () {
+
+		it("set the correct x and y size attributes", function () {
+			var expectedX = 96;
+			var expectedY = 100;
+			var sizedIcon = new L.Icon.Default({
+				iconUrl: icon1._getIconUrl('icon') + '?3',
+				iconSize: [expectedX, expectedY]
+			});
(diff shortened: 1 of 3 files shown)


src/layer/marker/Icon.js
```diff
@@ -47,8 +47,14 @@ L.Icon = L.Class.extend({
 	},
 
 	_setIconStyles: function (img, name) {
-		var options = this.options,
-		    size = L.point(options[name + 'Size']),
+		var options = this.options;
+		var sizeOption = options[name + 'Size'];
+
+		if (!L.Util.isArray(sizeOption)) {
+			options[name + 'Size'] = [options[name + 'Size'], options[name + 'Size']];
```
mourner: You can use `sizeOption` here twice for shorter code, right?

mourner: Also, you're overwriting the original options object, which is undesirable. Instead you could redefine `sizeOption`.

IvanSanchez:
Not sure about this. I'd rather extend the `L.point` factory method to allow single-parameter instantiation everywhere if we're gonna allow this.

mourner:
@IvanSanchez I'm not in favor of `L.point` accepting a single number because it can be a source of subtle, hard to track bugs.

src/layer/marker/Icon.js
rheh: Thanks - both of your comments have been implemented in the latest commit.

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -23,6 +23,55 @@ describe("Marker", function () {
...
+			map.addLayer(marker);
+
+			var beforeIcon = marker._icon;
+			marker.setIcon(icon2);
+			var afterIcon = marker._icon;
+
+			expect(beforeIcon).to.be(afterIcon);
+			expect(afterIcon.src).to.contain(icon2._getIconUrl('icon'));
+		});
+
```
jwoyame: I'm not looking at the whole file, but it looks like line 61 to 74 was accidentally duplicated?

rheh: I think you're right.  I'll update the pull request to correct this tomorrow.  Thanks.

Issue #4613, opened 2016-06-01
eslint breaks unit tests


IvanSanchez:
In my linux environment, the command-line unit tests are failing with the following stack trace:

```
ivan:~/devel/Leaflet\ $ jake test
Checking for JS errors...
Cannot find module 'eslint-config-package.json'
Error: Cannot find module 'eslint-config-package.json'
    at Object.ModuleResolver.resolve (/home/ivan/devel/Leaflet/node_modules/eslint/lib/util/module-resolver.js:75:19)
    at resolve (/home/ivan/devel/Leaflet/node_modules/eslint/lib/config/config-file.js:473:33)
    at Object.load (/home/ivan/devel/Leaflet/node_modules/eslint/lib/config/config-file.js:490:24)
    at loadConfig (/home/ivan/devel/Leaflet/node_modules/eslint/lib/config.js:64:33)
    at new Config (/home/ivan/devel/Leaflet/node_modules/eslint/lib/config.js:197:38)
    at CLIEngine.executeOnFiles (/home/ivan/devel/Leaflet/node_modules/eslint/lib/cli-engine.js:585:28)
    at Object.cli.execute (/home/ivan/devel/Leaflet/node_modules/eslint/lib/cli.js:181:95)
    at Object.<anonymous> (/home/ivan/devel/Leaflet/node_modules/eslint/bin/eslint.js:77:20)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
jake aborted.
```

I suspect this is due to the changes made in #4571. I need to check why this doesn't fail in Jenkins, and read some eslint documentation.

mourner:
Does the above fix it for you?

IvanSanchez:
@mourner Yeah, the tests run again after pulling the latest `master` branch.

Issue #4039, opened 2015-11-22
Tile pruning too agressive?


yohanboniface:
1.0 master:
![leaflet-tile-prune-1 0](https://cloud.githubusercontent.com/assets/146023/11323148/e01d9ed6-910a-11e5-9e36-118740d0c86c.gif)

0.7:
![leaflet-tile-prune-0 7](https://cloud.githubusercontent.com/assets/146023/11323134/7eb0b2b4-910a-11e5-86de-fe317da217c5.gif)

Git bisect points to https://github.com/Leaflet/Leaflet/commit/a4e8f4e1a8fd10c0cd701997c85bafe9894b216f

@mourner @IvanSanchez any clue, or should I investigate? :)

mourner:
Yeah, at some point I removed it and then didn't readd it back. Need to look whether it's worth doing at all.

yohanboniface:
> Yeah, at some point I removed it and then didn't readd it back. Need to look whether it's worth doing at all.

@mourner, not sure what "it" refers to in your sentence ;)

IvanSanchez:
I'm unable to see how the bisected commit affects tile pruning :-S

Maybe it's time to see if #3551 is not a bad idea after all?

mourner:
@yohanboniface I referred to retaining tiles on the current zoom when panning. I removed it when reworking the logic meaning to add it back some time later, but never did. I'm not sure we should though — or maybe we should but in a limited form (e.g. retain tiles only within X pixels from the current screen). Retaining already loaded tiles is different from #3551 though (which is loading yet unseen tiles).

hyperknot:
Can be checked visually on #4206. I recommend keeping at least one tile in memory on each sides.

What is the downside of keeping more tiles? Higher memory usage?

kaynz:
@hyperknot Bloated DOM,

mourner:
I think this can be fixed after rc1.

hyperknot:
One scenario where this will matter is smooth zooming with zoomSnap = 0, zoomAnimation = false and trackpad scrolling. 

On current master, this trigger thousands of tile loads even with a simple zoom-out, zoom-in.
http://playground-leaflet.rhcloud.com/cap/2/edit?output

(might need trackpad to reproduce smoothness)

jmullo:
The amount of tiles retained on each side should be configurable.

pke:
[...]

PR #4614, opened 2016-06-01
Add the 'bufferSize' option to GridLayer to increase tiles kept (#4039)

IvanSanchez:
Basically reuses some of the ideas from #3551 and the square-by-default tileSize option.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -120,7 +120,15 @@ L.GridLayer = L.Layer.extend({
 
 		// @option className: String = ''
 		// A custom class name to assign to the tile layer. Empty by default.
-		className: ''
+		className: '',
+
+		// @option bufferSize: Number|Point = [256, 256]
+		// How many pixels this grid extends out of the visible map container.
+		// This increases the amount of tiles to be loaded and kept. Lower values
+		// will decrease network usage, and higher values will prevent grey areas
+		// when dragging or zooming quickly. Use a number if the top/bottom and
+		// left/right buffer sizes are equal, or `L.point(leftRight, topBottom)` otherwise.
+		bufferSize: 256
 	},
 
 	initialize: function (options) {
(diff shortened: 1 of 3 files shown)


IvanSanchez:
Note to self: add `bufferSize: 0` to all unit tests involving tile counts, and fix #4613 in order to check locally.

nathancahill:
Awesome. Really excited to see this.

IvanSanchez:
@nathancahill Feel free to check out the branch and test ;-)

src/layer/tile/GridLayer.js
```diff
@@ -256,6 +264,11 @@ L.GridLayer = L.Layer.extend({
 		return s instanceof L.Point ? s : new L.Point(s, s);
 	},
 
+	_getBufferSize: function () {
+		var s = this.options.bufferSize;
+		return s instanceof L.Point ? s : new L.Point(s, s);
```
mourner: I would rather initialize `_bufferSize` in constructor.

mourner:
> Note to self: add bufferSize: 0 to all unit tests involving tile counts

Why add this if it's 0 by default anyway?

src/layer/tile/GridLayer.js
IvanSanchez: Shall we do the same with `getTileSize` then?

IvanSanchez:
> Why add this if it's 0 by default anyway?

I added a default of 256px of buffer, given the complaints from #4039.

hyperknot:
The problem with this approach is that this loads the hidden tiles even if the user has no intention of viewing them. For example, on tiny 600x400 map embed (cca. 6 tiles) we are loading 28 tiles now. That can eat through a lot of tile credits quite quickly.

A better solution would be to tweak the tile _pruning_ code, such that it keeps _buffer_ tiles in memory instead of cleaning + reloading them later.

pke:
@hyperknot yes the pruning code is way to aggressive imho. Panning back and forth a map view should not prune the borders of the map. Also for zooming it should keep more tiles in memory to make the experience seamless and not riddled with white tiles.

IvanSanchez:
Superseded by #4650

PR #4616, opened 2016-06-02
Add links onto wikipedia page about WMS, fix error in example

myshov:
Somewhat related to #4615 
Just an adding of useful links.


mourner:
Thanks!

PR #4618, opened 2016-06-03
Polygon contains

jcoupey:
Implementation of a `contains` method for polygons, as proposed at [uservoice](https://leaflet.uservoice.com/forums/150880-ideas-and-suggestions-for-leaflet/suggestions/6154146-add-method-to-determine-if-a-point-is-contained-by).

This doesn't add much to the codebase (+45 LOC) while providing a nice feature.

The different behaviour of `polygon.getBounds().contains(obj)` and `polygon.contains(obj)` can be viewed at  http://plnkr.co/FMWoraRGZPzXbKCdaxoL?p=preview, where the polygon bounding box is drawn for convenience.

Notes: 
- so far this method is OK for single-ring polygons and polygons with holes. To make it work for multipolygons, I would need more insight into how these should be handled;
- the _checkLeft function would probably fit better somewhere else in the code...

### DIFF
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -87,33 +87,38 @@ L.Polygon = L.Polyline.extend({
 	// @method contains (latlng: LatLng): Boolean
 	// Returns `true` if the polygon contains the given point.
 	contains: function (obj) {
-		if (!this.getBounds().contains(obj)) {
-			return false;
-		} else {
-			// check based on the winding number method
-			var wn = 0,
-			    nodes = this.getLatLngs()[0];
+		var contained = !this.isEmpty() &&
+				this.getBounds().contains(obj) &&
+				this._ringContains(this._latlngs[0], obj);
 
-			for (var i = 0, len = nodes.length; i < len; ++i) {
-				// through all edges of the polygon, with last edge linked to
-				// the first node
-				var next = (i < len - 1) ? i + 1 : 0;
-				if (nodes[i].lat <= obj.lat) {
-					if ((nodes[next].lat > obj.lat) &&
-							(this._checkLeft(obj, nodes[i], nodes[next]) > 0)) {
-						// upward crossing with obj on the left of current polygon
-						// edge
-						++wn;
-					}
-				}	else if ((nodes[next].lat <= obj.lat) &&
-									 (this._checkLeft(obj, nodes[i], nodes[next]) < 0)) {
-					// downward crossing with obj on the right of current
-					// polygon edge
-					--wn;
-				}
-			}
-			return (wn !== 0);
+		for (var i = 1, len = this._latlngs.length; contained && i < len; ++i) {
+			contained = contained && !this._ringContains(this._latlngs[i], obj);
 		}
+		return contained;
+	},
+
+	_ringContains: function (latLngs, obj) {
+		// check based on the wind number method
+		var wn = 0;
+
+		for (var i = 0, len = latLngs.length; i < len; ++i) {
+			// through all edges of the ring, with last edge linked to the
+			// first node
+			var next = (i < len - 1) ? i + 1 : 0;
+			if (latLngs[i].lat <= obj.lat) {
+				if ((latLngs[next].lat > obj.lat) &&
+						(this._checkLeft(obj, latLngs[i], latLngs[next]) > 0)) {
+					// upward crossing with obj on the left of current ring edge
+					++wn;
+				}
+			}	else if ((latLngs[next].lat <= obj.lat) &&
+								 (this._checkLeft(obj, latLngs[i], latLngs[next]) < 0)) {
+				// downward crossing with obj on the right of current ring
+				// edge
+				--wn;
+			}
+		}
+		return (wn !== 0);
 	},
 
 	_checkLeft: function (obj, latlng1, latlng2) {
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -84,6 +84,50 @@ L.Polygon = L.Polyline.extend({
 		return this._map.layerPointToLatLng(center);
 	},
 
+	// @method contains (latlng: LatLng): Boolean
+	// Returns `true` if the polygon contains the given point.
+	contains: function (obj) {
+		if (!this.getBounds().contains(obj)) {
+			return false;
+		} else {
+			// check based on the winding number method
+			var wn = 0,
+			    nodes = this.getLatLngs()[0];
+
+			for (var i = 0, len = nodes.length; i < len; ++i) {
+				// through all edges of the polygon, with last edge linked to
+				// the first node
+				var next = (i < len - 1) ? i + 1 : 0;
+				if (nodes[i].lat <= obj.lat) {
+					if ((nodes[next].lat > obj.lat) &&
+							(this._checkLeft(obj, nodes[i], nodes[next]) > 0)) {
+						// upward crossing with obj on the left of current polygon
+						// edge
+						++wn;
+					}
+				}	else if ((nodes[next].lat <= obj.lat) &&
+									 (this._checkLeft(obj, nodes[i], nodes[next]) < 0)) {
+					// downward crossing with obj on the right of current
+					// polygon edge
+					--wn;
+				}
+			}
+			return (wn !== 0);
+		}
+	},
+
+	_checkLeft: function (obj, latlng1, latlng2) {
+		// Return a positive (resp. negative) value if obj is left
+		// (resp. right) of the oriented line defined by latlng1 and
+		// latlng2 (using the (latlng1 latlng2)x(latlng1 obj) cross
+		// product z-component to check).
+		var u1 = latlng2.lng - latlng1.lng,
+		    u2 = latlng2.lat - latlng1.lat,
+		    v1 = obj.lng - latlng1.lng,
+		    v2 = obj.lat - latlng1.lat;
+		return u1 * v2 - u2 * v1;
+	},
+
 	_convertLatLngs: function (latlngs) {
 		var result = L.Polyline.prototype._convertLatLngs.call(this, latlngs),
 		    len = result.length;
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -87,33 +87,34 @@ L.Polygon = L.Polyline.extend({
 	// @method contains (latlng: LatLng): Boolean
 	// Returns `true` if the polygon contains the given point.
 	contains: function (obj) {
-		var contained = !this.isEmpty() &&
+		var point = this._map.latLngToLayerPoint(obj),
+		    contained = !this.isEmpty() &&
 				this.getBounds().contains(obj) &&
-				this._ringContains(this._latlngs[0], obj);
+				this._ringContains(this._rings[0], point);
 
-		for (var i = 1, len = this._latlngs.length; contained && i < len; ++i) {
-			contained = contained && !this._ringContains(this._latlngs[i], obj);
+		for (var i = 1, len = this._rings.length; contained && i < len; ++i) {
+			contained = contained && !this._ringContains(this._rings[i], point);
 		}
 		return contained;
 	},
 
-	_ringContains: function (latLngs, obj) {
+	_ringContains: function (ring, point) {
 		// check based on the wind number method
 		var wn = 0;
 
-		for (var i = 0, len = latLngs.length; i < len; ++i) {
+		for (var i = 0, len = ring.length; i < len; ++i) {
 			// through all edges of the ring, with last edge linked to the
 			// first node
 			var next = (i < len - 1) ? i + 1 : 0;
-			if (latLngs[i].lat <= obj.lat) {
-				if ((latLngs[next].lat > obj.lat) &&
-						(this._checkLeft(obj, latLngs[i], latLngs[next]) > 0)) {
-					// upward crossing with obj on the left of current ring edge
+			if (ring[i].y <= point.y) {
+				if ((ring[next].y > point.y) &&
+						(this._checkLeft(point, ring[i], ring[next]) > 0)) {
+					// upward crossing with point on the left of current ring edge
 					++wn;
 				}
-			}	else if ((latLngs[next].lat <= obj.lat) &&
-								 (this._checkLeft(obj, latLngs[i], latLngs[next]) < 0)) {
-				// downward crossing with obj on the right of current ring
+			}	else if ((ring[next].y <= point.y) &&
+								 (this._checkLeft(point, ring[i], ring[next]) < 0)) {
+				// downward crossing with point on the right of current ring
 				// edge
 				--wn;
 			}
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I must question whether this is any better than https://github.com/mapbox/leaflet-pip

jcoupey:
@IvanSanchez the ray-casting algorithm used there has some drawbacks, see detailed approaches comparison at http://geomalgorithms.com/a03-_inclusion.html.

leaflet-pip does way more than just point in polygon check (e.g. point location problem for a bunch of polygons). Whereas this PR just aims at adding a `contains` method as a core feature.

hyperknot:
I like the idea of having a robust `.containsPoint()` method for polygons in core. That can be actually quite useful for many users and would help plugin developers not to reimplement the same functionality.

Couple of notes:
- please call it `.containsPoint()` if it can only check for a point. `contains` for me would mean that we are checking for any other geometry object within a polygon, like an other polygon or a line.
- best place for helper functions is geometry / PolyUtil and LineUtil I believe?
- checkLeft should also be called a bit more descriptive, as it can only check a point with a line. checkPointLeft maybe in LineUtil?
- Can you impement multipolygon? I think it'd need to be complete before it can be added to core.

SuberFu:
For a bit of help. My understanding is that for multi-polygons, if a point is within any of the polygon, the point is considered to be "contained" in the multi-polygon.

jcoupey:
@hyperknot Thanks for your notes. So there should be both a `containPoint` and a `containLatLng` method? 

The checkLeft helper can be inlined if it is of no use anywhere else in the codebase (also making the implementation self-contained).

For multipolygon, the "containment" as described by @SuberFu is just a matter of iterating over the polygons.

hyperknot:
@SuberFu @jcoupey I think they can also make holes in each other, which makes things complicated but I don't know if Leaflet is using any specification here. Let's ask @mourner and @yohanboniface.

I believe LatLng is specifically geographical while this method is an abstract geometry one, thus I think a `.containsPoint` would be enough.

IvanSanchez:
@jcoupey I do agree that the method should be called `containsLatLng`. Something like `containsPoint` might lead to confusion.

If @hyperknot is OK about having this in the core, then I'm OK with having this in the core. Yes to multipolygon implementation, yes to moving helper funcs into `PolyUtil`.

Also please make sure your definition of "contains" conforms to https://en.wikipedia.org/wiki/DE-9IM . What happens to points which lay precisely on the edge of a polygon? Also, according to that, points inside an inner ring are "outside" the polygon.

IvanSanchez:
> I believe LatLng is specifically geographical while this method is an abstract geometry one, thus I think a `.containsPoint` would be enough.

Internally, `L.Polyline` and `L.Polygon` use `LatLng`s, not `Point`s. Naming it `containPoints` is confugins. consufing. confusing.

yohanboniface:
As often, PostGIS is a good friend: http://postgis.net/docs/manual-2.1/ST_Contains.html is a nice explanation on how `contains` should behave. :)

Also, if we add this to the core, we'll need a bunch of tests (MultiPolygons, hole, Polyline…).

hyperknot:
OK, I agree, I realised Point is really pixel specific. (But LatLng is geographic specific, leaving actually nothing for abstract points.)

hyperknot:
Also turf.js has a similar implementation:
https://github.com/Turfjs/turf/blob/master/packages/turf-inside/index.js#L64

SuberFu:
@IvanSanchez 

> Internally, L.Polyline and L.Polygon use LatLngs, not Points. Naming it containPoints is confugins. consufing. confusing.

Perhaps `containsLatLng`?

P.S. Should it be "contain" or "contains".

mourner:
I'm fine including this, but note that there is already a PIP algorithm in Leaflet core used for Canvas click detection: https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L346-L368

Let's DRY this up and I'll be 👍 on merging.

mourner:
`contains` is a good name and it's consistent with the same method for `LatLngBounds`.

SuberFu:
@jcoupey 
New question on this. Given that the polygon is, well, on a globe. Won't the "contain" calculation being implemented right now essentially assumes a straight lat-lng to x-y projection? How does it handle dateline crossing? How does it handle alternative projection scheme?

My I suggest that instead of calculating `contains` on the lat-lng, project them back to point using `this._map.project` before doing the contain calculation. This way you can support custom projection schemes.

SuberFu:
@mourner 

> I'm fine including this, but note that there is already a PIP algorithm in Leaflet core used for Canvas click detection: https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L346-L368
> 
> Let's DRY this up and I'll be :+1: on merging.

It looks like the existing hit detection uses the "evenodd" algorithm, and so is the rest of the leaflet. The suggested implementation by @jcoupey uses winding number. Granted, the differences doesn't matter if the polygon isn't self intersecting.

hyperknot:
@SuberFu raises a good point, that is that the visual lines of a polygon are in screen space, while the algorithm runs in coordinate space, thus it can happen that the two are not returning the same results. 

But with a simple Mercator it should not be an issue as it is linear, or isn't it? I think if there was a problem we would have run into it in the Canvas code linked by @mourner.

SuberFu:
@hyperknot 
> But with a simple Mercator it should not be an issue as it is linear, or isn't it? I think if there was a problem we would have run into it in the Canvas code linked by @mourner.

I believe the canvas code uses projected points, not lat-lng, for computation. As such, the containPoints operation operates on the L.Point in screen space (or something equivalent to a flat plane) instead of L.LatLng.

jcoupey:
So the remaining tasks would be:

- [ ] rename and move the helper function to a better place
- [ ] implement for multipolygons
- [ ] make sure the specs are ok (border inclusion, holes, `contains` for multipolygons...)
- [ ] adding corresponding tests

I'm happy to try handling this.

From what I gather from the above discussion, the method should use latLng only as argument but work internally with projected points.

@IvanSanchez Regarding boundaries inclusion, currently all edges facing southward or westward are considered inside the polygon while all edges facing eastward and northward are outside. Useful for polygons sharing an edge (a point on the edge is always in one of them only).
Discarding the points that are 'inside' holes (in the same sense) should be consistent for the above edges orientation rules.

yohanboniface:
>  rename and move the helper function to a better place

Bonus point if the function is a helper we can call with any latlngs array (or array of arrays), so we can use it without the context of a Polygon, or for example for knowing if a point is in a given shape of a multipolygon.

SuberFu:
@jcoupey 
May I suggest move it into PolyUtil? That seems to be a good place for this.

jcoupey:
@SuberFu As you previously suggested, I want to use projected points (not latlngs) for the check. In order to move the actual check to PolyUtil __and__ be able to call it with a latngs array as suggested by @yohanboniface, the projection should take place within PolyUtil. But then I don't have access to the `project` function from map. Or am I missing something here?

hyperknot:
@jcoupey maybe take map as an optional parameter which would be required if checking on projection is needed?

jcoupey:
This also questions whether it is relevant to work with points rather than latlngs, because projections involve rounding.
I have a test case where a `L.Polygon([[3, 1], [1, 3],...])` will fail to contain a edge-point like `L.LatLng(2, 2)`, only because the projected points are not quite in line. I think no-one would normally expect this to happen.

hyperknot:
@jcoupey but I believe that is what's supposed to happen. If you draw a line on screen which goes between [3, 1], [1, 3], then it won't go through [2, 2].

If we split the functions to containsLatLng and containsLatLngProjected, then the contains ones should be on the line but the projected one should not be.

jcoupey:
Helper functions renamed to  `pointToLineOrientation` in `LineUtil` and `ringContains` in `PolyUtil`. The latest can be used with points or latlngs (using the optional map parameter) in which case all latlngs are projected first using `map.project`.

As should be expected: being inside a "single" polygon means being inside the outer ring and in none of the holes (see `_singleContains`). Being inside a multipolygon means being inside at least one of the single polygons composing it (that's `contains`).

I updated the previous live example to http://plnkr.co/qpTCklct5l8XIS8xi4bU.

Complexity concern: for each check of a latlng against a polygon ring, all polygon latlngs are re-projected (if the checked latlng has not been ruled out as being outside the bounding box). I was hoping to use `this._rings` rather than `this._latlngs` in the polygon context to avoid this but the projections in the rings are rounded via `map.latLngToLayerPoint`, which leads to confusing results.

I tried to come up with relevant tests. Happy to have reviews and comments on the whole stuff.

src/layer/vector/Polygon.js
```diff
@@ -84,6 +84,34 @@ L.Polygon = L.Polyline.extend({
 		return this._map.layerPointToLatLng(center);
 	},
 
+	// @method contains (latlng: LatLng): Boolean
+	// Returns `true` if `latlng` is inside the (multi)polygon.
+	contains: function (latlng) {
```
jieter: Leaflet API usually also allows a simple array as `latlng` argument, shouldn't this function also allow that by adding a `latlng = L.latlng(latlng);` line?

src/layer/vector/Polygon.js
```diff
@@ -84,6 +84,34 @@ L.Polygon = L.Polyline.extend({
 		return this._map.layerPointToLatLng(center);
 	},
 
+	// @method contains (latlng: LatLng): Boolean
+	// Returns `true` if `latlng` is inside the (multi)polygon.
```
jieter: Should the documentation say that the polyline has to be on a map in order for this method to work?

jcoupey: Surely, and probably a good place to mention that all computations are performed on projected points.

src/geometry/PolyUtil.js
```diff
@@ -54,3 +54,45 @@ L.PolyUtil.clipPolygon = function (points, bounds, round) {
...
+	if (objs[0] instanceof L.Point && obj instanceof L.Point) {
+		point = obj;
+		points = objs;
+	} else if (map && objs[0] instanceof L.LatLng && obj instanceof L.LatLng) {
+		point = map.project(obj);
+		points = [];
+		for (var j = 0; j < objs.length; j++) {
+			points.push(map.project(objs[j]));
+		}
+	}
```
jieter: If this is considered a public API, we might want to raise an error about incompatible input arguments in the else clause here. Or just name the arguments `points, point`, and only have the `(map && objs[0] instanceof L.LatLng && obj instanceof L.LatLng)` check for the second case. Now it will fail with something like `points.length` is not defined, when called with `[point, point, point, ...], latlng`, which might be confusing.

src/geometry/PolyUtil.js
```diff
@@ -54,3 +54,45 @@ L.PolyUtil.clipPolygon = function (points, bounds, round) {
...
+	for (var i = 0, len = points.length; i < len; ++i) {
+		// through all ring edges (last edge linked back to the first node)
+		var next = (i < len - 1) ? i + 1 : 0;
+		if (points[i].y <= point.y) {
+			if ((points[next].y > point.y) &&
+					(L.LineUtil.pointToLineOrientation(point, points[i], points[next]) > 0)) {
+				// upward crossing with point on the "left" of current ring edge
+				++wn;
+			}
+		}	else if ((points[next].y <= point.y) &&
```
jieter: Is the tab character intended between `}` and `else if`?

src/layer/vector/Polygon.js
jcoupey: So maybe also throw an error if `this._map` is undefined?

mourner:
Nice progress! Still some items to check off the TODO before merge though, right?

jcoupey:
Oops, sorry I didn't check any of that to report the progress. 
I think everything is working fine now. I tried to fit in the leaflet development habits but I may still be missing things in this area, as @jieter pointed out.

Also tried to make things as clear as possible wrt the expected behaviour, both through the code comments and tests.

mourner:
@jcoupey it seems like there's still duplication between the code here and point-in-poly one in `Canvas.js`.

kkdd:
I realizes that ```L.Polygon.prototype._containsPoint``` in ```Canvas.js``` uses simple ray casting algorithm, and this pull request supplies very similar functionality by counting a winding number. Could these be unified?

For seeing the similarity and difference, see http://geomalgorithms.com/a03-_inclusion.html, or see pointInPolygon.html (https://gist.github.com/kkdd/f53277c3f5cff06ae16063c5ab781105), for example.

jcoupey:
> Could these be unified?

Yes they should. See previous comments from @mourner on this. I should find time soon to go further on this.

jcoupey:
How I gather things from the previous implementation at https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L346-L368 (please correct me if I'm wrong): at the end of the nested `for` loop when `i = 0`, `inside` has been flipped to `true` if `p` is in `this._parts[0]` (the exterior ring). Then `inside` goes back to `false` whenever `p` is in `this._parts[i]` for any `i` starting at 1 (i.e. if `p` is in one of the holes), hoping no two holes contains `p`.

So I simply replaced that by a `_singleContains` call using `_parts`. The good thing is that we no longer need to go through all parts as we now stop as soon as we are sure the point is out (out of the exterior ring or in any hole).

jcoupey:
Hi there, any update on this? As far as I can see, the above comments and advices have been taken into account. The last commit removes the duplicate code.

Maybe you need more time to review or want to wait until the current rc process is over on 1.0.0?

maximal:
Just wondering is it gonna be any update on this?

jcoupey:
@maximal I could probably find some time to try and resolve conflicts with current `master` branch. On the other hand I wonder if this is badly needed as the last update to this PR is nearly 4 years old?

MirzaeiDev:
Hello everyone,

It's disappointing to see that this issue has persisted for 8 years without a push in "master" branch. Unfortunately, it hasn't been fully documented or made available for others to benefit from.

That being said, I’ve worked on this issue and wanted to share my approach. I hope you can overlook any mistakes I made along the way.

### Here’s what happened:
We had an overlapping range problem in our panel where we define multiple ranges for product delivery. These ranges could overlap in certain cases, and we wanted to detect those overlaps using polygons.

### My Wrong Approach:
At first, I attempted a solution using this resource:
[http://plnkr.co/edit/FMWoraRGZPzXbKCdaxoL?p=preview](http://plnkr.co/edit/FMWoraRGZPzXbKCdaxoL?p=preview)

The `contains` function in that example is quite complex and relies on additional helper functions, so I thought including all related functions in my project would solve the issue. I added functions like:

- _ringContains
- _checkLeft
- _project
- _projectLatlngs
Unfortunately, it didn’t work as expected. If anyone is interested, I can share the code I wrote.

### The Right  Approach:
Ultimately, the solution is much simpler than I initially thought. It boils down to a function that checks whether a point is inside a polygon:

```javascript
const isPointInPolygon = (point, polygon) => {
  const x = point[1];
  const y = point[0];
  let inside = false;

  for (let i = 0, j = polygon.length - 1; i < polygon.length; j = i++) {
    const xi = polygon[i][1];
    const yi = polygon[i][0];
    const xj = polygon[j][1];
    const yj = polygon[j][0];

    const intersect = yi > y !== yj > y && x < ((xj - xi) * (y - yi)) / (yj - yi) + xi;
    if (intersect) inside = !inside;
  }

  return inside;
};
```
This function does the job without needing any extra dependencies.

kkdd:
Note that the singularity should be evaded when ```yj - yi = 0```.

jcoupey:
Closing as stale.

PR #4621, opened 2016-06-03
Beautify Marker Plugin Added

marslan390:
Lightweight plugin that adds colorful iconic markers without image and
gives full control of style to end user (i.e. Unlimited colors and CSS
styling).


marslan390:
Ah! it might be some issue with Anchor, i look into it.

marslan390:
Issue related to marker position change on map zoom level change has been fixed, also popupAnchor position set.

marslan390:
How much time will it take to review?

PR #4627, opened 2016-06-07
Fix typo in the description of getWheelDelta

myshov:
The event doesn't map itself, but value of delta does.


IvanSanchez:
No, the function takes in a `DOMEvent` and returns a number. The wording is OK right now.

myshov:
ok

PR #4628, opened 2016-06-07
Experiment: implements L.Handler.PathDrag

yohanboniface:
Experimental handler to allow dragging of path.
No tests, no doc. It's more a try to start a discussion. Works with both SVG and canvas. Also tested with touch.

Thanks in advance for giving feedback, specifically if it's worth cleaning the code to integrate in the core.

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -200,6 +200,12 @@ var deps = {
 		desc: 'Makes markers draggable (by mouse or touch).'
 	},
 
+	PathDrag: {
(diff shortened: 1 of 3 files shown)


kaynz:
Maybe @w8r can help you ;)

yohanboniface:
Sure :) We already have a parallel discussion in https://github.com/w8r/Leaflet.Path.Drag/issues/19 :)

PR #4632, opened 2016-06-08
Fixes control incompatibility with PrimeFaces

Lambdac0re:
PrimeFaces inserts a hidden input field with the id "javax.faces.ViewState" into each form tag on the page.
As this input was not created by the control, it has no layerId and causes these functions to crash.
Test case: Manually insert an input element into the control's form and watch the console while switching layers.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -316,7 +316,7 @@ L.Control.Layers = L.Control.extend({
 	},
 
 	_onInputClick: function () {
-		var inputs = this._form.getElementsByTagName('input'),
+		var inputs = this._form.getElementsByClassName('leaflet-control-layers-selector'),
 		    input, layer, hasLayer;
 		var addedLayers = [],
 		    removedLayers = [];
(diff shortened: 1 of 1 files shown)


myshov:
Hi @Lambdac0re!

I see two issues here:

1. You use method `getElementsByClassName()` which supported only IE9 and higher.
2. I've never used PrimeFaces, but I think it is problem with integration not with leaflet itself. It looks weird for me that PrimeFaces inserts additional inputs in every form on the page. So try to discuss this issue with developers of the framework or try to delete redundant input when page is loaded if it is possible.

IvanSanchez:
I agree with @myshov - Leaflet 1.0 aims to stay compatible with IE8, so `getElementsByClassName()` is not an option as of now.

Also, it's apparently possible to save the state of things in the server and not in the client (via `javax.faces.STATE_SAVING_METHOD`). So finding a way to make primefaces not add inputs everywhere in the page should be possible.

Lambdac0re:
I agree that if you aim to stay IE8 compatible, another selection should be used, but switching the state saving method is not an option on our systems.

IvanSanchez:
Alas, I don't think we'll change this here in upstream Leaflet - fixing the compatibility with a toolkit while breaking compatibility with our target platforms is a no-go. You're welcome to use your own patched version of Leaflet, though.

PR #4633, opened 2016-06-09
Add description to time argument of Util.throttle

myshov:
There was no description for `time` argument in `Util.throttle()`.


IvanSanchez:
Thanks :-)

PR #4638, opened 2016-06-10
Do not rely on options.draggable to compute draggableMoved

yohanboniface:
One can enable dragging of a marker after it has been initialized with draggable=false.

I've started adding uniitests to cover the area, but then remembered that, AFAIK, prosthetic does not fire a click on up at the moment.
So I'm not sure if I should remove those tests. @IvanSanchez I'm on your hands (LOL).

### DIFF
--- a/debug/map/markers.html
+++ b/debug/map/markers.html
@@ -32,12 +32,12 @@
 
 
         var markerDraggable = new L.Marker([0, 10], {
-            draggable: true,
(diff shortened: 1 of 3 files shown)


yohanboniface:
Rebased.
One test is skipped in Travis, but will be fixed when #4705 is merged (by loading CSS).

PR #4641, opened 2016-06-10
Add missing hasOwnProperty check

infostreams:
Without this I got a very cryptic "Cannot read property '_leaflet_id' of undefined" error. This fixes it.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -224,11 +224,13 @@ L.Control.Layers = L.Control.extend({
 		var baseLayersPresent, overlaysPresent, i, obj, baseLayersCount = 0;
 
 		for (i = 0; i < this._layers.length; i++) {
-			obj = this._layers[i];
-			this._addItem(obj);
-			overlaysPresent = overlaysPresent || obj.overlay;
-			baseLayersPresent = baseLayersPresent || !obj.overlay;
-			baseLayersCount += !obj.overlay ? 1 : 0;
+			if (this._layers.hasOwnProperty(i)) {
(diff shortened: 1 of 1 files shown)


nathancahill:
-1. `this._layers` is an array, so the error you're seeing must be coming from somewhere else (in `this._addItem` maybe?) Can you create an minimal example that reproduces the issue? https://playground-leaflet.rhcloud.com/

IvanSanchez:
@infostreams Can you publish a fiddle/codepen/[playground](https://playground-leaflet.rhcloud.com/) that displays this bug?

infostreams:
Basically I did this:

    L.control.layers(null, {
        key_a:new L.TileLayer("url"), 
        key_b:new L.TileLayer("url")
    }, {position: "bottomleft"}).addTo(map);

I can't get your playground to work (I tried, but the default thing it opens is showing javascript errors for me), but this is what triggered the `Cannot read property '_leaflet_id' of undefined` error for me.

If you assume that people are only going to pass in arrays into that function, then I think the original code works. Otherwise we have to add something more involved than what I just proposed

infostreams:
Huh? Wait, my version of leaflet-src.js says `for (i in this._layers) {` just above the `if (this._layers.hasOwnProperty(i)) {` I added, but the version I made the pull request for says `for (i = 0; i < this._layers.length; i++) {`....

infostreams:
Related thread: https://github.com/Leaflet/Leaflet/issues/4508. Uhm, close? I guess?

nathancahill:
Yep, looks like it's fixed in current master then. Thanks!

PR #4643, opened 2016-06-12
Improve documentation to clarify min/max zoom

### LINKED ISSUE #4034: Autoscaling behavior when zooming past a layer's maxZoom
I'm not sure if this is a bug or a misinterpreation of the documentation.

I'm working with a tile source whose maximum zoom level is 18.  We want the user to be able to zoom in up to level 21, so I assumed that we'd just need to set the map to `maxZoom: 21` and the layer to `maxNativeZoom: 18`, and the layer would autoscale the tiles as necessary for zoom levels higher than 18.  

However, when I set just these two properties, the layer failed to render any tiles for zoom levels higher than 18.  I found that I also had to set the tile layer to `maxZoom: 21` to get it to autoscale the tiles as expected.  

If this is the intended behavior, I think the documentation should state that the layer's maxZoom property needs to be explicitly set, otherwise it won't render any tiles for zoom levels higher than the default (18).

Here's sample code that demonstrates the issue:

``` javascript
var map = L.map('map', {
    center: [51.505, -0.09],
    zoom: 18,
    maxZoom: 21
});

L.tileLayer('https://{s}.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}', {
    subdomains: ['server', 'services'],
    maxNativeZoom: 18,

    // this MUST be set in order for the tiles to be autoscaled at zoom levels > 18
    maxZoom: 21
}).addTo(map);
```

rheh:
Fixes #4034


rheh:
I think I've added the appropriate comments as discussed in the ticket; however, I may have interpreted the comment incorrectly.  So a review would be appreciated.

src/map/Map.js
```diff
@@ -35,11 +35,11 @@ L.Map = L.Evented.extend({
 		zoom: undefined,
 
 		// @option minZoom: Number = undefined
-		// Minimum zoom level of the map. Overrides any `minZoom` option set on map layers.
+		// Minimum zoom level of the map. Any `minZoom` set on map layers overrides this value.
```
IvanSanchez: I think the right thing to do would be to change this into «`minZoom: Number = *`» and «Minimum zoom level of the map. If not specified and at least one `GridLayer` or `TileLayer` is in the map, the lowest of their `minZoom` options will be used instead.»

src/layer/tile/GridLayer.js
```diff
@@ -102,11 +102,11 @@ L.GridLayer = L.Layer.extend({
 		bounds: null,
 
 		// @option minZoom: Number = 0
-		// The minimum zoom level that tiles will be loaded at. By default the entire map.
+		// The minimum zoom level that tiles will be loaded at. By default the entire map.  This will override the [minimum zoom level](#map-minzoom) set on the map.
```
IvanSanchez: Would be better with something like «If the map's [`minZoom`](#map-minzoom) option is set to a lower value than this, then this layer will not be displayed when zooming fully out.»

src/layer/tile/TileLayer.js
```diff
@@ -35,11 +35,11 @@ L.TileLayer = L.GridLayer.extend({
 	// @aka TileLayer options
 	options: {
 		// @option minZoom: Number = 0
-		// Minimum zoom number.
+		// Minimum zoom number. This will override the [minimum zoom level](#map-minzoom) set on the map.
```
IvanSanchez: Copy-paste from the longer description from `GridLayer`. The reason this is here is because Leafdoc cannot reset the default value of an inherited option while keeping the description.

src/layer/tile/GridLayer.js
rheh: Added as suggested

src/layer/tile/TileLayer.js
rheh: Done thanks

src/layer/tile/GridLayer.js
```diff
@@ -102,11 +102,11 @@ L.GridLayer = L.Layer.extend({
 		bounds: null,
 
 		// @option minZoom: Number = 0
-		// The minimum zoom level that tiles will be loaded at. By default the entire map.  This will override the [minimum zoom level](#map-minzoom) set on the map.
+		// The minimum zoom level that tiles will be loaded at. By default the entire map. If the map's [minimum zoom level](#map-minzoom) option is set to a lower value than this, then this layer will not be displayed when zooming fully out.
 		minZoom: 0,
 
 		// @option maxZoom: Number = undefined
-		// The maximum zoom level that tiles will be loaded at.  This will override the [maximum zoom level](#map-maxzoom) set on the map.
+		// The maximum zoom level that tiles will be loaded at. By default the entire map. If the map's [maximum zoom level](#map-maxzoom) option is set to a hhigher value than this, then this layer will not be displayed when zooming fully in.
```
perliedman: Spelling error: "hhigher".

rheh: Whoops - fixed now, thanks.

mourner:
Please remove the committed `reference.html` build.

PR #4645, opened 2016-06-14
Only set the original popup offset once

peergynt:
It doesn't seem to make sense to set `_originalPopupOffset` more than once.

One way to illustrate the issue is to do a quick change in reuse_popups.html:
```
var marker=... 
var icon = new L.Icon.Default();
map.on('zoomend', function(event) {
	marker.setIcon(icon);
});
```

After opening reuse_popups.html, click on the bottom marker. The popup looks good. Close the popup.
Zoom out one level, click on the bottom marker again, the popup offset is now incorrect (twice the custom offset).
This is due to `setIcon` calling `bind_popup` after `_popup.options.offset` was already changed in `openPopup`.

### DIFF
--- a/src/layer/Layer.Popup.js
+++ b/src/layer/Layer.Popup.js
@@ -42,9 +42,10 @@ L.Layer.include({
 			this._popupHandlersAdded = true;
 		}
 
-		// save the originally passed offset
-		this._originalPopupOffset = this._popup.options.offset;
-
+		if (!this._originalPopupOffset) {
+			// save the originally passed offset
+			this._originalPopupOffset = this._popup.options.offset;
+		}
 		return this;
 	},
 


yohanboniface:
I agree with the issue, but not with the fix.
Firstly, `this._popup.options.offset` can be changed when `options` is passed to `bindPopup`, so at least should be taken into account.
More important: I think the issue is not `this._originalPopupOffset` being reset, but `popup.options.offset` being altered later in the code. That would need a bit of investigation.

yohanboniface:
> That would need a bit of investigation.

You mentioned that already in your description, sorry. That's `openPopup` changing `popup.options.offset`. I think this is a wrong pattern that should be fixed.

yohanboniface:
See #4662 for an alternative fix proposal.

PR #4648, opened 2016-06-15
Added tileRecycleDelay option to GridLayer

fnicollet:
Commit related to issue https://github.com/Leaflet/Leaflet/issues/4039
Tile is retained if it has been loaded in the last X seconds, which is configurable via the tileRecycleDelay option on any TileLayer/GridLayer (0 = disabled = default).


IvanSanchez:
Superseded by #4650

PR #4649, opened 2016-06-16
Fix #4039: Implement FIFO MRU list of tiles before pruning them

### LINKED ISSUE #4039: Tile pruning too agressive?

IvanSanchez:
Implements a non-strictly-First-In-First-Out More-Recently-Used list of off-screen tiles, preventing them to be pruned.

In other words: when a tile goes off-screen, it is not immediately pruned. Instead, it goes into a most-recently-gone-offscreen tiles. When a tile is ejected from that list (due to another tile going into it), then it will get pruned.

As title says, fixes #4039.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -120,7 +120,11 @@ L.GridLayer = L.Layer.extend({
 
 		// @option className: String = ''
 		// A custom class name to assign to the tile layer. Empty by default.
-		className: ''
+		className: '',
+
+		// @option maxOffScreenTiles: Number = 10
+		// How many tiles to keep loaded when they move off the screen.
+		maxOffScreenTiles: 10
 	},
 
 	initialize: function (options) {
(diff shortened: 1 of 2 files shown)


src/layer/tile/GridLayer.js
```diff
@@ -120,7 +120,11 @@ L.GridLayer = L.Layer.extend({
 
 		// @option className: String = ''
 		// A custom class name to assign to the tile layer. Empty by default.
-		className: ''
+		className: '',
+
+		// @option maxOffScreenTiles: Number = 10
+		// How many tiles to keep loaded when they move off the screen.
+		maxOffScreenTiles: 10
```
yohanboniface: I'd go for moar as default: 24, maybe 48?

IvanSanchez:
Superseded by #4650

PR #4650, opened 2016-06-16
Fix #4039: Use different L.Bounds for "marking as prunable" and loading tiles

### LINKED ISSUE #4039: Tile pruning too agressive?

IvanSanchez:
Another alternative for #4039. This one hooks up with the code that calculates the `tileRange` for loading tiles, and calculates another bigger `L.Bounds`. Tiles within those bounds will never be flagged for pruning.

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -679,4 +679,142 @@ describe('GridLayer', function () {
 
 	});
 
+	describe("configurable tile pruning", function () {
+		var clock, grid, counts;
+
+		beforeEach(function () {
+			clock = sinon.useFakeTimers();
+
+			grid = L.gridLayer({
+				attribution: 'Grid Layer',
+				tileSize: L.point(256, 256)
+			});
+
+			grid.createTile = function (coords) {
+				var tile = document.createElement('div');
+				tile.innerHTML = [coords.x, coords.y, coords.z].join(', ');
+				tile.style.border = '2px solid red';
+				return tile;
+			};
+
+			counts = {
+				tileload: 0,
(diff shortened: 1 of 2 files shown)


mourner:
Looks good!

hyperknot:
Did 3 tests for various scenarios, however because it is working well we are not triggering any grid.on('load') events so it doesn't trigger the last bit.
https://github.com/Leaflet/Leaflet/pull/4650/commits/79d203bee2132bec849b315e4c2997bbbacbc235#diff-19a7d53df5d545ee8cde69a53256057cR801

hyperknot:
Fixed the test using spy(). Also run the test on master and it fails "Loads map, moves forth and back by 512 px, default keepBuffer" with "expected 12 to equal 0" to confirm it works.

IvanSanchez:
Tests look good to me :-)

pke:
When will this be published on NPM?

yohanboniface:
With 1.0.0-rc2, but not date planned atm.

IvanSanchez:
It will be done when it's done. https://github.com/Leaflet/Leaflet/milestones/1.0-rc2

pke:
Ok this is now in the npm version. Thanks! The experience has improved.
However, an initial pan still leaves a lot of white space to each side. Is there currently a property that can be tweaked to pre-load 1 row on each side and diagonal?

PR #4651, opened 2016-06-16
SVG icon class

iatkin:
Hi. I've created an SVG based icon class and would like to add it to the official plugin list. I'm not sure how the Marker list is order so I hope I've included it in the right place.


yohanboniface:
Thanks!

PR #4662, opened 2016-06-23
Do not alter popup.options.offset (fix #4645)

yohanboniface:
Try at solving #4645 at the root.


peergynt:
Agreed that this is a much better fix!
Modifying popup.options.offset was definitely ugly.
I added PR #4665 to remove the code that is now unused.

yohanboniface:
Thanks, I've pulled in your commit :)

PR #4664, opened 2016-06-23
specify license on home page

brunob:
Not sure if the formulation is the best, but i think it's important to specify it on the home page as it was done before f3762cff39ed1739640c063c07ad4da816fc7182

Any thoughts about it ?

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 layout: v2
 ---
 
-<p>Leaflet is the leading open-source JavaScript library for mobile-friendly interactive maps.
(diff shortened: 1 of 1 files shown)


mourner:
Most people who read about the library don't need to know the licensing details right away. If they are important, they can immediately look it up in the repository. Thanks for the suggestion, but I'd like to keep the introductory sentence as simple as posslbe.

brunob:
You're right @mourner :) plus i've just discovered that the info is available in https://github.com/Leaflet/Leaflet/blob/master/FAQ.md#commercial-use-and-licensing

PR #4669, opened 2016-06-24
Do not resetView on GridLayer.moveEnd (fix #4617)

### LINKED ISSUE #4617: Gridlayer's outer edge snapping to vertical center of map
How to reproduce:
* Have a tall map container
* Init a map with a low zoom level (0, 1)
* Drag the map up/down
* When the center of the map is a point outside the [-85, 85] latitude range, the tiles will snap out of alignment

Any other non-gridlayer layers will be positioned properly. This is also reproducible at any other zoom level, if dragging the map near the polar regions.

Playground here: http://playground-leaflet.rhcloud.com/febi/edit?html,output

yohanboniface:
Try at fixing #4617.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -574,7 +574,7 @@ L.GridLayer = L.Layer.extend({
 	_onMoveEnd: function () {
 		if (!this._map || this._map._animatingZoom) { return; }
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Seems to fix it with no ill effects.

PR #4671, opened 2016-06-25
Added Leaflet-copyright-loader.

J-F-O:


mourner:
Just wondering, does HERE allow its tiles being used in Leaflet, per terms of use?

J-F-O:
Yes, you can use their map tiles with leaflet, using their free basic plan (100.000 "transactions" per month), or payed plan for more (as is today). But you do need to register to get your app id and app code.

https://developer.here.com/news/201602172#.V3EWLO10DeQ

IvanSanchez:
I'd rather provide a subclass of `L.TileLayer` to display the HERE maps, and include the attribution there, as done in https://github.com/gmaclennan/leaflet-bing-layer. Right now, the name of this plugin is misleading.

J-F-O:
@IvanSanchez

OK. Do it your way.

IvanSanchez:
@J-F-O [OK then](https://github.com/IvanSanchez/Leaflet.TileLayer.HERE).

J-F-O:
@IvanSanchez
That was fast, it seems hybrid.day gives: "providers[i].boxes undefined"
Most likely due to the "default copyright" have no boxes.

IvanSanchez:
Superseded by https://github.com/Leaflet/Leaflet/commit/810da921f2c72b57b51a11e7fd66352661bdc028 and https://github.com/Leaflet/Leaflet/commit/d0538aa29db19f71bd5a14be33b421dc7fc9c76a

J-F-O:
It's an outrage!

J-F-O:
@IvanSanchez 

Being that there can be only one, at least make it work:

https://ivansanchez.gitlab.io/Leaflet.TileLayer.HERE/demo.html

There is the following issues, at today's date:
1. For hybrid map, it should be the "default copyright(s)" and the boxes copyright as well.
2. Zooming out on the normal map seem to give all copyrights, they should filter out and clear on zoom level, there are numerous unnecessary duplicate copyrights as well.
3. And on Firefox (fedora distro) changing the map, from hybrid back to normal breaks the copyright, only the leaflet prefix shows.

If you need the help you can just look at my code, that should give you an idea about the copyright part. So you don't violate the terms.

You are welcome.

PR #4690, opened 2016-07-01
Use pointer-events: visiblePainted as fallback for IE <11

### LINKED ISSUE #4674: Mouse Events Regression for IE10 in 1.0-rc1
I am unable to get any GeoJSON layer-specific mouse events to fire in IE10 in 1.0-rc1. These work correctly in 0.7.7. This works fine in IE11, Edge, and the latest Chrome and Firefox.

When I use the debugger to capture the mouse event, what I find is in `_handleDOMEvent`,` e.target` is referencing the tile layer instead of the GeoJSON layer that is on top of the tiles. The result is that no mouse events reach the GeoJSON layers.

Perhaps there is a workaround or a different way to handle events that fixes this. (Tested against the modern.ie Win7/IE10 VM)

1.0-rc1: (Broken in IE10)
https://jsfiddle.net/nitehawk123/p8xqm2sr/19/

0.7.7: (Working in IE10)
https://jsfiddle.net/nitehawk123/xtqrkc3q/4/

Here is an alternate way to handle the mouse events that results in the same problem:
https://jsfiddle.net/nitehawk123/9o04nxh2/

perliedman:
Close #4674.

I've verified this fix in a IE10 VM

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -95,6 +95,7 @@
 .leaflet-control {
 	position: relative;
 	z-index: 800;
+	pointer-events: visiblePainted; /* IE 9-10 doesn't have auto */
 	pointer-events: auto;
 	}
 .leaflet-top,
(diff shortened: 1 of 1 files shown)


nitehawk42:
I've tried this as a patch to 1.0.0-rc1, works perfect. Thanks!

PR #4691, opened 2016-07-01
Prevent from adding twice the same event listener without context

### LINKED ISSUE #4477: Events: listener without context can be added twice
```javascript
var m = new L.Marker();
var x = function() {console.log("event fired")};

/* register without context */
m.on("test", x);
m.on("test", x);

/* fires twice */
m.fire("test");

/* register with context */
context = {}
m.on("test2", x, context);
m.on("test2", x, context);

/* fires once*/
m.fire("test2");

```

https://playground-leaflet.rhcloud.com/wum/1/edit?html,js,output

yohanboniface:
Fix #4477 

Because we are storing events as objects `{fn: fn}`, and because of Javascript checking objects equality by reference, I see no elegant options, either:

- refactoring so we can store directly the `fn` in the array: needs more work, ends with `events` being inconsistent, probability of breaking something bigger
- store somewhere else the `fn` reference (for example is some `this._listeners_refs`): bad footprint, denormalized event storage, thus more fragile
- loop over existing registered listeners to see: bad performance in theory

Given that we are only talking about listeners registered without context or `this` as context, I think the list of already registered listeners will always be small, so I've chosen the last option.

If someone has a better pattern, I'm open! :)

Also, I've added a `console.warn` in the case we receive a listener already registered, trying to be smart with the developer. That's not in the usual Leaflet patterns, so I'll understand if we prefer remove it.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -115,6 +115,12 @@ L.Evented = L.Class.extend({
 			// so simple array makes the memory footprint better while not degrading performance
 
 			events[type] = events[type] || [];
+			for (var i = 0, len = events[type].length; i < len; i++) {
+				if (events[type][i].fn === fn) {
+					// This given callback has yet been registered.
+					return;
+				}
+			}
 			events[type].push({fn: fn});
 		}
 	},
--- a/debug/tests/event-perf.html
+++ b/debug/tests/event-perf.html
@@ -13,37 +13,80 @@
 <body>
   <script>
 
-  var obj = new L.Evented();
+  function benchmark(name, fn, setup) {
+    //warmup
+    for (var i = 0; i < 100; i++) {
+      if (setup) {
+        setup();
+      }
+      fn();
+    }
+
+    var t = 0;
+    for (var i = 0; i < 100; i++) {
+      if (setup) {
+        setup();
+      }
+      var t0 = performance.now();
+      fn();
+      t += performance.now() - t0;
+    }
+    console.log(name, (t / 100) + ' ms/run');
+  }
+
   var fn = function (e) {called += e.count;},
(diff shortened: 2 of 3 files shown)


yohanboniface:
Oh, there is another pattern:

- let the listeners be registered twice, but only fire them once. Will look into that to see if it comes nicer.

yohanboniface:
Humm, that would need also taking care of removing both the events. DRY alert.

src/core/Events.js
```diff
@@ -115,6 +115,13 @@ L.Evented = L.Class.extend({
 			// so simple array makes the memory footprint better while not degrading performance
 
 			events[type] = events[type] || [];
+			for (var i = 0, len = events[type].length; i < len; i++) {
```
perliedman: While this seems like the only way to do it without some larger refactor of events, I'm not to fond of this change.

As I understand it, we handle the events code like "performance critical", it contains some optimizations to be fast.

With this change, adding _n_ listeners to an object changes from an _O(n)_ operation to an _O(n^2)_ operation, which is something I assume we want to avoid.

The other option would be to handle contextless and `this`-scoped listeners the same way as listeners with context, but that would also have performance implications.

Do we have any tests or other means to evaluate the performance differences in practice?

yohanboniface:
>  let the listeners be registered twice, but only fire them once. Will look into that to see if it comes nicer.

Other counter-argument for such a scenario: it's certainly worst to affect `fire` performance, because by definition `fire` will be used more often than `on`.

mourner:
Yeah, it's a bad solution performance-wise. I think we need to go different route here.

yohanboniface:
Much better approach with #4697

PR #4694, opened 2016-07-01
Fix #4592 by falling back to a legacy ScrollEvent property on Edge.

### LINKED ISSUE #4592: Scrollwheel zoom too fast in MS Edge

IvanSanchez:
The `ScrollEvent`s in Edge are a pain in the a**. The `deltaY` amount depends on the window geometry (or the ratio between window height and window width), so it can have values as high as 130px or as low as 25, depending on how the user resizes a browser window.

The only way to work around this right now is to rely on the *legacy* `wheelDeltaY` event property, which Edge keeps at +/-120 per tick.

I have **not** tried Edge with a proper touchpad.

Furthermore, changed the values of the guessed deltas to be exactly 60 (this prevents FFX scrolling by 54px and others by 60, which causes problems with small `zoomSnap`s.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -235,16 +235,16 @@ L.DomEvent = {
 	// Gets normalized wheel delta from a mousewheel DOM event, in vertical
 	// pixels scrolled (negative if scrolling down).
 	// Events from pointing devices without precise scrolling are mapped to
-	// a best guess of between 50-60 pixels.
+	// a best guess of 60 pixels.
 	getWheelDelta: function (e) {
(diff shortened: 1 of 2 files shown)


mourner:
Let's merge and see how that goes :)

Issue #4410, opened 2016-04-11
Jumpy zooming with Magic Mouse


DennisSchiefer:
This is a follow-up to #2154 . As requested there, I am opening a new issue as the problem still persists.

Test Setup:
- Latest master (63870f1) together with the code of the quick start example
- MacBook Pro Retina with OSX Yosemite and Magic Mouse

Problems:
- Zooming with the Magic Mouse is very sensitive. It is almost impossible to zoom by only one zoom level
- When zooming for more than one zoom level (e.g. with a quick swipe), the animation looks jumpy as it moves to each zoom level instead of one smooth zoom from the initial to the final level before reloading all of the tiles

IvanSanchez:
@DennisSchiefer: Any chance you could debug what's going on in debug `L.DomEvent.getMouseDelta()` ? I'd be good to know if the pixel values are OK-ish, and whether the magic mouse sends several `wheel` events when doing a quick swipe, or just one.

DennisSchiefer:
How would I do this?
I assume, I would need to checkout Leaflet and build it myself, or can this be done just with the zip provided on the download page?

IvanSanchez:
@DennisSchiefer I whipped up a quick playground for this: http://playground-leaflet.rhcloud.com/lod/edit?html,console,output

That should be enough to spy on the values of your mouse events, so please let us know if you see something weird after experimenting with different mice/touchpads.

If you want to establish breakpoints, you might need to check out the source and set up a similar example in the `debug/` directory.

DennisSchiefer:
Thank you for setting up that playground.
I tested it with the Magic Mouse and the Apple Trackpad on the Mac and with a Microsoft IntelliMouse Explorer 3 on Windows.

Apple:
Even a slight swipe triggers 10 events, each with deltaY =1, wheelDelta = -3; a fast swipe triggers even more events with deltaY going up to 2 or 3 and wheelDelta respectively going to -6 or -9.

Microsoft:
Slight turn of the wheel triggered one event, fast turn triggered at most 4 events. Each event had a larger deltaY and wheelDelta value.

IvanSanchez:
> Even a slight swipe triggers 10 events, each with deltaY =1, wheelDelta = -3;

What's the timing of this? (Add some `console.log( performance.now() )` to see how many milliseconds between events). Also, is the `target` of the event always the map container?

[...]

PR #4695, opened 2016-07-01
Blind attempt at solving #4410 by not accumulating scroll deltas in Safari

IvanSanchez:
This is a blind attempt rather than something properly tested.

Tries to overcome the steep scrolling in desktop safari by throttling down the scroll delta to the maximum (absolute) value during a `wheelDebounceTime` rather than accumulating everything.

Needs testing by somebody with an actual magic mouse.


perliedman:
I tried this and could unfortunately not see much difference compared to original behaviour. It might feel a little different, but not a clear improvement.

perliedman:
Another downside with this fix is that it's Safari (desktop) only - if I understand @DennisShiefer's report correctly, this is actually a problem in any browser with Magic Mouse on OS X.

IvanSanchez:
Yeah, I don't think this change will fix the UX problem.

PR #4697, opened 2016-07-01
refactoring of events, it should be about 4 times faster now

perliedman:
This is an updated version of #2755, originally by @fab1an.

I have merged with current master, resolved conflicts and fixed code style issues.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -70,7 +70,7 @@ L.Evented = L.Class.extend({
 
 		if (!types) {
 			// clear all listeners if called without arguments
-			delete this._events;
+			this._events = undefined;
 
 		} else if (typeof types === 'object') {
 			for (var type in types) {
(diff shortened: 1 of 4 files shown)


src/core/Events.js
```diff
@@ -70,7 +70,7 @@ L.Evented = L.Class.extend({
 
 		if (!types) {
 			// clear all listeners if called without arguments
-			delete this._events;
+			this._events = undefined;
```
yohanboniface: Not sure we need that change.

src/core/Events.js
```diff
@@ -90,95 +90,97 @@ L.Evented = L.Class.extend({
 
 	// attach listener (without syntactic sugar now)
 	_on: function (type, fn, context) {
+		this._events = this._events || {};
+
+		/* get/init listeners for type */
+		var typeListeners = this._events[type];
```
yohanboniface: Maybe, a more Leaflet way to do that would be something like:

```
var typeListeners = this._events[type] = this._events[type] || {listeners: {}, count: 0};
```

src/core/Events.js
```diff
@@ -188,26 +190,29 @@ L.Evented = L.Class.extend({
...
+			for (var contextId in typeListeners.listeners) {
+				var listeners = typeListeners.listeners[contextId];
 
-				for (i = 0, len = listeners.length; i < len; i++) {
-					listeners[i].fn.call(this, event);
+				// each fn in context
+				for (var i = 0, len = listeners.length; i < len; i++) {
+					var l = listeners[i];
+					if (l.ctx) {
+						l.fn.call(l.ctx, event);
```
yohanboniface: `l.fn.call(l.ctx || this, event);`?

src/core/Events.js
```diff
@@ -221,9 +226,8 @@ L.Evented = L.Class.extend({
 	// @method listens(type: String): Boolean
 	// Returns `true` if a particular event type has any listeners attached to it.
 	listens: function (type, propagate) {
-		var events = this._events;
-
-		if (events && (events[type] || events[type + '_len'])) { return true; }
+		var typeListeners = this._events && this._events[type];
+		if (typeListeners && typeListeners.count) { return true; }
```
yohanboniface: `if (this._events && this._events[type] && this._events[type].count) { return true; }` ?

src/core/Events.js
```diff
@@ -90,95 +90,97 @@ L.Evented = L.Class.extend({
...
+			typeListeners.listeners[contextId] = listeners;
+		}
 
-				// keep track of the number of keys in the index to quickly check if it's empty
-				events[indexLenKey] = (events[indexLenKey] || 0) + 1;
+		// check if fn already there
+		var found = false;
+		for (var i = 0, len = listeners.length; i < len; i++) {
+			if (listeners[i].fn === fn) {
+				found = true;
```
yohanboniface: We may return here and remove `found` check.

src/core/Events.js
```diff
@@ -188,26 +190,29 @@ L.Evented = L.Class.extend({
...
-				listeners = events[type].slice();
+			// each context
+			for (var contextId in typeListeners.listeners) {
+				var listeners = typeListeners.listeners[contextId];
 
-				for (i = 0, len = listeners.length; i < len; i++) {
-					listeners[i].fn.call(this, event);
+				// each fn in context
+				for (var i = 0, len = listeners.length; i < len; i++) {
+					var l = listeners[i];
```
yohanboniface: Not sure we need `l` either actually.

src/core/Events.js
fab1an: Mistake on my part: https://github.com/Leaflet/Leaflet/pull/2755#discussion-diff-15043284

src/core/Events.js
fab1an: I splitted this up after doing benchmarking, where more lookups are done when using the concise syntax. It's a micro-optimisation though, might not be true for modern browsers anymore.

fab1an:
Thanks to both of you.

mourner:
How does performance look in your benchmarks?

mourner:
Let's follow along in #4709 which includes the changes from this PR.

fab1an:
Should I close my original PR: #2755  as well?

perliedman:
Re-opening since [performance of these changes seem better than #4709](https://github.com/Leaflet/Leaflet/pull/4709#issuecomment-231302559).

mourner:
Let's rebase this on master (cleaning up history with fixups), and then I'm up for merging.

perliedman:
@mourner I rebased on `master`, but did you also mean squash commits?

I've often squashed commits when merging the PR, from GitHub's interface, but I can of course squash on this branch if you prefer it.

perliedman:
@mourner Ok, rebased and squashed it into two commits.

fab1an:
Awesome, I'm happy to see this merged, thanks guys!

PR #4698, opened 2016-07-01
Fix #4685 by always reposition popups even when hidden

### LINKED ISSUE #4685: Popup is misplaced when setting zoomAnimation to false in PopupOption
Hello,

Here is an issue that can be reproduced in master and RC1. Add a marker, bind a popup with zoomAnimation option as false and zoom in or out. The popup is misplaced:
https://playground-leaflet.rhcloud.com/bemi/1/edit?html,output

IvanSanchez:


src/layer/Popup.js
```diff
@@ -360,17 +360,11 @@ L.Popup = L.Layer.extend({
...
-		    anchor = this._getAnchor();
-
-		if (this._zoomAnimated) {
-			L.DomUtil.setPosition(this._container, pos.add(anchor));
-		} else {
-			offset = offset.add(pos).add(anchor);
-		}
-
-		var bottom = this._containerBottom = -offset.y,
+		    anchor = this._getAnchor(),
```
yohanboniface: I'd say invert `offset` and `anchor` declarations, then do:
```
offset = L.point(this.options.offset)._add(anchor),
```

yohanboniface: Erf, meant to comment on #4699 instead :p

johnd0e:
#4685 is already fixed.
What are this PR waiting for?

Falke-Design:
I close this because it is already resolved by #4841

PR #4699, opened 2016-07-01
Removed zoomAnimation option

fnicollet:
Removed zoomAnimation option in Popup
Fixes https://github.com/Leaflet/Leaflet/issues/4685

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -92,11 +92,6 @@ L.Popup = L.Layer.extend({
 		// the Map's [closePopupOnClick](#map-closepopuponclick) option).
 		autoClose: true,
 
-		// @option zoomAnimation: Boolean = true
-		// Whether to animate the popup on zoom. Disable it if you have
-		// problems with Flash content inside popups.
-		zoomAnimation: true,
-
 		// @option className: String = ''
 		// A custom CSS class name to assign to the popup.
 		className: '',
@@ -113,8 +108,6 @@ L.Popup = L.Layer.extend({
 	},
 
 	onAdd: function (map) {
-		this._zoomAnimated = this._zoomAnimated && this.options.zoomAnimation;
-
 		if (!this._container) {
 			this._initLayout();
 		}
@@ -237,12 +230,10 @@ L.Popup = L.Layer.extend({
 	getEvents: function () {
 		var events = {
 			zoom: this._updatePosition,
-			viewreset: this._updatePosition
+			viewreset: this._updatePosition,
+			zoomanim: this._animateZoom
 		};
 
-		if (this._zoomAnimated) {
-			events.zoomanim = this._animateZoom;
-		}
 		if ('closeOnClick' in this.options ? this.options.closeOnClick : this._map.options.closePopupOnClick) {
 			events.preclick = this._close;
 		}
@@ -286,7 +277,7 @@ L.Popup = L.Layer.extend({
 		var prefix = 'leaflet-popup',
 		    container = this._container = L.DomUtil.create('div',
 			prefix + ' ' + (this.options.className || '') +
-			' leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide'));
+			' leaflet-zoom-animated');
 
 		if (this.options.closeButton) {
 			var closeButton = this._closeButton = L.DomUtil.create('a', prefix + '-close-button', container);
@@ -362,11 +353,7 @@ L.Popup = L.Layer.extend({
 		    offset = L.point(this.options.offset),
 		    anchor = this._getAnchor();
 
-		if (this._zoomAnimated) {
-			L.DomUtil.setPosition(this._container, pos.add(anchor));
-		} else {
-			offset = offset.add(pos).add(anchor);
-		}
+		L.DomUtil.setPosition(this._container, pos.add(anchor));
 
 		var bottom = this._containerBottom = -offset.y,
 		    left = this._containerLeft = -Math.round(this._containerWidth / 2) + offset.x;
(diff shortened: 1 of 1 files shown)


IvanSanchez:
@fnicollet The unit tests are failing with the same symptoms I (and @yohanboniface) noticed today - we might need to check the popup offset manually before merging.

fnicollet:
Yes, I would say that this commit is the source of the unit test failing:
https://github.com/Leaflet/Leaflet/commit/bddfb1168f5ea859181857b5764ba1f16100ddd2

fnicollet:
Ok, I think I am beginning to understand. Before my PR, the tests fail when zoomAnimated is true, but succeed when zoomAnimated is false. The test is not just one the Popup's zoomAnimation : 
```js
this._zoomAnimated = this._zoomAnimated && this.options.zoomAnimation;
```
but also on "this._zoomAnimated", which comes from Layer.js:
https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L86

So it takes in account map's zoomAnimation. So my guess is that when running the tests in phantom, the map is set as not animated, probably this test is failing in that env:
https://github.com/Leaflet/Leaflet/blob/master/src/map/anim/Map.ZoomAnimation.js#L18

So basically, the tests have been written for the zoomAnimation=false edge case and not the zoomAnimation=true default case.

So I would say, let's fix the tests instead?

fnicollet:
Should this test be removed? 

It tests for marker1._popup._container.style.left, which now always depends on the popup's width/height but doesn't give any indication of the popup's position relative to its anchor
Because we are now in the zoomAnimated case, the popup's position is changed via "L.DomUtil.setPosition", which changes the CSS transform property and not the left/bottom properties.

So I would ditch this test. Tell me if you want me to add it to the PR

yohanboniface:
@fnicollet popup positioning would certainly deserve a bit of cleaning, but AFAIK even with your commit we are setting the bottom and left manually. So this test should pass IMHO.
Certainly, even when in zoomAnimated, we should take care of updating the offset with the anchor.

src/layer/Popup.js
```diff
@@ -362,11 +353,7 @@ L.Popup = L.Layer.extend({
 		    offset = L.point(this.options.offset),
 		    anchor = this._getAnchor();
```
yohanboniface: This should fix the failing test:
```
		    anchor = this._getAnchor(),
		    offset = L.point(this.options.offset).add(anchor),
````

Comment (yohanboniface) [inline src/layer/Popup.js]: Actually, `_add` instead of `add` to prevent a useless Point cloning?

**Comment by yohanboniface** [comment]
Though I think we may have an issue when `L.Browser.any3d` is `false`
.

**Comment by fnicollet** [comment]
Here is what I understood from the code:
This specific Spec tests the facts that popupAnchor is well interpreted. It does so by comparing the position of a popup with and without a "popupAnchor" on the icon.
- "popupAnchor" has an effect on _getAnchor, which drives the CSS transform of the container
- "offset" has no relationship with popupAnchor, it just affects the bottom/left properties of the container.

Both combined achieve the final placement of the popup. So if the test really wants to test popupAnchor, it should be changed to a test which account for both (using element.offsetLeft or something?) or removed

**Comment by fnicollet** [comment]
Just saw your other comment, there will indeed be an issue if L.Browser.any3d is false as setPosition will change top left, and then it will set left/bottom right after and override what's been done in setPosition.

File: src/layer/Popup.js [hunk 2]
```diff
@@ -286,7 +277,7 @@ L.Popup = L.Layer.extend({
 		var prefix = 'leaflet-popup',
 		    container = this._container = L.DomUtil.create('div',
 			prefix + ' ' + (this.options.className || '') +
-			' leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide'));
```
yohanboniface: I think we want to keep that.
Even if the Popup itself does not have a `zoomAnimation` option, we want to listen to the animation capabilities of the browser and hide the popup when no animation will be made.

src/layer/Popup.js
```diff
@@ -237,12 +230,10 @@ L.Popup = L.Layer.extend({
 	getEvents: function () {
 		var events = {
 			zoom: this._updatePosition,
-			viewreset: this._updatePosition
+			viewreset: this._updatePosition,
+			zoomanim: this._animateZoom
 		};
 
-		if (this._zoomAnimated) {
```
yohanboniface: I think we still need the test, for same reasons than below :)

yohanboniface: Note: keep in mind that `this._zoomAminated` comes from `L.Layer`.

fnicollet: ok, so if we keep this._zoomAnimated from L.Layer and remove the option
from Popup, we still need to fix the updatePosition test in the case where
zoomAnimated is false. Couldn't figure out the right calculation yet

On Mon, Jul 4, 2016 at 9:23 PM, Yohan Boniface <notifications@github.com>
wrote:

> In src/layer/Popup.js
> <https://github.com/Leaflet/Leaflet/pull/4699#discussion_r69486833>:
>
> >  		};
> >
> > -		if (this._zoomAnimated) {
>
> Note: keep in mind that this._zoomAminated comes from L.Layer.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/Leaflet/Leaflet/pull/4699/files/3df8411b7998221d5aa171a0c1343b927924d46b#r69486833>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABLtode2pe0gFPgQYn2iRilqM2_ru3_7ks5qSV2tgaJpZM4JDR9->
> .
>



-- 
Fabien Nicollet
Tel : +33 (0)6 60 54 72 20

IvanSanchez:
Unable to merge due to conflicts with the new tooltip/divoverlay code. Superseded by #4685

PR #4705, opened 2016-07-02
Allow to make screenshots in Phantomjs

yohanboniface:
For that, one needs to run this call:

    window.top.callPhantom({'render': 'screenshot.png'});

This PR also make PhantomJS load Leaflet CSS while running tests (separate commit).

Example of screenshot:

screenshot

Should I add a helper to make the screenshot command easier to remember (something like `takeScreenshot(optionalPath)` ?

### DIFF
--- a/build/build.js
+++ b/build/build.js
@@ -193,7 +193,7 @@ exports.test = function(complete, fail) {
 	var karma = require('karma'),
 	    testConfig = {configFile : __dirname + '/../spec/karma.conf.js'};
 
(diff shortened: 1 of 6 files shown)


nathancahill:
Bravo!

perliedman:
Looks good!

I'd vote for adding the shortcut function you mention above!

yohanboniface:
Added a helper `takeScreenshot` that takes an optional path argument (defaulting to `screenshot.png`).

PR #4708, opened 2016-07-04
Make L.Handler.enable/disable return this

yohanboniface:
Those are methods used very often, this can save us a lot of
useless lines.

```
this.handling.enable().doSomething();
```
instead of
```
this.handling.enable()
this.handling.doSomething()
```

Just like we have for `addTo` and such.


IvanSanchez:
@yohanboniface Docstrings should reflect the return value too.

yohanboniface:
Good catch :)
Fixed!

PR #4709, opened 2016-07-05
Refactor events futher for even more performance (continuing #2755 and #4697)

perliedman:
This adds bookkeeping of added listener functions to avoid looping over listeners to check existence.

In our synthetic tests (`debug/tests/event-perf.html`) it shows almost the same performance as current `master` for `on` and `off`, but keeps the more than double performance for `fire` from #4697.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -103,8 +103,8 @@ L.Evented = L.Class.extend({
 		}
 
 		var contextId = context && context !== this && L.stamp(context),
-			newListener = {fn: fn, ctx: context};
-			
+		    newListener = {fn: fn, ctx: context};
+
 		if (!contextId) {
 			contextId = 'no_context';
 			newListener.ctx = undefined;
@@ -119,13 +119,13 @@ L.Evented = L.Class.extend({
 
 		// check if fn already there
 		var found = false;
-		for (var i=0, len=listeners.length; i<len; i++) {
+		for (var i = 0, len = listeners.length; i < len; i++) {
 			if (listeners[i].fn === fn) {
 				found = true;
 				break;
 			}
 		}
-		
+
 		if (!found) {
 			listeners.push(newListener);
 			typeListeners.count++;
@@ -134,7 +134,7 @@ L.Evented = L.Class.extend({
 
 	_off: function (type, fn, context) {
 		if (!this._events) { return; }
-		
+
 		if (!fn) {
 			// clear all listeners for a type if function isn't specified
 			delete this._events[type];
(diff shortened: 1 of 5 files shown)


yohanboniface:
Have you monitored memory print roughly?

perliedman:
@yohanboniface gah, of course that one fails! But thanks, it would be a great failure not to test that now...

I'll try to have a look tomorrow.

src/core/Events.js
```diff
@@ -90,95 +90,116 @@ L.Evented = L.Class.extend({
...
+			fnIndices = {};
+			typeListeners.listeners[contextId] = listeners;
+			typeListeners.fnIndices[contextId] = fnIndices;
+		}
 
-			events[type] = events[type] || [];
-			events[type].push({fn: fn});
+		// check if fn already there
+		var fnStamp = L.stamp(fn);
+		if (fnIndices[fnStamp]) {
```
yohanboniface: If indice=0, this will be falsy ;)

src/core/Events.js
```diff
@@ -90,95 +90,116 @@ L.Evented = L.Class.extend({
...
+			if (listenerIndex >= 0) {
+				// find fn and remove it
+				var l = listeners[listenerIndex];
+				// set the removed listener to noop so that's not called if remove happens in fire
+				l.fn = L.Util.falseFn;
+				typeListeners.count--;
+
+				if (listeners.length > 1) {
+					if (!this._isFiring) {
+						listeners.splice(listenerIndex, 1);
```
yohanboniface: Humm, I see a sort of issue in altering the `listeners` array while relying on indices to retrieve the listeners.

perliedman: There's code handling this a few lines further down, modifying the relevant indices after a listener is removed. That code still relies on an annoying loop, though, and could potentially be slow depending on the order the listeners are removed in.

perliedman:
@yohanboniface I measured memory usage after modifying our perf test to add 10k listeners. Current master uses up 9.5MB on my machine, while with this change it uses roughly 10.5. So as expected memory usage is higher, but maybe not terrible.

For this maybe not very realistic scenario, firing events was 2.5x faster with the new code.

src/core/Events.js
perliedman: @yohanboniface also worth noting was that some tests broke before I added the code to fiddle with the indices, so I think we're good.

perliedman:
Looking at some more realistic performance test (`debug/tests/event-perf-2.html`, available from the #4697 PR), these changes seem to perform _worse_ than @fab1an's original code, while #4697 still out-performs both current `master` and the `on-twice` branch.

I'm closing this for now, and re-opening #4697.

fab1an:
This is probably to GC-pauses, which occur when creating so many objects?

fab1an:
Are you using Chrome's profiler or a different one?

perliedman:
@fab1an chrome's profiler! I'm by no means an expert with it, so I'm hoping I'm reading the numbers correctly.

fab1an:
Ok, then you could look at the flame chart view.

If a periodic operation takes longer than it's successors, you can zoom in, go down and see if the operation took longer because GC is called. I imagine this being the case because of the additional data-structures in your other branch.

PR #4710, opened 2016-07-06
Add L.Control.Layers.eachLayer

yohanboniface:
This should allow people extending the control to easily override
the order in which layers are processed (and thus added to the HTML).


IvanSanchez:
I'm in favour of #4711 (bias alert!), as overriding the `eachLayer()` method is not trivial for users. I can foresee users getting confused about the layers control failing completely due to a typo.

IvanSanchez:
Superseded by #4711.

PR #4711, opened 2016-07-06
Add a 'sortLayers' option to L.Control.Layers

IvanSanchez:
An alternative to #4710, this one with unit tests.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -55,12 +55,20 @@ L.Control.Layers = L.Control.extend({
 		// If `true`, the base layers in the control will be hidden when there is only one.
 		hideSingleBase: false,
 
-		// @option sortLayers : Function|Boolean = false
-		// Sort the layers with the given [compare function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/sort);
-		// this function may access user-defined properties of the layers.
-		// If not given, layers will keep the order in which they were added to
-		// the control.
-		sortLayers: false
+		// @option sortLayers: Boolean = false
+		// Whether to sort the layers. When `false`, layers will keep the order
+		// in which they were added to the control.
+		sortLayers: false,
+
+		// @option sortFunction: Function = *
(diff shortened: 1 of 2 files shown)


mourner:
Is there a way to sort by layer name with this PR?

IvanSanchez:
No - but it could be easily changed so the compare function receives the internal `{name: …, layer: …}` object.

src/control/Control.Layers.js
```diff
@@ -53,7 +53,14 @@ L.Control.Layers = L.Control.extend({
...
 		// If `true`, the base layers in the control will be hidden when there is only one.
-		hideSingleBase: false
+		hideSingleBase: false,
+
+		// @option sortLayers : Function|Boolean = false
+		// Sort the layers with the given [compare function](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array/sort);
+		// this function may access user-defined properties of the layers.
+		// If not given, layers will keep the order in which they were added to
+		// the control.
+		sortLayers: false
```
grischard: Should this simply be called `compareFunction` then?

yohanboniface:
Meeeh! ;)
I prefer #4710 because it does not limit the possibilities, while this option is specific to one usage (sorting). For example, one may want to filter out some layers, or amend, or whatever.

IvanSanchez:
Rebased, and addressed @mourner's concern about sorting by name. Also, split the on/off behaviour from the sorting function (inspired by @grischard), as to provide a sane default.

@grischard As I'm linking to the MDN docs about `sort()`, I don't think the implementation details about edge cases of equally-sorted values need to be explained in the Leaflet docs.

grischard:
I was glad when I asked about this this summer to see two PRs. This one suits me fine, and comes with a test case. I share @IvanSanchez's feelings on @yohanboniface's #4710 not making things easy, but could live with either. I would just be happy to have one of the two merged in.
 
Of course, I could pick one and run a local version. But then if Leaflet goes for the other, I have to re-do the work.

So I'd like to very humbly and politely ask for #4711 to be merged

PR #4712, opened 2016-07-06
Add Leaflet.Permalink to plugins.md

MarcChasse:


jieter:
[repo](https://github.com/MarcChasse/leaflet.Permalink)
[demo](https://marcchasse.github.io/leaflet.Permalink/)

(I always like to click instead of copy urls)

MarcChasse:
I'm not sure how I would go about extending leaflet-hash.
The Hash plugin loads the default map and then loads the map at the position in the link while permalink only loads the map at the specified position and zoom, that and i've added a bit more precision to the lat/long so higher zoom links will work without moving the map.

For example:
Hash default (18 img requests): [http://mlevans.github.io/leaflet-hash/map.html](http://mlevans.github.io/leaflet-hash/map.html)
Hash link (34 img requests): [http://mlevans.github.io/leaflet-hash/map.html#12/37.4317/-121.7927](http://mlevans.github.io/leaflet-hash/map.html#12/37.4317/-121.7927)
vs
Permalink default (20 img request): [https://marcchasse.github.io/leaflet.Permalink/](https://marcchasse.github.io/leaflet.Permalink/)
Permalink link (20 img request): [https://marcchasse.github.io/leaflet.Permalink/#12.28933,-82.98162,15z](https://marcchasse.github.io/leaflet.Permalink/#12.28933,-82.98162,15z)

I'm also partial to the way permalink displays the link:
Permalink: `<map url here>#12.28933,-82.98162,15z`
vs
Hash: `<map url here>#15/12.28933/-82.98162`

jieter:
Sound like you thought about it, nice. I asked mainly because fragmentation is not desirable. If your script does it different or better, no problem!

IvanSanchez:
Not sure about this. `L.Permalink` does look nicer than `L.Hash`, and it has nice features such as browser history handling. But I'd rather see it as a map handler `L.Map.Permalink = L.Handler.extend(…)`, not as a top-level class. This way devs can `enable()` and `disable()` it.

jieter:
also note this pull request: https://github.com/mlevans/leaflet-hash/pull/21/files, which adds browser history to leaflet-hash. Looks like it's a bit abandoned though, maybe we can ask @mlevans to transfer the repo to the leaflet org?

MarcChasse:
Thanks for looking it over guys, I'm not opposed to making it a map handler i'd just need some direction in getting started on the implementation. Looks like i'll need to implement [add and remove hooks](http://leafletjs.com/reference-1.0.0.html#handler) would [this](https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.Drag.js) be a good example to copy/modify?

yohanboniface:
Yes, addHooks/removeHooks are the only required methods, that should do anything needed on enabling and on disabling.
Map.Drag.js is a good example, but a part from addHooks/removeHooks, the code style is up to you ;)

MarcChasse:
I could use some help making it a map handler, is there any other resource I could check? 
I've got a [partially functioning demo](https://marcchasse.github.io/leaflet.Permalink/extentMap.html) setup based off of my [modified js](https://github.com/MarcChasse/leaflet.Permalink/blob/L.Map.Extend/leaflet.permalink.js) but it fails to adjust the link on moveend.
Any help would be greatly appreciated. Thank you!

PR #4714, opened 2016-07-06
Add Leaflet.ScaleFactor to plugins.md

MarcChasse:
[repo](https://github.com/MarcChasse/leaflet.ScaleFactor)
[demo](https://marcchasse.github.io/leaflet.ScaleFactor/)

### DIFF
--- a/plugins.md
+++ b/plugins.md
@@ -2308,6 +2308,15 @@ Allow the user to measure distances or areas.
 			<a href="https://github.com/nerik">Erik Escoffier</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/MarcChasse/leaflet.ScaleFactor">Leaflet.ScaleFactor</a>
+		</td><td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Looks good, but have a look at the issues I reported.

Also, consider not having the minified files in the repo - they are hard to maintain in the long run. Or the UMD header, for that matter. Also consider a `npm` package.

MarcChasse:
Thanks for looking it over Ivan, i'll need to figure out how to make an npm package.
Also I Included the UMD header based on the [Leaflet plugin guide](https://github.com/Leaflet/Leaflet/blob/master/PLUGIN-GUIDE.md#module-loaders), but i'm not using it so I could remove it.

IvanSanchez:
Well, there are different schools of thought about the whole UMD / npm / jspm / bower hell.

What I normally do is have "clean" files for development, which are versioned in the repo, and then the browserified version with the UMD header, which isn't versioned in the repo but can be fetched through NPM.

PR #4727, opened 2016-07-14
Plugins added Leaflet Polyline Offset

ghybs:
from Benjamin Becquet (also author of several other listed plugins).
https://github.com/bbecquet/Leaflet.PolylineOffset

By the way, the [Bus Lines example](http://bbecquet.github.io/Leaflet.PolylineOffset/examples/example_bus.html) is a really nice use case!


IvanSanchez:
Wow, nice! :-O

Issue #4646, opened 2016-06-14
Invalid out-of-bounds tiles are requested from the server when noWrap and bounds are set


joanrieu:
**Symptom**

Invalid out-of-bounds tiles are requested from the server, resulting in 404 errors in the console and the default *image-couldn't-be-loaded* rectangles being displayed instead of no tiles.

--

**Trigger**

This problem happens when a layer has both the `noWrap` and the `bounds` options.

--

**Visual comparison**

The layers are defined with `noWrap: true` and `bounds: [[-90, -180], [90, 180]]`. Incorrect behavior on the left, fixed one on the right.

![Visual comparison](http://i.imgur.com/ZfKoug5.png)

--

**Bug**

The following function located inside `GridLayer` does not respect the `noWrap` option: `map.wrapLatLng(...)` is always called.

```javascript
_tileCoordsToBounds: function (coords) {

	// ...

	nw = map.wrapLatLng(map.unproject(nwPoint, coords.z)),
	se = map.wrapLatLng(map.unproject(sePoint, coords.z));

	return new L.LatLngBounds(nw, se);
},
```

--

**Fix**

If the `noWrap` option is set on the layer, `map.wrapLatLng` *should not* be called.

--

**Explanation**

The `_tileCoordsToBounds(...)` function is called by `_isValidTile(...)`, which relies on it to validate a tile by checking that it is in the user-defined bounds (if any), as seen below.

```javascript
_isValidTile: function (coords) {

	// ...

	if (!this.options.bounds) { return true; }

	// don't load tile if it doesn't intersect the bounds in options
	var tileBounds = this._tileCoordsToBounds(coords);
	return L.latLngBounds(this.options.bounds).overlaps(tileBounds);
},
```

--

**Workaround** (until a fix is released)

You can patch the `layer._tileCoordsToBounds` function on your layers, changing the offending call.

--

**Version**

I am facing this bug in version `v1.0.0-rc.1` and the few changes to this file in `master` are completely unrelated.

joanrieu:
Related issues I found while tracking down the bug:
- Leaflet tries to get negative value images or images that don't exist #2981
- noWrap not working mstahv/v-leaflet#103

fyeah:
Thank you for discovering all this. Still took me a long time to get to work. If anyone else needs a quick fix for this see #4738 
I hope it get's fixed soon.

kaynz:
Is this somehow related to #2390?

JoshQuake:
[...]

PR #4738, opened 2016-07-19
only wrap tileBounds if noWrap is false. #4646

fyeah:
#4646 Thanks for discovering the problem. I really needed this to work. This is what I did.


IvanSanchez:
Your fix looks sane. However:

* There is a spurious call to `unproject` if `noWrap` is set to `false`
* Tile URLs are critical enough to have a unit test. There should be a test to check that the scenario described in #4646 results in just one tile loaded.

Even if this is not a complete bug fix, it's a very good starting point for one!!

IvanSanchez:
Superseded by #4908

Issue #4736, opened 2016-07-17
Dragging a marker doesn't provide 'oldLatLng' in 'move' event.


Rantanen:
Moving marker with `setLatLng` yields `oldLatLng` value in the event parameter.

Moving the marker by dragging it around doesn't.

http://playground-leaflet.rhcloud.com/lifu/1/edit?html,output

Rantanen:
Also discussed at Leaflet/Leaflet.markercluster#335

mourner:
Fixed in #4752.

PR #4752, opened 2016-07-25
Provide oldLatLng during marker dragging

snkashis:
References #4736

### DIFF
--- a/src/layer/marker/Marker.Drag.js
+++ b/src/layer/marker/Marker.Drag.js
@@ -60,6 +60,8 @@ L.Handler.MarkerDrag = L.Handler.extend({
 
 		// @event movestart: Event
 		// Fired when the marker starts moving (because of dragging).
+
+		this._marker._oldLatLng = this._marker.getLatLng();
 		this._marker
 		    .closePopup()
 		    .fire('movestart')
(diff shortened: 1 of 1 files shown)


src/layer/marker/Marker.Drag.js
```diff
@@ -60,6 +60,8 @@ L.Handler.MarkerDrag = L.Handler.extend({
 
 		// @event movestart: Event
 		// Fired when the marker starts moving (because of dragging).
+
+		this._marker._oldLatLng = this._marker.getLatLng();
```
yohanboniface: I'd store the tmp variable on `this` instead of `this._marker`, in order to reduce the risk of a clash for L.Marker users.

snkashis: Yeah, I went back and forth on this quickly for initial commit. You're right. New commit added.

src/layer/marker/Marker.Drag.js
```diff
@@ -60,6 +60,8 @@ L.Handler.MarkerDrag = L.Handler.extend({
 
 		// @event movestart: Event
 		// Fired when the marker starts moving (because of dragging).
+
```
yohanboniface: nit: I'd keep the docstring comment close to the `fire`.

PR #4755, opened 2016-07-27
Run *.png through zopflipng -m

scop:
6 bytes trimmed, yay :)


IvanSanchez:
Not particularly great savings, but why not.

PR #4764, opened 2016-07-29
Fix option misnamed snapZoom instead of zoomSnap

louh:


IvanSanchez:
Thanks for noticing this!

Issue #4475, opened 2016-04-21
Events: execution order is theoretically not defined


fab1an:
Event code uses `for` to iterate over object properties, like seen here: https://github.com/Leaflet/Leaflet/blob/master/src/core/Events.js#L194

This should work, but browser vendors could change the iteration order anytime. Also numerical properties are treated differently apparently: http://stackoverflow.com/a/280861/342947

IvanSanchez:
What's a failing use case for this? Can't really see when this can become a problem.

fab1an:
It probably won't ever fail, as long as there are no event names which start with numbers.

It's technically not correct though to rely on Javascript property iteration order, because it's not defined. Chrome could start iterating properties from behind (which it won't do). If you want predictable iteration you are required to use arrays.

IvanSanchez:
@fab1an Now that #4697 has been merged into `master`, is this still an issue?

PR #4769, opened 2016-07-30
Make sure we call events listeners in the order they are registered (fix #4743)

### LINKED ISSUE #4743: Vector rendering regression in rc2
I've loaded a GeoJSON polygon onto the map. I click and drag the map so the polygon is either partially or totally obscured, and then release the mouse. Then I click and drag the map so the polygon should be visible again, and then release the mouse. At this point, the portion of the polygon that was not visible does not get redrawn, but rather the polygon is simply closed with a joining line (or in the case where the polygon was completely obscured, nothing gets drawn at all). Not until I nudge the map again does the polygon redraw correctly.

Changing the clip padding doesn't make a difference. It only extends how far I have to drag the polygon past the viewport. When I drag the polygon back and release the mouse, it still doesn't redraw correctly.

https://jsfiddle.net/Lomtbax3/

Chrome Version 51.0.2704.84 m
Windows 7 Enterprise SP1
Leaflet 1.0.0-rc2

This almost seems like a repeat of #2674, #2814, #2945, or #3128, except those all specifically mention not redrawing while still dragging, which I understand is for performance reasons. My issue is that it's not redrawing correctly after releasing the mouse.

yohanboniface:
Attempt to fix #4743.

This makes sure the event listeners are called in the order they are registered.
The main change is not to store the events by context anymore (this being a hash, thus having an impredictable loop order).
The drawback being that when adding a listener the cost of finding duplicates is a higher in the case the current feature as a lot of listeners for a given event type.

This is the output of the event-perf debug page for master:
screenshot from 2016-07-30 20-44-32

And the same for the current branch:
screenshot from 2016-07-30 20-44-14

For comparison, this is for RC1:
screenshot from 2016-07-30 20-49-33

So as we can see, the current branch perform a bit less than the rc2 when adding twice the same function with same context (and still worst than rc1 that does not try to find duplicates), but this much better than rc1 on `fire`.
And I think keeping the listeners order is not negotiable.
We should keep in mind that the event-perf debug page has he bias of being a bit synthetical and not real use cases.

One option could be that we do no check for duplicates when adding events as it was until rc2.

As a side result, the code is also more simple, and thus more readable, as it is.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -96,30 +96,22 @@ L.Evented = L.Class.extend({
 		var typeListeners = this._events[type];
 		if (!typeListeners) {
 			typeListeners = {
-				listeners: {},
+				listeners: [],
 				count: 0
 			};
 			this._events[type] = typeListeners;
 		}
 
-		var contextId = context && context !== this && L.stamp(context),
-		    newListener = {fn: fn, ctx: context};
-
-		if (!contextId) {
-			contextId = 'no_context';
-			newListener.ctx = undefined;
+		if (context === this) {
+			// Less memory footprint.
+			context = undefined;
 		}
-
-		// fn array for context
-		var listeners = typeListeners.listeners[contextId];
-		if (!listeners) {
-			listeners = [];
-			typeListeners.listeners[contextId] = listeners;
-		}
+		var newListener = {fn: fn, ctx: context},
+		    listeners = typeListeners.listeners;
 
 		// check if fn already there
 		for (var i = 0, len = listeners.length; i < len; i++) {
-			if (listeners[i].fn === fn) {
+			if (listeners[i].fn === fn && listeners[i].ctx === context) {
 				return;
 			}
 		}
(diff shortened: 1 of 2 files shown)


src/core/Events.js
```diff
@@ -96,102 +96,81 @@ L.Evented = L.Class.extend({
...
 		// check if fn already there
 		for (var i = 0, len = listeners.length; i < len; i++) {
-			if (listeners[i].fn === fn) {
+			if (listeners[i].fn === fn && listeners[i].ctx === context) {
 				return;
 			}
 		}
 
-		listeners.push(newListener);
+		typeListeners.listeners.push(newListener);
```
perliedman: Why not `listeners` here? It looks like `listeners === typeListeners.listeners` ?

yohanboniface: No good reason ;)
I removed `listeners` local variable at first, and then apparently forgot to put it back here.
Fixing it, thanks! :)

perliedman:
:+1: LGTM!

The performance regression looks acceptable given that we win totally predictable and easy to understand call order for listeners.

mourner:
Merging this, but I'm still concerned because benchmarks may not cover some common use cases that can substantially regress from this (because of the quadratic behavior). But let's keep master in a working condition and see if we can improve on performance after that.

fab1an:
This should fix: #4475 as well.

src/core/Events.js
```diff
@@ -96,30 +96,22 @@ L.Evented = L.Class.extend({
 		var typeListeners = this._events[type];
 		if (!typeListeners) {
 			typeListeners = {
-				listeners: {},
+				listeners: [],
 				count: 0
```
fab1an: I think this count variable can be removed and replaced by `listeners.length`. 

Actually I think the whole `typeListeners` object can be replaced by the `listeners` array now.

perliedman: @fab1an you're correct! We did away with `typeListeners` yesterday! https://github.com/Leaflet/Leaflet/commit/017d29c4ebfaee201b4df48b53ca14e037308aa8

PR #4773, opened 2016-07-31
Better tooltip offset handling (fix #4772)

### LINKED ISSUE #4772: Tooltip offset not moving the labels in the right direction?
Positive numbers for x-axis moves the label towards the left, while negative numbers moves it to the right. I'm not sure if this is how this is designed though, couldn't find much references for this.
But for y-axis, the label doesn't move at all in any direction.

Sample code:
<code>
var lc = L.latLng(51.505, -0.09);

var map = L.map(document.getElementById('mapid'), {
	renderer: L.canvas()
});

var osm = new L.TileLayer('http://{s}.tile.osm.org/{z}/{x}/{y}.png',
	{ attribution: 'Map data &copy; <a href="http://openstreetmap.org">OpenStreetMap</a>' }).addTo(map)

map.setView(lc, 14);

var featureGroup = new L.FeatureGroup()
featureGroup.addTo(map)

var c = L.circle([lc.lat, lc.lng], { radius: 500 })
c.addTo(featureGroup)
c.bindTooltip('test', { permanent: true, offset: L.point({x: -80, y: 80}) })
</code>

yohanboniface:
- default offset is now [0, 0]
- both offset.x and offset.y are always honoured
- positive x always move the tooltip on the right
- positive y always move the tooltip on the bottom

Fix #4772

### DIFF
--- a/debug/map/tooltip.html
+++ b/debug/map/tooltip.html
@@ -49,7 +49,7 @@
 			iconSize: [50, 50]
 		});
 		L.marker([41.22, 9.48], {icon: icon}).addTo(map).bindTooltip('A div icon tooltip following mouse', {sticky: true, direction: 'auto'});
-		L.marker([41.23, 9.47], {icon: icon}).addTo(map).bindTooltip('A div icon tooltip');
+		L.marker([41.23, 9.47], {icon: icon}).addTo(map).bindTooltip('A div icon tooltip with custom offset', {direction: 'top', offset: [-25, -25]});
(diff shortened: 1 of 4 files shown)


remilev:
Hi, sorry for interrupting, but : 
when running `map.on('popupopen', function(e) { console.log(e.popup.options.offset) });`

I get Object {x: 10, y: 20} with leaflet v1beta2 and Array [10, 20] with leaflet RC3
(options.y or x undefined with RC3)
Is that intended?

Issue #4461, opened 2016-04-19
Add recursive to L.LayerGroup.eachLayer method


stefanocudini:
It might be useful to add an new option to *L.LayerGroup.eachLayer()* method for recursive behavior?
for example:

```javascript
....geojson loading...

var geo1 = L.geoJson(Geo1);
var geo2 = L.geoJson(Geo2);
var geo3 = L.geoJson(Geo3);
var geo4 = L.geoJson(Geo4);

var group1 = L.LayerGroup([geo1, geo2]);
var group2 = L.LayerGroup([geo3, geo4]);
var groupAll = L.LayerGroup([group1, group2]);

groupAll.eachLayer(function(lay, layParent) {
   console.log(lay, layParent);
}, {
 recursive: true
})

```

danielefenix:
+1

fabriziofenix:
+1

IvanSanchez:
Sounds like a cool little feature, but I'd rather have a new function call for recursively traversing layer groups (maybe something like `getSingleLayers()` and `eachSingleLayer()`). It should be straightforward to implement in very few lines of code, so I'll encourage any newbies to submit a pull request. cc @yourfirstpr

stefanocudini:
*getSubLayers()* or *getAllLayers()* ?

perliedman:
How about `eachLayerRecursive`? It's not a get-method, since it will use a callback to visit each layer, so I'd keep that out of the name. Also, since it's more or less the same method as `eachLayer`, it would be nice if the name makes that clear.

stefanocudini:
+1

ghybs:
Hi,

It seems to me that `getSingleLayers()` / `eachSingleLayer()` convey a different meaning from `eachLayerRecursive()`:

- `getSingleLayers()` / `eachSingleLayer()` sound like acting only on non-group layers, which is interesting in probably most situations (see Leaflet/Leaflet.markercluster#624 for getting non-group layers, stefanocudini/leaflet-search#101 for applying a method on them). However, it might be ambiguous whether they just skip group layers or if they continue recursively on children of group layers? => `getSingleLayersRecursive()` and `eachSingleLayerRecursive()`?

- `eachLayerRecursive()` sounds like doing the same as `eachLayer()` at the beginning (i.e. applying the method to _child group_ layers as well), but continuing to children of those group layers. Hence it sounds like it does **more** than `eachSingleLayerRecursive()`.

[...]

PR #4776, opened 2016-08-01
Get / Each single layers recursively

ghybs:
Following discussion in #4461, added the 2 methods to Layer Group that seem the most useful:
- `getSingleLayersRecursive()`: extract all **non-group** (i.e. single) layers from this Layer Group and from all its child Layer Groups recursively. When the method encounters a Layer Group, it re-applies itself on it and appends the result in place in the resulting array, instead of the Layer Group. However the final layers order is not necessarily guaranteed as it loops through child layers according to the ascending order of `_leaflet_id`, not to insertion order. This could replace [Leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) internal `_extractNonGroupLayers()` method (if ready to loop twice).
- `eachSingleLayerRecursive()`: similar but applying a method and optionally a context. This should be directly usable by [leaflet-search](https://github.com/stefanocudini/leaflet-search) and [leaflet-pip](https://github.com/mapbox/leaflet-pip) to replace `eachLayer()`.

I added corresponding test cases.

I also took the liberty to add an extra method `toSingleLayersArray()` to the `L.Layer` base class. The idea is to provide a mean to easily convert whatever layer into an array of single layers, without having to re-implement a Layer Group detection and having to call `getSingleLayersRecursive()` in case it is a group.

For example, [Leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) could now implement its `addLayer()` method (for single layer) through:
```javascript
addLayer: function (layer) {
    return this.addLayers(layer.toSingleLayersArray());
}
```
(it currently checks the Layer Group condition itself and acts accordingly)

I added corresponding test cases in Layer Group and Marker specs.

### DIFF
--- a/spec/suites/layer/LayerGroupSpec.js
+++ b/spec/suites/layer/LayerGroupSpec.js
@@ -54,6 +54,25 @@
 		});
 	});
 
+	describe("#getSingleLayersRecursive", function () {
+		it('gets all non-group layers recursively', function () {
+			var lg1 = L.layerGroup(),
+			    marker1 = L.marker([0, 0]),
+			    lg2 = L.layerGroup(),
+			    marker2 = L.marker([0, 0]),
+			    lg3 = L.layerGroup(),
+			    marker3 = L.marker([0, 0]);
+
+			lg1.addLayer(marker1);
+			lg1.addLayer(lg2);
+			lg2.addLayer(marker2);
+			lg2.addLayer(lg3);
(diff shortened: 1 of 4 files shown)


ghybs:
Well on 2nd thought, the new `toSingleLayersArray()` method is pretty useless / not worth being added to `L.Layer` base class.

I am sure you will not like it.

I will open a new PR without that commit.

PR #4781, opened 2016-08-02
Get / Each single layers recursively (no addition to base class)

ghybs:
This PR is a more simple version of PR #4776.

For the record, here is the corresponding description:

Following discussion in #4461, added the 2 methods to Layer Group that seem the most useful:
- `getSingleLayersRecursive()`: extract all **non-group** (i.e. single) layers from this Layer Group and from all its child Layer Groups recursively. When the method encounters a Layer Group, it re-applies itself on it and appends the result in place in the resulting array, instead of the Layer Group. However the final layers order is not necessarily guaranteed as it loops through child layers according to the ascending order of `_leaflet_id`, not to insertion order. This could replace [Leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster) internal `_extractNonGroupLayers()` method (if ready to loop twice).
- `eachSingleLayerRecursive()`: similar but applying a method and optionally a context. This should be directly usable by [leaflet-search](https://github.com/stefanocudini/leaflet-search) and [leaflet-pip](https://github.com/mapbox/leaflet-pip) to replace `eachLayer()`.

I added corresponding test cases.

### DIFF
--- a/spec/suites/layer/LayerGroupSpec.js
+++ b/spec/suites/layer/LayerGroupSpec.js
@@ -54,6 +54,25 @@
 		});
 	});
 
+	describe("#getSingleLayersRecursive", function () {
+		it('gets all non-group layers recursively', function () {
+			var lg1 = L.layerGroup(),
+			    marker1 = L.marker([0, 0]),
+			    lg2 = L.layerGroup(),
+			    marker2 = L.marker([0, 0]),
+			    lg3 = L.layerGroup(),
+			    marker3 = L.marker([0, 0]);
+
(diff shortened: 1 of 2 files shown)


IvanSanchez:
I think we can postpone this until 1.1. Now that we're at 1.0.0-rc3, it's time to focus on the remaining bugs and not on new features.

ghybs:
You are absolutely right, this small feature is just a convenient tool anyway, to facilitate usability of plugins that could have implemented it themselves (as already done in Leaflet.markercluster BTW).

Issue #3992, opened 2015-11-08
clicking on a marker whose popup is open make the popup close AND open at the same time (V1 beta2)


remilev:
http://jsfiddle.net/Pe5xU/261/

With the new leaflet v1 beta 2 :
1 - you click on the blue marker icon -> the popup open
2 - you REclick on the marker icon -> the popup close (fadeOut) AND re-open (fadIn)

With the older 0.7 version : 
1 - you click on the blue marker icon -> the popup open
2 - the popup is just closing. It is more what we expect.

To have the same result in v1 than in v0.7, we must use map.options.closePopupOnClick : false.
But of course, with that, a click on the map does'nt close the open popup anymore...

How to retrieve the old behaviour of v0.7 with v1 ?

IvanSanchez:
@yohanboniface Is this the expected behaviour with the 1.0.0 event handling?

yohanboniface:
> @yohanboniface Is this the expected behaviour with the 1.0.0 event handling?

Nope. I add that to my todo list, unless you have a look before ;)

JoranBeaufort:
Could it be that this problem still exists when binding popups to polygons? So if a popup is open and a user clicks somewhere else on the polygon that the `map.on('popupopen',function(e){})` or the `map.on('popupclose',function(e){})` are not fired correctly?

yohanboniface:
Which Leaflet version are you using?

JoranBeaufort:
Im using Leaflet 1.0.0-rc2: `<script src="https://npmcdn.com/leaflet@1.0.0-rc.2/dist/leaflet.js"></script>`. Maybe a bit more information: I am creating polygons with: `L.polygon()` and then bind popups to the polygons with: `polygon.bindPopup()`. In the popup, I have a div in witch I generate a SVG. So I only want to generate the SVG if the popup is actually open as to increase performance. So naturally, I use: 
```
map.on('popupopen',function(e){
   ... SVG GENERATION HERE ...
}
```
And when the popup closes, I destroy the SVG again:
```
    map.on('popupclose',function(e){
        var svg = d3.select("#my-SVG-div").selectAll("*").remove();
    });
```
It works fine for all cases, except when I click on the same polygon as has an open popup. Then the events do not seem to get fired.

yohanboniface:
May just be another face of #4743.
Can you try with #4769 by chance?

JoranBeaufort:
[...]

PR #4788, opened 2016-08-02
Only stop "preclick" on popup open for markers

yohanboniface:
We actually do not toggle the popup on click for L.Path.

Fixing what is described here https://github.com/Leaflet/Leaflet/issues/3992#issuecomment-236885662

Side note: I'm not a big fan of having the "toggling" only for Marker and not for Path. It's not consistent as a UX and it introduce ugly `if` in the code.
What about instead having it by default for all layers, and adding an option (`toggleOnClick`) to opt-out?


mourner:
Let's revisit after rc3. I'm up to an option and on by default for everything.

georgio8:
The inconsistency has been bothering users of: [my app](https://github.com/cyclestreets/cyclescape-issuemap)

PR #4791, opened 2016-08-04
Do not call zoomIn/zoomOut on zoomControl click if already at zoom limit

### LINKED ISSUE #4790: moveend event called when clicking on disabled zoom control
When map reaches max or min zoom and zoom control buttons are greyed out, clicking on them fires a `moveend` event, even though the map doesn’t change at all.

yohanboniface:
Fix #4790


IvanSanchez:
LGTM.

PR #4792, opened 2016-08-04
Docstrings: clarifications about noWrap and wrapLng

IvanSanchez:


src/geo/crs/CRS.js
```diff
@@ -80,17 +80,17 @@ L.CRS = {
...
 	//
 	// @property wrapLat: Number[]
-	// Like `wrapLng`, but for the latitude axis.
+	// Like `wrapLng`, but for the latitude (vertical) axis.
 
 	// wrapLng: [min, max],
 	// wrapLat: [min, max],
 
-	// @property infinite: Boolean = false
+	// @property infinite: Boolean
```
yohanboniface: why removing the default?

IvanSanchez: Because options have defaults and properties don't. See http://leafletjs.com/reference-1.0.0.html#crs-property

yohanboniface: oh, ok :)

Issue #4740, opened 2016-07-20
L.Polygon.getCenter() fails while not part of a map


Ceremony64:
Getting the center of a polygon (or other type) is not possible while it is not added to a map and results in one of two possible errors:
`TypeError: this._rings is undefined` or `TypeError: this._map is null`, depending on whether the polygon has been added to a map previously.

Heres an example:
https://jsfiddle.net/s81t48m3/

SuberFu:
It looks like getCenter() code have always depends having the polygon being part of a map , since it's calculated from the projected points instead of latlng and needed the map to perform projection (both on the final result and the polygon's `_rings`). 

It would be nice to have a way to invoke getCenter without needing to add it to a map. Perhaps a version where user supplies a version where use can supply their own CRS?

`getCenter()` performs centroid calculation using the map it is in, fallback to L.CRS.EPSG3857 (or whatever Leaflet map's default is) if it's not added to a map.
`getCenter(crs)` performs centroid calculation using the provided `crs`

IvanSanchez:
This is not possible, as `L.Polygon.getCenter()` works on the **projected** coordinates. This is not possible without a CRS, which means this is not possible until the polygon has been added to a map.

However, it would be a good idea to make the code throw an exception with a proper message instead.

snkashis:
@IvanSanchez how does the wording in this PR look?

IvanSanchez:
Fixed by #4820

PR #4820, opened 2016-08-14
Throw error on call to L.Polygon getCenter before map add

snkashis:
References #4740


IvanSanchez:
@snkashis Could we have an extra unit test for this? Something that checks than an error is thrown, like in https://github.com/Leaflet/Leaflet/blob/master/spec/suites/layer/vector/CircleSpec.js#L10

I'd also like to see the same sanity check in `L.Polyline.getCenter()`, and a note in the docstrings, if it's not too much to ask.

snkashis:
ok done.

IvanSanchez:
Looking good!! :-)

PR #4829, opened 2016-08-19
Fixes #4813: Added rounding of position to translate3d

### LINKED ISSUE #4813: Blurry tooltips due to unrounded CSS translation
leaflet.js version: "1.0.0-rc.3+27f05a2"

Tooltips were drawing half-pixel digits for the pos.x position (at least the only one it was doing it for me) so:

transform: translate3d(1024.5px, 359px, 0px); - was blurry
transform: translate3d(1024px, 359px, 0px); - was perfect (same with if I made it 1025px)

So, I hacked it (as in most likely not correct) line 1464 from:

'translate3d(' + pos.x + 'px,' + pos.y + 'px,0)') +

-to- 

'translate3d(' + Math.round(pos.x) + 'px,' + Math.round(pos.y) + 'px,0)') +

And everything with my tooltips has been perfect.

sisou:
Chrome renders half pixels when doing CSS transforms, which can make the tooltips blurry when the tooltip has an uneven hight or width.

`Math.round()` could also be added to the `layer/Tooltip.js` directly in the [_setPosition](https://github.com/Leaflet/Leaflet/blob/master/src/layer/Tooltip.js#L124), but I prefer handling exact numbers within the code and only rounding when rendering.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -190,7 +190,7 @@ L.DomUtil = {
 		el.style[L.DomUtil.TRANSFORM] =
 			(L.Browser.ie3d ?
 				'translate(' + pos.x + 'px,' + pos.y + 'px)' :
-				'translate3d(' + pos.x + 'px,' + pos.y + 'px,0)') +
+				'translate3d(' + Math.round(pos.x) + 'px,' + Math.round(pos.y) + 'px,0)') +
 			(scale ? ' scale(' + scale + ')' : '');
 	},
 


sisou:
Referring to https://github.com/Leaflet/Leaflet/issues/4813#issuecomment-241052557:

Sadly, I have no idea what I need to look at to verify. Everything looks sharp for me.

sisou:
The issue that this PR addresses is even visible in the Leaflet/debug/map/tooltip.html page:

beforeaftergithub

src/dom/DomUtil.js
```diff
@@ -190,7 +190,7 @@ L.DomUtil = {
 		el.style[L.DomUtil.TRANSFORM] =
 			(L.Browser.ie3d ?
 				'translate(' + pos.x + 'px,' + pos.y + 'px)' :
-				'translate3d(' + pos.x + 'px,' + pos.y + 'px,0)') +
+				'translate3d(' + Math.round(pos.x) + 'px,' + Math.round(pos.y) + 'px,0)') +
```
davojta: these rounds will affect even tiling mechanism. May be we should isolate it only for tooltips ?

sisou:
https://github.com/Leaflet/Leaflet/pull/5089 is a better solution.

PR #4830, opened 2016-08-19
Add new "corner" direction to tooltips

sisou:
This allows for a fixed positioning of the tooltip that is not dependant on the height or width of the DOM element.

My use case for this is that I have on my map many markers, that all look the same but have different permanent tooltips (some longer, some shorter, some multiline, some not).

Now, the equal positioning of these tooltips proves difficult with the available directions, because in my case, I do not want the tooltips vertically or horizontally centered, but want its upper left corner in the same place (relative to their markers) for all of them.

Thus I need a "direction", that does not include its height or width in its position calculation.

PS: The naming of the direction is debatable, I also thought about naming it `"fix(ed)"`.

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -138,6 +138,8 @@ L.Tooltip = L.DivOverlay.extend({
 			pos = pos.subtract(L.point(tooltipWidth / 2 - offset.x, -offset.y));
 		} else if (direction === 'center') {
 			pos = pos.subtract(L.point(tooltipWidth / 2 + offset.x, tooltipHeight / 2 - anchor.y + offset.y));
+		} else if (direction === 'corner') {
+			pos = pos.add(L.point(offset.x + anchor.x, offset.y + anchor.y));
 		} else if (direction === 'right' || direction === 'auto' && tooltipPoint.x < centerPoint.x) {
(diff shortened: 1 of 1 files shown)


perliedman:
Thanks, more alternatives for how to place the tooltips sounds like something that can be useful to many.

However, I'm thinking this PR might be a bit specific, and it would be better to generalize the tooltip placement and make it easier to add custom placement.

How about breaking out the conditional you added to (https://github.com/Leaflet/Leaflet/blob/master/src/layer/Tooltip.js#L135-147) into a new method, something like `getPosition(direction)`. This would make it trivial to inherit from `L.Tooltip` and add anything you want.

/cc @yohanboniface

yohanboniface:
Making easier to inherit from L.Tooltip to add custom behaviour sounds like a good option. I guess the tricky part is to have all the computation kept DRY and done once (offset, anchor, width, and so on). A bit in rush right now sadly to be more specific sorry :s

sisou:
Would it make sense to add a `custom` direction and, when that one is used, read from another option parameter a user-suppliedfunction that returns the position

yohanboniface:
My opinion is that the right answer for custom behavior is extending. So we may make this easier by refactoring the class a bit.

PR #4838, opened 2016-08-24
Make event firing reentrant

### LINKED ISSUE #4833: 'moveend' event handlers can be removed by a previous handler and throw an error
Reproduce the issue

1. Go to url https://jsfiddle.net/joongtang/zr0qxq8g/
2. Zoom out from level 10 to level 8 at once
3. View console tab.
4. Shown error message : "Uncaught TypeError: Cannot read property 'fn' of undefined"

Reproducible using Leaflet-1.0.0-rc3

perliedman:
Close #4833

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -341,6 +341,26 @@ describe('Events', function () {
 
 			expect(spy.called).to.be(false);
 		});
+
+		it('handles reentrant event firing', function () {
+			var obj = new L.Evented(),
+			    spy = sinon.spy();
+
+			obj
+				.addEventListener('test1', function () {
+					obj.fire('test2');
+				})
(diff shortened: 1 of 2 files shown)


spec/suites/core/EventsSpec.js
```diff
@@ -341,6 +341,26 @@ describe('Events', function () {
...
+		it('handles reentrant event firing', function () {
+			var obj = new L.Evented(),
+			    spy = sinon.spy();
+
+			obj
+				.addEventListener('test1', function () {
+					obj.fire('test2');
+				})
+				.addEventListener('test2', function () {
+					// NOP, just to make sure event actually fires
```
yohanboniface: Shouldn't we have a spy for that too then?

perliedman: Yes, you're right!

PR #4839, opened 2016-08-24
Add ColorIcon - a traditional marker icon of configurable color

jkufner:
Icon is implemented using data: URL to generate SVG image of requested
color. The shadow is embedded PNG file of original marker.

Thanks to use of data: URLs this icon does not use any external files.

The SVG image of the marker is the original Leaflet marker without
gradient and some redundant XML.

### DIFF
--- a/src/layer/marker/ColorIcon.js
+++ b/src/layer/marker/ColorIcon.js
@@ -1,0 +1,80 @@
+/**
+ * @class ColorIcon
+ * @aka L.ColorIcon
+ * @inherits Icon
+ *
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This should rather be a plugin - we don't want to change the default icon just yet, and Leaflet 1.0.0 must have compatibility with old browsers which don't support SVG.

jkufner:
Actually this was not meant to replace the default icont, but rather to complement DivIcon. However using this as a default would be nice when IE8 is long gone.

PR #4848, opened 2016-08-25
Edge also has problems with `bringToFront`

nikolai-b:


northlandiguana:
bringToFront() seems to remove any event listeners from the Path layer in Edge.

perliedman:
@northlandiguana could you please open a new issue for this, and provide an example of the issue?

PR #4356, opened 2016-03-30
Tutorial about making leaflet plugins

IvanSanchez:
Part of #4239, lengthy-ish, and hopefully not completely full of typos and mistakes.

Depends on #4356, because that's where I added the templates for the iframed examples.

Issue #4239, opened 2016-02-15
More tutorials needed


IvanSanchez:
The [tutorials](http://leafletjs.com/examples.html) page needs some love. Leaflet has a lot of features, but there are tutorials for just a bunch of them.

We should spend some time converting some of the pages in the `debug/` directory into tutorials/examples.

A non-comprehensive list:

* [ ] Complete newbie's guide to How To Use a Leaflet Plugin
* [ ] Image overlays (prettier version of `image-overlay.html`)
* [ ] Map bounds (with `bounceAtBounds`)
* [ ] Layer opacity
* [ ] WMS (with special attention to Leaflet/WMS nomenclature about the term "layer")
* [ ] Fractional zoom controls (prettier version of `zoomdelta.html`)
* [ ] setView, setZoom, fitBounds and flyTo (prettier version of `zoompan.html`)
* [ ] `L.CRS.Simple`
* [ ] Custom CRSs (using, for example, a [map of the galaxy](https://en.wikipedia.org/wiki/Milky_Way#/media/File:Artist%27s_impression_of_the_Milky_Way_%28updated_-_annotated%29.jpg) with galactic CRS)
* [ ] A primer on how to create Leaflet plugins (`L.Class`, `L.Class.include` vs `L.Class.extend`, etc)

mourner:
@IvanSanchez fantastic CRS tutorial, glad to see this being finally addressed! 
Also see #3271 for previous discussion on tutorials.

Kayoku:
Hi everyone.
I'm just discovering Leaflet, then i did [this](http://leafletjs.com/examples/quick-start.html) "Quick Start Guide". This is maybe stupid but, when you need to make sure that the map container has a defined height with : 
`#mapclass { height: 180px; }`
that's not working. Because we did this before : 
`<div id="mapid"></div>`
Just rename "mapclass" to "mapid" ? (For beginner, it's maybe more clear)
What do you think ?

mourner:
Hmm, yeah, that's a bug in the tutorial — well noted! Would you want to fix this yourself and make a pull request? It's as easy as clicking "edit" on this file https://github.com/Leaflet/Leaflet/blob/gh-pages/examples/quick-start.md, making a change right in the browser and saving.

Kayoku:
Ok thanks. Correction [#4317](https://github.com/Leaflet/Leaflet/pull/4317) open !

skorasaurus:
[...]

PR #4854, opened 2016-08-27
Docs: Add the plugin tutorials from #4356

IvanSanchez:
Replaces #4356, fulfills part of #4239


IvanSanchez:
I've rebased, after some conflicts popped up due to #4965. Should be fine for merging now. cc @mourner @yohanboniface @perliedman

perliedman:
_Really_ like the work you've put into these. Even as an experienced Leaflet user it gives some "huh, hadn't thought of it like that" moments. Great job!

I read through them, but of course I might still have missed some details - I think that's fine and we'll clean up the small things when we find them.

IvanSanchez:
@perliedman Appreciated :-)

With a bit of luck, I'll be able to put more work on the "zoom tutorials" of #5007, while @mourner fights the ES6 dragon.

PR #4869, opened 2016-08-30
Use translate() in DomUtil.setTransform

kernc:
Use translate() instead of translate3d() because the z transformation is 0 anyway, and browsers seem to better support the former transformation (i.e. the interpolation (e.g. when upscaling) remains cubic (in browsers that support it, e.g. Chromium)  instead of linear as noticeable (in Chromium) in https://jsfiddle.net/Lb6abxmu/).

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -188,9 +188,7 @@ L.DomUtil = {
 		var pos = offset || new L.Point(0, 0);
 
 		el.style[L.DomUtil.TRANSFORM] =
-			(L.Browser.ie3d ?
-				'translate(' + pos.x + 'px,' + pos.y + 'px)' :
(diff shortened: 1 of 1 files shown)


nathancahill:
translate3d() is faster than translate() because it triggers hardware acceleration.

kernc:
Could this be an [urban legend](https://stackoverflow.com/a/22337612/1090455) from the times CSS3 was only a draft?

For the [**test I run**](https://jsfiddle.net/533hc71h/174/) on Firefox 45, Chromium 52, Edge 25.10586, IE 11.545, Firefox 48 on Android and Android Browser 5.1, including [one of the original JSPerf tests](https://web.archive.org/web/20141001023754/http://jsperf.com/translate3d-vs-xy/36), I see exactly the opposite, namely that `translate()` is the slightly faster one. Could it be that, in 2016, all CSS transforms are pushed to the GPU?

Issue #4882, opened 2016-09-05
Marker events are not propagated to map


dnsBlah:
As the title mentioned, I made some markers, and drawing polylines in between.

I did put a click function in for zoom on click:
```
    map.on("click", function(e) {
        map.setView(e.latlng, map.getZoom() + 2);
    });
```

This is not being triggered when I click on a marker, but it does when clicking on polylines.
(I have popup on polylines, and would rather not to have it zooming when I click the polyline)

IvanSanchez:
So it appears that while `L.Path` interactive events are propagated to the map as documented and expected for 1.0.0, the same is not true for markers. A marker's `click` event is not propagated into the map, while a `click` event on a polyline is propagated, which makes the event propagation inconsistent.

(On a side note, the OP's problem is just a lack of `L.DomUtil.stopPropagation`, which highlights the need for #4239)

Playground demo'ing the issue:
https://playground-leaflet.rhcloud.com/mawe/edit?html,console,output

This *sounds* like an easy task of running `addEventParent` to `L.Marker`s when added to a map, but I'll let our event expert @yohanboniface weight in.

yohanboniface:
Humm, but that's expected behaviour https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Marker.js#L63 :)

IvanSanchez:
@yohanboniface Should we set events on `L.Path`s non-bubbly as well?

dnsBlah:
@IvanSanchez how is your sample not zooming on the line?
I see you don't have zoom on click, on the double click zoom.

Also I work with a json file with my lines.
```
var drawPolyline = function(pointList, tColor, tWeight = 3, tOpacity = 0.5) {
        var tLine = new L.Polyline(pointList, {
            color: tColor,
            weight: tWeight,
            opacity: tOpacity,
            smoothFactor: 1
        });
        return tLine;
    }
```

Load links, and draw
```
$.getJSON("/map/getLinks", function (jsonLinks) {
        $.each(jsonLinks, function(index, d){
            var pointA = new L.LatLng(d.fromLat, d.fromLng);
            var pointB = new L.LatLng(d.toLat, d.toLng);
            var pointList = [pointA, pointB];

            drawPolyline(pointList, d.color, 3, 1)
[...]

PR #4883, opened 2016-09-05
Turn nonBubblingEvents into a documented boolean option

IvanSanchez:
Derived from conversation in #4882, this turns the (hidden) `nonBubblingEvents` layer option into a new documented (boolean) option, `bubblingEvents`, with explicit different defaults for markers and paths.

This should keep behaviour the same as now, while documenting why markers don't bubble mouse events to the map (by default).

cc @yohanboniface

### DIFF
--- a/src/layer/Layer.Interactive.leafdoc
+++ b/src/layer/Layer.Interactive.leafdoc
@@ -8,6 +8,10 @@ Use the [event handling methods](#evented-method) to handle these events.
 @option interactive: Boolean = true
 If `false`, the layer will not emit mouse events and will act as a part of the underlying map.
 
+@option bubblingMouseEvents: Boolean = true
+When `true`, a mouse event on this layer will trigger the same event on the map
+(unless [`L.DomEvent.stopPropagation`](#domevent-stoppropagation) is used).
+
 @section Mouse events
 
 @event click: MouseEvent
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1317,6 +1317,8 @@ export var Map = Evented.extend({
 		this._fireDOMEvent(e, type);
 	},
 
+	_mouseEvents: ['click', 'dblclick', 'mouseover', 'mouseout', 'contextmenu'],
+
 	_fireDOMEvent: function (e, type, targets) {
 
 		if (e.type === 'click') {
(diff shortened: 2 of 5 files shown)


yohanboniface:
The tricky part here is that this option is only active when actually listening to that event.
So if you create a marker with `bubblingEvents: false` but do not listen for any event, it will still let the events go through the map. So I'm not sure that would be less confusing then than how it is currently. :s

IvanSanchez:
Maybe that's just a matter or wording in the docstrings?

Saulzi:
I could really do with this functionality, as when I have a bunch of markers I would like to know where the user clicks on the map rather than that they clicked on  a marker.

Is there any timeframe as to when this will be merged into master / 1.1 release

src/layer/Layer.Interactive.leafdoc
```diff
@@ -8,6 +8,10 @@ Use the [event handling methods](#evented-method) to handle these events.
 @option interactive: Boolean = true
 If `false`, the layer will not emit mouse events and will act as a part of the underlying map.
 
+@option bubblingEvents: Boolean = true
```
perliedman: How about calling the option `bubblingMouseEvents`, to make it more explicit which events it applies to (even though the doc string already says this)?

src/map/Map.js
```diff
@@ -1299,6 +1299,8 @@ export var Map = Evented.extend({
 		this._fireDOMEvent(e, type);
 	},
 
+	_nonBubblingEvents: ['click', 'dblclick', 'mouseover', 'mouseout', 'contextmenu'],
```
perliedman: Could we just call this `_mouseEvents` or similar? The mix of the option `bubblingEvents` and `nonBubblingEvents` in the condition below made my head spin for a while.

src/map/Map.js
```diff
@@ -1339,7 +1341,7 @@ export var Map = Evented.extend({
 		for (var i = 0; i < targets.length; i++) {
 			targets[i].fire(type, data, true);
 			if (data.originalEvent._stopped ||
-				(targets[i].options.nonBubblingEvents && Util.indexOf(targets[i].options.nonBubblingEvents, type) !== -1)) { return; }
+				(targets[i].options.bubblingEvents === false && Util.indexOf(this._nonBubblingEvents, type) !== -1)) { return; }
```
perliedman: With my suggestions, this condition would turn into:

```js
targets[i].options.bubblingMouseEvents === false && Util.indexOf(this._mouseEvents, type) !== -1)
```

...which looks clearer at least to me.

perliedman:
\o/ Feels good to finally close this!

PR #4894, opened 2016-09-09
Blog post about the FOSS4G (and 1.0.0 delay)

IvanSanchez:
Two things:

* I have a photo of Vlad, Per and me in the cafeteria area, but Sanghee's daughter is in the frame, and I don't feel comfortable uploading that (without his permission)
* Do we want to add more detail about the FOSS4G? Some more references to OpenLayers talks/features?


mourner:
I'm not sure we need another "very soon" blog post. I'd rather fix/postpone remaining two issues, then write up a final Leaflet 1.0 announcement instead, and maybe mix in a few FOSS4G pics, or just tweet them instead.

mourner:
We'll post a 1.0 blog post instead.

PR #4900, opened 2016-09-10
Fixed dead link #4899

JonasDralle:
this takes care of the broken link to this exact branch

issue #4899

have fun

PR #4901, opened 2016-09-10
Fixed dead link inside plugins.md

### LINKED ISSUE #4899: Dead Link in plugins.md
http://leafletjs.com/plugins.html#develop-your-own
The link at the very bottom of the page is broken because the branch was renamed

> pull request with the addition to Leaflet repo's [gh-pages branch](https://github.com/Leaflet/Leaflet/tree/gh-pages) (`plugins.md` file).

I already have an version where that is fixed. I only opened this Issue for reference inside the pull request.
Should be online in 10 min or less

JonasDralle:
this takes care of the broken link to this exact branch
this fixes #4899
have fun

same changes as in #4900


yohanboniface:
Thanks, that's better :)
But still the link itself should point to master and inside docs/ instead of the root of gh-pages* :)

JonasDralle:
*facepalms himself*
Yea you're right

PR #4902, opened 2016-09-10
fixed dead link

JonasDralle:
issue #4899

sorry that it took so long.

I already tried to fix this error twice (See #4900 and ##4901)


yohanboniface:
Thanks :)

PR #4904, opened 2016-09-10
Fix return type of GridLayer.getContainer in leafdoc

alejo90:
Return type of `GridLayer.getContainer` is currently listed as `String` when it should be `HTMLElement` in leafdoc.


alejo90:
ping @IvanSanchez

IvanSanchez:
@alejo90 It'd be nice if you could group all the tiny typos in the docstrings in one PR next time ;-)

PR #4905, opened 2016-09-10
Better check for undefined variable

mathiasrw:
Updated a single statement to the fastest way to check for undefined variable

### DIFF
--- a/src/geometry/LineUtil.js
+++ b/src/geometry/LineUtil.js
@@ -49,7 +49,7 @@ L.LineUtil = {
 	_simplifyDP: function (points, sqTolerance) {
 
 		var len = points.length,
-		    ArrayConstructor = typeof Uint8Array !== undefined + '' ? Uint8Array : Array,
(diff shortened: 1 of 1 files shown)


IvanSanchez:
How much faster is this, then?

mathiasrw:
Donno - and as jsperf.com is not up i wont find out today. 

Feel free to reject if you think its a silly change.

IvanSanchez:
The current way is due to a 4-year-old commit (https://github.com/Leaflet/Leaflet/commit/50c6c480b4f638ec31a74600edca0da28212a333), with the purpose of re-defining `undefined` in a closure, seemingly for helping a uglifier make files smaller. It should be worth it to run a few tests to see if there is indeed a speed improvement. If not, the current code will provide smaller file sizes.

mathiasrw:
I see - its actually the point then...

Ill close the PR for now and will come back if I find hard evidence of speed...

PR #4906, opened 2016-09-11
Adding an ISSUE_TEMPLATE.md

yohanboniface:
First suggestion. Thoughts?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -18,6 +18,6 @@
 ## Minimal example reproducing the issue
 
 - [ ] this example is as simple as possible
-- [ ] this example do not rely on any third party code
+- [ ] this example does not rely on any third party code
 
 Using http://playground-leaflet.rhcloud.com/ or any other jsfiddle like site.
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,23 @@
+- [ ] I'm reporting a bug, not asking for help
+- [ ] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
+- [ ] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
+- [ ] I've searched through the issues to make sure it's not yet reported
+
+----
+
+## How to reproduce
+
+- Leaflet version I'm using:
+- Browser (with version) I'm using:
+- OS/Platform (with version) I'm using:
+- step 1
+- step 2
+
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+- [ ] I'm reporting a bug, not asking for help
+- [ ] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure this is not a documented behaviour
```
IvanSanchez: Typo: «not documented behaviour». Maybe it's clearer if the wording is «to make sure the behaviour is documented and expected»

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] I'm sure this is a Leaflet code issue, not an issue with my own code
```
IvanSanchez: I'd add a line about ionic, cordova, react, ractive, vue and all the other JS frameworks we don't (really) care about.

IvanSanchez:
I'd also add a couple of lines about specifying the browser for browser-specific bugs (or the more generic word "platform" to include phones and such), and also (very important) which version the bug is reproducible in.

yohanboniface:
> I'd also add a couple of lines about specifying the browser for browser-specific bugs (or the more generic word "platform" to include phones and such),

Right! :)

> and also (very important) which version the bug is reproducible in.

Damn right! :)

ISSUE_TEMPLATE.md
yohanboniface: Changed :)

ISSUE_TEMPLATE.md
yohanboniface: Added :)

mourner:
I like it, although I would maybe simplify some statements and formatting to make the template shorter (longer templates are less likely to be adhered to).

yohanboniface:
I guess those two sentences can be made shorter:

> - [ ] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
>- [ ] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)

To you have any other idea in mind?

danzel:
The language on this bit could be improved I think:
> this example do not imply any third party code

I'm not sure what it is meant to mean.

perliedman:
@danzel a changed "imply" to "rely on", which at least for me makes more sense!

yohanboniface:
Great, thanks Per :)

mourner:
@yohanboniface maybe you meant "does not", not "do not"?

yohanboniface:
> @yohanboniface maybe you meant "does not", not "do not"?

Erf, fixed :p

Issue #4279, opened 2016-02-26
Layers should inherit pane from containing L.LayerGroup 


IvanSanchez:
I've been reading #4277 and started to check some of the code and think.

When a layer is added to a `LayerGroup` (or a child class like `FeatureGroup` or a `GeoJSON`), the layer does **not** inherit the `pane` option of the `LayerGroup` which contains it.

So, for example, this code from the "map panes" tutorial adds a layer to the `labels` pane:

```js
var positronLabels = L.tileLayer('http://{s}.basemaps.cartocdn.com/light_only_labels/{z}/{x}/{y}.png', {
	attribution: cartodbAttribution,
	pane: 'labels'
}).addTo(map);
```

But this won't:
```js
var positronLabels = L.tileLayer('http://{s}.basemaps.cartocdn.com/light_only_labels/{z}/{x}/{y}.png', {
	attribution: cartodbAttribution,
});

var group = L.layerGroup();
group.options.pane = 'labels';
group.addLayer(positronLabels).addTo(map);
```

So I wonder if `LayerGroup`s should force their `pane` option on layers contained in them, and if it would be possible to override this behaviour, or add an option like `forceGroupPaneOnLayers` or something like that. Or maybe it's better to handle the practical cases like #4277 and forget about a generic solution.

I would prefer a generic solution, but I'm afraid it might not behave correctly with vector renderers in custom panes somehow.

jwoyame:
I have two conflicting opinions on this:

1. Having the group change the pane of its child layers will be a source of error whenever Markers and Paths wind up in the same group, as they cannot be in the same pane. When loading a bunch of GeoJSON features in an array, that situation could come up.

2. It would be *GREAT* if some Layer-derivative could operate on a group of child layers as a pane. For example, with a GeoJSON dataset containing lots of river polylines, the LayerGroup has to remove/readd every layer to toggle its own visibility. If the group could be associated with a pane, that would enable far more simple toggling. Functions like `bringToFront`, `setZIndex`, etc. would enjoy greater performance if they could operate on panes as well.

[...]

PR #4909, opened 2016-09-12
Add a "membersPane" option to groups

IvanSanchez:
My attempt at providing a solution for #4277 / #4279. Haven't really tested it, but should work. *crosses fingers*

Now the question is the precedence we want for the `pane` options. We might want to push the `pane` option into a hidden `_defaultPane` option, so that explicit `pane`s from layers would override the group's `membersPane` which would override the `_defaultPane`. Or something like that.

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -63,7 +63,9 @@ L.Layer = L.Evented.extend({
 	// @method getPane(name? : String): HTMLElement
 	// Returns the `HTMLElement` representing the named pane on the map. If `name` is omitted, returns the pane for this layer.
 	getPane: function (name) {
-		return this._map.getPane(name ? (this.options[name] || name) : this.options.pane);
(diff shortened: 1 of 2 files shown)


ghybs:
Hi,

Pushing further the question on "what should be the generic behaviour for inheriting from group `pane`?":

_What to do in the case of a group added into another group?_

With the current PR, "nothing" happens: the parent group A defines the `_groupPane` option of the child group B, but the latter still has its own untouched `membersPane` option which dictates its child layer C `_groupPane`.
So currently C does not inherit A's `membersPane`, but only its immediate parent group B's `membersPane`.
And if B does not have a `membersPane`, C remains on its default `pane`.

It should be possible to correct this behaviour quite simply if desired, but it seems to me that such option inheritance scheme, with or without enforcing or explicit overriding, is currently unknown in Leaflet.
Therefore it might open the door to many confusion.

Issue #4566, opened 2016-05-12
fitBounds with custom projection acts really strange


theashyster:
At first I thought that this is fixed with this really easy fix https://github.com/Leaflet/Leaflet/issues/4528, but looks like it actually is not, and by doing the fix a new possible issue has occurred, not 100% sure. Also looks like the fix was actually not done in the right place, probably. This is all speculation, because I cannot reproduce the issue using plain Leaflet, but with a custom projection this issue pops up and does not go away.

By removing the `.add(padding)` from the `boundsSize` we now can ran into issues when working with bounds that are just a point and in that case `boundsSize` will be 0 and will cause trouble on this line calculating scale as `Infinity` and because of that `getBoundsZoom` returns `NaN`:
`var scale = Math.min(size.x / boundsSize.x, size.y / boundsSize.y);`

Working beta2: http://playground-leaflet.rhcloud.com/divo/1/edit?html,output
Working 0.7.7: http://playground-leaflet.rhcloud.com/nov/1/edit?html,output
Not working master: http://playground-leaflet.rhcloud.com/gin/1/edit?html,output
Not working rc1: http://playground-leaflet.rhcloud.com/cit/1/edit?html,output

The other issue I have is that in `getBoundsZoom` function we have this line:
`zoom = this.getScaleZoom(scale, zoom);`

Since zoom is set before like this `var zoom = this.getZoom() || 0` when we call `getScaleZoom` it always gives different result when we pass the function different zoom level and the same scale. If we do multiple fitBounds() calls in a row with the same bounds, but different starting zoom level (do one call, then do another from the resulting zoom level, then do another) this causes map to zoom around strangely. Size and boundsSize does not change, but the calculated zoom using `getScaleZoom` changes due to the changes in the starting zoom level.

Issue shown on rc1: http://playground-leaflet.rhcloud.com/faf/1/edit?html,output
Working as expected on beta2: http://playground-leaflet.rhcloud.com/gug/1/edit?html,output
Not working at all on master: http://playground-leaflet.rhcloud.com/quxi/1/edit?html,output
-- Click on the fitBounds button under the map to see the issue in action

Looks like this is the commit that kind of turned things ugly using custom projections https://github.com/Leaflet/Leaflet/commit/3fde3c0390e03c6d0187ede1c9418c3c57fd1765

So probably I should ask @IvanSanchez to take a look at this

perliedman:
Yes, we had some trouble with this before: #4395, #4402.

[...]

PR #4914, opened 2016-09-15
Fix handling of NaN in getScaleZoom

theashyster:
Fixed the issue with `getScaleZoom` returning `NaN` when `L.CRS.zoom()` used with a scale factor of `Infinity` which could return `NaN` causing this. Discussed here https://github.com/Leaflet/Leaflet/issues/4566


IvanSanchez:
One-line fixes, love those.

PR #4916, opened 2016-09-15
Ensure zoom is within span when adding a layer with min/maxzoom, fixes #4915

### LINKED ISSUE #4915: Map gets initialized past maxZoom level when adding tileLayer with maxZoom smaller than initial zoom
Hello,
I am not sure if this is an issue or it is the way it is supposed to work, because I can reproduce it in nearly every version of `Leaflet`.

When initializing the `map` I pass the option `zoom` to be `19`, later on I add a `tileLayer` to show my base layer and that by default has a `maxZoom` option set to `18` in `Leaflet`.

When the `map` is shown I get a `map` without the base layer shown and by pressing `zoomOut` on `zoomControl` the base layer is shown on the `map` (no matter if I enter `19` or `100` as my initial `zoom` level).

I see the point that this could be the way it is supposed to work, since you can control the init `options` of the `map` and set the `zoom` level that is not bigger than the `maxZoom` for `tileLayer`. On the other hand though if I run `map.setView(center, zoom);` after the `tileLayer` is added to the `map` with the same `zoom` level `19` it sets the `map` to the max zoomed in state and does not go any `zoom` levels further.

Demo time http://playground-leaflet.rhcloud.com/waxem/edit?html,output

The `map` will first load without the base layer, pressing the `setView` button will add the base layer.

IvanSanchez:
Fixes #4915, by running the same check from `L.Map.setMaxZoom` in `L.Map._updateZoomLevels`.

Could use a unit test, though.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -281,6 +281,20 @@ describe("Map", function () {
 			expect(map.getMinZoom()).to.be(2);
 			expect(map.getMaxZoom()).to.be(20);
 		});
+
+		it("layer minZoom overrides map zoom if map has no minZoom set and layer minZoom is bigger than map zoom", function () {
+			var map = L.map(document.createElement("div"), {zoom: 10});
+			L.tileLayer("{z}{x}{y}", {minZoom: 15}).addTo(map);
+
+			expect(map.getMinZoom()).to.be(15);
+		});
+
+		it("layer maxZoom overrides map zoom if map has no maxZoom set and layer maxZoom is smaller than map zoom", function () {
+			var map = L.map(document.createElement("div"), {zoom: 20});
+			L.tileLayer("{z}{x}{y}", {maxZoom: 15}).addTo(map);
+
+			expect(map.getMaxZoom()).to.be(15);
+		});
 	});
 
 	describe("#hasLayer", function () {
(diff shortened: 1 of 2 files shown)


theashyster:
Looks like it works fine, but adds a zoom animation and zoomControl is enabled for a moment then gets disabled when `this.setZoom(this._layersMaxZoom);` is called, but I guess that is the way it is, right?
http://playground-leaflet.rhcloud.com/hac/edit?html,output

IvanSanchez:
Yeah, you basically need the `setZoom()` call to set the right zoom level and trigger the disabling of the zoom button.

It's not very pretty, but this is an edge case, so I won't worry about it not being 100% beautiful.

theashyster:
Okay, I see, that is what I thought.
I agree, thanks for the fix. :+1:

theashyster:
@IvanSanchez Since I cannot push changes to your branch, I will paste the tests I created here.

```javascript
it("layer minZoom overrides map zoom if map has no minZoom set and layer minZoom is bigger than map zoom", function () {
	var map = L.map(document.createElement("div"), {zoom: 10});
	L.tileLayer("{z}{x}{y}", {minZoom: 15}).addTo(map);

	expect(map.getMinZoom()).to.be(15);
});

it("layer maxZoom overrides map zoom if map has no maxZoom set and layer maxZoom is smaller than map zoom", function () {
	var map = L.map(document.createElement("div"), {zoom: 20});
	L.tileLayer("{z}{x}{y}", {maxZoom: 15}).addTo(map);

	expect(map.getMaxZoom()).to.be(15);
});
```

PR #4933, opened 2016-09-23
DOCS GeoJSON update usage example to L.geoJSON

ghybs:
Replaced factory example `L.geoJson` (from Leaflet 0.7) by `L.geoJSON` for consistency with the new factory signature rendered just below in the generated docs.
http://leafletjs.com/reference-1.0.0.html#geojson-example


will02:
hello

Issue #3989, opened 2015-11-06
Leaflet-0.7.3 is not working anymore


leolux:
I did not make any code changes, but suddenly the map stays white and nothing gets displayed inside the map.

Is the version 0.7.3 supposed to work at the moment?
Has the compatibility for older versions beeing dropped ?

IvanSanchez:
0.7.3 works just fine: https://playground-leaflet.rhcloud.com/suy/edit?html,output

Have you tried using a different version? Using the debug console to check for any errors?

leolux:
Well, the link shows the error:
Mixed Content: The page at 'https://playground-leaflet.rhcloud.com/runner' was loaded over HTTPS, but requested an insecure script 'http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.js'. This request has been blocked; the content must be served over HTTPS.

leolux:
Here is a screenshot:
http://i65.tinypic.com/2zswknk.png

leolux:
I am running chrome browser Version 46.0.2490.80 m

leolux:
Even if I server leaflet-0.7.3.css and leaflet-0.7.3.js from my own server the tiles get blocked from the browser nevertheless. Here is an error message in chrome:
"Mixed Content: The page at 'https://localhost:8181/...' was loaded over HTTPS, but requested an insecure image 'http://b.tile.osm.org/18/137212/89481.png'. This content should also be served over HTTPS."

This seems to be a design problem. How to solve this without making changes to the browser security settings?

leolux:
I will close this issue, because map works in version 0.7.3 at the moment when javascript and css gets server over secure connection. 

But the browser keeps producing warning messages for all tile images ("Mixed Content"). Maybe this warning will be turned into a error in future which would actually stop it from working.

Thank you for your fast support!

leolux:
Related discussion can be found here: https://github.com/Leaflet/Leaflet/issues/3186

IvanSanchez:
@leolux Load that example over plain http or allow http scritps on a https page. There's a separate issue all about why the CDN doesn't allow https.

Still, my point stands: 0.7.3 just works.

PR #4941, opened 2016-09-26
Implement dump-tile-to-canvas functionality. Fixes #3575.

### LINKED ISSUE #3575: Space between tiles on fractional zoom levels in Webkit browsers
screen shot 2015-06-30 at 14 39 12

As the screenshot shows there is a small space between the tiles. This happens to me on fractional zoom levels in Safari, Chrome and Opera. It seems like this was fixed in #2377 but later removed in 3ccbe5b in favour of better performance. 

I tried reimplementing the fix to see the performance impact myself and it was pretty bad furthermore it didn't even fix the problem for Chrome.

Does anyone have an idea how this can be fixed? Fractional zoom levels would be a great feature for our app but as long as this bug remains we can't really use them.

IvanSanchez:
This is basically https://github.com/Leaflet/Leaflet.TileLayer.NoGap, but in the form of a pull request instead of a plugin.

So this adds a new option to `GridLayer` s (enabled by default on `TileLayer`s): instead of adding the tiles to a tile container, dump the tile contents into a `<canvas>`. That `<canvas>` will be inside the same map pane as the `GridLayer`, so it will perform the pan and zoom animations as expected. This is pretty much what OpenLayers 3 does.

The canvas will be as large as the map container (plus a bit of buffer). The whole content of that canvas will be shifted whenever the map is panned. This shift has decent performance (< 1 frame) in all browsers _except IE_. I should make a couple more tests in MS Edge anyway, just to make sure. It's a layer option that can be disabled, so any users hitting a performance problem can just disable this.

All the canvas operations are CORS-safe: even if the tiles come from an origin which does not allow cross-origin requests, everything will work. This is due to the code **not** using `ctx.getImageData`. The canvas will be tainted, but that's not a problem.

There are a few possible improvements, starting by clearing canvases from levels other than the active one when a tile from the active level is loaded - this will reduce artifacts on semitransparent tile sources. Further improvements to behave more like OL3 does are possible (loading lower zoom tiles with a fair buffer and using them when the canvas size is reset, for example).

A bit overkill for just fixing a bug, but I feel there's no other way around it. Plus, it will improve rendering of rotated tiles when time becomes to implementing that.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -51,7 +51,7 @@
  *         tile.height = size.y;
  *
  *         // draw something asynchronously and pass the tile to the done() callback
- *         setTimeout(function() {
+ *         setTimeout(function () {
  *             done(error, tile);
  *         }, 1000);
  *
@@ -125,7 +125,12 @@ L.GridLayer = L.Layer.extend({
 
 		// @option keepBuffer: Number = 2
 		// When panning the map, keep this many rows and columns of tiles before unloading them.
-		keepBuffer: 2
+		keepBuffer: 2,
+
+		// @option dumpToCanvas: Boolean = false
+		// Whether to dump loaded tiles to a `<canvas>` to prevent some rendering
+		// artifacts. Useful only when tiles are `<img>`s or `<canvas>`es.
+		dumpToCanvas: false
 	},
 
 	initialize: function (options) {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
I forgot to mention #3575 so we can have some cross-links around

IvanSanchez:
Note: I updated the code from https://github.com/Leaflet/Leaflet.TileLayer.NoGap , so it now handles `_invalidateAll`.

I'd like to see #3989 merged in, then add the "NoGap" idea as a new class between `GridLayer` and `TileLayer`. Some kind of maybe `PixelGrid`, with an extension function to dump pixels (in the form of calls to [`drawImage`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage), from a `CanvasImageSource` such as an `` or another `<canvas>`) to the canvas.

IvanSanchez:
Superseded by #5290

PR #4955, opened 2016-09-27
Add a 1.0 notice on frontpage

mourner:
cc @IvanSanchez @perliedman


mourner:
Looks like this:

image

Issue #4950, opened 2016-09-27
VML does not update bounds when panning


IvanSanchez:
It seems that the VML renderer is not updating its bounds (and thus, not updating the geometries' clipping box) when the map is panned around.

To reproduce:
- Open up IE8
- Go to http://leafletjs.com/examples/choropleth.html
- Zoom in a couple of times somewhere
- Pan somewhere
- See how the vector layers are cropped by the bounds of the map previous to the panning

![screenshot_20160927_163653](https://cloud.githubusercontent.com/assets/1125786/18878625/a00e79da-84d1-11e6-8e96-5a12b2cfdb58.png)

The solution might be to modify the `_update` method of `L.SVG.VML` and call `this.fire('update')` (as done in https://github.com/Leaflet/Leaflet/commit/c931d330fea6958342423550ea9cc58d1d78fcda ) so that the vector geometries re-clip themselves.

Anyone wanting to pick this up: fork the repo, clone it locally, apply the suggested fix, run some of the examples in the `debug/vector` directory with IE8. If successful, submit a pull request.

PR #4956, opened 2016-09-27
Fire update event from VML renderer

perliedman:
When refactoring updates, we accidentally created #4950, so vector drawing is currently broken for VML (IE8).

This adds firing the `update` event from the VML renderer, which is required since it _doesn't use it's parents `_update` method_, but rather `L.Renderer`'s `_update`, which doesn't fire this event.

I have verified that the GeoJSON debug page works again in IE8 after this fix.

### DIFF
--- a/src/layer/vector/SVG.VML.js
+++ b/src/layer/vector/SVG.VML.js
@@ -38,6 +38,7 @@ L.SVG.include(!L.Browser.vml ? {} : {
 	_update: function () {
 		if (this._map._animatingZoom) { return; }
 		L.Renderer.prototype._update.call(this);
+		this.fire('update');
 	},
 
 	_initPath: function (layer) {


IvanSanchez:
LGTM

t-sch:
Are you sure rendering still works on IE11, Chrome, Edge? From 1.0.0 to 1.0.1 my canvas rendering code (L.VectorGrid derivate) stopped working.

IvanSanchez:
@t-sch `VectorGrid` bugs don't belong to this repo.

t-sch:
I didn't say it is a VectorGrid bug but that code that worked on 1.00RC3 and 1.0.0 doesn't work anymore in 1.01.

IvanSanchez:
fiddle/codepen/[playground](https://playground-leaflet.rhcloud.com/) that displays this behaviour, or it didn't happen.

perliedman:
@t-sch vector rendering most definitely works in Chrome; haven't tested personally, but would be very surprised if it didn't work in IE11/Edge and we didn't hear about it.

That doesn't mean there aren't edge cases where there has been regressions between release candidates and 1.0.1, so if you believe you've discovered something like that, please open a new issue with details (preferably running code like @IvanSanchez says) regarding the problems you're seeing.

t-sch:
I'm currently trying to locate the reason, a comparison between 1.0.0 and 1.0.1 sources didn't show suspicious code.

perliedman:
@t-sch I think this PR is of interest for you: https://github.com/IvanSanchez/Leaflet.VectorGrid/pull/15

t-sch:
Yesss!

Issue #4578, opened 2016-05-18
Big GPU consumption with retina tiles on browser zoom to 75-33 %


davojta:
Browser: Chrome 50, 52
OS: Linux (reproducible on Windows too)
Leaflet version 1.0.0-rc.1
Issue is not reproducible with leaflet 0.7.7

Play test - http://playground-leaflet.rhcloud.com/qiwa/1

Steps - 
1. open link with retina tiles with map for all window - 100% width and 100% height
2. open chrome task manager (Shift + ESC)
3. reduce browser zoom to 50%
4. see GPU memory consumption (medium - near 120 mb, peak - more then 512)

Result - big memory consumption, on old or device with limited gpu memory - destroyеd map layer

Problem: GPU hardware acceleration was added to tiles. It's consist from 2 pieces will-change: 
- opacity property on tile element from (commit 6b67df023  make animations significantly smoother with will-change ) [https://github.com/Leaflet/Leaflet/commit/6b67df023234186e70647334c381fd66e4838721] 
- translate3d properties in addTile https://github.com/Leaflet/Leaflet/blame/master/src/layer/tile/GridLayer.js#L752

hyperknot:
If you are on a Retina device you already have 4x the number of pixels of a normal device (say 3840x2040 pixels). If you even zoom the browser to 50%, now you have 16x the number of pixels of a normal device (say 7680x4080 pixels). 

Obviously having a browser tab display 7680x4080 pixels will use that much GPU memory. But Retina devices are new and have enough memory to handle this. If not, just do not zoom the browser window to 50%, it will make your memory usage 1/4.

The GPU hardware acceleration is a key feature of Leaflet and the added properties are one reason why 1.0 is slightly faster then 0.7.

kaynz:
Wouldn't it be easy to avoid issues like this with a simple flag to disable GPU hardware acceleration?

perliedman:
@kaynz http://leafletjs.com/reference.html#global-l_disable_3d

davojta:
The main issues here is the tight case for retina tiles (on non-retina devices initially but it's can be fixed by override zooms) and acceleration on raster tile elements.
I fix it for our project by set .leaflet-fade-anim .leaflet-tile style to will-change: init; (it's recommended to not use widely will-change on mdn - https://developer.mozilla.org/en-US/docs/Web/CSS/will-change )
[...]

PR #4959, opened 2016-09-27
Fix GPU memory consumption with retina tiles and browser zoom

davojta:
1) switch off will-change
2) add no3d option for layer options
3) add test page

Discussed in https://github.com/Leaflet/Leaflet/issues/4578


mourner:
There's a reason the was `will-change` in CSS — as far as I remember, `flyTo` animations were excruciatingly slow without it on certain browsers/devices, so removing it will inevitably regress performance in some cases. We may revisit this now but test very thoroughly on iOS, Android, all desktop browsers etc.

davojta:
@mourner I think we can do it with special flag or just mention it in docs 
They main benefit is option to switch off translate3d for retina tiles and fallback to translate implementation without flooding GPU memory. There is a rare edge case anyway (browser zoom + retina tiles but cause browser crash )

kaynz:
Any news on this,  @mourner ? :)

src/dom/DomUtil.js
```diff
@@ -184,11 +184,11 @@ L.DomUtil = {
 	// Resets the 3D CSS transform of `el` so it is translated by `offset` pixels
```
Falke-Design, line 184: Docs: parameter `no3d` missing

src/dom/DomUtil.js
```diff
@@ -198,14 +198,13 @@ L.DomUtil = {
 	// Sets the position of `el` to coordinates specified by `position`,
```
Falke-Design, line 198: Docs: parameter `no3d` missing

Falke-Design (changes requested):

jonkoops:
A part of this has already landed in #7872. Are the changes of this PR still relevant?

PR #4963, opened 2016-09-28
docstrings: its to it's where appropriate in GridLayer.js

akkatracker:
Very nitpicky I know.


IvanSanchez:
Nitpicky is OK. I'll hold on a bit, because (from past experience) chances are that you're still reading comments through the code and find something else to nitpick about, and usually that happens during the 5 minutes just after I press the "merge" button ;-)

akkatracker:
That looks to be about it.

Thanks

On 28 September 2016 at 18:48, Iván Sánchez Ortega <notifications@github.com

PR #4972, opened 2016-09-29
Symlink reference-1.0.0.html to reference.html

IvanSanchez:
This replaces the `reference.html` file with a symlink to `reference-1.0.0.html` and updates a couple of links around.

Works locally, but needs to be tested in an OSX machine and a windows machine - symlinks might not play well in those platforms.


mourner:
Whoa, does git and gh pages support symlinks? Didn't know that.

jasonoverland:
+1. I recently noticed that I was looking at Leaflet 0.7.x docs because I followed a link from Google, and reference.html is the top result on Google for "leaflet documentation" (and other keywords).  Seems that reference.html should point to the latest, not 0.7.x.

I just checked out this branch on Windows and reference.html simply opens up as a file with "reference-1.0.0.html" as the contents. Windows does have some form of symlinks with the MKLINK command but git does not appear to be aware of it.  At least 'git status' shows everything as clean and not modified. I would imagine OS X shouldn't have any problems.

IvanSanchez:
superseded by #5490

PR #4977, opened 2016-09-29
Add quotes to css attribute value because of colon

peergynt:
Fix minor issue in css file. The quotes are consistent with the line above.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -522,7 +522,7 @@
 	margin: 0 auto;
 
 	-ms-filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678)";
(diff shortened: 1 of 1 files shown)


IvanSanchez:
No. That rule is specific for IE8, the above one specific for IE10. Unless there is some browser or linting tool that chokes on that CSS syntax, changing that rule could affect behaviour in IE.

See e.g. https://blogs.msdn.microsoft.com/ie/2009/02/19/the-css-corner-using-filters-in-ie8/ , https://stackoverflow.com/questions/7131356/filter-progiddximagetransform-microsoft-gradient-is-not-working-in-ie7#7136136

IvanSanchez:
Also, quoting from https://msdn.microsoft.com/en-us/library/ms530752(v=vs.85).aspx :

> When you use `-ms-filter`, enclose the progid in single quotes (`'`) or double quotes (`"`).

peergynt:
That's correct, this was a bad PR on my end.
The GWT InlineClientBundleGenerator was choking on this line (as well as a couple others) so I'll have to tweak these myself.

Issue #4849, opened 2016-08-25
leaflet.css and webpack css-loader


jmlopez-rod:
In the commit <https://github.com/Leaflet/Leaflet/commit/e9957cfa02b4201c62e98d43acc23acd5f3f9927>
there is the following change:

```
+.leaflet-default-icon-path {
+	background-image: url(images/);
+	}
```

But now, in webpack with the css-loader I get the following error:

```
ERROR in ./~/css-loader!./~/leaflet/dist/leaflet.css
Module not found: Error: Cannot resolve directory './images' in /Users/jmlopez/Workspace/ProjectName/node_modules/leaflet/dist
 @ ./~/css-loader!./~/leaflet/dist/leaflet.css 6:8715-8735
```

If I make it a proper url to a file then the compilation goes through. Anyone else having this issue? Should I bring this up with https://github.com/webpack/css-loader?

IvanSanchez:
Yeah, the method of detecting the image path in #4605 involves using a background image which is just a path, and doesn't exist.

It's a bit counter-intuitive, since we're using that just for path resolution and not actually for setting the background. Browsers don't even try to load the non-existent image, as there only element with that CSS class is destroyed before being used.

It's a fringe case, but we won¡t change any code when the incompatibility is due to a third party (it's not that we have something against webpack, but we ignore bugs related to R, Vue, React, Angular, Ionic, and so on and so on).

Maybe it's a good idea to make this a warning in Webpack instead of an error.

soyjavi:
IMHO, is better a warning... a lot of people uses 3rds like leaflet

nikolauskrismer:
Is there a workaround (maybe within webpack or by using arguments in argument given to require) that anybody knows of?
This prevents me from using leaflet 1.0.0 (since I use webpack)

IvanSanchez:
@nikolauskrismer Ask around the webpack folks and community. We'll be happy to apply a pull request to make webpack work nicer with Leaflet 1 **if** it doesn't break the functionality of marker icon path autodetection. None of the Leaflet team currently makes heavy use of webpack.

paralin:
I'm having this exact issue right now.

paralin:
[...]

PR #4979, opened 2016-09-29
Fix webpack using valid image file for default icon path (#4849)

jefbarn:


src/layer/marker/Icon.Default.js
```diff
@@ -42,8 +42,9 @@ L.Icon.Default = L.Icon.extend({
 		           L.DomUtil.getStyle(el, 'backgroundImage');	// IE8
 
 		document.body.removeChild(el);
+		console.log('marker path', path);
```
IvanSanchez: No `console.log()`s, plz

IvanSanchez:
Seems to work in FFX and phantom unit tests. Will make a couple tests in a couple more browsers tomorrow.

Looks good, except for the `console.log()`.

jefbarn:
Oops, forgot to pull that log message out, I'll fix it up.

jefbarn:
Grrr, I tried to fixup the commit messages, but just ended up making it worse.  I hate git sometimes.

mourner:
Let's try it out!

IvanSanchez:
Works for me. No problems on FFX, chrome nor ie11, tests pass.

lmcarreiro:
This pull request broke the case with webpack:

image

It's common to use `[name]-[hash].[ext]` with file-loader in webpack to avoid browser cache issues, the regex should be more flexible.

Issue #4912, opened 2016-09-14
class .leaflet-tile-loaded is added even in case of a loading error


Fil:
Class `leaflet-tile-loaded` is added even in case of a loading error.

(The bug seems to strike on Chrome not Firefox)

Demo:
http://bl.ocks.org/Fil/75b76aec31fa76060a0c872c68ae44da

Fix:
```
line 4583:
-		L.DomUtil.addClass(tile.el, 'leaflet-tile-loaded');
+		if (!err) L.DomUtil.addClass(tile.el, 'leaflet-tile-loaded');

```

![capture d ecran 2016-09-14 16 29 10](https://cloud.githubusercontent.com/assets/7001/18516326/ce73e79e-7a98-11e6-8deb-11f1a206bdee.png)

IvanSanchez:
So that code is around https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L814

IvanSanchez:
I think this is a regression - in 0.7.7, the `leaflet-tile-loaded` CSS class in only added on the `tileload` event, whereas it's not added on `tileerror`, see https://github.com/Leaflet/Leaflet/blob/v0.7.7/src/layer/tile/TileLayer.js#L566

We should keep in line with the 0.7.7 logic and not fire `tileload` events when there's a tile loading error, right?

@Fil Can you please try to make a pull request? A simple check wrapping the `addClass()` and `fire('tileload',...)` statements should suffice at this point. Anything else and we risk breaking the tile pruning logic.

Fil:
maybe just a `return;` here https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L797

IvanSanchez:
@Fil No, that has the potential of messing up the pruning algorithm if the last tile gives an error.

Fil:
Sorry I don't know enough Leaflet internals to help with this.

-- Fil

thachhoang:
@IvanSanchez By a simple check do you mean `if (!err)`? Should we add a test?

IvanSanchez:
@thachhoang Yes, just make sure that `src/tile/GridLayer`, around line 814, does *not* apply the `leaflet-tile-loaded` css class if there's been an error, and also it shouldn't fire the `tileload` event. The `_tileReady` method should look like:

```
if (err) {
    fire 'tileerror'
}

request tile fade-in

if (!err) {
[...]

PR #4980, opened 2016-09-29
BugFix: Dont load class .leaflet-tile-loaded when there's an error

jihdeh:
This PR help solve:

The issue on #4912 which still adds a class `.leaflet-tile-loaded` when there's an error.


IvanSanchez:
Looks good, thanks @jihdeh !

PR #4982, opened 2016-09-30
docstrings: L.GeoJSON filter and methods

ghybs:
Following https://stackoverflow.com/questions/39776919/how-to-reference-the-data-of-an-existing-geojson-leaflet-object/39787583#39787583, sounds like the current definition "decide whether to show a feature or not" implies that the data is retained in memory, but just not shown.

Replaced "show" by "include" to try to make it clearer that the filtered out data is _lost_ (not retained at all within the group).

Also took the opportunity to add a note about changing the `filter` option dynamically (no re-evaluation of child layers, whether they meet the new filter conditions or not).

Corrected docstrings of L.GeoJSON for methods (addData, resetStyle, setStyle) to be really counted as "methods" instead of static "functions" (geometryToLayer, coordsToLatLng, coordsToLatLngs, latLngToCoords, latLngsToCoords, asFeature)

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -78,7 +78,7 @@ L.GeoJSON = L.FeatureGroup.extend({
 		}
 	},
 
-	// @function addData( <GeoJSON> data ): Layer
+	// @method addData( <GeoJSON> data ): Layer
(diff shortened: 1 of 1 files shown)


mourner:
Great, thanks!

PR #4984, opened 2016-09-30
Bump v1.0.1

IvanSanchez:


mourner:
In the current form, we can't merge this PR because it assumes published 1.0.1 links in the live website, whereas to publish 1.0.1 to NPM, we need to update the versions first and merge to master first. Docs update and Leaflet source update must be separate PRs.

PR #4988, opened 2016-09-30
Add new class L.VideoOverlay

IvanSanchez:
I was looking at some other mapping APIs and became jealous of their ability to display videos. So, I added a tiny bit of functionality to the built-in `ImageOverlay`.

On top of that, I created a tutorial for that. I want to try and see if a "one feature, one tutorial" policy makes sense.


mourner:
This is great! The only thing I'm wondering is whether `L.imageOverlay(url, {video: true})` is the best API for this — it's not very intuitive. Would it be better to do `L.videoOverlay` that extends from `ImageOverlay`? After minor refactoring, this could also be more convenient to maintain.

IvanSanchez:
Yeah, having a `L.VideoOverlay` makes a lot of sense. Will refactor and rename later.

IvanSanchez:
@mourner OK, I refactored this into a `L.VideoOverlay` class, fixed a couple of bits of the documentation, made another example, add the new class to the dependency tree and the docs index.

PR #4989, opened 2016-09-30
ES6 modules & Rollup

### LINKED ISSUE #3229: Switch to a ES6 module system with Rollup
Have a version of leaflet which plays nice with SystemJs(and jspm) loader so we can import Leaflet components form ES6 code.
E.g.

```
import {Map, LatLng} form 'leaflet';
/*
var map = new Map({
  center: new LatLng(0,0)
})
*/
var map = Map({
  center: LatLng(0,0)
});
```

mourner:
This an in-progress PR that switches Leaflet to use ES6 modules system and Rollup for bundling. Closes #3229. It's A LOT of tedious work going through all the code, but it looks like it's going to be a major improvement.
- Dependencies will be imported explicitly in each file instead of depending on global namespaces, making the code easier to maintain and reason about.
- Things like `L.Util.bind` will turn into just `bind`, which means pretty much all references to Leaflet functions/classes will be properly minified, reducing the bundle size significantly (in theory).
- I love the syntax, it just comes along very naturally.
- We'll switch to a rollup-watch-based debug workflow, eliminating the `document.write` hack.
- People who use ES6 modules or Rollup in their apps will be able to benefit from tree shaking (import stuff they need directly and discard all the rest).

cc @IvanSanchez @patrickarlt @olanod @tmcw @perliedman @yohanboniface @hyperknot 

Status:
- [x] src/**
- [x] `publish.sh` script
- [x] Manually check `pointer` event handling hacks
- [x] Manually check `touch` event handling hacks
- [x] Cleanup old build script & custom builds system
- [x] Move unrunnable `L.Path` tests from `suites/layer/vector/PathSpec.js` to somewhere else
- [ ] Tweak build script to output file sizes
- [x] Rename output files from `leaflet-rollup-src.js` to `leaflet-src.js`
- [x] Wait until https://github.com/rollup/rollup/pull/1069 is fixed (IE8 is broken until then)
- [x] Rollup config/plugin to replace `L.version` to a string including revision


src/map/Map.js
```diff
@@ -1,3 +1,24 @@
...
+
+import {Evented} from '../core/Events';
+import {EPSG3857} from '../geo/crs/CRS.EPSG3857';
+import {Point, toPoint} from '../geometry/Point';
+import {Bounds, toBounds} from '../geometry/Bounds';
+import {LatLng, toLatLng} from '../geo/LatLng';
+import {LatLngBounds, toLatLngBounds} from '../geo/LatLngBounds';
+
+import {
+	any3d as isAny3D,
```
tmcw: at least within iD and Studio, we've tried to avoid `as` statements so that it's easier to track down code across the whole codebase. Minor nitpick, and not necessarily something you want to adopt, but I think it's worked well.

mourner: Yeah, I'd like to avoid `as` too, but the problem is that without namespaces, some of the exported names become awkward, and I wanted each file to export names that will end up in the namespace. See https://github.com/Leaflet/Leaflet/pull/4989/commits/a397b6385ed7eafa21da54471196657e448503c0 for how the `Browser` module looked, for example. Maybe we'll want to migrate to new names, but in a breaking way, in future.

src/geo/crs/CRS.Base.js
```diff
@@ -15,7 +15,7 @@ import {wrapNum} from '../../core/Util';
  * [Proj4Leaflet](https://github.com/kartena/Proj4Leaflet) plugin.
  */
 
-export var Base = {
+export var Base = extend({}, {
```
mourner: Why extend an empty object here here?

package.json
```diff
@@ -29,7 +29,8 @@
     "test": "jake test",
     "build": "jake build",
     "release": "./build/publish.sh",
-    "rollup": "rollup src/Leaflet.js -f umd -n L > bundle.js"
+    "lint": "node_modules/.bin/eslint --fix src/",
```
mourner: `node_modules/.bin` is not necessary in `package.json` scripts — it's in path automatically.

mourner: @IvanSanchez so `eslint` doesn't work but `rollup` and `jake` do? That's weird.

nathancahill:
I have the same experience. ./node_modules/.bin has to be specified to take priority over globally installed executables.

package.json
IvanSanchez: In my case, I put `rollup` and `jake` **manually** in my `$PATH` because I was tired of running them locally (but they have to be updated/upgraded separately). Not so with eslint.

build/rollup-config.js
```diff
@@ -0,0 +1,33 @@
+
+// Config file for running Rollup in "normal" mode (non-watch)
+
+// Needed to create files loadable by IE8
+import es3 from 'rollup-plugin-es3'
+import json from 'rollup-plugin-json'
+
+import gitRev from 'git-rev-sync'
+
+// TODO: There should be a way to skip the git branch+rev when doing a production build
```
Scarysize: `process.env.NODE_ENV` or a rollup production configuration would do the trick.

IvanSanchez: Yeah, but I'll wait for @mourner's input on the issue. This will affect the `build/publish.sh` script.

mourner: Will it? As far as I see, `gitRev` is only needed for the `L.version` property, and `publish.sh` takes the clean semver version from `package.json`.

IvanSanchez:
I tried to make some clever use of `rollup` in the browser, but hit a lot of roadblocks. Thus, rely on `npm run-script watch` or `yarn run watch` for the time being. The script includes a hack in the JS banner so devs are reminded to use a fresh bundle.

src/Leaflet.js
```diff
@@ -30,7 +31,7 @@ export {Class} from './core/Class';
 
 import {Evented} from './core/Events';
 export {Evented};
-export var Mixin = {Events: Evented.prototype};
+// export var Mixin = {Events: Evented.prototype};
```
IvanSanchez: @mourner I was not aware this was a backwards-compat issue, sorry 'bout that :-(

build/rollup-config.js
mourner: We can either set an env var (read from the config), or just leave git rev in the copyright and `L.version` — no big deal.

mourner:
Let's rebase to fix the Canvas conflict, and also maybe ditch all uses of `as` in favor of importing namespaces (which, as @IvanSanchez showed, are still neatly unrolled by rollup).

IvanSanchez:
@mourner Rebased. I'll take a break from this branch, and will let someone else rewrite the named `DomUtil`, `Util` and `DomEvent` imports.

mourner:
@IvanSanchez BTW what's up with the unrunnable Path tests?

IvanSanchez:
@mourner `L.Path` is no longer exported (because it should be something like an un-instantiable abstract class), thus the tests involving `new L.Path` cannot work.

Maybe we should export `L.Path` as well, but I'm unsure.

IvanSanchez:
After some rebasing hell and some fiddling with the VML code, the IE8 builds work again.

docs/index.html
```diff
@@ -5,7 +5,7 @@
 <div class="announcement">Jan 23, 2017 — <a href="http://leafletjs.com/2017/01/23/leaflet-1.0.3.html">Leaflet 1.0.3</a>, a bugfix release, is out.</div>
 
 <p>Leaflet is the leading open-source JavaScript library for mobile-friendly interactive maps.
-Weighing just about <abbr title="33 KB gzipped &mdash; that's 123 KB minified and 218 KB in the source form, with 10 KB of CSS (2 KB gzipped) and 11 KB of images.">33 KB of JS</abbr>,
+Weighing just about <abbr title="38 KB gzipped &mdash; that's 133 KB minified and 378 KB in the source form, with 10 KB of CSS (2 KB gzipped) and 11 KB of images.">38 KB of JS</abbr>,
```
mourner: Whoa, is `leaflet-src.js` really 378kb?

IvanSanchez: Yes, because we added docstrings to *everything*, and that kinda doubles the size of the file.

mourner: Oh, I forgot about that :)

perliedman:
```
per@linfro:~/Documents/Projects/Leaflet$ npm run build

> leaflet@1.0.3 build /home/per/Documents/Projects/Leaflet
> jake build


Bundling and compressing 67 files...
	Uncompressed: 372.64 KB (new)
	Saved to dist/leaflet-src.js
	Compressed: 141.92 KB (new)
	Saved to dist/leaflet.js
	Gzipped: 37.97 KB
```

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+control.layers = layers;
+control.zoom = zoom;
+control.scale = scale;
+control.attribution = attribution;
 
-// define Leaflet as an AMD module
-} else if (typeof define === 'function' && define.amd) {
-	define(L);
-}
+export {Control, control};
```
simon04: Adding a file `./control/index.js` and doing the preparation of `Control`/`control` there would streamline the code in `Leaflet.js` down to 
```javascript
export {Control, control} from './control';
```

build/build.js
```diff
@@ -1,237 +0,0 @@
...
-    return (bytes / 1024).toFixed(2) + ' KB';
-};
-
-exports.build = function (callback, version, compsBase32, buildName) {
-
-	var files = getFiles(compsBase32);
-
-	console.log('Bundling and compressing ' + files.length + ' files...');
-
-	var copy = fs.readFileSync('src/copyright.js', 'utf8').replace('{VERSION}', version),
```
simon04: The file `copyright.js` can be removed.

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+import {Keyboard} from './map/handler/Map.Keyboard';
+Map.Keyboard = Keyboard;
+import {ScrollWheelZoom} from './map/handler/Map.ScrollWheelZoom';
+Map.ScrollWheelZoom = ScrollWheelZoom;
+import {Tap} from './map/handler/Map.Tap';
+Map.Tap = Tap;
+import {TouchZoom} from './map/handler/Map.TouchZoom';
+Map.TouchZoom = TouchZoom;
+
+export {Map, createMap as map} from './map/Map';
```
simon04: Adding a file `./map/index.js` and doing the preparation of `Map` there would streamline the code in `Leaflet.js` down to 
```javascript
export {Map, map} from './map';
```

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+export {DivIcon, divIcon} from './layer/marker/DivIcon';
+export {Marker, marker} from './layer/marker/Marker';
+
+// layer, tile
+export {GridLayer, gridLayer} from './layer/tile/GridLayer';
+import {TileLayer, tileLayer} from './layer/tile/TileLayer';
+import {TileLayerWMS, tileLayerWMS} from './layer/tile/TileLayer.WMS';
+TileLayer.WMS = TileLayerWMS;
+tileLayer.wms = tileLayerWMS;
+export {TileLayer, tileLayer};
```
simon04: Adding a file `./layer/tile/index.js` and doing the preparation of `TileLayer`/`tileLayer` there would streamline the code in `Leaflet.js` down to 
```javascript
export {GridLayer, gridLayer, TileLayer, tileLayer} from './layer/tile';
```

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+export {FeatureGroup, featureGroup} from './layer/FeatureGroup';
+import {GeoJSON, geoJSON, geoJson, geometryToLayer, coordsToLatLng, coordsToLatLngs, latLngToCoords, latLngsToCoords, getFeature, asFeature} from './layer/GeoJSON';
+GeoJSON.geometryToLayer = geometryToLayer;
+GeoJSON.coordsToLatLng = coordsToLatLng;
+GeoJSON.coordsToLatLngs = coordsToLatLngs;
+GeoJSON.latLngToCoords = latLngToCoords;
+GeoJSON.latLngsToCoords = latLngsToCoords;
+GeoJSON.getFeature = getFeature;
+GeoJSON.asFeature = asFeature;
+export {GeoJSON, geoJSON, geoJson};
```
simon04: Maybe add the static functions to the GeoJSON class in `./layer/GeoJSON.js`? Or prepare and export a `GeoJSON_with_static_functions` object in `./layer/GeoJSON.js`?

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+import {Simple} from './geo/crs/CRS.Simple';
+
+CRS.Earth = Earth;
+CRS.EPSG3395 = EPSG3395;
+CRS.EPSG3857 = EPSG3857;
+CRS.EPSG900913 = EPSG900913;
+CRS.EPSG4326 = EPSG4326;
+CRS.Simple = Simple;
+
+export {CRS};
```
simon04: Adding a file `./geo/crs/index.js` and doing the preparation of `CRS` there would streamline the code in `Leaflet.js` down to 
```javascript
export {CRS} from './geo/crs';
```

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+export {Transformation} from './geometry/Transformation';
+
+// geo
+
+export {LatLng, toLatLng as latLng} from './geo/LatLng';
+export {LatLngBounds, toLatLngBounds as latLngBounds} from './geo/LatLngBounds';
+
+// geo/projection
+
+import * as Projection from './geo/projection/Projection';
```
simon04: Rename `Projection.js` to `index.js` to comply with other suggestions and the [node.js module import](https://nodejs.org/api/modules.html).

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+export {Tooltip, tooltip} from './layer/Tooltip';
+
+import {Icon} from './layer/marker/Icon';
+export {icon} from './layer/marker/Icon';
+import {IconDefault} from './layer/marker/Icon.Default';
+Icon.Default = IconDefault;
+export {Icon};
+
+export {DivIcon, divIcon} from './layer/marker/DivIcon';
+export {Marker, marker} from './layer/marker/Marker';
```
simon04: Combine all markers in `./layer/marker/index.js` and export them here all at once:
```javascript
export{Icon, icon, DivIcon, divIcon, Marker, marker} from './layer/marker';
```

src/Leaflet.js
```diff
@@ -1,29 +1,170 @@
...
+export {Canvas, canvas} from './layer/vector/Canvas';
+export {SVG, svg} from './layer/vector/SVG';
+import './layer/vector/Renderer.getRenderer';	// This is a bit of a hack, but needed because circular dependencies
+
+export {Path} from './layer/vector/Path';
+export {CircleMarker, circleMarker} from './layer/vector/CircleMarker';
+export {Circle, circle} from './layer/vector/Circle';
+export {Polyline, polyline} from './layer/vector/Polyline';
+export {Polygon, polygon} from './layer/vector/Polygon';
+export {Rectangle, rectangle} from './layer/vector/Rectangle';
```
simon04: Combine all vector layer classes in `./layer/vector/index.js` and export them here all at once:
```javascript
export{Renderer, Canvas, canvas, SVG, svg, Path, CircleMarker, circleMarker, Circle, circle, Polyline, polyline, Polygon polygon, Rectange rectangle} from './layer/vector';
```

IvanSanchez:
@simon04 You make a few good points! As far as I'm aware, Rollup (*et al*) can tree-shake better if the modules are not assigned, but just imported/exported.

However, right now I think it'd be better to just merge this as it is, and worry about cleaner ES6 module exports later.

As long as the ES5 bundle works exactly the same, all that can be postponed.

PR #4992, opened 2016-10-01
Better description

Jmuccigr:
Replaced awkward text with easier to understand version.


mourner:
This file is autogenerated. You need to update the text in the source files (search).

PR #4994, opened 2016-10-02
Better description

Jmuccigr:
Language improvement (hopefully to the right file this time).


IvanSanchez:
Yes, you got the right file this time :-)

Remember that you can run `npm install`, then `jake docs` to see how the documentation will look. We usually wait until the next release to update the published documentation files.

Jmuccigr:
Thanks for the tip. I haven't installed it locally yet, but may as I'm starting to use it a lot more.

PR #4995, opened 2016-10-02
Normalize line endings to UNIX-style

fbonzon:
There is currently a mix of Windows-style (CR-LF) and Unix-style (LF) line endings in the Leaflet repository.

This PR makes all text files consistently use Unix-style (LF) line endings.

The inconsistent EOL-style is particularly problematic for the file `/dist/leaflet-src.js`, constructed from different source files, and which produces a mix of CRLF and LF line endings within the same file.


mourner:
I'm not willing to merge a PR like this with 16k line changes because it will wreak havoc on existing PRs, make `git blame` worse, etc. Also, I never experienced problems because of mixed line endings. They will eventually get replaced but not like this.

PR #4997, opened 2016-10-02
Shift+scroll wheel zoom

### LINKED ISSUE #3521: Shift + scroll wheel zoom

yamov:
This pull request is supposed to resolve the "Shift + scroll wheel zoom" issue (fixes #3521).

When I started resolving it I found out that in the current version of Leaflet holding the SHIFT key and scrolling the mouse wheel produces no effect and does not zoom a map. Strange thing was that it was working in Firefox and was not in Chrome and Opera. After some research I found out that Chrome and Opera have a feature  - holding the SHIFT key pressed transforms vertical scrolling into horizontal. And horizontal scrolling was explicitly ignored by Leaflet. So, I thought it would be a good idea to treat vertical and horizontal wheel scrolling in a similar manner for the sake of resolving this issue and for the sake of consistency as well.

The current version of Leaflet already provides the SHIFT + zoom feature for zoom buttons and key presses (+/-). However, the zoomDetlata multiplier is hard-coded and I don't think it is a really good idea.
My proposal is to extend the API and implement it as a separate map option (I came up with the name "zoomDeltaFactor"). Implementing it as a separate map option makes it flexible (the default value is 3 but developers can change it depending on situation) and explicit.

So, the summary of this PR:
- The getWheelDelta method of the L.DomEvent object has been changed to treat horizontal and vertical scrolling similarly,
- Holding the Shift key while scrollwheel-zooming now multiplies the zoom delta by 3 (by default),
- The API has been extended and provides an option to change the zoom delta multiplier.

_PS. This is my first attempt to contribute to an open source project. So, if I am doing it wrong, please, be kind and guide me to the right path.
Thanks ;)_

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -63,13 +63,13 @@ L.Control.Zoom = L.Control.extend({
 
 	_zoomIn: function (e) {
 		if (!this._disabled && this._map._zoom < this._map.getMaxZoom()) {
-			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? this._map.options.zoomDeltaFactor : 1));
 		}
 	},
 
 	_zoomOut: function (e) {
 		if (!this._disabled && this._map._zoom > this._map.getMinZoom()) {
-			this._map.zoomOut(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomOut(this._map.options.zoomDelta * (e.shiftKey ? this._map.options.zoomDeltaFactor : 1));
 		}
 	},
 
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -237,12 +237,11 @@ L.DomEvent = {
 	// Events from pointing devices without precise scrolling are mapped to
 	// a best guess of 60 pixels.
 	getWheelDelta: function (e) {
-		return (L.Browser.edge) ? e.wheelDeltaY / 2 : // Don't trust window-geometry-based delta
-		       (e.deltaY && e.deltaMode === 0) ? -e.deltaY / L.DomEvent._wheelPxFactor : // Pixels
-		       (e.deltaY && e.deltaMode === 1) ? -e.deltaY * 20 : // Lines
-		       (e.deltaY && e.deltaMode === 2) ? -e.deltaY * 60 : // Pages
-		       (e.deltaX || e.deltaZ) ? 0 :	// Skip horizontal/depth wheel events
-		       e.wheelDelta ? (e.wheelDeltaY || e.wheelDelta) / 2 : // Legacy IE pixels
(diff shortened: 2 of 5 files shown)


IvanSanchez:
> *This is my first attempt to contribute to an open source project. So, if I am doing it wrong, please, be kind and guide me to the right path.*

I think the PR is quite good, actually :smile: 

But.

You're forgetting a use case: laptop trackpads which fire tiny horizontal scroll events when scrolling vertically. That's the reason why I implemented https://github.com/Leaflet/Leaflet/commit/fa3da6503fbe9ef0d7b2e2461b165f9d2821590a long ago in 1.0.0rc1. I think this might happen on some Mac trackpads/magic mice as well.

So if the shift key is **not pressed**, you've got to check for **vertical** scroll **only**, and if the shift key is **pressed**, you've got to check for horizontal scroll **only**.

And then there's the question of what should be the right behaviour when pressing shift and doing a vertical/horizontal scroll in such a trackpad.

And then, this has to be tested and double-tested on multiple browsers, and on *multiple hardware*.

Your code looks good, the way you're implementing the feature looks good... but until we can be sure this won't break behaviour in those weird trackpads, we cannot merge the code.

And even so, I'd be more worried about fixing https://github.com/Leaflet/Leaflet/issues/4410 first. Mac touchpads are way more common than users shift-scrolling horizontally, I'm afraid.

yamov:
@IvanSanchez, thank you very mach for the reply.

In fact, I did not forget about the use case you have described. I spent some time testing it and the thing is I actually was not able to get the scrolling in two directions at the same time, not even when scrolling diagonally. I guess, modern touchpads are quite smart and can guess your real intended scroll direction and ignore the unintended direction.

I was testing it with the following code snippet:
```
document.addEventListener('wheel', function(e) {
    console.log(e.deltaY);
    console.log(e.deltaX);
});
```
And I was not able to scroll in two directions simultaneously. It was always either vertical or horizontal (i.e. one of the number was always equal to 0).

However, I totally agree with your point that it should be tested in multiple browsers and on multiple hardware. Currently I have only tested it on the DELL laptop touchpad in Win 10 (all major browsers) and in Xubuntu 14.04 (Chrome and FF; the OS is running in virtual machine, though).
Unfortunately, right now I don't have the opportunity to test in OS X on my iMac using Magic Mouse (I have just moved to another country and the iMac has not yet arrived). But I will definitely do that as soon as I can.

On the other hand, I wonder if it would be a problem even if under some configuration a user would get a 2-directional scroll using a touchpad. In my code I use (e.deltaY || e.deltaX), so the vertical scrolling is checked first (which probably is the user's intention in 99% of the cases) and tiny horizontal scrolls are ignored anyway. Are they not? This way we only check for horizontal scrolling when there was no vertical scrolling at all.

What do you think about it?

stell:
Is Shift+scrolling still a topic or not of relevance anymore?

PR #5002, opened 2016-10-03
Use rollup-plugin-buble and use shorthand ES6 class methods

IvanSanchez:
This uses Rich Harris' [Bublé](https://buble.surge.sh/guide/) in order to use "Object shorthand methods and properties", so instead of

```
var foo = L.Class.extend({
    method: function(par1, par2) {
         ...
    }
});
```

One can write:

```
var foo = L.Class.extend({
    method(par1, par2) {
         ...
    }
});
```

...thus saving a few bytes on the ES6 code.

We might also want to use things like default values for function parameters, arrow functions, and other things that transpile well.

cc @mourner

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "phantomjs-prebuilt": "^2.1.12",
     "prosthetic-hand": "^1.3.1",
     "rollup": "^0.36.1",
+    "rollup-plugin-buble": "^0.14.0",
     "source-map": "^0.5.6",
     "uglify-js": "~2.7.3"
   },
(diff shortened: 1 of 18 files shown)


IvanSanchez:
Note this targets `rollup2`, not `master`

mourner:
Buble is pretty cool! I don't see why we wouldn't want to use it (it's pretty much free if we're already using rollup), but I'd probably want to hold on upgrading to ES6-isms until we land ES6 modules to avoid merge complications.

IvanSanchez:
Sounds like a sane thing to do. I'll close down this PR, but let's revisit this branch when #4989 is done.

PR #5010, opened 2016-10-10
Fix doc: add clarification for projection; closes #4817

### LINKED ISSUE #4817: SphericalMercator.project() does not accept [lat,lng] array
The documentation for LatLng in Leaflet 0.7.x indicates that:

> All Leaflet methods that accept LatLng objects also accept them in a simple Array form and simple object form (unless noted otherwise)...

However, I was getting NaN results when using this form.  In examining the code it appears that that particular call (on line 11 of the Projection.SphericalMercator.js ) assumes the latlng passed in is a LatLng object.  Checked the latest source in master and it appears this is still the case.

Workaround is obvious, but implementation does seem inconsistent with the documentation.

I have not checked the other projections for this bug.

erickzhao:
Adds clarification in the documentation that specifies that the`project` and `unproject` methods cannot take in arrays, but only real `L.point` instances.

This is my first time trying to add a fix to an open-source project, so I hope I got the workflow down okay. I have used Leaflet in the past, but never actually touched this specific method. I hope I didn't misunderstand the Issue.

Closes #4817

### DIFF
--- a/src/geo/projection/Projection.leafdoc
+++ b/src/geo/projection/Projection.leafdoc
@@ -8,9 +8,9 @@ a flat surface (and back). See [Map projection](http://en.wikipedia.org/wiki/Map
 The bounds where the projection is valid
 
 @method project(latlng: LatLng): Point
-Projects geographical coordinates into a 2D point.
+Projects geographical coordinates into a 2D point. Only accepts actual `L.LatLng` instances, not arrays.
 
 @method unproject(point: Point): LatLng
-The inverse of `project`. Projects a 2D point into a geographical location.
+The inverse of `project`. Projects a 2D point into a geographical location. Only accepts actual `L.Point` instances, not arrays.
 
 
--- a/src/geometry/Transformation.js
+++ b/src/geometry/Transformation.js
@@ -29,7 +29,7 @@ L.Transformation = function (a, b, c, d) {
 L.Transformation.prototype = {
 	// @method transform(point: Point, scale?: Number): Point
 	// Returns a transformed point, optionally multiplied by the given scale.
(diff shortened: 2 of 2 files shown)


src/geo/projection/Projection.leafdoc
```diff
@@ -8,9 +8,9 @@ a flat surface (and back). See [Map projection](http://en.wikipedia.org/wiki/Map
 The bounds where the projection is valid
 
 @method project(latlng: LatLng): Point
-Projects geographical coordinates into a 2D point.
+Projects geographical coordinates into a 2D point. Only accepts real `L.Point` instances, not arrays.
```
IvanSanchez: You mean "real `L.LatLng` instances" instead of `L.Point`, right?

IvanSanchez:
@erickzhao Workflow looks fine ;-)

Just be aware of one typo I commented. Also, I'd like to use the word "actual" instead of "real", just to avoid confusion about coordinates using real numbers.

Can you fix those and push another commit?

erickzhao:
@IvanSanchez 

Just updated the issues and force pushed to my fork! I ran `jake docs` and it seems to look just fine on my end.

screen shot 2016-10-11 at 8 59 30 am

Just another note on "real" vs "actual", the `transform` and `untransform` methods also both say "Only accepts real L.Point instances, not arrays". Should that be changed to "actual" as well?

IvanSanchez:
Looking good so far!

Yeah, feel free to change the text on `transform` and `untransform` too.

erickzhao:
Updated! PTAL but I think it should be all good!

IvanSanchez:
Looks good, thanks!!!

PR #5013, opened 2016-10-12
WIP: ES6-ify more stuff

IvanSanchez:
This is just a WIP, which applies the @mourner-style ES6fications (#4989) to: 

* Marker
* Icon
* DivIcon
* DomUtil
* Draggable
* GridLayer
* TileLayer
* Path
* Popup (was not fully ES6ified)
* All map interaction handlers

Note this targets the `rollup2` branch.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -231,6 +231,7 @@ export var disableImageDrag;
 export var enableImageDrag;
 export var preventOutline;
 export var restoreOutline;
+let _outlineElement, _outlineStyle, _userSelect;
 
 (function () {
 	// prefix style property names
@@ -275,14 +276,14 @@ export var restoreOutline;
 		disableTextSelection = function () {
 			if (userSelectProperty) {
 				var style = document.documentElement.style;
-				this._userSelect = style[userSelectProperty];
+				_userSelect = style[userSelectProperty];
 				style[userSelectProperty] = 'none';
 			}
 		};
 		enableTextSelection = function () {
 			if (userSelectProperty) {
-				document.documentElement.style[userSelectProperty] = this._userSelect;
-				delete this._userSelect;
+				document.documentElement.style[userSelectProperty] = _userSelect;
+				_userSelect = undefined;
 			}
 		};
 	}
@@ -311,19 +312,19 @@ export var restoreOutline;
 		}
 		if (!element || !element.style) { return; }
 		restoreOutline();
-		this._outlineElement = element;
-		this._outlineStyle = element.style.outline;
+		_outlineElement = element;
+		_outlineStyle = element.style.outline;
 		element.style.outline = 'none';
-		on(window, 'keydown', restoreOutline, this);
+		on(window, 'keydown', restoreOutline);
 	};
 
 	// @function restoreOutline()
 	// Cancels the effects of a previous [`L.DomUtil.preventOutline`]().
 	restoreOutline = function () {
-		if (!this._outlineElement) { return; }
-		this._outlineElement.style.outline = this._outlineStyle;
-		delete this._outlineElement;
-		delete this._outlineStyle;
-		off(window, 'keydown', restoreOutline, this);
+		if (!_outlineElement) { return; }
+		_outlineElement.style.outline = _outlineStyle;
+		_outlineElement = undefined;
+		_outlineStyle = undefined;
+		off(window, 'keydown', restoreOutline);
 	};
 })();
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -1,3 +1,5 @@
+import {on, off, preventDefault} from './DomEvent';
+
 /*
  * @namespace DomUtil
  *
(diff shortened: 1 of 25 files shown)


src/dom/Draggable.js
```diff
@@ -1,3 +1,12 @@
+import {Evented} from '../core/Events';
+import {touch} from '../core/Browser';
+import {on, off, preventDefault} from '../dom/DomEvent';
```
mourner: We might want to do `on as onDOMEvent` or similar (I've used this earlier in rollup2 branch), because `on` without a namespace is too generic.

src/layer/marker/DivIcon.js
```diff
@@ -1,3 +1,6 @@
+import {Icon} from './Icon';
+import {toPoint as point} from '../../geometry/Point';
```
mourner: And here, `as` is not needed — `toPoint` is descriptive and doesn't conflict with `point` variables.

src/dom/DomUtil.js
```diff
@@ -9,226 +11,233 @@
...
-	// @function get(id: String|HTMLElement): HTMLElement
-	// Returns an element given its DOM id, or returns the element itself
-	// if it was passed directly.
-	get: function (id) {
-		return typeof id === 'string' ? document.getElementById(id) : id;
-	},
+// @function get(id: String|HTMLElement): HTMLElement
+// Returns an element given its DOM id, or returns the element itself
+// if it was passed directly.
+export function get (id) {
```
IvanSanchez: `eslint --fix` to the rescue!

src/dom/Draggable.js
IvanSanchez: I'm not really convinced on the aliases for that, but I agree it might be confusing. Will take no action on this, maybe I'll change my mind in a later iteration.

IvanSanchez:
@mourner Regarding `L.DomEvent.on()` *et al*, I think the solution is to do something like:

```js
import * as DomEvent from './DomEvent.js';

DomEvent.on(...);
```

This will be tree-shaken properly, see http://rollupjs.org/?version=0.36.3&shareable=JTdCJTIyb3B0aW9ucyUyMiUzQSU3QiUyMmZvcm1hdCUyMiUzQSUyMmNqcyUyMiUyQyUyMm1vZHVsZU5hbWUlMjIlM0ElMjJteUJ1bmRsZSUyMiUyQyUyMmdsb2JhbHMlMjIlM0ElN0IlN0QlN0QlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwKiUyMGFzJTIwRG9tRXZlbnQlMjBmcm9tJTIwJy4lMkZEb21FdmVudC5qcyclM0IlNUNuJTVDbkRvbUV2ZW50Lm9uKCUyMHdpbmRvdyUyQyUyMCdyZWxvYWQnJTJDJTIwZnVuY3Rpb24oKSUyMCU3QiU3RCklM0IlMjIlN0QlMkMlN0IlMjJuYW1lJTIyJTNBJTIyRG9tRXZlbnQuanMlMjIlMkMlMjJjb2RlJTIyJTNBJTIyZXhwb3J0JTIwZnVuY3Rpb24lMjBvbiUyMCglMjBlbCUyQyUyMHR5cGUlMkMlMjBmbiUyMCklMjAlN0IlNUNuJTVDdGVsLmFkZEV2ZW50TGlzdGVuZXIodHlwZSUyQyUyMGZuKSUzQiU1Q24lN0QlNUNuJTVDbmV4cG9ydCUyMGZ1bmN0aW9uJTIwb2ZmJTIwKCUyMGVsJTJDJTIwdHlwZSUyQyUyMGZuJTIwKSUyMCU3QiU1Q24lNUN0ZWwucmVtb3ZlRXZlbnRMaXN0ZW5lcih0eXBlJTJDJTIwZm4pJTNCJTVDbiU3RCUyMiU3RCU1RCU3RA==

mourner:
@IvanSanchez yes, the only reason to prefer importing specific methods rather than as a namespace is that it will be uglified better.

src/layer/marker/Icon.Default.js
```diff
@@ -29,15 +28,15 @@ export var IconDefault = Icon.extend({
 	},
 
 	_getIconUrl: function (name) {
-		if (!this.prototype.imagePath) {	// Deprecated, backwards-compatibility only
-			this.prototype.imagePath = this._detectIconPath();
+		if (!this.__proto__.imagePath) {	// Deprecated, backwards-compatibility only
+			this.__proto__.imagePath = this._detectIconPath();
```
mourner: `__proto__` is IE11+, we can't use it.

IvanSanchez:
@mourner Tree-shaking will be applied before uglifying, and rollup flattens the namespaced functions, so (right now) I don't see any benefit about not importing whole namespaces.

Maybe this happens in a platform other than `rollup`?

mourner:
@IvanSanchez Oh wow, I didn't know Rollup flattened namespaces! Awesome!

mourner:
@IvanSanchez let's work in the `rollup2` branch now.

PR #5019, opened 2016-10-15
Add leaflet-corridor to the list of plugins in docs

mikhailshilkov:


jieter:
[link to demo](http://mikhail.io/demos/leaflet-corridor/).

Note that Leaflets `L.Circle` [now uses an option instead of an argument](https://github.com/Leaflet/Leaflet/issues/3785), which might be an API choice you want to follow too.

mikhailshilkov:
Updated plugin API to move corridor parameter into options, as @jieter suggested

PR #5021, opened 2016-10-17
Remove duplicated test in Draggable

yohanboniface:
We already are testing for this._enabled some lines before.


perliedman:
:+1:

PR #5022, opened 2016-10-17
Fix language

Jmuccigr:
1. Improve some of the language to be more readable
2. Correct reference to night style base map, which isn't actually used
3. Add detail that it's possible to style the text in the control


mourner:
A much-needed cleanup, thanks for taking on this!

PR #5024, opened 2016-10-19
Fix #5011 (canvas remove+add)

### LINKED ISSUE #5011: CircleMarker not showing when setting `preferCanvas` to true in map option
- I'm reporting a bug, not asking for help
- I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
- I've searched through the issues to make sure it's not yet reported

----

## How to reproduce

- Leaflet version I'm using: 1.0.0
- Leaflet plugins used: proj4leaflet v0.7.2 and leaflet.markercluster 1.0.0
- Browser (with version) I'm using: Chrome (but this can be seen in Mozilla Firefox and Edge)
- step 1 check link in minimal example
- step 2 use the scroll or zoom in/out buttons

 - Expected behaviour:  the map zooms in and the object geometries are visible at each zoom level (if not clustered);
 - Actual behaviour: the map zooms in/out but the geometries disappear (totally or partially) when zooming in / out (the clustering functionality still works)

- using leaflet + marker cluster + svg [ works as expected ]: https://codepen.io/anon/pen/YGLbGq?editors=1010
- using leaflet + marker cluster + canvas [ fails to work ]: https://codepen.io/anon/pen/PGZONy?editors=1010
- using leaflet + canvas [ no clustering works as expected]: https://codepen.io/anon/pen/wzjbgP?editors=1010

ghybs:
Bug when using map option `preferCanvas: true` (canvas renderer for paths) and removing then adding back a vector within the same animation frame (typically in the same sequence, as done by Leaflet.markercluster at `"zoomend"` event, to remove layers and clusters outside visible bounds).

This commit clears the `_removed` flag from paths when they are added to the canvas.

Also added corresponding test suites.

Note: Leaflet #5011 was moved to https://github.com/Leaflet/Leaflet.markercluster/issues/727 due to bug being initially visible only with MCG plugin.

Demo of result: http://playground-leaflet.rhcloud.com/waw/1/edit?html,output

### DIFF
--- a/spec/suites/layer/vector/CanvasSpec.js
+++ b/spec/suites/layer/vector/CanvasSpec.js
@@ -157,4 +157,36 @@ describe('Canvas', function () {
 		});
 	});
 
+	it('removes vector on next animation frame', function (done) {
+		var layer = L.circle([0, 0]).addTo(map),
+		    layerId = L.stamp(layer),
+		    canvas = map.getRenderer(layer);
+
+		expect(canvas._layers.hasOwnProperty(layerId)).to.be(true);
+
+		map.removeLayer(layer);
+		// Defer check due to how Canvas renderer manages layer removal.
+		L.Util.requestAnimFrame(function () {
+			expect(canvas._layers.hasOwnProperty(layerId)).to.be(false);
+			done();
+		}, this);
+	});
+
+	it('adds vectors even if they have been removed just before', function (done) {
+		var layer = L.circle([0, 0]).addTo(map),
+		    layerId = L.stamp(layer),
+		    canvas = map.getRenderer(layer);
+
+		expect(canvas._layers.hasOwnProperty(layerId)).to.be(true);
+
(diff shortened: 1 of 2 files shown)


ghybs:
Thinking about the flag, it does not work if the layer is _moved_ from one pane (therefore renderer) to another (e.g. if developer wants to dynamically change z-index), because the flag is attached to the layer, and we cannot distinguish if the canvas is the same or not.

This is a problem for current version and even with this PR.

To cover such a use case, we would need either to:
- store a temporary array / hash map of the layers to be removed (like `needsRemoving` in Leaflet.markercluster), within each canvas, or
- rename the flag to make it specific to each canvas (e.g. `["_removed" + canvas._leaflet_id]`), so that each one operates independently.

mourner:
Thanks! Let's follow-up on the remaining rare case.

Issue #4968, opened 2016-09-28
L.Icon.Default brings a wrong image url


ArminMueller:
- [ x] I'm reporting a bug, not asking for help
- [ ] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
- [x ] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [ ] I've searched through the issues to make sure it's not yet reported

The image url leaflet prsents me is https://uismedia.geo-info-manager.com/apis/leaflet_1/imagesmarker-icon-2x.png. It seems there is a missing "/"
Additionally i have an error 
leaflet.min.js:5 GET https://uismedia.geo-info-manager.com/apis/leaflet_1/images/ 403 (Forbidden)

IvanSanchez:
* Is there any public webpage in your server we can visit, so we can reproduce the problem ourselves?
* What OS and web browser are you using?

jasongrout:
> leaflet.min.js:5 GET https://uismedia.geo-info-manager.com/apis/leaflet_1/images/ 403 (Forbidden)

This may be from the same issues discussed in https://github.com/Leaflet/Leaflet/issues/4849

IvanSanchez:
Indeed. Which why I'm curious as to know the circumstances this can be reproduced, so we can make unit tests against those.

Greigrm:
I had the same issue moving from RC3 to 1.0.1 - in my code I had the line L.Icon.Default.imagePath = 'images';  - can't remember quite why this was, but commenting it out solved the issue - might be worth checking you don't have similar

codeofsumit:
suddenly got the same issue in two completely different projects, both using webpack and leaflet.
If I add markers to the map, the images aren't found. Browser throws this error:
![image](https://cloud.githubusercontent.com/assets/2399810/19414064/38e62bca-9342-11e6-97ce-b03df80ede82.png)

codeofsumit:
![image](https://cloud.githubusercontent.com/assets/2399810/19419501/47d92578-93d9-11e6-8860-6ab6415406c3.png)
seems like webpack is loading the image path correctly from the CSS file, but then there is some inline stuff going on and the name of the icon is added to the image path. And that isn't correct. Still investigating.

codeofsumit:
ok I got something.

So a marker looks like this currently because the images (icon and shadow) aren't found.
[...]

PR #5041, opened 2016-10-23
#4968 Fix path for data URLS

Radu-Filip:
related to https://github.com/Leaflet/Leaflet/issues/4968

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -33,7 +33,8 @@ L.Icon.Default = L.Icon.extend({
 		// `L.Icon.Default` will try to auto-detect the absolute location of the
 		// blue icon images. If you are placing these images in a non-standard
 		// way, set this option to point to the right absolute path.
-		return (this.options.imagePath || L.Icon.Default.imagePath) + L.Icon.prototype._getIconUrl.call(this, name);
+		var path = this.options.imagePath || L.Icon.Default.imagePath;
+		return path.indexOf("data:") === 0 ? path : path + L.Icon.prototype._getIconUrl.call(this, name);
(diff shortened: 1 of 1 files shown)


codeofsumit:
with this PR, the marker looks like this for me
image

The shadow image has the same data:image url as the marker icon.

deeg:
@mourner, is anything needed to get this PR merged? This is causing issues for people who are using webpack.

Seems like a pretty straightforward fix that is worth merging.

src/layer/marker/Icon.Default.js
```diff
@@ -25,26 +25,26 @@ L.Icon.Default = L.Icon.extend({
 	},
 
 	_getIconUrl: function (name) {
-		if (!L.Icon.Default.imagePath) {	// Deprecated, backwards-compatibility only
-			L.Icon.Default.imagePath = this._detectIconPath();
-		}
+		// if (!L.Icon.Default.imagePath) {	// Deprecated, backwards-compatibility only
```
perliedman: It looks like this would introduce a breaking change (setting L.Icon.Default.imagePath manually no longer supported), which I believe is unnecessary. Also, it will cause the code to detect the icon path every time an icon is changed (i.e. when a marker is created), which potentially is bad for performance.

I guess you need this change since `_getIconUrl` will be called twice and for your use case you need to return different paths for icon and shadow. However, I think you can achieve the same effect and still support manually setting `L.Icon.Default.imagePath` _and_ not do this for every call to `_getIconUrl`, but it would require some restructuring.

Also, if it's actually required to remove this support, the lines should just be removed, not commented out. But as mentioned above, I do not think removing them as is is a good idea.

PR #5042, opened 2016-10-24
Plugins: Add Leaflet.LabelTextCollision

yakitoritabetai:
Add Leaflet.LabelTextCollision to the list of plugins in docs.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1252,7 +1252,7 @@ These plugins provide new markers or news ways of converting abstract data into 
 			<a href="https://github.com/yakitoritabetai/Leaflet.LabelTextCollision">Leaflet.LabelTextCollision</a>
 		</td>
 		<td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
The plugin demo looks good, but I feel that the documentation is lacking. There is no mention that this plugin extends the `L.Canvas` renderer, nor any mention of the `text` option for `L.Path`s.

I would rather see [Leaflet.LayerGroup.Collision](https://github.com/MazeMap/Leaflet.LayerGroup.Collision) updated to handle tooltips and the such. Specially since your implementation of the collision algorithm has a huge algorithmic complexity compared to r-trees.

yakitoritabetai:
Thank you pointed out.
I wrote a readme.md in detail.

Collision detection algorithm is complex and difficult.
I will take a look at the Leaflet.LayerGroup.Collision.


I'm sorry that my English is difficult to read.

PR #5046, opened 2016-10-25
Fix marker shadow if max-width is set

brunob:
complement to c75af44ca2ed736cbff98a8b6d417b04355e275c


IvanSanchez:
LGTM. Did a small playground to test this: https://playground-leaflet.rhcloud.com/tuji/edit?html,output and seems to work OK. Didn't even notice the shadows were broken.

brunob:
Yes, it was broken only if `max-width: 100% !important;` was set on img, thx for the merge @IvanSanchez :)

PR #5049, opened 2016-10-25
L.GeoJSON.asFeature should pass through FeatureCollections

### LINKED ISSUE #5045: LayerGroup.toGeoJSON() can produce invalid GeoJSON
## How to reproduce
- Leaflet version I'm using: unpkg.com/leaflet@1.0.1/dist/leaflet-src.js
- Browser (with version) I'm using: Firefox 49.0, Chrome  54.0.2840.71
- OS/Platform (with version) I'm using: Mac 10.11.6

I expect the output of LayerGroup.toGeoJSON() to be valid GeoJSON. If the object can't be represented with valid GeoJSON I should get an error or warning while calling toGeoJSON().

Instead, the object created by toGeoJSON() is invalid GeoJSON and cannot be consumed by L.GeoJSON and reports.

`Uncaught Error: Invalid GeoJSON object` 

Since FeatureGroup and GeoJSON both extend LayerGroup, this also affects them. 

```
function test(){
    var lg = new L.LayerGroup();
    var startingJson = {"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Point","coordinates":[78.3984375,56.9449741808516]}}]};
    var gj = new L.GeoJSON(startingJson);
    lg.addLayer( gj );

    var json2 = lg.toGeoJSON();
    console.log( json2 );
       // Results in the following. The nesting is all weird. 
       // {"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"FeatureCollection","features":[{"type":"Feature","properties":{},"geometry":{"type":"Point","coordinates":[78.3984375,56.9449741808516]}}]}}]}

    gf = new L.GeoJSON(json2);
}
test();
```

IvanSanchez:
Should fix #5045. Unit tests wanted.

### DIFF
--- a/spec/suites/layer/LayerGroupSpec.js
+++ b/spec/suites/layer/LayerGroupSpec.js
@@ -68,4 +68,28 @@
 			}, ctx);
 		});
 	});
+
+	describe("#toGeoJSON", function () {
+		it('should return valid GeoJSON for a layer with a FeatureCollection', function () {
+			var geoJSON = {
+				"type":"FeatureCollection",
+				"features":[
+					{
+						"type":"Feature",
+						"properties":{},
+						"geometry": {
+							"type":"Point",
+							"coordinates": [78.3984375, 56.9449741808516]
+						}
+					}
+				]
+			};
+
+			var layerGroup = L.layerGroup();
+			var layer = L.geoJSON(geoJSON);
+			layerGroup.addLayer(layer);
+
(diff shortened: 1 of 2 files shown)


perliedman:
Added the requested unit test. I have verified that this test fails without @IvanSanchez's fix.

PR #5050, opened 2016-10-25
Events performance refactoring

Chosko:
This pull request should fix the #5044, solving the optimization issue with many event listeners.

To to this I created a ListenerCollection class, which aggregates and manages event listeners, and replaced  each `this._event[type]` array with an object of type `L.ListenerCollection` 

While adding less than  L.ListenerCollection.MAX_ARRAY listeners, the ListenerCollection acts as a normal array. When adding more listeners, the behaviour is switched to an hash with constant access time (but more memory is used).

In a similar way, when removing elements down to L.ListenerCollection.MIN_OBJECT listeners, the ListenerCollection switches its behaviour back to an array.

I kept MAX_ARRAY and MIN_OBJECT values separated to avoid an obscillating switching effect between array an hash behaviours.

The class is tested with 100% coverage. I didn't remove any test (just added them) and I tried to keep the behaviour exactly as it was before.

It should work but in a single day I didn't have the time to test it in production, or to benchmark the performances, so it may be not production-ready yet.

Tomorrow I will test performances in a production environment and I will benchmark everything.

In the meantime, I'll appreciate any feedback.

I'm open to see this get rejected: there aren't breaking changes but it's a big refactoring, and I'm not sure you want to keep it.

### DIFF
--- a/build/deps.js
+++ b/build/deps.js
@@ -3,6 +3,7 @@ var deps = {
 		src: ['Leaflet.js',
 		      'core/Util.js',
 		      'core/Class.js',
+          'core/ListenerCollection.js',
 		      'core/Events.js',
 		      'core/Browser.js',
 		      'geometry/Point.js',
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -603,5 +603,4 @@ describe('Events', function () {
 			expect(obj.listens('test')).to.be(false);
 		});
 	});
-
 });
--- a/spec/suites/core/ListenerCollectionSpec.js
+++ b/spec/suites/core/ListenerCollectionSpec.js
@@ -12,6 +12,46 @@ describe('ListenerCollection', function () {
 		});
 	});
 
+	describe('#each', function () {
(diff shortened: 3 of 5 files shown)


Chosko:
I changed the performance test made by @IvanSanchez in #5044, updating only the library source

Before the PR
https://playground-leaflet.rhcloud.com/foq/1/edit?html,output
Time elapsed: 4371.655000000002msec

After the PR
https://playground-leaflet.rhcloud.com/bal/edit?html,output
Time elapsed: 1089.2650000000003msec

Chosko:
Warning: I just found a performance issue while removing a layer.

in the old version, this line https://github.com/Leaflet/Leaflet/blob/master/src/core/Events.js#L160 takes a few time to duplicate the listeners.

in my version, the duplication is made with https://github.com/Leaflet/Leaflet/pull/5050/files#diff-0e79846ee2c407aa1ac0ccbfd8b15a44R133, which is implemented as an array.slice

IvanSanchez:
Superseded by #5054. The approach is interesting, but *for the time being* it's simpler to not use events for the problem in #5044.

Issue #4919, opened 2016-09-15
L.latLngBounds swapping longitude of Northeast and Southwest point


aaani:
Consider the example below
```
var southWest = L.latLng(-7.12725588946924, 159.14), northEast = L.latLng(70.0415297179124, 24.855900000000005);
var bounds = L.latLngBounds(southWest, northEast);

//logs - LatLng(-7.12726, 159.14), LatLng(70.04153, 24.8559)
console.log(southWest.toString(), northEast.toString());

//logs - LatLng(-7.12726, 24.8559)
console.log('bounds SW: ', bounds.getSouthWest().toString());
//logs - LatLng(70.04153, 159.14)
console.log('bounds NE: ', bounds.getNorthEast().toString());
```

Longitude for SouthWest and NorthEast points are swapped in the bounds. Why? This seems like a very bad bug.
Issue reproduction - http://jsbin.com/huziqi/edit?html,console

ghybs:
Hi,

If I am not mistaken, the conventional [cardinal directions](https://en.wikipedia.org/wiki/Cardinal_direction) and latitude / longitude ranges are:
```
             North (+90)
               |
(-180) West ———+——— East (+180)
               |
             South (-90)
```

Therefore your `southWest = L.latLng(-7.12725588946924, 159.14)` would actually be more East than your `northEast = L.latLng(70.0415297179124, 24.855900000000005)`.

When creating your `bounds` out of these corners, Leaflet detects the correct positions for you.
Is not it great? :-)

IvanSanchez:
This is a good example of Postel's lay: [«an implementation should be conservative in its sending behavior, and liberal in its receiving behavior»](https://en.wikipedia.org/wiki/Jon_Postel#Postel.27s_Law).

Perhaps it should be a good idea to change the docstrings and rename the parameters for the `LatLngBounds` constructor into `corner1` and `corner2` or something like that, to prevent confusion.

aaani:
@ghybs Some more context - 
![image](https://cloud.githubusercontent.com/assets/3347504/18590516/c45a970a-7bf4-11e6-861f-2b1c65f00f16.png)

Here is what I am expecting it to do - Region in the blue box is what I expect my bounds to be, while region is the red box is what leaflet is picking up.

ghybs:
@aaani Look at your blue polygon, it crosses the dateline.

A very easy workaround would have been to use a longitude of 159.14 - 360 = -200.86 to reach the left copy of the world.
[...]

PR #5059, opened 2016-10-27
Changed LatLngBounds arg names to corners

ghybs:
as proposed by @IvanSanchez in issue #4919.

This change should not affect the code, it is only for consistency with docstrings.

docstrings: replaced `southWest` and `northEast` argument names by `corner1` and `corner2`, to try to prevent confusion that Leaflet will keep these corners at those places, whereas it actually "re-orders" them to keep lowest values West-most / South-most, as if they were just corners.

### DIFF
--- a/src/geo/LatLngBounds.js
+++ b/src/geo/LatLngBounds.js
@@ -7,9 +7,9 @@
  * @example
  *
  * ```js
- * var southWest = L.latLng(40.712, -74.227),
- * northEast = L.latLng(40.774, -74.125),
- * bounds = L.latLngBounds(southWest, northEast);
+ * var corner1 = L.latLng(40.712, -74.227),
+ * corner2 = L.latLng(40.774, -74.125),
+ * bounds = L.latLngBounds(corner1, corner2);
  * ```
  *
  * All Leaflet methods that accept LatLngBounds objects also accept them in a simple Array form (unless noted otherwise), so the bounds example above can be passed like this:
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Me like.

Chris-Petty:
When does this get into the live site/docs? Would have saved me a couple hours 😃 .

ghybs:
Hi @Chris-Petty,

It should be live already?
http://leafletjs.com/reference-1.0.3.html#latlngbounds

For your issue, depending on your specific context and data characteristics, you may be able to simply implement a 360 degrees offset for whatever value is between -180 and 0 degrees longitude for example (you choose the cut-off, it may be different from 0), as proposed [here](https://github.com/Leaflet/Leaflet/issues/4919#issuecomment-247639618) or [there](https://stackoverflow.com/questions/40966062/sun-marker-position-render-outside-main-view-at-any-zoom-level/40967553#40967553).

Chris-Petty:
Thanks @ghybs yea looking now it seems updated, but yesterday I swear it was still SW/NE.

Speaking of which I think there is a minor inconsistency in the docs. The coordinates/points in examples in the PR are still SW/NE. In the docs for [bounds](http://leafletjs.com/reference-1.0.3.html#bounds), it talks about corner1 and corner2 in terms of "topLeft" and "bottomRight" (so effectively NW/SE). Would it be good to make the coordinates in the example code here match `bounds` conceptually?

ghybs:
Bounds docs change to corners should be live with the next release...
https://github.com/Leaflet/Leaflet/pull/5487

Chris-Petty:
Ah I see the ref further up now. Assumed they'd be together if accounted for.

PR #5065, opened 2016-10-31
Finish drags when a L.Draggable is removed, fix #5063.

### LINKED ISSUE #5063: L.Draggable._dragging stuck
## How to reproduce
- Leaflet version I'm using: //cdnjs.cloudflare.com/ajax/libs/leaflet/1.0.0/leaflet.js
- Browser (with version) I'm using: FF 49.0.2
- OS/Platform (with version) I'm using: Ubuntu 16.04

I have a temporary draggable markers that appear over a feature (on hover) and unless are dragged - I disappear them should the mouse hover away. Unless dragged, i.e.: unless they received 'dragstart'.
When i drag them real quick however (for some reason reproduces more easily on FF - seems slower) then the following happens:
1. `L.Draggable._onDown` sets the global `L.Draggable._dragging = true;`
2. `mousemove` reaches my code and at this point:
   - I haven't received 'dragstart' yet. 
   - the mouse is far enough for me to decide to remove the marker
3. I remove the marker and `L.Draggable._onUp` is never called leaving `L.Draggable._dragging` rot in its sorry state forever.

Obviously the offending consequence is that nothing can be dragged ever.
Understanding what's cooking I have a workaround, so no rush with this.

To see for youself:
https://playground-leaflet.rhcloud.com/qaha/1/edit?html,output

start dragging the marker and do not stop until it disappears. When it disappears the map is not draggable/panable.

I would expect the code be more resilient to this scenario - more specifically:
- if a draggable is being removed amidst a drag it should clear `L.Draggable._dragging`

IvanSanchez:


src/dom/Draggable.js
```diff
@@ -62,6 +62,8 @@ L.Draggable = L.Evented.extend({
 	disable: function () {
 		if (!this._enabled) { return; }
 
+		if (this._lastTarget) { this._onUp({}); }
```
yohanboniface: I think  this will fail because `_onUp` expect an argument `e`.

yohanboniface:
I was actually planning something in this direction: extracting from onUp the whole "finish dragging" functionnality, so it can be called programmatically :)

src/dom/Draggable.js
IvanSanchez: Nope, because I'm passing a `{}` ;-)

IvanSanchez:
@yohanboniface This is not meant as nice code, just a quick&dirty solution to #5063 :-)

src/dom/Draggable.js
remster: See here: https://playground-leaflet.rhcloud.com/vikez/edit?html,output
changed 'dragstart' to 'mousedown' and problem persists.
This is because you set Draggable._dragging on mousedown and you set this._lastTarget on mousemove.
I see a design problem with this. Perhaps set this._lastTarget in mousedown as well?
Or set Draggable._dragging to refer the very Draggable and clear on disable iif Draggable._dragging === this?

perliedman:
I changed the logic to use `L.Draggable._dragging` as [@remster suggested](https://github.com/Leaflet/Leaflet/pull/5065/files#r85732526). This fix also works with the playground example he provided (https://playground-leaflet.rhcloud.com/vikez/edit?html,output).

If we like this method, we should perhaps rename `L.Draggable._dragging` to `L.Draggable._currentDraggable` or something like that?

PR #5070, opened 2016-11-03
Move attribution option and getAttribution to L.Layer

### LINKED ISSUE #5051: L.geoJSON layer attribution not shown on map control
## How to reproduce
- Leaflet version I'm using: 1.0.0
- Browser (with version) I'm using: Chrome 54
- OS/Platform (with version) I'm using: OS X 10.11.6
- Add L.geoJSON layer with an attribution option to a map

Attribution statement not added to attribution control on map, in the same way that `L.tileLayer` is.

https://github.com/MashSoftware/place/blob/master/mash_place/templates/counties.html#L65

perliedman:
Since we now have a common base class for all layers, we can easily allow all layers to add attribution, not just `GridLayer` and `ImageOverlay`. This fixes #5051, since also `L.GeoJSON` can add attribution.

This PR moves the `attribution` option and the `getAttribution` method to `L.Layer`. I also updated some of the debug pages to reflect this change.

### DIFF
--- a/debug/map/image-overlay.html
+++ b/debug/map/image-overlay.html
@@ -35,7 +35,8 @@
 
 		var overlay = new L.ImageOverlay("https://www.lib.utexas.edu/maps/historical/newark_nj_1922.jpg", bounds, {
 			opacity: 0.5,
-			interactive: true
+			interactive: true,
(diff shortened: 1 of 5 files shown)


IvanSanchez:
Me like!

PR #5071, opened 2016-11-03
add margin to LatLngBounds equals method

miguelcobain:


danzel:
A test would be good here

miguelcobain:
@danzel tests were added. 👍

Issue #5073, opened 2016-11-03
Control.Layers collapse() method - icon is not showing


falqn:
- Leaflet version I'm using: 1.0.1
- Browser (with version) I'm using: Chrome 54.0.2840.71 m
- OS/Platform (with version) I'm using: Win7 64bit

Control.Layers collapse() method should collapse the control and icon of collapsed control should be visible but is not.

Take a look: https://playground-leaflet.rhcloud.com/tin/edit?html,output

PR #5074, opened 2016-11-03
Init the Control.Layers toggler when control collapses the 1st time

IvanSanchez:
This is a blind, untested solution to #5073. When a `collapsed: true` control is created, the toggler link is *not* created, thus the controls is invisible if `collapse()` is called.


yohanboniface:
Closing in favor of #5087

PR #5084, opened 2016-11-10
Added opacity to CircleMarker

### LINKED ISSUE #4683: Aligning setOpacity functionality between markers and circle markers
`L.markers` and `L.CircleMarker` both have a property called opacity, but only markers has a `setOpacity()` method for altering the opacity.  

Should we add a method to the L.Path(L.CircleMarker extends Path) so we can use the same function call to alter both the stroke and fill opacity of a circle marker as we use with a regular marker? 

I imagine the logic for the path method would would allow for either a single argument to set both: `.setOpacity(<number> opacity)`, or two arguments to set stroke and fill opacity separate: `.setOpacity(<number> stroke opacity, <number> fill opacity)`. 
Each method would return `this`.

A potential use of this would be when working with geojson, feature groups, or layer groups that includes multiple types of data where some data may be polygonal, some data may be point data, and some may be other types, where you want to quickly adjust the opacity of all of the sub-layers in the layer group.

Thoughts?

DecSander:
Closes #4683 

I wasn't sure how to set fill opacity since DomUtils only has setOpacity, which only seems to apply to stroke opacity.

Also amended test cases to match new option

### DIFF
--- a/src/layer/vector/CircleMarker.js
+++ b/src/layer/vector/CircleMarker.js
@@ -15,13 +15,18 @@ L.CircleMarker = L.Path.extend({
 
 		// @option radius: Number = 10
 		// Radius of the circle marker, in pixels
-		radius: 10
+		radius: 10,
+
+		// @option opacity: Number = 1.0
+		// The opacity of the marker
+		opacity: 1
 	},
 
 	initialize: function (latlng, options) {
 		L.setOptions(this, options);
 		this._latlng = L.latLng(latlng);
 		this._radius = this.options.radius;
+		this._opacity = this.options.opacity;
 	},
 
 	// @method setLatLng(latLng: LatLng): this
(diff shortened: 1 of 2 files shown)


src/layer/vector/CircleMarker.js
```diff
@@ -15,13 +15,18 @@ L.CircleMarker = L.Path.extend({
 
 		// @option radius: Number = 10
 		// Radius of the circle marker, in pixels
-		radius: 10
+		radius: 10,
+
+		// @option opacity: Number = 1.0
+		// The opacity of the marker
+		opacity: 1
```
IvanSanchez: This is not needed - opacity is already an option of `L.Path`

IvanSanchez:
Even though the code looks nice, and that I love having more tests, I disagree with the approach taken.

The `opacity` option comes from `L.Path`, not from `L.CircleMarker`, and is handled directly in the renderers. So it should make more sense to have `setOpacity` and `getOpacity` methods in `L.Path` instead.

Furthermore, even when the tests test the value of the `_opacity` private property, they don't test the opacity of the SVG path, or of the canvas drawn pixels.

And if you look at how the renderers update the style, the code can be much much simpler, just a decorator/wrapper over `setStyle` / `options`:

``` 
L.Path = L.Layer.extend({
   ...

   getOpacity: function(){ return this.options.opacity; },
   setOpacity: function(o){ return this.setStyle({ opacity: o }); }
});
```

I mean, if you (re)set the opacity, you're triggering a redraw anyway, so there's no penalty in using `setStyle()`.

So given that it takes about 20 characters of wrapper code over `setStyle`, I'm partial to rejecting this PR, and asking for a tutorial on styling instead (see #4239).

Issue #2012, opened 2013-09-08
Border around layercontrol not consistent with border around zoom control on mobile


ICarryTheDustOfAJourney:
On a full up-to-date Galaxy S3, S2 and Nexus 7 (Chrome and Firefox) the borders around the +- zoom control are transparent, but around the layer control its completely opaque.

PR #5085, opened 2016-11-10
Adding background-clip to .leaflet-bar

ddproxy:
Issue #2012 covers mobile devices

Leaflet/Leaflet.draw#629 detected an issue with svg in .leaflet-bar for retina or 'zoomed in' on chrome that would ignore scaling when using negative background positions.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -261,6 +261,7 @@
 .leaflet-bar {
 	box-shadow: 0 1px 5px rgba(0,0,0,0.65);
 	border-radius: 4px;
+	background-clip: padding-box;
 	}
 .leaflet-bar a,
 .leaflet-bar a:hover {


IvanSanchez:
@ddproxy This looks simple enough, but I must question why this change should be applied to `leaflet.css` instead of `leaflet.draw.css`

ddproxy:
Ah yes, the why.

This is a somewhat difficult issue to diagnose if you don't know what you are looking for. Potentially every plugin or leaflet application that uses svg's in their toolbar will have this issue on chrome in these conditions.

This is in the latest release of leaflet.draw in it's css, so this is not an issue anymore for Leaflet.draw. So this is just a sneaky patch for everyone else who doesn't know they need it yet.

Eventually this should be patched by Chrome (it's their problem, they should fix it) but in the meantime this is the best I've got 😬 

It would be okay to decline this, but a thought. https://github.com/Leaflet/Leaflet/pull/5085/files#diff-e3d63482ae6ce17380258fdb6cb730deL443  has overlap with this css tag. Someone more familiar with the css could place this somewhere better and get the same effect.

johnd0e:
> Eventually this should be patched by Chrome (it's their problem, they should fix it)

Is it still an issue?

Malvoz:
It's likely that Chrome fixed [the bug](https://github.com/Leaflet/Leaflet.draw/issues/629#issuecomment-259600703) (which exactly?) a long time ago.

@ddproxy Please respond if you still think this is needed.

PR #5087, opened 2016-11-10
Always create L.Control.Layers as collapsed; expand if collapse option not set

perliedman:
This is an alternative fix for #5073. @IvanSanchez's fix is in #5074. This is minor details, but it seemed a bit easier to me to avoid any state for fixing this.

I've tested both approaches with the example provided in #5073 and they both do their job.


perliedman:
Sure, I'll try to add a test.

Issue #5082, opened 2016-11-09
Canvas renderer mouseover event precedence is incorrect when features have overlapping pixels


jbccollins:
## How to reproduce

- Leaflet version I'm using: 1.0.1
- Browser I'm using: Chrome
- OS/Platform I'm using: Windows 10

## What behaviour I'm expecting and which behaviour I'm seeing

I expect that when two features render with overlapping pixels (which does not necessarily mean overlapping geometries) the feature that renders on top will be the only feature to receive mouseover events. The default SVG renderer handles overlapping features in this way. The canvas renderer triggers mouseover events on all layers that contain the mouseover point. It's a bit hard to put into words so for the sake of brevity here are two gifs

SVG (expected):
![mouseover_correct](https://cloud.githubusercontent.com/assets/8915552/20155636/ca2da78c-a69a-11e6-88e9-abb7c872225e.gif)

Canvas: (unexpected)
![mouseover_bug](https://cloud.githubusercontent.com/assets/8915552/20155638/cbcae532-a69a-11e6-9dd1-509cac26d8d8.gif)

I believe that this was introduced (or at least made more apparent) by https://github.com/Leaflet/Leaflet/commit/609a79279272c88f1089bc2119d309b18c169110 which removes the break statement.

## Minimal example reproducing the issue
- A jsfiddle with a very simple example: http://jsfiddle.net/jcollin6/6k05szap/7/

## Possible solution

I would be willing to put up a PR for this but I don't know how. I tested these changes by directly modifying my local copy of leaflet.

```javascript
	_handleMouseHover: function(e, point) {
	    var id, layer, drawnLayerKeys;
            // reverse the keys so that we check layers from the top down 
	    drawnLayerKeys = Object.keys(this._drawnLayers).concat().reverse();
	    for (var i = 0; i < drawnLayerKeys.length; i++) {
	        id = drawnLayerKeys[i];
	        layer = this._drawnLayers[id];
	        if (layer.options.interactive && layer._containsPoint(point)) {
	            L.DomUtil.addClass(this._container, 'leaflet-interactive'); // change cursor
	            if (this._hoveredLayer && (this._hoveredLayer._leaflet_id.toString() !== id.toString())) {
                        // If we find a match but another layer is already 'hovered' 
                        // then trigger mouseout on that layer 
	                this._fireEvent([this._hoveredLayer], e, 'mouseout');
	            }
	            this._fireEvent([layer], e, 'mouseover');
[...]

PR #5090, opened 2016-11-11
Make sure mouseout for previous hovered layer before mouseover for new

perliedman:
This addresses #5082, where `mouseout` is sometimes not fired before `mouseover` is fired for a new layer.


perliedman:
This also fixes a problem (probably actually the same problem) with the `debug/vector/geojson.html` example - before this fix, it rendered incorrectly if you set `preferCanvas: true`.

IvanSanchez:
Looks sane, and I trust @perliedman when it comes to event handling.

PR #5098, opened 2016-11-14
Remove layer from map before calling onRemove

joukewitteveen:
This makes sure that the popupclose event fires with an actually closed popup.
function onPopupClose (e) {
  e.popup.isOpen(); // Previously, this would be true
}

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -173,6 +173,8 @@ L.Map.include({
 
 		if (!this._layers[id]) { return this; }
 
+		delete this._layers[id];
+
 		if (this._loaded) {
 			layer.onRemove(this);
 		}
@@ -180,8 +182,6 @@ L.Map.include({
 		if (layer.getAttribution && this.attributionControl) {
 			this.attributionControl.removeAttribution(layer.getAttribution());
 		}
-
-		delete this._layers[id];
 
 		if (this._loaded) {
 			this.fire('layerremove', {layer: layer});


IvanSanchez:
@joukewitteveen What's the error this is supposed to fix? Is there a unit test, bug report, etc?

joukewitteveen:
Consider this PR a bug report by itself.
Suppose we want to maintain a list of opened popups (or tooltips, or ...) and update this list on the `popupopen` and `popupclose` events by iterating over all popups (or tooltips, or ...) and including those that are open according to `.isOpen()`. Currently, this approach does not work (http://playground-leaflet.rhcloud.com/roni/edit?html,console,output).

joukewitteveen:
A workaround is possible: http://playground-leaflet.rhcloud.com/mese/edit?html,console,output
Watch the console when you close the popup.
The workaround looks like this:
```
    marker.on('popupclose', function (e) {
      e.popup.once('remove', <fn>);
    });
```

perliedman:
Hi @joukewitteveen, and thanks for both contributing this as well as the explanation of the issue you're seeing.

While I understand your particular use case, I think changing this behavior can just as easily break applications that rely on the current implementation, which is worse.

As I see it, the docs does not clearly say, but [`onRemove`](http://leafletjs.com/reference.html#ilayer-onremove) says: "Should contain all clean up code that removes the layer's elements from the DOM" - arguably, the layer isn't really removed until `onRemove` has run, which would justify the current behavior.

I believe we risk having regressions from this change, and the use case it addresses seems like an edge case that can be worked around. So, I'm voting for closing this.

Without having seen your code, I kind of wonder why you're not using the `popup` property of the `popupclose`/`popupopen` events to figure out which are open/closed instead of iterating?

Issue #5091, opened 2016-11-11
Canvas renderer click events have unexpected propagation


jbccollins:
## How to reproduce

- Leaflet version I'm using: https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/master/leaflet-src.js
- Browser I'm using: Chrome
- OS/Platform I'm using: Windows 10

## What behaviour I'm expecting and which behaviour I'm seeing
Related to: #5082 which was resolved by @perliedman and @IvanSanchez.

I expect that when two features render with overlapping pixels (which does not necessarily mean overlapping geometries) the feature that renders on top will be the only feature to receive click events. The default SVG renderer handles overlapping features in this way. The canvas renderer triggers click events on all layers that contain the clicked point.
Here are two .gif's demonstrating the differences in behaviour. In both gifs "mouseover" on a feature will color it blue and "click" on a feature will color it black.

SVG (expected):
![click_expected](https://cloud.githubusercontent.com/assets/8915552/20223810/43b913b0-a809-11e6-86b5-576be7ffc01e.gif)

Canvas: (unexpected)
![click_unexpected](https://cloud.githubusercontent.com/assets/8915552/20223817/4a683754-a809-11e6-9466-288b949d82d9.gif)

## Minimal example reproducing the issue
http://jsfiddle.net/jcollin6/rvk38ga5/2/

perliedman:
I should have seen this one coming, I guess the click logic is very similar to the old hover logic. I will look into it as soon as I can find the time.

jbccollins:
@perliedman yeah, sorry, I hate to be a bother. Hopefully the click/hover logic is similar enough that this won't be too much trouble for you.

jbccollins:
Thanks guys! Any idea when the next npm version that will contain these changes is scheduled to be deployed?

PR #5100, opened 2016-11-14
Fix canvas firing click event on multiple overlapping layer on canvas

perliedman:
As reported in #5091, if overlapping layers are clicked with canvas renderer, each of the overlapping layers get a click event. This PR makes sure only the topmost layer gets the click.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -252,17 +252,17 @@ L.Canvas = L.Renderer.extend({
 	// so we emulate that by calculating what's under the mouse on mousemove/click manually
 
 	_onClick: function (e) {
-		var point = this._map.mouseEventToLayerPoint(e), layers = [], layer;
+		var point = this._map.mouseEventToLayerPoint(e), layer, clickedLayer;
 
 		for (var id in this._layers) {
 			layer = this._layers[id];
 			if (layer.options.interactive && layer._containsPoint(point) && !this._map._draggableMoved(layer)) {
-				L.DomEvent._fakeStop(e);
-				layers.push(layer);
+				clickedLayer = layer;
 			}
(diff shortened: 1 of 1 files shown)


jbccollins:
Looks good to me fwiw.

src/layer/vector/Canvas.js
```diff
@@ -252,17 +252,17 @@ L.Canvas = L.Renderer.extend({
...
 	_onClick: function (e) {
-		var point = this._map.mouseEventToLayerPoint(e), layers = [], layer;
+		var point = this._map.mouseEventToLayerPoint(e), layer, clickedLayer;
 
 		for (var id in this._layers) {
 			layer = this._layers[id];
 			if (layer.options.interactive && layer._containsPoint(point) && !this._map._draggableMoved(layer)) {
-				L.DomEvent._fakeStop(e);
-				layers.push(layer);
+				clickedLayer = layer;
```
yohanboniface: You may fire event here directly and return, no?

jbccollins: No I don't think so.  I believe this is looping over all the layers starting from the lowest layer to the highest layer. If you break on the first layer that contains the clicked point then that will be the lowest layer.

perliedman: Exactly like @jbccollins says: since we render by looping `this._layers`, the last element will be drawn on top, and we can't break until we have checked that, if we want the topmost layer to get the click.

yohanboniface: Good point!

jbccollins:
Looks like this just needs a quick rebase before it merges?

IvanSanchez:
@jbccollins I think you mean "needed", in past tense :-P

PR #5106, opened 2016-11-16
Fixes Layer.isPopupOpen when no popup is bound

gvangool:
This fixes uncaught exception when checking whether a popup isOpen(), when the layer has no popup bound to it.

### DIFF
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -328,3 +328,65 @@ describe("L.Map#openPopup", function () {
 	});
 
 });
+
+describe('L.Layer#_popup', function () {
+	var c, map, marker;
+
+	beforeEach(function () {
+		c = document.createElement('div');
+		c.style.width = '400px';
+		c.style.height = '400px';
+		map = new L.Map(c);
+		map.setView(new L.LatLng(55.8, 37.6), 6);
+		marker = L.marker(L.latLng(55.8, 37.6)).addTo(map);
+	});
+
+	afterEach(function () {
+		if (document.body.contains(c)) {
+			document.body.removeChild(c);
+		}
+	});
+
+	it("only adds a popup to the map when opened", function () {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Oooh, I like this. Could it be possible to add a unit test, which makes sure that a call to `isPopupOpen()` does not throw an error?

gvangool:
Sure, but I wasn't able to find other unit tests for the `L.Layer` popup functions.

Do they exists?

IvanSanchez:
No, maybe you should add them to https://github.com/Leaflet/Leaflet/blob/master/spec/suites/layer/PopupSpec.js . If you look at that file, you'll see it has tests for both `L.Popup` and the popup methods/options for `L.Map`. Having tests for the popup methods/options for `L.Layer` in the same place makes sense.

gvangool:
Alright, I'll have a look tonight!

gvangool:
@IvanSanchez this is my initial pass at the test cases, I kept 1 separate since I'm unsure whether this is actually expected behavior.

IvanSanchez:
This looks quite good :+1: . One thing, though: As the current undesired behaviour is that a method call throws an error, the tests should explicitly check that the method call does indeed *not* throw an error (and check that, without your fix, the unit tests fail).

For reference, see https://github.com/Leaflet/Leaflet/blob/88d107ca969be9d7a5e7b17fafc4aadab466e18f/spec/suites/control/Control.LayersSpec.js#L104

mepdr:
Tooltip has the same issue https://github.com/Leaflet/Leaflet/blob/master/src/layer/Tooltip.js#L371

IvanSanchez:
@medpr Feel free to write a fix and some tests, then.

gvangool:
@IvanSanchez changed the commits to first add the tests and only then change the behavior

perliedman:
@gvangool could you rebase this on `master`? Other than that, I think we're good to merge

PR #5110, opened 2016-11-17
add Redraw event

PowerPan:
Fires an event after Redrawing a Path Layer

### DIFF
--- a/docs/reference-1.0.0.html
+++ b/docs/reference-1.0.0.html
@@ -6514,9 +6514,28 @@ precedence over the map&#39;s <a href="#map-renderer">default renderer</a>.</td>
 </div>
 
 </section><section>
-<h3 id=''>Events</h3>
+<h3 id='path-event'>Events</h3>
+
+	<section >
 
 
+
+
+		<table><thead>
+		<tr>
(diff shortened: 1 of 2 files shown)


IvanSanchez:
What problem does this solve?

PowerPan:
It is a feature
with this, you can register at the redraw event and you can do than some stuff.
I need this feature for one of my projects. Reading the docs I found no other solution to do this.

IvanSanchez:
I don't think there's a good rationale for this. The only calls to `L.Path.redraw` come from `setLatLngs` (polyline, polygon), `setLatLng` and `setRadius` (circlemarker). If I had lots of those, I'd rather stay away from event handlers and just make an extra call after `setLatLngs` or whatever.

In the meantime, feel free to monkey-patch Leaflet in your code with something like:

```js
L.Path.include({
  redraw:  function () {
		if (this._map) {
			this._renderer._updatePath(this);
		}
		return this.fire('redraw');;
	},
})
```

If you think this should be a Leaflet feature, please **explain your rationale**. Give a couple of examples of something that could never be done without this feature.

PR #5112, opened 2016-11-17
Class name option for layer control items

Neorth:
I ran into an issue where I had to indent some of the layers in the layer control. This PR makes it possible to pass a class name for every item added to the layer control. The class name will be set on the label for the item in the list of layers.

With this PR the layer passed to the layer control can be replaced by an object with the following structure:
`{ layer: layer, className: className }`

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -215,12 +215,15 @@ L.Control.Layers = L.Control.extend({
 		}
 	},
 
-	_addLayer: function (layer, name, overlay) {
+	_addLayer: function (obj, name, overlay) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
No need.  You should be able to do something like

```js
L.control.Layers({
    '<span class="foo">Foo</span>': L.tileLayer(whatever)
}, {}, {}).addTo(map);
```

Neorth:
That will only give you control over the span content, i.e. not the input boxes. My specific use case it to indent individual items. I guess it also could be used to style the input boxes individually (or by groups). Another use case I can think of is to track (user initiated) layer changes to Analytics.

PR #5130, opened 2016-11-22
Use openstreetmap.org, not osm.org

systemed:
osm.org isn't an officially supported endpoint, doesn't have SSL certificates etc. - it's just a convenience for typing into your browser URL bar


IvanSanchez:
@systemed  As stated in CONTRIBUTING.md, please don't just change the generated `reference-version.html` files, but the docstrings in `src/layer/tile/TileLayer.js` as well.

If not, the next release of the docs will have the same error.

systemed:
And done. I think.

PR #5131, opened 2016-11-22
Only collapse layer control if collapsed: true

### LINKED ISSUE #5126: L.control.layers collapses after first mouse hover
- Browser (with version) I'm using: Chrome
- OS/Platform (with version) I'm using: Win7

I'm initializing the map with 'collapse:false' option. After first mouse hover, the control collapses, but it should stay expanded at all times.

https://playground-leaflet.rhcloud.com/bipo/edit?html,output

perliedman:
Fixes regression caused by #5087. Close #5126.

### DIFF
--- a/spec/suites/control/Control.LayersSpec.js
+++ b/spec/suites/control/Control.LayersSpec.js
@@ -140,6 +140,46 @@ describe("Control.Layers", function () {
 		});
 	});
 
+	describe("collapse when collapsed: true", function () {
+		it('expands when mouse is over', function () {
+			var layersCtrl = L.control.layers(null, null, {collapsed: true}).addTo(map);
+			happen.once(layersCtrl._container, {type:'mouseover'});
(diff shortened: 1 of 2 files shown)


jieter:
It still seems to have different behaviour: moving the mouse over the control and then back to the map makes the control collapse, while before it stayed open with `collapsed: false`

[playground with leaflet 1.0.2](https://playground-leaflet.rhcloud.com/vup/edit?html,output) (collapses after hovering control)
[playground with leaflet 1.0.0](https://playground-leaflet.rhcloud.com/ticu/1/edit?html,output) (not collapsing, expected behaviour)

IvanSanchez:
@jieter The behaviour in `master` after this PR should be the same as 1.0.0.

jieter:
Ah, OK, the changelog description tricked me into thinking that this is part of 1.0.2, which it is apparently not.

PR #5135, opened 2016-11-23
Sanity check in LatLngBounds.contains (fixes #5132)

### LINKED ISSUE #5132: Null Reference passing object to LatLngBounds.contains
## How to reproduce

- Leaflet version I'm using: 1.0.1 & 1.0.2
- Browser (with version) I'm using: Firefox 49.02 and Edge 25
- OS/Platform (with version) I'm using: MacOS 10.11.6 for Firefox/Windows 10 for Edge
- call LatLngBounds.contains with an object-style LatLng, e.g. `{lat:0,lon:0}`

expecting: the function to return true or false
seeing: a null reference exception

http://codepen.io/plaidphantom/pen/YpVZap

I noticed this issue originally while using 1.0.1, but putting together the CodePen I tested it with 1.0.2 and get identical behavior.

According to the [LatLng](http://leafletjs.com/reference-1.0.0.html#latlng) documentation i think this should work, as there's nothing "noted otherwise" on [.contains](http://leafletjs.com/reference-1.0.0.html#latlngbounds-contains)

IvanSanchez:
Note that the new unit tests fail without the fix.

### DIFF
--- a/spec/suites/geo/LatLngBoundsSpec.js
+++ b/spec/suites/geo/LatLngBoundsSpec.js
@@ -138,9 +138,19 @@ describe('LatLngBounds', function () {
 	});
 
 	describe('#contains', function () {
-		it('returns true if contains latlng point', function () {
+		it('returns true if contains latlng point as array', function () {
 			expect(a.contains([16, 20])).to.eql(true);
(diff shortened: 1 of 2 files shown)


perliedman:
@IvanSanchez could you just rebase and I think this is good to merge.

PR #5137, opened 2016-11-23
Fix max/min calculation for getBoundsZoom (fixes #5136)

### LINKED ISSUE #5136: getBoundsZoom with option "inside?" does not always set zoom to inside the bounds
## How to reproduce

- Leaflet version I'm using: 1.0.2 (issue occurs from rc.1)
- Browser (with version) I'm using: Chrome 55
- Set zoom level using getBoundsZoom(bounds, true) where bounds is a different aspect ratio to the map.
   - For example bounds of (-85,-180), (85,180) with a map of 600x400 px

When using getBoundsZoom with the option "inside" set to true I expect the zoom level to be set such that the map view is entirely inside the bounds set as per the [documentation](http://leafletjs.com/reference-1.0.0.html#map-getboundszoom).
When using version of leaflet >= 1.0.0-rc1 this is not always the case.

Note I added a polygon to the map to showing the bounds.

Using Leaflet version 1.0.2 the view is outside bounds: https://playground-leaflet.rhcloud.com/yaca/edit?html,output

Same for Leaflet Version 1.0.0-rc.1: https://playground-leaflet.rhcloud.com/juc/edit?html,output

However in version 1.0.0-beta.2 it works as expected with the view completely contained within the bounds: https://playground-leaflet.rhcloud.com/tom/1/edit?html,output

IvanSanchez:
When `inside` is true, the code shall use `Math.max` instead of `Math.min` when comparing the horizontal and vertical scale factors between the bbox and the map viewport size.


src/map/Map.js
```diff
@@ -785,9 +785,11 @@ L.Map = L.Evented.extend({
 		    se = bounds.getSouthEast(),
 		    size = this.getSize().subtract(padding),
 		    boundsSize = this.project(se, zoom).subtract(this.project(nw, zoom)),
-		    snap = L.Browser.any3d ? this.options.zoomSnap : 1;
+		    snap = L.Browser.any3d ? this.options.zoomSnap : 1,
+		    scalex = size.x / boundsSize.x,
+		    scaley = size.y / boundsSize.y,
+		    scale = inside ? Math.max(scalex, scaley) : Math.min(scalex, scaley);
```
IvanSanchez: This could also be written as `scale = Math.[ inside ? 'max' : 'min' ](size.x / boundsSize.x, size.y / boundsSize.y)`.

But I'd rather make the code a bit more readable, even if it means making it a few bytes bigger.

IvanSanchez:
@perliedman Added one test for the `inside` parameter. Should be enough.

Issue #4534, opened 2016-05-03
Remove peerDependencies in all official Leaflet plugins


hyperknot:
This came up on an Editable issue: https://github.com/Leaflet/Leaflet.Editable/commit/eaaa6cd

I recommend to get rid of peerDependencies in all official Leaflet plugins.

1. They provide no value at all, no one is going to install a Leaflet
plugin without Leaflet being available (or they might use Leaflet from CDN, etc.).
2. The canonical place for version compatibility information is in the
README.md, like it has always been for all Leaflet plugins.
3. npm actually doesn't do anything useful with peer dependencies, it just
shows that warning that peer dependencies are not installed automatically.
4. On the other hand, we can get crazy build failure situations like https://github.com/Leaflet/Leaflet.Editable/commit/eaaa6cd8c2f626f19b08d5f5950c57ce971d2dc8#commitcomment-17338411
5. If Leaflet is required for building a plugin, it should be specified in devDependencies, not peer.

=> I recommend removing peerDependencies in all official Leaflet plugins

theashyster:
I think this is fixed. I went through all of the repositories available here https://github.com/Leaflet and could not find `peerDependencies` in `package.json` in any of those.

PR #5142, opened 2016-11-24
Change dependencies to peerDependencies

Ralf8686:
If I some plugin include leaflet to dependencies and I have local version leaflet, then I have two version - leaflet from plugin and my.


theashyster:
I am confused now - so this was a wrong thing to do? https://github.com/Leaflet/Leaflet/issues/4534

semone:
Plus one on the confusion

theashyster:
I would suggest the same thing as @hyperknot in https://github.com/Leaflet/Leaflet/issues/4534
`If Leaflet is required for building a plugin, it should be specified in devDependencies, not peer.`

PR #5160, opened 2016-12-02
Remove incrementation of uninitialized typeListeners attribute 'count'

### LINKED ISSUE #5159: typeListeners.count never initialized
This line initializes all type listeners:
https://github.com/Leaflet/Leaflet/blob/a97b87d0333fdc4b0c18d7b2ee9c15b393ae76ff/src/core/Events.js#L98

However this line will always try to increment `NaN` because `count` is never initialised. 
https://github.com/Leaflet/Leaflet/blob/a97b87d0333fdc4b0c18d7b2ee9c15b393ae76ff/src/core/Events.js#L117

I can only assume this is a leftover, because the `count` is never referenced again somewhere, is it?

Brikky:
`typeListeners.count` was removed in an earlier update to event handling, but was still being referenced and producing an error when incremented.

Fixes #5159

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -114,7 +114,6 @@ L.Evented = L.Class.extend({
 		}
 
 		listeners.push(newListener);
-		typeListeners.count++;
 	},
 
 	_off: function (type, fn, context) {


perliedman:
:+1: Great, thanks for contributing this fix!

PR #5161, opened 2016-12-02
Correct latLng ordering in examples

jasonoverland:


jasonoverland:
Ok, I amended to include requested changes to the source code.  I also corrected an issue with the outer ring & hole examples.

IvanSanchez:
Looks good, thanks! :smile:

Issue #5149, opened 2016-11-29
TileLayer/GridLayer requesting tiles outside of specified bounds


mrMetalWood:
## How to reproduce

- Leaflet version I'm using: 1.0.2
- Browser (with version) I'm using: Chrome 54.0.2840.98
- OS/Platform (with version) I'm using: macOS Sierra 10.12

- Add a tile layer with bounds to the map.

## What behaviour I'm expecting and which behaviour I'm seeing

I would expect that only tiles inside the bounding box are requested and displayed.
Instead there are tiles requested and displayed that are outside the boundingbox. More specific tiles that are located near "The International Date Line".

![bildschirmfoto 2016-11-29 um 14 53 25](https://cloud.githubusercontent.com/assets/2902795/20712316/c9d73178-b643-11e6-8c94-a874cb7f67cd.png)

## Minimal example reproducing the issue

In the example there are tiles requested and displayed that are located inside the red box and are not part of the bounds.
There has already been some conversation with @IvanSanchez in the leaflet Gitter channel about this topic on November 24, 14:05.
Most likely its some kind of issue that has to do with coordinates being wrapped around the date line and not correct boundingboxes being returned from a private GridLayer function. Here the concluding quote from Ivan:

> The L.GridLayer._tileCoordsToBounds private method is not returning good bboxes for tiles with a x-coordinate of -1

https://playground-leaflet.rhcloud.com/lulic/1/edit?html,output

IvanSanchez:
Note: That playground shows the expected bounds of the `{x: -1, y: 1, z: 2}` tile in red, and the actual in blue.

mrMetalWood:
I think the other way around. Actual bounds in red, expected in blue.

IvanSanchez:
@mrMetalWood Ahem.

```js
    // Actual tile bounds of a tile with a x-coordinate of -1 (red)
    // i.e. {x: -1, y: 1, z: 2}
    var tileRect = L.rectangle(
      position._tileCoordsToBounds(L.point({x: -1, y: 1, z: 2})),
      {color: 'blue', weight: 5}
    ).addTo(map);
    

    // Expected tile bounds of a tile with a x-coordinate of -1 (blue)
    // i.e. {x: -1, y: 1, z: 2}
    var tileRect = L.rectangle(
      [[0,-180],[66.51326044311188,-270]],
      {color: 'red', weight: 1}
    ).addTo(map);
```

[...]

PR #5164, opened 2016-12-04
Fix for wrapping inside the antimeridian (issue #5149)

DiogoMCampos:
Related to issue #5149, check my comment there.

I believe the problem was that, when calling the ```wrapLatLng()``` function, sometimes one of the coordinates was outside the range while the other one was inside (as explained in my last edit in the aforementioned comment). This possible solution would force both points (```nw``` and ```se```) to be moved by the same amount.

This still wouldn't move the blue rectangle in the issue #5149 example to the place of the red rectangle, but that is due to the ```noWrap``` option being set to false by default. If the user wants to select something outside the [-180, 180] longitude range, that option should be set to true.

I am not sure if the code fits the project guidelines in terms of coding practices, so please do correct me if anything is wrong.

Hopefully this is a valid solution. If not, I am definitely available to keep working on it.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -684,8 +684,18 @@ L.GridLayer = L.Layer.extend({
 		    se = map.unproject(sePoint, coords.z);
 
 		if (!this.options.noWrap) {
-			nw = map.wrapLatLng(nw);
-			se = map.wrapLatLng(se);
+			var newNw = map.wrapLatLng(nw);
+			var newSe = map.wrapLatLng(se);
+
+			if (!newNw.equals(nw)) {
+				se.lat += (newNw.lat - nw.lat);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This PR looks good!

However, I'm not sure how the code will behave when presented with edge cases like bounds equal or larger than the map bounds, or bounds that just touch (intersect) but not overlap the antimeridian. And the bug this fixes (#5149) is exactly about touching but not overlapping the antimeridian.

Realizing those edge cases (and the lack of tests in this PR) is what prompted me to work on #5166.

I'll let other team folks give an opinion on the matter and choose a PR to fix the bug.

IvanSanchez:
#5166 has been merged, so I have to close this PR.

Sorry your code didn't finally make it - but we appreciate the PR a lot!

PR #5168, opened 2016-12-06
Point to correct PR for the layers control  PR

jieter:
Was pointing to #5090 Make sure mouseout for previous hovered layer before mouseover for new
should point to #5087 Always create L.Control.Layers as collapsed; expand if collapse option not set


jieter:
Hmm, should be #5131

IvanSanchez:
@jieter Maybe we want to point to all the involved PRs instead?

jieter:
Yes, but that makes the description inaccurate. So I need to allocate more brain-cycles to fix this.

PR #5171, opened 2016-12-07
Flip the order of the 'moveend' and 'viewreset' events in map (fixes #5170)

### LINKED ISSUE #5170: L.Canvas not redrawing properly on non-animated map pan
I report several problems in L.Canvas implementation: using map.setView without animation causes geometries to be drawn twice even when they should not be displayed at all.
This is shown in http://playground-leaflet.rhcloud.com/yuxo/2/edit?html,console,output

IvanSanchez:
Flipping the order of the `moveend` and `viewreset` events in the map fixes #5170, but I'm not sure if this might have any side effects.

In particular, I don't know if this will be a regression of https://github.com/Leaflet/Leaflet/pull/4193

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1084,12 +1084,12 @@ L.Map = L.Evented.extend({
 		this
 			._moveStart(zoomChanged)
 			._move(center, zoom)
-			._moveEnd(zoomChanged);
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Closing in favour of #5250.

PR #5172, opened 2016-12-07
Different _reset methods for Canvas & SVG (fixes #5170)

### LINKED ISSUE #5170: L.Canvas not redrawing properly on non-animated map pan

IvanSanchez:
An alternative to #5171, this also fixes #5170.

Rather than changing the order of the `L.Map` events, this changes the order of the calls in a `L.Canvas` reset/redraw, to trigger the reprojection of the paths' pixel coordinates **before** drawing them.

Adds a bit more code, but the impact in the rest of the code (with edge cases like the `GridLayer` pruning and so on) should be greatly reduced.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -92,6 +92,15 @@ L.Canvas = L.Renderer.extend({
 		this.fire('update');
 	},
 
+	_reset: function () {
+		for (var id in this._layers) {
+			this._layers[id]._reset();
+		}
+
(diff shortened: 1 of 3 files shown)


IvanSanchez:
Closing in favour of #5250.

PR #5185, opened 2016-12-12
Be more picky about pointer events (blind attempt to fix #5180)

### LINKED ISSUE #5180: Duplicated 'dblclick' and phantom 'touchstart' events with Chrome >= 55
Starting with chrome 55.0.2883.75 'dblclick' is fired twice

It may be related to issue #4127

Example from issue #4127 is showing the problem
http://playground-leaflet.rhcloud.com/coh/edit?html,console,output

On Firefox 50.0.2, Safari 10.0.1, Opera 41.0.2353.69, Chromium 53.0.2785.143, 
You get expected behavior:
"map click!"
"map click!"
"map double click!"
"body double click!"

While with chrome 55.0.2883.75 you get:
"map double click!"
"body double click!"
"map click!"
"map double click!"
"body double click!"

And Chrome in beta 56.0.2924.21 you get:
"map click!"
"map double click!"
"body double click!"
"map click!"
"map double click!"
"body double click!"

I still have a test machine on Chrome 54 and it has expected behavior.

IvanSanchez:
This PR aborts the handling of `PointerEvent`s in two cases:

* In `DomEvent.DoubleTap`, do not check for double taps if the event is a pointer event of type `mouse` when not in Edge.
  This gets rid of the synthetic doubleclicks in Chrome >=55. Chrome still needs that code so that Leaflet fires `dblblick`s when using a finger.
  This also means that mouse `dblclick`s in Chrome >=55 are handled by the OS **only**, and touch `dblclick`s are handled by the Leaflet code.

* In `DomEvent.Pointer.js`, skip handling the event if the `pointerType` is a `mouse`. There is an extra check for IE (IE11???) which uses the `e.MSPOINTER_TYPE_MOUSE` constant, which doesn't exist in Chrome. That's what's causing the spurious touch events.

This should fix #5180, hopefully. Testing this PR would help.

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -17,6 +17,9 @@ L.extend(L.DomEvent, {
 			var count;
 
 			if (L.Browser.pointer) {
+				if (!L.Browser.edge && e.pointerType === 'mouse') {
+					return;
+				}
 				count = L.DomEvent._pointersCount;
 			} else {
 				count = e.touches.length;
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -50,7 +50,7 @@ L.extend(L.DomEvent, {
 
 	_addPointerStart: function (obj, handler, id) {
 		var onDown = L.bind(function (e) {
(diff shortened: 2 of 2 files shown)


perliedman:
Looks good, I tested this in Chrome 55 and Edge - in Edge, I couldn't test touch events because of my setup, so we'll have to just hope it still works (although I see not reason it shouldn't).

PR #5187, opened 2016-12-13
Fix typo in touch-action CSS rules

IvanSanchez:
Apparently @perliedman introduced a very-hard-to-spot typo in the `touch-action` CSS rules in #4552, and nobody noticed until #5180. :disappointed:


perliedman:
Obviously looks good, good catch.

PR #5189, opened 2016-12-14
Fix L.Browser.pointer for chrome, Edge

pranavk:
window.PointerEvent is available on chrome, Edge, this makes
L.Browser.pointer true even if there is no pointer.

Keeping navigator.maxTouchPoints as one of the checks does well
across all the browsers.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -30,7 +30,7 @@
 
 	    mobile = typeof orientation !== 'undefined' || ua.indexOf('mobile') !== -1,
 	    msPointer = !window.PointerEvent && window.MSPointerEvent,
-	    pointer = window.PointerEvent || msPointer,
+	    pointer = (window.PointerEvent && navigator.maxTouchPoints) || msPointer,
 
 	    ie3d = ie && ('transition' in doc.style),
 	    webkit3d = ('WebKitCSSMatrix' in window) && ('m11' in new window.WebKitCSSMatrix()) && !android23,


danzel:
We've had `maxTouchPoints` detection before, but we removed it.

https://github.com/pranavk/Leaflet/commit/afa0d1d87b4c522ca3cbeef785c829b886355982#diff-bc4c02f919ba9f4e3259e37fa12927a7
https://github.com/pranavk/Leaflet/commit/c6d858775341667dceefd0f7b9f0718cf7d52c0e#diff-bc4c02f919ba9f4e3259e37fa12927a7

Not sure we should be bringing it back.

perliedman:
Hm, exactly what is the issue we're trying to fix here? The docs for [`L.Browser.pointer`](http://leafletjs.com/reference-1.0.2.html#browser-pointer) says: "true for all browsers supporting pointer events."

This _is_ true for Chrome and Edge, they support pointer events, so it is indeed correct that `L.Browser.pointer` is `true`. Or am I missing something?

IvanSanchez:
Chrome >55 and Edge do, in fact, fire `PointerEvent`s even if the computer only has a traditional mouse.

The `L.Browser.Pointer` flag is there to account for the `PointerEvent` behaviour, not for the multi-touch or multi-pointer behaviour.

If this creates a problem, please open a bug report, including browser version, pointing devices, and reproducible example.

skalb:
@IvanSanchez I noticed that L.Browser.touch is also true on Chrome 55 / OSX, is that also intended? 

This causes the leaflet-touch style to be applied to leaflet-bar buttons, for example.

PR #5206, opened 2016-12-18
Implemented polyline wrapping

williamnf:
As discussed on issue  #2645, we have developed an implementation for Polylines that allow for two points to link using the shorter path around the international date line instead of the direct path when the difference between the longitudes of each point is higher than half the size of the CRS bounds (180). This is activated by setting the noWrap variable of the L.polyline object to true. 
However, this is based on hard-coded values of the range. We aren't sure if it is possible to use the wrapLng values of the CRS.Earth file on the geo/crs folder.

### DIFF
--- a/debug/vector/polyline.html
+++ b/debug/vector/polyline.html
@@ -1,0 +1,72 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>Leaflet debug page</title>
+
+    <link rel="stylesheet" href="../../dist/leaflet.css" />
+
+    <link rel="stylesheet" href="../css/screen.css" />
+    <script type="text/javascript" src="../../build/deps.js"></script>
+    <script src="../../dist/leaflet-src.js"></script>
+</head>
+
+<body>
+    <div id="mapid" style="width: 1000px; height: 700px; border: 1px solid #ccc"></div>
+
+    <script src="route.js"></script>
+    <script>
+        var mymap = L.map('mapid').setView([51.505, -0.09], 1);
+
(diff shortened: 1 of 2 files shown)


IvanSanchez:
OK, I pointed out a few things. I can summarize as:

Good things:

* The code works
* There is an example that can be tested out (this is awesome!)
* The solution goes above and beyond what I originally asked for (this actually *splits* linestrings at the antimeridian, whereas I thought about just adding +360/-360 to the longitudes).

Currently:
screenshot_20161219_134808

What I was thinking about:
screenshot_20161219_134744

What you implemented:
screenshot_20161219_134640

Clerical things:

* There is a merge conflict due to using an outdated `master` branch as the base for your fork. Either `git rebase` or `git merge` from the current Leaflet `master`. This is just a common thing that happens when lots of people use the same git repo concurrently.
* There are no unit tests. I don't think this feature can be merged without a couple of those.

Bad algorithmic things:

* I'm worried about hard-coding the wrapping bounds and the map projection (the calls to `Math.tan`) in the wrapping/splitting code. But this cannot be done when the `L.Polyline` is instantiated, because the polyline is not aware (and cannot be aware) of the map's CRS. This is IMHO a big concern.
* I'm not sure how this works for `L.Polygon`s. Remember, in Leaflet `L.Polygon` inherits from `L.Polyline` .
* This solution does not address the use case of drawing a **continuous** line across a ship route in the pacific ocean. I'm sure there are nice use cases for splitting, but I have good use cases for just offsetting (e.g. duplicating a vector renderer to create multiple copies of vector geometries while minimizing the rendering artifacts)

I'd approach the CRS problem by deferring the wrapping/splitting until the polyline has been added to the map the first time. Probably hook up the code inside the `_project()` private method.

By deferring this, you can use `this._map`, which means using `this._map.options.crs`, which means `this._map.options.crs.project()` and `this._map.options.crs.wrapLng` and `this._map.options.crs.wrapLat`.

*Maybe* `L.CRS` needs more stuff, like a method to return the size of the "wrapping box", or a method to return a `LatLng` which is less than half a "wrapping box" away from another `LatLng`.

How does this sound?

PR #5219, opened 2016-12-27
Add Leaflet.Viewpoint to plugins list

ggolikov:
Leaflet.Viewpoint extends L.CircleMarker and represents marker with multiple directions. Useful to show photos taken from one point.


ggolikov:
demo: https://ggolikov.github.io/Leaflet.Viewpoint/example/

PR #5225, opened 2017-01-01
prevent exception if no callback is provided

Benvorth:


perliedman:
Hi @Benvorth, can you provide a use case where `whenReady` is called without a callback, which makes this fix necessary?

Benvorth:
If you use `leaflet` with `react-leaflet` and `leaflet-routing-machine` you get an error on first loading the map:

    import React from 'react'
    import L from 'leaflet'
    import {Map, Marker, Popup, TileLayer} from 'react-leaflet'
    import Routing from './routing'
    
    export default React.createClass({
      render () {
        let zoom = 6
        let position = [51.1657, 10.4515]
        return(
          <Map center={position} zoom={zoom} >
            <TileLayer
              url='http://korona.geog.uni-heidelberg.de/tiles/roads/x={x}&y={y}&z={z}'
              attribution='Imagery: <a href="http://giscience.uni-hd.de/">GIScience Research Group</a> &mdash; Map data &copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>'
            />
            <Routing from={[57.74, 11.94]} to={[57.6792, 11.949]}/>
          </Map>
        )
      }
    })
    
`routing.js` (following https://github.com/PaulLeCam/react-leaflet/issues/92):

    import {MapLayer} from 'react-leaflet'
    import L from 'leaflet'
    import 'leaflet-routing-machine'
    import 'leaflet-routing-machine/examples/Control.Geocoder'

    export default class RoutingMachine extends MapLayer {

      componentWillMount() {
        super.componentWillMount()
        const {from, to} = this.props
        let waypoints = [
          L.latLng(from[0], from[1]),
          L.latLng(to[0], to[1]),
        ]
        this.leafletElement = L.Routing.control({
          plan: L.Routing.plan(waypoints, {
            createMarker: (i, wp) => {
              return L.marker(wp.latLng, {
                draggable: true,
                title: 'maker ' + i,
                icon: L.icon({
                  iconUrl: app.util.getMarkerIconFor(i),
                  iconSize:     [25, 41], 
                  iconAnchor:   [12, 41], 
                  popupAnchor:  [0, -45], 
                  shadowUrl: require('./img/leafletjs/marker-shadow.png'),
                  shadowSize:   [41, 41], 
                  shadowAnchor: [20, 41]
                })
              })
            },
            geocoder: L.Control.Geocoder.nominatim(),
            routeWhileDragging: true
          }),
          routeWhileDragging: true,
          routeDragTimeout: 250,
          showAlternatives: true,
          altLineOptions: {
            styles: [
              {color: 'black', opacity: 0.15, weight: 9},
              {color: 'white', opacity: 0.8, weight: 6},
              {color: 'blue', opacity: 0.5, weight: 2}
            ]
          },
          collapsible: true
        }).addTo(map)
      }

      render() {
        return null
      }
    }
2017-01-03 15_35_44-neue benachrichtigung

danzel:
I feel that the client code shouldn't be calling this method with `null`/`undefined`.

I don't see this is leaflets responsibility to handle. It seems like better behavior that you get an error rather than having an erroneous call silently handled.

perliedman:
I agree with @danzel, the problem is calling `whenReady` with an undefined callback.

@Benvorth If you don't know when this happens already, I would put a conditional breakpoint in `whenReady` to break when it's called without a callback.

seclace:
Hi there! I'm having the same error and it is very discrourage me in my progress. What am I doing wrong when adding MapLayer to map?

My `MapContainer.js` component:

    import React, { Component } from 'react';
    import { Map, Marker, Popup, TileLayer } from 'react-leaflet';

    import './style.css';

    import Routing from '../Routing';

    const position = [51.505, -0.09];

    class MapContainer extends Component {
      render() {
        return (
          <div id="map" className="map-container">
            <Map ref='map' className="map-container-inner" center={position} zoom={13}>
              <Routing map={this.refs.map} from={[57.74, 11.94]} to={[57.6792, 11.949]} />
              <TileLayer
                url='http://{s}.tile.osm.org/{z}/{x}/{y}.png'
                attribution='&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
              />
              <Marker position={position}>
                <Popup>
                  <span>A pretty CSS3 popup.<br/>Easily customizable.</span>
                </Popup>
              </Marker>
            </Map>
          </div>
        )
      }
    }

    export default MapContainer;

My `Routing.js` component:

    import { MapLayer } from 'react-leaflet';
    import L from 'leaflet';
    import 'leaflet-routing-machine';

    export default class Routing extends MapLayer {

      createLeafletElement(props) {
        const {map, from, to} = this.props;
        this.leafletElement = L.Routing.control({
          position: 'topleft',
          waypoints: [
            L.latLng(from[0], from[1]),
            L.latLng(to[0], to[1]),
          ],
        }).addTo(map);
      }

      // componentWillMount() {
      //   super.componentWillMount();
      //   const {map, from, to} = this.props;
      //   this.leafletElement = L.Routing.control({
      //     position: 'topleft',
      //     waypoints: [
      //       L.latLng(from[ 0 ], from[ 1 ]),
      //       L.latLng(to[ 0 ], to[ 1 ]),
      //     ],
      //   }).addTo(map);
      // }
    }

Nothing from this not working (createLeafletElement & componentWillMount) becouse the map is null after the moment of mounting MapContainer component, `ref` is not set yet but it been in props of `Routing.js`.
Becouse of this I've tried to get the map from context (as specified in docs) and then get the same error as in this issue. Screenshot here: https://yadi.sk/i/SUvwzeKJ3LgpWS

Could you give the community the correct way of creating this element (I think it's not really hard, but can't find the solution for now). Thanks for this library!

perliedman:
@seclace hi! Please Note that this is an issue tracker, not a support forum. Unfortunately we don't have the time and resources to dig into other peoples code, especially not when it involves a lot of other libraries which we might not have experience with - for example react-leaflet, in this particular case.

Having said that, looking at the screenshots, it looks like the `MapLayer` component from react-leaflet assumes that `leafletElement` is in fact a `Layer`, but what you actually set it to in your code is a `Control`. Later, this causes an error since controls don't have the internal method `_layerAdd`, which layers have.

So, given the naming (`MapLayer`), it might in retrospect not come as a surprise that it doesn't work with controls. I'm sorry to say I have no idea what the right way to approach this is in react-leaflet.

perliedman:
Thinking about this, we could make sure to throw an error if what's being added doesn't have a `_layerAdd` method.

PR #5226, opened 2017-01-04
Added z-index support to ImageOverlay

Saulzi:
Currently, the ImageOverlay Layer type does not support z-index like the LayerGroup / GridLayer,

This adds similar functionality to GridLayer.

Just after finishing this I have realized that I can ensure I have the correct z-index for all layers by ensuring that each individual layer is in its own pane.

I thought I would share as other's may find this functionality usefull.


almostcake:
Good day,
You are going to include this feature in next version, and how soon it will be released?
Thanks

Saulzi:
This still has not been merged. I have just created another PR which will conflict with this as it adds an additional event and missing documentation for the load event.

perliedman:
@Saulzi sorry that this has been left hanging. Could you resolve the conflict and rebase on `leaflet/master`, so that we can merge this?

perliedman:
@Saulzi also, it looks like you have made this PR from your `master` branch, so it now includes the commit from your other PR.

Please do not make PRs from `master`, since you are likely to add other commits, and they will also be part of your original PR. Always make PRs from a fresh branch.

So, for this to be merged, you must back out the last commit first.

IvanSanchez:
Superseded by #5418

PR #5247, opened 2017-01-12
add source to moveend and zoomend events

paulgreg:
As [described on leaflet.uservoice.com](https://leaflet.uservoice.com/forums/150880-ideas-and-suggestions-for-leaflet/suggestions/17371396-adding-origin-to-events), we need to refresh some markers when the map has moved or zoomed but only if it’s coming from user interaction, not from our code calling leaflet API.

Here‘s a pull request which adds a 'source' param to moveend and zoomend events.
The source can be either 'user' or 'leaflet', depending on the action that triggered the event.
To illustrate:
- The user drags the map, 'moveend' is triggered with source=user.
- Calling 'fitbounds' on a list of markers will trigger 'moveend'' with source=leaflet

The commit may not be clean enough to integrate but it should give you an idea of the problem we trying to solve and so, hopefully, start a discussion.

So what do you think ?

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -63,13 +63,13 @@ L.Control.Zoom = L.Control.extend({
 
 	_zoomIn: function (e) {
 		if (!this._disabled && this._map._zoom < this._map.getMaxZoom()) {
-			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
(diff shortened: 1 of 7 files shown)


adrien-delhorme:
I need to solve the same problem. How can I help to speed up the merge of this pull request?

IvanSanchez:
Superseded by #6929

PR #5248, opened 2017-01-12
Blind attempt to fix Edge dblclicks (side effects of #5185)

### LINKED ISSUE #5180: Duplicated 'dblclick' and phantom 'touchstart' events with Chrome >= 55

IvanSanchez:
Based on my observations and on the feedback given by @dtapuska, I now think that the right behaviour is to rely on native `dblclick`s whenever we're on something other than Edge **or** the pointer type is a mouse.

The rationale is that Edge fires native `dblclick`s when using a mouse, but not when using a touchscreen. Chrome always fires native `dblclick`s. IEMobile (not edge) always fires native `dblclick`s.

This should fix #5180 (or at least make the code behave better in more browsers)

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -17,7 +17,7 @@ L.extend(L.DomEvent, {
 			var count;
 
 			if (L.Browser.pointer) {
-				if (!L.Browser.edge && e.pointerType === 'mouse') {
+				if ((!L.Browser.edge) || e.pointerType === 'mouse') {
 					return;
 				}
 				count = L.DomEvent._pointersCount;


dtapuska:
@IvanSanchez  have you filed an issue against [Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/) in its tracking database? @Teddink @RByers

IvanSanchez:
@dtapuska No, because the behaviour of the `dblclick` events is really left up to the browser's implementation.

See the specification at https://www.w3.org/TR/DOM-Level-3-Events/#event-type-dblclick : it was written well before touchscreens were commonplace, so it doesn't address double-taps. Also, quoting:

> The definition of a double click depends on the environment configuration

So it's perfectly fine for the Edge guys to say "touches don't trigger dblclicks" for whatever reason they want.

dtapuska:
@IvanSanchez  Browser vendors need to know about these type of interops so that we can all agree on what is the correct behavior so web developers don't have to user agent detection for scenarios like this.

RByers:
We could also discuss this as a UI events spec issue?
IIRC Safari also doesn't send `dblclick`, not sure about Firefox.

PR #5249, opened 2017-01-12
Make sure path's pixel coords are up to date when updating

perliedman:
This is _yet another_ fix for #5170, I believe the simplest. From looking at the code, it seems to be an oversight that `_updatePaths()`, which is just an optimization for doing `_update(layer)` for all the canvas' layers, does not call `project(layer)`, which `_update(layer)` does.

Adding this fixes the provided example for #5170, and it's just a one line fix.


IvanSanchez:
Closing in favour of #5250.

PR #5261, opened 2017-01-17
Leafet.migrationLayer

Sylvenas:
leafet.migrationLayer is used to show migration data such as population,flight,vehicle,traffic and so on.Data visualization on map.


IvanSanchez:
Wow, this looks really cool! :+1:

PR #5264, opened 2017-01-18
New control positions (middle & center)

yigityuce:
Vertical center(middle) and horizontal center(center) positions are added to controls


IvanSanchez:
I'm going to ask for an update of the docstrings around https://github.com/Leaflet/Leaflet/blob/v1.0.2/src/control/Control.js#L14 , so the functionality will be up-to-date with the documentation.

Has this been tested?

yigityuce:
Oh sorry, i fixed it (https://github.com/yigityuce/Leaflet/commit/9882b8fb6dcac4a1717fea2a2d0acff1b97ce7c3)

Test results are like this:

turuncu@turuncu:/var/www/Leaflet/dist$ /opt/node-v6.9.2-linux-x64/bin/jake lint
Checking for JS errors...
	Check passed.

turuncu@turuncu:/var/www/Leaflet/dist$ /opt/node-v6.9.2-linux-x64/bin/jake build[3vvvvvv,custom]
Managing dependencies...
 * Core
 * EPSG3395
 * GridLayer
 * TileLayer
 * TileLayerWMS
 * ImageOverlay
 * Marker
 * DivIcon
 * Popup
 * Tooltip
 * LayerGroup
 * FeatureGroup
 * Path
 * Polyline
 * Polygon
 * Rectangle
 * CircleMarker
 * Circle
 * SVG
 * VML
 * Canvas
 * GeoJSON
 * MapDrag
 * MouseZoom
 * TouchZoom
 * BoxZoom
 * Keyboard
 * MarkerDrag
 * ControlZoom
 * ControlAttrib
 * ControlScale
 * ControlLayers

Bundling and compressing 67 files...
	Uncompressed: 371.91 KB (+59 bytes)
	Saved to dist/leaflet-custom-src.js
	Compressed: 141.61 KB
	Saved to dist/leaflet-custom.js
	Gzipped: 37.90 KB
turuncu@turuncu:/var/www/Leaflet/dist$ /opt/node-v6.9.2-linux-x64/bin/jake test
Checking for JS errors...
	Check passed.

Checking for specs JS errors...
	Check passed.

Running tests...

................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
..................
PhantomJS 2.1.1 (Linux 0.0.0): Executed 578 of 582 (skipped 4) SUCCESS (11.127 secs / 10.563 secs)
	Tests ran successfully.

yigityuce:
Hi @perliedman ,

Actually i couldn't understand what should i fix. Could you explain more please? Thanks for your review btw, hope it works for you and all 😄

perliedman:
@yigityuce the vertical alignment for `middlecenter` differs from `middleleft` and `middleright`.

For `middleleft` and `middleright`, the control is truely in the vertcial middle. For `middlecenter`, the control's _top_ is centered, the rest of the control is below the center. (At least when I tested.)

ghybs:
Hi @yigityuce,

I guess your `middlecenter` position just concatenates the `leaflet-center` and  `leaflet-middle` classes, in which case the latter will override the CSS `transform` of the former, making its value only `translateX(-50%)`, thus no longer vertically correcting for the control container height.

yigityuce:
Hi @perliedman and @ghybs 

I fixed the middlecenter position from top. Can you control and inform me please. 07a9d1f6c790e79e1d0176867367de556af7d7ca

iH8:
Awesome feature but please do not introduce a middle and a center. It will only confuse users. Stick to one of those. I would suggest center. [Example](https://embed.plnkr.co/qq3y9KctDDZ7AsZIKgzB/)

perliedman:
@iH8 the problem is that HTML/CSS uses the term "center" for horizontal center, but "middle" for vertical. You could argue that following established terminology is a good thing.

ghybs:
Hi,

If naming is so controversial, a possible solution would simply be to keep this functionality in a separate plugin. That way, Leaflet base is not bloated, and the plugin can provide any sugar you want.

After all, this functionality is only for a minority of people, and quite easy to patch (remember the SO post [How to locate leaflet zoom control in a desired position](https://stackoverflow.com/questions/33614912/how-to-locate-leaflet-zoom-control-in-a-desired-position))

ghybs:
Hi @yigityuce,

What do you think about exporting this functionality into a Leaflet plugin?
I have seen that you have already authored at least 1 plugin, so you might be in a good position to build it.

However since I mentioned that idea which _might_ have decreased the probability of your PR being merged, I am ready to spend the time to build that plugin, should you agree that I do it on your behalf.
Of course I can cite you as the author and transfer the repo ownership to you later on.

Please let me know what you think.

gassc:
+1 for this to be included *out-of-the-box.*

Having five additional positional options that follow the semantic logic of the existing options is more intuitive to me than having to add a plugin or patch for what seems like fundamental functionality.

I'm coming from a desktop GIS and graphic design background, and I suspect a lot of folks wading into web mapping and finding this library are as well&mdash;certainly not a minority of users. It'd be useful to have a least a *little bit more* freedom to experiment with positioning map elements.

That [other maps API](https://developers.google.com/maps/documentation/javascript/3.exp/reference#ControlPosition) uses "center" instead "middle" nomenclature, if that is the sticking point here. As an end user either way is fine with me as long as the documentation is clear; I don't think that should be the deal breaker in merging the PR.

Is there anything I can do to help?

perliedman:
Closed in favor of #5554.

PR #5265, opened 2017-01-18
Scrubbing of detached DOM elements, prevents memory leaks

### LINKED ISSUE #5263: Map isn't removed from memory on map.remove() - Causes memory leak
## How to reproduce

- Leaflet version I'm using: Leaflet 1.0/1.0.2
- Browser I'm using: Chrome Version 55.0.2883.87 
- Repro steps:
  a) Render map using GeoJSON.
  b) On button click - remove the existing map using map.remove() and initialize the map again (Using existing GeoJSON/new GeoJSON).
  c) Browser memory usage spikes every time the map is cleared and initialized again. (Check Windows Task Manager)

- Expected behavior: The old map should be disposed and the browser shouldn't hold the memory that was allocated for the old map.
- Actual behavior: Browser memory usage spikes/stacks every time the map is re-initialized

- Link to JSFiddle: https://jsfiddle.net/6wu97g1L/
- Repro steps:
  a) Open JSFiddle link in new browser window.
  b) Check initial browser memory usage in Task Manager
  c) Click on "Color map" button (2-3 times).
  d) Check Chrome browser memory usage in Task Manager after every click.
  e) Repeat steps (c) & (d)

Memory on initial load:

image

Memory after 2 clicks:

image

Memory after 4 clicks:

image

IvanSanchez:
Fixes #5263.

This scrubs all the DOM elements which would be become detached elements after a `map.remove()` call. It's necessary to both `L.DomUtil.remove()` the element and `delete` every reference to them, even some implicit closures that we made for convenience.

Also adds a new way to call `L.DomEvent.off()`, as to scrub all (known) native event handlers.


IvanSanchez:
Before: the `HTMLCanvasElement` becomes a detached element, leaking memory. Also observe five `HTMLDivElement`s for the map panes, and 8 for the control corners, animation proxy, and draggable proxy.

screenshot_20170118_172053

After: I've managed to scrub the detached elements, except for a reference to `L.Draggable` that I cannot locate right now.

screenshot_20170118_172134

mourner:
Hmm, I thought that if you remove an element from the DOM, and don't have any references to it left, all the children nodes will also be garbage-collected properly. Do we have to remove all the children like this too?

IvanSanchez:
Apparently not - I had to both remove the nodes from the DOM and `delete` their references in order for Chrome to garbage-collect them. If I skip removing some explicitly declared element from the parent, then remove the parent, the children is somehow kept in memory.

I might be wrong, though. This PR was a 20-minute hack anyway.

perliedman:
@mourner from a quick look, one problem we have is that adding mouse/touch listeners from a layer adds a reference from the map container to these layers, and they are not always properly cleaned up when the layer is removed; for example `Canvas` leaves a `dblclick` etc. from the map container. @IvanSanchez's fix addresses this.

As for the other changes, I haven't looked at them in detail, but I would expect similar patterns.

PR #5270, opened 2017-01-19
Remove trailing column in Object definition

batje:
The trailing column in the Layer default Options Object definition crashes IE8.


IvanSanchez:
Hhhmm, I wonder why the linting tools didn't catch this. :pensive:

PR #5271, opened 2017-01-20
add leaflet.magicMarker plugin

Sylvenas:
Adding magical animation effect to a marker while loading.


IvanSanchez:
@Sylvenas, I'm slightly disappointed by this, because whenever I read "magic" I expect sparkles and unicorns around!

:unicorn: :rainbow: :sparkles:

PR #5279, opened 2017-01-24
Explicitly stop propagation of wheel events in layers control

### LINKED ISSUE #5277: Сan not scroll list of layers in Google Chrome
- Browser (with version) I'm using: Chrome & FF

I can not scroll list of layers in layers control. Error in chrome 55, but all is well in Firefox 50. View gif.  In Leaflet 0.7.3 works fine.

Chrome 

Firefox

IvanSanchez:
Fixes #5277.

I don't know why this change in the behaviour of Chrome, though.

Dunno if it's worth to submit a bug report upstream. According to this behaviour, event handlers in a parent element take precedence over the default action on children elements.


IvanSanchez:
Superseded by #5280

PR #5280, opened 2017-01-25
Stop scroll propagation in L.Layers.Control in chrome>55

### LINKED ISSUE #5277: Сan not scroll list of layers in Google Chrome

IvanSanchez:
Fixes #5277.


wladich:
If it is fixed this way, it it will be necessary to fix similar usages in several other places (zoom buttons size, touch tolerances, ...) and in third-party plugins and custom components that rely on this flag.
In my opinion it is better to fix Browser.touch detection which was changed in
a1371507bfe898eac1dabbb34c8f7692195e74d9 and works incorrectly in last Chrome version.

a1371507bfe898eac1dabbb34c8f7692195e74d9 removes deprecated "navigator.pointerEnabled" condition from Browser.pointer flag. We can use navigator.maxTouchPoints mentioned in #5189.

src/control/Control.Layers.js
```diff
@@ -163,7 +163,7 @@ L.Control.Layers = L.Control.extend({
 		container.setAttribute('aria-haspopup', true);
 
 		L.DomEvent.disableClickPropagation(container);
-		if (!L.Browser.touch) {
+		if ((L.Browser.chrome && L.Browser.pointer) || !L.Browser.touch) {
```
perliedman: Maybe I'm missing something here, but why do we have this conditional at all? Why should scroll events propagate for mobile/touch but not for desktop/non-touch?

Issue #5278, opened 2017-01-24
L.Transformation factory


anetz89:
Hi there,

first of all - great job with your 1.0 release! I really appreciate your work!

As I use custom, projected maps with my own CRS, I have to dig deeply into your code and found out, that there's currently no factory function for L.Transformation like it is for custom elements like L.marker() for L.Marker.

Do you have plans to include factory functions like these? In case of the L.Transformation it would be helpful to also allow a passed array.

Currently I do it in my code, but maybe it is helpful for others, too?

    L.transformation = function(a, b, c, d) {
        if (L.Util.isArray(a)) {
            return L.transformation.apply(this, a);
        }
        return new L.Transformation(a || 1, b || 0, c || -1, d || 0);
    };

Regards
Markus

perliedman:
Hi, and thanks for this suggestion. Although maybe not the most common thing, it looks like a nice improvement of the API.

One thing: if we add this to Leaflet, the array logic should go into the constructor, and not only the factory, right?

Feel free to submit a pull request, and ideally add a unit test for this!

anetz89:
Added the array logic to the constructor. There might be some documentation issue with b, c and d as they get optional when an array is passed although they should not be optional when no array is passed. That's why I added default properties for a, b, c and d (corresponding to CRS.Simple).

I'm also not sure about the test, as it's quite basic. Feel free to adjust the code the way you like!

Please see: https://github.com/Leaflet/Leaflet/pull/5282

jieter:
@perliedman Seems this can be closed since #5282 is merged

PR #5282, opened 2017-01-25
added transformation factory

anetz89:
fulfilled my duty:
https://github.com/Leaflet/Leaflet/issues/5278


src/geometry/Transformation.js
```diff
@@ -9,21 +9,33 @@
...
+ * var transformation = L.transformation(2, 5, -1, 10),
  * 	p = L.point(1, 2),
  * 	p2 = transformation.transform(p), //  L.point(7, 8)
  * 	p3 = transformation.untransform(p2); //  L.point(1, 2)
  * ```
  */
 
 
-// factory new L.Transformation(a: Number, b: Number, c: Number, d: Number)
+// factory new L.Transformation(a: Number|Number[], b?: Number, c?: Number, d?: Number)
```
perliedman: @IvanSanchez I don't know Leafdoc well enough to say if this is the correct way to specify that `a` is now either `Number` or an array of `Number` - does it look ok to you?

src/geometry/Transformation.js
```diff
@@ -9,21 +9,33 @@
...
+    this._d = a[3];
+    return;
+  }
+	// standard behavior, not sure if default values are helpful? Documentation implies 
+	// that b, c and d are optional, so some kind of default value (or a better 
+	// documentation) is needed
+	this._a = a || 1;
+	this._b = b || 0;
+	this._c = c || -1;
+	this._d = d || 0;
```
perliedman: I'd say leave this unchanged - `b`, `c` and `d` should only be optional if `a` is an array. We should instead describe in the docs how it works.

src/geometry/Transformation.js
```diff
@@ -9,21 +9,33 @@
...
-// factory new L.Transformation(a: Number, b: Number, c: Number, d: Number)
+// factory new L.Transformation(a: Number|Number[], b?: Number, c?: Number, d?: Number)
 // Creates a `Transformation` object with the given coefficients.
 L.Transformation = function (a, b, c, d) {
-	this._a = a;
-	this._b = b;
-	this._c = c;
-	this._d = d;
+	// very conservative way to allow a only. not sure if needed.
+  if (L.Util.isArray(a) && a.length === 4 && !b && !c && !d) {
```
perliedman: Just `L.Util.isArray(a)` is fine here, no need to be overly cautious. If `a` is an array and you specify more args, things will break in surprising ways if implemented this way.

perliedman:
Also notice the broken tests, you have a couple of code style issues that need to be sorted out.

src/geometry/Transformation.js
IvanSanchez: Leafdoc is very lenient on these issues, so it might work.

However, I'd prefer the `@alternative` syntax, like [in the `L.latLng()` factory method](https://github.com/Leaflet/Leaflet/blob/release-v1.0.3/src/geo/LatLng.js#L95). Makes things much clearer.

anetz89:
Thanks for your reviews! Look forward to your next one :)

Have you thought about the usage of an editorconfig file already? this would make all contributer's life much easier. See http://editorconfig.org/
Most Editors support this config and adjust their settings according to the configuration.
I added the configuration that fulfills your requirements.

src/geometry/Transformation.js
anetz89: What about "a.length === 4"? If a.length <4, index-errors may appear which did not appear with the current state of the code.

perliedman:
:+1: Thanks for contributing this, and thanks for great work on following up on the details!

PR #5283, opened 2017-01-26
Proof of concept of named classes

IvanSanchez:
This is just a proof of concept of named class inheritance.

I've been looking at ES6 `class`es, and the only real nice thing is the classes are named. This PR is just a proof of concept, to see if we like the idea.

Before: all classes are named `NewClass` because of how `L.Class.Extend` works:

screenshot_20170126_113427

After: An `eval()` trick allows to give each subclass its own name:

screenshot_20170126_113204

I think this would be quite useful to devs while developing.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -10,21 +10,31 @@ import * as Util from './Util';
 
 export function Class() {}
 
-Class.extend = function (props) {
+// @function extend(className: String, props: Object): Function
+// [Extends the current class](#class-inheritance) given the properties to be included.
+// Returns a Javascript function that is a class constructor (to be called with `new`).
+// The given name is shown in the stack traces when debugging.
 
-	// @function extend(props: Object): Function
-	// [Extends the current class](#class-inheritance) given the properties to be included.
-	// Returns a Javascript function that is a class constructor (to be called with `new`).
-	var NewClass = function () {
+// @alternative extend(props: Object): Function
+// Extends the current class. The resulting constructor is unnamed, and will
+// show up as `NewClass` in the stack traces when debugging.
+Class.extend = function (className, props) {
 
+	if (props === undefined) {
+		props = className;
+		className = '';
+	}
+
+	className = className.replace(/\W/, '');	// Remove spaces and non-letters
+
+	var NewClass = eval('var NewClass = function ' + className + '() {' +
 		// call the constructor
-		if (this.initialize) {
-			this.initialize.apply(this, arguments);
-		}
+		'if (this.initialize){ this.initialize.apply(this, arguments);}' +
 
 		// call all constructor hooks
-		this.callInitHooks();
-	};
+		'this.callInitHooks();' +
+
+		'};NewClass;'); //
 
 	var parentProto = NewClass.__super__ = this.prototype;
 
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -45,7 +45,7 @@ import {Point} from '../../geometry/Point';
  */
 
 
-export var Polyline = Path.extend({
+export var Polyline = Path.extend('Polyline', {
 
 	// @section
 	// @aka Polyline options
--- a/src/layer/vector/SVG.js
+++ b/src/layer/vector/SVG.js
@@ -43,7 +43,7 @@ export var create = Browser.vml ? vmlCreate : svgCreate;
  * ```
  */
 
-export var SVG = Renderer.extend({
+export var SVG = Renderer.extend('SVG', {
 
 	getEvents: function () {
 		var events = Renderer.prototype.getEvents.call(this);
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -29,7 +29,7 @@ import {PosAnimation} from '../dom/PosAnimation';
  *
  */
 
-export var Map = Evented.extend({
+export var Map = Evented.extend('Map', {
 
 	options: {
 		// @section Map State Options


mourner:
I was also thinking whether we could rely on ES6 class syntax for that, e.g. create a `Class` ES6 class and then extending from it with the usual `class Path extends Class { ... }`. Apart from names, it's also really convenient to be able to use `super`.

IvanSanchez:
I thought about ES6 classes too. The main roadblock right now is [the lack of instance properties in the class constructor](https://stackoverflow.com/questions/22528967/es6-class-variable-alternatives/22986568#22986568). If you read the [grammar specification](http://wiki.ecmascript.org/doku.php?id=strawman:maximally_minimal_classes#class_bodies), you'll see that a ES6 class definition can only have methods, generators, getters and setters.

So while it's possible to do this:

```
class TileLayer extends GridLayer {
  createTile(coords) {...}  // Yay, class method
}
```

It's not possible to do this:

```
class TileLayer extends GridLayer {
  options = {subdomains:'abcd'}  // Parse error, there's no grammar for this
}
```

So yeah, ES6 classes are cool but we cannot specify options to be inherited, or properties which are to be applied to each instance, and Leaflet does that for all the subclassing.

:disappointed:

fuzhenn:
Hi, everyone

I'm a fun of Leaflet Class utilities, so I have tried to migrate those handy methods to standard ES6 classes.

Below is the codes and it works well till now, and I would like to share it here.

Hope it can be helpful. 😄 

```Class.js```
```javascript

export default class Class {
    constructor(options) {
        if (!this || !this.callInitHooks) {
            throw new Error('Class instance is being created without "new" operator.');
        }
        this.setOptions(options);
        this.callInitHooks();
    }

    callInitHooks() {
        const proto = Object.getPrototypeOf(this);
        this._traverseInitHooks(proto);
    }

    setOptions(options) {
        if (!this.hasOwnProperty('options')) {
            this.options = this.options ? Object.create(this.options) : {};
        }
        if (!options) {
            return this;
        }
        for (let i in options) {
            this.options[i] = options[i];
        }
        return this;
    }
    _traverseInitHooks(proto) {
        if (this._initHooksCalled) {
            return;
        }
        const parentProto = Object.getPrototypeOf(proto);
        if (parentProto._traverseInitHooks) {
            parentProto._traverseInitHooks.call(this, parentProto);
        }
        this._initHooksCalled = true;
        const hooks = proto._initHooks;
        if (hooks && hooks !== parentProto._initHooks) {
            for (let i = 0; i < hooks.length; i++) {
                hooks[i].call(this);
            }
        }
    }

    /**
     * add a constructor hook
     * @param {string|function} fn - a hook function or name of the hook function
     * @param {any[]} args           - arguments for fn
     * @static
     */
    static addInitHook(fn, ...args) {
        const init = typeof fn === 'function' ? fn : function () {
            this[fn].apply(this, args);
        };
        const proto = this.prototype;
        const parentProto = Object.getPrototypeOf(proto);
        if (!proto._initHooks || proto._initHooks === parentProto._initHooks) {
            proto._initHooks = [];
        }
        proto._initHooks.push(init);
    }

    /**
     * method for adding properties to prototype
     * @param  {object} props - additional instance methods or properties
     * @static
     */
    static include(...sources) {
        for (let i = 0; i < sources.length; i++) {
            extend(this.prototype, sources[i]);
        }
        return this;
    }

    /**
     * merge new default options to the Class
     * @param  {object} options - default options.
     * @static
     */
    static mergeOptions(options) {
        const proto = this.prototype;
        const parentProto = Object.getPrototypeOf(proto);
        if (!proto.options || proto.options === parentProto.options) {
            proto.options = proto.options ? Object.create(proto.options) : {};
        }
        extend(proto.options, options);
        return this;
    }
}
```

```ClassSepc.js```
```javascript
describe('Class', () => {
    it('addInitHook', () => {
        const D = class extends Class {
            constructor() {
                super();
            }
            addHandler(name, clazz) {
                if (!this.handlers) {
                    this.handlers = [];
                }
                var handler = this[name] = new clazz();
                this.handlers.push(handler);
            }
        };
        const E = class extends D {
        };
        const F = class extends E {
        };

        const EHandler = class extends Class {
            getV() {
                return 'e';
            }
        };

        const FHandler = class extends Class {
            getV() {
                return 'f';
            }
        };

        E.addInitHook('addHandler', 'e', EHandler);
        F.addInitHook('addHandler', 'f', FHandler);
        const d = new D();
        const e = new E();
        const f = new F();
        expect(d.handlers).to.be(undefined);
        expect(e.handlers.length).to.eql(1);
        expect(f.handlers.length).to.eql(2);
        expect(e['e'].getV()).to.eql('e');
        expect(f['f'].getV()).to.eql('f');
    });

    it('include', () => {
        const D = class extends Class {
            constructor() {
                super();
            }
        };
        D.include({
            foo(bar) {
                return bar;
            }
        });
        expect(new D().foo('bar')).to.be.eql('bar');
    });

    it('mergeOptions', () => {
        const D = class extends Class {
        };
        D.mergeOptions({
            'foo' : 'd',
            'bar' : 'd'
        });

        const E = class extends D {
        };

        E.mergeOptions({
            'foo' : 'e'
        });

        const d1 = new D();
        const d2 = new D();
        d2.config('foo', 'd2');
        const e = new E();
        expect(d1.options['foo']).to.be.eql('d');
        expect(d1.options['bar']).to.be.eql('d');
        expect(d2.options['foo']).to.be.eql('d2');
        expect(d2.options['bar']).to.be.eql('d');
        expect(e.options['foo']).to.be.eql('e');
        expect(e.options['bar']).to.be.eql('d');
    });

    it('throws an error when creating without new operator', () => {
        expect(() => { Class(); }).to.throwException();
    });

    it('call parent\'s method by super', function () {
        const D = class extends Class {
            constructor() {
                super();
            }

            foo(name, gender) {
                return name + ',' + gender;
            }
        };
        const E = class extends D {
            constructor() {
                super();
            }

            foo2() {
                return super.foo.apply(this, arguments);
            }
        };
        expect(new E().foo2('bar', 'male')).to.be.eql('bar,male');
    });
});
```

mourner:
@IvanSanchez I think there's currently a ES proposal for property initialization in the class syntax; there might even be a Rollup/Buble plugin for this, not sure.

IvanSanchez:
@mourner I think you mean https://github.com/tc39/proposal-class-public-fields ?

Please note that:
* The syntax is pretty much TypeScript
* That's for ES7 / ES2016, **not** ES6
* It's a *proposal* at the *draft* stage, so do not expect support from runtimes or transpilers (https://kangax.github.io/compat-table/esnext/#test-class_properties)

jonkoops:
@mourner What is your opinion about using ECMAScript classes at this point? Are there any plans to make them a 1st class citizen like the custom class mechanism Leaflet introduces?

johnd0e (approved):

johnd0e:
I like the idea very much

Issue #4844, opened 2016-08-25
Pluggable TileLayer canvas renderer


IvanSanchez:
#4637 (1px gap between tiles) is difficult, and has no fix in sight.

An idea is to use a `<canvas>` to composite the tiles together, in order to apply the zoom animations or fractional zoom scaling to the whole canvas, not to the individual tiles.

This doesn't seem trivial and would need something like:
- Detecting the size of the canvas needed - up to double the map's size to account for `n+0.5` fractional zoom levels
- Make `TileLayer`s have an option for different renderers - DOM (current) and CanvasTile or something like that.
- Fall back to DOM renderer if the `crossOrigin` option is set to `false`.
- Set the "canvas tile origin" to the lower X,Y of the visible tiles
- On tile load, dump the tile bitmap into the canvas into its appropiate location.
- Ponder how to do tile fade-in properly. `requestAnimationFrame` + dumping bitmap each frame???
- On map move, check if the canvas is no longer overlapping the map's visible bounds; shift its contents and update. Perhaps it's possible to do this less than once per (tile size) pixels moved.
- Most importantly, hope that performance is OK-ish.

rnixx:
@IvanSanchez You implemented a plugin with a fix (https://github.com/Leaflet/Leaflet.TileLayer.NoGap). Is there a plan that this code gets integrated into leaflet core?

Falke-Design:
@IvanSanchez do you think the plugin [L.TileLayer.Canvas](https://github.com/GIAPspzoo/L.TileLayer.Canvas) is enough to close this issue?

skerit:
> @IvanSanchez do you think the plugin [L.TileLayer.Canvas](https://github.com/GIAPspzoo/L.TileLayer.Canvas) is enough to close this issue?

That plugin draws the individual tiles to their own canvas element. So instead of having a lot of img elements, you have a lot of canvas elements. That might be better for performance, but it doesn't help with the gaps.

IvanSanchez:
Superseded by #9732

PR #5290, opened 2017-01-31
New PixelGrid class. Implements #4844, fixes #4637 and #3575.

### LINKED ISSUE #4637: 1px lines while animating in 1.0
There are 1px horizontal and vertical lines appearing while animating pan or zoom in 1.0. 
In 0.7 these are not present.

master:
http://playground-leaflet.rhcloud.com/jeqig/1/edit?html,output

0.7:
http://playground-leaflet.rhcloud.com/jeqig/2/edit?html,output

Present in Chrome and Firefox, so it cannot be the same as the fractional zoom bug. More visible on a black background.

IvanSanchez:
Fixes the dreaded "half-pixel gap between tiles at fractional zoom" bug from #4637 / #3575 by adding a new class which sits between `GridLayer` and `TileLayer`, as described by #4844.

Code is basically https://github.com/Leaflet/Leaflet.TileLayer.NoGap , refactored as ES6 modules. Comments from #4941 apply to this PR.

The downside is that this new class adds about ~6 KiB~ 4KiB of unminified code / ~4 KiB~ 3KiB of minified code / ~0.5 KiB~ 0.45 KiB of gzipped code. ~I don't like rewriting the `GridLayer` methods just to add one more line of code inside a loop, but I see no other alternative right now (or should the `GridLayer` make a check for `dumpToCanvas`?)~

### DIFF
--- a/build/docs-index.leafdoc
+++ b/build/docs-index.leafdoc
@@ -9,6 +9,7 @@ This file just defines the order of the classes in the docs.
 @class Popup
 @class Tooltip
 
+@class PixelGrid
 @class TileLayer
 @class TileLayer.WMS
 @class ImageOverlay
--- a/build/leafdoc-templates/html.hbs
+++ b/build/leafdoc-templates/html.hbs
@@ -47,6 +47,7 @@
 			</ul>
 			<h4>Raster Layers</h4>
 			<ul>
+				<li><a href="#pixelgrid">PixelGrid</a></li>
 				<li><a href="#tilelayer">TileLayer</a></li>
 				<li><a href="#tilelayer-wms">TileLayer.WMS</a></li>
 				<li><a href="#imageoverlay">ImageOverlay</a></li>
--- a/debug/map/pixelgrid.html
+++ b/debug/map/pixelgrid.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script src="../leaflet-include.js"></script>
+	<style>
+		#map {
+			position: absolute;
+			width: 100vw;
+			height: calc( 100vh - 5em );
+			bottom: 0px;
+
+			background: cyan;
+		}
+		body {margin: 0}
+	</style>
+</head>
+<body>
+	There should be no gaps between the tiles.
+
+	<div id="map" class="map"></div>
+
+	<script type="text/javascript">
+		var mapopts =  {
+			center: [35, -122],
+			zoom : 5.5,
+			zoomSnap: 0.1,
+			zoomDelta: 0.1,
+		};
+
+		var map = L.map('map', mapopts);
+
+		var cartoAttrib = '&copy; <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>, &copy;        <a href="https://carto.com/attribution">CARTO</a>';
+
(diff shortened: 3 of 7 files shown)


IvanSanchez:
I've left a playground at https://playground-leaflet.rhcloud.com/mizuv (https://playground-leaflet.rhcloud.com/mizuv/edit?html,output to see source code).

themre:
Sample of still showing gaps: 
leafletgaps

IvanSanchez:
Yeah, those gaps are due to the fade-in animation of the tiles. Tiles are not stitched into the `<canvas>` until they become fully opaque after 250msec. Stitching a fading tile would require extra logic, and dumping pixels at every frame.

hyperknot:
Impressive work!
Just some feedback:
- It works really great on both Chrome and Safari!
- The biggest problem seems to me is animation performance. It can be a bit annoyingly jerky / non-smooth on wheel zoom, which is not present in the legacy code. Can we debounce / throttle it to only run after an animation has finished?
- Afaik, the issue was not present in Firefox. Is the IE condition there because IE is also non-affected, or because it has some problems with canvas? I recommend adding Firefox as well to the condition. But it might require user agent detection, or something reliable to detect Firefox, as I don't know any way to detect if this bug is present in a browser.
- Are we not opening the can of CORS now?

IvanSanchez:
@hyperknot Thanks for having a look :smile: 

> Can we debounce / throttle it to only run after an animation has finished?

This is a possibility. But it will make the "gaps during zoom animation" problem (shown in a gif above) worse. I can try that and see how it performs.

> Is the IE condition there because IE is also non-affected, or because it has some problems with canvas?

IE performance about anything `<canvas>`-related is.... suboptimal, *to say the least*.

> I recommend adding Firefox as well to the condition.

Not sure. PixelGrid also gets rid of some minor rendering artifacts when showing tiles past `maxNativeZoom` (antialiasing at the tile edges). If a developer wants to disable it in Firefox it should be a simple matter of...

```
L.tileLayer(url, {
  dumpToCanvas: !L.Browser.ie && !L.Browser.gecko
})
```

> Are we not opening the can of CORS now?

Impressively enough, no, we're not!! PixelGrid uses [`drawImage`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage) instead of [`putImageData`](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/putImageData). `getImageData`/`putImageData` involve a lot of CORS restrictions, but I'm very glad that `drawImage` does not. This means that PixelGrid works with tilesets which don't have CORS headers on them.

The only side effect is that the canvas becomes tainted and the JS code cannot query pixels from the pixelgrid, but that's negligible for the use cases we're dealing with.

hyperknot:
You are right, even while zooming the pixel bug is much less visible. It's still there sometimes, but I'd say something like 10-20% of the original. 

So I guess it's a matter of performance vs. display quality here. I'm interested in other's opinion as well. It'd be a hard sell to have a new version which is slower than the old one. Also it might be interesting to see how the new method works on an older Android phone for example.

IvanSanchez:
FYI, I considered other options such as offscreen 2D canvases, and WebGL. But browser support, code complexity and CORS limitations are big obstacles for those solutions.

hyperknot:
@mourner as this one is affecting performance, what is your opinion?

mourner:
I still think this should be a plugin rather than be in the core. Getting rid of the gaps is really an edge issue that should be irrelevant for 99% of Leaflet users.

IvanSanchez:
How about keeping the `_onOpaqueTile`, `_onUpdateLevel`, `_onRemoveLevel` and `_onCreateLevel` hook points?

The reason I wanted to put this in the core was to not duplicate the tile/level pruning algorithms... in its current form, [Leaflet.TileLayer.NoGap](https://github.com/Leaflet/Leaflet.TileLayer.NoGap) duplicates the logic and it would be hard to keep bugfixes up-to-date.

IvanSanchez:
Superseded by #5476

Issue #5293, opened 2017-01-31
Errors when enabling dragging with a marker that is not on the map


danzel:
Found in https://github.com/Leaflet/Leaflet.markercluster/issues/755

If you enable dragging on a marker that has been removed from the map, it will fail.

Reproducing case:
https://playground-leaflet.rhcloud.com/jaro/edit?html,output

Not sure if we should be doing some more checks inside leaflet, or if the calling application shouldn't do this to layers that are not on the map.

danzel:
I'm taking a look in to this now.

PR #5295, opened 2017-01-31
Remove marker.dragging when not on the map

danzel:
Fix for #5293 
Not 100% sure this is the right way to do this, but it is probably the easiest ;)

An issue we have now is that interacting with the dragging object while the marker is not on the map is not really valid as the dragging handler depends on the icon, which doesn't exist while not on the map.

This is probably a breaking change, although I don't expect it to actually affect many users.


danzel:
Added a note for the docs. Can't get jekyll to work on windows, so not 100% sure the formatting is right.

PR #5303, opened 2017-02-03
Add CSS for -webkit-tap-highlight-color, fixes #5302

### LINKED ISSUE #5302: Chrome>55 highlights SVG paths on tap
As seen in https://stackoverflow.com/questions/42011444/how-to-avoid-leaflet-path-from-getting-selected-on-touch-devices

Environment: Chrome >= 55, latest Leaflet code from master.

How to reproduce: 
* Create a map with at least one SVG path or marker
* Use a touchscreen device, or Chrome's device emulation
* Click on the path, or make a drag less than 3 pixels (easier if no popups or interactions on the path)
* Somehow the interaction gets registered as a selection of the SVG path, highlighting it

screenshot_20170203_094007

IvanSanchez:
fixes #5302

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -60,6 +60,12 @@
 	-ms-touch-action: none;
 	touch-action: none;
 }
+.leaflet-container {
+	-webkit-tap-highlight-color: transparent;
+}
+.leaflet-container a {
+	-webkit-tap-highlight-color: rgba(51, 181, 229, 0.4);
+}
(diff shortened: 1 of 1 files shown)


IvanSanchez:
> If it's just SVGs, maybe we can target just those - or is this also a problem for other elements?

As far as I could tell, elements such as marker ``s and the zoom controls were also affected. :-(

That's why I'm adding the rule to the whole container.

PR #5304, opened 2017-02-03
Fix popup "autoClose" and "closeOnClick" options documentation, fixes #5040

### LINKED ISSUE #5040: Popup.options.closeOnClick documentation missing, autoClose documentation wrong
As far as I can tell, the `closeOnClick` option of `Popup` closes the popup when clicking somewhere else on the map (equivalent to `Map.option.closePopupOnClick`). The `autoClose` option closes the popup when another popup is opened.

The documentation of the `closeOnClick` option is missing, and the `autoClose` documentation instead describes the behaviour of the `closeOnClick` option.

zemadz:
"autoClose" documentation was actually about "closeOnClick" and "autoClose" documentation was missing.

### DIFF
--- a/docs/reference-1.0.0.html
+++ b/docs/reference-1.0.0.html
@@ -2916,13 +2916,19 @@ off of the screen while it is open.</td>
 		<td><code>true</code></td>
 		<td>Controls the presence of a close button in the popup.</td>
 	</tr>
+	<tr id='popup-closeonclick'>
+		<td><code><b>closeOnClick</b></code></td>
+		<td><code>Boolean</code>
+		<td><code>*</code></td>
+		<td>Set it if you want to override the default behavior of the popup closing when user clicks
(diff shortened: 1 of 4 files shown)


zemadz:
I've changed the docstrings in the code for `autoClose`. Also I noticed that I had missed previously that `closeOnClick` doesn't have a default "option" value, instead it is based on maps `closePopupOnClick`. Is "*" the correct "default" value in such case?

IvanSanchez:
Yeah, we use `*` in other bits of the documentation when the default value depends on something.

Please read https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#api-documentation , and make sure that the generated documentation looks as it should. I see no docstrings for `closeOnClick` in `src/layer/Popup.js` yet.

IvanSanchez:
Looks good! Thanks!

Issue #5281, opened 2017-01-25
Feature: add a 'error' event to ImageOverlay


IvanSanchez:
As per https://stackoverflow.com/questions/41841381/leaflet-imagelayer-seturl-error-handling , it should be useful to have a `error` event in `L.ImageOverlay`, akin to the `tileerror` event in `L.TileLayer`.

AP-whitehat:
Looking into this issue.
The problem is in 
https://github.com/Leaflet/Leaflet/blob/master/src/layer/ImageOverlay.js
I suppose ... ?

IvanSanchez:
Yeah. See:

https://github.com/Leaflet/Leaflet/tree/ad75456fa54aee4d38e92dd7654128819cf8e31a/src/layer/ImageOverlay.js#L167
and also:
https://github.com/Leaflet/Leaflet/tree/ad75456fa54aee4d38e92dd7654128819cf8e31a/src/layer/tile/TileLayer.js#L144

perliedman:
@AP-whitehat yes, exactly. From a quick look, an error handler should probably be added in [`_initImage`](https://github.com/Leaflet/Leaflet/blob/master/src/layer/ImageOverlay.js#L160).

AP-whitehat:
What would be better to use as a dummy placeholder  for error links
http://www.birthdaydirect.com/popup_image.php?pID=16068
or http://hdimagesnew.com/wp-content/uploads/2016/09/404.png
or something else .

I'll have to include the png in dist/images/

AP-whitehat:
Issue not yet Fixed ,
am I going in the right direction ?

please point out my mistakes 😛

IvanSanchez:
@AP-whitehat Please open up a pull request, so we can comment on specific lines of your code :-)

AP-whitehat:
@IvanSanchez 
Please Review my PR

rhberro:
A pull request related to this issue was merged, can we close this one?

perliedman:
Closed by #5307!

PR #5307, opened 2017-02-04
Error handler in ImageOverlay for 404 links

AP-whitehat:
This fix isn't working as of now ,  Needs some improvements .
Issue : https://github.com/Leaflet/Leaflet/issues/5281

### DIFF
--- a/src/layer/ImageOverlay.js
+++ b/src/layer/ImageOverlay.js
@@ -39,11 +39,11 @@ export var ImageOverlay = Layer.extend({
 
 		// @option crossOrigin: Boolean = false
 		// If true, the image will have its crossOrigin attribute set to ''. This is needed if you want to access image pixel data.
-		crossOrigin: false
+		crossOrigin: false,
 
 		// @option errorOverlayUrl: String = ''
 		// URL to the overlay image to show in place of the overlay that failed to load.
-		errorOverlayUrl: '',
+		errorOverlayUrl: ''
 	},
 
 	initialize: function (url, bounds, options) { // (String, LatLngBounds, Object)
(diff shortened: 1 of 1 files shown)


AP-whitehat:
Thanks 
I'll remove any unnecessary changes . 
I'll also try adding auto test for this .

AP-whitehat:
Added Error handler in L.ImageOverlay similar to that in L.TileLayer.

Question : 
Are the @options updated in realtime 
or do I have to set a default value for them 
see [https://github.com/Leaflet/Leaflet/pull/5307/files?diff=unified#diff-4c0b0ad9071509997408374e182dc111R46](url)

Suggest improvements @perliedman

perliedman:
Sorry for not following up on this sooner! Merged now, thanks for contributing! :+1:

src/layer/ImageOverlay.js
```diff
@@ -165,6 +169,7 @@ export var ImageOverlay = Layer.extend({
 		img.onmousemove = Util.falseFn;
 
 		img.onload = Util.bind(this.fire, this, 'load');
+		img.onerror = Util.bind(this._overlayOnError, this, 'error');
```
crabmusket: @perliedman @APwhitehat why is `'error'` being passed to `_overlayOnError`?

perliedman: To avoid the pretty weird case where the error image itself 404s. `TileLayer` has the same logic, and I think this is more or less a copy of that.

APwhitehat: Yes, I basically mimicked the `'error'` from `TileLayer`.
I thought it was meant to be default error decription.

crabmusket: It looks like `_overlayOnError` takes no arguments so I think that `'error'` string will just disappear into the aether! https://github.com/Leaflet/Leaflet/blob/61ff641951fa64ba4730f71526988d99598681c8/src/layer/ImageOverlay.js#L231

PR #5309, opened 2017-02-04
Removed type attribute in HTML5 files

uzerus:
Removed unneeded type attribute in <script> form in .html files.

type attribute is not needed in HTML5 cos its default scripting language in html


perliedman:
Hi! Thanks for taking the time to clean this up! It's a detail, but always good to remove unnecessary stuff.

I manually squashed and merged this in ce6975057f60c5769acf1b57765c6b927cf36d19.

PR #5329, opened 2017-02-13
Add index.js files to subdirectories/namespace

simon04:
This allows to streamline the import/export in the main file.

This PR incorporates the suggestions from https://github.com/Leaflet/Leaflet/pull/4989#pullrequestreview-18713630.


IvanSanchez:
It's good to see this as a PR, but it seems that the unit tests are failing. Apparently the exports are not as they should be in the bundle :-(

simon04:
Interesting. The exports in `dist/leaflet-src.js` are the same (modulo some reordering). When using the built file in `karma.conf.js`, the tests pass:

```patch
diff --git a/spec/karma.conf.js b/spec/karma.conf.js
index ad817795..259497e4 100644
--- a/spec/karma.conf.js
+++ b/spec/karma.conf.js
@@ -10,7 +10,7 @@ module.exports = function (config) {
                "spec/sinon.js",
                "spec/expect.js",
 
-               "src/Leaflet.js",
+               "dist/leaflet-src.js",
 
                "spec/after.js",
                "node_modules/happen/happen.js",
```

simon04:
Using [karma-rollup-plugin](https://github.com/TrySound/karma-rollup-plugin) instead of karma-rollup-preprocessor fixes the tests, see ccb6573.

IvanSanchez:
> Using `karma-rollup-plugin` instead of `karma-rollup-preprocessor` fixes the tests

This kind of stuff never ceases to amaze me. :open_mouth:

PR #5330, opened 2017-02-13
Provide missing SVG static methods.

jkuebart:
Currently, `SVG.create()` and `SVG.pointsToPath()` are unavailable when imported as a module.

### DIFF
--- a/src/layer/vector/index.js
+++ b/src/layer/vector/index.js
@@ -1,6 +1,9 @@
 export {Renderer} from './Renderer';
 export {Canvas, canvas} from './Canvas';
-export {SVG, svg} from './SVG';
+import {SVG, create, pointsToPath, svg} from './SVG';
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This looks good to me, but I also want to ask...

Should we have tests for checking if the exported names are as expected? I mean, a "unit test" just to check that every expected exported symbol exists in the default bundle. With the modules architecture, it sounds like a nice thing to have. We can take it in other issue if we want.

jkuebart:
An easy way to get this might be to write the unit tests using »fine-grained« ES5 imports, i.e. without '*' or default.

IvanSanchez:
Yeah, you're right, we're using the whole `L.Foo.bar()` namespaces in the tests. If something is not caught by the tests, it's because there is no explicit call for it.

PR #5331, opened 2017-02-14
Do not stop keypress on escape if no popup is open

### LINKED ISSUE #5234: Keyboard handler does not propagate escape key events
When using Leaflet with [keyboard navigation enabled](http://leafletjs.com/reference-1.0.2.html#map-keyboard), the propagation of escape keydown events is stopped by [`L.Map.Keyboard`](https://github.com/Leaflet/Leaflet/blob/63fd4edc76893ab2a2f83d54e703e0a4da73de7b/src/map/handler/Map.Keyboard.js#L159-L166):

```js
_onKeyDown: function (e) {
  var key = e.keyCode;
  if (...) {
    // ...
  } else if (key === 27) { // 27 is escape key's keyCode
    map.closePopup();
  } else {
    return;
  }
  L.DomEvent.stop(e);
}
```

I need to capture escape key events in my application, but creating a Leaflet [`keypress`](http://leafletjs.com/reference-1.0.2.html#map-keypress) handler with `map.on` — or using `map._container.addEventListener` directly — does not work.

Is this intentional? Why was the check for [`map._popup`](https://github.com/Leaflet/Leaflet/commit/1eae1719bb51cef00382b882ec368c5238c873e7) removed in 1eae1719bb51cef00382b882ec368c5238c873e7?

IvanSanchez:
Fixes #5234 by reverting 1eae1719bb51cef00382b882ec368c5238c873e7.

As a bonus, some unit tests for keyboard handler.

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -92,6 +92,7 @@
 	<!-- /map/handler -->
 	<script src="suites/map/handler/Map.DragSpec.js"></script>
 	<script src="suites/map/handler/Map.TouchZoomSpec.js"></script>
+	<script src="suites/map/handler/Map.KeyboardSpec.js"></script>
 
 	<!-- /geo/crs -->
 	<script src="suites/geo/CRSSpec.js"></script>
--- a/spec/suites/map/handler/Map.KeyboardSpec.js
+++ b/spec/suites/map/handler/Map.KeyboardSpec.js
@@ -101,6 +101,36 @@ describe("Map.Keyboard", function () {
 		});
 	});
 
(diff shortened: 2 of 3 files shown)


IvanSanchez:
ping @snkashis @perliedman

pshassett:
The esc key still doesn't seem to raise a `keypress` event in chrome 75.0 or firefox 68.0.  
```js
map.on('keypress', function(e){
var event = e.originalEvent
alert(event.key)
});
```
Am I missing something in my implementation or could this a browser-based problem?

PR #5336, opened 2017-02-15
plugins: Add Leaflet.Zoominfo

flaviocarmo:
Adding my leaftlet plugin to the plugin list


IvanSanchez:
I cannot run the examples, because in your example `https://flaviocarmo.github.io/Leaflet.zoominfo/examples/L.Control.Zoominfo.js` is returning 404.

Can you please fix that? We'd like to have a look at a working demo before adding the plugin to the list.

flaviocarmo:
Fixed. Sorry, i was updating other links and missed this.

IvanSanchez:
Looks interesting :+1: 

I suggest adding a couple of small screenshots to your readme, to make even more clear what the plugin does.

Issue #5144, opened 2016-11-28
Remove tolerance on mouse events in IE/Edge mouse+touchscreens


Ernie23:
## How to reproduce

- Leaflet version I'm using: Leaflet 1.0.2
- Browser (with version) I'm using: IE11 and Edge
- OS/Platform (with version) I'm using: Windows 10

1. Important: set map option preferCanvas on true
2. Loading GeoJson layer with options
    - onEachFeature: register mouseover and mouseout event on layer
3. Set map on Zoomlevel 15

## What behaviour I'm expecting and which behaviour I'm seeing

When I move the mouse close to the border, I expect the correct area to highlight immediately.
Right now, I have to move the mouse a bit further into the area. The best way to see this is at zoom 15+.

The left area should be highlighted and not the right one, because the cursor moves over the area.
![hittest](https://cloud.githubusercontent.com/assets/5499652/20664863/5aad5a52-b55d-11e6-9ce7-fac900531b81.PNG)

## Minimal example reproducing the issue
Sample:  https://jsfiddle.net/qd68r2wt/12/

perliedman:
Hi, and thanks for your report.

I think is because your browser supports touch.

For touch enabled browsers, Leaflet sets click tolerance to at last 10 pixels (https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Path.js#L134).

This click tolerance is later used when determining if a point is on the edge of a polygon (https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L476 and https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L436).

Forcibly disabling touch support makes this error go away for me: https://jsfiddle.net/uL565rks/

The increased click tolerance makes sense for things like points, but I agree it gives an unexpected and undesired effect in cases like this. We should probably look into improving this.

Ernie23:
That's interesting.
Can you tell me, why Leaflet thinks that IE supports touch?
How does the leaflet decide?

I've no touch hardware connected with my pc. 
Chrome and Firefox works fine without setting L.Browser.touch = false.

perliedman:
@Ernie23 the code is your friend, remember this is open source. Here's the code that checks for touch support: https://github.com/Leaflet/Leaflet/blob/master/src/core/Browser.js#L41

IvanSanchez:
[...]

PR #5339, opened 2017-02-16
Add an option to L.Path to allow the user to set the clicktolerance of a single layer

geoloep:
This is a suggestion for a new path option. It will only affect Canvas layers.

On non-touch browsers the clickTolerance is set to the weight of the stroke. In my real world projects the weight of polylines is way to small to create a usable clickable area. Current available options to alleviate this issue is to increase the weight or draw a second invisible polyline with increased weight. Both have their drawbacks.

It would be convenient to set a buffer area around the polyline that is also clickable. Since the _clickTolerance function is already present the only required change is to allow the user to influence the result of this function. By adding this as an path option, only specific layers will be affected.

I haven't contributed code to a project such as Leaflet before, the additions are mainly to show what I had in mind. I hope it is clear. I can expand on this request if necessary.

[Demo](http://demo.geoloep.nl/clickTolerance/)


IvanSanchez:
> I haven't contributed code to a project such as Leaflet before, the additions are mainly to show what I had in mind.

The code and the PR is actually good, don't worry about that :-D

> By adding this as an path option, only specific layers will be affected.

I'm not fully sure this is the best way? In my head, it would be a bit cleaner to set a `clickTolerance` option in the `L.Canvas` renderer, e.g.

```
return (this.options.stroke ? this.options.weight / 2 : 0) + (touch ? this._renderer.options.clickTolerance || 0 : 0);
```

> By adding this as an path option, only specific layers will be affected.

What's your use case for a per-layer click tolerance?

geoloep:
> What's your use case for a per-layer click tolerance?

I can imagine the additional clicktolerance would be annoying on polygons. Especially when multiple polygons touch each other. I came across this relevant issue: #5144. Closely clustered circleMarkers would give the same problem. There isn't really a use case on layers other than polylines I suspect.

I myself would only need added tolerance on certain polylines, and then the required tolerance might also depend on the weight of the line and perhaps the zoom level where I expect interaction to occur. I'm a developer that likes to have fine grained control over these matters and will test which setting "feels" just right.

When I want to make a change to one of my layers I would argue that it is cleaner to make it on the layer level than to apply it application wide via the renderer (and deal with the side effects if they occur).

geoloep:
I created a plugin that realizes the same functionality as I intended with this pull request: [Leaflet.Clicktolerance](https://github.com/geoloep/Leaflet.ClickTolerance).

IvanSanchez:
@geoloep Cool!! Please remember to add the plugin to the plugins list.

ValentinH:
I came across this issue while using paths on Canvas. After looking at the code, I'm wondering why the click detection doesn't use the `tapTolerance` option for the touch case?

PR #5351, opened 2017-02-21
Have uglifyjs spit out a sourcemap for leaflet.js too.

### LINKED ISSUE #5341: Source map for minified leaflet.js
There is a source map for the non-minified leaflet-src.js, but nothing for the minified leaflet.js. Even if such a source map simply mapped back to leaflet-src.js rather than the "true" source I would still find it useful, as I'm sure would many others.

danzel:
Fixes #5341 
The resulting source map points back to the original files too :+1:

### DIFF
--- a/build/publish.sh
+++ b/build/publish.sh
@@ -18,7 +18,7 @@ npm run-script build
 
 echo "Creating git tag v$VERSION..."
 
-git add dist/leaflet-src.js dist/leaflet.js dist/leaflet-src.map -f
+git add dist/leaflet-src.js dist/leaflet.js dist/leaflet-src.js.map dist/leaflet.js.map -f
 
 git commit -m "v$VERSION"
 
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "lintfix": "eslint src --fix; eslint spec/suites --fix; ",
     "rollup": "rollup -c build/rollup-config.js",
     "watch": "rollup -w -c build/rollup-watch-config.js",
-    "uglify": "uglifyjs dist/leaflet-src.js -c -m -o dist/leaflet.js"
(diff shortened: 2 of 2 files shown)


danzel:
re my second commit.
I noticed that the resulting map file for `-src` is called `leaflet-src.js.map`, but the `publish.sh` file was adding `leaflet-src.map` (the 1.0.3 tagged build has `-src.map` as the file name).
Not sure if that was broken or what was going on, looks like it is a recent change.

IvanSanchez:
@perliedman As I said in #5341 I don't really see the usefulness, but if you and @danzel agree, then it's OK.

IvanSanchez:
Ah, oh, ok :smile: 

According to the [sourcemap spec](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#heading=h.9ppdoan5f016):

> Optionally, a source map will have the same name as the generated file but with a “.map” extension.  For example, for “page.js” a source map named “page.js.map” would be generated. 

So ideally I'd go for `leaflet-src.js.map` and `leaflet.js.map`.

IvanSanchez:
```
ivan@~/devel/Leaflet\ [pr/5351]$ ls -l dist                                                                                                              
total 1836                                                                                                                                                              
drwxr-xr-x 1 ivan ivan    146 des.  23 12:44 images
-rw-r--r-- 1 ivan ivan  13888 feb.   8 16:08 leaflet.css
-rw-r--r-- 1 ivan ivan 134439 feb.  21 13:41 leaflet.js
-rw-r--r-- 1 ivan ivan 617195 feb.  21 13:41 leaflet.js.map
-rw-r--r-- 1 ivan ivan 380727 feb.  21 13:41 leaflet-src.js
-rw-r--r-- 1 ivan ivan 725203 feb.  21 13:41 leaflet-src.js.map
```

LGTM.

PR #5352, opened 2017-02-21
Docstrings: recover lost docstring for L.Util.lastId

IvanSanchez:
Related to #5349. It seems that the docstrings for `L.Util.lastId` was lost in the big rollupJS commit, [over here](https://github.com/Leaflet/Leaflet/commit/703ae02aa8cbd0b87be5b01e77754b83ad732267?w=1#diff-4b892233603abc3c8825cfb5905b7453L59).

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -45,6 +45,8 @@ export function bind(fn, obj) {
 	};
 }
 
+// @property lastId: Number
+// Last unique ID used by [`stamp()`](#util-stamp)
 export var lastId = 0;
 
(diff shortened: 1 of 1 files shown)


perliedman:
Nice catch, I noticed I didn't see any docs when reviewed #5349, but just assumed it must be in some surprising location.

IvanSanchez:
@perliedman That's kinda the point of having docstrings, to keep the docs in **un**surprising locations :laughing:

perliedman:
@IvanSanchez lol, good point.

PR #5354, opened 2017-02-22
Fix white lines in Webkit

themre:
I used `translate` property for webkit browsers and also added `will-change` property for Chrome which doesn't cause white lines while panning/zooming.
https://github.com/Leaflet/Leaflet/issues/3575


themre:
this demo also doesn't work for me and I suspect that function isn't called. because I've builded whole leaflet.js and included that leaflet.js into fresh html and it doesn't give me lines like in your demo.
I've mainly read through this discussion where some hints were provided:
https://bugs.chromium.org/p/chromium/issues/detail?id=600120

perliedman:
You're right, the recent ES6 changes makes it impossible to monkey patch Leaflet like in my example.

I modified the playground to use 1.0.3 as base, where you can still monkey patch, and put a debug output to make sure the custom `setTransform` is actually used.

It appears to help initially, but once I zoom out two zoom levels or so, the lines are back.

themre:
ok. yeah, on rare occasions the lines are preset, but paning doesn't give me any white lines like in latest version which is really annoying. we can hope that this webkit "bug" or "feature" will be soon fixed.

perliedman:
I wouldn't hold my breath on Webkit fixing this, I get the feeling from the discussion in the page you linked to that they will not be able to do that.

For now, #5290 is our best bet; it's complicated, but looks like it does the job.

themre:
ok, then we'll wait for this new PixelGrid implementation. for now i'll use my fix in our apps. 
but probably this isn't final solution of PixelGrid, right? because I'm still getting more white lines than in my example. But will this result in having `canvas` element instead of regular `img` tiles?

IvanSanchez:
I've tried @perliedman's playground with a fractional zoom (`var map = new L.Map('map', {center: myCenter, zoom: 15.5, zoomSnap: 0.1});`), and #3575 is still visible :-(

While this is a good approach, it doesn't fix the bug, so I'll close this PR.

> because I'm still getting more white lines than in my example.

Under which conditions? I can only see tile gaps during the tile fade-in animation (which makes them very very subtle)

> But will this result in having `canvas` element instead of regular `img` tiles?

Yeah, that's the whole point of #5290. The only reliable way (that I know of) getting rid of the gaps between tiles is to stitch the tiles together in a `canvas`. :-|

themre:
I've added gif to your PR. I hope you manage to solve it and thanks for all the work!

PR #5359, opened 2017-02-25
fix invalid GeoJSON produced by nested LayerGroups

Resonance1584:
When a LayerGroup contains another LayerGroup .toGeoJSON returns an invalid GeoJSON object as it will return nested FeatureCollection objects.

This fix squashes nested FeatureCollections by concatenating their features array with the parent FeatureCollection's features array.

The GeoJSON spec states that a FeatureCollection's features property contains an array of Feature objects.

http://geojson.org/geojson-spec.html#feature-collection-objects
> The value corresponding to "features" is an array. Each element in the array is a feature object as defined above.

Currently LayerGroup.toGeoJSON will return a FeatureCollection where the features array may contain one or more FeatureCollection objects.

Example:
https://playground-leaflet.rhcloud.com/pin/2/edit?html,console

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -356,4 +356,17 @@ describe("L.LayerGroup#toGeoJSON", function () {
 			features: []
 		});
 	});
+
+	it('should return only one FeatureCollection for nested LayerGroups', function () {
+		var layerGroup = new L.LayerGroup([
+			new L.LayerGroup([new L.Marker([-41.3330287, 173.2008273])]),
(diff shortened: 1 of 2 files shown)


spec/suites/layer/GeoJSONSpec.js
```diff
@@ -356,4 +356,17 @@ describe("L.LayerGroup#toGeoJSON", function () {
 			features: []
 		});
 	});
+
+	it('should return only one FeatureCollection for nested LayerGroups', function () {
```
IvanSanchez: Yay tests!! :+1: :smile:

PR #5361, opened 2017-02-25
Update toolbar inner border radius

### LINKED ISSUE #5360: Toolbar item border radius is too big
## How to reproduce

- Browser (with version) I'm using: Chrome (56.0.2924.87)
- OS/Platform (with version) I'm using: OSX  Sierra
- step 1
Use default map

Looking closely at the inner border radius of the zoom controls.  It seems that the white border does not touch the grey outline, showing a tiny bit of map through the controls at the edges.

screen shot 2017-02-24 at 8 22 55 pm

And is far more noticable if you change to a non-white background:

screen shot 2017-02-24 at 8 19 57 pm

Using http://playground-leaflet.rhcloud.com/ or any other jsfiddle-like site.
N/A see above screenshots.

newmanw:
Flatten toolbar inner border radius to match grey outline.  fixes #5360

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -309,7 +309,14 @@
 	height: 30px;
 	line-height: 30px;
 	}
-
+.leaflet-touch .leaflet-bar a:first-child {
+	border-top-left-radius: 2px;
+	border-top-right-radius: 2px;
+	}
+.leaflet-touch .leaflet-bar a:last-child {
+	border-bottom-left-radius: 2px;
+	border-bottom-right-radius: 2px;
+	}
 
 /* zoom control */
 


perliedman:
Hi and thanks for contributing this fix!

A question: from the screenshots in #5360, it looked like you tested this with a touch enabled device (controls have a somewhat thick border rather than a box shadow). Did you try this fix also without touch support (you can do this by manually setting [`L.Browser.touch`](http://leafletjs.com/reference-1.0.3.html#browser-touch) before instantiating the map)?

newmanw:
@perliedman thanks for looking at this, you are absolutely correct, sorry about that.  Non touch devices look great, I updated the pull request to modify the border radius on touch devices only.  Please take a look and let me know if the new PR makes more sense.

Thanks!

perliedman:
:+1: Looks good, thanks!

PR #5362, opened 2017-02-27
Adding Micello Indoor Maps link / description

marioatmicello:


IvanSanchez:
I'm going to say "no" to this, because it doesn't look like a plugin, but rather like an SDK on top of Leaflet. If we were to accept this kind of projects in the list, I would have added https://api.mazemap.com/js/v1.2.9/docs/ (by yours truly) a long time ago.

marioatmicello:
Hey Ivan

With respect, how is our request any different than the ESRI plugin or 
the existing indoor maps plugin already listed?? Our SDK is basically a 
glorified layer being added to the map.

------ Original Message ------

PR #5372, opened 2017-03-06
Update plugins.md

flowfyre:
add Leaflet.FlowEcharts plugin docs


johnd0e:
Please clarify plugin's relations to Baidu and to https://github.com/wandergis/leaflet-echarts

Issue #5350, opened 2017-02-20
Feature: add callback for animations


dawadam:
Hello,
I'm using version 1.0.3

You can't set the boundary and use an animation at the same time, like this:

 ```
This.map.flyToBounds (myBounds)
This.map.setMaxBounds (myBounds)
```
Exemple : http://playground-leaflet.rhcloud.com/yeya/1/edit

One solution is by adding "moveend" listener who will launch setMaxBounds once the animation finishes, but it's complicated and we must filter events.

Why not add a "end of the animation" callback function in parameters ?

**Comment by perliedman** [comment]
It doesn't look like the linked example provides any useful information, am I missing something?

You could add a one-time `moveend` listener specifically for your `flyTo` animation:

```js
map.flyToBounds(bounds);
map.once('moveend', function() {
    map.setMaxBounds(bounds);
});
```

dawadam:
I think you aremissing somthing, look at the javascript tab.
Thanks for "once", it work.

dawadam:
There is a problem.
If the move of the map is automatic, by "flyTo" method and if we stop via a drag with mouse, event is not called.

perliedman:
Ok, could you provide an illustration of the issue?

I modified your previous example with a `once('moveend')`, and it works as I would expect if I pan the map in the middle of the fly to: the map's max bounds are correctly set and the map is locked around Paris.

dawadam:
Ok, I've updated the example.
Now at the end of the movement, a marker is added to the center of Paris.

http://playground-leaflet.rhcloud.com/xiwi/1/edit?html,js,output

In some cases, when moving is stopped, the marker appears a little later (probably corresponding to the end of automatic movement).

In the other, it does not appear at all.
It appear if new moving on of the map.

perliedman:
I was able to reproduce this problem by just clicking (not dragging) the map while the flyTo animation was running.

My guess is that the click aborts the animation since the user begins a pan operation, but since the map is actually never dragged, no `moveEnd` event is fired. I would consider this a bug, so we will have to investigate further.

dawadam:
Ok,
Good news.

I repeat, sometimes the event is still launched.
Test clicking at the end of the movement to stop the map.

PR #5378, opened 2017-03-08
Stop map on drag start instead of pointer down

perliedman:
Looking at #5350, I found out that `Map.Drag` stops an ongoing `flyTo` animation once a pointer down event is received, and not when the user actually starts dragging the map. This leaves the possibility that the user only taps the map, without starting to drag. In this situation, the animation will be aborted, but no map movement will be registered by `Map.Drag`, so no `moveend` event will be fired, neither for the `flyTo` animation or the drag.

This in turn becomes a problem since it's reasonable to assume a `flyTo` call will be succeeded by a `moveend` at some point, either when the animation finishes or when it's aborted - clearly the map has moved by then.

I looked through other usages of `Map._stop`, and it appears it is otherwise only used where a `moveend` will be fired, so `Map.Drag` is an exception, which is not guaranteed to fire a `moveend`.

An alternative to this fix would be to make `Map._stop` fire a `moveend` if an animation is interrupted - this would also make sense to me, but at the same time seems to be a potentially more intrusive fix: we would sometimes fire multiple `moveend` events where we today only fire one, potentially causing problems for existing apps.

### DIFF
--- a/src/map/handler/Map.Drag.js
+++ b/src/map/handler/Map.Drag.js
@@ -61,7 +61,6 @@ export var Drag = Handler.extend({
 			this._draggable = new Draggable(map._mapPane, map._container);
 
 			this._draggable.on({
-				down: this._onDown,
 				dragstart: this._onDragStart,
 				drag: this._onDrag,
 				dragend: this._onDragEnd
@@ -93,9 +92,6 @@ export var Drag = Handler.extend({
 
 	moving: function () {
 		return this._draggable && this._draggable._moving;
-	},
-
-	_onDown: function () {
 	},
 
 	_onDragStart: function () {
--- a/src/map/handler/Map.Drag.js
+++ b/src/map/handler/Map.Drag.js
@@ -96,12 +96,12 @@ export var Drag = Handler.extend({
 	},
 
 	_onDown: function () {
-		this._map._stop();
 	},
 
 	_onDragStart: function () {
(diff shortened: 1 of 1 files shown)


src/map/handler/Map.Drag.js
```diff
@@ -96,12 +96,12 @@ export var Drag = Handler.extend({
 	},
 
 	_onDown: function () {
```
IvanSanchez: If the body of `_onDown` is empty, we might as well remove the method definition and the `this._draggable.on('down': this._onDown)` line up.

perliedman: Yes, absolutely. Just me being lazy here. Will fix if we think this is otherwise a good solution.

IvanSanchez: Yeah, the solution LGTM. I don't think that the behaviour in the cases it affects (inertia drag, then the user presses the button, waits, then drags) will be noticeable.

perliedman: There is a difference though, previously you could just tap the map during a flyTo, and it would interrupt the animation. With this change, tapping does nothing, you actually have to pull on the map to stop the animation.

IvanSanchez: I'm thinking "fix the bug now, revisit later if users complain about the behaviour" :-)

PR #5379, opened 2017-03-08
Update CONTRIBUTING.md: Use "yarn global add"

BjoernSchilberg:
Use "yarn global add" instead of "yarn install -g". Because "yarn install" is used to install all dependencies for a project [1]. "yarn global"  install packages globally on your operating system [2].

[1] https://yarnpkg.com/en/docs/cli/install
[2] https://yarnpkg.com/en/docs/cli/global


mourner:
Good catch, thanks!

Issue #5353, opened 2017-02-21
Add double quotes for CSS / filter to prevent warnings in Safari debugger


robertharm:
A user of our plugin reported at https://gitlab.com/IvanSanchez/Leaflet.GridLayer.GoogleMutant/issues/24#note_23896582 that the following CSS is producing "Unexpected CSS token" warnings in Safari debugger:

```
.leaflet-oldie .leaflet-popup-tip {
	width: 24px;
	margin: 0 auto;
	-ms-filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678)";
	**filter: progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678);**
	}
.leaflet-oldie .leaflet-popup-tip-container {
	margin-top: -1px;
	}
```

We fixed this by adding double quotes which according to our research is the suggested fix for those kind of warnings: `filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678);"`

Perhaps this change should also be added to the leaflet repo.

IvanSanchez:
Duplicate of #5066, #4977, #4122.

ahmad-ben:
Always add a **" "** double quote to the filter value in the correspondent path.
In my case:
<img width="786" alt="Screenshot 2024-02-27 173448" src="https://github.com/Leaflet/Leaflet/assets/114083017/60b6ffe4-fb0a-41e6-b15b-01807c02d3b9">
The path is **./node_modules/leaflet/dist/leaflet.css:538:15**
So i went to  **./node_modules/leaflet/dist/leaflet.css** and search for **filter property**:
<img width="896" alt="Screenshot 2024-02-27 173635" src="https://github.com/Leaflet/Leaflet/assets/114083017/8a8e75fe-ff94-437c-96b5-2ac51c85bf4f">
Then added the double quote:
<img width="905" alt="Screenshot 2024-02-27 173650" src="https://github.com/Leaflet/Leaflet/assets/114083017/0a679dd9-a408-40b8-98c0-82fd3723824a">

Issue #5066, opened 2016-10-31
progid:DXImageTransform causes error on ios


tuvokki:
- [x] I'm reporting a bug, not asking for help
- [x] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported
There is a pull request that fixes this issue, but it has been closed without merge: https://github.com/Leaflet/Leaflet/pull/4122/commits/0a2a736d57eee2308922d1a4cfdd57984c87add8

----

## How to reproduce

- Leaflet version I'm using: 1.0.1
- Browser (with version) I'm using: Safari on ios (9.3)
- OS/Platform (with version) I'm using: ios (9.3)
- include a map in an ios build. In this case it is an ionic application
- open to see the following:
![screen shot 2016-10-31 at 15 40 09](https://cloud.githubusercontent.com/assets/181719/19858068/693d16ae-9f80-11e6-963f-843a5c01f640.png)

## What behaviour I'm expecting and which behaviour I'm seeing
No warnings

The above pullrequest fixes it.

IvanSanchez:
Duplicate of https://github.com/Leaflet/Leaflet/pull/4977, https://github.com/Leaflet/Leaflet/pull/4122. 

Besides, that's an error, not a warning. Compatibility for old IEs is more important than warnings in Safari.

tuvokki:
Does quoting the line break anything?

IvanSanchez:
Yes. See #4977.

tuvokki:
Clear. I only read (and referenced) #4122 thanks for clearing this up.

PR #5380, opened 2017-03-08
added " " to avoid warning on safari

pierre-leon:


mourner:
Have you confirmed that it doesn't break any versions of old IE? I think it was intentional.

IvanSanchez:
Duplicate of #5353, #5066, #4977, #4122. Rejected for the same reasons.

Issue #3240, opened 2015-02-18
Show GH stars and date of last update in plugin list


mourner:
Here's a nice example we could follow when revamping the plugins page: http://www.staticgen.com/
It uses a static generator script to fetch number of stars for each repo / etc. A bit more cumbersome to maintain, but will make the plugins list more useful. It starts to get out of hand. Dynamic search / sorting would also help a lot.

geomajor56:
Date of last update would be nice too

On Wednesday, February 18, 2015, Vladimir Agafonkin <
notifications@github.com> wrote:

> Here's a nice example we could follow when revamping the plugins page:
> http://www.staticgen.com/
> It uses a static generator script to fetch number of stars for each repo /
> etc. A bit more cumbersome to maintain, but will make the plugins list more
> useful. It starts to get out of hand. Dynamic search / sorting would also
> help a lot.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Leaflet/Leaflet/issues/3240>.
>


-- 
Sent from Gmail Mobile

mourner:
Agreed.

IvanSanchez:
Maybe something entirely client-side with octokat.js (https://github.com/philschatz/octokat.js) would be possible.

Also: add another column indicating whether the plugin is available through Bower / NPM / [insert name of favourite framework or package manager here].

mourner:
@IvanSanchez doing this completely client-side would be a GitHub API abuse — imagine making 200 requests on one page load just to show stars... I think the only way to make it work currently well is precalculating stars. 

Although I just realized it would be so awesome to have a Node service running somewhere that would track number of stars to a set number of repos, cache them and serve. It would be useful not only for Leaflet but for other projects.

Column for package managers would make the list too verbose and noisy I think. One column I think would be useful is one indicating the Leaflet version requirements (e.g. does it support master or only 0.7.3).

jieter:
Is having this list in a html like structure a requirement? I would say the number of plugins justifies a  sortable and searchable list/table with some more information available when selecting a plugin. This is where we can add things like npm/bower package links without cluttering the main list.

Making this table searchable and having good tags/keywords for each plugin will make the categories less important and the reduces the discussion about in which category we need to put a plugin.

I once started to make something like the above, but I cannot find it any more...

IvanSanchez:
> Is having this list in a html-like structure a requirement?

I'd say "yes". In my mind, the plugins should be:

[...]

PR #5387, opened 2017-03-14
Replace the big HTML list of plugins with structured data and jekyll magic

IvanSanchez:
I felt bothered with the "how can we tell if a plugin is compatible with 0.7/1.0?" and also "what is the name for the npm package of this plugin?", so I decided to experiment making the list of plugins into a data structure, and some jekyll/liquid magic to create the static list.

Kinda addresses https://github.com/Leaflet/Leaflet/issues/3240 , as this would make that simpler.

Results looks good so far. I'm not sure about the badges, or about displaying the `npm install leaflet-plugin` bit. I'm wondering how to add some generic `<script src="https://unpkg.com/leaflet-plugin">` also.
screenshot_20170314_175356

---

Things that need to be resolved/discussed before putting more work on this:

- [ ] How to properly convey compatibility with Leaflet 0.7 / 1.0. The badges seem too intrusive.
- [ ] Replace links to the repo with a github/gitlab/bitbucket logo+link??
- [ ] Replace `npm install foo` with a npmjs/unpkg logo+link??
- [ ] Replace author link with an array of links + logos (github/twitter/whatever)
- [ ] File naming. Do we want one file for all the plugins, or one per plugin category?
- [ ] Create a template for the table (like `docs/_includes/frame.html`), then reuse it once per category
- [ ] Add a field for the software license too? This should have the same concept as debian's "contrib": "promotes or depends on a non-free service or library"
- [ ] Are there logos/badges for licenses?
- [ ] Try a two-row + `rowspan=2` layout, so the rightmost column can be two cells with author on top and badges on bottom


IvanSanchez:
Note that this is a WIP as I've only worked on just one section.

perliedman:
Great work (as always), @IvanSanchez! I was toying with the same idea a couple of months ago, but never got anything done.

I think the advantages are obvious, and adds flexibility to do lots of interesting things with the plugin page.

A possible downside, though, is that this will make it harder for people to add new plugins: people know how to edit a HTML page, but not necessarily our custom YML format. (Come to think of it, you might see that as an advantage as well :stuck_out_tongue_winking_eye: )

perliedman:
Related: it would be great if the plugin page could also automatically show stars and last activity for each plugin, when hosted on Github. But that is really outside the scope of what you're doing here.

IvanSanchez:
> people know how to edit a HTML page, but not necessarily our custom YML format.

C'mon, it's just copy-pasting a few lines :-P

---

There are a few things I'm not convinced about and would like some input about, I'll leave them as a to-do list.

perliedman:
I rebased this and changed the structure a bit; now the whole plugin page's structure with categories, subcategories and the plugins themselves are all in the `plugins.yaml` file.

I spent quite some time on a script which extracts all the data about the current plugins.

Some things to note:

* The initial "table of contents" still isn't auto generated (not sure how to deal with the columns automatically)
* Some data is missing from my auto-script (no automatic way to know demo link and version compatibility status)
* The structure of the YAML file became pretty messy when it was auto generated; in particular, the order of the category and plugin properties isn't friendly to a human; we should probably clean this up if we expect humans to edit the file

perliedman:
Oh, yes one more thing: the internal links on the page are currently broken.

Falke-Design:
@IvanSanchez do you want to keep this PR open?

PR #5394, opened 2017-03-18
Submission of new plugin list entry "Leaflet.PolylineMeasure"

ppete2:
Submission of new plugin list entry "Leaflet.PolylineMeasure"


IvanSanchez:
Looks good :+1:

PR #5398, opened 2017-03-21
Add Leaflet.TileLayer.ColorPicker to plugins list

frogcat:
Leaflet.TileLayer.ColoPicker adds getColor(latLng) to TileLayer. It will be useful when you just want to pick a color (or elevation encoded as color) from displaying tile.

color picker demo: <https://frogcat.github.io/leaflet-tilelayer-colorpicker/>
elevation picker demo: <https://frogcat.github.io/leaflet-tilelayer-colorpicker/mapbox-terrain-rgb.html>


IvanSanchez:
Very nice! 👍

PR #5399, opened 2017-03-21
build sets e in catch(e) to argument name

batje:
The current uglify job generates the following code:

`_setOpacityIE:function(t,e){var i=!1,n="DXImageTransform.Microsoft.Alpha";try{i=t.filters.item(n)}catch(t){if(1===e)return}e=Math.round(100*e),i?(i.Enabled=100!==e,i.Opacity=e):t.style.filter+=" progid:"+n+"(opacity="+e+")"}`

The catch(t) statement might *not* return (if value != 1), which changes the argument t to the caught error. I tried alternatives to catch(e) eg. catch(me), catch(you) catch(acold), but only testing for the error to have happened helps.

I get this error when I click on a cluster from the markercluster module.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -174,7 +174,7 @@ function _setOpacityIE(el, value) {
 	if (filter) {
 		filter.Enabled = (value !== 100);
 		filter.Opacity = value;
-	} else {
+	} else if (!el.name == 'Error') {
 		el.style.filter += ' progid:' + filterName + '(opacity=' + value + ')';
 	}
 }
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -174,7 +174,7 @@ function _setOpacityIE(el, value) {
 	if (filter) {
 		filter.Enabled = (value !== 100);
 		filter.Opacity = value;
-	} else if (!el.name == 'Error') {
+	} else if (!el.name === 'Error') {
 		el.style.filter += ' progid:' + filterName + '(opacity=' + value + ')';
 	}
 }


IvanSanchez:
Hhhmmm, I see. `uglifyjs` is overwriting the `el` parameter with the caught `e`xception.

I'm partial to submitting a bug report upstream to uglify (or at the very least try updating the dependencies) before implementing a workaround.

IvanSanchez:
Updating to uglify 2.8 doesn't help. I've tried to create a minimal case which makes uglify reuse variable names, with no avail. :-(

batje:
Fair point. I updated to:
```
├─┬ eslint@3.18.0 
│ ├─┬ babel-code-frame@6.22.0 
│ │ └── js-tokens@3.0.1 
│ ├── doctrine@2.0.0 
│ ├── esquery@1.0.0 
│ ├── shelljs@0.7.7 
│ └── strip-json-comments@2.0.1 
├─┬ rollup-watch@3.2.2 
│ └── require-relative@0.8.7 
└── uglify-js@2.8.14 

```

The result is very similar:
`function yt(t, i) {
  var e = !1,
    n = "DXImageTransform.Microsoft.Alpha";
  try {
    e = t.filters.item(n)
  } catch (t) {
    if (1 === i)
      return
  }
  i = Math.round(100 * i), e ? (e.Enabled = 100 !== i, e.Opacity = i) : "Error" === !t.name && (t.style.filter +=
    " progid:" + n + "(opacity=" + i + ")")
}`

johnd0e:
@batje 
Is it reproducible with current Leaflet?
I cannot find `_setOpacityIE` in uglified code anymore.

batje:
I would say that if you can't find _setOpacityIE anymore, this issue has become irrelevant.

PR #5411, opened 2017-03-26
Doc(LayerGroup): add hasLayer method call with id

ghybs:
With the current `hasLayer` code, we can call this method by passing the `_leaflet_id` and get the same result as passing the corresponding layer object.
Added the alternative method call, as done for `removeLayer` method.


IvanSanchez:
LGTM

PR #5415, opened 2017-03-27
Added event for when loading image fails + documentation

Saulzi:
Added event for when loading image fails on ImageOverlay layer and added missing documentation for load event

PR #5416, opened 2017-03-27
Added event for when loading image fails + documentation

Saulzi:
Added event for when loading image fails on ImageOverlay layer and added missing documentation for load event

### DIFF
--- a/spec/suites/layer/ImageOverlaySpec.js
+++ b/spec/suites/layer/ImageOverlaySpec.js
@@ -14,4 +14,67 @@ describe('ImageOverlay', function () {
 			expect(overlay._bounds).to.equal(bounds);
 		});
 	});
+	describe("_image", function () {
+		var c, map, overlay;
+		// Url for testing errors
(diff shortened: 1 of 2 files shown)


Saulzi:
This is the same as PR #5415 without the extra documentation file.

spec/suites/layer/ImageOverlaySpec.js
```diff
@@ -14,4 +14,67 @@ describe('ImageOverlay', function () {
...
+		function raiseImageEvent(event) {
+			var domEvent = document.createEvent('Event');
+			domEvent.initEvent(event);
+			overlay._image.dispatchEvent(domEvent);
+		}
+
+		describe('when loaded', function () {
+			it('should raise the load event', function () {
+				var loadraised = false;
+				overlay.once('load', function () { loadraised = true; });
```
perliedman: While this works, you can also use `sinon` to do this and similar tasks in a more explicit fashion, see for example https://github.com/Leaflet/Leaflet/blob/master/spec/suites/core/EventsSpec.js#L20

spec/suites/layer/ImageOverlaySpec.js
```diff
@@ -14,4 +14,67 @@ describe('ImageOverlay', function () {
...
+				raiseImageEvent('load');
+				expect(loadraised).to.be(true);
+			});
+		});
+
+		describe('when load fails', function () {
+			it('should raise the error event', function () {
+				var errorRaised = false;
+				overlay.once('error', function () { errorRaised = true; });
```
perliedman: ...same here, `sinon` would be great.

Saulzi: So basically the same as jasmine as you can do jasmine.spy. 

Will bear that in mind for any future contributions :)

PR #5422, opened 2017-03-28
Control.Layers: Only add layer events to layers when we are on the map

### LINKED ISSUE #5421: L.control.layers don't fire "baselayerchange, overlayadd, overlayremove" events after setPosition()
## Setup
- Leaflet version I'm using: Leaflet 1.0.3
- Browser (with version) I'm using: latest Chrome, Firefox
- OS/Platform (with version) I'm using: macOS Sierra
- Open: https://playground-leaflet.rhcloud.com/yame/1/edit?html,output

## How to reproduce
1. Change layers -> events are fired and logged 🆗
2. Click "change pos" button `layerControl.setPosition('bottomright') `-> layercontrol is moved to bottom-right. 🆗
3. Change layers -> events don't fire anymore and aren't logged ⛔

## URL
https://playground-leaflet.rhcloud.com/yame/1/edit?html,output

danzel:
Can be seen as an extension of the fixes done in #4214 / #4213
Fixes #5421

### DIFF
--- a/spec/suites/control/Control.LayersSpec.js
+++ b/spec/suites/control/Control.LayersSpec.js
@@ -29,6 +29,26 @@ describe("Control.Layers", function () {
 			expect(spy.args[1][0].layer).to.be(baseLayers["Layer 2"]);
 		});
 
+		it("works after removing and readding the Control.Layers to the map", function () {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Lookin' good! 👍

Issue #5381, opened 2017-03-09
Missing tiles after fast zoom in and out with wouse-wheel


oliverheilig:
I'm experiencing missing tiles when i zoom in and out on the map. I first thought it's an issue with my tile-provider. But after extensive tests i can reproduce it with any Leaflet map and any provider. It's a bit tricky to provoke this though.

## How to reproduce

- I've tested it with LL 1.0.3, and the current trunk
- I can reproduce it with Chrome, FireFox and Edge on Windows 10
- I can reproduce this with a zoom-gesture on my notebook and a real mouse on my PC
- I even can reproduce it with the Leaflet quickstart sample http://leafletjs.com/examples/quick-start/example.html:
- First zoom out to world view.
- Then zoom in very fast with the mouse-wheel
- Immediately afterwards zoom out two wheel-clicks.

You need the right timing to get this result, i manage this about 20% of the time. All i can see is that the missing tile isn't even requested (using fiddler).

![missing-tiles](https://cloud.githubusercontent.com/assets/958768/23729423/0d7eb95c-0462-11e7-8564-3fb7a7b52768.png)

IvanSanchez:
Since I cannot reproduce this easily, I'm gonna ask you to find some reproducible timings with synthetic `WheelEvent`s.

Something like:

```
    function doTheWheel(amount) {
      var mapSize = map.getSize();
      var ev = new WheelEvent('wheel', {

        deltaX: 0,
        deltaY: amount,
        deltaZ: 0,
        deltaMode: 0,

        clientX: mapSize.x / 2,
        clientY: mapSize.y / 2
      });

      map.getContainer().dispatchEvent(ev);
    }

    function doTheWheelClosure(amount) {
      return function() {
        return doTheWheel(amount);
      }
    }

    setTimeout(doTheWheelClosure(-50), 600);  // Scroll "up"/"in" three times in very quick succession
    setTimeout(doTheWheelClosure(-50), 610);
    setTimeout(doTheWheelClosure(-50), 620);
    
    setTimeout(doTheWheelClosure(50), 1500); // Scroll "down"/"out" twice some time later
    setTimeout(doTheWheelClosure(50), 1700);
```

See https://playground-leaflet.rhcloud.com/gaxu/edit?html,output for a half-baked example. Changing the times and scrolling distances should be everything you'd need to do.

oliverheilig:
et voilà,

https://playground-leaflet.rhcloud.com/tefi/edit?html,output

[...]

PR #5431, opened 2017-04-01
On update set current tiles active to avoid pruning (#5381)

oliverheilig:
The missing tiles after fast zoom in/out seems to be a race condition. 

When the tiles of the level have been loaded before they are re-used. But it can occur that this tile is not marked as active (maybe the opacity animation wasn't completed before), and this removes the current tile at the purging function.

Forcing the current tile to active solves this problem.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -653,6 +653,7 @@ export var GridLayer = Layer.extend({
 				var tile = this._tiles[this._tileCoordsToKey(coords)];
 				if (tile) {
 					tile.current = true;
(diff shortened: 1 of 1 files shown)


oliverheilig:
Uh ok, 

i see what you mean.

oliverheilig:
OK, 

i also have another fix for the missing tiles which doesn't cause the regression. Not changing any property of the currently loaded tiles fixes  #5381 without the flickering. 

http://176.95.37.29/vectortest/missing-tiles-fixed2.html

Not really understanding the pruning code in detail, this is just an educated guess.

PR #5432, opened 2017-04-04
Escape html elements inside code sample

daturkel:
The markup renderer will still literally render html tags inside "pre" and "code" unless they're escaped. The result was that the example code would be rendered as:
```
var baseMaps = {
	"Grayscale": grayscale,
	"Streets": streets
};
```
rather than
```
var baseMaps = {
	"<span style='color: gray'>Grayscale</span>": grayscale,
	"Streets": streets
};
```
because the "Grayscale" text would be literally styled as a span with gray color, and then recolored by the syntax highlighter.

Escaping the span tags fixes this.


daturkel:
Only caught it because I was confused while following the tutorial myself!

perliedman:
@daturkel yeah, I can imagine it looked a bit suspicious... Just naming it "Grayscale" makes it gray automatically :)

PR #5436, opened 2017-04-05
[zoom-control] Use `touchstart` event if in Mobile devices

yanglang:
### Description
Instead of always listening to `click` event, we should actually listen to `touchstart` event if in mobile devices as it's more reliable way to see if the button is being clicked. 

We are seeing a bug on mobile devices (iOS with Chrome and Safari) where the first mouse click on zoom buttons is always ignored, and have to click twice to trigger the correct action. After digging into the issue, we found out that the `click` event will not be called in some scenarios, as documented here https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent. And changing to `touchstart` event seems fixed the issue. 

I think in other cases, we may need to have better solution instead of use `touchstart` to replace `click`, but since the target here is just a button with no other complicated interactions, so I think it's okay here.

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -1,6 +1,7 @@
 
 import {Control} from './Control';
 import {Map} from '../map/Map';
+import * as Browser from '../core/Browser';
 import * as DomUtil from '../dom/DomUtil';
 import * as DomEvent from '../dom/DomEvent';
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
No. Touch support does not imply that a browser *will* fire a `TouchEvent` instead of a `MouseEvent`.

This is particularly troubling in the case of Chrome >= 55, and more widespread usage of `PointerEvent`s. See https://github.com/Leaflet/Leaflet/issues/5180 and related discussion, also see the work done in the [pointerify branch](https://github.com/Leaflet/Leaflet/tree/pointerify).

While your concern is valid, your solution does not cover some of the edge cases that we know are problematic (devices with both mouse and touchscreens, etc).

PR #5438, opened 2017-04-06
Relative path for images in CSS

alanevEugene:
It is necessary when styles are imported using webpack

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -345,12 +345,12 @@
 	border-radius: 5px;
 	}
 .leaflet-control-layers-toggle {
-	background-image: url(images/layers.png);
+	background-image: url(./images/layers.png);
 	width: 36px;
 	height: 36px;
 	}
 .leaflet-retina .leaflet-control-layers-toggle {
-	background-image: url(images/layers-2x.png);
+	background-image: url(./images/layers-2x.png);
 	background-size: 26px 26px;
 	}
 .leaflet-touch .leaflet-control-layers-toggle {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
How does the failure look like, right now? We did fix some webpack problems back in 1.0.1 and haven't heard of any problems since then

alanevEugene:
I had such a problem:
```bash
ERROR in ../~/css-loader?{"modules":true,"sourceMap":true,"importLoaders":1}!../~/postcss-loader!../~/leaflet/dist/leaflet.css
Module not found: Error: Can't resolve 'images/layers.png' in '/home/tom/projects/addycar-advertisers-ui/node_modules/leaflet/dist'
 @ ../~/css-loader?{"modules":true,"sourceMap":true,"importLoaders":1}!../~/postcss-loader!../~/leaflet/dist/leaflet.css 6:6656-6684
 @ ../~/leaflet/dist/leaflet.css
```

meendoo:
Have you tried [resolve-url-loader](https://www.npmjs.com/package/resolve-url-loader)? Maybe it helps. I dunno.

perliedman:
Hi! I'm going to close this, as there's not been any action for a long time here.

At the moment, it is not clear to me why this change is needed, it also appears a lot of other webpack users do not see this issue, although they instead have other problems (#4968).

If you still feel this PR adds value to Leaflet, don't feel discouraged, just tell me why we need this and we'll consider re-opening!

alanevEugene:
This problem is not actual for me no more, and I even don't remember how to reproduce this error.
Thank you for closing PR, I should have done this before

PR #5452, opened 2017-04-11
Fix box zoom race condition

ppaskaris:
The deferred call to `_resetState` can interrupt the next box zoom if
the user initiates it before the timeout fires. This causes the mouse
move handler to create a second box zoom element, orphaning the first
one and leaving it in the DOM.

It is possible to reproduce this issue in Chrome by using the Timeline
panel's CPU throttle (20x) and box zooming in rapid succession.

### DIFF
--- a/src/map/handler/Map.BoxZoom.js
+++ b/src/map/handler/Map.BoxZoom.js
@@ -25,6 +25,7 @@ export var BoxZoom = Handler.extend({
 		this._map = map;
 		this._container = map._container;
 		this._pane = map._panes.overlayPane;
+		this._resetStateTimeout = 0;
 		map.on('unload', this._destroy, this);
 	},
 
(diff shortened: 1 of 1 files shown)


src/map/handler/Map.BoxZoom.js
```diff
@@ -113,7 +122,10 @@ export var BoxZoom = Handler.extend({
 		if (!this._moved) { return; }
 		// Postpone to next JS tick so internal click event handling
 		// still see it as "moved".
-		setTimeout(Util.bind(this._resetState, this), 0);
+		if (this._resetStateTimeout !== 0) {
```
perliedman: This is more or less exactly the same code as line 57-62; could we break this out to a private method, something like `_clearDeferredResetState`?

perliedman:
Thanks :+1:

PR #5461, opened 2017-04-17
Web accessibility

Mirodil:
so tabbing through the page make the link(in the popup) to the first entry focussed before the 'x' for close


IvanSanchez:
I don't think there is a [WAI-ARIA guideline](https://www.w3.org/TR/wai-aria-practices-1.1/) for this kind of thing, but it looks sane to me. 👍 

Please check the continuous integration tests - there is some trailing whitespace that is making them fail. 👎

Issue #5118, opened 2016-11-18
Documentation for L.Map.FitBounds is incorrect? 


HarelM:
- [x] I'm reporting a bug, not asking for help
- [x] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behavior is documented and expected
- [x] I've searched through the issues to make sure it's not yet reported

FitBounds documentation describes that options parameter is not optional, I'm not sure this is the case as the codes tell me otherwise, am I wrong?

P.S. you have a spelling mistake in the report issue template for the word "behaviour"... :-)

IvanSanchez:
Yeah, there should be a question mark to describe that parameter as optional. Do you know how to do a PR for that?

P.S.: [No, we don't](http://www.merriam-webster.com/dictionary/behaviour).

HarelM:
I do, but I'm not sure how the documentation is generated (I assume it is).
Just add the question mark in the code's comment itself?
I'm starting a migration to version 1.0 and I'm using typescript so every issue with the documentation is creating a compilation error on my side (at least for the functions I use).

Same question for ```map.setZoom()```?

IvanSanchez:
Read CONTRIBUTING.md, how to build the documentation is explained there :-)

If you can fix several docstrings at once (in the same PR), please do so. From our perspective it creates cleaner merges.

HarelM:
OK, I'll do my best, I'm not promising anything. If it'll take me too much time or I find it hard to get started I'll let you know so you can fix it instead of me.
I'll finish the migration (make the entire code base compile) and then create a PR so that I'll know that I took care of everything I know of.

I suggest to try and compare the previous documentation to the new one and see if the things that changed were suppose to change - it has better chances of finding all kinds of small issues like this one.

Thanks for a great library BTW.

IvanSanchez:
Sounds good. Just drop a comment when ready.

HarelM:
Well, unfortunately, I couldn't make the migrated code work, I decided to downgrade back to version 0.7.7.
Please update the documentation with the two fixes above for fitBounds and setZoom.
It seems that angularJS 1.x, typescript and leaflet are not meant to work together. :-(
I couldn't find a document for migration so if you can point me out to it, it would be great, although I have very little hope after struggling with it most of the day...

rodrijuarez:
@IvanSanchez I've already created the PR referencing this issue, fixing the documentation for those two functions and another problem that I had with the opening/closing tags in a template

PR #5462, opened 2017-04-17
Issue #5118 Fix documentation for method setZoom

gatsbimantico:
As mentioned in #5118 the method setZoom needed to fix the docs.
Checked the code and it should be optional.


danzel:
Thanks

PR #5463, opened 2017-04-18
Remove global L usage

Trufi:
I found global L usage in code.
Was it kept for some purpose or just forgot to change?


IvanSanchez:
It seems that we forgot to change those. Very good catches!!

PR #5468, opened 2017-04-19
Add subresource integrity information & scripts

### LINKED ISSUE #5466: Generate integrity hashes during build
Problem: Leaflet can be subject of a script injection attack on the CDNs (see https://github.com/Leaflet/Leaflet/issues/5457 ). And, as https://github.com/Leaflet/Leaflet/issues/5460 points out, some users can be very security-conscious.

Solution: Calculate [subresource integrity hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) during the build process, and put those in the download page of the website.

e.g. instead of 

```html
<script src="https://unpkg.com/leaflet@1.0.3/dist/leaflet.js"></script>
```

tell users to use something like 

```html
<script src="https://unpkg.com/leaflet@1.0.3/dist/leaflet.js" integrity="sha384-abcdefghijklmnopq"></script>
```

The [MDN page on integrity hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) has some recipes to calculate the hashes on a CLI.

* [ ] Calculate integrity hashes on a build and output them
* [ ] Make the release script modify `docs/_config.yml` to overwrite the value of the latest release, and the latest integrity hash
* [ ] Modify `docs/download.md` to put the integrity hash somewhere

Ideally hashes should be provided for the uglified and non-uglified flavours of `leaflet.js` and `leaflet-src.js`; hopefully this won't add much more noise to the download webpage.

IvanSanchez:
Should fix #5466.

### DIFF
--- a/build/publish.sh
+++ b/build/publish.sh
@@ -33,4 +33,4 @@ git checkout master
 git branch -D build
 
 echo "All done."
-echo "Remember to run 'npm run-script integrity' to upload the information in the website's download page"
+echo "Remember to run 'npm run-script integrity' and then commit the changes to the master branch, in order to update the website."
--- a/build/publish.sh
+++ b/build/publish.sh
@@ -33,3 +33,4 @@ git checkout master
 git branch -D build
 
 echo "All done."
+echo "Remember to run 'npm run-script integrity' to upload the information in the website's download page"
--- a/build/integrity.js
+++ b/build/integrity.js
@@ -1,0 +1,29 @@
+
+// This script calculates the integrity hashes of the files in dist/ , and
+// **overwrites** the values in the documentation.
+
+var ssri = require('ssri');
+var fs   = require('fs');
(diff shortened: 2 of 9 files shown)


build/publish.sh
```diff
@@ -33,3 +33,4 @@ git checkout master
 git branch -D build
 
 echo "All done."
+echo "Remember to run 'npm run-script integrity' to upload the information in the website's download page"
```
perliedman: The `integrity` task doesn't actually upload the information, or am I missing something? So the text should probably read something like

```
Remember to run 'npm run-script integrity' _and_ upload the information in the website's download page
```

(Added emphasis to show the difference.)

IvanSanchez: Done. The new wording should be a bit clearer.

PR #5476, opened 2017-04-25
Add hook points to allow for a proper NoGap plugin

IvanSanchez:
This is just a subset of changes from #5290.

Motivation is @mourner's comments in https://github.com/Leaflet/Leaflet/pull/5290#issuecomment-286386260 , about the tile gap not being a core issue. This PR just allows plugin authors to run some code whenever a `GridLayer`  creates or deletes a tile zoom-level container, and whenever a tile becomes fully opaque. That, in turn, will allow much cleaner code for [L.TileLayer.NoGap](https://github.com/Leaflet/Leaflet.TileLayer.NoGap/blob/master/L.TileLayer.NoGap.js), since right now it reimplements the whole tile pruning logic.


IvanSanchez:
Once/if this is merged, then the NoGap plugin can finally take form. See https://github.com/Leaflet/Leaflet.TileLayer.NoGap/pull/3

PR #5478, opened 2017-04-25
Sanity check to prevent loading tiles when bounds are `Infinity`

### LINKED ISSUE #5477: Setting map center to L.LatLng(Infinity, Infinity) causes infinite loop
## How to reproduce
- Leaflet version I'm using: `master`, 1.0.3
- Browser (with version) I'm using: Chrome, Firefox (latest versions)
- OS/Platform (with version) I'm using: Windows 8.1

I expect that passing `Infinity` in to the `L.LatLng` constructor like this `new L.LatLng(Infinity, Infinity);` would throw an error. Instead this is allowed and is causing the browser to freeze when adding a tile layer, for example.

**Beware, this will freeze your browser**
Example - http://playground-leaflet.rhcloud.com/qem/edit?html,output

IvanSanchez:
Fixes #5477

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -631,6 +631,12 @@ export var GridLayer = Layer.extend({
 		    noPruneRange = new Bounds(tileRange.getBottomLeft().subtract([margin, -margin]),
 		                              tileRange.getTopRight().add([margin, -margin]));
 
+		// Sanity check: panic if the tile range contains Infinity somewhere.
+		if (tileRange.min.x === Infinity ||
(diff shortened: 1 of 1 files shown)


IvanSanchez:
@perliedman Wait, this didn't take into account the case of `-Infinity`

IvanSanchez:
Huh. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite

perliedman:
@IvanSanchez doh, so much for my reviewing skills! :disappointed:

PR #5479, opened 2017-04-25
Better sanity checks for avoiding loading infinite tiles

IvanSanchez:


IvanSanchez:
See https://github.com/Leaflet/Leaflet/pull/5478#issuecomment-297015214

Issue #5475, opened 2017-04-24
L.Bounds methods and parameters


dom1n1k:
Docs: http://leafletjs.com/reference-1.0.3.html#bounds

Factory:

`L.bounds(<Point> topLeft, <Point> bottomRight)` [link](http://leafletjs.com/reference-1.0.3.html#bounds-l-bounds)

But in methods:

`getBottomLeft()` - Returns the bottom-left point of the bounds [link](http://leafletjs.com/reference-1.0.3.html#bounds-getbottomleft)
`getTopRight()` - Returns the top-right point of the bounds [link](http://leafletjs.com/reference-1.0.3.html#bounds-gettopright)

Why not `getTopLeft` and `getBottomRight`? I think this is inconsistent.

ghybs:
Hi,

I think this is similar to https://github.com/Leaflet/Leaflet/pull/5059

I.e. actually you can pass whatever points / "corners" to `L.bounds` factory (not just `topLeft`, `bottomRight`, could be `topRight` and `bottomLeft`), and it will just re-order the coordinates to get:

- the smaller `x` in `min.x` (i.e. left-most),
- the higher `x` in `max.x` (i.e. right-most),
- the smaller `y` in `min.y` (i.e. top-most) and
- the higher `y` in `max.y` (i.e. bottom-most).

perliedman:
@ghybs is right here, in practice it makes no difference: as long as you pass min and max for x as well as y, the order is insignificant.

Having said that, I have to agree that the naming is weird, and goes against what I think is established conventions. We should fix this.

Just renaming the factory/constructor parameters `bottomLeft` and `topRight` seems right. Contribution to address this very welcome!

ghybs:
Maybe `corner1` and `corner2` like in #5059 would be closer to the functionality, even though farther from conventions?

BTW, we could even push further the refinement by directly accepting an indefinite length of point arguments - i.e. corners (instead of a single array or 2 points), by replacing `points` by `arguments` if we detect more than 1 argument (or if the 1st one is not an array). Same applies for `L.LatLngBounds`.

perliedman:
@ghybs :+1: consistency with #5059 seems like a good idea, updated suggestion above.

dom1n1k:
Maybe add 2 extra methods `getTopLeft` and `getBottomRight`? Additional, not instead.
In `LatLngBounds` there are methods for all 4 corners:
[...]

PR #5488, opened 2017-04-29
Feat(Bounds): add new methods for 2 missing corners

ghybs:
As suggested in https://github.com/Leaflet/Leaflet/issues/5475, this PR adds `getTopLeft()` and `getBottomRight()` methods to `Bounds` to retrieve the 2 missing corners (besides `getBottomLeft()` and `getTopRight()` already existing).

Having now the 4 corners make `Bounds` more in line with `LatLngBounds`.

Also added 4 tests (for each of the corner retrieval methods).

Closes https://github.com/Leaflet/Leaflet/issues/5475 together with the previous PR https://github.com/Leaflet/Leaflet/pull/5487

### DIFF
--- a/src/geometry/Bounds.js
+++ b/src/geometry/Bounds.js
@@ -74,13 +74,13 @@ Bounds.prototype = {
 	},
 
 	// @method getTopLeft(): Point
-	// Returns the top-left point of the bounds.
+	// Returns the top-left point of the bounds (i.e. [`this.min`](#bounds-min)).
 	getTopLeft: function () {
 		return this.min; // left, top
 	},
 
 	// @method getBottomRight(): Point
-	// Returns the bottom-right point of the bounds.
+	// Returns the bottom-right point of the bounds (i.e. [`this.max`](#bounds-max)).
 	getBottomRight: function () {
 		return this.max; // right, bottom
 	},
--- a/src/geometry/Bounds.js
+++ b/src/geometry/Bounds.js
@@ -73,6 +73,18 @@ Bounds.prototype = {
 		return new Point(this.max.x, this.min.y);
 	},
 
+	// @method getTopLeft(): Point
+	// Returns the top-left point of the bounds.
+	getTopLeft: function () {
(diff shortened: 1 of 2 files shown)


src/geometry/Bounds.js
```diff
@@ -73,6 +73,18 @@ Bounds.prototype = {
...
+	// @method getTopLeft(): Point
+	// Returns the top-left point of the bounds.
+	getTopLeft: function () {
+		return new Point(this.min.x, this.min.y); // left, top
+	},
+
+	// @method getBottomRight(): Point
+	// Returns the bottom-right point of the bounds.
+	getBottomRight: function () {
+		return new Point(this.max.x, this.max.y); // right, bottom
```
mourner: You can write it as `return this.max.clone();`.

ghybs: Hi,

Thanks for the review and catch!

Now that I realize (:ashamed:) that we just return the same values  as `this.max`, I wonder if we should even clone it at all?

`LatLngBounds` do not bother cloning: https://github.com/Leaflet/Leaflet/blob/master/src/geo/LatLngBounds.js#L100-L110

Cloning is the exact transcription of my initial code.
Doing so is for sure the safest way, but slightly expensive performance-wise (e.g. see https://github.com/Leaflet/Leaflet.markercluster/pull/536#issuecomment-149535377).

Not cloning is the most simple, but could lead to app mistakes if the developer modifies them without cloning first.
But it is entirely consistent with `LatLngBounds`.

Furthermore, as these 2 methods are new, it should not break anything (yet) exposing them with potentially error-prone behaviour, should the app developer be not cautious enough.

For simplicity and consistency with `LatLngBounds`, I assume you would be ok to not clone and directly output `this.max` (same for `this.min`), so I will modify this PR accordingly.

Please feel free to comment back should you have a different position.

ghybs:
Took the opportunity to add a documentation comment that the 2 new methods are just shorthands for the already available properties `this.min` and `this.max`, with links.

PR #5493, opened 2017-05-03
Replace reference.html symlink with redirect 

alyhegazy:


alyhegazy:
Yes, worked fine for me using bundle exec jekyll serve.

perliedman:
@alyhegazy looks like it's working on the live site as well. Thanks for submitting this!

alyhegazy:
@perliedman anytime!

PR #5495, opened 2017-05-03
leaflet-geopackage plugin added

danielbarela:
Added the leaflet-geopackage plugin to the Overlay data formats section


perliedman:
👍

Issue #5499, opened 2017-05-04
pressing Enter key results in click event


egoroof:
## How to reproduce

- Leaflet version I'm using: 1.0.3
- Browser (with version) I'm using: Chrome 60 (also firefox, safari and others)
- OS/Platform (with version) I'm using: macOS Sierra 10.12.4 (also Windows 10 and Ubuntu)
- go to http://leafletjs.com/examples/quick-start/example.html
- open browser console
- click to map
- press Enter key

## What behaviour I'm expecting and which behaviour I'm seeing

Press Enter key results in click event (you may catch it by adding click event listener on map). This behaviour causes issues in different modules. In the example clicking on geometry (circle or polygon) and press Enter key logs `Cannot read property 'toString' of undefined`, clicking in other point and press Enter key logs `Cannot read property 'lat' of undefined`.

I expect suppressing Enter key as click event so I wont need to check it in every module manually.

![screen shot 2017-05-04 at 18 11 54 2](https://cloud.githubusercontent.com/assets/2605831/25701977/022733a4-30f9-11e7-834c-963e25529de6.png)

![screen shot 2017-05-04 at 18 09 07 2](https://cloud.githubusercontent.com/assets/2605831/25701976/02257f96-30f9-11e7-852d-40fa8d4b43c7.png)

perliedman:
The conversion from `keypress` to `click` stems from 40179ed11f8f74958ea6da5789486913df2fb397, which addresses #1355. Originally this only applied to markers, but apparently we do this for any enter keypress now.

Firing `click` events without `latlng` seems like a bad idea, but on the other hand keyboard accessibility is a good thing. We need to come up with a better solution.

IvanSanchez:
This is made by the browser, by design, for accessibility purposes. If you press `Tab` to cycle through clickable (and focusable) elements in the page, then press `Enter`, the browser itself will fire a `click` event.

I cannot reproduce any problems in my main machine (Linux) when the map is focused. Can any other user with MacOSX confirm?

IvanSanchez:
Note that these clicks are fired by the browser and not by Leaflet, by trying out something like:

```
<!DOCTYPE html>
<html>
<head>
</head>
<body>
  <a id='mylink' href='#'>Foobar</a>

  <script>
    
document.getElementById('mylink').addEventListener( 'click', (ev)=>{
  console.log(ev);
  console.log('Link has been "clicked"');
} );
    
  </script>
</body>
</html>
```

[...]

PR #5507, opened 2017-05-11
Don't turn enter keypress into clicks on map

perliedman:
Addresses #5499 by:

* removing the general transformation of enter `keypress` into `click`
* adding a `keypress` listener for markers with bound popups

Possible downside: popups or other `click` handlers that added without `bindPopup` will no longer fire when enter is pressed on a focused marker.

I first tried to fix this in a more general way, but it ends up more of a hack, where we can still end up with a `click` event without mouse coordinates and `latlng`, exactly like #5499 again.

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -383,7 +383,7 @@ Layer.include({
 		if (!this._popupHandlersAdded) {
 			this.on({
 				click: this._openPopup,
-				keypress: this._popupKeypressHandler,
+				keypress: this._onKeyPress,
 				remove: this.closePopup,
 				move: this._movePopup
 			});
@@ -399,7 +399,7 @@ Layer.include({
 		if (this._popup) {
 			this.off({
 				click: this._openPopup,
-				keypress: this._popupKeypressHandler,
+				keypress: this._onKeyPress,
 				remove: this.closePopup,
 				move: this._movePopup
 			});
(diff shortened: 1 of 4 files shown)


src/map/Map.js
```diff
@@ -1300,9 +1300,9 @@ export var Map = Evented.extend({
 	_handleDOMEvent: function (e) {
 		if (!this._loaded || DomEvent.skipped(e)) { return; }
 
-		var type = e.type === 'keypress' && e.keyCode === 13 ? 'click' : e.type;
+		var type = e.type;
 
-		if (type === 'mousedown') {
+		if (type === 'mousedown' || e.type === 'keypress') {
```
egoroof: `e.type` -> `type` ?

src/layer/Popup.js
```diff
@@ -515,5 +517,11 @@ Layer.include({
 
 	_movePopup: function (e) {
 		this._popup.setLatLng(e.latlng);
+	},
+
+	_popupKeypressHandler: function (e) {
```
IvanSanchez: I'd personally name this `_onKeyPress` instead, but it's just a matter of taste

perliedman: Yep, agreed that was a weird name to come up with!

egoroof:
Thanks!

PR #5510, opened 2017-05-12
enable/disable inertia in runtime

treecon:
I haven't found a way to enable/disable inertia (and other options) in runtime.

I suppose something like 

map.inertia.enable()
map.inertia.disable()

Is this easy to be added?


perliedman:
Hi and thanks for bringing this question up. It seems reasonable that inertia should be possible to enable or disable at runtime.

However, you've opened a _pull request_, which is something you do if you want to contribute code (ask us to _pull_ your code into our code). You should open an issue instead. You can of course also open a pull request that actually fixes this.

Closing this for now.

PR #5515, opened 2017-05-17
Add imageOverlayer image rendering option

diversoft:
add image randdering option,  imageoverlayer can zoom using interpolation algorithm: nearest-neighbor

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -206,6 +206,7 @@
 .leaflet-zoomPixelated
 {
 	-ms-interpolation-mode: nearest-neighbor; /*IE 7+ (non-standard property)*/
+	-ms-interpolation-mode: nearest-neighbor;
 	image-rendering: -webkit-optimize-contrast; /*// Safari 6, UC Browser 9.9*/
 	image-rendering: -webkit-crisp-edges; /*// Safari 7+*/
 	image-rendering: -moz-crisp-edges; /*// Firefox 3.6+*/
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -196,15 +196,9 @@
 	visibility: hidden;
 	}
 
+/* zoom imageoverlay pixelated */
 
-/* cursors */
-
-.leaflet-interactive {
-	cursor: pointer;
-	}
-
-.leaflet-zoomPixelated
-{
+.leaflet-zoompixelated{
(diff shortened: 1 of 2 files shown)


src/layer/ImageOverlay.js
```diff
@@ -47,7 +47,10 @@ export var ImageOverlay = Layer.extend({
 
 		// @option zIndex: Number = 1
 		// The explicit [zIndex](https://developer.mozilla.org/docs/Web/CSS/CSS_Positioning/Understanding_z_index) of the tile layer.
-		zIndex: 1
+		zIndex: 1,
+		// @option zoomPixelated: String = ''
+		// The explicit [zoomPixelated](https://developer.mozilla.org/zh-CN/docs/Web/CSS/image-rendering) of the tile layer.
+		zoomPixelated:"NN"
```
perliedman: Three things here:

1. As implemented now, this is a boolean switch, it's either on or off; the naming also sounds like it's a boolean. I'd either keep the name and make the value `true`/`false`, or change the name to something like `imageRendering`, since that's the name used in CSS
2. The default should not be nearest neighbor, we should keep default as it is now (`image-rendering: auto`)
3. `NN` is not a very descriptive value; if we're not making this a boolean (see point 1 above), I'd prefer `pixelated` or `nearest-neighbor` to keep to the CSS terminology

dist/leaflet.css
```diff
@@ -202,6 +202,16 @@
 .leaflet-interactive {
 	cursor: pointer;
 	}
+
+.leaflet-zoomPixelated
```
perliedman: To keep with naming standard in our CSS, this should be `leaflet-zoom-pixelated`.

diversoft:
I changed my code, how can i re-pullrequest?

perliedman:
No need to do a new request, your new commits automatically end up in this request, as you can see.

Your changes look good, but I'm on a phone right now, so I will take a closer look when I'm by a real computer.

perliedman:
Hi, and thanks again for taking the time to contributing this code!

After some consideration, I've decided to approach this in a slightly more general way: instead of adding a specific option for rendering, we would allow to add any custom CSS class to an `ImageOverlay` using a the new option `className`. This would not only allow you to add rendering hints like making the image pixelated, but really any styling you can apply to an image.

Although this code change didn't get merged, submitting ideas like this is essential for improving Leaflet's quality and flexibility!

Closing in favor of #5555.

PR #5517, opened 2017-05-19
Div icon accept node element as option

MGMsystems:
I have added the option to pass a DOM element to divIcon.
This will append the element in the div created for the icon.
If both html and element are passed. The html option will be set.

### DIFF
--- a/src/layer/marker/DivIcon.js
+++ b/src/layer/marker/DivIcon.js
@@ -29,13 +29,9 @@ export var DivIcon = Icon.extend({
 		// iconAnchor: (Point),
 		// popupAnchor: (Point),
 
-		// @option html: String = ''
+		// @option html: String|DOM Element = ''
 		// Custom HTML code to put inside the div element, empty by default.
 		html: false,
-
-		// @option element: DOM element = ''
-		// Custom DOM element to append inside the div element, empty by default. If html is also passed. The passed html wil be set.
-		element: false,
 
 		// @option bgPos: Point = [0, 0]
 		// Optional relative position of the background, in pixels
@@ -48,10 +44,12 @@ export var DivIcon = Icon.extend({
 		var div = (oldIcon && oldIcon.tagName === 'DIV') ? oldIcon : document.createElement('div'),
 		    options = this.options;
 
-		div.innerHTML = options.html !== false ? options.html : '';
-
-		if (options.element !== false && options.html === false && options.element.nodeType === 1) {
-			div.appendChild(options.element);
+		if (options.html === false) {
+			div.innerHTML = '';
+		} else if (typeof options.html === 'object' && options.html.nodeType === 1) {
+			div.appendChild(options.html);
+		} else {
+			div.innerHTML = options.html;
 		}
 
 		if (options.bgPos) {
--- a/src/layer/marker/DivIcon.js
+++ b/src/layer/marker/DivIcon.js
@@ -29,7 +29,7 @@ export var DivIcon = Icon.extend({
 		// iconAnchor: (Point),
 		// popupAnchor: (Point),
 
-		// @option html: String|DOM Element = ''
+		// @option html: String|HTMLElement = ''
 		// Custom HTML code to put inside the div element, empty by default.
 		html: false,
 
--- a/src/layer/marker/DivIcon.js
+++ b/src/layer/marker/DivIcon.js
@@ -33,6 +33,10 @@ export var DivIcon = Icon.extend({
 		// Custom HTML code to put inside the div element, empty by default.
 		html: false,
 
+		// @option element: DOM element = ''
+		// Custom DOM element to append inside the div element, empty by default. If html is also passed. The passed html wil be set.
+		element: false,
+
 		// @option bgPos: Point = [0, 0]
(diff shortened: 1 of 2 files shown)


src/layer/marker/DivIcon.js
```diff
@@ -33,6 +33,10 @@ export var DivIcon = Icon.extend({
 		// Custom HTML code to put inside the div element, empty by default.
 		html: false,
 
+		// @option element: DOM element = ''
```
IvanSanchez: The idea is good, but I think that the approach is wrong. Adding one more option might be confusing.

Instead, look at the API surface of `L.Popup`: there is a `setContent()` method which accepts **either** a string (of HTML text), or an instance of `HTMLElement`. The code then checks the type of that parameter and [either sets `innerHTML` or appends the child element](https://github.com/Leaflet/Leaflet/blob/master/src/layer/DivOverlay.js#L167).

I think the approach we currently follow in `L.Popup` (and `L.Tooltip`) is cleaner than having two parallel ways of adding content. What would you think about following this pattern instead?

MGMsystems: I was doubting to the same thing. I will adapt the code.

MGMsystems:
Adapted to the change request.

MGMsystems:
@IvanSanchez Can you review this merge request again?
Thanks!

IvanSanchez:
@MGMsystems I'd like the changes requeste at https://github.com/Leaflet/Leaflet/pull/5517/files#diff-6b6567b636fafffc6acd1c0b3987ced5R32 to be resolved

MGMsystems:
@IvanSanchez  I changed the code to either accept a `string` or a `HTMLelement` 
https://github.com/Leaflet/Leaflet/pull/5517/files#diff-6b6567b636fafffc6acd1c0b3987ced5R47
```javascript
if (options.html === false) {
div.innerHTML = '';
} else if (typeof options.html === 'object' && options.html.nodeType === 1) {
div.appendChild(options.html);
 } else {
div.innerHTML = options.html;
}
```

I've updated the documentation to say `HTMLelement`.
Or am I missing something in the change request?

IvanSanchez:
I think it's all good now. Nice job :-)

MGMsystems:
Thanks :)
Very fast responses from you guys!

MGMsystems:
@IvanSanchez When will you be able to release a new version with this change?
We kind of need it in our project.

ghybs:
Hi,

Sorry to comment after this PR has been merged, but it seems to me that it is somehow a duplicate of PR #2800, which has not been merged, if I understand correctly, mainly due to this [comment](https://github.com/Leaflet/Leaflet/pull/2800#issuecomment-49487087) from @mourner:

> The problem with this approach is that conceptualy, icon instance is something that defines how marker DOM elements are created, it doesn't represent DOM elements itself, so having references to a single DOM element there is awkward. E.g. you can't create an icon with a single DOM element and then use it for several markers. I agree this is a valid request, but we need to approach the API differently.

Here is a quick demo of re-using the same DivIcon (built from an HTMLElement) in more than 1 Marker, while this PR is effective:
http://playground-leaflet.rhcloud.com/guvas/1/edit?html,output
=> the HTMLElement is just moved to the last marker (expected behaviour of `appendChild`), hence any previous marker "loses" its content.

As stressed out by the above comment, the current API treats Icon as a generator, not as a placeholder.

@MGMsystems: if you need a different API that fits your specific situation (in particular, no re-use of an Icon), then why not just extending the `Icon` or `DivIcon` class? You could even turn it into a plugin if desired, as long as you clearly explain that such an icon _cannot_ be re-used, or by implementing some deep [cloning](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) feature. But I feel such a feature might be counter-productive, since one of the advantage of providing an exact HTMLElement is to keep a reference to it, is not it?

There might be some ambiguous wording to be fair.
How should we know that an `Icon` instance is supposed to generate many icons / DOM elements?
Whereas `Marker` is a specific instance at a given position, for example?
Maybe a class name like "`IconGenerator`" might have been more appropriate?

I have seen several SO posts with such mistake, but I could not find them right now.
That one is also related somehow: https://github.com/Leaflet/Leaflet.markercluster/issues/786

MGMsystems:
I can see why this might indeed not be a good idea. For us however we use the icons a bit different. We create a new icon for each marker and each icon has a different content.
That's how I forgot that you could use an icon in multiple markers and so losing it's content

perliedman:
@MGMsystems hi, and thanks again for providing this functionality! It's a convenient and logical extension of the API, similar to how `Popup`s and `Tooltip`s work.

However, @ghybs raises very valid concerns and I think it would be a mistake to release a new Leaflet version with the current changes, given that we know it simply will not work for a pretty common use case (re-using the same `DivIcon` instance for multiple markers).

I would suggest changing this so that the `html` option is either a string (as in earlier versions) or a _function_ that returns either a string or a DOM element; so this function is a factory method that can create the necessary DOM elements. This solves to problem of reusing an icon instance, and also aligns with `bindPopup`, which can also accept a function to generate the popup contents.

I imagine this would be not too much trouble to implement, would it be possible for you to make this change?

PR #5520, opened 2017-05-19
Added Interactive pan/zoom entry..

elrobis:
Added Interactive pan/zoom entry for L.Control.ZoomBar.


IvanSanchez:
Looks good!

I suggest adding a screenshot to your readme - it will make much easier to people to know how the plugin works without visiting the demo

PR #5525, opened 2017-05-22
Plugins: Adds Leaflet.Tilelayer.Colorizr

hnrchrdl:
I wrote a plugin for modifying the colors of tilelayers. Would be happy to add it to the list.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -300,6 +300,17 @@ Plugins for loading basemaps or GIS raster layers in common (albeit non-default)
 			<a href="https://github.com/IvanSanchez/">Iván Sánchez</a>
 		</td>
 	</tr>
+    <tr>
+		<td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
With other plugins such as the greyscale tilelayer, Leaflet.TileLayer.PixelFilter and L.TileLayer.GL, I don't see much benefit from this one. I guess it'll fit a small niche between those, so let's add it to the list.

hnrchrdl:
This plugin brings great flexibility that is not yet covered with the existing plugins: tweaking contrast, brightness, boosting or swapping color channels and much more. I will add a bunch of demos to show the purpose. Thanks.

IvanSanchez:
I will argue that L.TileLayer.GL does that (and more), and that Colorizr covers browsers without WebGL support. That's why I'm a bit reluctant.

hnrchrdl:
Ok, I was not aware that L.TileLayer.GL could do that, as this is not super obvious from it's name & description. It is your choice after all, thanks again.

PR #5526, opened 2017-05-23
Added in-browser print map plugin with print modes and custom print background.

Igor-Vladyka:
Added a link to an in-browser print plugin that allows you to choose one of 4 different modes for printing, and use custom tiles layer as print background. 

Supported v0.7.7 and v1.0.3 with all basic leaflet layers.

[Demo with text and custom print tiles layer](https://igor-vladyka.github.io/leaflet.browser.print/)


IvanSanchez:
Looking good!

PR #5529, opened 2017-05-24
Docs(Marker): modify icon option description

ghybs:
To explain that a _common_ / generic instance of `L.Icon.Default` is used, instead of leaving room to think that a _new instance_ of `L.Icon.Default` is created for every Marker without specified `icon` option.

See https://github.com/Leaflet/Leaflet.markercluster/issues/786#issuecomment-302893446

The current behaviour is to re-use the same instance for all markers with default icon.

Also replaced "icon class" by "icon instance", since it needs an actual _instance_ of `L.Icon`, not the class.

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -24,7 +24,7 @@ export var Marker = Layer.extend({
 	// @aka Marker options
 	options: {
 		// @option icon: Icon = *
-		// Icon class to use for rendering the marker.
+		// Icon instance to use for rendering the marker.
 		// See [Icon documentation](#L.Icon) for details on how to customize the marker icon.
 		// If not specified, a common instance of `L.Icon.Default` is used.
 		icon: new IconDefault(),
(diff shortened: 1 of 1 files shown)


IvanSanchez:
@ghybs Have you tried compiling the docs? Right now I'm not sure if the newline in the docstrings might break something

ghybs:
Hi @IvanSanchez,

No particular issue when building the docs, the table cell is filled with a normal inline text.
The HTML source code indeed reflects the line break from the docstring, but it does not affect the rendered text, as expected.

Many other descriptions use similar line breaks with no impact on compiled docs, e.g.:
- Tooltip [`direction`](https://github.com/Leaflet/Leaflet/blob/v1.0.3/src/layer/Tooltip.js#L36-L41) option ([docs](http://leafletjs.com/reference-1.0.3.html#tooltip-direction))
- Map [`maxBounds`](https://github.com/Leaflet/Leaflet/blob/v1.0.3/src/map/Map.js#L49-L54) option ([docs](http://leafletjs.com/reference-1.0.3.html#map-maxbounds))
- etc.

Leafdoc is great :-)

IvanSanchez:
> Leafdoc is great :-)

I should trust my own code more often :-)

PR #5532, opened 2017-05-25
Typo in Projection.LonLat.js doc.

jjimenezshaw:
In Projection.LonLat.js documentation, says it is used in EPSG:3395, while it is really used in EPSG:4326. It appears also in leaflet reference webpage.


IvanSanchez:
Thanks for noticing this!

Issue #5116, opened 2016-11-18
L.Control.Layer throws error if <input> inside layer label


IvanSanchez:
Small non-critical bug on an edge case, courtesy of @tomchadwin from [a gis.stackexchange question](https://gis.stackexchange.com/questions/218286/adding-input-to-leaflet-1-0-0-layers-control-throws-an-error/218290).

If a `L.Control.Layers` has an `<input>` inside the label for a layer, like...

```js
    var layersControl = L.control.layers({
      "<input id='positron-slider' type='range' min='0' max='1' step='0.1'></input> Positron": positron,
      "<input id='darkmatter-slider' type='range' min='0' max='1' step='0.1'></input> Dark Matter": darkmatter,
    }).addTo(map);
```

(Full playground at https://playground-leaflet.rhcloud.com/wipo/edit?html,output )

Then the logic inside the `_checkDisabledLayers` and `_onInputClick` will try to iterate through that input as well, and throw an error during the process.

The problem lies with the way that the code iterates through the checkboxes and radiobuttons. There is an in-depth explanation in the gis.stackexchange link above:

```js
		var inputs = this._form.getElementsByTagName('input'),
```

The idea is to replace that `getElementsByTagName()` query with a new private property like `this._inputs`. That should be a plain array, populated in the `_addItem` private method, with something like `this._inputs.push(input);`. Care should be taken when calling `removeLayer`.

This sounds fairly easy and non-critical, so I'm going to let this up for grabs (ping @yourfirstpr ). Anybody wanting to work on this should:

- Say something in the issue comments
- Fork the repo
- Clone the fork
- Copy-paste some file from the `debug/test` directory, and put the code from my playground in there. That's the easiest way to debug this.
- Make the necessary modifications to [`/src/control/Control.Layers.js`](https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.Layers.js)
- Test, test, test
- For bonus points, add a unit test to `/spec/suites/Control.LayersSpec`
- Commit the changes and push to your fork
- Create a pull request

IvanSanchez:
@serdarkacka @tomchadwin Anybody wants to take this one? Should be easy enough.

tomchadwin:
I'll take a look, but I'm not confident to be able to fix it myself, so by all means others can have a look as well.

tomchadwin:
[...]

Issue #145, opened 2011-06-20
Popup option closeMapOnClick/closePopupOnClick docs discrepancy


jgerigmeyer:
The docs refer to Popup option closeMapOnClick, but the code uses closePopupOnClick instead.

PR #5537, opened 2017-05-28
Fix for issue #5116

iZucken:
As this filters out custom inputs created from property names (supposedly using innerHTML), they still will become unusable whenever _update() is fired on control.
I believe that this is a different issue coming from design of a module.


iZucken:
@perliedman, Please, check line #145 - array gets flushed there. Also, do note that flush rather than in e.g. splice is needed, because after that in _update it gets repopulated by calling _addItem (line #272) on every layer again. That is also the cause for unbinding of events on DOM blocks inside captions.

perliedman:
@iZucken ah, sorry, missed that line.

Good analysis, that explains the issue clearly. So, given this, I interpret this as even if you sort of fixed the issue, it's still not really meaningful to have `input`s in the labels, at least not if you want to attach event handlers to them?

iZucken:
It can be. I can see many applications where a plain statement of this fact in documentation could suffice to a degree. Knowing that, user will just have to rebind his events after "destructive" operations.

iZucken:
I rebased commits to local master, but it appears that one cannot change head branch of a request (probably not truth). Should I create a new one (and close this)?

PR #5544, opened 2017-06-01
Fix for issue #5116

iZucken:
Rebased from #5537


perliedman:
Thanks for following through here!

PR #5554, opened 2017-06-10
Control position extend

perliedman:
This is a continuation of #5264.

Vertical and horizontal center is added with the new control positions `topcenter`, `leftcenter`, `rightcenter`, `bottomcenter` and `center` (vertical and horizontal center).

This is mostly @yigityuce's work. 

After some discussion, the consensus seems to be that the using the CSS terminology "middle" for vertical center is confusing, so we now use "center" for horizontal as well as vertical. We also happen to use the same terminology as [another map library](https://developers.google.com/maps/documentation/javascript/3.exp/reference#ControlPosition) (thanks @gassc for this hint).


mourner:
How do we go about legacy browsers that don't support transform here? Is there a fallback in behavior?

perliedman:
@mourner hm, yes, you're right, this will not work in IE < 9 if I understand correctly.

We'll have to bring out the annoying IE workarounds once again, I guess.

Sawalilaode (commented):
Tidak akurat gps nya

Malvoz:
Closing in favor of #8358, to discuss this thoroughly first.

PR #5561, opened 2017-06-15
Refactor(fitBounds): move sanity check into _getBoundsCenterZoom

ghybs:
after conversion of `bounds` argument into LatLngBounds, so that we avoid duplicating the conversion and inconsistency in several public API's that use this private method (namely [`flyToBounds`](http://leafletjs.com/reference-1.0.3.html#map-flytobounds)).

This PR tries to address https://github.com/Leaflet/Leaflet/issues/4345#issuecomment-308301449.

However I am not exactly sure if the remaining `isValid` sanity check is not totally redundant with `LatLng` conversion own checks.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -248,6 +248,10 @@ export var Map = Evented.extend({
 		options = options || {};
 		bounds = bounds.getBounds ? bounds.getBounds() : toLatLngBounds(bounds);
 
+		if (!bounds.isValid()) {
+			throw new Error('Bounds are not valid.');
+		}
+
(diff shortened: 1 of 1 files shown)


IvanSanchez:
If I read #4345 correctly, it means that the sanity check is not currently applied to `flyToBounds()` calls. Do you think you could add a unit test for that?

johnd0e:
> However I am not exactly sure if the remaining `isValid` sanity check is not totally redundant with `LatLng` conversion own checks.

It has no sanity check: https://github.com/Leaflet/Leaflet/blob/19da01346d8f69b075d02d9b714bbf5a0f2c03e5/src/geo/LatLngBounds.js#L33-L41

PR #5566, opened 2017-06-20
Plugins: add SuperMap iClient

songyumeng:
Plugins: add SuperMap iClient


danzel:
Yeah, this is a product that uses leaflet from what I can tell.

PR #5573, opened 2017-06-22
Docstrings(Map/Grid+TileLayer): improve minZoom and maxZoom explanations

ghybs:
so that it is more explicit how Map's options may get automatically computed from its Grid/Tile layers, and what is the exact effect of these options on Grid/Tile layers.

The explanations for Map `minZoom` and `maxZoom` options are taken from https://github.com/Leaflet/Leaflet/pull/4643#r66759552

Replaces / closes https://github.com/Leaflet/Leaflet/pull/4643
Fixes https://github.com/Leaflet/Leaflet/issues/4034

Note: the "By default the entire map." sentence on Grid Layer `minZoom` option looks misplaced. It might have been intended for the above `bounds` option. But in that case, I am not sure the wording is appropriate? I would have said that if not set, there is no horizontal limit to Tile Loading… which is what the current docstrings says anyway, by describing the opposite situation.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -106,11 +106,11 @@ export var GridLayer = Layer.extend({
 		bounds: null,
 
 		// @option minZoom: Number = 0
-		// The minimum zoom level down to which this layer will be displayed (included).
+		// The minimum zoom level down to which this layer will be displayed (inclusive).
 		minZoom: 0,
 
 		// @option maxZoom: Number = undefined
-		// The maximum zoom level up to which this layer will be displayed (included).
+		// The maximum zoom level up to which this layer will be displayed (inclusive).
 		maxZoom: undefined,
 
 		// @option maxNativeZoom: Number = undefined
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -106,11 +106,11 @@ export var GridLayer = Layer.extend({
 		bounds: null,
 
 		// @option minZoom: Number = 0
-		// The minimum zoom level that tiles will be loaded at. By default the entire map.
+		// The minimum zoom level down to which this layer will be displayed (included).
 		minZoom: 0,
 
 		// @option maxZoom: Number = undefined
-		// The maximum zoom level that tiles will be loaded at.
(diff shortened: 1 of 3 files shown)


src/layer/tile/GridLayer.js
```diff
@@ -106,11 +106,11 @@ export var GridLayer = Layer.extend({
 		bounds: null,
 
 		// @option minZoom: Number = 0
-		// The minimum zoom level that tiles will be loaded at. By default the entire map.
+		// The minimum zoom level down to which this layer will be displayed (included).
 		minZoom: 0,
 
 		// @option maxZoom: Number = undefined
-		// The maximum zoom level that tiles will be loaded at.
+		// The maximum zoom level up to which this layer will be displayed (included).
```
perliedman: I'm not a native english speaker myself, but isn't "inclusive" better than "included" in this context?

src/layer/tile/TileLayer.js
```diff
@@ -42,11 +42,11 @@ export var TileLayer = GridLayer.extend({
 	// @aka TileLayer options
 	options: {
 		// @option minZoom: Number = 0
-		// Minimum zoom number.
+		// The minimum zoom level down to which this layer will be displayed (included).
```
perliedman: Same here, possibly "inclusive" instead of "included".

src/layer/tile/TileLayer.js
```diff
@@ -42,11 +42,11 @@ export var TileLayer = GridLayer.extend({
 	// @aka TileLayer options
 	options: {
 		// @option minZoom: Number = 0
-		// Minimum zoom number.
+		// The minimum zoom level down to which this layer will be displayed (included).
 		minZoom: 0,
 
 		// @option maxZoom: Number = 18
-		// Maximum zoom number.
+		// The maximum zoom level up to which this layer will be displayed (included).
```
perliedman: Same here, possibly "inclusive" instead of "included".

ghybs:
Not a native english speaker myself either, but your comment sounds good to me: https://en.wiktionary.org/wiki/inclusive

Anyway we can still refactor this text later on.

perliedman:
:+1:

Issue #5585, opened 2017-06-27
Documentation Link Broken in GitHub Pages


cnatis:
## How to reproduce
Go to http://leafletjs.com/ click "Docs" link in heading

## What behaviour I'm expecting and which behaviour I'm seeing
I am seeing a 404, I expect the documentation page as before

http://leafletjs.com/reference-1.0.3.html

IvanSanchez:
Looking at this.

IvanSanchez:
I don't understand. https://github.com/Leaflet/Leaflet/blob/master/docs/reference-1.0.3.html is still there.

cnatis:
@IvanSanchez at least I can still view them for now in the repo, maybe an issue with github pages, but this was working yesterday before the 1.1.0 update

IvanSanchez:
Yeah, I don't see any problem on the repo side of things. Please use http://leafletjs.com/reference-1.1.0.html in the meantime.

IvanSanchez:
We've updated the link in the header to point to `1.1.0`, but the `1.0.3` docs still show a 404 error. I'll leave this open for a while to see if github pages magically decides to behave.

cnatis:
Thank you for the quick response! Great work!

ghybs:
As for the `1.0.3` broken page, could it be related to the missing Jekyll front matter in `1.1.0` page?
https://help.github.com/articles/configuring-jekyll/#front-matter-is-required

IvanSanchez:
The front matter in 1.1.0 is fixed already at https://github.com/Leaflet/Leaflet/commit/11d716f0964d8bc0d058ca09e9ba8003451b4b8d , but the 1.0.3 docs still are a 404 :-( 😖

ghybs:
I guess there is no specific Jekyll error message in the repo settings?
https://help.github.com/articles/viewing-jekyll-build-error-messages/#viewing-jekyll-build-failure-messages-in-your-repository

Might try to re-commit the file with a dummy modification?

ghybs:
before attempting this blind fix, what about temporarily switching the GitHub pages settings to use a `gh-pages` branch again, see if it makes any difference?
Then we can switch back to `master/docs`?

ghybs:
Here you go: https://github.com/Leaflet/Leaflet/tree/gh-pages
We can remove that branch again later on obviously…
Should we give it a try?

flother:
[...]

PR #5592, opened 2017-06-28
Whitelist the jekyll-redirect-from

cyrilf:
Since this [commit](https://github.com/Leaflet/Leaflet/commit/b6d21653faf8894f1db05e10958f55305767c27d) the `reference.html` links on the homepage don't work anymore: **well-documented API** or **API documentation**

Based on the [installation](https://github.com/jekyll/jekyll-redirect-from#installation) procedures of this plugin it's recommended to whitelist it if Jekyll use the safe mode (which I guess you're using as it's hosted on Github Pages).

I didn't try locally (so maybe double check) but hopefully it will solve this issue. :sparkles:

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -9,6 +9,9 @@ latest_leaflet_version: 1.1.0
 
 gems:
   - jekyll-redirect-from
+  
+whitelist:
+  - jekyll-redirect-from
 
 # Integrity hashes for both leaflet.js and leaflet-src.js
 # These will be shown in the downloads page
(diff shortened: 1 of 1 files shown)


perliedman:
Hi and thanks!

Weird thing is the redirect worked up until yesterday, when we _changed_ the redirect. Maybe something in GitHub's Jekyll setup changed since the site was last generated?

Anyway, you suggestion looks very promising, I've scratched my head over this since yesterday, so I'm going to merge this in the hope that it resolves the problem!

perliedman:
Unfortunately does not seem to have improved the situation :(

Related to #5585. I've contacted GitHub support, lets hope they can explain what's going on.

cyrilf:
You mean it worked on the live website or locally?  
Because based on your [Contributing page](https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#improving-documentation) you do `jekyll serve --watch` which may work.  
But I think Github Pages launch it with: `jekyll serve --safe` and the `safe` might be the issue.

But ok, :crossed_fingers: because I have played with Jekyll for years.

cyrilf:
Ok :sob:.
I tried; I failed. :see_no_evil:

perliedman:
It worked on the live website up until yesterday. Still works locally for me, but as you say that might be because of some other mode/config.

PR #5596, opened 2017-06-29
Prevent using functions from the global scope with es6

atd-schubert:
I got the following error while using leaflet as a submodule in our [angular2 implementation of Leaflet](https://github.com/yagajs/leaflet-ng2/issues/243).

```
ReferenceError: L is not defined
      at NewClass.disable (node_modules/leaflet/dist/leaflet-src.js:5659:7)
      at NewClass.removeHooks (node_modules/leaflet/dist/leaflet-src.js:12824:19)
      at NewClass.disable (node_modules/leaflet/dist/leaflet-src.js:5569:8)
      at MapComponent.set [as draggingEnabled] (lib/map.component.js:9:20941)
      <...>
```

I think this is because of the use of Leaflet functions from the global scope. I also try to find other occurrences and fix them, too.

After the fix everything seems to work fine for me.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -75,7 +75,7 @@ export var Draggable = Evented.extend({
 
 		// If we're currently dragging this draggable,
 		// disabling it counts as first ending the drag.
-		if (this._dragging === this) {
+		if (_dragging === this) {
 			this.finishDrag();
 		}
 
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -75,7 +75,7 @@ export var Draggable = Evented.extend({
 
 		// If we're currently dragging this draggable,
 		// disabling it counts as first ending the drag.
-		if (L.Draggable._dragging === this) {
+		if (this._dragging === this) {
 			this.finishDrag();
 		}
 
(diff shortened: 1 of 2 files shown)


IvanSanchez:
This is a very nice catch. We thought that we removed every reference to `L.` from the code, but apparently we were wrong. 😇

atd-schubert:
My chance to make my first contribution to Leaflet 😎

...after using it for years

IvanSanchez:
There's always a first time :-D

src/dom/Draggable.js
```diff
@@ -75,7 +75,7 @@ export var Draggable = Evented.extend({
 
 		// If we're currently dragging this draggable,
 		// disabling it counts as first ending the drag.
-		if (L.Draggable._dragging === this) {
+		if (this._dragging === this) {
```
perliedman: This should just be `_dragging`, since it's actually referring to a class global ("static" in Java/C# terminology) defined here: https://github.com/atd-schubert/Leaflet/blob/21f31fa825acaaab0f78663c00854c45b99db16b/src/dom/Draggable.js#L24

perliedman: This also highlights the fact that the comment is misleading, since it's not "this draggable" we're checking, but actually any draggable.

atd-schubert: I have not dived into this so deeply. Thanks for this advice! I already make a new commit...

perliedman:
Thanks, great work! :+1:

PR #5598, opened 2017-06-30
Use same condition for adding as well as removing double tap listener

### LINKED ISSUE #5588: Logic to remove Leaflet's double tap listener for dispatching simulated dbclick events presumes that such a listener exists, when it might not.
## How to reproduce

- Leaflet version I'm using: 1.0.3 and 1.1.1
- Browser (with version) I'm using: Microsoft Edge on a Windows 10 Tablet
- OS/Platform (with version) I'm using: Windows 10 Tablet

The logic to remove Leaflet's double tap listener for dispatching simulated dbclick events presumes that such a listener exists, when it might not. Leaflet 1.0.3 and Leaflet1.1.0 only add such a listener if the browser is not a pointer-based browser and is not Chrome, however the removal logic fails to take this into account, resulting in Javascript errors.

See lines 2234 and 2281 and of Leaflet 1.1.0/leaflet-src.js

Line 2234, inside function addOne():
	} else if (touch && (type === 'dblclick') && addDoubleTapListener &&
	           !(pointer && chrome)) {
		// Chrome >55 does not need the synthetic dblclicks from addDoubleTapListener
		// See #5180
		addDoubleTapListener(obj, handler, id);

Line 2281, inside function removeOne():
	} else if (touch && (type === 'dblclick') && removeDoubleTapListener) {
		removeDoubleTapListener(obj, id);

Line 2281 is missing the pointer and chrome conditions from line 2234. This leads to Javascript errors when removing client-based dblclick listeners for pointer-based browsers, such as Microsoft Edge.

- [ ] this example is as simple as possible
- [ ] this example does not rely on any third party code

Using http://playground-leaflet.rhcloud.com/ or any other jsfiddle-like site.

perliedman:
Removal logic in `DomEvent` is missing part of the condition used when adding a listener, which can cause Leaflet to try to remove a listener that has not been added before.

Fixes #5588.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -134,7 +134,8 @@ function removeOne(obj, type, fn, context) {
 	if (Browser.pointer && type.indexOf('touch') === 0) {
 		removePointerListener(obj, type, id);
 
-	} else if (Browser.touch && (type === 'dblclick') && removeDoubleTapListener) {
+	} else if (Browser.touch && (type === 'dblclick') && removeDoubleTapListener &&
+	           !(Browser.pointer && Browser.chrome)) {
 		removeDoubleTapListener(obj, id);
 
 	} else if ('removeEventListener' in obj) {


perliedman:
@hanginwithdaddo it would be great if you could test this change and verify that it solves your issue before we merge.

hanginwithdaddo:
@perliedman I apologize for not getting back to you sooner. I've been away for a number of weeks and finally had a chance to go back to this. I looked at your change it is the same that I made locally to a copy of Leaflet 1.0.3 and I know works. Given that I found this while using the Leaflet.Draw plugin, I can't test this until "Fix non-extendable objects regression of 1.1.0 #5658" gets pushed out as part of an official build. The draw plugin chokes because or #5658. At this point, I suggest merging this change into master so that it is available for the next release. Again, please accept my apologies.
Dave

perliedman:
@mourner @IvanSanchez @yohanboniface @ghybs anyone have time to review this?

I was always going to review my own code, but I _know_ I will mess up sooner or later if I start doing that.

IvanSanchez:
Counterpoint? Counterpart? Counterline-of-code? Me can no english.

PR #5600, opened 2017-07-02
addTo() can be called with a map or layer group.

Timmmm:
Does leafdoc support sum types?


perliedman:
And yes, Leafdoc supports the syntax you used here.

Issue #5594, opened 2017-06-29
L.DomEvent.disableScrollPropagation returns nothing


Eschon:
## How to reproduce

- Leaflet version I'm using: 1.1.0
- Browser (with version) I'm using: Chrome Version 59.0.3071.109
- OS/Platform (with version) I'm using: Mac OS X 10.11.6

I came across this problem when using a modified version the [Rrose Plugin](https://github.com/erictheise/rrose).
I reduced the code as much as possible for a reproduction example:
http://playground-leaflet.rhcloud.com/bijew/edit?html,console,output

The part of the code that causes the Problem is this:
```
L.DomEvent
    .disableClickPropagation(wrapper)
    .disableScrollPropagation(this._contentNode)
    .on(wrapper, 'contextmenu', L.DomEvent.stopPropagation);
```

Changing it to 
```
L.DomEvent
    .disableClickPropagation(wrapper)
    .disableScrollPropagation(this._contentNode);
L.DomEvent.on(wrapper, 'contextmenu', L.DomEvent.stopPropagation);
```
fixes the Problem. 

Prior to 1.1.0 this worked and the [Documentation](http://leafletjs.com/reference-1.1.0.html#domevent) states that `disableScrollPropagation` should return this (which should be `L.DomEvent`).

IvanSanchez:
Looks like we missed a `return this` during the rewrite to ES2015 modules, see https://github.com/Leaflet/Leaflet/blob/v1.1.0/src/dom/DomEvent.js#L182 and `disableClickPropagation` a few lines below.

This looks like an easy fix. @Eschon would you like to submit a pull request?

Eschon:
@IvanSanchez Thanks for the quick reply!  
Sure I can try to make a PR.

IvanSanchez:
Thanks for the PR :-D

I think there is a bigger issue here - in a ES2015 module, `this` doesn't have any meaning as far as I recall. Chaining stuff will not be possible using modules :-/

perliedman:
@IvanSanchez not sure I follow - when do we / would we want to chain using modules? Chaining seems to me to be something you do with an object's methods, where `this` has meaning, but I guess I'm missing some possible new feature here?

Anyway, I hope it's ok to close this and open a new issue if module chaining (or the lack of it) is something we need to discuss further.

perliedman:
[...]

PR #5601, opened 2017-07-02
Fix(DomEvent): DomEvent.off should return this

jelhan:
Following docs `DomEvent.off()` should return `this` but it was broken in 14c5f1602cdd337df14c37ea8df777578219f744. Leaflet 1.1.0 is affected by this but.


perliedman:
See https://github.com/Leaflet/Leaflet/issues/5594#issuecomment-311919846 for a potential gotcha regarding ES6 modules and `this`.

lasersox:
https://github.com/yafred/leaflet-responsive-popup is also broken by this issue.

PR #5604, opened 2017-07-04
GeoJSON remark about the reversed coords

e7o-de:
Oh man, that took me a while of debugging ... seriously? I mean, the way Leaflet is doing it is also for me the "correct" order, but a little sentence in documentation would have been great about that GeoJSON dudes failure. Was debugging a while because I thought my layer is somehow just hidden behind the map or I'm not passing the right data until I figured out my points being rendered in Somalia ...

### DIFF
--- a/docs/examples/geojson/geojson.html
+++ b/docs/examples/geojson/geojson.html
@@ -107,6 +107,8 @@ title: Using GeoJSON with Leaflet
 };
 </code></pre>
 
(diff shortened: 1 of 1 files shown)


perliedman:
Hi! To be fair, it isn't that GeoJSON is somehow wrong or weird, it's just that axis order in GIS is one of those subjects where there exist two competing conventions, both have their own merits, more or less like left- or right-driving countries, one is not more right than the other. For more information see this summary of lon/lat vs. lat/lon: https://macwright.org/lonlat/

Anyway, a great idea to add a comment about this, it's definitely one of those things that you _are_ going to get wrong when starting to work with web maps.

Issue #5610, opened 2017-07-07
Basic example yields a truncated map locally, works OK on JSFiddle


wsw70:
## How to reproduce

- Leaflet version I'm using: 1.1.0
- Browser (with version) I'm using: Chrome 59, FF 54.0.1
- OS/Platform (with version) I'm using: Windows 10

As a first time user, I tried to reproduce the example on the home page (with OSM) and in the Quick Start Guide (Mapbox). Both failed locally. The same code works on JSFiddle.

**I understand if this issue is closed as "localized" (because it works on JSFiddle/CodePen). I have however tried to debug as much as possible and a solution could help someone having the same issue.** If it is closed I will move the question to a more general forum (SO) and still update the case here, should there be a solution.

## What behaviour I'm expecting and which behaviour I'm seeing

Expected behaviour: a square map (see code below)

What I see: a square canvas with a truncated map: http://i.imgur.com/zoc1yTC.png (example uncommented in the code below, the commented one gives a map truncated horizontally at 50%)

The tiles which are not truncated can be moved, but the canvas is sometimes filled in with missing parts, sometimes no.

The network capacities are I believe sufficient (symmetrical 1 Gbps), there is no filtering and all sites work OK, including the examples on the Leaflet web pages.

There are no errors in the Chrome DevTools console. The source files are served locally  by a local server (`python -m http.server`), the result is the same if they are accessed via `file://...`

I retrieve (local server, the case which does not work) 16 tiles for this example, as well as `marker-icon.png` and `marker-shadow.png` - which indicates that the data is correctly downloaded (at least I assume that so many tiles suggest so). Interestingly, the CodePen version retrieves only 9 tiles (and the map is complete)

## Minimal example reproducing the issue

Working examples: https://codepen.io/WoJWoJ/pen/owPZQp and https://jsfiddle.net/zotwu4yw/

the exact same code ran locally:

HTML:

```
<!DOCTYPE html>
<html lang="en">
    <head>
    </head>
    <body>
        <div id="map"></div>
    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.1.0/dist/leaflet.css" />
    <script src="https://unpkg.com/leaflet@1.1.0/dist/leaflet.js"></script>
    <script src="worldmap.js"></script>
    <link href="worldmap.css" rel="stylesheet">
    </body>
</html>

```
Javascript (two examples)

```
/*
var mymap = L.map('map', {
    center: [51.505, -0.09],
    zoom: 13
[...]

PR #5614, opened 2017-07-07
Update quick-start example in docs to remind that CSS goes before JS

colbin8r:
docs(quick-start): add warning to place CSS before JS to quick start example (#5610)


perliedman:
Thanks! :+1:

colbin8r:
Thank YOU for such an awesome package!

Issue #5609, opened 2017-07-06
Cancel tile requests in _onTileRemove


jbccollins:
## How to reproduce

- Leaflet version I'm using: 1.1.0
- Browser (with version) I'm using: Google Chrome
- OS/Platform (with version) I'm using: Windows 10

## What behaviour I'm expecting and which behaviour I'm seeing

My jsfiddle demonstrates the issue. You can see what I'm talking about if you open up the network tab of the developer console and disable browser caching.

The zoomend event in my fiddle will trigger a url change for the tile layer (toggle between normal osm tiles and greyscale osm tiles). Despite the tile url changing the requests for tiles using the old url are not canceled.  While this is not a huge issue when using osm tiles (osm is super fast and caching is done well) it affects my production use case where our tiler uses dynamic tiling to generate "up to the minute" tiles. Tile generation is not cached and is kinda slow (which is outside of my control unfortunately) and stacking these kind of tile requests puts a burden on our server. If you examine the network tab when using my jsfiddle you will see that on zoomend both the greyscale and normal osm tile requests are completed although one of them will not be necessary.

Right now Leaflet currently will cancel some tile requests via the [_abortLoading](https://github.com/Leaflet/Leaflet/blob/6bfa764cb08d1010c13dc71b2050e457a4c0ad3d/src/layer/tile/TileLayer.js#L230) function by overriding the src attribute of the tile png when conditions dictate that a tile is no longer visible. I propose changing the _onTileRemove function in TileLayer.js to do something similar.

```javascript
  _onTileRemove: function (e) {
    e.tile.onload = null;
    e.tile.setAttribute('src', L.Util.emptyImageUrl); // add this line
  },
```

## Minimal example reproducing the issue
http://jsfiddle.net/yzsgdq13/3/
- [x] this example is as simple as possible
- [x] this example does not rely on any third party code

perliedman:
Hi, thanks for reporting. This does indeed sound reasonable: there is no reason to continue loading tiles when settings a new URL.

From a quick look, it looks like setting a new URL with `setUrl` normally causes a `redraw` - although your suggested change looks reasonable, perhaps `redraw` should call `_abortLoading` (if available), since there is no reason to continue loading if the tiles should be requested again (as the docs for `redraw` says).

Anyway, I would be open to merging this change if you supply a pull request!

jbccollins:
[...]

Issue #137, opened 2011-06-17
Zooming under Android 3 crashes the browser


mourner:
The browser crashes when zooming on Android 3 tablet. Have no idea why this happens and how to debug this, so any help is much appreciated.

kaynz:
Is this still an issue?
`		if (!L.Browser.android) {
			this.on('tileunload', this._onTileRemove);
		}`

My app is running fine on Crosswalk 22 (Chromium 52) without the upper check.

PR #5615, opened 2017-07-07
Cancel http requests for outdated tiles

jbccollins:
#5609


mourner:
I have a feeling we had this logic before but removed it for some reason. Worth digging the history

jbccollins:
@mourner There is a similar line in the ```_abortLoading``` function. @perliedman and I had a discussion on [the issue](https://github.com/Leaflet/Leaflet/issues/5609) about why I did not use or add anything to the ```_abortLoading``` function. Is this what you are referring to?

perliedman:
@mourner looks like you have a good memory: #137 - specifically these changes look related: https://github.com/Leaflet/Leaflet/commit/5f07e8bfc24fdc5e097974a051bc64b9ac82c3eb and https://github.com/danzel/Leaflet/commit/0b913df152a6763f2cacfdf867ddd7a677735fde

Don't know if these are still applicable, it looks like we don't do any checks for `Browser.android` anywhere in the code these days. I would also be surprised if this happens on Chrome/Android, so just checking the Android platform would not be ideal.

So, if we think this is still something we need to guard against, maybe add a new `Browser` check for the old Android only browser and add checks against this before setting `emptyImageUrl`?

perliedman:
@jbccollins sorry for keeping this waiting.

What I mean is that on Android, we should not cancel requests is this is the native Android browser (not Chrome, not Firefox, etc.). Not sure how we best detect that, but hopefully there's a way.

perliedman:
So, I merged this - if someone has an actual device with a stock Android browser, I'd very much like a test to see if this works, and if `L.Browser.androidStock` is actually set correctly - coding in the dark here.

Issue #5489, opened 2017-05-02
window.L not set if using a module system


perliedman:
With the new Rollup build, the UMD loader will not set `window.L` if you're using a module system, like when building with Browserify. 

This is a breaking change, and breaks a number of plugins, which rely on `window.L` being set instead of explicitly importing or requiring Leaflet.

Related discussion in #2364, where the conclusion was that 1.0 (or 0.8 as it was back then) _should_ set `window.L`.

ghybs:
Hi,

FWIW, many developers that I see asking questions (SO, GIS SE…) and using build tools (not so many people in the end, to be honest) indeed seem to expect `window.L` to be set.
E.g. see [this post on SO](http://stackoverflow.com/questions/42428251/initializing-leaflet-map-in-angular2-component-after-dom-object-exists/42431059#42431059), and even https://github.com/Leaflet/Leaflet/issues/4964.
And so do many plugins, as already mentioned.

Here is an [interesting post](http://ifandelse.com/its-not-hard-making-your-library-support-amd-and-commonjs/#update) that does propose a workaround to expose globally even through the UMD wrapper (it also cites the d3 example BTW).
Unfortunately I guess that would require making a Feature Request to Rollup?

As a side note, I think it could be interesting to provide a simple ES module entry file for power users actually using an ES module loader and expecting their build tool to optimize the bundle (I guess that it is what is implied in [that comment](https://github.com/Leaflet/Leaflet/issues/2364#issuecomment-285853971)).

Here is a [quick example](https://plnkr.co/edit/v44CXnKqxrRoSVxwMOfx?p=preview) in SystemJS.
It basically uses an entry file almost like [`src/Leaflet.js`](https://github.com/Leaflet/Leaflet/blob/master/src/Leaflet.js), but without the `version` import from `package.json` (which would require a special plugin for JSON in SystemJS, if I understand correctly) nor the `noConflict` code.

And [with a simple plugin](https://plnkr.co/edit/Z8yRSu3aN5F6zfhnTGfZ?p=preview) wrapped with a new `import` and `export` wrapper, instead of the standard UMD.
(using [Leaflet.FeatureGroup.SubGroup](https://github.com/ghybs/Leaflet.FeatureGroup.SubGroup) plugin, but could probably be applied to most plugins that have a build process)

As previously, even that example demonstrates that the biggest drawback is breaking plugins that do not adapt to the new pattern…

But at least this is just an extra option that is cheap to add.
[...]

PR #5620, opened 2017-07-10
Feat(ESmodule): provide an entry file compatible to module loaders

ghybs:
so that application developers can import Leaflet using any module loader (not just Rollup) and benefit from the latter's advantages, typically bundle optimization / tree shaking.

This is a follow-up from https://github.com/Leaflet/Leaflet/issues/5489#issuecomment-313845806

Here is the updated Plunkr demo, using the `index.js` entry file at https://rawgit.com/ghybs/Leaflet/8b7fba889368684c0066bbde151a5979033e4931/src/index.js instead of the inline one in the previous demo version:
https://plnkr.co/edit/987oafSDqBblon8gl0vD?p=info

In the case of **SystemJS**, this module loader automatically detects AMD usage.
Therefore plugins with such AMD wrapper can be used without issue, provided that the application developer imports the modified symbols from `leaflet`.
An easy way to achieve that is simply to import everything into the `L` namespace, in which case we re-create a similar situation as having `L` globally available:
```javascript
import * as L from 'leaflet';
import 'leaflet.featuregroup.subgroup'; // or any AMD-wrapped Leaflet plugin.

// Use as if leaflet and the plugin were loaded through browser <script> tags…
L.featureGroup.subGroup();
```
Demo with plugin: https://plnkr.co/edit/Do6MdN0V8I3e1WFSMrXB?p=preview

Now we should very probably do a similar exercise for other module loaders?


perliedman:
Ok, so lets see if I understand these changes correctly.

For new modern build systems that understand modules, they will `import` from this new `index.js`, which _just_ exports the Leaflet namespace as an object, but does not set any global, etc.

This will only work with plugins that are written using the same method, that also import Leaflet this way, and doesn't rely on the global `window.L`. However, they can still attach to Leaflet's namespace (since you use `L.featureGroup.subGroup` in your code).

A question regarding the last point, attaching to Leaflet's namespace, is if this is something we want to continue advocating as a "best practice" for plugins? I've really like the "attach to Leaflet's namespace" as long as we used a global (then there's at least only _one_ global). With modules, it occurs to me this is starting to be a bit counter-intuitive rather than helpful. Any thoughts on this?

I think you're right we need to try this out with other module loaders as well, I'm adding those as a checklist here:

* [ ] Webpack
* [ ] Rollup
* [ ] Browserify

colbin8r:
I'd like to weigh on the Leaflet plugin/global namespace issue. I'm writing an ES6 Leaflet plugin now. I wanted both to expose my ES6 modules and follow the best practice of attaching to the global `L`. Here's a gist showing some code that should do that: https://gist.github.com/colbin8r/795c89433a6bc0098cd4df30ccac4e04
(It should probably also check for the existence of a global L now that I think about it... 🤔)

It's been well-established that global variables are a bad practice in general. Also, having Leaflet depend on globals has been a thorn in my side (see #5613). Probably the best solution is to only expose `L` if you know you're in a browser environment (e.g. check for `window` or something).

mourner:
We should definitely do this, although keeping a separate `index.js` for this makes it harder to maintain. @ghybs want to explore if json2module-based solution looks nicer in Leaflet?

ghybs:
Hi @mourner,

As for the entry file, I would see a few possible options indeed:
- Use `json2module` from `d3`, but needs to duplicate data from `package.json` during the build step. Therefore in some cases that data may be missing (directly importing the entry file from a cloned repo, instead of from a built copy like `npm`).
- Import `index.js` in `Leaflet.js`, like a wrapper for Rollup build. A nice advantage is that the `index.js` could be as much lean and compliant with ES modules as possible (but the version issue from previous point could still be problematic), while we still provide some workarounds for the UMD build (global, `noConflict`…)

ghybs:
@perliedman,

> With modules, it occurs to me [attaching to Leaflet's namespace] is starting to be a bit counter-intuitive rather than helpful.

I would even say that keeping the namespace by default totally defeats one of the main points of ES modules! :-) I think modules should delegate naming to consumer code.

> However, they can still attach to Leaflet's namespace […]

Well I think this is actually specific to SystemJS, which automatically detects AMD-wrapped libraries and behaves accordingly. I think there is something similar with Rollup through [rollup-plugin-commonjs](https://github.com/rollup/rollup-plugin-commonjs) plugin.

I am trying to come up with a scheme to do something similar as Leaflet for the plugins: having a clean ES module (that does not try to rely on global), with a Rollup build to UMD version, that would behave more or less like currently.
Will post when I have something working.

ghybs:
Here is already a new commit to demonstrate the re-use of `index.js` entry file for ES module in `Leaflet.js` for UMD Rollup build, so that all exports are managed in a single file.

The resulting build is exactly the same.

jjimenezshaw:
@ghybs 
> I would even say that keeping the namespace by default totally defeats one of the main points of ES modules! :-) I think modules should delegate naming to consumer code.

Making a plugin makes sense to do it inside `L` namespace, so it is a bit redundant. However, while making a webpage "using" Leaflet, I would prefer to know what is Leaflet and what is not. Then the namespace makes totally sense for me. Even `Map` or other modules can conflict with my webpage modules (or get mixed in an ugly way). How do I proceed in this case?

ghybs:
Hi @jjimenezshaw,

Thank you for your input!

Please can you kindly clarify the situation(s) you are referring to, so that we make sure we understand each other?
- `import {pluginClass} from 'leaflet-plugin'` (ES6 module)
- `var plugin = require('leaflet-plugin')` (CommonJS)
- `require(['leaflet', 'leaflet-plugin'], callback)` (AMD)
- `<script src="leaflet-plugin.js"></script>` (currently global)

jjimenezshaw:
@ghybs Sorry, I misunderstood something. I can import Leaflet and use it normally with `import L from 'leaflet';` Later I can do `L.map` and other methods as usual (using `L` namespace I mean). Forget my previous post ;)

colbin8r:
@jjimenezshaw Even better, you can import only the parts of Leaflet you actually need:

```js
let myMap = new Map(...);
...
```

This allows for you to include in your final bundle or code *only the parts of Leaflet you actually depend on* through a technology called [tree shaking](https://github.com/rollup/rollup#tree-shaking). Since the `Map` module exported by Leaflet probably depends on other Leaflet components, those components are included as well, but you eliminate extraneous code from your final/production build.

jjimenezshaw:
@colbin8r Thanks. But if I want to 'keep' the `L`, how do I do "this"? (because "this" doesn't work)
```javascript
import { Map as L.Map } from 'leaflet'; 
let myMap = new L.Map(...);
...
```
See that I want `L` not only on Map, but on every dependency it imports.

colbin8r:
@jjimenezshaw I'm not sure why you would want to keep `L` other than for visual similarity to existing Leaflet code. If that's the case, then you may want to reconsider because I think Leaflet as a whole is moving away from namespacing everything under a global `L` anyway, so this style will likely change. One of the benefits of using ES6 modules is that you can use call the exports from other modules whatever you prefer.

With your last code example, you don't have direct access to any of the `Map` modules dependencies anyway because they aren't imported into your current scope. If you want access to them, you have to import them yourself.

But to answer your question, you have a few options.

1. Import everything like you were and forget about tree-shaking:
```js
import L from 'leaflet';
let myMap = new L.Map(...);
```

2. Manually namespace it yourself.
```js
// note the ES6 object literal here
// https://github.com/lukehoban/es6features#enhanced-object-literals
L = { Map };
let myMap = new L.Map(...);

Let me know if I missed something.

jjimenezshaw:
I like namespaces for libraries. I love it. All my life hearing 'Do not use global variables/functions', and as far as I see, when you import a module you create some stuff in the global scope. It was cleaner when there was only one, `L`.

On the other side, it is great including only what I use, and not everything.

But unfortunately I cannot have both things. It is a pity that I cannot import a module, and directly include it in a namespace I choose (and everything that imports it as well... like `L`).

Anyway, thanks for you help.

IvanSanchez:
Hhhhhmmm. Doesn't https://github.com/Leaflet/Leaflet/pull/6021 supersede this PR?

ghybs:
Hi @IvanSanchez,

Thank you for pointing it out!

Hum indeed, both PR address the same objective.
Having the global export and `noConflict` in Rollup config `outro` (as done in #6021) avoids having to split `Leaflet.js` in 2 chained imports (as done in this PR), at the expense of coding them as strings.

There is still the concern about importing the `version` from `package.json`, but that could be addressed in a separate PR.

I also quite advanced regarding plugins as ES modules and buildable through Rollup.
I will probably post that as a separate thread then.

PR #5626, opened 2017-07-13
Filter published files on npm

TrySound:
For now not gzipped leaflet package is 23mb. There's weird 4mb file `.goutputstream-N6HL8X` and a lot other stuff which increases node_modules size. `files` field can fix this with specified whitelist.


perliedman:
Hi and thanks! This does indeed look like a very good idea.

A question to understand this change: with this change, we include `src` although not strictly necessary (the `main` attribute is set to `dist/leaflet-src.js`). Is this to make it possible to import specific parts of Leaflet as modules, or is the intention to still supply a somewhat complete version of Leaflet? If the later, would it also make sense to include `spec`?

I'm not too familiar with how npm packages are normally distributed, so trying to understand what would be best for us here.

TrySound:
Yeah. Source is useful to reduce bundle size, classes are not treeshakable by bundlers. I don't see spec necessary in published package.

TrySound:
Sure, forgot to remove :)

PR #5628, opened 2017-07-13
Adds Leaflet.HtmlLegend to docs/plugins.md

ka7eh:


yohanboniface:
Thanks!

Adding a link to the plugin example/demo is a good habit, and it makes the plugin more attractive, and the plugins list more useful, so I'd suggest to add one, given there already is a [demo](https://consbio.github.io/Leaflet.HtmlLegend/) page :)

ka7eh:
Of course. Just added the link to the demo page.

yohanboniface:
👍

PR #5631, opened 2017-07-15
Remove reference to deps.js in L.VideoOverlay debug example

thachhoang:
This fix is similar to #5417. The L.VideoOverlay example was added shortly after so it missed out.


ghybs:
Thank you for catching it!

Issue #5622, opened 2017-07-10
Mouse Wheel Zoom Experience Changed


marioatmicello:
I'm seeing a different zooming experience on leaflet 1.1.0 compared to 1.0.0 when using the mouse wheel.

If I slowly move the mouse wheel forward / backward and change the zoom one level at a time I do not see any issue.

However, if I quickly spin my mouse wheel in either direction, the map completely disappears, then reappears after the desired zoom level is set. 

This happens on tiles that have already been loaded into the browser so I ruled out the possibility that this is a case of tiles that have no been loaded yet.

The "expected" behavior here is to have the tiles either be blurry or much smaller (based on whether I am zooming in or out) before the desired zoom level tiles could replace the current zoom level tiles.

I see this on leafletjs.com using Chrome, Edge and Firefox (all latest or close to it )

perliedman:
Hi!

It does indeed look like we have some kind of regression here:

* looking at 1.0.0, zooming in quickly clearly shows "blurry" tiles until new ones have loaded: http://playground-leaflet.rhcloud.com/kiwus/
* using latest Leaflet doing the same shows grey background until new tiles load: http://playground-leaflet.rhcloud.com/dada/

I would be suspicious of #5431. Lets look at what happens if we back out that patch first, otherwise I'll guess we have to bisect to track this down.

oliverheilig:
@perliedman  yes, can confirm this. Reverting the changes fixes the regression, but reintroduces the missing tiles. Again, searching for the code where these tiles are actually evicted...

oliverheilig:
@perliedman I've reverted the patch #5634 and found the real cause for #5381

perliedman:
Should be fixed by #5634, that was just merged.

PR #5634, opened 2017-07-16
Fix Missing Tiles Error #5381 While Maintaining Zoom Experience #5622

### LINKED ISSUE #5381: Missing tiles after fast zoom in and out with wouse-wheel

oliverheilig:
Ok,

i reverted the changes from #5431 and tracked down the true error of #5381: on _abortLoading() the incomplete tiles were removed as DOM-element, but not from the _tiles collection, which left them in a kind of zombie-state. So this now fixes #5381 while maintaining the behaviour mentioned at #5622.

You can test the (hopefully final) version here: http://176.95.37.29/vectortest/missing-tiles-fixed3.html

Hope everyone is happy now :sweat_smile:


perliedman:
@mourner if you have the time, could you check that this looks sane to you as well?

perliedman:
By accident, I stumbled upon this old comment from myself a couple of years ago: https://github.com/Leaflet/Leaflet/issues/3002#issuecomment-62960089

> Shouldn't the tiles be removed from the _tiles cache when they're aborted?

Looks like we've gone a long way to find out that yes, they probably should. :smile:

perliedman:
Merging this and hoping for the best.

Thanks again, @oliverheilig. Sorry this has dragged out for a while, I forgot about this still being open during my vacation. It would have been great if this had been merged before 1.2.0...

theashyster:
@perliedman I remember this post http://leafletjs.com/2016/09/27/leaflet-1.0-final.html and the part about new release cycle and I was really excited that you guys decided to switch the mindset to have shorter release cycles and smaller bugfix releases. So maybe after a week or two it would be a perfect time to do a release for version `1.2.1`.

It really is much more fun when new releases comes out more often and the bugs are squashed. Then you are able to remove those ugly hacks you have added just for the sake of fixing things till the next release of the library you use.

Keep up the good work 👍

PR #5639, opened 2017-07-17
Add plugin: Leaflet.UTFGrid

brendan-ward:
Note: this is different than the existing `Leaflet.utfgrid` plugin already listed on plugins page.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -527,7 +527,7 @@ The following plugins change the way that tile or image layers are displayed in 
 		<td>
 			<a href="https://github.com/consbio/Leaflet.UTFGrid">Leaflet.UTFGrid</a>
 		</td><td>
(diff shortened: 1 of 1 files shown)


danzel:
I like the highlighting.
Do you think this should replace the existing utfgrid plugin?

brendan-ward:
@danzel we created `Leaflet.UTFGrid` based on Leaflet 1.0, and have been trying to track Leaflet versions with this and other plugins we've created - so other than the feature highlighting, that is probably one of the bigger differences between the two at the moment.

This was inspired quite a bit by `Leaflet.utfgrid` (thanks for the work there!), but it looks like that plugin hasn't been updated in quite some time?  I haven't checked it for compatibility with Leaflet 1.0; if it is not it might make a case for at least labeling that one with the highest supported version of Leaflet in the plugin registry and README for that plugin.

PR #5643, opened 2017-07-18
util is now exported as a plain object

### LINKED ISSUE #5589: L.Util L.DomUtil is not extensible in leaflet 1.1.0?
Сan not add a custom property to an L.Util or L.DomUtil object since leaflet version 1.1.0:
https://jsfiddle.net/rendrom/Lj63c72u/

But it works in 1.0.3 and earlier:
https://jsfiddle.net/rendrom/scvpb4s2/

This is a new behavior or ES6 Rollup build result?

Example incompatibility in plugins:
https://github.com/w8r/Leaflet.Modal/blob/master/src/modal.js#L16

calvinmetcalf:
This changes the method of exporting the methods of util so that it's an extendable object

fixes #5589 and calvinmetcalf/leaflet-ajax#49

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -236,3 +236,27 @@ export function cancelAnimFrame(id) {
 		cancelFn.call(window, id);
 	}
 }
+
+export default {
+	cancelAnimFrame: cancelAnimFrame,
+	requestAnimFrame: requestAnimFrame,
+	cancelFn: cancelFn,
+	requestFn: requestFn,
+	emptyImageUrl: emptyImageUrl,
+	isArray: isArray,
+	indexOf: indexOf,
+	template: template,
+	getParamString: getParamString,
+	setOptions: setOptions,
+	splitWords: splitWords,
+	trim: trim,
+	formatNum: formatNum,
+	falseFn: falseFn,
+	wrapNum: wrapNum,
+	throttle: throttle,
+	lastId: lastId,
+	stamp: stamp,
+	bind: bind,
+	create: create,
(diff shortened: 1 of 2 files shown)


hyperknot:
Sorry, isn't this against the whole point of ES6 modules? `import * as Util from './Util'` is the recommended way to do this.

calvinmetcalf:
yes and no, they key thing is just `import * as Util from './Util'` exports a module object which is **not** a plain vanilla JS object which means when you assign it to L.Util that ends up acting very different then it did before.   In all other respects they are more or less identical and since neither of them use named exports none of the nice ES6 advantages apply to it.

The key thing is that the way it currently is actively breaks plugins, I've got bug reports of this for my plugins and I we've seen bugs in our sites that accidentally upgraded

calvinmetcalf:
@hyperknot the 'es6 way' would be to use named imports everywhere instead of `import * as foo from 'bar'` which would avoid creating any of the exotic module objects.

hyperknot:
@calvinmetcalf I think the grand issue here is that the ES6 module version of Leaflet breaks some plugins. Can you open an issue for this and explain what plugins are breaking in exactly what way? I believe the core team is more than happy to look for resolving compatibility issues for plugin developers.

calvinmetcalf:
leaflet draw and leaflet-ajax are the 2 I've run into, but anything that adds a method to any namespace in leaflet is going to be effected.

mourner:
Let's do a more generic fix like #5658 so that it's not only for `Util`.

PR #5651, opened 2017-07-20
feat(Marker.Drag): autoPan implementation

### LINKED ISSUE #3572: Move map at the edge of map on marker dragging

alenaksu:
Hi,
this is an implementation of auto panning for the marker (should fix #3572). Unlike #3643, this doesn't modify  the Draggable class.
It also implements a sort of incremental speed that will increase as far as you move closer the edge.

Hope it helps!

### DIFF
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -1,0 +1,373 @@
+describe("Marker.Drag", function () {
+	var map,
+	    div;
+
+	beforeEach(function () {
+		div = document.createElement('div');
+		div.style.width = div.style.height = '600px';
+		div.style.top = div.style.left = 0;
+		div.style.position = 'absolute';
+		document.body.appendChild(div);
+
+		map = L.map(div).setView([0, 0], 0);
+	});
+
+	afterEach(function () {
+		document.body.removeChild(div);
+	});
+
+	describe("drag", function () {
+		it("drags a marker with mouse", function (done) {
+			var marker = new L.Marker([0, 0], {
+				draggable: true
+			});
+			map.addLayer(marker);
+
+			var hand = new Hand({
+				timing: 'fastframe',
+				onStop: function () {
+					var center = map.getCenter();
+					expect(center.lat).to.be(0);
+					expect(center.lng).to.be(0);
+
+					var markerPos = marker.getLatLng();
+					// Marker drag is very timing sensitive, so we can't check
+					// exact values here, just verify that the drag is in the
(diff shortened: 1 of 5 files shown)


alenaksu:
I saw your changes and I have to say that we have decreased the speed in our app too, so I think 10 is better.

I'm looking forward to see this PR merged :)

src/layer/marker/Marker.js
```diff
@@ -36,6 +36,18 @@ export var Marker = Layer.extend({
...
 
+		// @option autoPan: Boolean = false
+		// Set it to `true` if you want the map to do panning animation when marker hits the edges.
+		autoPan: false,
+
+		// @option autoPanPadding: Point = Point(5, 5)
+		// Equivalent of setting both top left and bottom right autopan padding to the same value.
+		autoPanPadding: [50, 50],
+
+		// @option autoPanSpeed: Number = 100
```
theashyster: Isn't this value responsible about what will be shown in the docs as the default value?
If that is the case then these are wrong for both `autoPanPadding` and `autoPanSpeed`

spec/suites/layer/marker/Marker.DragSpec.js
```diff
@@ -0,0 +1,373 @@
...
+				}
+			});
+			var toucher = hand.growFinger('mouse');
+
+			toucher.wait(100).moveTo(300, 280, 0)
+				.down().moveBy(5, 0, 20).moveBy(290, 32, 1000).wait(100).up().wait(100);
+		});
+	});
+
+	// describe("#setIcon", function () {
```
theashyster: Why commenting out such a large chunk of tests and not just removing them?

perliedman: Ouch, my bad. `Marker.DragSpec.js` is a new file, but as you can see I copied it from another file and forgot to remove the old contents.

Thanks for keeping an eye on me! 😄

src/layer/marker/Marker.js
perliedman: :+1: Good catch!

spec/suites/layer/marker/Marker.DragSpec.js
theashyster: Heh, yep, saw that. No problem 😄

PR #5655, opened 2017-07-23
[PR] Better replace for icon image

xusy2k:
Whitenoise append a suffix not contemplated in replace

Found at https://github.com/makinacorpus/django-leaflet/issues/188

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -52,7 +52,7 @@ export var IconDefault = Icon.extend({
 		if (path === null || path.indexOf('url') !== 0) {
 			path = '';
 		} else {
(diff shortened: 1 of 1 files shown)


perliedman:
Hi, and thanks for notifying us of this problem.

From my understanding, this is yet another issue caused by some form of bundler or preprocessing tool messing with Leaflet's CSS, which in turn causes issues when Leaflet tries to figure out the icon's default URL. So, it's very similar to #4968, where Webpack turns the same URLs into data-URLs.

As discussed in #4968, we hopefully have a more general solution to this in the workings, which will hopefully resolve this problem with Whitenoise (which I honestly know nothing about), as well as Webpack.

I'm going to close this in favor of a more general solution, but I really appreciate the effort of not only notifying us, but also taking the time to submit a correct to the problem you were seeing! Thanks!

PR #5658, opened 2017-07-24
Fix non-extendable objects regression of 1.1.0

### LINKED ISSUE #5650: in 1.1.0 all objects are non extendable
summing up #5643 and #5589, due to how the exports are done with the switch to rollup the leaflet objects are now all non extendable, so no plugin can add a new method to `L` or `L.Util`. 

Suggestions: republish 1.0.3 as 1.1.1 ASAP, this will prevent things from breaking, then either publish 1.1.0 as 2.0.0 or redo the switch to rolllup avoiding using `import * as foo from 'bar'` and `export * from 'foo'`;

mourner:
Closes #5650 until we figure out a less hacky solution. Tested it locally and it seems to work. I put the redefine in `Util` because it is currently included first in the Rollup build, but this is fragile. The advantage of this solution is that it changes just a few lines of code. cc @calvinmetcalf @perliedman  @IvanSanchez


mourner:
Looks like this hack makes Phantom go crazy. Not sure how to fix the tests...

mourner:
OK the tests seem to be running OK!

jdalton:
This is breaking in an environment where `Object` methods are frozen and so they cannot be overwritten.

jdalton:
Maybe wrapping it in a try-catch would help.

jdalton:
PR created https://github.com/Leaflet/Leaflet/pull/6806

PR #5659, opened 2017-07-24
Make L.Draggable._dragging "public" again

yohanboniface:
To make Leaflet.Editable happy

See https://github.com/Leaflet/Leaflet.Editable/issues/130


yohanboniface:
> I think we should do Draggable._dragging rather than L.Draggable._dragging,

Good point! Fixed :)

PR #5660, opened 2017-07-24
Load retina after touch

KroniK907:
Previously the css would load the touch class after the retina class. This means that if viewed on a retina touch device, the retina sizing is ignored because the touch class overrides it, essentially making the retina class useless. 

This swaps the positions so that if a device is both a touch device, and a retina device, it will load the retina sizing correctly.


KroniK907:
Ignore this pull request. This is not an issue.

PR #5667, opened 2017-07-28
Make L.LineUtil._flat public as L.LineUtil.isFlat

yohanboniface:
Also, keep retrocompat for both L.LineUtil._flat and L.Polyline._flat (<1.1).


perliedman:
:+1:

PR #5669, opened 2017-07-29
Docs(TileLayer): workaround leafdoc issue with @ symbol

### LINKED ISSUE #5668: TileLayer url documentation rendered misses some text
While looking into a retina issue at leaflet-providers, I noticed that part of the text of the `L.TileLayer` documentation is missing.

[the rendered html looks like this](http://leafletjs.com/reference-1.1.0.html#tilelayer-url-template):

> A string of the following form:
> ```
> 'http://{s}.somedomain.com/blabla/{z}/{x}/{y}{r}.png'
> ```
> You can use custom keys in the template, which will be evaluated from TileLayer options, like this:
> ```
> L.tileLayer('http://{s}.somedomain.com/{foo}/{z}/{x}/{y}.png', {foo: 'bar'});
> ```

but the entire [line 29](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/TileLayer.js#L29) is missing.

@IvanSanchez probably a leafdoc issue?

ghybs:
when used as a comment in docstring instead of as a directive start.
Simply replaced by "`&commat;`" as a workaround for now.

Fix #5668

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -26,7 +26,7 @@ import * as DomUtil from '../../dom/DomUtil';
  * 'http://{s}.somedomain.com/blabla/{z}/{x}/{y}{r}.png'
  * ```
  *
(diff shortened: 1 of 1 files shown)


perliedman:
:+1:

PR #5670, opened 2017-07-29
Do not prevent default on source of simulated event if simulated event is unhandled

### LINKED ISSUE #5425: ios: dragging=false but scrolling on map not possible
## How to reproduce

- Browser (with version) I'm using: Safari/602.1
- OS/Platform (with version) I'm using: iOS 10

When `dragging` is disabled, i expect the page to scroll when i drag over the map on ios. instead, nothing happens. on android i can scroll by dragging if dragging=false.

On ios, it's not possible to scroll the page when dragging on the map:

https://jsfiddle.net/geg8yLe3/

This is similar to #5343

perliedman:
This allows for example `touchstart` events to propagate so that they can be used to scroll the map when `dragging` is disabled. This fixes #5425.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1152,6 +1152,17 @@ describe("Map", function () {
 			happen.click(layer._icon);
 		});
 
+		it('does not preventDefault on touches when no other listener catches simulated events', function () {
+			var checkNotPrevented = function (e) {
+				L.DomEvent.off(map._container.parentNode, 'touchstart', checkNotPrevented);
+				expect(e.defaultPrevented).not.to.be.ok();
+			};
+
+			map.dragging.disable();
+			var spy = sinon.spy();
+			L.DomEvent.on(map._container.parentNode, 'touchstart', checkNotPrevented);
+			happen.touchstart(map._container, {touches: [{target: map._container}]});
+		});
 	});
 
 	describe('#getScaleZoom && #getZoomScale', function () {
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1302,7 +1302,7 @@ export var Map = Evented.extend({
 			if (src === this._container) { break; }
 			src = src.parentNode;
 		}
-		if (!targets.length && !dragging && !isHover && DomEvent.isExternalTarget(src, e)) {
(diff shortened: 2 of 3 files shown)


herrernst:
This unfortunately now has a bad side effect: In Safari on iOS, **enabling** dragging now results in not only scrolling the map, but also the page. But this "double-scrolling" doesn't happen anymore after you have pinch-zoomed once. See attached video (this branch with default options). I have not found out why this happens, maybe you have an idea. 

leaflet-ios-safari-drag

perliedman:
@herrernst can you share a code example which exhibits this problem? Also, I don't have any iOS devices, so any help you could provide to examine why this happens would be of great help.

herrernst:
Okay, that wasn't so easy, but I found the reason: Safari on iOS 10 has a weird behaviour where `preventDefault` on `touchmove` events doesn't stop scrolling when the handler is added "later" (like in Draggable, where it is [added in the `touchmove` handler](https://github.com/Leaflet/Leaflet/blob/c29dabe2f99b0bfe15f590e0fea483d30b885d46/src/dom/Draggable.js#L119)):
http://www.openradar.me/28479522
http://www.openradar.me/28359335

So one thing you could do is to `preventDefault()` already in the `touchstart` handler, [like how it's done in Map.TouchZoom](https://github.com/Leaflet/Leaflet/blob/c29dabe2f99b0bfe15f590e0fea483d30b885d46/src/map/handler/Map.TouchZoom.js#L64).
I have done so in my fork, and it seems to work, I hope it has not any unwanted side effects: https://github.com/herrernst/Leaflet/commit/1eadeabc359ed0f319618d9d9f67fb91cc388c9e

(By the way, if you do have a Mac, you can install Xcode and use the iOS Simulator.)

perliedman:
Wow, thanks, really nice find! You just have to love these little quirks in certain OS/browser combinations... 😖 

I'll have to think for a bit about the potential consequences of your simple and elegant fix. Hopefully we can use it.

SimonEast:
Did anyone get any further with @herrernst's tweak? We're currently building a map application for the World Health Organisation and we'd *love* to have iOS scrolling fixed.

Issue #2033, opened 2013-09-18
Layer control not working on FF V23 & 24, Android 4.x


ICarryTheDustOfAJourney:
Almost no reaction when selecting a different base layer or overlay. The bullseye/checkmark changes, but not the map. 
In some rare cases there is some kind of reaction (race condition?)

Observable on SGS2 SGS3 N7, FF V23 & 24.
Not the case on Chrome. 
LL V0.6.4

PR #5672, opened 2017-07-30
When a layer is repeated in Control.Layers, behavior is not consistent.

jjimenezshaw:
If you include the same (base) layer in a Control.Layer more than once (obviously with a different name), the behavior is not consistent when selecting it. Usually no base layer at all is included on the map when it is selected (it works with the other layers perfectly), so you see it grey.

Why do I include a layer more than once? Well, if I want some logic in the names, maybe makes sense to have two names for the same layer. Could be:
`{ 'OpenStreetMap': osm, 'Satellite': sat, ... , 'OSM Color': osm, 'OSM B&W': osm_bw, ...}`

The problem is that two (or more) `input` has the same layerId, so the logic looking what to remove and add does not work in this case.

The fix just check if the map has the layer after the removal phase, so it is added again if was one of the repeated layers.

By the way #2033 (fixed here #2171) creates since then a strange effect, and calls twice the layers update. It was a bug in FF for Android that now it's very old. I would suggest to remove the hack, or call the hack only in FF for Android, as @JoaoFTCoutinho suggested in 2014. Should I do it?

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -373,7 +373,7 @@ export var Layers = Control.extend({
 			layer = this._getLayer(input.layerId).layer;
 			hasLayer = this._map.hasLayer(layer);
 
-			if (input.checked && !hasLayer) {
+			if (input.checked) {
 				addedLayers.push(layer);
 
 			} else if (!input.checked && hasLayer) {
@@ -386,7 +386,9 @@ export var Layers = Control.extend({
 			this._map.removeLayer(removedLayers[i]);
 		}
 		for (i = 0; i < addedLayers.length; i++) {
-			this._map.addLayer(addedLayers[i]);
+			if (!this._map.hasLayer(layer)) {
+				this._map.addLayer(addedLayers[i]);
+			}
 		}
 
 		this._handlingClick = false;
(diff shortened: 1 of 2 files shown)


src/control/Control.Layers.js
```diff
@@ -373,7 +373,7 @@ export var Layers = Control.extend({
 			layer = this._getLayer(input.layerId).layer;
 			hasLayer = this._map.hasLayer(layer);
 
-			if (input.checked && !hasLayer) {
+			if (input.checked) {
 				addedLayers.push(layer);
 
 			} else if (!input.checked && hasLayer) {
```
perliedman: Even if not strictly necessary, could you move this `hasLayer` check down to the layer removal, so that it works the same way as the loop for adding layers? Easier to understand the code when it is symmetric.

perliedman:
If possible, could you also add a unit test to cover your scenario? This would ensure that we don't accidentally break this in the future.

jjimenezshaw:
Sure, I will try now a unit test.

jjimenezshaw:
There you have a unit tests, that fails on master ;)

perliedman:
@jjimenezshaw thanks a lot! Wish all PRs where this smooth to handle! :grinning:

PR #5676, opened 2017-07-31
fixed #5652 panning inertia executes even when pausing

### LINKED ISSUE #5652: Drag panning inertia executes even if pan movement is paused at the end
## How to reproduce

- Leaflet version I'm using:
Latest (1.1.0) - whatever both Craigslist and CellMapper Beta maps are using
http://leafletjs.com/ demo maps
craigslist.org - "map view" of sales/houses/etc
cellmapper.net/testmap/
- Browser (with version) I'm using:
Chrome 58
- OS/Platform (with version) I'm using:
Sabayon Linux - Gentoo derivative - rolling release

Click and hold mouse button on map 
Drag Map to point you want it, perhaps quickly
Remove finger from button

Expected behavior: Map just stays in place
What I am seeing: Map pan continues further

http://leafletjs.com/ demo maps

Edit 2017-07-21: explicit reference to example, removed all caps text

Git-Lior:
slowingRatio may need to be tweaked for optimizing the result

### DIFF
--- a/src/map/handler/Map.Drag.js
+++ b/src/map/handler/Map.Drag.js
@@ -184,11 +184,19 @@ export var Drag = Handler.extend({
 		var map = this._map,
 		    options = map.options,
 
-		    noInertia = !options.inertia || this._times.length < 2;
+		    noInertia = !options.inertia || this._times.length < 2,
+
+			// Check if drag is slowing down
+		    slowingRatio = 2.5,
+		    length = this._positions.length,
+		    secondToLastPoint = this._positions[length - 2],
+		    thirdToLastPoint = this._positions[length - 3],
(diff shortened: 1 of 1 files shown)


perliedman:
Hi and thanks a lot for this! Looking at your changes definitely made me understand this problem much better than I had before.

While I think I understand your solution, I feel reluctant to enter more constants and calculations to the inertia calculation - it's already a bit hard to follow, IMHO.

I made an alternative fix that relies more on the already existing velocity calculations for inertia, but makes sure only recent positions are actually used (as it already is when you end the drag in the middle of a motion). I submitted this as #5690 - it's bit less extra logic, but I don't know if we can come up with some corner case covered by this PR that my solution doesn't cover.

perliedman:
Closing in favor of #5690.

Issue #5681, opened 2017-08-02
Control.Layers.js: TODO keyboard accessibility


jjimenezshaw:
I have been looking around `Control.Layers.js` recently, and there is a line in method `_initLayout` that says:

`// TODO keyboard accessibility`

I would kindly implement it (or at least try), but I do not know what is the purpose of "keyboard accessibility". With the current version, you can reach the control with the "tab", open it in Chrome with "enter" (in Firefox it opens automatically when focused) and navigate inside again with the tab and cursors. Standard (and slow) navigation with the keyboard.

What is missing? If we want to add another functionality, I can work on that. If it is just a leftover, I can remove the comment.

PS The comment was added in 2011 in 50e36cc

perliedman:
@jjimenezshaw hm, I don't think there's any real plans on keyboard accessibility for the layers control.

I think what you describe is sufficient, although we're certainly open to any suggestions for improvement if you or anyone else have them.

In summary, I think it's fine to remove that comment for now, and just improve this if someone figures out this is an accessibility issue and there's an actionable suggestion for improving it.

PR #5684, opened 2017-08-03
remove comment in Control.Layers.js: TODO keyboard accessibility

jjimenezshaw:
As mentioned in #5681, just remove the comment.


perliedman:
:+1: Thanks for helping us keep the code clean!

PR #5688, opened 2017-08-05
[Plz merge] Release 1.2.0

perliedman:
This contains preliminary changes for making the 1.1.1 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ Leaflet Changelog
 
 An in-progress version being developed on the `master` branch.
 
+## 1.1.1 (release date TBD)
+
+### API Changes
+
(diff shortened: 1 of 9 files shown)


perliedman:
Pasting a Gitter comment I made on the version number of this release, feel free to add your thought on this:

> Do you have thoughts on the version number? 1.1.1 seems logical since it's a "mostly bug fixes" release, but if we want to adhere more strictly to semantic versioning I guess it should be 1.2.0 since there is new functionality/API changes.
> I'm leaning towards trying to do semantic versioning correctly from here on, so it would be 1.2.0.

danzel:
I vote `1.2.0`, the non-extendable objects regression warrants it IMHO.

perliedman:
I think we have some sort of consensus that 1.2.0 is the right version for this!

PR #5689, opened 2017-08-05
Throw if argument to addLayer is not actually a layer

perliedman:
It's a somewhat common error trying to add something which isn't a layer with `addLayer` - it can for example happen when using a plugin written before Leaflet 1.0, where layers didn't have to inherit from `Layer`, but it can also happen in other cases like #5225.

This adds a clearer error message if a non-layer is accidentally passed to `addLayer`.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -419,6 +419,13 @@ describe("Map", function () {
 			map.addLayer(layer);
 		});
 
+		it("throws if adding something which is not a layer", function () {
+			var control = L.control.layers();
+			expect(function () {
+				map.addLayer(control);
+			}).to.throwError();
+		});
+
 		describe("When the first layer is added to a map", function () {
 			it("fires a zoomlevelschange event", function () {
 				var spy = sinon.spy();
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -155,6 +155,10 @@ Map.include({
 	// @method addLayer(layer: Layer): this
 	// Adds the given layer to the map
 	addLayer: function (layer) {
+		if (!layer._layerAdd) {
+			throw new Error('The provided object is not a Layer.');
+		}
+
 		var id = Util.stamp(layer);
 		if (this._layers[id]) { return this; }
 		this._layers[id] = layer;


perliedman:
BTW, I tried to add a test for this using @ghybs' example (http://playground-leaflet.rhcloud.com/teba/1/edit?html,output) as basis, but my prosthetic hand skills falied me - it works well when run as a standalone test, but not when run together with the automatic tests.

HarelM:
Hi, I didn't try to migrate to 1.2 yet but I did went through the changelog to see what has changed and got to this pull request.
I'm using angular and typescript as my framework for classes etc and using leaflet's framework is both an overkill and somewhat weird in my code.
Wouldn't the right check for a layer should be the API documented method `onAdd` instead of `_layerAdd` which is a non documented internal method that restricts the usage of other frameworks to create layers?
My layer code looks like so:
```js
export class PoiMarkerLayer extends L.Layer {
...
public onAdd(map: L.Map): this {
        this.enabled = true;
        this.updateMarkers();
        map.addLayer(this.markers);
        return this;
    }
```
As I said, I haven't tried to switch to 1.2 yet, but I assume it will stop working.
Any thoughts? I would very much like to hear what you think...
Here's a full class definition in case you are interested:
https://github.com/IsraelHikingMap/Site/blob/master/IsraelHiking.Web/sources/application/services/layers/routelayers/route.layer.ts

perliedman:
@HarelM since Leaflet 1.0, layers not have to have `onAdd`/`onRemove`, they actually have to extend `L.Layer`; in retrospect, `instanceof Layer` might have been a better check here, but the result is the same: you have to actually extend `Layer` to be able to add it as a layer.

If your layer doesn't have the `_layerAdd` method, it definitely will not work anyway, Leaflet uses that method when adding (see #5225). The only difference is that you got a much more cryptic error message before.

HarelM:
I'll double check myself later against the actual generated code, but the code I have linked to is working without using the `L.extend(...)` method using leaflet 1.1.0.
What you are basically saying is that if it used to work before this commit, it will continue to work after this commit, right?

perliedman:
@HarelM Yes. Look at the code for `addLayer`: https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L172: it will not call `onAdd` directly, it calls `_layerAdd`, which in turn call `onAdd` (https://github.com/Leaflet/Leaflet/blob/master/src/layer/Layer.js#L110) - if you don't have `_layerAdd` in your layer, it will not work.

PR #5690, opened 2017-08-06
No inertia if drag is stopped before ending the drag

### LINKED ISSUE #5652: Drag panning inertia executes even if pan movement is paused at the end

perliedman:
This is an alternative to #5676 that also fixes #5652.

Currently, we accumulate positions in an array while dragging, and successively push out old positions, maintaining a list of recent positions. However, if the drag is paused for a while, and then immediately ended (by releasing mouse button/touch), we don't update the list of recent positions to reflect the pause, so it can still look as though we were dragging quickly when the drag ended.

This patch makes sure we update the positions list right before we calculate the current drag velocity and inertia.

### DIFF
--- a/src/map/handler/Map.Drag.js
+++ b/src/map/handler/Map.Drag.js
@@ -129,15 +129,19 @@ export var Drag = Handler.extend({
 			this._positions.push(pos);
 			this._times.push(time);
 
-			if (time - this._times[0] > 50) {
-				this._positions.shift();
-				this._times.shift();
-			}
+			this._prunePositions(time);
 		}
 
 		this._map
 		    .fire('move', e)
 		    .fire('drag', e);
+	},
+
+	_prunePositions: function (time) {
+		while (this._positions.length > 1 && time - this._times[0] > 50) {
+			this._positions.shift();
+			this._times.shift();
+		}
 	},
 
 	_onZoomEnd: function () {
@@ -192,6 +196,7 @@ export var Drag = Handler.extend({
 			map.fire('moveend');
 
 		} else {
+			this._prunePositions(+new Date());
 
 			var direction = this._lastPos.subtract(this._positions[0]),
 			    duration = (this._lastTime - this._times[0]) / 1000,


perliedman:
BTW, I tried to add a test for this using @ghybs' example (http://playground-leaflet.rhcloud.com/teba/1/edit?html,output) as basis, but my prosthetic hand skills falied me - it works well when run as a standalone test, but not when run together with the automatic tests.

perliedman:
Here's a test that does work (as part of `Map.DragSpec.js`) when run in a browser, but not when run in PhantomJS; it appears panning only works in PhantomJS when `fastframe` mode is enabled, but this has the side effect that the last `wait(250)` doesn't delay long enough for the inertia to actually take place. Hopefully there's a way around this that I'm not yet aware off.

```js
		it("does not apply inertia when drag is paused before releasing", function (done) {
			var container = document.createElement('div');
			container.style.width = container.style.height = '600px';
			container.style.top = container.style.left = 0;
			container.style.position = 'absolute';

			document.body.appendChild(container);

			var map = new L.Map(container, {
				dragging: true,
				inertia: true
			});
			map.setView([0, 0], 1);

			var hand = new Hand({
				timing: 'frame',
				onStop: function () {
					var center = map.getCenter();
					var zoom = map.getZoom();
					document.body.removeChild(container);
					expect(center.lat).to.be.within(20, 22.0);
					expect(center.lng).to.be.within(-181, -179);

					done();
				}
			});
			var mouse = hand.growFinger('mouse');

			// We move 5 pixels first to overcome the 3-pixel threshold of
			// L.Draggable.
			mouse.wait(100).moveTo(200, 200, 0)
				.down().moveBy(5, 0, 20).moveBy(256, 32, 200).wait(1000).up().wait(250);
		});
```

PR #5730, opened 2017-08-24
Feature/#5727/optional esc on close

Mickyfen17:
References #5727 
Add option to disable Escape key default close popup functionalty.

### DIFF
--- a/src/map/handler/Map.Keyboard.js
+++ b/src/map/handler/Map.Keyboard.js
@@ -141,8 +141,8 @@ export var Keyboard = Handler.extend({
 		if (e.altKey || e.ctrlKey || e.metaKey) { return; }
 
 		var key = e.keyCode,
-		map = this._map,
-		offset;
+		    map = this._map,
+		    offset;
 
 		if (key in this._panKeys) {
 
--- a/src/map/handler/Map.Keyboard.js
+++ b/src/map/handler/Map.Keyboard.js
@@ -142,7 +142,6 @@ export var Keyboard = Handler.extend({
 
 		var key = e.keyCode,
 		map = this._map,
-		mapPopup = map._popup,
 		offset;
 
 		if (key in this._panKeys) {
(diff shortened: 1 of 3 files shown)


src/map/handler/Map.Keyboard.js
```diff
@@ -162,7 +163,7 @@ export var Keyboard = Handler.extend({
 		} else if (key in this._zoomKeys) {
 			map.setZoom(map.getZoom() + (e.shiftKey ? 3 : 1) * this._zoomKeys[key]);
 
-		} else if (key === 27 && map._popup) {
+		} else if (key === 27 && mapPopup && mapPopup.options.closeOnEscapeKey) {
```
yohanboniface: Not sure we need an intermediate variable to save out 2 chars on two occurrences ;)

Mickyfen17: 😄 I went back and forth deciding whether or not to use that variable. I can remove it and just go back to `map._popup` if that's best?

yohanboniface: I'd say so, yes :)

Mickyfen17: Change made 🙂

src/map/handler/Map.Keyboard.js
```diff
@@ -141,8 +141,8 @@ export var Keyboard = Handler.extend({
 		if (e.altKey || e.ctrlKey || e.metaKey) { return; }
 
 		var key = e.keyCode,
-		    map = this._map,
-		    offset;
+		map = this._map,
+		offset;
```
Motardo: Minor nitpick: Could you revert the white space here? It was 2 tabs of indentation plus 4 spaces of alignment. Thanks!

Mickyfen17: Spaces re-added.
For some reason when I first ran `jake test` I got 2 `Expected indentation of 2 tab characters but found 4` errors on line 144 & 145. Also linting errors in my text editor, therefore I just fixed the spacing as per the linter. 
Added back the 4 additional spaces of indentation and tests still passing.

IvanSanchez:
Looks good. Thanks!

PR #5731, opened 2017-08-24
Update VideoOverlay.js

Dzwiedzminator:
wrong keyword uses - className should be used instead
found this when testing new leaflet on ie8 - map doesn't load because of this


IvanSanchez:
Hhhmm, nice catch! Apparently this was introduced by #5661.

I guess it would be better to use `L.DomUtil.addClass()` instead - if the user is providing an instance of `HTMLVideoElement`, overwriting the existing CSS classes of that DOM element looks like a bad idea.

PR #5737, opened 2017-08-28
Replace the old Overpass plugin by its fork

GuillaumeAmat:


mmd-osm:
Old plugin has severe performance issues, pages using is almost instantly get 429 Too many requests error, see briefkastenkarte.de for an example. Those issues are addressed by the fork.

Issue #5741, opened 2017-08-30
.getBounds().pad() not working with negative values


Shooshte:
This referes to a 2 year old issue found here:
https://github.com/Leaflet/Leaflet/issues/2673

The problem:
When passing negative values into the .pad() function, the result is not smaller than the current map container (-10 and 10 produce results that differ only slightly in size, but are both bigger than the map)

Steps to reproduce:
Open the leaflet homepage map

Enter:
```
L.rectangle(map.getBounds().pad(-10)).addTo(map)
L.rectangle(map.getBounds().pad(10)).addTo(map)
```
into the console

Zoom out, see two rectangles bigger than the actual map at the time of drawing.

CalvinWilliams1012:
I beleive that should fix the issue. However it may give a new issue as it does not check if the value is <-100 or >100.

danzel:
I think the issue here is your understanding of what the function does.

The parameter `bufferRatio` is a percentage, where 1 = 100%.
So 10 means, pad 1000% bigger, and -10 means -1000% bigger (which would be slightly smaller than 1000% I think).
To make it smaller you need to use a negative number that is between 0 and -0.5.
Padding by 0 would not change the size, padding by -0.5 would make it size 0.
(I think? Maybe 0.33333 would make it size 0)
This doesn't really make a lot of sense though, haha :). I don't think `.pad()` is designed for making a `LatLngBounds` smaller.
Maybe we could have another function that takes a percent, and resizes the box to that percent of the original size?

CalvinWilliams1012:
@danzel Ya you're right I tested it on the leaflet main page with rectangles as you can see in my mspaint pic 
![image](https://user-images.githubusercontent.com/15316956/29905263-8ff01e3c-8ddb-11e7-88e2-ccfd21cfcc84.png)

You can close/deny my pull request as I see what the function is for now. I am unsure of if we should add a new function for percentResize but I would be willing to add it :)

IvanSanchez:
I don't think this is a bug, it's just a misunderstanding of what the method does, and the fact that we are using the word "percentage" in the docstrings rather than "ratio".

Some clarification in the docstrings for that method (e.g. `Returns bigger bounds created by extending the current bounds by a given ratio in each direction. For example, a ratio of 0.5 extends the bounds by 50%. Negative values yield smaller bounds instead.` or something like that) would prevent this confusion in the future.

@CalvinWilliams1012, maybe you would like to make such a change in the documentation?

[...]

PR #5744, opened 2017-08-31
Pad method fixed (#5741)

CalvinWilliams1012:
New contributor here I think I've done everything right but let me know if I haven't.

The Pad method in src/geo/LatLngBounds.js was not doing what it said it was doing(see Issue #5741 ). 

The previous pad method was not using percentages(as the comment said it was) and would therefore give you the wrong bounds so I fixed it!

Here is the answers to the contributing questions:

1. Are you sure that this new feature is important enough to justify its presence in the Leaflet core? Or will it look better as a plugin in a separate repository?

It is a bug fix with an updated comment, so in my opinion very important in the core.

2. Is it written in a simple, concise way that doesn't add bulk to the codebase?

Very simple.

I ran the tests using
```
jake test
```
and did not get any errors other than in some file I didn't touch had 4 spaces instead of tabs but I let that be.

### DIFF
--- a/src/geo/LatLngBounds.js
+++ b/src/geo/LatLngBounds.js
@@ -77,12 +77,12 @@ LatLngBounds.prototype = {
 	},
 
 	// @method pad(bufferRatio: Number): LatLngBounds
-	// Returns bigger bounds created by extending the current bounds by a given percentage in each direction.
+	// Returns bounds created by extending or retracting by a percentage.
 	pad: function (bufferRatio) {
 		var sw = this._southWest,
 		    ne = this._northEast,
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I must reject this pull request. It radically changes the semantics of the `L.LatLngBounds.pad()` method, and makes some unit tests fail (see the Travis CI integration in this PR).

Also please don't disregard linting errors (the whitespace thing). If the linting doesn't pass, the unit tests don't run.

I appreciate the effort in sending a PR (I know it can be scary at first!) but unfortunately these changes don't help the codebase. Please don't let discourage you!

Issue #5739, opened 2017-08-29
Multiple canvas renders not binding popups


jaymmartin:
## How to reproduce

- Leaflet version I'm using: 1.2.0
- Browser (with version) I'm using: Chrome 61, Firefox 55
- OS/Platform (with version) I'm using: Windows 10

Steps 
1. Add a vector layer to the map using a specified canvas renderer and bind a popup to it
2. Add another vector layer to the map using a different canvas renderer and bind a popup to it

## What behaviour I'm expecting and which behaviour I'm seeing
I expect that the popup bound on the layer from step 1 to open when clicked, but it does not. The popup bound to layer 2 does work, however. 

The cursor pointer also does not change to indicate the step 1 layer is clickable.

## Minimal example reproducing the issue
https://codepen.io/jaymmartin/pen/yoQJzw

perliedman:
Hi!

Thanks for reporting this. The issue is that a canvas renderer acts as a layer covering the whole map container. Although the canvas is transparent (except for what's drawn on it), it will in fact swallow pointer events. So, putting two canvases on top of each other, only the topmost will receive the pointer events.

This is not something specific to Leaflet, but rather a fact of how the browser's DOM works.

BTW, I have a kludge workaround for this that you can have a look at: https://gist.github.com/perliedman/84ce01954a1a43252d1b917ec925b3dd

jaymmartin:
I see now. 

Multiple canvas renderers are useful, as redraw performance is much better if you can segment the layers on different canvases. For my case, I am using mouseover/mouseout, so that workaround will not work.

What about a solution where a canvas object forwards an event to the Map object, which then forwards it to the next canvas object? L.Map would need to track all of its canvas renderers and order in the DOM if it doesn't already. I'll see if I can get something out of this and update here.

```javascript
// L.Canvas
onAdd: function () {
	Renderer.prototype.onAdd.call(this);

	// register with map as canvas
	this._map.registerCanvasRenderer(this);

	// Redraw vectors since canvas is cleared upon removal,
	// in case of removing the renderer itself from the map.
	this._draw();
}

_handleMouseHover: function (e, point) {
	var layer, candidateHoveredLayer;

	for (var order = this._drawFirst; order; order = order.next) {
		layer = order.layer;
		if (layer.options.interactive && layer._containsPoint(point)) {
			candidateHoveredLayer = layer;
		}
	}

	if (candidateHoveredLayer !== this._hoveredLayer) {
[...]

PR #5752, opened 2017-09-01
Enable mouse events on multiple canvas renderers

jaymmartin:
Fix for #5739

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -342,14 +342,14 @@ export var Canvas = Renderer.extend({
 	// Canvas obviously doesn't have mouse events for individual drawn objects,
 	// so we emulate that by calculating what's under the mouse on mousemove/click manually
 
-	_dispatchEvent: function(e) {
-		switch(e.type) {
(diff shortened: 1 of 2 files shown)


perliedman:
Hi, and sorry for the late feedback on this PR.

I like the idea and the somewhat simple implementation, given the complexity of the issue.

However, I see an issue with how forwarding is done - to be useful, I guess that forwarding will need to be in top to bottom order, so that the topmost canvas layer handles the event first, and then forwards it to the layer directly underneath it. However, since new layers are just pushed to the `_canvasRenderers` array, there is no guarantee the render order will be the same as in that array.

I'm not sure how to best address this, but I really hope there's a good way solve this!

perliedman:
Ok, I accidentally merged this PR since I had multiple tabs with PRs open at the same time... as mentioned  in https://github.com/Leaflet/Leaflet/pull/5752#issuecomment-345266979, I don't think this is ready to merge yet. If you have time to continue discuss this, I'd be happy to do so if you open a new pull request. Sorry for any inconvenience.

Issue #5756, opened 2017-09-03
When an SVG file is used as an image overlay, expose its DOM to the browser


endquote:
I want to use an SVG file as an image overlay. This works fine, however Leaflet creates it as an `img` tag, which means I cannot access the DOM of the SVG file from code.

A way around this would be to use an `object` tag and nest it in there.

endquote:
I'm not sure if this PR will get accepted, so here's a monkey-patch way to achieve the same thing.

```
const overlay = L.imageOverlay('maps/map.svg', [[0, 0], [-height, width]]).addTo(this.map);

// Super-janky way of replacing the SVG <img> tag with an <object> tag that embeds it.
const layers = overlay._image.parentNode;
overlay._image.parentNode.removeChild(overlay._image);
const obj = document.createElement('object');
obj.data = overlay._url;
obj.type = 'image/svg+xml';
overlay._image.classList.forEach(c => obj.classList.add(c));
obj.setAttribute('style', overlay._image.getAttribute('style'));
layers.appendChild(obj);
overlay._image = obj;

// In Chrome, the object tag needs to be redrawn after a zoom for some reason.
// This causes an unfortunate one-frame flicker.
if (L.Browser.chrome) {
  this.map.on('zoomend', () => {
    overlay._image.style.display = 'none';
    overlay._image.style.display = 'block';
  });
}
```

perliedman:
As @IvanSanchez mentioned in response to #5757, the best way to achieve this is probably to inherit from `ImageOverlay` like `VideoOverlay` does, this way you can make the DOM accessible anyway you want. Avoid monkey patching if you can.

endquote:
Agreed that a new type of overlay like L.VideoOverlay makes much more sense. Thanks for tolerating the new guy.

PR #5757, opened 2017-09-03
add useObjectTag option to ImageOverlay

endquote:
Re: https://github.com/Leaflet/Leaflet/issues/5756

This adds a `useObjectTag` option `ImageOverlay`, which causes it to nest the image into an `object` tag. For SVG images, this exposes the SVG tree to the DOM. For raster images, it doesn't appear to make any difference.

Unfortunately there's a weird issue in Chrome where the contents of the `object` tag don't look right after a zoom. Perhaps this could be fixed elsewhere, but I just did a janky force-redraw.

### DIFF
--- a/src/layer/ImageOverlay.js
+++ b/src/layer/ImageOverlay.js
@@ -52,6 +52,10 @@ export var ImageOverlay = Layer.extend({
 		// @option className: String = ''
 		// A custom class name to assign to the image. Empty by default.
 		className: '',
+
+		// @option useObjectTag: Boolean = false
+		// Nest the image in an object tag. For SVG elements it allows access to the SVG DOM.
+		useObjectTag: false,
 	},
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm gonna reject this, because I think that adding more complexity to `L.ImageOverlay` is not the way to go.

Instead, have a look at the code for `L.VideoOverlay`, which extends `L.ImageOverlay`. See also https://github.com/Leaflet/Leaflet/pull/4988#issuecomment-250762538 - this here is a very similar scenario.

Also note that the unit tests are failing. Be sure to run `npm test` before committing your code to prevent this.

endquote:
Agreed that a new type of overlay like `L.VideoOverlay` makes much more sense. Thanks for tolerating the new guy.

PR #5772, opened 2017-09-10
Height width fix

### LINKED ISSUE #5636: contributions not visible in fullscreen map using chrome on android
## How to reproduce

- Leaflet version I'm using: 1.1.0
- Browser (with version) I'm using: chrome 59
- OS/Platform (with version) I'm using: android 6
- visit : http://leafletjs.com/examples/mobile/example.html

I'm expecting to see the contributions on the right bottom of the map, but i don't.

CalvinWilliams1012:
closes #5636 

Searched through all of the example files for occurrences of "100vh" and these three files is what I came up with. If I missed any let me know!

### DIFF
--- a/docs/examples/extending/class-diagram.md
+++ b/docs/examples/extending/class-diagram.md
@@ -13,7 +13,7 @@ title: Leaflet class diagram
 		maxBounds: bounds
 	});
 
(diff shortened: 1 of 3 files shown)


mourner:
Thank you!

PR #5789, opened 2017-09-20
docs show 'zoomanim' event under "Other events" section

luckylooke:
Seems to be copy-paste mistake "Other Methods" instead of "Other Events"


luckylooke:
aaa, didn't notice that it is generated docs.. I don't have whole project locally.. I will try to make some time, but it is not probable to be soon.. I am quite busy now :/

Issue #4981, opened 2016-09-30
Need unit tests for tileerror events


IvanSanchez:
Currently we don't have unit tests for the `tileerror` events on `L.GridLayer`s. This is particularly useful since the behaviour changes in #4912/#4980.

PR #5805, opened 2017-09-26
Add tileerror tests to GridLayer spec

msiadak:
First attempt at addressing #4981 

I'm relatively new to this, so feedback is appreciated both on the tests themselves and where I placed them in the spec.  I'd be happy to revise and resubmit.


mourner:
I think this looks great! Love seeing contributions to the test suite :)

PR #5822, opened 2017-10-04
use zoom parameter if passed

broncha:
We have an offline layer, which caches the tiles and we need to generate URL for various zoom level (https://github.com/mWater/offline-leaflet-map). We did override this method there but I think this should also be supported in leaflet itself!

It only supplies the z parameter from the current zoom level of the map right now

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -164,7 +164,7 @@ export var TileLayer = GridLayer.extend({
 			s: this._getSubdomain(coords),
 			x: coords.x,
 			y: coords.y,
-			z: this._getZoomForUrl()
+			z: coords.z ? coords.z : this._getZoomForUrl()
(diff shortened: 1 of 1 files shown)


lpikora:
Unfortunately it breaks TileLayer's zoomOffset option.  https://github.com/Leaflet/Leaflet/issues/6004

ghybs:
Hi,

Thank you @lpikora for pointing out that PR!

> we need to generate URL for various zoom level

Sounds to me that the proper place for such a modification / override for that purpose _might_ have been in `_getZoomForUrl` method then.

But in the first place, I think we could have spotted the regression earlier should this PR have been first reported as an issue, with all required information in it (in particular: Leaflet version, bug description and minimal test case reproduction), and this PR had included a test spec.

Follow-up of corrective action in issue #6004.

PR #5837, opened 2017-10-11
Docs(Download): add CDN alternatives

ghybs:
Hi,

This PR follows https://github.com/Leaflet/Leaflet/pull/5814.

Instead of duplicating the example code with several CDN's path (as done in the above mentioned PR), I propose to list the CDN alternatives at the end of the "Using a Hosted Version of Leaflet" section.

All 3 listed CDN's automatically update based on npm registry, so there is nothing to do on Leaflet side when releasing new versions.
(but `cdnjs` might need a PR on the [autoupdate configuration file](https://github.com/cdnjs/cdnjs/blob/master/ajax/libs/leaflet/package.json) in case the `dist/` folder changes)

An advantageous side effect is that we can link to the package "page" provided by these services, which usually lists the available files and gives access to different versions through a drop-down menu.

I also added a disclaimer to try preventing questions and/or complaints about these services, especially in case of availability issues.

Closes https://github.com/Leaflet/Leaflet/pull/5814

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -55,6 +55,8 @@ Leaflet is available on the following free CDN's:
 - cdnjs: https://cdnjs.com/libraries/leaflet
 - jsDelivr: https://www.jsdelivr.com/package/npm/leaflet?path=dist
 
+_Disclaimer: these services are external to Leaflet; for questions or support, please contact them directly._
+
 ### Using a Downloaded Version of Leaflet
 
 Inside the archives downloaded from the above links, you will see four things:
--- a/docs/download.md
+++ b/docs/download.md
@@ -32,7 +32,7 @@ so please read the changelog carefully when upgrading to it.
 
 ### Using a Hosted Version of Leaflet
 
-The latest stable Leaflet release is hosted on a CDN &mdash; to start using
+The latest stable Leaflet release is available on several CDN's &mdash; to start using
 it straight away, place this in the `head` of your HTML code:
 
     <link rel="stylesheet" href="https://unpkg.com/leaflet@{{ site.latest_leaflet_version}}/dist/leaflet.css" />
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm a bit worried about versioning. Do those CDNs host all released versions, and can users point to specific versions as well? If they switch to a higher version automatically I would expect trouble.

ghybs:
Absolutely fair concern!

As for hosting:
- `jsDelivr` advertises caching all files in their own [S3 permanent storage](https://github.com/jsdelivr/jsdelivr#usage) and automatically fetching new files from `npm` registry.
- `cdnjs` commits files into their [own GitHub repo](https://github.com/cdnjs/cdnjs/tree/master/ajax/libs/leaflet), [automatically updating](https://github.com/cdnjs/cdnjs/blob/master/CONTRIBUTING-WIP.md#21-auto-update-overview) from `npm` registry using their [`auto-update`](https://github.com/cdnjs/cdnjs/blob/master/auto-update.js) tool
- `unpkg` sounds to simply [cache files in the CDN](https://unpkg.com/#/about). It automatically updates alongside `npm` registry.

As for versioning:
They all follow `unpkg` scheme with specific `@version` or `/version` in each file URL path.

Actually, [`unpkg`](https://unpkg.com/) and [`jsDeliver`](https://www.jsdelivr.com/features#alias) allow semantic version range and `latest` (which is the point of your concern if I understand correctly).
`cdnjs` accepts only exact version, no version range or [`latest`](https://github.com/cdnjs/cdnjs#latest-version-url-support).

IvanSanchez:
> They all follow `unpkg` scheme with specific `@version` or `/version` in each file URL path.

@ghybs Could you please add one example for each of the CDNs, as we do in https://github.com/Leaflet/Leaflet/blame/v1.2.0/docs/download.md#L38 ? That'd make devs' lifes easier.

ghybs:
Hi,

While I personally agree one code example is enough (https://github.com/Leaflet/Leaflet/pull/5814#issuecomment-333402016), I think there could be an "elegant" solution to provide ready-to-use code for different CDN's while keeping the Download page "simple":

Like [slate code samples](https://lord.github.io/slate/#introduction), we could use _tabs_ for each CDN. We could even make the SRI inclusion a _checkbox_ option (somehow like [jsDelivr files configuration tool](https://www.jsdelivr.com/package/npm/leaflet?path=dist&tab=collection)).

But then we need some dedicated JS, ensure browsers compatibility, and would probably wonder about accessibility… Therefore the implementation would not be that "simple", and we would probably have to use a well tested library.

So if I may, I can propose to proceed in 2 steps:
1. Implement current PR as is, so that we can already close https://github.com/Leaflet/Leaflet/pull/5814.
2. Create a Feature Request to discuss further if we want to include more code examples, and how to do it (tabs? which library for implementation?)

(yeah I know I may be overthinking simple things sometimes…)

mourner:
Let's merge and follow up with any potential improvements later.

PR #5839, opened 2017-10-13
remove unfunctional plugin from listing

obama:
the plugin does not work in the current state and wasnt updated since 4 years so its obviously not maintained anymore


obama:
actually it works, i pasted the wrong lines

perliedman:
Thanks @obama !

(Sorry, but it's friday and it had to be said.)

PR #5855, opened 2017-10-19
Update docs for L.DomEvent.off()

### LINKED ISSUE #4731: L.DomEvent.off() undefined _leaflet_id on removing all events
Tested on 1.0.0rc1
According to the docs, `L.DomEvent.off(obj, types, fn, context)` "Removes a previously added listener function. If no function is specified, it will remove all the listeners of that particular DOM event from the element."
However this is not working as stated, if one tries to remove listeners from an object without passing the listener function `fn`, the `L.stamp(fn)` call at the beginning of `L.DomEvent._off()` tries to set property `_leaflet_id` of an undefined object and raises an exception.
Perhaps the docs are outdated?

77ganesh:
this commit fixes #4731


perliedman:
Thanks! Always great to see someone taking care of the details, especially a first time contributor! :+1:

Issue #5821, opened 2017-10-03
`moveend` fires prematurely, before adjusting/rounding of lat/lng occurs


yeedle:
A known issue with `setView` and `panTo` (see https://github.com/Leaflet/Leaflet/issues/426 and https://github.com/Leaflet/Leaflet/issues/5512 for example) is that it often sets the view to a slightly rounded lat/lng on higher zoom levels.

The issue is that the `moveend` event fires before the rounding occurs. Which means that if I set up an `once` event listener to fire on `moveend`, it will fire prematurely, and my events will occur not in the right time. (In my use case, I'm using `moveend` to open a modal, and automatically close it when the user moves the map. In the cases that rounding occurs, the modal opens and closes immediately because the latlng are rounded and the map moves after `moveend` has already been fired.)

## How to reproduce
- Leaflet version I'm using: Leaflet 1.2
- Browser (with version) I'm using: Chrome 61
- OS/Platform (with version) I'm using: Ubuntu 16.4

set up an event listener on `moveend` and then call `setView` while continuously calling `getCenter`, and watch how the center gets updated even after `moveend` has been called.

## What behaviour I'm expecting and which behaviour I'm seeing
I'm expecting `moveend` to fire after rounding has occured and the map won't move anymore.
I'm seeing `moveend` being fired and the map being moved ever so slightly afterwards.

## Minimal example reproducing the issue
click on the map once, see the lat/lng that `moveend` get's fired at. Click it again, and see the actual rounded lat lng. (This is tricky, because rounding is based on several factors like device size and pixel density so the bug will only occur under certain conditions)
http://jsfiddle.net/Yeedle/04y19h84/

perliedman:
Hi, 
thanks for reporting this and providing an example illustrating the issue (but please use a recent version in your example - the fiddle uses Leaflet 0.7.2 which is _really_ old - fortunately the same issue exists in the latest version as well).

After looking at this, I think I don't share your assesement that this is because `moveend` is fired before the rounding, but rather that two different methods for moving the map are used.

The first move, which changes zoom and center drasticly, does not pan or animate the zoom, but rather just moves the center and zoom in one step. Using this method, you get the _exact_ center coordinate that you asked for.

[...]

PR #5858, opened 2017-10-20
Round pan offset towards zero

perliedman:
As [observed in #5821](https://github.com/Leaflet/Leaflet/issues/5821#issuecomment-334979512), we currently use `floor` on the offset when panning, causing negative offsets to be rounded towards negative infinity, instead of towards zero, which is probably the intention.

This PR adds the `trunc`/`_trunc` methods to `Point` and use it to round the pan.

Since `Math.trunc` is only available in somewhat recent browsers, I included a polyfill.


IvanSanchez:
@perliedman Does this actually fix #5821?

perliedman:
@IvanSanchez from what I can tell with this updated fiddle, it fixes #5821: http://jsfiddle.net/apsrpk8k/

PR #5868, opened 2017-10-25
Redirect to Plunker, Codepen and JSFiddle

iH8:
Redirects to Plunker, Codepen or JSFiddle with corresponding template. For usage in issuetemplate, issues, StackOverflow etc.

Usage Examples:
- http://leafletjs.org/edit.html // Defaults to Plunker, latest version
- http://leafletjs.org/edit.html#provider:codepen;version:1.1.0 // Codepen, version 1.1.0
- http://leafletjs.org/edit.html#provider:jsfiddle;version:master // JSFiddle, master version.

### DIFF
--- a/docs/edit.html
+++ b/docs/edit.html
@@ -7,7 +7,7 @@
     <body>
         <script>
 
-            const config = {
+            var config = {
                 provider: 'plunker',
                 title: 'Leaflet',
                 description: "Leaflet is the leading open-source JavaScript library for mobile-friendly interactive maps. Weighing just about 37 KB of gzipped JS code, it has all the mapping features most developers ever need.",
@@ -18,11 +18,11 @@
                 js: "var map = new L.Map('leaflet', {\n\tlayers: [\n\t\tnew L.TileLayer('//{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {\n\t\t\t'attribution': 'Map data © <a href=\"http://openstreetmap.org\">OpenStreetMap</a> contributors'\n\t\t})\n\t],\n\tcenter: [0, 0],\n\tzoom: 0\n});"
             };
 
-            const hash = location.hash.substr(1);
+            var hash = location.hash.substr(1);
 
             if (hash) {
-                hash.split(';').forEach((str) => {
-                    const arg = str.split(':');
+                hash.split(';').forEach(function (str) {
+                    var arg = str.split(':');
                     config[arg[0]] = arg[1];
                 });
             }
@@ -39,11 +39,11 @@
             }
 
             function createForm (action) {
-                const form = document.createElement('form');
+                var form = document.createElement('form');
                 form.setAttribute('method', 'post');
                 form.setAttribute('action', action);
                 form.createInput = function (name, value) {
-                    const input = document.createElement('input');
+                    var input = document.createElement('input');
                     input.setAttribute('type', 'hidden');
                     input.setAttribute('name', name);
                     input.setAttribute('value', value);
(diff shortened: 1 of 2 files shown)


ghybs:
Hi,

Wow that is super nice! :+1:

How about linking to `dist/leaflet-src.js` instead of the minified version, to facilitate debugging?

I do not know Plunker API at all, but is it possible to specify tag attributes as well?
(typically to include SRI)

I guess that would be impossible in the case of CodePen and JSFiddle.

perliedman:
Yes, agree with @ghybs - it's definitely better to link `leaflet-src.js`, debugging with the minified source is a real headache.

iH8:
I knew Ghybs would be happy 👍 Will change to `leaflet-src.js` and send a PR. As for the tags in Plunker, they can be changed. Just change [the HTML template](https://github.com/Leaflet/Leaflet/blob/master/docs/edit.html#L61). As for SRI (which i'm not so familiar with),  i was assuming the SRI differs per version so that would be more difficult to implement since i don't know where to get the hash or how to generate it. But if that's needed, i'm willing to look into it. Just point me in the right direction.

ghybs:
Hi iH8,

Yes you are absolutely right, SRI is per file, therefore it is different for each version.

I do not think it is very critical for a debugging tool, but since we show a full HTML code in Plunker, I would try to show a good example that enforces SRI directly… :smile: 

As for implementation, I see that you just need to pass Plunker an HTML string indeed, that you have nicely templated.
In that case, you could try to match the requested version, and in case SRI hashes are known for that version, include them into the `<link>` and `<script>` tags.

I have just made such a list for some of Leaflet versions:
https://github.com/ghybs/Leaflet.MarkerCluster.Freezable/blob/ddb0f9a7df983d884c855f8fcbb7517d9f7b411b/docs/examples/mcg-freezable-libs-setup.js#L21-L55

You can also use the online [SRI Hash Generator](https://www.srihash.org/) tool for other files and versions if needed.

perliedman:
We already have the SRI hashes for the latest version here: https://github.com/Leaflet/Leaflet/blob/master/docs/_config.yml - we could but hashes for older versions there, or in a new YAML file in the `_data` directory, if we want to leverage Jekyll to solve this for us.

iH8:
Ok, i understand and i am all for good complete examples. I wrote this with in the back of my mind that we wouldn't have to maintain it or at least keep maintenance to a minimum. So i did not hardcode the versions, it's completely oblivious as to which version is latest. In order for that to work we would have to keep a list of versions and corresponding SRI's (which i don't think is a bad thing to have) and keep it up to date automaticly when doing a release. Now i see @perliedman's response and indeed, something like that would work. I'm working on the tooling, landed at release script at the moment and can easily take this along 👍 Only obstacle for me is that i would like it as a JSON file since that's the easiest to read from browser and write from node script. It's must more transportable than YML. Just output the SRI's to the config.yml like is done now with an ugly string replace and on top of that have a sri.json somewhere:
```
{
    "v1.2.0": {
        "source": "sha512-YLT+I34kEPlk5OqR5XObf40B7sInrIU+bGe5VcwSpfR5OrFVjExFxfhVoJQEPZQWMyB53o3AU/bb5J91nc8CPA==",
        "minified": "sha512-lInM/apFSqyy1o6s89K4iQUKg6ppXEgsVxT35HbzUupEVRh2Eu9Wdl4tHj7dZO0s1uvplcYGmt3498TtHq+log==",
        "css": "sha512-M2wvCLH6DSRazYeZRIm1JnYyh22purTM+FDB5CsyxtQJYeKq83arPe5wgbNmcFXGqiSH2XR8dT/fJISVA1r/zQ=="
    },
    "v1.1.0": {
        ...
    }
}
```

perliedman:
Jekyll can handle JSON as well, AFAIK, so that should be good. I'm just thinking it might be good if we can use the generated hashes in Jekyll templates (like we do today) as well as dynamically reading it from the client, if we need that.

Issue #4510, opened 2016-04-26
Document Event.layer


nathancahill:
It looks like event.layer holds the previous (child's) target when an Event is being propagated up through the parents: https://github.com/Leaflet/Leaflet/blob/master/src/core/Events.js#L265

I'm guessing it was written for FeatureGroups, so event listeners on the feature group can access the layer that was clicked. Reading the code, it looks like the property is added to any event that propagates.

Is that the case? Should it be documented alongside Event.target? Just want to make sure I'm not missing anything.

PR #5880, opened 2017-10-26
Clean up and document event propagation properties

perliedman:
I had somehow missed #4510 and was a bit surprised to find out about this undocumented hack to get `FeatureGroup` to work.

Thinking about it, I think the approach in itself is sane, but the real problem lies in the naming of the event property - `layer` is very specific to `FeatureGroup`'s utilization of the property.

On the other hand, there's probably a lot of code out there that use this property, so I'm going for deprecating this property, although actually removing it is not something we need to do soon.

Instead, I added two properties:

* `sourceTarget` - this is the original target of the event, the object it was first fire on before any propagation occured
* `propagatedFrom` - this is exactly the same as the current `layer`; that is, the previous object in the chain of object propagating the event (to be honest, I don't feel 100% about the name of this property, so feel free to suggest a better one)

Ideally, we should have made this analog to the DOM's `target` and `currentTarget`, but we already use the property `target` with the same semantic as the DOM's `currentTarget`, so that will unfortunately not work.

### DIFF
--- a/src/core/Events.leafdoc
+++ b/src/core/Events.leafdoc
@@ -21,7 +21,16 @@ The base event object. All other event objects contain these properties too.
 @property type: String
 The event type (e.g. `'click'`).
 @property target: Object
-The object that fired the event.
+The object that fired the event. For propagated events, the last object in
+the propagation chain that fired the event.
+@property sourceTarget: Object
+The object that originally fired the event. For non-propagated events, this will
+be the same as the `target`.
+@property propagatedFrom: Object
+For propagated events, the last object that propagated the event to its
+event parent.
+@property layer: Object
+**Deprecated.** The same as `propagetedFrom`.
 
 
 @miniclass KeyboardEvent (Event objects)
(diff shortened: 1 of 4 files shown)


src/core/Events.leafdoc
```diff
@@ -21,7 +21,16 @@ The base event object. All other event objects contain these properties too.
...
+The object that fired the event. For propagated events, the last object in
+the propagation chain that fired the event.
+@property sourceTarget: Object
+The object that originally fired the event. For non-propagated events, this will
+be the same as the `target`.
+@property propagatedFrom: Object
+For propagated events, the last object that propagated the event to its
+event parent.
+@property layer: Object
+**Deprecated.** The same as `propagetedFrom`.
```
IvanSanchez: Typo here: `propageted` vs `propagated`

perliedman: Nice catch :+1:, fixed now.

PR #5922, opened 2017-11-17
Remove touch based click tolerance

### LINKED ISSUE #5921: Remove touch based click tolerance for canvas renderer
The use of `L.Browser.touch` to increase click tolerance when using the canvas renderer (https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Path.js#L143) is confusing and produce incorrect events for features close to each other (#5339, #5144, probably more).

As [suggested by @IvanSanchez](https://github.com/Leaflet/Leaflet/pull/5339#issuecomment-280316123), it would probably be better to have the extra tolerance as an option settable in `Canvas`, and defaulting it to 0 (which is how it works with the SVG renderer anyway).

Muscot:
Fixes #5921

### DIFF
--- a/src/layer/vector/Renderer.js
+++ b/src/layer/vector/Renderer.js
@@ -34,7 +34,11 @@ export var Renderer = Layer.extend({
 		// @option padding: Number = 0.1
 		// How much to extend the clip area around the map view (relative to its size)
 		// e.g. 0.1 would be 10% of map view in each direction
-		padding: 0.1
+		padding: 0.1,
+
+		// @option tolerance: Number = 0
+		// How much to extend click tolerance round a path/object on the map
+		tolerance : 0
 	},
 
 	initialize: function (options) {
--- a/spec/suites/layer/vector/PathSpec.js
+++ b/spec/suites/layer/vector/PathSpec.js
@@ -80,5 +80,17 @@ describe('Path', function () {
 			}, 300);
 		});
 
+		it('it should return tolerance with stroke', function () {
+			var path = new L.Polyline([[1, 2], [3, 4], [5, 6]]);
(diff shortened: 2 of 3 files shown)


src/layer/vector/Renderer.js
```diff
@@ -34,7 +34,11 @@ export var Renderer = Layer.extend({
 		// @option padding: Number = 0.1
 		// How much to extend the clip area around the map view (relative to its size)
 		// e.g. 0.1 would be 10% of map view in each direction
-		padding: 0.1
+		padding: 0.1,
+
+		// @option tolerance: Number = 0
+		// How much to extend click tolerance round a path/object on the map
+		tolerance : 0
```
IvanSanchez: Shouldn't the default of this be `Browser.touch ? 10 : 0`?

Muscot: I think that we should try to build so little logic as possible round browser.touch in the library, because it's difficult to detect if it's a touch device or not. But still give the options for the users to be able to set tolerance. What do you think?

perliedman: Since I've supervised this PR, I think I will have to agree with @Muscot :smile: 

@IvanSanchez from a backwards compatibility point of view, you _are_ right, but I think the old default does not make sense: SVG does not have the same (larger tolerance when `touch` is true), and actually setting the tolerance to anything other than `0` will cause problems with lines/polygons that are right next to each other (see #5144).

So, I'd argue actually defaulting to `0` is what we've always should have done.

IvanSanchez: If @perliedman says it's right to break a bit of backwards compatibility, then I have no problem with that.

PR #5935, opened 2017-11-22
Use more stable form of Haversine formula.

### LINKED ISSUE #5932: Some points have non-zero distance to themselves
If `x` is e.g. `[60.1882535, 24.8300111]`, then `L.latLng(x).distanceTo(x)` -- which is expected to give zero -- gives 0.09493529796600342, i.e. ~10cm. If you change the latitude a bit, say, to 60.1882537, then the formula gives zero. Several other random points have this property, too (this exact same non-zero value for `L.latLng(x).distanceTo(x)`). For some values of `x` the distance is 0.13425878593145293. These are the only two non-zero values I've seen in this context.

Leaflet version: 1.2.0 stable.

JSFiddle to check distance of random points to themselves: https://jsfiddle.net/egrqg3qa/4/

Reproduced on Windows 10 with
- Chrome Version 62.0.3202.94 (Official Build) (64-bit)
- Firefox 57.0 (64-bit)
- Microsoft Edge 40.15063.674.0

jussimattas:
This PR fixes #5932. The new formula does not give a non-zero distance from a point to itself. For pairs of distinct points, the new formula gives the same results as the old one, to precision ~1e-7 (this large differences are basically only encountered with near antipodal points, more often the difference is much smaller, around 1e-9).

JSFiddle comparing the versions: https://jsfiddle.net/Ljacqhpj/1/

Performance-wise the formulas are pretty much identical.

### DIFF
--- a/src/geo/crs/CRS.Earth.js
+++ b/src/geo/crs/CRS.Earth.js
@@ -24,10 +24,9 @@ export var Earth = Util.extend({}, CRS, {
 		var rad = Math.PI / 180,
 		    lat1 = latlng1.lat * rad,
 		    lat2 = latlng2.lat * rad,
-		    dLat = (latlng2.lat - latlng1.lat) * rad,
-		    dLon = (latlng2.lng - latlng1.lng) * rad,
-		    a = Math.sin(dLat / 2) * Math.sin(dLat / 2) +
-		        Math.cos(lat1) * Math.cos(lat2) * Math.sin(dLon / 2) * Math.sin(dLon / 2),
+		    sinDLat = Math.sin((latlng2.lat - latlng1.lat) * rad / 2),
+		    sinDLon = Math.sin((latlng2.lng - latlng1.lng) * rad / 2),
+		    a = sinDLat * sinDLat + Math.cos(lat1) * Math.cos(lat2) * sinDLon * sinDLon,
 		    c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
 		return this.R * c;
 	}
--- a/src/geo/crs/CRS.Earth.js
+++ b/src/geo/crs/CRS.Earth.js
@@ -24,9 +24,11 @@ export var Earth = Util.extend({}, CRS, {
 		var rad = Math.PI / 180,
 		    lat1 = latlng1.lat * rad,
 		    lat2 = latlng2.lat * rad,
-		    a = Math.sin(lat1) * Math.sin(lat2) +
-		        Math.cos(lat1) * Math.cos(lat2) * Math.cos((latlng2.lng - latlng1.lng) * rad);
-
-		return this.R * Math.acos(Math.min(a, 1));
+		    dLat = (latlng2.lat - latlng1.lat) * rad,
+		    dLon = (latlng2.lng - latlng1.lng) * rad,
+		    a = Math.sin(dLat / 2) * Math.sin(dLat / 2) +
+		        Math.cos(lat1) * Math.cos(lat2) * Math.sin(dLon / 2) * Math.sin(dLon / 2),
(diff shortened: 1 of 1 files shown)


jussimattas:
Here's the Node snippet I used to evaluate performance:
```
const R = 6371000;
const distance = (latlng1, latlng2) => {
	var rad = Math.PI / 180,
	lat1 = latlng1.lat * rad,
	lat2 = latlng2.lat * rad,
	a = Math.sin(lat1) * Math.sin(lat2) +
      Math.cos(lat1) * Math.cos(lat2) * Math.cos((latlng2.lng - latlng1.lng) * rad);

	return R * Math.acos(Math.min(a, 1));
};

const distance2 = (latlng1, latlng2) => {
	var rad = Math.PI / 180,
	lat1 = latlng1.lat * rad,
	lat2 = latlng2.lat * rad,
	dLat = (latlng2.lat - latlng1.lat) * rad,
	dLon = (latlng2.lng - latlng1.lng) * rad,
	a = Math.sin(dLat / 2) * Math.sin(dLat / 2) +
      Math.cos(lat1) * Math.cos(lat2) * Math.sin(dLon / 2) * Math.sin(dLon / 2),
	c = 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
	return R * c;
};

const np = 100000;
let x1, x2, d1 = new Array(np), d2 = new Array(np);
console.time('d1');
for (let i = 0; i < np; i++) {
	x1 = {lat: 180 * Math.random() - 90, lng: 360 * Math.random() - 180};
	x2 = {lat: 180 * Math.random() - 90, lng: 360 * Math.random() - 180};
	d1[i] = distance(x1, x2);
}
console.timeEnd('d1');

console.time('d2');
for (let i = 0; i < np; i++) {
	x1 = {lat: 180 * Math.random() - 90, lng: 360 * Math.random() - 180};
	x2 = {lat: 180 * Math.random() - 90, lng: 360 * Math.random() - 180};
	d2[i] = distance2(x1, x2);
}
console.timeEnd('d2');
```

perliedman:
@mourner think your comments on this would be valuable. I don't really know the history behind selecting the current formula, but perhaps you remember?

src/geo/crs/CRS.Earth.js
```diff
@@ -24,9 +24,11 @@ export var Earth = Util.extend({}, CRS, {
 		var rad = Math.PI / 180,
 		    lat1 = latlng1.lat * rad,
 		    lat2 = latlng2.lat * rad,
-		    a = Math.sin(lat1) * Math.sin(lat2) +
-		        Math.cos(lat1) * Math.cos(lat2) * Math.cos((latlng2.lng - latlng1.lng) * rad);
-
-		return this.R * Math.acos(Math.min(a, 1));
+		    dLat = (latlng2.lat - latlng1.lat) * rad,
+		    dLon = (latlng2.lng - latlng1.lng) * rad,
+		    a = Math.sin(dLat / 2) * Math.sin(dLat / 2) +
```
IvanSanchez: Please store `Math.sin(dLat / 2)` in a separate variable. The current code will make two calls to `Math.sin()`, and I do not expect the JITs to magically optimize this. 

Same goes for `Math.sin(dLon / 2)`.

IvanSanchez:
I've set up a jsperf benchmark at https://jsperf.com/haversine-distance - it seems that @jussimattas is actually faster than the current implementation, even with the duplicated calls to `Math.sin()`, at least in FFX 57.

jussimattas:
Updated JSFiddle: https://jsfiddle.net/Ljacqhpj/2/

jussimattas:
Alright. Still needs @IvanSanchez's approval.

PR #5936, opened 2017-11-22
Optimise images.

grischard:
Saved 1.1 MB out of 9.7 MB. 22.3% per file on average (up to 62%).

Most of those are in the docs. The most significant is probably the 4.7% (2586 -> 2464 bytes) saving in dist/images/marker-icon-2x.png after running it through pngout.

For every file, [ImageOptim](https://github.com/ImageOptim/ImageOptim) compared the different tools and picked the smallest file that was produced by or by a combination of `advpng`, `pngout` and `zopfli` for PNGs, `gifsicle` for GIFs, `svgo` for SVGs or `jpegoptim` and `mozjpeg` for JPEGs.

No related issues, no code change, no functionality change, no visual differences in images (compression is lossless).


IvanSanchez:
I wasn't aware that we could squeeze the marker image a further 5% 😄

PR #5941, opened 2017-11-28
Docs: add mouseup event, update

germansokolov13:
The `mousedown` is documented and the `mouseup` isn't but it works! In the Interactive Layer and those inheriting.


IvanSanchez:
Looks good. Thanks for catching this!

PR #5944, opened 2017-12-01
Fixed #5848 - Map drags correctly when in disabled FieldSet

### LINKED ISSUE #5848: Panning doesn't release mouse in firefox when map is inside a disabled fieldset
## How to reproduce

- Leaflet version I'm using: 1.2.0 (older versions are the same, also tested 0.7.7)
- Browser (with version) I'm using: Firefox 54.0 64 bit (also on 55, 56 on Windows)
- OS/Platform (with version) I'm using: Ubuntu Linux 16.04
- Tested in other Browsers/platforms: The problem does **not** occur in chrome/linux, chrome/windows, edge/windows

1. Place a leaflet map inside a HTML fieldset and set the disabled property.
2. Use the mouse to start panning

I expect normal panning behaviour.
When I am panning and release the mouse button, I am still panning. 

https://plnkr.co/edit/bU5RxPt11kWJqEgh1nKg

jguthrie100:
Dragging functionality was setting the `eventHandler` on the `document` object (which must've not been responding to events due to the `disabled` tag.

I changed it to assign the `eventHandler` directly to the relevant object (and interestingly enough the object specified when actually setting the mousedown (`_dragStartTarget`) clickhandler!)

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -116,8 +116,8 @@ export var Draggable = Evented.extend({
 
 		this._startPoint = new Point(first.clientX, first.clientY);
 
-		DomEvent.on(document, MOVE[e.type], this._onMove, this);
-		DomEvent.on(document, END[e.type], this._onUp, this);
+		DomEvent.on(this._dragStartTarget, MOVE[e.type], this._onMove, this);
+		DomEvent.on(this._dragStartTarget, END[e.type], this._onUp, this);
 	},
 
 	_onMove: function (e) {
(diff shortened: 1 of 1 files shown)


jguthrie100:
I was wondering if anyone reading this would like to have a go at pair programming a solution.

I've tracked down the bug as far as the wrong eventHandlers getting processed when the disabled tag is set.

When a mouseUp event occurs, the generic _handleDOMEvent function gets called, rather than the Draggable.js specific _onUp.

Then strangely, when clicking outside of the fieldset, the correct eventHandler gets called (_onUp) and the map stops dragging.

jguthrie100:
@cherniavskii

Falke-Design:
Suggest to fix #5848

johnd0e (changes requested):
Conflicts

PR #5949, opened 2017-12-03
Removed markdown syntax in plugin description

idawave:
Sorry to bother, just thought it would be cleaner with removing markdown syntax (as not parsed), as the docs file is exposed publicly from the website.


ghybs:
Hi,

Thank you for your nice concern regarding formatting on the website! :-)

If you wish, you can use `<em>` tag for italic font style and `<code>` for inline code styling.

idawave:
I'm fine with the current formatting, I don't think it needs more. The markdown was adding some mess but now that's fine.

Issue #5940, opened 2017-11-27
TileLayerWMS width and height report NaN with Tilesize of point


Spaction:
- Leaflet version I'm using: 1.2
- Browser (with version) I'm using: Chrome 62.0.3202.94
- OS/Platform (with version) I'm using: Windows 10

With update of Leaflet 1.2 the height and width of a WMS tile can be a point. The wmsParams however was not updated and results in a NaN value when this is done.

In Leaflet src js the following needs to be changed to allow for the tileSize to be a point (shows as line 11477, however just searching for wmsParams is enough to get you in the area)

wmsParams.width = wmsParams.height = options.tileSize * (options.detectRetina && retina ? 2 : 1);

perliedman:
Hi, could you please provide an example that reproduces the problem? It's not clear to me what we're talking about here.

It looks like the line you're referring to is in the built/bundled JavaScript; please refer to the actual source files instead. Even better, if you have a fix, open a pull request so we can check it out and merge it into the source.

IvanSanchez:
So the offending line seems to be:

https://github.com/Leaflet/Leaflet/blob/623fa6161e25338a85bbd05c229e632d542c0797/src/layer/tile/TileLayer.WMS.js#L83

And I guess this is a regression since #3573. I guess the fix is to make a call to `this.getTileSize()`, then multiply that vector by the retina factor with `multiplyBy`, then set the `wmsParams`'s width and height from that.

Bonus points if a new unit test is added (checking that the WMS params do in fact contain the width and height of the non-square tiles)

Spaction:
@IvanSanchez  you are 100% correct with what caused the issue.

The fix I did myself was to replace the line with:
```
 if (isNaN(options.tileSize)) {
            wmsParams.width = options.tileSize.x * (options.detectRetina && retina ? 2 : 1);
            wmsParams.height = options.tileSize.y * (options.detectRetina && retina ? 2 : 1);
        }
        else {
            wmsParams.width = wmsParams.height = options.tileSize * (options.detectRetina && retina ? 2 : 1);
        }
```
@perliedman 
**set-up is pretty simple:**

```
 var options = {
        tileSize: new L.Point(500, 300),
    };

var wmsLayer = L.tileLayer.wms(""www.testURL", options)
```

Then ensure the underlying wmsLayer.wmsParams.height & width are not NaNs themselves

Issue #5094, opened 2016-11-12
RemoveLayer from LayerGroup (GeoJSON) does not remove layer


sainsb:
Using leaflet version 1.0.1, Windows 7, Chrome Version 54.0.2840.99 m

I'm expecting to see layers removed from the L.GeoJSON when I call geojsonlayer.removeLayer(0);

Here is an example of how to reproduce:
http://playground-leaflet.rhcloud.com/soje/edit?html,output

Click the click button 3 times. The subsequent layers are not removed.

Am I missing something?

sainsb:
my bad

PR #5958, opened 2017-12-08
fix wms tilesize

songyumeng:
Fixes [#5094](https://github.com/Leaflet/Leaflet/issues/5940).Same issue. See discussion there.


IvanSanchez:
The bugfix looks fine 😄 but you introduced whitespace changes, and those make the linter fail ☹️ 

@songyumeng Could you please try and set the whitespace as it was before?

Issue #5960, opened 2017-12-11
Invisible vector layers on SVG Renderer after Renderer is removed then re-added on map


abigaildelarosa:
## How to reproduce

- Leaflet version I'm using: 1.2.0
- Browser (with version) I'm using: Chrome 62.0.3202.62 (tried with firefox and it does the same)
- OS/Platform (with version) I'm using: Ubuntu 16.04
- step 1
- step 2
Application summary:
I do have several geoJson files. I want to create buttons/links to select which file should be use. When the user click the button/link then a specific geoJson file will be display. Once the file is displayed several layers out of the file are created and then the user can select which layer does he needs to use.

## What behaviour I'm expecting and which behaviour I'm seeing
I am expecting to see the layers. Unfortunately the layers do not show. I checked the code, the dots are there but it seems that there is no color, but when I resize the window then the layers show in the map. 

## Minimal example reproducing the issue

*The javascript looks something similar like this:
As it appears in the documentation I am creating a map, then add to it a base layer. Then I create a layer control with the baselayer. 

var map = L.map('map');
      map.setView([40.703312, -73.97968], 10);
      var baselayer_1 = L.esri.basemapLayer('Topographic');
      baselayer_1.addTo(map);
      var baseLayer = {
          "Waiting for Selection": baselayer_1
      };
      layercontrol = L.control.layers(baseLayer).addTo(map);

Then in other function "multiple layers"(assume that I have another funciton to create layer), I create the layers out of the geoFile, with a specific color depending of x category.
baselayer_1.addTo(map);
layer_1 = createlayer(geoFile, "category for color in layer");
layer_2 = createlayer(geoFile, "category for color in layer");
allLayers = {
            "one": layer_1,
            "two": layer_2,
        };
layercontrol = L.control.layers(allLayers).addTo(map);

*On the onclick event button I do have something like:
1.- remove first all the layers from the map (because we might have layers from other geoFile)
2.- then remove old control.
3.- call the function to run multiple layers and load it.
<a href="#" onclick="map.eachLayer(function(layer){map.removeLayer(layer)});
    map.removeControl(layercontrol);
    displayDotsTest(geoFile1);
     ">geoFile1</a>

[...]

PR #5966, opened 2017-12-15
Fix(SVGrenderer): reset svgSize on remove from map

ghybs:
Hi,

This PR resets the `_svgSize` cache on SVG Renderer, so that when it is re-used and re-added on map, in `_update` method, the check on `_svgSize` triggers and re-assigns `width` and `height` attributes on the `_container`.

Otherwise, the latter no longer has width and height, therefore is invisible, until the viewport size is changed.

I will try to include a unit test later on.

Fix https://github.com/Leaflet/Leaflet/issues/5960


perliedman:
Nice :+1:

PR #5975, opened 2017-12-20
We'd like to add a new routing plugin (leaflet.tripgo.routing)

gimenna:
Doc update: Add leaflet.tripgo.routing in routing section.


gimenna:
thanks!

PR #5979, opened 2017-12-31
Set alt attribute for img-based Icons only

### LINKED ISSUE #5953: Markers should always have an alt attribute, even if it is empty
## How to reproduce

- Leaflet 1.2.0
- Chrome 62 / Ubuntu
- Inspect the marker with dev tools and look at the `alt` attribute on the marker image (or absence thereof)

When adding a marker to the map, if you don't specify any alt text, it does not render the `alt` attribute on the image. This can cause assistive tech such as screenreaders to read out the full path to the marker image `src` which is not a great experience. Automated accessibility checkers will flag this as a failure.

Images which are "decorational" should have an empty alt text such as ``. It is important that the `alt` attribute is _present_, and _empty_ in order to classify as a decorational image. The corresponding WCAG guideline can be found here:

https://www.w3.org/WAI/tutorials/images/decorative/

https://codepen.io/andymantell/pen/QOPXvR

msiadak:
Closes #5953 

I considered and ultimately decided against following @IvanSanchez's suggestion to create a private _setAlt function in the Icon and DivIcon classes.  Because the alt attribute is only valid for img elements in the context of a marker, we can just do a quick test for the icon's tagName and only set alt if it's an img.  This seemed like a cleaner solution to me than implementing _setAlt for img based Icon subclasses and implementing a no-op _setAlt for non-img based subclasses.

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -190,6 +190,21 @@ describe("Marker", function () {
 			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
 			expect(marker._shadow.parentNode).to.be(map._panes.shadowPane);
 		});
+
+		it("sets the alt attribute to an empty string when no alt text is passed", function () {
+			var marker = L.marker([0, 0], {icon: icon1});
+			map.addLayer(marker);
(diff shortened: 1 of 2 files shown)


mourner:
Thank you! Seems like a good fix.

PR #5980, opened 2018-01-02
Update README.md

rbrtmrtn:
Minor edits for style, simplicity, and readability


README.md
```diff
@@ -1,14 +1,14 @@
 <img width="600" src="https://rawgit.com/Leaflet/Leaflet/main/src/images/logo.svg" alt="Leaflet" />
 
 Leaflet is the leading open-source JavaScript library for **mobile-friendly interactive maps**.
-Weighing just about 39 KB of gzipped JS plus 4 KB of gzipped CSS code, it has all the mapping [features][] most developers ever need.
+Weighing in at about 39 KB of gzipped JS plus 4 KB of gzipped CSS code, it has all the mapping [features][] most developers ever need.
 
-Leaflet is designed with *simplicity*, *performance* and *usability* in mind.
+Leaflet was designed with *simplicity*, *performance* and *usability* in mind.
```
jonkoops, line 6: Why use past tense here? This makes it sound like Leaflet is no longer actively being worked on.

jonkoops (commented):
I don't see the added value of these changes.

Malvoz:
> I don't see the added value of these changes.

Several people have thumbed-up these changes for some reason. I approved them because I was under the impression that it needed a review per https://github.com/Leaflet/Leaflet/issues/7830#issuecomment-1047889429. But if they don't add any value perhaps it should just be closed?

jonkoops:
@Malvoz I think it's better to close this, the PR has been open since 2018 and a couple of thumbs up is not unexpected. This change seems quite arbitrary and does not really seem to improve much.

It's also not worth the time to review these kinds of things unless they fix actual glaring issues.

PR #6012, opened 2018-01-18
Enable mouse events on multiple canvas renderers #2

jaymmartin:
See #5752. I know this needs work, but creating for additional discussion.

Sorry for the delay here. I just got some time to look at this again. So if I'm understanding @perliedman correctly, there are cases where the order in which canvas renderers are added to the map is not necessarily the order in which they appear in the DOM.

If so, could someone point me in the direction of how/when that happens? Is it add canvas A, add canvas B, remove canvas A? Setting z-index on the individual canvas elements? Or something else entirely?

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -342,14 +342,14 @@ export var Canvas = Renderer.extend({
 	// Canvas obviously doesn't have mouse events for individual drawn objects,
 	// so we emulate that by calculating what's under the mouse on mousemove/click manually
 
-	_dispatchEvent: function(e) {
-		switch(e.type) {
-			case 'mousemove':
-				this._onMouseMove(e);
-				break;
-			case 'click':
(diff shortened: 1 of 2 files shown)


danwild:
Hey @jaymmartin, just a heads up that I started [a plugin](https://github.com/danwild/leaflet-event-forwarder) that does something similar. 

The challenging part for leaflet `v1+` is that it now [supports custom panes](https://leafletjs.com/examples/map-panes/), so the order which elements are added/removed, nor their order in the DOM can really be relied upon for knowing z ordering.

Falke-Design:
Does contain the same code as #5752 and should be more a issue then a PR

johnd0e:
> Does contain the same code as #5752

But 5752 was reverted.

jonkoops:
I'm closing this PR as the author is unresponsive and the changes have been stale for too long.

Issue #5190, opened 2016-12-14
preferCanvas problem with panes


mrit64:
## How to reproduce

- Leaflet version I'm using: 1.0.2
- Browser (with version) I'm using: Chrome and FF
- OS/Platform (with version) I'm using: Windows 10

## What behaviour I'm expecting and which behaviour I'm seeing
I have crated a map with preferCanvas option (or renderer: L.canvas()). It's working well without panes. When panes are associated with layers, SVG renderer is used despite the preferCanvas option.
Maybe because of the following instruction in L.Map._getPaneRenderer  function :

		if (renderer === undefined) {
			renderer = (L.SVG && L.svg({pane: name})) || (L.Canvas && L.canvas({pane: name}));

...which favour SVG renderer in all cases ?

Best regards

perliedman:
@mrit64 could you provide a running example on http://playground-leaflet.rhcloud.com/ or similar?

mrit64:
Here : 
https://jsfiddle.net/mrit64/wrq77j6e/

Lax-Phorell:
I have the same problem, but i think it is connected with
```javascript
getPaneRenderer: function (name) {
    if (name === 'overlayPane' || name === undefined) {
        return false;
    }
...
```
I think by default, Leaflet uses the overlayPane when rendering to canvas. If you use a different name for your pane, then
```javascript
renderer = (L.SVG && L.svg({pane: name})) || (L.Canvas && L.canvas({pane: name}));
```
kicks in and you end up with a SVG renderer and therefore `this.options.preferCanvas` isn't even checked.

IvanSanchez:
Closed by #6019

PR #6019, opened 2018-01-20
Respect the preferCanvas option in all panes

mjumbewu:
When creating a renderer for a pane other than `overlayPane`, use the same logic as default to determine whether to create a `canvas` or `svg` renderer.

(Current behavior is for SVG renderer to be created for all panes other than the `overlayPane`)

### DIFF
--- a/src/layer/vector/Renderer.getRenderer.js
+++ b/src/layer/vector/Renderer.getRenderer.js
@@ -1,6 +1,6 @@
 import {Map} from '../../map/Map';
-import {Canvas, canvas} from './Canvas';
-import {SVG, svg} from './SVG';
+import {canvas} from './Canvas';
+import {svg} from './SVG';
 
 Map.include({
 	// @namespace Map; @method getRenderer(layer: Path): Renderer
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Thanks for this! 😄 

@mjumbewu One small thing, though: next time please try to write something like "fixes #5190" in the description of your pull req; this makes it easier to keep track of the bugs - see https://help.github.com/articles/closing-issues-using-keywords/

jcimoch:
@IvanSanchez I think the issue is still present in latest leaflet version. If I force to use canvas and try to display GeoJSON on pane other than overlayPane, leaflet starts using DOM instead.

ghybs:
Hi @jcimoch,

If you think the bug is still there, please can you open a [new issue](https://github.com/Leaflet/Leaflet/issues/new), with relevant details?
Make sure to provide the info asked in the issue template (provided when you open the issue), and in particular a reproduction example.

That would greatly help in understanding the problem, and hopefully to fix it!
Thank you.

jcimoch:
@ghybs Yes, I will try to reproduce it in isolated env, maybe the issue is connected with my specific setup. I will let you know when it's ready.

PR #6021, opened 2018-01-23
add pkg.module

Rich-Harris:
This is a follow-up to a [Twitter conversation](https://twitter.com/Rich_Harris/status/955851611225706499) — essentially, it adds an ES module build of Leaflet which means more efficient bundles (and a slightly easier workflow for Rollup users, since they no longer need the commonjs plugin to use Leaflet).

It would also allow people to do this sort of thing in modern browsers:

```html
<script type='module'>
  import * as L from 'https://unpkg.com/leaflet?module';

  // look ma, no build step!
  let map = L.map('map', {
    center: [51.505, -0.09],
    zoom: 13
  });
</script>
```

### DIFF
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -25,7 +25,7 @@ const banner = `/* @preserve
 `;
 
 const outro = `var oldL = window.L;
-function noConflict() {
+exports.noConflict = function() {
 	window.L = oldL;
 	return this;
 }
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -3,8 +3,9 @@
 import rollupGitVersion from 'rollup-plugin-git-version'
 import json from 'rollup-plugin-json'
 import gitRev from 'git-rev-sync'
+import pkg from '../package.json'
 
-let version = require('../package.json').version;
+let {version} = pkg;
 let release;
 
 // Skip the git branch+rev in the banner when doing a release build
@@ -23,15 +24,33 @@ const banner = `/* @preserve
  */
 `;
 
+const outro = `var oldL = window.L;
+function noConflict() {
+	window.L = oldL;
+	return this;
+}
+
+// Always export us to window global (see #2364)
+window.L = exports;`;
+
 export default {
 	input: 'src/Leaflet.js',
-	output: {
-		file: 'dist/leaflet-src.js',
-		format: 'umd',
-		name: 'L',
-		banner: banner,
-		sourcemap: true
-	},
+	output: [
+		{
+			file: pkg.main,
+			format: 'umd',
+			name: 'L',
+			banner: banner,
+			outro: outro,
+			sourcemap: true
+		},
+		{
+			file: pkg.module,
+			format: 'es',
+			banner: banner,
+			sourcemap: true
+		}
+	],
 	legacy: true, // Needed to create files loadable by IE8
 	plugins: [
 		release ? json() : rollupGitVersion()
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "uglify-js": "~3.0.26"
   },
   "main": "dist/leaflet-src.js",
+  "module": "dist/leaflet-src.esm.js",
   "style": "dist/leaflet.css",
   "files": [
     "dist",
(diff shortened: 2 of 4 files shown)


IvanSanchez:
Wait, what? This is too simple and must be some form of sorcery 🧙✨

Rich-Harris:
dumbledore

IvanSanchez:
This seems to fail due to a failed test at https://github.com/Leaflet/Leaflet/blob/b2cb8f79e82a6ba956c23ff746c1bfc8a186d98c/spec/suites/core/GeneralSpec.js#L9
...because `noConflict` becomes is no longer part of the source, but part of the outro.

But this is... clever, insanely clever. And points to the possibility of loading the Leaflet source code without a build system if there were `index` files instead of `index.js` files in the `src` subdirectories.

Other than that, LGTM. Very.

Rich-Harris:
Could add a `export function noConflict(){...}` to the ESM build, though it's probably best not to as ESM eliminates the possibility of conflicts. Let me know if you need me to make any changes to the PR.

IvanSanchez:
@Rich-Harris Could you please try adding the outro to https://github.com/Leaflet/Leaflet/blob/34251900ae5e7c64800ff41b09476ff1c9e95109/spec/karma.conf.js#L50 , and then running `npm run test` in your machine?

Karma triggers its own Rollup build, with its own config there.

Rich-Harris:
I added it, but Karma doesn't seem to acknowledge the `outro` option. Not sure why. Unfortunately I can't look into it right now, am on deadline!

IvanSanchez:
@Rich-Harris Gotcha. The outro has to use `exports.noConflict` - notice how the removed code used `export function noConflict()` instead of just `function noConflict`.

This means that making this changes will make the build to work and the tests to pass again.

I can't seem to be able to add a commit to this PR. Rich, could you please add a commit with this?

```diff
diff --git a/build/rollup-config.js b/build/rollup-config.js
index 261f2f20..54f28562 100644
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -25,7 +25,7 @@ const banner = `/* @preserve
 `;
 
 const outro = `var oldL = window.L;
-function noConflict() {
+exports.noConflict = function() {
        window.L = oldL;
        return this;
 }
diff --git a/spec/karma.conf.js b/spec/karma.conf.js
index 1e2c128f..07bafb33 100644
--- a/spec/karma.conf.js
+++ b/spec/karma.conf.js
@@ -20,6 +20,15 @@ module.exports = function (config) {
                {pattern: "dist/images/*.png", included: false, serve: true}
        ];
 
+    var outro = "var oldL = window.L;\n" + 
+        "exports.noConflict = function() {\n" + 
+        "      window.L = oldL;\n" + 
+        "      return this;\n" + 
+        "}\n" + 
+        "\n" + 
+        "// Always export us to window global (see #2364)\n" + 
+        "window.L = exports;";
+    
        config.set({
                // base path, that will be used to resolve files and exclude
                basePath: '../',
@@ -52,7 +61,8 @@ module.exports = function (config) {
                                json()
                        ],
                        format: 'umd',
-                       name: 'L'
+                       name: 'L',
+                       outro: outro
                },
```

build/rollup-config.js
```diff
@@ -23,15 +24,33 @@ const banner = `/* @preserve
  */
 `;
 
+const outro = `var oldL = window.L;
+function noConflict() {
```
IvanSanchez: This has to be changed to `exports.noConflict = function()` too.

IvanSanchez: If not, the UMD bundle for the tests will work, but the one for release won't. 😐

Rich-Harris:
d'oh! I was close. Tests pass now!

> I can't seem to be able to add a commit to this PR

Yeah, I think you have to use the web interface, which is a pain

IvanSanchez:
LGTM now :smile:

IvanSanchez:
@mourner I think we can go ahead with merging the code and write some docs later. Something in the API docs about `noConflict`, and then a short blog post.

PR #6032, opened 2018-01-28
Fix DomEvent.off() Not Removing Events When Calling with One Argument

### LINKED ISSUE #6013: DomEvent.off() not working correctly when called with one parameter
## How to reproduce

- Leaflet version I'm using: 1.3.0
- Browser (with version) I'm using: Chome 63.0.3239.132
- OS/Platform (with version) I'm using: Linux
- Assing event handler on an element using DomEvent.on(element, 'eventType', callback) 
- Remove all event handlers from element using DomEvent.off(element)
- Expect that after firing event, the callback will not be executed. (It is not working)

```
var div = document.getElementById('leaflet');
var eventHandler = function(){
  alert('Handler fired!');
}
L.DomEvent.on(div, 'click', eventHandler);
L.DomEvent.off(div);
//now manually click on div element -> alert is shown
```
https://plnkr.co/edit/yQ1GzAFD9oDNqnV0Nhok?p=preview

## Possible solution
I think that problem is in DomEvent.js
```
(...)
	} else {
		for (var j in obj[eventsKey]) {
                        //this does nothing cause j variable is not parsed
			removeOne(obj, j, obj[eventsKey][j]);
		}
		delete obj[eventsKey];
	}
(...)
```
In this case `j` is a key which isn't event type. It is a event handler id (e.g. resize37_35)
In order to fix it, `j= j.replace(/[^a-z]/g, '')` or something similar.

caleblogan:
fixes #6013 

`removeOne()` uses the original function to get the id for the wrapper handler attached to the object. When calling `off()` with only one argument, we don't have the original function, but we do have `type` which will be equal to the id (`type, stamp, context`) as a string. It may make more sense to pull this logic up out of `removeOne()`, but I wasn't sure how to do that without changing a lot of code.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -61,7 +61,7 @@ export function off(obj, types, fn, context) {
 		}
 	} else {
 		for (var j in obj[eventsKey]) {
-			removeOne(obj, j, obj[eventsKey][j]);
+			removeOne(obj, j);
 		}
 		delete obj[eventsKey];
 	}
@@ -122,9 +122,14 @@ function addOne(obj, type, fn, context) {
 }
 
 function removeOne(obj, type, fn, context) {
-
-	var id = type + Util.stamp(fn) + (context ? '_' + Util.stamp(context) : ''),
-	    handler = obj[eventsKey] && obj[eventsKey][id];
+	var id;
+	if (!fn) {
+		id = type;
+		type = type.split(/\d/)[0];
+	} else {
+		id = type + Util.stamp(fn) + (context ? '_' + Util.stamp(context) : '');
+	}
+	var handler = obj[eventsKey] && obj[eventsKey][id];
 
 	if (!handler) { return this; }
 
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -47,6 +47,10 @@ var eventsKey = '_leaflet_events';
 // @alternative
 // @function off(el: HTMLElement, eventMap: Object, context?: Object): this
 // Removes a set of type/listener pairs, e.g. `{click: onClick, mousemove: onMouseMove}`
+
+// @alternative
+// @function off(el: HTMLElement): this
+// Removes all known event listeners
 export function off(obj, types, fn, context) {
 
 	if (typeof types === 'object') {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
Also this will need an `// @alternative` docstring block to document the restored functionality.

caleblogan:
Thanks for the review! I added a few tests. I think they cover enough, but if not let me know.

src/dom/DomEvent.js
```diff
@@ -122,9 +126,14 @@ function addOne(obj, type, fn, context) {
 }
 
 function removeOne(obj, type, fn, context) {
-
-	var id = type + Util.stamp(fn) + (context ? '_' + Util.stamp(context) : ''),
-	    handler = obj[eventsKey] && obj[eventsKey][id];
+	var id;
+	if (!fn) {
```
johnd0e, line 130: This is error prone.
May be better check `arguments.length`.

src/dom/DomEvent.js
```diff
@@ -122,9 +126,14 @@ function addOne(obj, type, fn, context) {
 }
 
 function removeOne(obj, type, fn, context) {
-
-	var id = type + Util.stamp(fn) + (context ? '_' + Util.stamp(context) : ''),
-	    handler = obj[eventsKey] && obj[eventsKey][id];
+	var id;
+	if (!fn) {
+		id = type;
+		type = type.split(/\d/)[0];
```
johnd0e, line 132: [Util.splitWords](https://leafletjs.com/reference-1.6.0.html#util-splitwords)

johnd0e:
I'm not sure that this api would be really useful.

And it's implementation relies on some redundancy in current code, which is removed here:
https://github.com/Leaflet/Leaflet/pull/7027/commits/04ca20d0cce85cc287f7d6d90686bd54f209cffd

johnd0e:
BTW, some functionality is missing:
```js
// @function off(el: HTMLElement, types: String): this
// Removes all previously added listeners of given types.
```

johnd0e:
My main objection against this PR: it is not related to included tests.
I mean that all 3 added tests pass on current `master`.

So tests may be valuable, as well as doc additions.
But code changes seems have no affect at all.

johnd0e:
Superseded by #7125

PR #6034, opened 2018-01-31
Add option “usedLevelForFractional” for GridLayer.

boundaryfree:


ghybs:
Hi,

Thank you for proposing your first PR here!

Please can you kindly provide some explanation on the reason why you think this PR is useful for Leaflet users?
While we can guess from the code what it does, having context greatly helps in making sure the way the PR is implemented is relevant, and it also helps history.

Additionally, note that Leaflet provides some [tests](https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#running-the-tests) that need to pass before we can merge a PR.
In this case, the lint step triggered some style-related errors.
Please could you kindly fix those?

We would also highly appreciate that you include a unit test related to the new feature that you introduce.

Keep up the good work!

YorkshireChemist:
Just to say that I believe this will be very useful for some Leaflet users. A number of people (e.g., at stackoverflow) mentioned that Leaflet images are blurred at non-fractional zoom. This is particularly important for people using Leaflet to display high-quality non-geographic images (e.g., pyramids via Zoomify or DeepZoom plugins). Clamping zoom at the ceiling solves this problem, and I have seen some already modifying Leaflet core along the lines of the proposed PR.

ghybs:
Hi @YorkshireChemist,

Thank you for your input!

Please feel free to open a proper issue with detailed description of the use case, and in particular linking the Stack Overflow questions you mention.
That would make a very good reference for the usage of this feature.

Please feel free also to improve this PR, in particular to fix the style and potentially affected test(s).

YorkshireChemist:
Hi @ghybs ,

Thanks, will do. It might take a little while but I will do it.

src/layer/tile/GridLayer.js
```diff
@@ -543,8 +547,21 @@ export var GridLayer = Layer.extend({
 		return zoom;
 	},
 
+	_usedZoom: function(zoom) {
+		var options = this.options;
+		if (options.usedLevelForFractional === 'round') {
+			return this._clampZoom(Math.round(zoom));
+		}
+		else if (options.usedLevelForFractional === 'ceil') {
```
ij1: A quick grep shows that Leaflet repo seems to use else ifs/elses on the same line as the previous block's closing brace so you might want to fix it so that it won't prevent taking this in (I need this very same feature myself :-)).

src/layer/tile/GridLayer.js
```diff
@@ -146,7 +146,11 @@ export var GridLayer = Layer.extend({
 
 		// @option keepBuffer: Number = 2
 		// When panning the map, keep this many rows and columns of tiles before unloading them.
-		keepBuffer: 2
+		keepBuffer: 2,
+
+		// @option usedLevelForFractional: String = 'round'|'ceil'
+		// Determin which level (upper or lower) image to use for the fractional level.
```
ij1: Determine

src/layer/tile/GridLayer.js
```diff
@@ -543,8 +547,21 @@ export var GridLayer = Layer.extend({
 		return zoom;
 	},
 
+	_usedZoom: function(zoom) {
```
ij1: Space before opening parenthesis

ij1:
My take on this same feature is in #6505.

PR #6036, opened 2018-01-31
Update plugins.md

dtutic:
Add plugin EdgeScaleBar that draw scale bars along map edges in Web Mercator projection.


IvanSanchez:
Looks good!

Two suggestions, though:
* Add a screenshot to your Readme file. It gives users an impression on how it looks without having to see the demo.
* Make the relevant scale ticks more prominent. There are two kinds of ticks, and it's not clear if the scale applies to the longer ticks or to the shorter ticks.

PR #6039, opened 2018-02-02
add Leaflet Truesize

moklick:


moklick:
thanks we will check that

PR #6041, opened 2018-02-05
Issue #6013

### LINKED ISSUE #6013: DomEvent.off() not working correctly when called with one parameter

mwfontes:
Probably fixed #6013, the problem of passing only the DOM element to the off() method… + added the option of passing only the type and no handler.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -57,11 +57,15 @@ export function off(obj, types, fn, context) {
 		types = Util.splitWords(types);
 
 		for (var i = 0, len = types.length; i < len; i++) {
-			removeOne(obj, types[i], fn, context);
+			// Adds the possibility of only the type being specified.
+			for (var key in obj[eventsKey]) {
+				removeOne(obj, types[i], obj[eventsKey][key], context);
+			}
 		}
 	} else {
 		for (var j in obj[eventsKey]) {
-			removeOne(obj, j, obj[eventsKey][j]);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm going to close this in favour of #6032. @mwfontes, we appreciate your time, but #6032 covers the bug and it's better to not duplicate efforts.

PR #6064, opened 2018-02-21
Do not use an incorrect comparison of strings with an integer.

OriginalSin:
Hi Guys!
Do not use an incorrect comparison of strings with an integer.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -365,7 +365,7 @@ export var GridLayer = Layer.extend({
 		if (zoom === undefined) { return undefined; }
 
 		for (var z in this._levels) {
-			if (this._levels[z].el.children.length || z === zoom) {
+			if (this._levels[z].el.children.length || z == zoom) {
 				this._levels[z].el.style.zIndex = maxZoom - Math.abs(zoom - z);
 				this._onUpdateLevel(z);
 			} else {
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -365,12 +365,13 @@ export var GridLayer = Layer.extend({
 		if (zoom === undefined) { return undefined; }
 
 		for (var z in this._levels) {
-			if (this._levels[z].el.children.length || z == zoom) {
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Hi @OriginalSin , 

Why is that comparison incorrect? Under which circumstances does it provoke an error? When proposing a change, it's always better to explain what problem you're trying to solve, or what use case is improved by it.

Also note that we enforce usage of `===` in the linting rules, so changing this actually breaks the build.

OriginalSin:
Because this code print `false`:
var hash = {};
hash[5] = true;
for( var key in hash) {
  console.log(key === 5)
}

OriginalSin:
Also check the _removeTilesAtZoom, which does not work if the parameter is a string.

IvanSanchez:
But how does this fail when using a leaflet map with a `TileLayer`?

OriginalSin:
I think if you zooming very active.

cherniavskii:
@OriginalSin can you please provide an example, which reproduces the issue solved by this PR?

OriginalSin:
@cherniavskii Hi, I can only suggest inserting `console.log` or `alert` in the code:

_removeTilesAtZoom: function (zoom) {
       for (var key in this._tiles) {
            if (this._tiles [key] .coords.z! == zoom) {
               continue;
            }
            console.log ('is this code execute sometimes ????', this._tiles [key] .coords.z, zoom, this._tiles [key] .coords.z === zoom);
            this._removeTile (key);
       }
},

and try to find such a case when we see the message.
I say that **this function does nothing - never**.

OriginalSin:
Yes, I understand that.
The only consequences (that I think it was possible by this code) is that with active zooming on DOM there are left containers from severa zooms.
default

johnd0e:
> Because this code print `false`:

@OriginalSin
Nice catch!

But this wrong pattern seems common in Leaflet, so we need to fix all such cases.

johnd0e:
Cleaner solution would be using [`Object.keys()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys) instead, but it is available only since IE 9.

mourner:
Superceded by #7123

Issue #5994, opened 2018-01-11
Marker drifts on zoom / marker jumps around on zoom


ghost:
Issue 1. Marker drifts on zoom (does not stay centered on the map, when marker and map is on same coordinates)
Reproduce: go to https://jsfiddle.net/o0yfm5zy/1/
Click button **first**, and then zoom. 

Issue 2. marker jumps around on zoom (only about a pixel or so but it is noticeable, especially when you use flyTo() on retina screen).
Reproduce: on the jsfiddle, click the button again (or re-run) and zoom

I experience the same behaviour on firefox, chromium/opera and ios.

feature7:
Bump!

Any workaround to make the marker and map always on the same exact location?
And to keep the marker at the same exact place through zoom-levels?

daverayment:
Hi. I can confirm the issue. This 'drift' seems to be caused by pixel coordinates repeatedly being used to calculate the centre of the map after an initial pan or pan-and-zoom operation. I think I can see a simple fix in the `setView` function, but I need to confirm with some further testing. This should also take care of the 'wobble' you're seeing on zoom, too.

Interestingly, there's already a test called `zoom-remain-centered.html` in the source, but it doesn't cover the pre-panning case, so this must be why it hasn't previously been detected.

feature7:
@daverayment thanks for looking into this, looking forward to a solution.

ghybs:
Hi,

Probably related to https://github.com/Leaflet/Leaflet/issues/3737 (issue description sounds like dupe https://github.com/Leaflet/Leaflet/issues/5512)

daverayment:
Yes, this looks like #3737 - sorry for not spotting that. My initial simple fix didn't work, but I have another attempt which looks more promising. Unfortunately, it's a fair bit more involved, so I'll write it up for further discussion once I've done some more testing. Sorry about the delay, @feature7

daverayment:
@feature7 could you try the following please?

https://jsfiddle.net/o0yfm5zy/20/

I've had to copy in all the Leaflet code, so it will be a little slow! Hopefully, you'll still be able to check whether it tackles the drifting issue for you, though. Let me now how it goes.

feature7:
Great!
From what I can see, the drifting issue is solved, chromium, firefox, ios, marker remains centred. Tried some random locations as well. Hope to see it in next release.

feature7:
Did some more checking with the new code, @daverayment your fix is solid.

[...]

Issue #3737, opened 2015-08-11
mouse zoom doesn't preserve mouse location


Sumbera:
While I am doing some experimental SVG rendering   I found that  leaflet doesn't preserve same location of mouse pointer when I zoom in/out with mouse. It is easy to reproduce: 

1. zoom deep in
2. move mouse for example to the center of your map, or where is your POI
(or go to  http://leafletjs.com/ and move your mouse on the markup there.)
3. do not move mouse, only scroll far out and then back in
4. you will arrive back somewhere else (shifted) than before

would be great to preserve mouse location when zooming  as for example google maps  or OL 2.x do.

IvanSanchez:
Right now Leaflet uses the map center as a reference. Making that variable (and coincidental with the mouse pointer) might be nice, but I'm afraid that it requires a *lot* of refactoring to do. Big refactors = not something we want this close to 1.0.

Sumbera:
thanks for response,  FYI, in in Map.js I have removed rounding in two conversion functions to get it work correctly:
```javascript
latLngToLayerPoint: function (latlng) { // (LatLng)
    var projectedPoint = this.project(L.latLng(latlng)); // -- SSU removed  ._round();
		return projectedPoint._subtract(this.getPixelOrigin());
	}, 

_getNewPixelOrigin: function (center, zoom) {
  		var viewHalf = this.getSize()._divideBy(2);
		return this.project(center, zoom)._subtract(viewHalf)._add(this._getMapPanePos());// --  SSU removed ._round();
	},
```
please re-consider to fix this for 1.0, if possible.

IvanSanchez:
@Sumbera Wow, I didn't expect such a concise solution (somehow I was expecting floating point arithmetic to be the culprit). Could you please make a pull request for that?

I guess that the `round()` is there for some reason, maybe setting integer CSS offsets in IE8. Probably will need some testing.

mourner:
The round is there because we need to position stuff on round coordinates, otherwise map will get blurry in many browser/device combinations. And yeah, they may break in legacy browsers.

IvanSanchez:
Moving the `round()` to `DomUtil.setTransform` and `DomUtil.setPosition` might just do the trick.

Sumbera:
FYI, I have put sample of what I am doing with SVG here:
http://bl.ocks.org/Sumbera/raw/7e8e57368175a1433791/
[...]

PR #6066, opened 2018-02-21
Ensure map centre is maintained for zoom operations

daverayment:
An issue exists where the rounding operations necessary to produce pixel-aligned map views can result in 'drifting' of the map centre when zooming in and out. For example, if you pan to a specific landmark, zoom all the way out, then zoom in again, the centre will have shifted. (Note that you should use keyboard + and - or the on-screen buttons to zoom into/out of the centre, _not_ your mousewheel or pinch-to-zoom.)

This fix regards the `center` parameter passed to `setView `as the 'last known good' map centroid, storing it in `_lastCenter`, which already exists for a similar purpose. Subsequent `drag`, `move`, `zoom` and `setView` operations set or reset `_lastCenter` and an associated `_movedAfterCentered` field. The `getCenter` function has been changed to return the `_lastCenter` latlng if `_movedAfterCentered` is not set, similar to the existing `_moved()` check (which does the same for if the map hasn't been moved from its origin). This means that `layerPointToLatLng` is not called with the pixel-based centre coords when the geographical centre is available and still accurate, removing the `round` operation and the accumulated error as you zoom in and out.

A test file is included - `debug\map\zoom-remain-centered-after-pan.html` - which does an initial animated pan when initially loaded. Zooming in and out on the new version should be stable, whereas the same operations should show the drift on Leaflet version 1.3.

I haven't been able to produce an automated test. If one's required, I'd appreciate assistance setting one up.

The issues addressed by the fix are: #3737 and #5994.

### DIFF
--- a/debug/map/zoom-remain-centered-after-pan.html
+++ b/debug/map/zoom-remain-centered-after-pan.html
@@ -1,0 +1,47 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script src="../leaflet-include.js"></script>
+</head>
+<body>
+
+
+	<div id="map"></div>
+
+	<script>
+		// Check the 'center' setting of the scroll-wheel, double-click and touch-zoom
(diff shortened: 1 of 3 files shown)


jarrodwhitley:
@daverayment any update on this?

johnd0e:
Should this fix #6712?

PR #6071, opened 2018-02-26
Make L.Class.Include() extends the options of the prototype instead of replacing them

### LINKED ISSUE #6070: L.Class.include() removes inherited options
I noticed a bug when trying out https://github.com/Leaflet/Leaflet.TileLayer.NoGap with the latest Leaflet. 

The [minimally reproducible example](https://plnkr.co/edit/YuFXdvAzObSEIiBQd0OX?p=preview) is:

```
L.TileLayer.include({
  options: {
    MyCustomOption: 'myCustomValue'
  }
});
```

This crashes because during initialization of a `TileLayer`, `this.getPane()` gets called. That call tries to look for the value of `this.options.pane`, which is empty. Setting a debug breakpoint there and looking into `this.options` shows that only the options of the `TileLayer`  exist (options that should be inherited from `GridLayer` do not!!).

This breaks the `NoGap` plugin, and potentially more plugins out there.

Problem appears at least with Leaflet 1.3.1 in Chrome 64 and Firefox 58 on Linux. Looks like a browser-independent bug, so I haven't tested more platforms. I haven't bisected the problem yet.

IvanSanchez:
Fixes #6070. Includes unit test.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -86,7 +86,14 @@ Class.extend = function (props) {
 // @function include(properties: Object): this
 // [Includes a mixin](#class-includes) into the current class.
 Class.include = function (props) {
+	var options;
+	if (props.options) {
+		options = Util.extend({}, this.prototype.options, props.options);
+	}
 	Util.extend(this.prototype, props);
+	if (options) {
+		this.prototype.options = options;
+	}
 	return this;
 };
 
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -84,7 +84,7 @@ describe("Class", function () {
 			expect(a.mixin2).to.be.ok();
 		});
 
-		it("merges options instead of replacing them", function () {
+		it("extend()ing merges options instead of replacing them", function () {
 			var KlassWithOptions1 = L.Class.extend({
 				options: {
 					foo1: 1,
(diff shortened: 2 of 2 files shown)


IvanSanchez:
An alternative would be to modify `L.Util.extend()` so that it performs "deep" extends of nested objects, but I'm not comfortable changing the logic for such a critical piece of code.

src/core/Class.js
```diff
@@ -86,7 +86,14 @@ Class.extend = function (props) {
 // @function include(properties: Object): this
 // [Includes a mixin](#class-includes) into the current class.
 Class.include = function (props) {
+	var options;
+	if (props.options) {
+		options = Util.extend({}, this.prototype.options, props.options);
```
ghybs: Hi,

Why not simply leverage [`mergeOptions`](https://github.com/Leaflet/Leaflet/blob/v1.3.1/src/core/Class.js#L93-L98) to handle `include({options})` special case?
That would avoid breaking the reference with the parent class' options.
Example: https://plnkr.co/edit/Eth0KHPhwETsXUQwD9fq?p=preview

ghybs: BTW should we also treat as special cases [`statics`](https://github.com/Leaflet/Leaflet/blob/v1.3.1/src/core/Class.js#L43-L47) and… [`includes`](https://github.com/Leaflet/Leaflet/blob/v1.3.1/src/core/Class.js#L49-L54)?

IvanSanchez: Yeah, that sounds good. My changes were rather a quick&dirty fix, but I certainly agree with fixing `statics` and `includes` too. Will need to write unit tests for those.

johnd0e: Why do we need to introduce some unobvious 'special cases' where we can achieve the same in clear explicit way?

Falke-Design:
The linked issue was closed by #7756, so I close this too

Issue #5558, opened 2017-06-12
No way to track when L.Icon is removed


kuuup-at-work:
Hi.

When using L.Marker you can listen to the `remove` event. This is not possible with `L.Icon` or `L.DivIcon` so there is not way to run cleanup code when an icon is removed. Extending `L.Marker` doesn't help because you can attach a single `L.Icon` factory to multiple markers. This approach also fails when you use markercluster.

SONUSADDAM8:
hi

perliedman:
Revisiting this issue, I think a quite straight forward solution would be to add methods `destroyIcon` and `destroyShadow` to `Icon`; these correspond to already existing methods `createIcon` and `createShadow`. The default implementation would be to do nothing, but extending `Icon` you could add custom actions here.

We would then need to call `destroyIcon`/`Shadow` in `Marker`'s `_removeIcon` and `_removeShadow`.

williamscs:
Hey @perliedman. I was taking a look at this issue... it looks like the _option member of Marker.js isn't actually an Icon object. It's the actual DOM object.

I see that in `setIcon` that we use this.options.icon to assign the actual Icon object. We don't seem to clear that in the `_removeIcon` or `_removeShadow` methods. I think that value should probably be set to null when `_removeIcon` is called, if nothing else, for clarity.

I intend to use this.options.icon to call the `destroyIcon` method. I'm not sure how to continue with the _shadow member, though. I'm probably just not going to create the `destroyShadow` method

perliedman:
@williamscs what you're describing is intentional: the `Icon` instance does not represent a single icon, it is an instance shared by any marker using that icon. The `Icon` instance acts as a factory for DOM elements, so the marker calls `createIcon`/`createShadow` to make new DOM elements - but it's still just a single `Icon` instance.

So, since `createIcon` returns a DOM element, not an `Icon` instance, the `destroyIcon`/`destroyShadow` methods should accept the same DOM elements as arguments, and clean them up.

If you clear `this.options.icon`, it's no longer possible to re-add a marker to the map after removing it, since its icon is now cleared, which isn't what we want.

johnd0e:
`L.Icon` is removed only when it gets unreferenced and [garbage-collected](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#garbage_collection).

And there is no possibility to track that in js code, so you should consider some alternative approach in your own code.

PR #6077, opened 2018-03-03
#5558 feat(Icon) - Add extendable destroyIcon method.

williamscs:


williamscs:
Will update documentation. Getting environment set up for working with jekyll.

williamscs:
Howdy, @perliedman. Sorry for the delay, but I've pushed up some more changes. Hopefully it's more in line with what you're talking about. The destroyIcon and destroyShadow methods now are passed the HTMLElement instances.

src/layer/marker/Marker.js
```diff
@@ -269,6 +270,7 @@ export var Marker = Layer.extend({
 
 	_removeShadow: function () {
 		if (this._shadow) {
+			this.options.icon.destroyShadow(this._icon);
```
perliedman: Should be `this._shadow`, right?

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -330,4 +330,18 @@ describe("Marker", function () {
...
+	describe('removal', function () {
+		it('should call destroyIcon', function () {
+			var marker = new L.Marker([0, 0], {icon: icon1});
+			var destroyIconSpy = sinon.spy(icon1, 'destroyIcon');
+			var destroyShadowSpy = sinon.spy(icon1, 'destroyShadow');
+			map.addLayer(marker);
+
+			map.removeLayer(marker);
+
+			expect(destroyIconSpy.calledOnce).to.equal(true);
```
perliedman: Given the issue I mentioned below, it would be nice to not only check that the methods are called, but also that they are called with the expected arguments (the icon element for `destroyIcon` and the shadow for `destroyShadow`).

williamscs:
Sorry, this is super old and stale. Closing.

PR #6078, opened 2018-03-04
new plugin added (geogrid.js)

mocnik-science:


IvanSanchez:
LGTM.

I could nitpick about a few implementation details (e.g. why the absolute need for WebGL?), but overall looks nice.

mocnik-science:
Thanks!

Regarding your question: the plugin can actually also use SVG to render the grid (see option `renderer`), but performance is much worse than in case of WebGL.

IvanSanchez:
I guess as much, but in that case I'd push for a full-fledged WebGL `L.Renderer` for Leaflet. It comes down to development priorities and available development time, I guess.

PR #6090, opened 2018-03-12
Update VideoOverlay.js

zhuang-hao-ming:
The example comments are wrong which confuse reader.


IvanSanchez:
Thanks for noticing this! :-)

PR #6096, opened 2018-03-16
Change 'Form' to 'div'

dczunigap:
**Control Layers** create a form with inputs dynamically for panel of layers. So, on this panel I see checkbox and labels for hide or show layers on map.

But, when I use **jQuery Validate**, the plugin check this form and launch error about undefined rules.
then jQuery Validate plugin catch it and search rules for validate inputs of form.
error_jquery_validate
form_dynamically

**Change form to div no affect behavior of Control Layers.**

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -76,7 +76,7 @@ L.Control.Layers = L.Control.extend({
 			L.DomEvent.on(container, 'click', L.DomEvent.stopPropagation);
 		}
 
-		var form = this._form = L.DomUtil.create('form', className + '-list');
+		var form = this._form = L.DomUtil.create('div', className + '-list');
 
 		if (this.options.collapsed) {
 			if (!L.Browser.android) {


ghybs:
Hi,

Thank you for submitting your first PR here!

If I understand correctly, your report a compatibility issue with [jQuery validation plugin](https://jqueryvalidation.org/).

Filing an issue first could have been more appropriate, so that we can make sure maintainers have enough information about the situation you are trying to improve, and therefore can in a next step properly evaluate your PR and if it is relevantly implemented.

Your proposed PR is to modify Leaflet so that when you use it on your site with the above jQuery plugin, the latter does not complain about missing custom validation rules?

With this modification, you make Leaflet build some `<input>`'s outside a `<form>`. While this is still valid as per HTML spec, I am not sure what other implications this may have, e.g. regarding accessibility?

Furthermore, I am unable to reproduce the error message you show in your screenshot. I may not use jQuery validation properly, so please provide a full reproducible example:

- Merely loading jQuery and jQuery validation plugin does not seem to do anything, therefore not causing any error due to Leaflet's `<form>`.
- Explicitly catching all page `<form>`'s and trying to `.validate()` them, even with `debug: true` option, does not trigger any error: https://plnkr.co/edit/3W9P0vCTtwfjhGi7Us7P?p=preview
You can check that Leaflet's `<form>` has been properly taken into account because it now has the `novalidate` attribute set by jQuery validation.

Please can you kindly provide more details about your situation, so that maintainers can properly evaluate how relevant is your PR?

IvanSanchez:
I'm strongly inclined to reject this PR, for two reasons:

- This targets the `0.7` branch, and we don't really do 0.7 maintenance anymore.
- The jQuery validate plugin allows the user to validate specific forms - thus, if it's validating the layers control, I'm inclined to think about an implementation error.

ghybs:
Hi @dczunigap,

If you still have your problem and you are sure the problem is due to Leaflet rather than how your app is configured, please open an issue with relevant details and a reproduction example.

Also we will likely not be able to provide support for Leaflet 0.x.

dczunigap:
Chales! @ghybs you're right... jquery validate plugin no produce conflict when I configure it in javascript file. The conflict is show when use an angular directive which implements jquer validate plugin in my project, as you say and @IvanSanchez.

So...  I think I owe you an apology. Thanks for the feedback and give me a chance for try contribute

PR #6100, opened 2018-03-20
Add zoomOffset option to fitBounds

sds-dubois:
Add an option to offset the zoom level after `fitBounds` - I feel like it's common use case to want to fit the bounds and then zoom out.

I've had some trouble calling `map.zoomOut()` right after `fitBounds`, even when disabling the animation (eg, tiles not loaded, marker positions are wrong...) and thought it would just be better to add this as an option to `fitBounds`.

Let me know what you think

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -257,6 +257,9 @@ export var Map = Evented.extend({
 
 		    zoom = this.getBoundsZoom(bounds, false, paddingTL.add(paddingBR));
 
+		if (typeof options.zoomOffset === 'number') {
+			zoom += options.zoomOffset;
+		}
 		zoom = (typeof options.maxZoom === 'number') ? Math.min(options.maxZoom, zoom) : zoom;
 
 		if (zoom === Infinity) {
--- a/src/map/Map.methodOptions.leafdoc
+++ b/src/map/Map.methodOptions.leafdoc
@@ -105,4 +105,7 @@ Equivalent of setting both top left and bottom right padding to the same value.
 @option maxZoom: Number = null
 The maximum possible zoom to use.
 
+@option zoomOffset: Number = null
+The zoom level will be offset by this value, eg. use -1 to zoom out after
+setting the view to fit bounds.
 


ghybs:
Hi,

Thank you for submitting your first PR here!

If I understand correctly, you would like your map to `fitBounds`, but with 1 zoom level lower than the current computation result?

I can understand that you would like an extra option in Leaflet core to fit your use case, but it seems to me that you could already achieve something similar with the current API, using [`map.getBoundsZoom(bounds)`](http://leafletjs.com/reference-1.3.0.html#map-getboundszoom):
```javascript
// Compute the appropriate zoom level, without having the map actually change its view.
var fitZoom = map.getBoundsZoom(bounds);

map.setView(bounds.getCenter(), fitZoom - 1);
```
Example: https://plnkr.co/edit/W3dJjQBeP2RFLRTQLlRJ?p=preview

sds-dubois:
> If I understand correctly, you would like your map to fitBounds, but with 1 zoom level lower than the current computation result?

Exactly, although it could be anything and not just 1 zoom level lower.

Thanks for pointing how to achieve the same result with the current API - but my point with this PR was more to suggest adding this option, just like there's the `maxZoom` option. I thought it could be useful to have it in `fitBounds` since it would also work with arrays of latlongs, padding, maximum zoom etc

But feel free to close it if you don't think that's a common enough use case.

ghybs:
Hi,

Sorry for not getting back to you earlier.
I will merge this if you can add a test for this new feature.

sds-dubois:
Hi @ghybs, I just added 2 tests:
- the first one is a simple test to verify the behavior of `zoomOffset` (cf the test above "Snaps zoom level to integer by default" which demonstrates zoom would be 2 otherwise)
- a second one to test that `zoomOffset` does not override `maxZoom` if set

Btw there's no option `minZoom`, so let me know if you think it would be useful and I can just add it along with some tests.

johnd0e:
> I feel like it's common use case to want to fit the bounds and then zoom out.

I'm not sure that it is really common case.
And if you really need this - why not use [`pad`](https://leafletjs.com/reference-1.6.0.html#latlngbounds-pad) method?

> I've had some trouble calling `map.zoomOut()` right after `fitBounds`, even when disabling the animation (eg, tiles not loaded, marker positions are wrong...) and thought it would just be better to add this as an option to `fitBounds`.

For me such option would look a bit weird. May be better address such issue in more direct way (I mean fix `zoomOut` method).

PR #6117, opened 2018-04-04
Fix for memory leak in canvas renderer

aj8k:
1. Use `canvas` renderer
2. Add polylines (40K for example)
3. Remove them and add a bunch of other polylines (once more 40K)

AR: Page memory usage almost doubled

The root of the bug is in `_drawnLayers` object in canvas renderer. References to layers removed everywhere except this place.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -161,6 +161,8 @@ export var Canvas = Renderer.extend({
 			this._drawFirst = next;
 		}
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
@uygardonduran That seems to be a leftover from https://github.com/Leaflet/Leaflet/commit/4c484462dc4d646cb43adc0585401d28d878330d. Why don't you send a PR removing all references to `_drawnLayers`, and see if everything works fine after that?

PR #6118, opened 2018-04-05
Add fractional zoom support for tile servers

lapo-luchini:
Fractional zoom tiles are a bit blurry (even trying changing `Math.round` with `Math.ceil` to always use a higher resolution one instead of the closest one), but since our internal tile server always supported fractional zoom levels too I've added `zoomSnap` support to tiles too, in order to always show the user 1:1 rendered tiles.

Option name (and part of the description) is heavily inspired to the one in `Map`, feel free to change that as desired.

(this was a part of #1309 that was not implemented by the more comprehensive change that closed that bug)


markrian:
This also affects `GridLayer` subclasses that want to be able to render fractional zoom levels.

PR #6119, opened 2018-04-05
Apply #5498 for Polygons too

BakuCity:
Extend the fix on #5498 to handle Polygons with empty arrays.


sowelie:
Is there any plan for when this will be part of a release? I am using the canvas renderer for polygons and getting JS errors due to this fix not being in 1.3.1.

PR #6124, opened 2018-04-06
Add locks and logs to .gitignore.

denysdovhan:
When new contributors run `npm install` or `yarn` in this project, they get corresponding lockfiles. If they get an error, they get `*.log` files.  Since Leaflet is a _library_, there's no reason to commit lockfiles to the repo. However, these files are not ignored, so newbie is gonna get something like this:

screen shot 2018-04-06 at 20 59 25

This PR ignores `*.log`s and `*.lock`s, so newcomers cannot commit them.

P.S: Let me know if you prefer to use `.npmrc`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ bower.json
 component.json
 debug/local/
 Gemfile.lock
+*lock*
+*.log


denysdovhan:
No problem. @mourner, see #6127.

bozdoz:
this makes no sense to me. Why wouldn't you want the lock file in source control so that the package managers install the correct versions each time?  I believe yarn does this by default, and npm can do it with `npm ci`.  Removing the lock file means some day down the road nested version dependencies *could* break the build.

Issue #3870, opened 2015-09-25
Convert indentation from smart tabs to 4-space


mourner:
I've been a fan of smart tabs (tabs for indentation, spaces for alignment) and persevered with it through many years, but have to admit that I regret this choice now and 4 spaces provide a much less painful experience for both maintainers and contributors. I now use 4 spaces in all my other projects (smooth sailing), and it would be great to convert Leaflet as well eventually.

These are the big problems I see with smart tabs:

- they're very hard to enforce on the linting level; there are all sorts of edge cases and the approach is not used often enough to drive fixing related issues and bringing related features on projects like ESLint
- most contributors are unfamiliar with the approach so I have to write "please fix whitespace" comments in PRs over and over again; it will get easier with https://github.com/Leaflet/Leaflet/commit/301473ad4d83ec14254464e2f598e9827a904927 but that's still a hack
- all-spaces indentation is much better supported in editors unlike mixed tabs and spaces, which is more hassle to write
- smart tabs bring complexity to whitespace management; all-spaces are barebones simple

These are the issues that prevent me from doing a repo-wide conversion:

- it'll **bork all the open PRs** (authors will have to painfully rebase and solve lots of conflicts); does anyone know a solution to this? I know there's at least an option to ignore whitespace when showing diffs, but is there an easy way to merge code with different whitespace in git?
- it'll screw `git blame` so that you'll have to run it again on commits before the conversion, but I guess that's a necessary evil

Should we go ahead with this and deal with consequences, or just leave this alone and continue to deal with whitespace annoyances?

cc @IvanSanchez @yohanboniface @danzel @jfirebaugh

yohanboniface:
I'd support this and think the 1.0 is the good opportunity to do so. Will be painful, but necessary for a long term view imho.

IvanSanchez:
Apparently it is possible to rewrite history and replace all tabs with spaces without borking `blame`, by using https://git-scm.com/docs/git-filter-branch and a `sed` replacement.

mourner:
@IvanSanchez nah, that's just too scary — it would change all commit refs and break everything for sure. Not an option.

> WARNING! The rewritten history will have different object names for all the objects and will not converge with the original branch. You will not be able to easily push and distribute the rewritten branch on top of the original branch. Please do not use this command if you do not know the full implications, and avoid using it anyway,

danzel:
Breaking blame is gonna suck, but it is easy to work around when necessary.
[...]

PR #6125, opened 2018-04-06
Add Editorconfig

denysdovhan:
See: http://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# EditorConfig
+# http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.js]
+indent_style = tab
+tab_width = 4
+
+[{package.json,.travis.yml}]
+indent_style = space
+indent_size = 2


denysdovhan:
Would you prefer to use this set of settings?

```
# http://editorconfig.org
root = true

[*]
indent_style = tab
indent_size = 4
end_of_line = crlf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

Falke-Design:
Because there was no request for this since 2018 anymore, I suggest to close this

.editorconfig
```diff
@@ -0,0 +1,20 @@
+# EditorConfig
+# http://EditorConfig.org
```
Malvoz, line 2:
```suggestion
# https://EditorConfig.org
```

.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.js]
+indent_style = tab
+tab_width = 4
+
+[{package.json,.travis.yml}]
+indent_style = space
+indent_size = 2
```
Malvoz, line 20: What's the basis for these settings? They'd have to be kept in sync with the ESLint config right? (https://github.com/Leaflet/Leaflet/issues/8176 / https://github.com/Leaflet/Leaflet/issues/3870)

jonkoops (changes requested):
I don't see the value of adding a config that can become out of sync with our existing linting tools (which are checked by CI). Checking this config is also in no way enforced by any tooling on CI at the moment.

Malvoz:
Closing per https://github.com/Leaflet/Leaflet/pull/6125#pullrequestreview-951256183, https://github.com/Leaflet/Leaflet/pull/6125#issuecomment-968363321 and https://github.com/Leaflet/Leaflet/pull/2347#issuecomment-32362778.

PR #6127, opened 2018-04-07
Ignore package-lock.json and yarn.lock explicitly

denysdovhan:
https://github.com/Leaflet/Leaflet/pull/6124#issuecomment-379366811


ghybs:
Thank you!

PR #6128, opened 2018-04-09
Fix setStyle method for Circle class

martinRenou:
The Circle class inherits from CircleMarker class, which has a `setStyle` method which sets the radius. There was no equivalent in Circle class (the `setStyle` method was the one from the Path class), so we couldn't set the Circle radius with the `setStyle` method

### DIFF
--- a/src/layer/vector/Circle.js
+++ b/src/layer/vector/Circle.js
@@ -1,5 +1,4 @@
 import {CircleMarker} from './CircleMarker';
-import {Path} from './Path';
 import * as Util from '../../core/Util';
 import {toLatLng} from '../../geo/LatLng';
 import {LatLngBounds} from '../../geo/LatLngBounds';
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Why is this needed, when there is already a `setRadius()` method?

martinRenou:
I'm currently working on https://github.com/jupyter-widgets/ipyleaflet, we use python models on the backend for defining Leaflet widgets on the frontend. Every time the model changes on the backend, we update the frontend with a simple `setStyle` call passing the options defined in the model. This works fine for each attribute of the Circle class (those defined in the Path class: color, opacity, fillColor...), except for the radius.

SylvainCorlay:
`setStyle` appears to take the options dictionary and `radius` is listed in the options for the circle marker.

IvanSanchez:
BTW, this could be a much cleaner change if you just replace `setStyle: Path.prototype.setStyle,` with `setStyle: CircleMarker.prototype.setStyle,`.

IvanSanchez:
On second thought, it seems that the behaviour of `setStyle` is an oversight dating back to https://github.com/Leaflet/Leaflet/commit/508a75f7a836a0a5bd1e9472cb09434183046860. I'm partial to accepting this, but I'll wait for other maintainers to chip in.

Maybe we need to add some docstrings to make it clearer that `setStyle` on both `Circle` and `CircleMarker` do set the radius too.

mourner:
I don't think this is an oversight. Radius not being a part of Circle "style" is intentional, because unlike CircleMarker, it'a physical property of the circle, just like geographical location. It's not a visual thing.

johnd0e:
@mourner
Should we close this?

PR #6129, opened 2018-04-09
Add missing layerType to LayerEvent

beschoenen:
Adds missing documentation for `layerType` on `LayerEvent`.


IvanSanchez:
`LayerEvent`s in core Leaflet do not have a `layerType` property. In fact, `layerType` appears a total of zero times in the core Leaflet code.

Perhaps you meant to fix the documentation for the Leaflet.Draw plugin?

beschoenen:
Yes, my mistake.

Issue #6099, opened 2018-03-20
Markes leaves their position on two-finger zoom on IOS


ikhudo:
Leaflet 1.3.1
IOS 11,
Browser: Safari

it is only reproducible with `zoomAnimation: false` parameter

![leaflet 2](https://user-images.githubusercontent.com/35076442/37661426-11a4ebde-2c55-11e8-8df2-dfca3f50882a.gif)

[JSFiddle](http://jsfiddle.net/igorko82/4v6Luacz/11/)

Run it on IOS device and try to zoom in/out with pinch gesture.

designgears:
This seems to be an issue with canvas, even with zoomAnimation: true while panning if you touch a second finger to the screen to perform a pinch to zoom the markers do the same thing as in your gif.

manubb:
Apparently there are two different bugs reported in this issue. PR #6126 only solves one of them.
The line:
```js
setPosition(this._mapPane, new Point(0, 0));
```
in `Map.prototype._resetView` causes problem with `zoomAnimation: false`. I will open a new PR for this.

IvanSanchez:
This seems to be a tricky scenario. @manubb How do you feel about creating a reproducible, *automatic* example of this behaviour using `prosthetic-hand`, like at https://github.com/Leaflet/Leaflet/blob/master/debug/tests/touch-shake.html ? That should make a bit easier to track down the pinch-zoom code that relies on the `zoomAnimation` option; I'm curious about why it impacts pinch-zoom at all.

mourner:
Please check out the latest beta https://github.com/Leaflet/Leaflet/releases/tag/v1.8.0-beta.3 and see if everything's resolved?

PR #6133, opened 2018-04-10
Fix for bad zoom center when zoomAnimation === false

manubb:
I'm not 100% sure this is really safe but this fixes a part of issue #6099.
(Step to reproduce from a map with `{zoomAnimation:false}` with a scrollwheel mouse:
-press left button and perform some panning
-use the scrollwheel to change zoom while keeping left mouse button pressed

Then the new map center is not the expected map center.

This can also be seen with a touch device with a panning followed by a pinch zoom but it is easier to reproduce with a mouse.)

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1131,8 +1131,6 @@ export var Map = Evented.extend({
 
 	// @section Map state change events
 	_resetView: function (center, zoom) {
-		DomUtil.setPosition(this._mapPane, new Point(0, 0));
-
 		var loading = !this._loaded;
 		this._loaded = true;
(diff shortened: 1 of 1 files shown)


johnd0e:
Just for record: this code was there from the very beginning: https://github.com/Leaflet/Leaflet/blame/00717ab05b6046397db7d249aeea732af6a2fa12/src/Map.js#L137

The last change was here: https://github.com/Leaflet/Leaflet/commit/d13aaac837791849f219d8d857124dd767a985e8#diff-607b516b5121104eef3007cb5727e69cR526

Issue #6139, opened 2018-04-23
In geolocation event handler view bounds are calculated incorectly


wladich:
In Map._handleGeolocationResponse value of Coordinates.accuracy is interpreted as diameter of "confidence circle", but it should be radius.
Error is at line https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L659:
`
bounds = latlng.toBounds(pos.coords.accuracy),
`
LatLng.toBounds according to manual:
> Returns a new LatLngBounds object in which each boundary is sizeInMeters/2 meters apart from the LatLng.

But after reading the definition of Coordinates.accuracy  from MDN https://developer.mozilla.org/en-US/docs/Web/API/Coordinates/accuracy
> The Coordinates.accuracy read-only property is a strictly positive double representing the accuracy, with a 95% confidence level, of the Coordinates.latitude and Coordinates.longitude properties expressed in meters.

and accuracy term in wikipedia https://en.wikipedia.org/wiki/Accuracy_and_precision I'm sure Coordinates.accuracy should be the radius of "confidence circle"

I attached demo illustrating wrong behavior of Map._handleGeolocationResponse:
https://jsfiddle.net/4zjLzqv1/

Here blue circle radius equals `Coordinates.accuracy` and red circle radius equals `Coordinates.accuracy / 2`.
Blue circle is larger then view bounds and can be seen only when zoomed out one step.

Conclusion:
Line https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L659 should be
`
bounds = latlng.toBounds(pos.coords.accuracy * 2),
`

Error was introduced in first commit of geolocation support:
https://github.com/Leaflet/Leaflet/blame/e87e010e7ddb4acec5088362a43f5e9227e0a9eb/src/map/ext/Map.Geolocation.js#L63

IvanSanchez:
Sounds like an easy fix. Care to send a pull request?

spmsupun:
+1

wladich:
#6140

ghybs:
Fixed by #6140

PR #6140, opened 2018-04-23
Fixed view bounds calculation in geolocation event handler (#6139)

wladich:


perliedman:
Nice! Thanks for spotting this nasty mistake as well as fixing it!

PR #6154, opened 2018-05-04
Make _detectIconPath play nice with fingerprinting

mb21:
This way `_detectIconPath` works correctly when the CSS was changed to e.g. `url(/foo/marker-icon-908e25f-hash.png)`, as is the case when using [Rails asset pipeline fingerprinting](http://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark).

Test:

    > var path = "url(foo/bar/marker-icon-123asdf.png)"
    > path.replace(/^url\(["']?/, '').replace(/marker-icon[^/]*\.png["']?\)$/, '');
    "foo/bar/"

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -52,7 +52,7 @@ export var IconDefault = Icon.extend({
 		if (path === null || path.indexOf('url') !== 0) {
 			path = '';
 		} else {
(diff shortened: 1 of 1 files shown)


mb21:
Just saw that this is an ad-hoc-fix for https://github.com/Leaflet/Leaflet/issues/4968, you should probably merge https://github.com/Leaflet/Leaflet/pull/5771 instead.

Issue #5392, opened 2017-03-18
Leflet map in iframe: page scrolls down to bottom on first click on map


lukashanssondn:
Hi
I have noticed an annoying problem when I embed  a leflet into a iframe. On my mobile iphone 5 the page scrolls down to the bottom on the first click on the map.

The thing is easily reproducible but only appers in mobile phones, web works fine.

surf to this page with your phone and test 
http://ext.dn.se/qs/widgets/test/iframeTest2.html

Or make a testpage
Here is the code

Thanks for a fantasic library anyway!!

<!doctype html>
<html>
<head>
<meta charset="UTF-8">
<meta charset="UTF-8">
<meta charset="UTF-8">
<meta content="initial-scale=1.0, width=device-width,height=device-height,minimum-scale=1.0,maximum-scale=1.0,user-scalable=no" name="viewport">
<meta content="telephone=no" name="format-detection">
<meta content="yes" name="apple-mobile-web-app-capable">
<meta content="black-translucent" name="apple-mobile-web-app-status-bar-style">
<meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
<meta content="IE=edge,chrome=1" http-equiv="X-UA-Compatible">
<title>test</title>
</head>
<body>
<div style="width:100%">
some text her to fill the page
 Lorem ipsum dolor sit amet, nibh ac consectetuer wisi cras dapibus, commodo odio commodo felis ac quis, lacus sed corrupti pede, sed id. Lacus non ipsum, id eu est porttitor, risus vitae quis scelerisque nec, cras lacus nam sed in sed velit. Est nulla praesent, neque nulla ut eget wisi montes rutrum, mus praesent sed. Dapibus ac morbi aliquam. Dapibus mi blandit at diam, id ut imperdiet nec, at parturient, dolor mauris integer fringilla sapien nunc, sem tellus nisl vel turpis aliquam ac. Magnis lorem dui, sem risus sapien.
</div>

<iframe id="test"  style=" height:400px; width:100%;" scrolling="no"  frameborder="0" src="http://leafletjs.com/examples/mobile/example.html"></iframe>

<div style="width:100%">
some text her to fill the page
[...]

PR #6157, opened 2018-05-09
get rid of focus for map container as it forces iframe scrolling in iOS

kazhe:
Dealing with this bug: https://github.com/Leaflet/Leaflet/issues/5392
The issue here is that due to a bug in Webkit on iOS devices (both on Chrome and Safari) a focus() event for an element in an item that is already in viewport scrolls the parent to the bottom, which certainly is not a desired behaviour.

### DIFF
--- a/src/map/handler/Map.Keyboard.js
+++ b/src/map/handler/Map.Keyboard.js
@@ -82,8 +82,6 @@ export var Keyboard = Handler.extend({
 		    top = body.scrollTop || docEl.scrollTop,
 		    left = body.scrollLeft || docEl.scrollLeft;
 
-		this._map._container.focus();
-
 		window.scrollTo(left, top);
 	},
 


IvanSanchez:
@kazhe Thanks for the pull req! :-)

However, I'm afraid that this might break existing functionality. With your changes, what happens when a user clicks outside of a map, then inside of a map, then uses the keyboard controls?? 

Please note that the keyboard controls need the map to have the focus.

I'm afraid that your approach might have undesired side effects. Perhaps the solution is not to remove the `focus` event, but rather preventing the browser from scrolling when the map is focused.

kazhe:
actually - this can now be fixed by setting keyboard: false in map properties if it is called on a touch device (it did not work with an older version of leaflet when the keyboard binding was performed regardless of whether keyboard was true or false; considering that keyboard navigation is not a top priority on an iPhone, this seems to be an adequate solution)

PR #6183, opened 2018-06-09
Added rotate check for setTransform function.

Boroshno:
Since Leaflet doesn't support rotation of marker I doing it manualy by changing transform property of markers. But in case of zoom leaflet rewrites transform property of objects. And it doesn`t check if it has rotate in it. So it removes it by adding only translate and scale to new transform value. 

I`ve added regex to check if it has rotate in it to add it to the new value. In this case rotate wouldn`t be losed.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -208,11 +208,16 @@ export function testProp(props) {
 export function setTransform(el, offset, scale) {
 	var pos = offset || new Point(0, 0);
 
+	//Here we need to get rotate value, if it exists, not rewrite it by new tranform value
+	var rotateRegexp = /(rotate)(\(.*(?:deg\)))/g; //regex to match rotateZ(...deg)
+	var rotate = rotateRegexp.exec(el.style[L.DomUtil.TRANSFORM]);
+
 	el.style[TRANSFORM] =
 		(Browser.ie3d ?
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I just don't like this approach. Running a regexp on **every** layer **every** time it's repositioned seems like it would spend a non-trivial amount of CPU time.

I would prefer something along the lines of:

```
export function setTransform(el, offset, scale, rotation) {
```

And then something like:

```
L.Marker.Rotated = L.Marker.extend({
  _setPos: function(pos) {
    /* possibly recalculate position based on the centre of rotation before this */
    L.DomUtil.setTransform(this._icon, pos, undefined, this.options.rotation);
  }
})
```

This increases the API surface (which is bad), but removes any use of regexps around (which is good).

Boroshno:
I think you are right. As rotation maybe not needed for **every** layer, but only for points. Why I am not like solution with addidng one more API is because of scale. It is optional too. So in our case we need to overload the function twice. But I agree with you about regEx problem.

If you have some more ideas or decisions how to solve the problem of rotation - I will be very gratefull!

Falke-Design:
I closing this because regex is not performant and the other solution is coverd in #7140

johnd0e:
Reopened as draft for sake of discussion.

PR #6200, opened 2018-06-22
Canvas: allow reset of dashArray and allow array-type as option

McBen:
A style update was not able to clear a dashed line.
Additional it allows to use an array for 'dashArray' in canvas mode too. 

The name 'dashArray' is misleading. I seen many developers assigning an array to it.
Since the SVG-Render doesn't care if it's an array or string (undocumented feature of 'SetAttribute'?) nobody even realize that this is wrong.


mourner:
Thanks!

> undocumented feature of 'SetAttribute'?

This is probably due to the fact that arrays are stringified with comma by default.

PR #6201, opened 2018-06-22
Test(GridLayer): add 2 new tests for zoom-in/-out in graphical browsers

ghybs:
This PR adds 2 extra tests for GridLayer zoom-in and zoom-out animation on "graphical browsers" (i.e. not PhantomJS).
They are similar to these already existing 2 tests:
- GridLayer number of 256px tiles loaded in synchronous animated grid @800x600px Loads 32, unloads 16 tiles zooming in 10-11
- GridLayer number of 256px tiles loaded in synchronous animated grid @800x600px Loads 32, unloads 16 tiles zooming out 11-10

But instead of using `clock.tick` to accelerate the tests and have to use rAF at the appropriate moments, these tests let the animation execute on its own.
Therefore they should be less prone to breaking when the animation is internally modified.

The drawback is that they take slightly longer time to pass (a few hundreds ms).
But there are already much longer tests (a few seconds for some of them).

Hopefully this PR will not conflict once the previous one gets merged (https://github.com/Leaflet/Leaflet/pull/6199)

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -773,6 +773,62 @@ describe('GridLayer', function () {
 			clock.tick(250);
 		});
 
+		// NOTE: This test has different behaviour in PhantomJS and graphical
+		// browsers due to CSS animations!
+		// Similar to previous test but without using sinon.useFakeTimers, in order to avoid having to remain too close to how
+		// the tiles animation work internally, but remain at a more external level.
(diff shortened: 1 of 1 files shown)


mourner:
@ghybs just noticed the tests are failing (maybe intermittently) in master after this commit: https://travis-ci.org/Leaflet/Leaflet/builds/396715938?utm_source=github_status&utm_medium=notification

We should either revert or make them stable.

Issue #6195, opened 2018-06-20
retina not acurate


JLuc:
## How to reproduce

- Leaflet version I'm using: 1.3.1 but also on http://leaflet-extras.github.io/leaflet-providers/preview/
- Browser (with version) I'm using: Firefox 60.0.2 or Chromium  66.0.3359.181 
- OS/Platform (with version) I'm using: Ubuntu 17.10

## What behaviour I'm expecting and which behaviour I'm seeing

Documentation says "detectRetina : If true and user is on a retina display, it will request four tiles of half the specified size and a bigger zoom level in place of one to utilize the high resolution."

But it's not used to compute `L.Browser.retina` (true for browsers on a high-resolution "retina" screen.)
`var retina = (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1;`

On my screen, `window.devicePixelRatio` is 2 when browser display is zoomed to 200%
so `L.Browser.retina` is true as soon as display zoom is > 100% and false when browser zoom is < 100%

That is not what "retina" is meant to be.

## Minimal example reproducing the issue

- Open browser, zoom display to 150% (this is browser display, not map zoom)
- Browse to http://leaflet-extras.github.io/leaflet-providers/preview/
- Choose `Stamen toner` or another provider with retina tiles
- Open developer 's javascript console
- See the value for L.Browser.retina
- See the value for window.devicePixelRatio
- Open "Network" tab and see the uploaded tiles : they are @2x version
- Unzoom browser (dont unzoom the map, but the browser display, using its menu)
- See values...

Window has to be refreshed CTRL+F5 after browser display zoom has changed; so retina value is computed anew

ghybs:
Hi,

You might confuse the Tile Layer [`detectRetina`](https://leafletjs.com/reference-1.3.0.html#tilelayer-detectretina) option and the `{r}` in Tile Layer [urlTemplate](https://leafletjs.com/reference-1.3.0.html#tilelayer-url-template) argument, which adds `@2x` when `L.Browser.retina` flag is `true`, as you have noticed.

While these 2 features target "retina" / high DPI devices (or when a user has increased their browser native zoom display as you noticed), they provide 2 different mechanisms (which can be combined, even though it might not be desirable):
[...]

PR #6203, opened 2018-06-22
more acurate 'Browser.retina' explanation

JLuc:
retina is true for both retina screens and for any screens when browser's display is zoomed. 
See https://github.com/Leaflet/Leaflet/issues/6195#issuecomment-399510164

### DIFF
--- a/docs/reference-1.3.0.html
+++ b/docs/reference-1.3.0.html
@@ -18599,7 +18599,7 @@ touch events.</td>
 	<tr id='browser-retina'>
 		<td><code><b>retina</b>
 		<td><code>Boolean</code></td>
-		<td><code>true</code> for browsers on a high-resolution &quot;retina&quot; screen.</td>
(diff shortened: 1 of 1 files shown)


ghybs:
Hi,

Thank you for your PR!

I like the improved documentation.
Please note that the HTML file is automatically generated, so you need to also change the docstring in `src` so that your text is rendered in next version.

In this case it would be in `src/core/Browser.js`:
https://github.com/Leaflet/Leaflet/blob/v1.3.1/src/core/Browser.js#L112-L114

JLuc:
Done 

(but i created a new https://github.com/Leaflet/Leaflet/pull/6206 PR when instead i wanted to merge with the previous https://github.com/Leaflet/Leaflet/pull/6203)

ghybs:
Ok with PR https://github.com/Leaflet/Leaflet/pull/6206

Issue #5078, opened 2016-11-03
Grid lines showing around tiles


ttback:
## How to reproduce

- Leaflet version I'm using: 1.0.1
- Browser (with version) I'm using: Chrome
- OS/Platform (with version) I'm using: OSX 10
Steps to reproduce:
Open http://timesmachine.nytimes.com/timesmachine/2001/09/12/issue.html and you'd see grid lines.

## What behaviour I'm expecting and which behaviour I'm seeing
We are not doing any fancy things with the map rendering but the default API calls through tilelayer() etc. It has been working in production for 2 years, but after upgrading to Leaflet 1.0.1, it's start showing lines on Desktop version only.

I tried `.leaflet-tile { border: 0px; }` but it can't get rid of the lines

ttback:
I am not sure if this has anything to do with https://github.com/Leaflet/Leaflet/issues/1066

perliedman:
Hi, this is a duplicate of #3575. If you need a fix ASAP, you can use this plugin: https://github.com/Leaflet/Leaflet.TileLayer.NoGap, or you can wait for #4941 to fix this.

PR #6234, opened 2018-07-12
Fix setTransform function for desktop webkit

Jesus89:
### Abstract

This PR solves an old (2016) and a common issue regarding gaps between the tiles :white_square_button: The following GIF proves that his fix already works: no gaps are appreciated.

dark

### History

I attach below the related tickets and pull requests regarding this issue, only inside Leaflet's repository (there are more tickets talking about the same issue in other projects that use Leaflet):

Tickets related:
- https://github.com/Leaflet/Leaflet/issues/3575
- https://github.com/Leaflet/Leaflet/issues/3339
- https://github.com/Leaflet/Leaflet/issues/1379
- https://github.com/Leaflet/Leaflet/issues/1066
- https://github.com/Leaflet/Leaflet/issues/5078

PRs related:
- https://github.com/Leaflet/Leaflet/pull/4941 (superseded)
- https://github.com/Leaflet/Leaflet/pull/5290 (superseded)
- https://github.com/Leaflet/Leaflet/pull/5354 (ignored)
- https://github.com/Leaflet/Leaflet/pull/5476

We see that these PRs https://github.com/Leaflet/Leaflet/pull/4941, https://github.com/Leaflet/Leaflet/pull/5290 tried to fix the issue but finally, they were superseded by this one https://github.com/Leaflet/Leaflet/pull/5476 that enabled the [Leaflet.TileLayer.NoGap](https://github.com/Leaflet/Leaflet.TileLayer.NoGap) plugin.

The problem is that this Leaflet.TileLayer.NoGap doesn't work at all. Only navigating in the git-tree you can find pseudo-working states. And there is another issue of that implementation: using a canvas for the transitions has a very bad performance.

### Solution

There was an ignored PR (https://github.com/Leaflet/Leaflet/pull/5354) closed in favour of the non-working TileLayer.NoGap plugin. The point is that this approach goes in the right direction to fix the issue and that's why this PR is based on that one.

This solution is based on using `translate` instead of `translate3d` in the transformation function.

A key difference implemented in this PR is that the bug is not happening in mobile devices so it shouldn't be fixed there. That's why the condition is `!mobile && (ie3d || webkit3d)`.

Final notes. We have this solution successfully working at CARTO, using a forked Leaflet, so it would be nice for us to move this forward and have the solution in the upstream, that will allow also other users with the same problem to benefit from it.

Thanks in advance!

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -208,8 +208,14 @@ export function testProp(props) {
 export function setTransform(el, offset, scale) {
 	var pos = offset || new Point(0, 0);
 
-	el.style[TRANSFORM] =
-		(Browser.ie3d ?
+	el.style.willChange = scale !== 1 ? 'auto' : 'transform';
+	el.style[L.DomUtil.TRANSFORM] =
+		// Based on https://github.com/Leaflet/Leaflet/pull/5354:
+		// We don't want to use `translate3d` for webkit since it causes the
+		// white lines issue (tile's gaps) while panning or zooming:
+		// https://bugs.chromium.org/p/chromium/issues/detail?id=600120#c15
+		// This issue does not appear on mobile devices.
+		(!Browser.mobile && (Browser.ie3d || Browser.webkit3d) ?
 			'translate(' + pos.x + 'px,' + pos.y + 'px)' :
 			'translate3d(' + pos.x + 'px,' + pos.y + 'px,0)') +
 		(scale ? ' scale(' + scale + ')' : '');


ramiroaznar:
cc @IvanSanchez :-)

cherniavskii:
Hi,
Thanks for opening a PR!

Did you check fractional zoom? See https://github.com/Leaflet/Leaflet/pull/5354#issuecomment-281608329.
Also, we need to test both desktop and mobile browsers to push this PR forward.
I'll try to find some time for that.

Jesus89:
Hello. Thanks for addressing this PR :)

Yes. I have checked that this solution is not perfect when you force the zoom to snap non-integer values, for example with this configuration `{center: [0,0], zoom: 2.1, zoomSnap: 0.1}`, so it behaves the same.

However, it fixes the common cases of transitions between integer values of zoom for both `buttons` and `scroll` zoom. I attach some GIF to see the differences and improvements.

NOTE: this PR does not affect mobile devices since this bug is only relevant for desktop WebKit browsers.

### Buttons zoom

| Bug |
|---|
| bug-dzoom | 

| Fix |
|---|
| fix-dzoom |

### Scroll zoom

| Bug |
|---|
| bug-szoom |

| Fix |
|---|
| fix-szoom |

mourner:
Changing the way `will-change` is applied could also have significant performance implications, so we should test very thoroughly and compare FPS on flyTo animations with before and after this patch.

nparley:
Just to note there are other problems with subpixel translate 3d which can cause the text in popups to be blurry on chrome. E.g. this pr https://github.com/Leaflet/Leaflet/pull/6240. Moving to translate from translate3d does not fix this problem. Is the better solution to round?

ryanalta:
While trying to figure out where a blurring was coming from, I ended up here via #3297 and #6069. For my own purposes, it was enough to throw in some `Math.floor` on `setTransform` to test. Perhaps someone more willing to dive into this would find a good cross-browser solution to deal with Chrome.

```
// browser doesn't support 3D CSS transforms.
function setTransform(el, offset, scale) {
	var pos = offset || new Point(0, 0);
    
	el.style[TRANSFORM] =
		(ie3d ?
			'translate(' + Math.floor(pos.x) + 'px,' + Math.floor(pos.y) + 'px)' :
			'translate3d(' + Math.floor(pos.x) + 'px,' + Math.floor(pos.y) + 'px,0)') +
		(scale ? ' scale(' + scale + ')' : '');
}
```

therealPaulPlay:
Any updates regarding this? Why has this fix never made it into prod?

PR #6235, opened 2018-07-14
:bulb: plugins.md  added

dayjournal:
- :bulb: Leaflet.Control.Opacity added


ghybs:
Thank you for your plugin!

Issue #6232, opened 2018-07-09
Leaflet 


EugenRakhimov:
## How to reproduce
It's the same issue as the closed one: https://github.com/Leaflet/Leaflet/issues/3297
 Leaflet version I'm using: 1.3.1
Browser (with version) I'm using: Chrome Version 67.0.3396.99 (Official Build) (64-bit)
- OS/Platform (with version) I'm using: MacOS Sierra Version 10.12.6
- step 1
Go to official leaflet site, first page
https://leafletjs.com/
- step 2
Look at the map with the popup.
Popup is blured, text is blured.
![screen shot 2018-07-10 at 10 40 20](https://user-images.githubusercontent.com/12945598/42479900-1d519cfc-842f-11e8-814f-036491942651.png)

## What behaviour I'm expecting and which behaviour I'm seeing
I expect to see sharp text and sharp popup as in Firefox.
![screen shot 2018-07-10 at 10 40 51](https://user-images.githubusercontent.com/12945598/42479925-351396b0-842f-11e8-8a43-d47d47c56993.png)
But I got the blured view
![screen shot 2018-07-10 at 10 40 20](https://user-images.githubusercontent.com/12945598/42479900-1d519cfc-842f-11e8-814f-036491942651.png)

## Minimal example reproducing the issue
Just open https://leafletjs.com/  with chrome on mac OS.

Using http://leafletjs.com/edit.html or any other jsfiddle-like site.

EugenRakhimov:
I think it's the result of positioning with transform.

EugenRakhimov:
So, for whom it may concern, I fixed blurring in my project by assigning
`window.L_DISABLE_3D = true ` before leaflet initialisation. 
Apparently it tells leaflet to not use translate3D, and use left and top instead. Haven't noticed drop of speed, so it may help to someone like me.

sjoodin:
Experienced the same blurry behaviour. Noticed that I had set chrome's page zoom to 90%. Setting the page zoom back to 100% resolved the problems for me.

Falke-Design:
I recognized this sometimes too, but I never investigated it

Issue #6069, opened 2018-02-24
Blurring due to fractional CSS transform


brianreavis:
Continuing from https://github.com/Leaflet/Leaflet/issues/6068.

## How to Reproduce

Open http://leafletjs.com/examples/quick-start/ on a non-retina screen. Pan the map slowly, and note the tiles sometimes crisper, sometimes blurrier due to interpolation. To further see the effect, click on the map to get the "You clicked the map at..." popup, which is also blurry depending on the current translation.

- Leaflet version I'm using: (current)
- Browser (with version) I'm using: **Chrome Version 64.0.3282.186**
- OS/Platform (with version) I'm using: OSX High Sierra 10.13.3 (17D47)

## Screenshots

![image](https://user-images.githubusercontent.com/118480/36634877-bce7c4d0-1968-11e8-9aac-f2ced539d58c.png)

![image](https://user-images.githubusercontent.com/118480/36633507-5c852416-1954-11e8-94dd-da17c623b0b1.png)

This can also lead to subtle seam artifacts, depending on how you've panned:

![image](https://user-images.githubusercontent.com/118480/36634696-97421684-1965-11e8-9fe2-0e88114d9252.png)

ghybs:
Hi @mbullington,

Thank you for your look into this!

Please would you be able to publish a patch of your fixed version, so that other users (like @EugenRakhimov) can test it in their configuration and report if it works in their case?

A simple way would be to setup e.g. a Plunker (https://leafletjs.com/edit) with:
```javascript
L.Draggable.include({
  _updatePosition: yourModifiedFunction
})
```
before instantiating the map.

You can also send a PR of course.

mbullington:
Thanks @ghybs!

I implemented the fix here using the tool you suggested: https://plnkr.co/edit/wmVxVPD6S7trPlaRDBF3?p=preview

This test case can be contrasted with https://leafletjs.com/examples/quick-start/ , where the issue persists. I'm using a normal 1080p display connected to a mid-2012 MacBook Pro.

@brianreavis @EugenRakhimov Could you please test the above link on your devices and see if it fixes the issue?

EugenRakhimov:
@mbullington Looks fine to me.

MikaelEdebro:
We're also seeing this behaviour. However not only on Pan, but always.
Here is a screenshot comparing on the same screen:
1. The map in QGis
2. The WMS tile opened in new tab
3. The map in leaflet

[...]

Issue #3297, opened 2015-03-10
Chrome: Blurry map


molcochet:
Hi,
When loading a map in chrome, the text and lines are slightly blurry. The blurriness changes as the map is zoomed. IE and Firefox do not that be the same issue are a re clear and distinct at all zoom levels. I think it might have something to do with the transform property not correctly getting tagged as -webkit.

perliedman:
Could you provide Chrome version and OS. Preferrably screenshots, as well.

Also, are you seeing this issue on the Leaflet demo map on the Leaflet site, or where do you see it?

perliedman:
Made a comparison myself, this is from Chrome 40.0.2214.115 on Windows:

-------

![image](https://cloud.githubusercontent.com/assets/1246614/6575631/dd6ad522-c730-11e4-931e-74111df9d9ca.png)

--------

Compared to Internet Explorer 11 on the same machine:

--------
![image](https://cloud.githubusercontent.com/assets/1246614/6575642/eb4f7f44-c730-11e4-92c8-c9d98a9c3ea1.png)

--------

Map is from Leaflet's homepage: http://leafletjs.com/

As can be seen, Chrome is visibly blurrier, and also the popup har som small artifact around the popup's arrow. However, it's not only the map that is blurry, the font rendering is also much blurrier, which should be outside Leaflet's control. Also worth noting is that these screenshots are from a setup with dual monitors, of which one is high-DPI, so the blurriness might be related to Chrome scaling things.

@molcochet, how does this compare to the issue you're seeing? What equipment do you use, etc.?

molcochet:
Hi,
Sorry for the lack of detail earlier. I also used  http://leafletjs.com/ to recreate after noticing it on my configuration. Your screenshots are exactly the issue I have observed.

My test configuration: 
Browsers: 
Chrome 40.0.2214.115 m
Firefox: Developer 38.0a2 (2015-03-06)
IE: IE11 11.0.9600.17633

OS: Windows 7 SP1

Hardware:
Dual Monitor, one High DPI. Also recreated using only low DPI configuration (disconnected High DPI monitor and restarted chrome to verify.

Leaflet:
0.7.3 & 0.8-Dev & whichever version is currently running the website as a stable base for comparison
Additional plugins:
Leaflet-label

molcochet:
I tried an older version of our product, using 0.7.2, and the issue does not occur. Screenshots below.

Chome, Leaflet 0.7.2:
[...]

PR #6240, opened 2018-07-17
On non-retina screens, round Draggable values to prevent bad subpixel rendering

mbullington:
This PR fixes the issue brought up by #6069 (and potentially #6232 #3297 as well) where dragging the map pane on a non-retina screen causes the map to become blurry. This is because the resulting CSS transform is using fractional values.

You can try a demo here: https://plnkr.co/edit/wmVxVPD6S7trPlaRDBF3?p=preview

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -185,7 +185,7 @@ export var Draggable = Evented.extend({
 		// Fired continuously during dragging *before* each corresponding
 		// update of the element's position.
 		this.fire('predrag', e);
-		DomUtil.setPosition(this._element, this._newPos);
+		DomUtil.setPosition(this._element, !Browser.retina ? this._newPos.round() : this._newPos);
 
 		// @event drag: Event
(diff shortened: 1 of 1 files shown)


mourner:
As far as I remember (very vaguely), we used to do rounding but dropped it because it introduced noticeable zigzag-like pattern when panning. I wonder if it's only noticeable on Retina? We should explore the implications. Another option would be to round when the drag ends, but not during panning.

EugenRakhimov:
@mbullington just in case you need it only for web-kit based browsers. Firefox and even IE do a good job.

nparley:
Any movement on this PR, the text blurring on sub pixel translate3d is quite noticeable in chrome. I can confirm for me there is no issue on retina.

lulzmachine:
Any progress on this? It looks like the pull requests are satisfactory.

johnd0e:
> the text blurring on sub pixel translate3d is quite noticeable in chrome

MacOS?

PR #6248, opened 2018-07-21
Enhance Map accessibility

Muhnad:
**Changes**:
- Replace anchor tags with buttons.
- Add Alt attribute to images.
- Add roles to elements.
- Make tooltip focusable.

### DIFF
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -103,7 +103,7 @@ describe("Marker.Drag", function () {
 				onStop: function () {
 					var center = map.getCenter();
 					expect(center.lat).to.be(0);
-					expect(center.lng).to.be.within(0, 30);
+					expect(center.lng).to.be.within(10, 30);
 
 					var markerPos = marker.getLatLng();
 					// Marker drag is very timing sensitive, so we can't check
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -26,6 +26,7 @@ describe("Marker.Drag", function () {
 			var hand = new Hand({
 				timing: 'fastframe',
 				onStop: function () {
+
 					var center = map.getCenter();
 					expect(center.lat).to.be(0);
 					expect(center.lng).to.be(0);
@@ -102,7 +103,7 @@ describe("Marker.Drag", function () {
 				onStop: function () {
 					var center = map.getCenter();
 					expect(center.lat).to.be(0);
-					expect(center.lng).to.be.within(10, 30);
+					expect(center.lng).to.be.within(0, 30);
 
 					var markerPos = marker.getLatLng();
 					// Marker drag is very timing sensitive, so we can't check
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -159,7 +159,7 @@ describe("Marker", function () {
 			} else {
 				expect(marker._icon.src).to.contain('marker-icon.png');
 			}
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
 		});
 
 		it("changes an image to a DivIcon", function () {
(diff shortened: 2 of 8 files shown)


IvanSanchez:
It seems these changes break the continuous integration tests - it seems that the automated tests for dragging markers are failing. Can you please check that?

Muhnad:
Hi @IvanSanchez, sorry I'm busy these days so I didn't check what happened in testing phase :D 

Do you have any idea about why the maker falls when we wrap it in button?

fiinix:
Has the fix for this issue been abandoned?

spec/suites/layer/marker/Marker.DragSpec.js
```diff
@@ -102,7 +103,7 @@ describe("Marker.Drag", function () {
 				onStop: function () {
 					var center = map.getCenter();
 					expect(center.lat).to.be(0);
-					expect(center.lng).to.be.within(10, 30);
+					expect(center.lng).to.be.within(0, 30);
```
johnd0e, line 106: Why change this?

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -159,7 +159,7 @@ describe("Marker", function () {
 			} else {
 				expect(marker._icon.src).to.contain('marker-icon.png');
 			}
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
```
johnd0e, line 162: Parent of parent is an overkill.
Maybe something wrong in marker design.

src/control/Control.Zoom.js
```diff
@@ -80,23 +80,18 @@ export var Zoom = Control.extend({
...
-
-		return link;
+		var button = DomUtil.create('button', className, container);
+		button.innerHTML = html;
+		button.title = title;
+
+		button.setAttribute('aria-label', title);
+
+		DomEvent.disableClickPropagation(button);
+		DomEvent.on(button, 'click', DomEvent.stop);
```
johnd0e, line 90: Originally this was needed to prevent default action of link.
Now it is not needed.

src/layer/marker/Icon.js
```diff
@@ -136,9 +136,10 @@ export var Icon = Class.extend({
 		}
 	},
 
-	_createImg: function (src, el) {
+	_createImg: function (src, el, alt) {
 		el = el || document.createElement('img');
 		el.src = src;
+		el.alt = alt || '';
```
johnd0e, line 142: In other part you use 'Location' string instead of empty string.

src/layer/marker/Marker.js
```diff
@@ -205,17 +209,14 @@ export var Marker = Layer.extend({
 			}
 
 			if (icon.tagName === 'IMG') {
-				icon.alt = options.alt || '';
+				icon.alt = options.alt || 'Location';
```
johnd0e, line 212: Instead of hardcoding some arbitrary string you better change `options.alt` defaults.

johnd0e:
Other thoughts:
- I do not think that button is good for Marker
- Even if I'm wrong, proposed implementation is bad anyway.
- And I haven't reviewed Popup implementation yet.

spec/suites/layer/marker/Marker.DragSpec.js
Muhnad: Actually I don't remember anything literally anything about that, I forget I opened this PR :D

spec/suites/layer/marker/MarkerSpec.js
Muhnad: I agree with you, it such a stupid thing I did.

src/layer/marker/Icon.js
Muhnad: It depends on the use case.

in this case, you create a new image tag so you need to add alt value or leave it with an empty string.

src/layer/marker/Marker.js
Malvoz: "location" seems opinionated, and may not make sense in a non-geographic map. How about "Marker"?

src/layer/Popup.js
```diff
@@ -179,7 +179,11 @@ export var Popup = DivOverlay.extend({
 			' leaflet-zoom-animated');
 
 		var wrapper = this._wrapper = DomUtil.create('div', prefix + '-content-wrapper', container);
-		this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+		var wrapperContent = this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+
+		wrapperContent.setAttribute('role', 'tooltip');
```
Malvoz, line 184: [`role="tooltip"`](https://www.w3.org/TR/wai-aria-1.1/#tooltip) doesn't seem appropriate for popups, rather [`role="dialog"` ](https://www.w3.org/TR/wai-aria-1.1/#dialog).

PR #6261, opened 2018-07-30
Add the need for attribution to the quickstart

pietervdvn:
I noticed that people new to OSM and needing a map often follow the leaflet tutorial, but remove the attribution from the map because it is 'an unwanted advertisement' or 'visual noise'. It would help that [the quick-start](https://leafletjs.com/examples/quick-start/) would explicitly note the obligation to have attribution.

I added an extra paragraph outlining this need, which hopefully clears the confusion with new users.


ghybs:
:+1:

pietervdvn:
Thanks!

PR #6263, opened 2018-08-01
Tried Fixing flickering of zoom levels

### LINKED ISSUE #6259: maxNativeZoom can cause flickering
Hello,

I am finding that if you have two map layers (with different zIndexes) and one of them has maxNativeZoom set to the same value as it's maxZoom, the map will flicker between the two layers when panning.

Here is an example of the problem.  All you need to do is pan the map left and right and you will see the layers flicker:

https://codepen.io/happicow/pen/WKXazR

Mitigation techniques:
- If you remove the maxNativeZoom value the issue resolves.
- If you set maxNativeZoom to a value greater than maxZoom (6 for example) the issue resolves.
- If you downgrade to Leaflet v1.0.0 the issue resolves.  (v1.0.3 seems to be when this bug appeared)

This has been tested on the latest versions of Firefox, Chrome and Safari on both Windows and MacOS with the same results.

Thanks.

mjguru:
I am not able to test it locally. I would rather love to chat more about this and understand more about the issue. Still, this tries to fix #6259

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -544,10 +544,12 @@ export var GridLayer = Layer.extend({
 	},
 
 	_setView: function (center, zoom, noPrune, noUpdate) {
-		var tileZoom = this._clampZoom(Math.round(zoom));
+		var tileZoom = 0;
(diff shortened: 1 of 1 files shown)


johnd0e:
Superseded by #6310

Issue #6136, opened 2018-04-16
Drop support for Internet Explorer


perliedman:
> **Note**: Support for IE is being dropped in [v2.0](https://github.com/Leaflet/Leaflet/milestone/39)!

I think it's really time to discuss dropping support for IE8, and also possibly 9 and 10. These browsers are supported by few web sites these days, and for users who really need to support them, we can point to older versions of Leaflet.

Maintaining support for olders IE versions is a pain, and it appears we rarely test it anyway (we broke IE8 in a previous release without noticing and nobody really told us).

perliedman:
Some discussion on our IE support in this thread: #4745.

IvanSanchez:
I guess it's time to [run this informal poll again](https://twitter.com/LeafletJS/status/685052811374129152)?

mourner:
I'm up for dropping IE8 support, and agree that we can just offer older versions of Leaflet for people who need to support it. Even jQuery no longer supports that version (https://jquery.com/browser-support/), so people using such a browser should expect the majority of the web to be mostly broken.

We should consider dropping support for IE9 and IE10 too. I just fetched the latest data from http://gs.statcounter.com/ and the breakdown is as follows:

- IE 11: 2.7%
- IE 10: 0.11%
- IE 9: 0.11%
- IE 8: 0.17%

All the IE8–10 values dropped by 2x in the last year.

tomhughes:
I had a look at the openstreetmap.org statistics at the request of @IvanSanchez and our numbers for yesterday are:

* IE 11: 9% (11128 visitors)
* IE 10: 0.1% (99 visitors)
* IE 9: 0.2% (198 visitors)
* IE 8: 0% (4 visitors)
* IE 6: 0% (2 visitors)

I suspect anything before IE9 barely works other parts of the site anyway, if it's actually functional at all.

codeofsumit:
I'd vote for dropping everything below IE11.

franciscop-invast:
I would say that libraries websites such as openstreetmap.org are *not* representative; some developers will be doing sanity checks on those pages to make sure they work. Before considering any library and if there's a demo, I open it in the browsers I need to support to make sure the basics are working. Now this implies IE11, but before I would use browserstack to open libraries demos in other browsers, very likely slightly tweaking the analytics.

septatrix:
Pinging @mourner because I think a year and a half later it may be time to again discuss this and come to a conclusion?

amb26:
[...]

PR #6279, opened 2018-08-08
remove trailing commas

helbling:


helbling:
To have the same syntax style as in the rest of the code base - there is no trailing comma anywhere else
Our syntax checker throws a warning as some IE versions cannot handle trailing commas. Not sure if these versions are still relevant.

IvanSanchez:
Hm, I supposed that `eslint` would catch such a thing, and that the rollupJS-based build process (which specifies the `legacy` option for rollup, ensuring ES3 syntax) would change this accordingly. Or maybe `uflifyJS` is taking care of the dangling commas so the minified version works on IE8 but the non-minified version spits out errors.

@mourner @perliedman Any thoughts about changing the linting rules, or deprecating IE8 altogether (#6136)?

mourner:
Let's make sure our linter catches this now (along with merging the PR), and consider dropping IE8 separately.

PR #6285, opened 2018-08-14
Plugins: add leaflet.repeatedmarkers

### LINKED ISSUE #1945: Markers Don't Repeat
I need markers to repeat across as different world maps are shown as you pan.

I have created a jsfiddle at: 
http://jsfiddle.net/deliriousrhino/u387Q/3/ 

If you look at New Zealand you will see my issue. There should be markers on either side of the country.

I tried worldCopyJump = true, however it throws an error "Set map center and zoom first."

Also I am no sure worldCopyJump is the functionality I am after since I need all the markers around the -180 and 180 longitude to be shown together.

IvanSanchez:
This should close #1945 forever and ever.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1643,6 +1643,17 @@ These plugins provide new markers or news ways of converting abstract data into 
 			<a href="https://webkid.io">webkid</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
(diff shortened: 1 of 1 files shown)


perliedman:
:tada: Amazing! I had just written a lengthy reply to whining about people whining, but this is a much better way to address it 😄

PR #6290, opened 2018-08-16
fixed capitalization of 'webkitTransform'

tuckergordon:
Fixing capitalization issue that broke transformations on certain webkit browsers (e.g. IOS 8.1). 

This was the root of this issue on StackOverflow: https://stackoverflow.com/questions/51861944/leaflet-tiles-not-loading-correctly-in-angular-on-mobile?noredirect=1#comment90689799_51861944


IvanSanchez:
For reference: https://stackoverflow.com/questions/708895/how-to-set-the-style-webkit-transform-dynamically-using-javascript

PR #6302, opened 2018-08-21
Added custom HTML attributes for Marker icon.

vasiledraghita:
Send markerAttributes prop for Marker component with custom html attributes in order to set them for  tag. Eg:  markerAttributes={{ customAttr: 'value' }}

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -215,6 +215,15 @@ export var Marker = Layer.extend({
 			icon.tabIndex = '0';
 		}
 
+		if (options.markerAttributes) {
+			// Add custom HTML attributes for Marker icon.
+			for (var item in options.markerAttributes) {
+				var attr = document.createAttribute(item);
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Thanks for the PR, but I don't think this is a good addition to Leaflet.

Allowing arbitrary attributes to an `L.Icon`'s attributes would allow users to mess up Leaflet's positioning logic way too easily. If users want to use custom HTML in their markers, they can already resort to a `L.DivIcon` with arbitrary contents. Those are nested, creating some guarantee that the positioning will not break. Subclassing `L.Icon` is also a cleaner option that allowing arbitrary attributes.

Besides, icon stuff should be in `L.Icon` (or subclasses) and there are no docstrings for the new option.

And also, here in Leaflet we don't use reactjs jargon ("props") nor jsx syntax (`{{ customAttr: 'value' }}`).

vasiledraghita:
Thanks for the review. It worked with L.divIcon for what I wanted to do.

PR #6310, opened 2018-08-23
moves _clampZoom to else conditional in _setView to run after max and minZoom checks

robertbernardbrown:
Hi all, took a shot at fixing issue #6259  - apologies if I'm stepping on any toes.

I took my cue from @IvanSanchez on how to fix the bug and moved the zoom clamp into the max/minZoom conditional check. Basically _setView will now set the value of tileZoom conditionally based off the max/minZoom.

I received a "success" from the tests and tested against #3814 and @happicow 's example in the origin issue. I don't detect any flickering when panning maps nor do I experience any pan-pinch-zoom tile-rendering issues. I tested on Chrome and Safari on my macbook, Chrome and Safari on my iPhone, and Chrome and Safari on my iPad. Here's a gif of the iPhone test on Chrome:
leaflet

This is my first PR on an open source project. Please let me know if there are anything I missed or if anything needs adjustment.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -544,10 +544,12 @@ export var GridLayer = Layer.extend({
 	},
 
 	_setView: function (center, zoom, noPrune, noUpdate) {
-		var tileZoom = this._clampZoom(Math.round(zoom));
-		if ((this.options.maxZoom !== undefined && tileZoom > this.options.maxZoom) ||
-		    (this.options.minZoom !== undefined && tileZoom < this.options.minZoom)) {
+		var tileZoom;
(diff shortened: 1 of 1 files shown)


DanielHabib:
This was driving me crazy! Thanks for fixing

WHI-CH:
Wasted a lot of time on this, thanks for the fix!
So why isn't this in master yet?

robertbernardbrown:
I'm not sure. I'm hoping it can still find it's way in soon. I'm assuming other priorities keep popping up for the maintainer

rcollette:
1 year for 5 lines of code with multiple confirmations of the fix and its still not in?  No unit tests though.

This doesn't bode well for my evaluation of this library.

Issue #6158, opened 2018-05-09
setUrl causes tiles loaded with tileLayer to flicker


kwankong:
One behaviour of setUrl I find odd is that suppose you have tiles=tileLayer(url) loaded, when you run tiles.setUrl(url), the tiles would flicker, which suggests the tiles were removed and then reloaded, even if you use the same url.  But if you have image=imageOverlay(url) loaded, image.setUrl(url) does not cause the image to flicker at all.  Has anyone noticed this?  Why would setUrl behave differently when applied toward tileLayer?

IvanSanchez:
Can you publish a fiddle/codepen/[plunkr](https://leafletjs.com/edit) that displays this behaviour?

kwankong:
Okay, Ivan,
I have a couple of demos set up using
(1) imageOverlay: https://plnkr.co/edit/sa74aFv4u2jgsFBU93G6?p=preview
(2) tileLayer: https://plnkr.co/edit/nxnga7MqeC3bgUEM07Dz?p=preview

perliedman:
Looking at the code, I can see this:

* Calling `TileLayer.setUrl` will always cause a `redraw` unless the second argument to `setUrl` is `true`: https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/TileLayer.js#L117
* `ImageOverlay.setUrl` will update the `img` element's `src` attribute, which I assume the browser is smart enough to optimize into a no-op if `src` is the same

We could quite easily add the same optimization in `TileLayer.setUrl`, checking first if the new URL differs from the old. The only downside I can see is if someone relies on `setUrl` to always make a redraw.

kwankong:
A couple of questions:
1. When TileLayer.setUrl is called, does it redraw only the visible tiles or all the tiles, for all zoom levels?
2. Is it possible to update the src attribute in TileLayer similar to how it is done in ImageOverlay?

IvanSanchez:
> When TileLayer.setUrl is called, does it redraw only the visible tiles or all the tiles, for all zoom levels?

It redraws only visible tiles (but will use the new URL template for future tiles)

Keep in mind that (in the default CRS), for a zoom level `z`, there are `4^z` tiles. That's like 17 thousand million for zoom level 17, and no browser is loading (nor reloading) that amount of tiles, ever.

> Is it possible to update the src attribute in TileLayer similar to how it is done in ImageOverlay?

Yes, it is. But no one has written the code for it yet.

[...]

PR #6313, opened 2018-08-26
Avoid unnecessary redrawing in TileLayer.setUrl() when URL does not change.

JackNeus:
I've written a potential fix for issue #6158.

I essentially added a check in TileLayer.setURL() to see if the current url is the same as the new one. If this is the case AND noRedraw is undefined, I set noRedraw to true, which avoids redrawing a tile that hasn't actually changed. If, for some reason, a developer wants to redraw, they can still do that by passing a value of true to the noRedraw parameter.

I received "SUCCESS" when testing as detailed in CONTRIBUTING.md. I also tested against @kwankong's example (https://plnkr.co/edit/nxnga7MqeC3bgUEM07Dz?p=preview).  The flickering stops with my fix.

This is my first open source contribution, so please let me know if there's anything I can do better. I'd appreciate all the feedback I can get!

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -114,7 +114,13 @@ export var TileLayer = GridLayer.extend({
 
 	// @method setUrl(url: String, noRedraw?: Boolean): this
 	// Updates the layer's URL template and redraws it (unless `noRedraw` is set to `true`).
+	// If the URL does not change, the layer's URL template will not be redrawn unless
+	// the noRedraw parameter is set to false.
 	setUrl: function (url, noRedraw) {
+		if (this._url === url && noRedraw === undefined) {
+			noRedraw = true;
+		}
+
 		this._url = url;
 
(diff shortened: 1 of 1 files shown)


src/layer/tile/TileLayer.js
```diff
@@ -114,7 +114,13 @@ export var TileLayer = GridLayer.extend({
 
 	// @method setUrl(url: String, noRedraw?: Boolean): this
 	// Updates the layer's URL template and redraws it (unless `noRedraw` is set to `true`).
+	// If the URL does not change, the layer's URL template will not be redrawn unless
```
perliedman: I would phrase this as "[...] the layer will not be redraw unless [...]", since it's in fact the layer, not the URL template that is redrawn.

JackNeus:
Hi @perliedman, sorry for the delay. I've made the requested change!

perliedman:
Great, thanks!

PR #6319, opened 2018-08-29
Add setLastLatLng to Polyline

danielthegray:
I added a new method to set the last point of a Polyline to a certain value (without copying, modifying and replacing the original array each time).

I think it is a useful feature so I have added the method. I have respected the coding style the best I could and `npm run lint` doesn't return any errors. I have compiled the rolled up JS file and used it in my project successfully.

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
 		return this.redraw();
 	},
 
+	// @method setLastLatLng(latlng: LatLng, latlngs? LatLng[]): this
+	// Modifies the last point of the polyline to be the specified point.
+	// By default, adds to the first ring of the polyline in case of a
+	// multi-polyline, but can be overridden by passing a specific ring as a
+	// LatLng array (that you can earlier access with
+	// [`getLatLngs`](#polyline-getlatlngs)).
+	setLastLatLng: function (latlng, latlngs) {
+		latlngs = latlngs || this._defaultShape();
+		latlng = toLatLng(latlng);
+		latlngs.pop();
+		latlngs.push(latlng);
+		this._bounds.extend(latlng);
+		return this.redraw();
+	},
+
 	_setLatLngs: function (latlngs) {
 		this._bounds = new LatLngBounds();
 		this._latlngs = this._convertLatLngs(latlngs);


IvanSanchez:
This looks fine at first sight, but it suffers from some subtle problems, and I don't think it's that useful - it's safer to let developers run `getLatLngs()` and `setLatLngs()` instead of providing specific methods for each possible operation. Should leaflet expose `setFirstLatLng()` or `spliceLatLngs()`? Feels a bit like feature creep.

Also this passes the lint and build :smile: but doesn't have any tests :worried:

src/layer/vector/Polyline.js
```diff
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
 		return this.redraw();
 	},
 
+	// @method setLastLatLng(latlng: LatLng, latlngs? LatLng[]): this
+	// Modifies the last point of the polyline to be the specified point.
+	// By default, adds to the first ring of the polyline in case of a
+	// multi-polyline, but can be overridden by passing a specific ring as a
```
IvanSanchez: This is confusing. Multi-polylines don't have rings, because only polygons (and multi-polygons) have rings.

src/layer/vector/Polyline.js
```diff
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
...
+	// LatLng array (that you can earlier access with
+	// [`getLatLngs`](#polyline-getlatlngs)).
+	setLastLatLng: function (latlng, latlngs) {
+		latlngs = latlngs || this._defaultShape();
```
IvanSanchez: There is no guarantee that `latlngs` is related at all to the polyline's coordinate set. Not only for the obvious case of specifying an explicitly unrelated `latlngs`, but also for the subtle cases where shallow copies of the arrays are created. In those cases, this method just modifies the parameter passed by reference, not the data contained in `this`.

src/layer/vector/Polyline.js
```diff
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
...
+		latlng = toLatLng(latlng);
+		latlngs.pop();
```
IvanSanchez: This doesn't cover the edge case of a polyline without any `latlng`s. `var line = L.polyline([]);` is a valid thing to do.

src/layer/vector/Polyline.js
```diff
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
...
+		latlngs.push(latlng);
+		this._bounds.extend(latlng);
```
IvanSanchez: This will only extend the bounds, never make them smaller. This allows for an undesirable edge case where calling `setLatLngs()` with random points all over the globe and finally with a "good" point will make the polyline's bounds way larger than they should be.

src/layer/vector/Polyline.js
```diff
@@ -172,6 +172,21 @@ export var Polyline = Path.extend({
...
+		return this.redraw();
```
IvanSanchez: ...and `redraw()` will redraw the whole geometry anyway, which means any performance gains from this new method are gonna be negligible.

src/layer/vector/Polyline.js
danielthegray: This comment is similarly present in the `addLatLng()`, so I would think that any problem with this comment here should also apply to `addLatLng()` (just the previous method in this class)

src/layer/vector/Polyline.js
danielthegray: It does :) it is precisely my use-case (I start with an empty polyline and add points along the way). If you test the following in your browser:
```>> a=[]
Array []
>> a.pop()
undefined```
You can see that popping an empty list is no problem.

src/layer/vector/Polyline.js
danielthegray: I was hoping to gain performance from avoiding the instantiation + copying of a potentially long array on each mouse move. I'll try with the array copying and see if I run into any performance problems.

danielthegray:
Thanks for your review!

danielthegray:
I experimented and you were right, the biggest performance drain is redrawing it every time. Thank you!

Issue #6333, opened 2018-09-06
[website doc issue] non working link in plugins page


solsticedhiver:
On https://leafletjs.com/plugins.html, all the in-page links after "Clustering/decluttering" (included) are non functionnal.

It is easy to see tthat the page is generated from some markdown code because one can see on the page something like

`</table> ### Overlay animations These plugins animate markers or some geometries. See also [geometries with time or elevation](#geometryinteraction-time).`

and below some headers are not headers and not links because they are like

`### Clustering/Decluttering When you are displaying a lot of data, these plugins will make your map look cleaner.`

mgred:
@solsticedhiver This is caused by an additional `<tr>` tag in the list of Markers & Renderers.

IvanSanchez:
Fixed by #6334. @mgred, thanks for fixing this!

PR #6334, opened 2018-09-07
fix(docs): remove additional <tr>

mgred:
### Problem

See #6333 

### Solution

Remove an additional <tr> tag from the table of Markers & Renderers that causes display issues and links not working.

### Extra

This commit also replace an html link with md syntax, which was not related to the issue (#6333).


mgred:
Cool

PR #6337, opened 2018-09-08
Fix #6336 : use openstreetmap tiles in quickstart guide

### LINKED ISSUE #6336: "Leaflet Quick Start Guide" should offer a simpler Tile Provider
Hi, i've just read someone saying that quick start doesn't work since it use MB tiles with no accessToken defined. See also #6280 & #6268 and as @IvanSanchez said : 

> apparently the mapbox tokens are confusing more people.

I propose to change the base tiles on this example, and maybe it would be better to apply this everywhere MapBox tiles are used in examples pages. Any thought about it ?

brunob:
et voilà :)

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -13,7 +13,7 @@ Every tutorial here comes with step-by-step code explanation and is easy enough 
 page="quick-start/"
 thumbnail="quick-start/thumbnail.png"
 title="Leaflet Quick Start Guide"
-description="A simple step-by-step guide that will quickly get you started with Leaflet basics, including setting up a Leaflet map (with Mapbox tiles) on your page, working with markers, polylines and popups, and dealing with events."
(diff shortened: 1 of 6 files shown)


brunob:
@yohanboniface ping ;)

ghybs:
Thank you for having proposed this PR.

As explained in https://github.com/Leaflet/Leaflet/issues/6336:
- Leaflet pages receive quite some high traffic, which could be considered a "heavy use". Hence we should not use OSM tiles.
- [OSM Tile Usage Policy](https://operations.osmfoundation.org/policies/tiles/) also advises _against_ hardcoding OSM tile URL into an app, so providing it in Leaflet code snippets could also be interpreted as a violation.

matkoniecz:
@ghybs I there some reason to close this PR without closing #6336 itself?

PR #6350, opened 2018-09-27
Add general purpose metric grid overlay

bill-chadwick:
A better metric grid overlay suitable for British, Irish, UTM grids etc. With example and definitions for British and Irish grids.


ghybs:
Thank you for contributing this nice plugin! :+1:

PR #6362, opened 2018-10-09
Fix(Map): set internal flags at beginning of map initialization

ghybs:
Hi again,

Fix https://github.com/Leaflet/Leaflet/issues/6165

As explained in the above issue, due to assignment of some internal flags in the middle of the map initialization, there is an edge case where `invalidateSize` will have trouble computing the size difference when called.

By setting these flags at the beginning of the map init (in particular, before any call to `setView`, which is internally executed when the map is initialized with `center` and `zoom` options), we avoid an inconsistent internal state and restore the correct behaviour of `invalidateSize`.

Sending this PR in 2 steps:
1. with added test only, to show that it fails in current state.
2. with actual fix, to show that it now passes.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -735,6 +735,7 @@ describe("Map", function () {
 			// Now notifying the map that the container size has changed,
 			// it should return new values and correctly position coordinates
 			map.invalidateSize();
+			
 			expect(map.getSize().x).to.equal(600);
 			expect(map.latLngToContainerPoint(center).x).to.equal(300);
 		});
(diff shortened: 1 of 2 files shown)


ghybs:
Hum not sure why [that test](https://github.com/Leaflet/Leaflet/blob/09fed5d72fb2c592b0359a9d4cdf8dbda406664d/spec/suites/layer/tile/GridLayerSpec.js#L598-L680) failed… given that it passed in [that other PR](https://github.com/Leaflet/Leaflet/pull/6363), it passes on my local environment, and there is no code related to GridLayer in this PR…

PR #6365, opened 2018-10-10
Fix: autoPan doesn't work when popup-content gets updated while a panning-animation is running

### LINKED ISSUE #6327: autoPan doesn't work when popup-content gets updated while a panning-animation is running
## How to reproduce

- Leaflet version I'm using: 1.3.5
- Browser (with version) I'm using: Google Chrome 68.0.3440.106
- step 1 Open a Popup with `autoPan`-Option enabled
- step 2 Update the content of the opened Popup while a panAnimation is running

I expect to see the complete popup content when i call the `setContent` or `update` method on popups. This does not happen when any panning-animations are running because of this line: 

https://github.com/Leaflet/Leaflet/blob/9fda888df2791ca0df6f182a9d91d393ee49034c/src/layer/Popup.js#L238

I've come up with 2 possible solutions for this problem:
1. Any running panning-animations get stopped when `_adjustPan`  gets called:
```javascript
if (!this.options.autoPan) { return; }
this._map._panAnim && this._map._panAnim.stop();
```
2. The new  `_adjustPan`-Request gets queued when there already is an animation running:
```javascript
if (!this.options.autoPan) { return; }
if (this._map._panAnim && this._map._panAnim._inProgress) {
  this._map._panAnim.once('end', this._adjustPan, this);
  return;
}
```
These 2 solutions behave slightly different and I am not sure what solution would be best.
Feedback is appreciated!

- [X] this example is as simple as possible
- [X] this example does not rely on any third party code

**IMPORTANT:** The marker have to get aligned with the top of the map-viewport for the example to work

https://plnkr.co/edit/qPnk16uSxBiOLHHTV2Ke?p=preview

Schleuse:
Fixes #6327

### DIFF
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -326,6 +326,24 @@ describe('Popup', function () {
 				autoPanPadding: L.point(10, 10)
 			});
 		});
+
+		it("should pan map to show popup content if autoPan is enabled even when animating", function (done) {
+			map.on('popupopen', function (e) {
+				var popupTopOffset = getPopupOffset(map, e.popup);
+				expect(popupTopOffset).to.be(10);
+				done();
+			});
+
+			map.panTo([55.8, 40.7], {
+				animate: true,
+				duration: 1
+			});
+
+			map.openPopup('<div style="height: 400px;"></div>', L.latLng(58.4, 37.6), {
+				autoPan: true,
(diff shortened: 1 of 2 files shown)


perliedman:
Hi!

Nice catch and also a nice fix. I think this is the preferred solution of the two you outline in #6327, since it's simpler to understand.

Good work adding tests for these as well. :+1:

PR #6366, opened 2018-10-11
Fix #6335 by specifically checking for correspondingElement

### LINKED ISSUE #6335: JavaScript Error when L.DomUtil.getClass is called on a linked SVG-Element in IE11
## How to reproduce

- Browser (with version) I'm using: Internet Explorer 11
- The JavaScript-Error occurs when you click on a linked SVG-Element in IE11 with a click-listener wich tries to get the class of the clicked target. The target-Attribute is in this case a <code>SVGElementInstance</code>-Object which doesn't have a <code>className</code> property. <code>L.DomUtil.getClass</code> tries to get the <code>baseVal</code>-property and runs into a NullPointer.

We're using Linked SVG-Elements as our Markers. Some plugins we use have a click-listener on the map-container and try check if the Element that got clicked has a specific CSS-class. This doesn't work in IE11 because <code>L.DomUtil.hasClass</code> uses <code>L.DomUtil.getClass</code>. 

It would be nice if <code>L.DomUtil.getClass</code> could handle this, admittingly, weird case. <code>SVGElementInstance</code> has a <code>correspondingElement</code> and <code>correspondingUseElement</code> property which could be used to support this.

Codepen for editing:
https://codepen.io/Schleuse/pen/YOEVeZ

Codepen for testing (Codepen doesn't support IE11 any more in the editing view): 
https://codepen.io/Schleuse/full/YOEVeZ/

Schleuse:
Fixes the issue as I've described in #6335

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -149,6 +149,10 @@ export function setClass(el, name) {
 // @function getClass(el: HTMLElement): String
 // Returns the element's class.
 export function getClass(el) {
+	// Check if the element is an SVGElementInstance and use the correspondingElement instead
+	if (el.correspondingElement) {
+		el = el.correspondingElement;
+	}
 	return el.className.baseVal === undefined ? el.className : el.className.baseVal;
 }
 


perliedman:
LGTM!

Added a little to your comment, since it is often hard to remember why we have all these special cases in the code; might be nice the day we can remove IE11 compatibilty! :wink:

Schleuse:
Nice catch - that comment was definitely needed. 
What a happy day that will be when IE is finally dead! 🍻🎉

PR #6373, opened 2018-10-15
Update plugins.md

bill-chadwick:
Add Leaflet.MetricGrid demo links


bill-chadwick:
Update with demo links as suggested.

Issue #6344, opened 2018-09-21
toGeoJSON precision parameter is not documented


tylkomat:
Please Document the `precision` Parameter for the `toGeoJSON` method.
I took me quite some time to figure out why my calculated area where not same for leaflet layer and geojson layer.

I don't know why this precision even exists. When I convert a layer to geojson I would like to keep the precision as it was originally. Why would I want to loose it?

rkatka:
Addressed in PR: https://github.com/Leaflet/Leaflet/pull/6375

perliedman:
@tylkomat the reason is that the [GeoJSON spec recommends 6 decimal places](https://tools.ietf.org/html/rfc7946#section-11.2), which we default to.

PR #6375, opened 2018-10-17
Added LeafDoc for toGeoJSON's precision - Issue #6344

rkatka:
Added LeafDoc for toGeoJSON's precision argument - Issue #6344


perliedman:
Nice! Thanks!

Issue #5332, opened 2017-02-14
Allow changing the duration of zoom animations


IvanSanchez:
Right now there is a `duration` option for pan animations, and for zoom/pan animations (fitbounds, setview). It would be nice to have a `duration` option also for the zoom animation, e.g.:

`map.zoomIn(1, { duration: 0.5 })`

The duration of the zoom animation is defined in two places: in the CSS definition of `.leaflet-zoom-animated`, and in the private `_animateZoom` method of `L.Map`.

The idea would be to get the duration out of the CSS, and define it dynamically in `_animateZoom` via `this._mapPane.style[DomUtil.TRANSITION] = '...'`, along for the timeout for `_onZoomTransitionEnd`.

hd719:
Hello can I work on this?

perliedman:
@hd719 you're very welcome to do so, as far as I know no one is currently working on this.

hd719:
Hey @perliedman I forked over the repo and I followed the instructions in the `CONTRIBUTING.md` file. When setting up the build system I ran both the commands `npm install -g jake` and `npm install`, but now how do I see the changes in the development environment? Do I have to run a specific command? 

-Thank you.

perliedman:
@hd719 if you run `npm run watch`, the build system will continously monitor the source for changes and rebuild.

hd719:
Hey @perliedman I am having some trouble starting this I found the _animateZoom function in `Map.js` and having trouble moving on. Any advice? 

-Thank you!

shadeland:
Is this still open? I'm going to look into it

perliedman:
@shadeland AFAIK there hasn't been anything done in this area.

It looks like I failed to respond to @hd719's question, but to be honest I'm not too familiar with this part of the code myself, so I will not be able to help with specifics on this issue.

Isabella370:
I have also facing this issue but know You help me a lot There is another method Which may help you If you share the PowerPoint window first before entering Slide Show mode it confuses Zoom and it will stop sharing the window. Because 
this is Slide Show mode, all animations and transitions work as expected. You can still see and access the Zoom controls 
by moving your mouse to the top of the screen. So in this way [PowerPoint animations](https://apkoptions.com/) also work in zoom.

tomasmusil:
Hello, I have developed a solution of my own and just noticed a very similar PR #9020 which already solves this.
Is there any interest in merging this for the v2? I don't want to spam you with (almost) duplicate PRs.

PR #6378, opened 2018-10-20
[WIP] Allow changing the duration of zoom animations #5332

ctyo:
add zoomIn duration options. issue #5332
`map.zoomIn(1, {animate:true, duration : 2000})`

zoom

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -181,6 +181,7 @@ export var Map = Evented.extend({
 		center = this._limitCenter(toLatLng(center), zoom, this.options.maxBounds);
 		options = options || {};
 
+		// FIXME : 具合が悪い
 		if (options.zoom && options.zoom.animate && options.zoom.duration) {
 			options.animate = options.zoom.animate;
 			options.duration = options.zoom.duration;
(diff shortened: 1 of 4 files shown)


src/map/Map.js
```diff
@@ -1636,8 +1654,9 @@ export var Map = Evented.extend({
 			noUpdate: noUpdate
 		});
 
+		var zoomLimit = duration ? duration.match(/([\\.0-9]+)s?$/)[1] * 1000 : 250;
```
IvanSanchez: regexps in the code might be confusing, and the name "zoomLimit" is also confusing. I'd rather use something like `parseFloat` and a name like `zoomEndTime`.

src/map/Map.js
```diff
@@ -181,6 +181,11 @@ export var Map = Evented.extend({
 		center = this._limitCenter(toLatLng(center), zoom, this.options.maxBounds);
 		options = options || {};
 
+		if (options.zoom && options.zoom.animate && options.zoom.duration) {
```
IvanSanchez: This is not how options passed to `setView` work. Just use `options.duration` instead. Am I missing a documented use case here?

src/map/Map.js
```diff
@@ -1625,7 +1630,20 @@ export var Map = Evented.extend({
 			this._animateToCenter = center;
 			this._animateToZoom = zoom;
 
+			if (duration) {
+				if (typeof (duration) === 'number') {
+					duration = duration + 's';
+				}
+				this._zoomStyle = document.createElement('style');
+				this._zoomStyle.innerHTML = '.leaflet-zoom-anim .leaflet-zoom-animated {' +
```
IvanSanchez: Nope nope nope nope. This will mess up things when there are several map instances in the same page zooming at the same time. This should rather be done by setting `this._container.style.transition` (or rather, `this._container.style[DomUtil.TRANSITION]` for compatibility).

Also, with the current implementation new `<style>` elements would be added to the map pane, but never removed (even after their contents are set to an empty string on the timeout).

PR #6380, opened 2018-10-21
Fix/different element for layer switcher

### LINKED ISSUE #6377: Layers switcher uses form and prevents using map inside a form
The layer switcher control adds a form to the html code.
The html specifications forbid to put a form inside another form [html nesting](http://jkorpela.fi/html/nesting.html).
So it is not possible to put a Leaflet map inside a form.
   
  maxx

hundekoerper:
Hi 👋 

I gave #6377 a shot for hacktoberfest. I choose a `section` element instead of a generic `div` because it seemed semantically more appropriate. I hope all is well and in accordance with your contribution guidlines :)

thanks, have a nice day

Resolves #6377


ghybs:
Hi,

As a side note, `<section>` has been introduced in HTML5:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section
https://caniuse.com/#search=section

IE<9 will treat it as an unknown element, still rendering it, but I do not know exactly with what style.

PR #6404, opened 2018-11-15
Fix circles height calculation

thomassuckow:
Circle height was not being calculated for nonstandard crs. Changing the crs resulted in strange behaviours where the height would appear to change as you zoomed.

### DIFF
--- a/src/layer/vector/Circle.js
+++ b/src/layer/vector/Circle.js
@@ -91,10 +91,12 @@ export var Circle = CircleMarker.extend({
 			this._radiusY = p.y - top.y;
 
 		} else {
-			var latlng2 = crs.unproject(crs.project(this._latlng).subtract([this._mRadius, 0]));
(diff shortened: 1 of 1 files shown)


ghybs:
Hi,

Thank you for your first PR here! :+1: 

Please can you link to the issue this PR solves?
If there is none yet, please would you care opening one?

That would make it much clearer what use case this PR addresses.

Thanks!

thomassuckow:
Also note, the circle doesn't appear to be the right size anyway (it is too small). This PR fixes the vertical size of the circle not being calculated, but does not fix the overall size of the circle being wrong with non-standard CRS.

thomassuckow:
Closed because the code this PR modified is fallback code assuming the radius is pixels, while it still should set _radiusY the fallback should not be used anyway.

CRS's should specify a displacement function that given a point, bearing, distance return a new latlng. An improved Circle implementation would then create a polygon using that function that may not look like a circle when near the poles.

Issue #4978, opened 2016-09-29
Circle.getBounds() throw error in 1.0.0 version


atomiomi:
This bug was already fixed in previous version https://github.com/Leaflet/Leaflet/issues/1068. But it occurs in new leaflet version.

Here is fiddle - https://jsfiddle.net/kecmw7ms/5/
Tested in Chrome Version 53, Firefox Version 47, Safari Version 10.0

Notes:
- Adding `circle` to the map makes it work
- Removing `setView` from map breaks it again

IvanSanchez:
This is expected. The bounds will be different depending on the CRS, and a circle itself doesn't have a CRS - it depends on the map to access the CRS and do the proper calculations.

It's possible to throw a more verbose error or add to the documentation, but that's about it.

atomiomi:
@IvanSanchez so it also expected that `getBounds` throws error when view was not set on map and that `getBounds` returns different LatLngBounds depending on current map view?

IvanSanchez:
> so it also expected that `getBounds` throws error when view was not set on map

OK, that *might* be a bug. The calls to `this._map.layerPointToLatLng()` might be able to be changed into `this._map.options.crs.project()`.

rgbkrk:
Possibly related - within our test suite we're running into the bounds being invalid. I've gone ahead and created a repository for reproducing the issue:

https://github.com/rgbkrk/leaflet-issue-reproduction

```bash
git clone https://github.com/rgbkrk/leaflet-issue-reproduction
cd leaflet-issue-reproduction
npm i
node reproduce.js
```

Expect to see:

```js
/Users/kylek/code/src/github.com/rgbkrk/leaflet-issue-reproduction/node_modules/leaflet/dist/leaflet-src.js:1609
    throw new Error('Invalid LatLng object: (' + lat + ', ' + lng + ')');
    ^

Error: Invalid LatLng object: (NaN, NaN)
    at Object.L.LatLng (/Users/kylek/code/src/github.com/rgbkrk/leaflet-issue-reproduction/node_modules/leaflet/dist/leaflet-src.js:1609:9)
    at Object.L.Projection.SphericalMercator.unproject (/Users/kylek/code/src/github.com/rgbkrk/leaflet-issue-reproduction/node_modules/leaflet/dist/leaflet-src.js:2019:10)
[...]

PR #6413, opened 2018-11-23
Fix Error when calling L.Circle.getBounds (#4978)

Schleuse:
I think I've found a solution for #4978. Instead of using the map's projection to determine the bounds, I'm expanding the center LatLng by the Radius of the circle.
The results are pretty close to each other.

I've build a plunkr to demontrate: https://plnkr.co/edit/ZSxogETQ1sVj8l84abrJ?p=preview

### DIFF
--- a/spec/suites/layer/vector/CircleSpec.js
+++ b/spec/suites/layer/vector/CircleSpec.js
@@ -30,6 +30,14 @@ describe('Circle', function () {
 			expect(bounds.getSouthWest()).nearLatLng(new L.LatLng(49.99820, 29.99720));
 			expect(bounds.getNorthEast()).nearLatLng(new L.LatLng(50.00179, 30.00279));
 		});
+
+		it('should return approximate bounds when not added to the map', function () {
+			circle = L.circle([50, 30], {radius: 200});
+
+			var bounds = circle.getBounds();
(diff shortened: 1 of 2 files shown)


ghybs:
Hi,

Thank you a lot for this PR!

Yeah your solution is appealing…

…but looking at how Leaflet implements the [`LatLng.toBounds()`](https://github.com/Leaflet/Leaflet/blob/v1.3.4/src/geo/LatLng.js#L85-L94) method you cleverly use, it seems to me that it actually assumes the CRS to represent the Earth.
This seems to be a leftover from Leaflet 0.x, more than 5 years ago… (https://github.com/Leaflet/Leaflet/commit/5558987272e2776b6fb4761af7a790df9a22c00d)

Therefore this fix would only be suitable when the map uses a CRS close to the assumptions made in this method computation, and I suspect it would break e.g. if using CRS.Simple.

Schleuse:
Hmm... good point. One solution for this problem could be to move the logic of `LatLng.toBounds()` to the `L.CRS.Earth` (like the other functions in `L.LatLng`) and add a reference to a CRS in `L.LatLng` defaulting to `L.CRS.Earth`.

I've played around with this idea a bit and made a branch for testing: https://github.com/Schleuse/Leaflet/tree/add-crs-to-latlng

I've added a new optional argument to the constructor of `L.LatLng` to be able to pass a CRS. I'm not really sure this would be viable solution at all - feedback is appreciated ;D

PR #6440, opened 2018-12-20
Add Leaflet.Streetview plugin

adriankwiat:


adriankwiat:
I will make a new plugin

PR #6450, opened 2018-12-30
Update rotate branch to latest Leaflet release

Keats:
I initially did the PR on https://github.com/fnicollet/Leaflet/pull/32 so it's probably why it cannot be automatically merged, I'll fix that if this change is acceptable.

As mentioned in the original PR:
It seems to mostly work fine except with touch gestures on iOS where it sometimes go crazy. I didn't manage to reproduce it reliably sadly. A bad quality gif showing the issue: https://imgur.com/PtcOi1o

There is also an issue with polygons where they move a lot while using gestures but end up repositioning themselves correctly at the end, you can see the debug/rotate-mobile template and is the same thing happening as in https://github.com/va2ron1/Leaflet/issues/16


johnd0e:
@Keats 
Thanks for trying, but you see - now it's just impossible to review (200+ commits).
To correctly update `rotate` branch it would be better to [rebase](https://git-scm.com/docs/git-rebase) it on `master`.

If you do not have write access to leaflet repo - just create new branch in own fork, and make new PR against `master`.

Issue #6457, opened 2019-01-04
Implement getIcon() for Markers


OsamaShabrez:
I am trying to optimize my application to avoid unnecessary redraws. I always get a collection of markers from my database and then I have to do a comparison to see if I need an update.

It would be cool if we can have a getIcon / getIconUrl / getCustomIconUrl function which will return the URL to the marker image which can be then used to identify if a marker needs to be redrawn.

For instance, in my application, I have a different icon for currently selected marker but it appears there is no way I can check what is the current icon and then only update a single marker rather than removing all markers and recreating them from scratch.

PR #6459, opened 2019-01-04
Add getIcon function to retrieve marker icon - Issue #6457

OsamaShabrez:
As suggested in #6457.


OsamaShabrez:
I would need help on how I can update the docs to reflect this change?

PR #6478, opened 2019-01-16
doc: Elliptical Mercator projection - Earth is an ellipsoid

jjimenezshaw:
Minor change in documentation:
Elliptical Mercator projection (EPSG:3395) assumes that the Earth is an ellipsoid, not a geoid (in comparison with Spherical Mercator (EPSG:3857), that assumes the Earth is a perfect sphere).

https://sciencing.com/difference-between-geoid-ellipsoid-8638149.html

### DIFF
--- a/src/geo/projection/Projection.Mercator.js
+++ b/src/geo/projection/Projection.Mercator.js
@@ -6,7 +6,7 @@ import {Point} from '../../geometry/Point';
  * @namespace Projection
  * @projection L.Projection.Mercator
  *
- * Elliptical Mercator projection — more complex than Spherical Mercator. Takes into account that Earth is an ellipsoid, not a perfect sphere. Used by the EPSG:3395 CRS.
+ * Elliptical Mercator projection — more complex than Spherical Mercator. Assumes that Earth is an ellipsoid. Used by the EPSG:3395 CRS.
  */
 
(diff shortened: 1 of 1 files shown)


src/geo/projection/Projection.Mercator.js
```diff
@@ -6,7 +6,7 @@ import {Point} from '../../geometry/Point';
  * @namespace Projection
  * @projection L.Projection.Mercator
  *
- * Elliptical Mercator projection — more complex than Spherical Mercator. Takes into account that Earth is a geoid, not a perfect sphere. Used by the EPSG:3395 CRS.
+ * Elliptical Mercator projection — more complex than Spherical Mercator. Takes into account that Earth is an ellipsoid, not a perfect sphere. Used by the EPSG:3395 CRS.
```
mourner:
```suggestion
 * Elliptical Mercator projection — more complex than Spherical Mercator. Assumes that Earth is an ellipsoid. Used by the EPSG:3395 CRS.
```

jjimenezshaw: Sounds good.

jjimenezshaw:
@mourner I have applied your suggestion, but manually. Should I do anything else?

PR #6488, opened 2019-01-22
Explicitly remove controls from the map on map unload

### LINKED ISSUE #6487: control.onRemove isn't called when map.remove() is called

**The issue**

I am currently trying to track down a memory leak in my application. One potential place I've found is that `control.onRemove` isn't called when `map.remove()` is called.

https://leafletjs.com/reference-1.4.0.html#control

I'm doing stuff in `control.onAdd` that needs tearing down when the map is removed, but `control.onRemove` isn't called when `map.remove()` is called.

In order to help prevent memory leaks, it would be useful if `control.onRemove` were called when the map is removed.

**Environment**
- Leaflet version: 1.4.0
- Browser (with version): all
- OS/Platform (with version): all

https://jsbin.com/qofetabumu/1/edit?html,js,output **warning: resource intensive**

the memory usage of the page goes up by about 50 MB a second, and because it's being used by an event listener that isn't removed, it can't be garbage collected.

this is a totally unrealistic example but my actual case is super complicated and i'm not sure how to simplify it

----

I'd be willing to have a crack at fixing this but would need pointing in the right direction as I haven't contributed to leaflet before :)

IvanSanchez:
This is a blind attempt (i.e. *untested*) to fix #6487. A map being destroyed will fire the `unload` event, and any controls from that map can use that event to remove themselves from the map.

### DIFF
--- a/src/control/Control.js
+++ b/src/control/Control.js
@@ -79,6 +79,8 @@ export var Control = Class.extend({
 			corner.appendChild(container);
 		}
 
+		this._map.on('unload', this.remove, this);
+
 		return this;
 	},
 
@@ -95,6 +97,7 @@ export var Control = Class.extend({
 			this.onRemove(this._map);
 		}
 
+		this._map.off('unload', this.remove, this);
 		this._map = null;
 
 		return this;


callumacrae:
Sweet, `onRemove` is being called now 👍

callumacrae:
It looks like layers have the same problem - should the same code be added there too?

IvanSanchez:
> Sweet, `onRemove` is being called now :+1:

I'm more worried about the memory leaks - does this help?

> It looks like layers have the same problem (untested, just looked at the code) - should the same code be added there too?

No need, because

https://github.com/Leaflet/Leaflet/blob/8108ff9f073bbdfa360de8247c39858720418d37/src/map/Map.js#L797-L799

callumacrae:
> I'm more worried about the memory leaks - does this help?

It helps in the example I posted, it doesn't help (measurably) in the actual application I'm working on - which probably just means the problem is somewhere else, probably outside of leaflet. I'll carry on investigating and keep you posted if I find anything else!

IvanSanchez:
> It helps in the example I posted, 

That's good enough for me! :smile:

callumacrae:
Thanks for the quick fix!

Issue #6489, opened 2019-01-23
getZoomScale returns NaN, breaks zooming, when zoom level >= crs.scales.length


scaddenp:
Leaflet 1.4, (and earlier versions), any browser.
Problem occurs in getZoomScale when toZoom or fromZoom is larger than crs.scale.length-1. NaN returned and rendering goes south.

Firstly there is an obvious fix/workaround in client code using Leaflet- set the options.maxZoom on map to the crs.scale.length - 1. I have done this (checking crs particularly when fetching the crs data from a remote server like ArcGIS).

However, default leaflet behavior is to set maxZoom to largest of zooms on any tile/grid sets not smallest. This mechanism doesnt seem to use any limits which crs.scale might impose. Adding a check to the crs in _updateZoomLevels would be an option.

Alternatively a fix could be check in crs.scale getZoomscale and effectively limit the zoom level from that result. (I think returning a scale of 0.5 would work but havent tried it).

IvanSanchez:
Can you publish a fiddle/codepen/[plunkr](https://leafletjs.com/edit) that displays this behaviour?

scaddenp:
I guess I could but it seems a bit obvious. In getZoomScale code you have
```
getZoomScale: function (toZoom, fromZoom) {
		// TODO replace with universal implementation after refactoring projections
		var crs = this.options.crs;
		fromZoom = fromZoom === undefined ? this._zoom : fromZoom;
		return crs.scale(toZoom) / crs.scale(fromZoom);
	},
```

and crs.scale(toZoom) is obviously going to fail to if toZoom > crs.scale.length.

scaddenp:
Plunkr https://plnkr.co/edit/qIdL3erde8B1v7MXcqRG?p=preview

Zoom it in two more levels and it breaks. The crs has 12 scales (derived from the resolutions) and so zooming to 12 will kill. Cant unzoom either once it breaks.

IvanSanchez:
I'm seeing your plunkr and I'm happy to see that zooming isn't entirely "broken". I can still zoom back.

I think the easiest fix would be for the CRS to override the map's `minZoom` and `maxZoom` options. I'll have a quick look at Proj4Leaflet's code.

PR #6491, opened 2019-01-25
Add minZoom and maxZoom properties to L.CRS as well

IvanSanchez:
A blind approach to #6489: Let `L.CRS` instances to have `minZoom` and `maxZoom` properties, and check them in `getMinZoom()`/`getMaxZoom()`.

Increase readability of `getMinZoom()` by using `Math.max()` and infinite values for the defaults of `map.options.minZoom` and `_layersMinZoom` (idem for `max)`.


IvanSanchez:
Weird. Unit tests for `baselayerchange` fail on CI, but work locally :confused:

PR #6499, opened 2019-01-30
Adding Leaflet.Multispectral to plugins list

jywarren:
From https://github.com/publiclab/leaflet-multispectral, released today! 

Demo: https://publiclab.github.io/leaflet-multispectral/

Uses https://github.com/publiclab/image-sequencer

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -623,6 +623,15 @@ The following plugins change the way that tile or image layers are displayed in 
 			<a href="https://github.com/valkenburg">Wessel Valkenburg</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/publiclab/leaflet-multispectral">Leaflet.Multispectral</a>
+		</td><td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Seems nice, but I guess it only works for `ImageOverlay`s, right? There's some overlap with `TileLayer.GL`, which can also perform fancy raster manipulations. :thinking:

jywarren:
Yeah! I actually have a second library specifically for tiles coming next,
I had separated them to keep things modular. Was about to publish it too!
They're quite similar bc of the image processing engine.

This one will be interesting also because the processing for a large area
will be offloadable to a serverless API with an upcoming release of image
sequencer.

IvanSanchez:
Do check your demo, as the buttons hide behind the basemap. And think about using WebGL - performance will be way better that way.

jywarren:
Yeah - we have an issue open on this but part of the spec of Image
Sequencer is that it be runnable in nodejs identically to browser, so we
don't use canvas at all. There are a few attempts to implement webgl on
nodejs, and we're looking at those, but webassembly may also be a big speed
improvement, so we're exploring both possibilities. Either of these would
appear in this lib via upstream once an implementation is released in Image
Sequencer. Thanks!

IvanSanchez:
See https://www.npmjs.com/package/gl - it works. And don't waste much time with WebAssembly: it might speed things up by 2x or 3x, but I've seen WebGL speed things up to 1000x. Really.

PR #6505, opened 2019-02-05
Add ceilZoom to GridLayer to select zoom level used between integers

ij1:
By default, a fractional zoom level is rounded, that is, the tiles switch zoom level at halfway point which creates undesirable >1 scale when a canvas is used within a GridLayer tile. This change allows avoiding >1 scales by setting ceilZoom to true that makes the tile zoom level to be always larger than a non-integer zoom level.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -128,6 +128,12 @@ export var GridLayer = Layer.extend({
 		// from `minNativeZoom` level and auto-scaled.
 		minNativeZoom: undefined,
 
+		// @option ceilZoom: boolean = false
+		// If set, tiles with fractional zoom level will be selected using ceil
+		// instead of rounding. It can be used to prevent larger than 1 scaling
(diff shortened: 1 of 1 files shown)


johnd0e:
> which creates undesirable >1 scale when a canvas is used within a GridLayer tile.

Is there some issue reported here?
Or may be you could post some sample to illustrate the issue?

ij1:
> > which creates undesirable >1 scale when a canvas is used within a GridLayer tile.
> 
> Is there some issue reported here?
> Or may be you could post some sample to illustrate the issue?

Whenever scale is >1, the tile no longer looks "accurate" to eye as its pixels are larger than the pixels on the screen.

johnd0e:
Well, this may be useful.
But what if we make this in more robust way?
I'd prefer more smart action instead of just 'floor-or-ceil'.
E.g. round it.
Or even more - set separate threshold values for zoom-in/out.

PR #6507, opened 2019-02-10
Make sure geojson polygon objects have at least 4 coordinates

### LINKED ISSUE #6494: geoJSON layer accepts invalid polygon object
**Expected behavior**
A geojson polygon feature with less than 4 coordinates must raise error.

**Current behavior**
Leaflet accepts such feature as a valid geojson object.

**Environment**
- Leaflet version: 1.4.0
- Browser (with version): N/A
- OS/Platform (with version): N/A

**Additional context**
A geojson polygon feature must have at least four coordinates, and the first and last items in the coordinates list must be the same. So the following object is considered invalid:
```json
{
  "type": "Polygon",
  "coordinates": [[[-0.5, -0.5], [0.5, 0.5], [-0.5, -0.5]]]
}
```
But leaflet treats it as a valid object.

See `scripts.js` in the this example: https://plnkr.co/edit/IpLcBvzV2D1akpzlKiA5?p=preview

ka7eh:
Resolves #6494

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -193,7 +193,7 @@ export function geometryToLayer(geojson, options) {
 
 	case 'Polygon':
 	case 'MultiPolygon':
-		latlngs = coordsToLatLngs(coords, geometry.type === 'Polygon' ? 1 : 2, _coordsToLatLng);
+		latlngs = coordsToLatLngs(coords, geometry.type === 'Polygon' ? 1 : 2, _coordsToLatLng, 4);
 		return new Polygon(latlngs, options);
 
 	case 'GeometryCollection':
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm actually **against** this change, because I like to follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle):

> Be conservative in what you do, be liberal in what you accept from others.

If anything, Leaflet should make sure that the [`L.Polygon.toGeoJSON` implementation](https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/layer/GeoJSON.js#L340) returns sanitized GeoJSON (i.e. making sure that the first and last pair of coordinates are the same).

ka7eh:
@IvanSanchez I see your point and it makes sense. The reason I submitted this here was because of some invalid geometry issues originated in some leaflet plugins (e.g. `leaflet.pm`) and I was thinking of preventing such errors in Leaflet itself. But I see why it's better to stop the issue at its source instead of here.

IvanSanchez:
Yeah, fixing Leaflet.pm and Leaflet.Edit so they output well-formed GeoJSON is the preferred way.

PR #6509, opened 2019-02-11
Use const instead of magical number for earth R

johnd0e:
This change is [Reviewable](https://reviewable.io/reviews/leaflet/leaflet/6509)


IvanSanchez:
This breaks the unit tests. Please review the results, and run the tests locally.

johnd0e:
Fixed, thanks.

PR #6522, opened 2019-02-20
Fix GridLayer occassionally leaving tiles from two layers active

ij1:
If GridLayer finds a tile that is not current but is still active, set a flag to do a prune if at the end of the fade.

I suspect the original problem might be due to a race between _updateOpacity and timed prune from _tileReady. As it seems cheap to calculate when a clearing prune is needed at the end of the fade, just do it.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -313,11 +313,15 @@ export var GridLayer = Layer.extend({
 
 		var now = +new Date(),
 		    nextFrame = false,
-		    willPrune = false;
+		    willPrune = false,
+		    endPrune = false;
 
 		for (var key in this._tiles) {
 			var tile = this._tiles[key];
-			if (!tile.current || !tile.loaded) { continue; }
+			if (!tile.current || !tile.loaded) {
+			  if (tile.active) { endPrune = true; }
+			  continue;
+			}
 
 			var fade = Math.min(1, (now - tile.loaded) / 200);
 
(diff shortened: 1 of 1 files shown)


johnd0e:
> I suspect the original problem might be due to [...]

What is mentioned 'original problem'?
Is there some issue reported?

ij1:
> > I suspect the original problem might be due to [...]
> 
> What is mentioned 'original problem'?
> Is there some issue reported?

"The original problem" just referred to what's in the title, that is, gridlayer leaves tiles from two different zoom levels active. The one which has a large scale() looks very blurry and mixes with the tiles that are for the currently correct zoom level. To me it looked as if the pruning is somehow timer dependent and failed to do its work if timer fired early enough but I'm no expert on how it is supposed to work.

johnd0e:
@ij1 
> "The original problem" just referred to what's in the title, that is, gridlayer leaves tiles from two different zoom levels active.

Do you have sample code to reproduce original problem?
We need to check if it's gone after #7123 (not sure if it's related without sample code)

joshuahhh:
I've run into this problem myself, running Leaflet 1.7.1. It's hard to reproduce – it only happens sometimes, and only in a complex setup using [Leaflet.PixiOverlay](https://github.com/manubb/Leaflet.PixiOverlay).

Here's a GIF of the problem, in Leaflet 1.7.1. Labels are on a layer above the background. Note that the first zoom works out ok, but not the second.
map2

I will see if I can merge & test this change.

joshuahhh:
I just tested this change by manually editing leaflet-src.js in my node_modules folder. It seems to completely eliminate the problem. If it's necessary, I can try to find a minimal reproduction of the original problem, but, as I mentioned, that may be difficult.

PR #6530, opened 2019-02-25
Fix minNativeZoom causing browser crash

tagorecaue:
Fixed browser crash when user set minNativeZoom in a tilelayer.

When the mapZoom is 1 and minNativeZoom is 17, for example, the funcion _getTiledPixelBounds result in a huge bounds due to scale, then the function _update try to load a big tileRange, ocasioning browser crash.

What I did was check if tile has bounds and just try to load tiles inside this bounds.

It improves a lot the performance of tile loading

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -631,7 +631,20 @@ export var GridLayer = Layer.extend({
 		    pixelCenter = map.project(center, this._tileZoom).floor(),
 		    halfSize = map.getSize().divideBy(scale * 2);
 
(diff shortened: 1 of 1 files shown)


tagorecaue:
Debug example.
browser crashes

tagorecaue:
@IvanSanchez do you think if it can be reviewed soon?

cherniavskii:
@tagorecaue can you publish an example using https://leafletjs.com/edit.html or any other jsfiddle-like site?

johnd0e:
@tagorecaue?

PR #6534, opened 2019-02-28
Added GeoRasterLayer Plugin

DanielJDufour:
First time adding a plugin to the site.  Let me know if I did something wrong or have any questions.  Thank you for your consideration :-)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -368,6 +368,15 @@ Plugins for loading basemaps or GIS raster layers in common (albeit non-default)
 	</tr>
 	<tr>
 		<td>
+			<a href="https://github.com/geotiff/georaster-layer-for-leaflet">GeoRasterLayer</a>
+		</td><td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
If the plugin is meant to load GeoTIFFs, why is it **not** called GeoTIFFsomethingsomething? :thinking:

DanielJDufour:
@IvanSanchez  , excellent question! The package used to be called geotiff-layer-for-leaflet (http://npmjs.com/package/geotiff-layer-for-leaflet).  However, we renamed it because we are in the process of adding NetCDF (CF Convention) support and also support for geo-referenced JPG and PNG files.  I felt it premature to mention these other file formats though because support for them is still in the works.  Open to your thoughts! =)

jmbajo:
Excellent Plugin, I hope to see it anytime soon.

Rennzie:
👍🏻 🚀

IvanSanchez (approved):

PR #6543, opened 2019-03-05
Round point before CSS translate is set

mdorda:
related to #3575

Some additional testing is required. We do not use scaling, not even decimal zooming. But this commit definitely solved pixel gap issue for all our cases.


johnd0e:
> Some additional testing is required.

Definitely.
Fractional values are important when [`devicePixelRatio`](https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio) > 1.

PR #6544, opened 2019-03-05
Use translate instead of translate3d

mdorda:
related to #3575
do not merge together with #6543

This change should not affect performance. As you can see [here](https://jsperf.com/translate3d-vs-xy), modern browsers use GPU acceleration for both properties.


johnd0e:
> do not merge together with #6543

So which approach should be preferred?

johnd0e:
Compare to #6234

mdorda:
Actually I do not remember which change fixed the issue. We are using our own modified leaflet library due to long responses to issues and merge requests. It would be better to not merge this.

PR #6550, opened 2019-03-11
make popup closable by pressing ESC-key

Schleuse:
Add's the ability to close a open popup by pressing the <kbd title="Escape">Esc</kbd>-Key

### DIFF
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -295,6 +295,19 @@ describe('Popup', function () {
 		expect(map.hasLayer(layer._popup)).to.be(true);
 	});
 
+	it("can close a open popup with esc keypress when marker has focus", function () {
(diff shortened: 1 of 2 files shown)


IvanSanchez:
...but the map's keyboard handler already does this, at 

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/map/handler/Map.Keyboard.js#L163-L164

Does this PR address any case not covered by that? Or, does this PR make that bit of the keyboard handler obsolete?

Schleuse:
Oops! I didn't notice that code. I've analyzed why this feature doesn't work in my use-case and I am going to open a new issue for that.

Thanks for the quick feedback @IvanSanchez 🥇

Schleuse:
See #6551

PR #6559, opened 2019-03-15
Remove double background definition

laoneo:
For the control attribution container are two background CSS commands defined. Better to use the white one so we meet the requirements of WCAG 2.0 AA for the contrast between the link and it's background which is inherited from the container.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -397,7 +397,6 @@
 
 .leaflet-container .leaflet-control-attribution {
 	background: #fff;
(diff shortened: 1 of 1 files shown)


mourner:
This is on purpose — for old IE browsers that don't support `rgba` to have a readable attribution control. At the same time, I don't want to make it fully white — de-emphasizing it so that it's not too distracting from the map is a design decision.

laoneo:
So you want to support IE < 9 and not be WCAG 2.0 AA compatible?

PR #6571, opened 2019-03-22
Make DivIcon html option accept Element

oscar-sensornet:
Adds the possibility to use a DOM `Element` instance as the `html` option for a `DivIcon`.

### DIFF
--- a/src/layer/marker/DivIcon.js
+++ b/src/layer/marker/DivIcon.js
@@ -44,7 +44,12 @@ export var DivIcon = Icon.extend({
 		var div = (oldIcon && oldIcon.tagName === 'DIV') ? oldIcon : document.createElement('div'),
 		    options = this.options;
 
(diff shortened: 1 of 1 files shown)


src/layer/marker/DivIcon.js
```diff
@@ -44,7 +44,12 @@ export var DivIcon = Icon.extend({
 		var div = (oldIcon && oldIcon.tagName === 'DIV') ? oldIcon : document.createElement('div'),
 		    options = this.options;
 
-		div.innerHTML = options.html !== false ? options.html : '';
+		if (options.html instanceof Element) {
+			div.innerHTML = '';
```
IvanSanchez: I'd rather see `L.DomUtil.empty()` in use here.

oscar-sensornet: Sure. No problem.

Issue #6572, opened 2019-03-23
Chrome on Android events propagation after pinch zoom


7Ds7:
- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
- Open https://codepen.io/binaryunit/pen/OqdbyP 
- Click the image and pinch to zoom it
- Close the modal
- Pinch zoom on map
- Click image again and try to pinch zoom it again it does NOT receive any events

**Expected behavior**
Pinch Zoom / touch events should work on other elements after pinch zooming on the leaflet map

**Current behavior**
Touch Events are not properly propagated on other elements once you zoom on the map

**Environment**
- Leaflet version: 1.3.2
- Browser (with version): Chrome
- OS/Platform (with version): Android

**Additional context**
I am using other frameworks and components but i isolated them and reproduce the issue here
https://codepen.io/binaryunit/pen/OqdbyP it was very tricky to understand what was going on, but at least now i can reproduce it flawlessly every time.
Please note that this only happens on Chrome/Mobile and not any other Desktop/Mobile browser that i had tried (FF, Safari)

- [x] this example is as simple as possible
- [ ] this example does not rely on any third party code 
(i am using [viewerjs](https://github.com/fengyuanchen/viewerjs) but since it only happens after on pinch zooming on leaflet map i am assuming it is something that leaflet does to the global events after pinch zooming, keep in mind that pan or any other interactions do not cause this behaviour.

I looked into the several options as well as repo issues namely [#5310](https://github.com/Leaflet/Leaflet/issues/5310) which seem to be the closest behaviour but that was not it, any directions to follow or solutions would be greatly appreciated.

7Ds7:
After some investigation looking at the touch event handler
https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.TouchZoom.js#L29

if we change the on off methods to the map container instead of document it seems that it fixes the issue, not so sure what would be the implications though.

```
	_onTouchStart: function (e) {
		...

		DomEvent.on(this._map._container, 'touchmove', this._onTouchMove, this);
		DomEvent.on(this._map._container, 'touchend', this._onTouchEnd, this);

...
	},

	_onTouchMove: function (e) {
[...]

PR #6578, opened 2019-03-25
#6572 touch zoom events only to container instead of document

7Ds7:


7Ds7:
This although disables the possibility of touch events on any other element than the map element itself

johnd0e:
> although disables the possibility of touch events on any other element than the map element itself.

Do you mean that you still have some real issue with current code?

johnd0e:
> we want map panning or pinch zooming to continue if one of the fingers went off the map.

Perhaps for most browsers we can use [setPointerCapture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) for that purpose.

johnd0e:
Compare on and off:
https://github.com/Leaflet/Leaflet/blob/e213469a296ea00e6f9e6e6beb49d081e8d1c4fe/src/map/handler/Map.TouchZoom.js#L61-L62
https://github.com/Leaflet/Leaflet/blob/e213469a296ea00e6f9e6e6beb49d081e8d1c4fe/src/map/handler/Map.TouchZoom.js#L115-L116

I suspect that there 'off' misses a context:

```js
		DomEvent.off(document, 'touchmove', this._onTouchMove, this);
		DomEvent.off(document, 'touchend', this._onTouchEnd, this);
```

PR #6583, opened 2019-03-29
Added WP Mapit

phpwebdev11:
Added a new wordpress plugin WP MapIt which uses Leaflet and Open street maps to display maps on the wordpress website.


phpwebdev11:
Thank you.

PR #6598, opened 2019-04-14
Adding my plugin

iboates:
It looks like maybe this file actually has some broken HTML on the original, unless I just can't figure out how to properly read it. I added my plugin to the "Map Events" section, since it seems like the best fit. It is a plugin which makes points get dynamically highlighted near the user's mouse cursor. Link: https://github.com/iboates/leaflet-spotlight

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -3522,7 +3522,7 @@ These plugins extend Leaflet event handling.
 		</td><td>
 			Dynamically highlight features near the mouse cursor with a customizable shape
 		</td><td>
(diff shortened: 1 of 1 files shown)


iboates:
@IvanSanchez I suggest you roll it back so I can try it again, the page formatting has definitely broken:

https://leafletjs.com/plugins.html (scroll to Events)

IvanSanchez:
Oh, there was a missing `<`

IvanSanchez:
Fixed in https://github.com/Leaflet/Leaflet/commit/d11050974bda04109a526ea0cbb7f15d4c166c14 .

PR #6613, opened 2019-04-19
Layer: refactor openPopup/openTooltip: use common function

johnd0e:
`Layer`: refactor `openPopup`/`openTooltip`: use common function `DivOverlay._openOverlay`.

---
This change is [Reviewable](https://reviewable.io/reviews/leaflet/leaflet/6613)

### DIFF
--- a/src/layer/DivOverlay.js
+++ b/src/layer/DivOverlay.js
@@ -1,4 +1,5 @@
 import {Layer} from './Layer';
+import {FeatureGroup} from './FeatureGroup';
 import * as Util from '../core/Util';
 import {toLatLng} from '../geo/LatLng';
 import {toPoint} from '../geometry/Point';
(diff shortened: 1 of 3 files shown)


johnd0e:
Perhaps better name function differently, like `_initOverlay`.

johnd0e:
> maybe `_updateOverlay` would reflect what the code is doing best?

There are already several methods prefixed with `update`, so we need reflect function specific somehow 

Besides, I do not feel that `update` is right word, as this code may be invoked _before_ tooltip opened. It is a sort of preparatory step: here we determine what layer will be set as source, and calculate target latlng.

So may be `_preOpen`?

mourner:
> It is a sort of preparatory step

`_prepareOpen` then? or `_prepareOverlay`

johnd0e:
> `_prepareOpen` then? or `_prepareOverlay`

Ok, PR updated.

I've added one more commit, please check if it's acceptable.

PR #6614, opened 2019-04-19
Minor: Background color CSS declaration

milesflo:
For the sake of compatibility, it's best to make as few assumptions about the browser as possible.

On the current docs website, it's assumed that the background will be `white`. As a result of my distro's dark theme, that's not the case... The website looks like this on my machine:
image

With one small declaration, the site will look as intended on any browser:
image


mourner:
Good catch, thanks!

Issue #3690, opened 2015-07-29
z-index change breaks jQueryUI use


gaelicmichael:
I've been using Leaflet 0.7.3 very successfully with jQueryUI. It is clear that the z-index values or logic has changed in 1.beta, as my own divs that were once working are now overlaid by Leaflet and jQueryUI modal dialog boxes are hidden underneath the map pane.

mourner:
cc @patrickarlt

mourner:
Yeah, the z-index values were bumped. Can you set up a JSFiddle to demonstrate the problem?

Eschon:
I don't know about jQueryUI, but I had some divs over my map.
After updating to 1.0 beta 1 the were below the map so I just increased the z-index on my divs and it worked fine.

gaelicmichael:
It's obvious how to modify my own CSS to give things a great z-index, but the issue is compatibility with a large library like jQueryUI, which had no previous problems. It's not reasonable to have to modify the jQueryUI code itself to enable inter-operability, though I haven't looked yet for means of specifying the use of z-index values for modals and other widgets -- I'm going to look this morning.

Regardless, the usage and values of z-index in Leaflet needs to be well documented in order for developers to anticipate such issues.

gaelicmichael:
This entry may help those with this problem: http://stackoverflow.com/questions/16917430/jquery-ui-1-10-dialog-and-zindex-option

mourner:
So, any chance to see a JSFiddle?

jaymmartin:
See fiddle here: https://jsfiddle.net/onwu4kv6/

Seems like by default jQueryUI dialog z-index is 100 and default map pane has a z-index of 100.

IvanSanchez:
This can be worked around very easily by manually specifying a lower `z-index` for the map.

The underlying problem is that the map panes share the same [stacking context](https://developer.mozilla.org/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) as the siblings of the map container. So it's not the z-index of the map container, it's the z-index of the panes.

[...]

PR #6634, opened 2019-05-03
CSS z-index bug

ilblog:
Unfortunately a bug in Chrome browser does not respect CSS stacking context regarding z-index. Therefore tile-pane with 400 z-index overlays the stuff, that is even outside map container.

The patch is applied on Windy.com and works flawlessly more than 1 month

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -94,8 +94,6 @@
 	-moz-user-select: none;
 	}
 
-.leaflet-pane         { z-index: 400; }
-
 .leaflet-tile-pane    { z-index: 200; }
 .leaflet-overlay-pane { z-index: 400; }
 .leaflet-shadow-pane  { z-index: 500; }


IvanSanchez:
> Unfortunately a bug in Chrome browser

Which one?

---

Looks related to https://github.com/Leaflet/Leaflet/issues/3690

johnd0e:
@ilblog?

I've re-added `.leaflet-pane         { z-index: 400; }` back on Windy.com haven't observed any difference.
Please provide some sample to reproduce the issue.

Malvoz:
https://github.com/Leaflet/Leaflet/pull/6634#issuecomment-615312862:

> I've re-added `.leaflet-pane { z-index: 400; }` back on Windy.com haven't observed any difference.

If I re-add the styles I see that all the controls disappear.

https://github.com/Leaflet/Leaflet/pull/6634#issuecomment-488979965:

> Looks related to https://github.com/Leaflet/Leaflet/issues/3690

It looks like this indeed would fix that issue. It's not entirely clear to me what the implications would be if `z-index` was removed for `leaflet-pane` as this PR suggests though. Could that introduce other problems? `isolation: isolate` as suggested in https://github.com/Leaflet/Leaflet/issues/3690#issuecomment-229584356 would also fix the issue described here for Windy.com.

PR #6651, opened 2019-05-09
Enhance Popup/Tooltip content function handling

johnd0e:
Allow return undefined (in this case Popup/Tooltip is not shown)

Note: _This adds single commit to already reviewed #6639_

---
As said in docs to [`FeatureGroup`](https://leafletjs.com/reference-1.5.0.html#featuregroup) (and thus `GeoJSON`, as it extends `FeatureGroup`), _`bindPopup` binds a popup **to all of the layers at once** (likewise with `bindTooltip`)_.

And what is implied (but not said explicitly) - function **must** return value for every layer (otherwise we get exception).

But in my practice it's more common when only **some** of layers need Popup/Tooltip.
This PR adds such possibility.

### DIFF
--- a/src/layer/DivOverlay.js
+++ b/src/layer/DivOverlay.js
@@ -259,5 +259,12 @@ Map.include({
 		}
 
 		return this;
+	},
+
+	_closeOverlay: function (overlay) {
+		if (overlay) {
+			this.removeLayer(overlay);
+		}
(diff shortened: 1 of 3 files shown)


johnd0e:
@mourner 
Should new behavior to be documented explicitly?

PR #6661, opened 2019-05-12
Add files via upload

harrishudson:
Please consider adding this to the plugins page

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,4 +1,4 @@
----
+﻿---
 layout: v2
 title: Plugins
 bodyclass: plugins-page
@@ -1752,6 +1752,19 @@ These plugins provide new markers or news ways of converting abstract data into 
 			<a href="https://github.com/iDerekLi/">Derek Li</a>
 		</td>
 	</tr>
+	<tr>
+		<td>
+			<a href="https://github.com/harrishudson/Leaflet.zoomSVG">Leaflet.zoomSVG</a>
+		</td>
+		<td>
+			A Leaflet SVG Renderer that does not re-project or redraw features 
+upon map zoom change. No external dependencies.
+		</td>
+		<td>
+			<a href="https://github.com/harrishudson/">Harris Hudson</a>
+		</td>
+	</tr>
+
 </table>
 
 


docs/plugins.md
```diff
@@ -1,4 +1,4 @@
----
+﻿---
```
ghybs: Please what is this change for? It might break the page metadata?

harrishudson: @ghybs thanks for the review.  That particular change you highlighted was not intentional (I don't have any client side source control installed and it may have been an artifact when I did file upload).  Thanks for any correction.

IvanSanchez:
I'm rejecting this for the time being - the file header might break things, and the plugin doesn't do anything that can't be achieved by setting the map's `zoomAnimation` option to false.

harrishudson:
Thanks for the update and reviewing this. My apologies - the file header glitch was an unintended artifact in the file when I did the file upload. I simply haven’t had the time to rectify that. The intention of this plugin was to allow the efficient rendering of very large geometry layers by not redrawing SVG paths upon zoom level change and just let the browser’s native SVG renderer perform the scaling. This is all now possible thanks to the SVG vector-effect “non-scaling-stroke” property.  My further findings were that the initial projection of very large layers was still problematic however - so this plugin didn’t perform completely as I hoped. I have sort of abandoned working on it for now. I also had some bugs in that it wasn’t drawing markers correctly. However this wasn’t an attempt to simply disable zoomAnimation but rather an attempt to utilise the browser native SVG scaling capabilities and make the rendering of geometries more efficient. Hopefully some future incarnation will be useful. The use case I was trying to solve relates to displaying extremely large geometry layers. Thanks for your review.

IvanSanchez:
> The intention of this plugin was to allow the efficient rendering of very large geometry layers by not redrawing SVG paths upon zoom level change and just let the browser’s native SVG renderer perform the scaling. 

Keep in mind that Leaflet redraws SVG geometries because of a *performance optimization*: it *clips* geometries which are outside the visible part of the map. Skipping that optimization might be counter-productive.

Your idea is interesting, though. Maybe feature detection of SVG2 plus `vector-effect: non-scaling-stroke` plus handling large values of [`L.Renderer`'s `padding` option](https://leafletjs.com/reference#renderer-padding) could lead to fewer SVG redraws. However, most use cases for performance in vector datasets are covered already with VectorGrid, so I'm doubtful of the gains of this approach :-/

harrishudson:
Thanks.  I am not that familiar with VectorGrid - I will take a look.  
However I suspect that may not be useful in my
particular use case where I am dealing with a large number of geometries 
already downloaded in the browser.  I am typically
finding that trying to render Leaflet GeoJSON layers of more than about 
120,000 geometries may start to cause response time
and memory issues within the browser.  If there is any other suggested 
way to make such rendering more scalable in Leaflet,
I would be interested to hear.  You can take a look Visual Field here;  
http://visualfield.org  This is an app that I recently
released that I leverage Leaflet to do the map rendering.  However, to 
be practical, I had to put limits on rendering large
CSV (with WKT) layers to no more than 120,000 geometries.  If there is 
some other client side scalable alternative, I would
be interested.

PR #6663, opened 2019-05-13
GeoJSON.resetStyle: allow invocation without an argument

### LINKED ISSUE #4298: L.geoJSON.resetStyle should be callable without arguments.
`geoJSON.setStyle()` does not require or even utilize a target layer thus, I believe in the spirit of consistency `geoJSON.resetStyle()` should also operate without a target layer.

[My proposed change is here.](https://github.com/Leaflet/Leaflet/compare/master...JrFolk:GeoJSON-api)

I believe `layer.DefaultOptions` was unused because the style was taken from the `geoJSON` layer's *current* style rather than the `featureGroup`'s 'default' style.

Previously, if additional featureGroups were added to a `geoJSON` layer then resetting them all, on a `mouseout` event for instance, would have to be done with `geoJSON.eachLayer()`.  Additionally, the propagated style would be taken from the *last `geoJSON` layer to fire the event*, which could vary depending on which featureGroup initially fired the event, resulting in un-intuitive behavior

With this change, if you add additional `featureGroup`s to a `geoJSON` layer it will retain the 'default' style that it was instantiated with rather than being forced to use the current `geoJSON`'s style.

&nbsp;

As an example, I load several geoJSON features with varying styles into a single `geoJSON` layer.  I want them all to highlight when any feature is hovered over to indicate their grouping, but return to their original style on `mouseout`.  This allows a simple event callbacks to accomplish this:

```javascript
layer.on('mouseover', function(){ this.setStyle(/* highlight features */) })
     .on('mouseout', function(){ this.resetStyle() })
```

&nbsp;

It also allows the addition of a `geoJSON` layer to another `geoJSON` layer and the `styleReset` will propagate as described above.  Previously, because the `geoJSON` layer has no 'default' style, it would end up using no style, which defaulted to the Leaflet style.  While this is probably not recommended, it is possible and thus I think expected behavior should be supported.

joukewitteveen:
When no argument is provided, the style of each of the layers in the
GeoJSON layer is reset.

Fixes #4298

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -127,9 +127,13 @@ export var GeoJSON = FeatureGroup.extend({
 		return this.addLayer(layer);
 	},
 
-	// @method resetStyle( <Path> layer ): this
+	// @method resetStyle( <Path> layer? ): this
 	// Resets the given vector layer's style to the original GeoJSON style, useful for resetting style after hover events.
+	// If `layer` is omitted, the style of all features in the current layer is reset.
 	resetStyle: function (layer) {
+		if (layer === undefined) {
+			return this.eachLayer(this.resetStyle, this);
+		}
 		// reset any custom styles
 		layer.options = Util.extend({}, layer.defaultOptions);
 		this._setLayerStyle(layer, this.options.style);


IvanSanchez:
Seems like a nice & simple addition, but your PR is breaking the continuous integration tests. Please check https://travis-ci.org/Leaflet/Leaflet/builds/531842818? and fix accordingly. Looks like it's just `eslint` complaining about the coding style.

joukewitteveen:
fixed :-)

ghybs:
Hi,

Thank you for your contribution and sorry for the delay before merging this!

PR #6674, opened 2019-05-23
Add Leaflet.IconMaterial plugin

ilyankou:


IvanSanchez:
Looks good, but I can't help to feel that it's kinda redundant, given Leaflet.Icon.Glyph :-|

Maybe there's potential for merging those plugins?

ilyankou:
In my understanding Leaflet.Icon.Glyph is not SVG but a png, right?

IvanSanchez:
> In my understanding Leaflet.Icon.Glyph is not SVG but a png, right?

Yeah, and I can see how the implementation of L.IconMaterial (which BTW should be named L.Icon.Material as per https://github.com/Leaflet/Leaflet/blob/master/PLUGIN-GUIDE.md#plugin-api ) is novel.

My concern is just that there are 4 or 5 marker/icon plugins that do basically the same in different ways, and there should be a way to refactor all that :-|

PR #6676, opened 2019-05-24
Add module build

cedeber:
It exposes the module build file for tools like @pikapkg

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "uglify-js": "~3.5.10"
   },
   "main": "dist/leaflet-src.js",
+  "module": "dist/leaflet-src.esm.js",
   "style": "dist/leaflet.css",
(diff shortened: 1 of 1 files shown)


mourner:
Please see #6239 for the reason we removed this previously.

cedeber:
Oh, it sounds sad in 2019.

deorder:
What is the correct way to let rollup find the correct file `dist/leaflet-src.esm.js` without having to modify `package.json` and add `"module": "dist/leaflet-src.esm.js",` manually after every npm update or reinstall?

johnd0e:
More related comments are in #7055.

mathe42:
Solution for rollup: https://www.npmjs.com/package/@ec-nordbund/rollup-plugin-leaflet

PR #6681, opened 2019-05-29
Allow overriding of GeoJSON output

inkstak:
**Background:**

I reported in #6586 a problem about accuracy in `L.Util.formatNum` rounding method.
Following my pull request, a performance issue has been reported in #6668, and my changes have been reverted.

**Problem:**

I agree with the arguments advanced in this last thread.
In the end, I especially have a problem with the inaccuracy of the `toGeoJson` function.

The solution for me was to overwrite the `GeoJSON.latLngToCoords` but it's not so easy because Leaflet uses internal API.
If I overwrite this function, it won't be used by `GeoJSON.latLngsToCoords`.
If I also overwrite this function,  it won't be used by every `toGeoJSON` declared for each feature.
Et cetera. At the end, I have to copy so many functions.

**Solution:**

I just replaced `latLngToCoords` with `GeoJSON.latLngToCoords`

EDIT : the first PR text was submitted by error.

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -471,3 +471,44 @@ describe("L.LayerGroup#toGeoJSON", function () {
 		});
 	});
 });
+
+describe("Overriding #toGeoJSON formatting", function () {
+	it("is possible to overwrite output of #toGeoJSON coordinates", function () {
+		var marker = new L.Marker([-1.4837191022531273, 43.49222084042808]),
+		    layerGroup = new L.LayerGroup([marker]);
+
+		expect(layerGroup.toGeoJSON()).to.eql({
+			type: 'FeatureCollection',
+			features: [{
+				type: 'Feature',
+				properties: {},
+				geometry: {
+					type: 'Point',
+					coordinates: [43.492221, -1.483719]
+				}
+			}]
+		});
+
+		L.GeoJSON.latLngToCoords = function (latlng, precision) {
+			if (typeof precision === 'number') {
+				return latlng.alt !== undefined ?
+					[L.Util.formatNum(latlng.lng, precision), L.Util.formatNum(latlng.lat, precision), L.Util.formatNum(latlng.alt, precision)] :
+					[L.Util.formatNum(latlng.lng, precision), L.Util.formatNum(latlng.lat, precision)];
+			} else {
+				return latlng.alt !== undefined ? [latlng.lng, latlng.lat, latlng.alt] : [latlng.lng, latlng.lat];
+			}
+		};
+
+		expect(layerGroup.toGeoJSON()).to.eql({
+			type: 'FeatureCollection',
+			features: [{
+				type: 'Feature',
+				properties: {},
+				geometry: {
+					type: 'Point',
+					coordinates: [43.49222084042808, -1.4837191022531273]
+				}
+			}]
+		});
+	});
+});
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -258,7 +258,7 @@ export function latLngsToCoords(latlngs, levelsDeep, closed, precision) {
 	for (var i = 0, len = latlngs.length; i < len; i++) {
 		coords.push(levelsDeep ?
 			latLngsToCoords(latlngs[i], levelsDeep - 1, closed, precision) :
-			latLngToCoords(latlngs[i], precision));
+			GeoJSON.latLngToCoords(latlngs[i], precision));
 	}
 
 	if (!levelsDeep && closed) {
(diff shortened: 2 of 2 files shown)


inkstak:
This request intends to facilitate overwriting of `toGeoJSON` output.

An other approach to my main problem of accuracy may be that `toGeoJSON` skip numbers formatting when not precision (or a specific value) is provided. This is the approach I used in my monkey-patch for now (see the spec attached).

johnd0e:
@inkstak
I have new performance data, see https://github.com/Leaflet/Leaflet/issues/6668#issuecomment-614649784

johnd0e:
@inkstak 
#6587 was reverted, but you may consider "2nd best" approach with epsilon:
```js
  function formatNum_epsilon(num, digits) {
  	var pow = Math.pow(10, (digits === undefined ? 6 : digits));
  	return Math.round((num + Number.EPSILON) * pow) / pow;
  }
```
With this function we get almost same precision as in your original PR:
```
L.Util.formatNum(1.005, 2) === 1.01
L.Util.formatNum(1.555, 2) === 1.56
L.Util.formatNum(-1.4837191022531273, 18) === -1.483719102253127 // last digit is lost
```
Performance is great: https://jsperf.com/leaflet-issue-6668

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -471,3 +471,44 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			type: 'FeatureCollection',
+			features: [{
+				type: 'Feature',
+				properties: {},
+				geometry: {
+					type: 'Point',
+					coordinates: [43.49222084042808, -1.4837191022531273]
+				}
+			}]
+		});
```
johnd0e, line 512:
```suggestion
		var latLngArr = [-1.4837191022531273, 43.49222084042808];
		var lngLatArr = [latLngArr[1], latLngArr[0]];
		var latLng = L.latLng(latLngArr);

		var latLngToCoords = L.GeoJSON.latLngToCoords; // keep ref to original func

		// override with custom function, to prevent latlngs formatting when precision is not specified
		// (to avoid round-off errors)
		L.GeoJSON.latLngToCoords = function (latlng, precision) {
			if (typeof precision === 'number') {
				return latLngToCoords.apply(this, arguments);
			}
			return latlng.alt !== undefined ? [latlng.lng, latlng.lat, latlng.alt] : [latlng.lng, latlng.lat];
		};

		expect(L.GeoJSON.latLngToCoords(latLng, 6)).to.eql([43.492221, -1.483719]);
		expect(L.GeoJSON.latLngToCoords(latLng)).to.eql(lngLatArr);

		var marker = new L.Marker(latLng);
		expect(marker.toGeoJSON()).to.eql({
			type: 'Feature',
			properties: {},
			geometry: {
				type: 'Point',
				coordinates: lngLatArr
			}
		});
		L.GeoJSON.latLngToCoords = latLngToCoords;
```

johnd0e:
I understand the issue, but I do not like proposed solution, as it is too specific.

Currently most of internals routines are not overridable. Even those that look like overridable.

E.g., the most straight-forward way to solve your issue could be overriding just single function: `L.Util.formatNum`, look here:
https://github.com/Leaflet/Leaflet/blob/1ae785b73092fdb4b97e30f8789345e9f7c7c912/src/layer/GeoJSON.js#L258-L261

Unfortunately, if you look into built `leaflet-src.js`, you see that that wouldn't work:
```js
function latLngToCoords(latlng, precision) {
	precision = typeof precision === 'number' ? precision : 6;
	return latlng.alt !== undefined ?
		[formatNum(latlng.lng, precision), formatNum(latlng.lat, precision), formatNum(latlng.alt, precision)] :
```

Thus allowing overriding of some arbitrary function looks like inconsistent decision.

johnd0e:
I would better think about alternative approaches:
1. Refactor to allow overriding of most internal routines
2. Or solve your issue in some explicit way, e.g. what if we change function in this way:
```js
// @function latLngToCoords(latlng: LatLng, precision?: Number or Function): Array
// precision:
// - if given value is `number` than apply `L.Util.formatNum` function to latlng values
// - if it is `function` than apply it to to format latlng values
function latLngToCoords(latlng, precision) {
```
3. Alternatively: make overridable copy of `formatNum` in `L.GeoJSON` namespace.
4. ???

What do think?

johnd0e:
I have an idea: enhance `L.Util.formatNum` to allow special `null` value of `digits`:
```js
	if (digits === null) { return num; } // no formatting
```

inkstak:
Hi @johnd0e, thank you for looking into this old problem.

I like your idea, and the possibility to skip precision with `toGeoJSON(false)`
With this solution, this PR is no longer needed

inkstak:
Closed in favor of #7100

PR #6690, opened 2019-06-07
Check for elements in e.touches

dangowans:
My map wouldn't drag.  After inspecting, I found that the touches element existed on the object, but had no elements.  Doing a check for the number of elements avoided an array out of bounds issue.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -112,7 +112,7 @@ export var Draggable = Evented.extend({
 		// Fired when a drag is about to start.
 		this.fire('down');
 
-		var first = e.touches ? e.touches[0] : e,
+		var first = e.touches && e.touches.length > 0 ? e.touches[0] : e,
 		    sizedParent = DomUtil.getSizedParentNode(this._element);
 
 		this._startPoint = new Point(first.clientX, first.clientY);


mourner:
@dangowans interesting! What device / OS / browser was this on?

dangowans:
It's a Bootstrap 4 website I'm working on.  Firefox 67.0.1 64bit on Windows 10.  I've used Leaflet in the past without this issue.  First time with Bootstrap 4, but not sure if that has anything to do with it.

IvanSanchez:
Can you publish a fiddle/codepen/[plunkr](https://leafletjs.com/edit) that displays this behaviour?

dangowans:
Try as I may on CodePen, I can't seem to recreate the problem.  I guess that's a good thing.

dangowans:
Oddly enough, the code in my application appears to be working now as well.  Perhaps it needed a weekend break.  If it happens again, I'll try to document how it happened.

Although, it still may not hurt checking for an element in the array first.

src/dom/Draggable.js
```diff
@@ -112,7 +112,7 @@ export var Draggable = Evented.extend({
 		// Fired when a drag is about to start.
 		this.fire('down');
 
-		var first = e.touches ? e.touches[0] : e,
+		var first = e.touches && e.touches.length > 0 ? e.touches[0] : e,
```
johnd0e: In this line we are handling `touchstart` event.
And it MUST https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/touches property.

And if you find that it doesn't - then we should find we that ever happened.

Never try to fix issues blindly.

dangowans: Fair enough.  Thank you for your insight.

Issue #5004, opened 2016-10-05
Speed improvement by using "passive" event handlers


fbonzon:
Background: https://blog.chromium.org/2016/05/new-apis-to-help-developers-improve.html

When inspecting a simple Leaflet 1.0.1 map (from the [Leaflet Quick Start Guide](http://leafletjs.com/examples/quick-start/)) on my Android 6.0 phone with USB debugging on Chrome on my desktop machine, I see repeated entries of this warning in the Console:

> Handling of 'touchmove' input event was delayed for … ms due to main thread being busy. Consider marking event handler as 'passive' to make the page more responsive.

Is there any code in Leaflet that prevents a scroll to happen, and if not could we declare the touchmove event listeners (and maybe others) as passive ?

IvanSanchez:
From https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md (linked from that blog post):

> **By marking a touch or wheel listener as passive, the developer is promising the handler won't call `preventDefault` to disable scrolling.**

Fact is, Leaflet **does** call `preventDefault` whenever a touch drag happens: it needs to prevent the browser's behaviour (scrolling or zooming the whole page) in order to do stuff on the map (pan it or zoom it).

While *theoretically* it should be possible to use some clever `overflow` properties in the map container to allow touch-capable browsers to pan it around, it would be very finicky and prone to misbehaviours across different browsers (and then there would be the question of how to attach that to two-finger touch gestures and prevent things like https://github.com/Leaflet/Leaflet/issues/3798 from happening).

> Is there any code in Leaflet that prevents a scroll to happen

Yes, and in a few critical places: https://github.com/Leaflet/Leaflet/search?q=preventdefault

sandstrom:
I have [asked the Chrome team](https://bugs.chromium.org/p/chromium/issues/detail?id=489802#c42) to consider removing the warning if passive is explicitly set.

dtapuska:
FWIW... Chrome supports pointerevents so I believe new versions of leaflet detect that and should use that code path instead of touchstart/touchmove.

vinayakkulkarni:
Just did a Lighthouse scan:

<img width="848" alt="screen shot 2018-07-04 at 5 14 00 pm" src="https://user-images.githubusercontent.com/19776877/42275370-b0da1430-7fad-11e8-83a3-0c657a15eb8a.png">

https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners

[...]

PR #6694, opened 2019-06-11
Use passive event listeners

danielkorte:
Even though Leaflet does call preventDefault this makes it explicitly so and thus passes Lighthouse testing. #5004

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -126,7 +126,9 @@ export var passiveEvents = (function () {
 		});
 		window.addEventListener('testPassiveEventSupport', Util.falseFn, opts);
 		window.removeEventListener('testPassiveEventSupport', Util.falseFn, opts);
-	} catch (e) {}
+	} catch (e) {
+		// Errors can safely be ignored since this is only a browser support test.
+	}
 	return supportsPassiveOption;
 });
 
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -1,3 +1,4 @@
+import * as Util from './Util';
 import {svgCreate} from '../layer/vector/SVG.Util';
 
 /*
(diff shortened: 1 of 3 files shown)


danielkorte:
Updated!

danielkorte:
Good call. Updated to include a passive events support check. Support function stolen from [Modernizr](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/dom/passiveeventlisteners.js).

mourner:
Thanks!

zipzit:
~~Hmmm.  I'm running this change (via build from source) I've verified that my cloned content has the lines shown here, but I'm just not seeing the expected results.  Is it possible that the changes made to src/core/Browser.js , src/dom/DomEvent.DoubleTap.js, src/dom/DomEvent.js aren't propogating over to the dist folder correctly after `$ npm install` and `$ npm run build` ?~~

Oops.  I didn't correctly clear Safari cache for local network testing.  Scratch that.  (I'm leaving my comment in, for anybody else who gets stuck in the same place...)

This update totally fixed my issue of missing tiles on mobile browser.  Many thanks to @danielkorte and support folks.

sk33lz:
Would you be able to tag a new release with this fix? The latest release, 1.5.1, does not include this, so it's not available on any CDNs.

iamandrewluca:
@mourner any ETA when a new release will be made? Thanks!

gacallea:
@danielkorte thanks for the inclusion in 1.6.0 :)

however, I still get a lighthouse error as leaflet doesn't seem to be using passive listener. is there anything I am missing that I need to add to my maps script? or should this work by default?

please and thank you

mourner:
Heads up that this wasn't actually working until #6930

gacallea:
> Heads up that this wasn't actually working until #6930

does this mean we have to wait for 1.6.1? (I'm confused sorry).

Ivan-Perez:
It was actually working in modern browsers, and in older ones there may be some unwanted behaviour because the code always tries to use passive events without checking browser support of that feature.

gacallea:
> It was actually working in modern browsers, and in older ones there may be some unwanted behaviour because the code always tries to use passive events without checking browser support of that feature.

I see, but I have the issue on modern browsers (I don't use old browsers at all).

xstable:
I had the same issue on 1.6.0 within Gatsby: 
image

Can be tested here: https://onepager.erfolgsbeginn.de/
It's part of gatsby, so don't worry about the filename. Look at it source and beautify the contained js, and you see thats leaflet.

johnd0e:
> I had the same issue on 1.6.0 within Gatsby:

You should try latest `master`, it is more recent.

PR #6697, opened 2019-06-14
RIP Mapzen: update Pelias geocoder plugin info

missinglink:
This PR updates information about what was previously called the `mapzen/leaflet-geocoder` and is now named `pelias/leaflet-plugin`.

I've updated the service URL to point to 'geocode.earth' which is run by the same people that ran the 'Mapzen Search' (full disclosure, I am one of them).

I just merged https://github.com/pelias/leaflet-plugin/pull/179 to fix the demo to remove all the old `mapzen.com` links.
The geocode.earth API key hasn't been enabled on our rate-limiter yet but should be available for use shortly.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -4209,10 +4209,10 @@ External services that transform an address or the name of a place into latitude
 	</tr>
 	<tr>
 		<td>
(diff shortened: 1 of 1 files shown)


missinglink:
The build is failing, I'm assuming that wasn't me!? 🤷‍♂

Issue #6552, opened 2019-03-11
window is not defined


ilostit:
Hi,

i have this issue after update reactivemaps "@appbaseio/reactivemaps"
leaflet 1.4.0

XXX/node_modules/leaflet/dist/leaflet-src.js:7
typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
^
ReferenceError: window is not defined
at XXX/node_modules/leaflet/src/core/Util.js:217:24
at version (XXX/node_modules/leaflet/dist/leaflet-src.js:7:65)
at Object. (XXX/node_modules/leaflet/dist/leaflet-src.js:10:2)
at Module._compile (internal/modules/cjs/loader.js:689:30)
at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
at Module.load (internal/modules/cjs/loader.js:599:32)
at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
at Function.Module._load (internal/modules/cjs/loader.js:530:3)
at Module.require (internal/modules/cjs/loader.js:637:17)
at require (internal/modules/cjs/helpers.js:20:18)

any idea ?

IvanSanchez:
Hi, and thanks for taking the time to open a bug report in Leaflet.

However, in this repository we only handle bugs in "[vanilla](http://vanilla-js.com/)" Leaflet. This means that we do **not** handle bugs which are specific to frameworks such as:
* React
* Ractive
* AngularJS
* Bootstrap
* Leaflet for R
* Joomla
* Wordpress
* Ionic
* Cordova
* Vue
* Electron
* Polymer
* TypeScript
* ...etc

Please understand that we only have the time and energy to test Leaflet in plain web browsers.

Please try to either reproduce the bug without using any frameworks, or submit a bug to the appropiate repo.

HoraceShmorace:
This error isn't specific to any framework. It merely shows up in those frameworks because of the Leaflet dependency.

This is happening when trying to include Leaflet as a dependency in a _universal_ web app, which gets server-side rendered first (SSR). This is a standard SEO requirement for any content-based site.

This error occurs when simply including the `leaflet` dependency in an SSR app (in my case, via npm), and is happening because Leaflet is trying "to do stuff" that depends on the window object before an instance is even instantiated, which is a bad practice. Initialization should be explicit, via the constructor (or some other explicitly executed closure).

IvanSanchez:
Let me quote myself, emphasis mine:

[...]

Issue #3220, opened 2015-02-10
ReferenceError: window is not defined


irshad-platalytics:
I am facing issue "ReferenceError: window is not defined" using leaflet with nodeJs in as:

c:\Projects\Leaflet\node_modules\leaflet\dist\leaflet-src.js:9180
}(window, document));
  ^
ReferenceError: window is not defined
    at Object.<anonymous> (c:\Projects\Leaflet\node_modules\leaflet\dist\leaflet-src.js:9180:3)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (c:\Projects\Leaflet\app.js:7:9)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)

jfirebaugh:
Based on your backtrace it looks like you're trying to use it in node or some other non-browser environment, which isn't going to work. Leaflet is a browser library.

PR #6715, opened 2019-07-03
Expose a server api

jamestalmage:
Exposes a limited subset of the API that functions in Node.js.

Includes the following modules:
  * core
  * geo
  * geometry

The following modules are NOT included (because they depend on browser/dom):
  * control
  * dom
  * layer
  * map

I moved around a few functions `core/Util` and `core/Browser` to make swapping the implementation for Node easier. Specifically, any references to `window` were moved out of Util, and into the browser package (Util still exports exactly the same functionality, there should be no breaking changes).

The test suites for the exported API are now also run in Node.

Uses `esm` to provide module support. No bundler / transpiler necessary!

Usage:

```js
const L = require('leaflet/server');
const crs = L.CRS.EPSG3857;

crs.latLngToPoint(L.latLng(0, 0));
```

### DIFF
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -2,6 +2,7 @@
 
 import rollupGitVersion from 'rollup-plugin-git-version'
 import json from 'rollup-plugin-json'
+import resolve from 'rollup-plugin-node-resolve'
 import gitRev from 'git-rev-sync'
 import pkg from '../package.json'
 
@@ -53,6 +54,9 @@ export default {
 	],
 	legacy: true, // Needed to create files loadable by IE8
 	plugins: [
-		release ? json() : rollupGitVersion()
+		release ? json() : rollupGitVersion(),
+		resolve({
+			browser: true
+		})
 	]
 };
--- a/build/rollup-watch-config.js
+++ b/build/rollup-watch-config.js
@@ -2,6 +2,7 @@
 // This adds a sanity check to help ourselves to run 'rollup -w' as needed.
 
 import rollupGitVersion from 'rollup-plugin-git-version'
+import resolve from 'rollup-plugin-node-resolve'
 import gitRev from 'git-rev-sync'
 
 const branch = gitRev.branch();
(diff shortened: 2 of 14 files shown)


jamestalmage:
CRS zoom and scale test passes in Node 10, but in Node 12 produces the following error:

```
      Error: expected 2.4999999999999996 to sort of equal 2.5
```

Implemented an `almostEqual` assertion to handle the floating point error.

jamestalmage:
Is there any interest in merging this?

ghybs:
Hi and thank you for this very nice first contribution to this repository!

I know that some people are looking for such possibility of using Leaflet in node (e.g. #3220), even though mostly for SSR (e.g. #6332, #6552).

Unfortunately, it seems to me that the core maintainers have no such use case, therefore I doubt we would be able to seriously maintain such a large contribution. :worried:

However, if I understand correctly how the PR works, you provide a server specific entry point (server.js > LeafletServer.js), and provide 2 versions of Browser features detection so that it is compatible with node, and resolved by rollup for the browser?
In that case, it seems to me that it should be possible to make it a consumer package, that will directly reference the appropriate src files (possibly replacing the core index and Browser to use your replacement instead)
That way, you would still have a server-ready lib, that you can easily upgrade with Leaflet, and you would keep maintenance control (rather than relying on maintainers who have no stake in this feature...)

The test tweaks might still be interesting to keep though, especially the ability to run in node without a headless browser, if you would care to submit a PR just for those (and if that still makes sense without the rest of the modifications?)

slutske22:
I am in need of this!

NimitTRT:
Is there a way to use Leaflet in SSR?

mathe42:
> Is there a way to use Leaflet in SSR?

Not that easy. You can get SSG with for example puppetier - I think. But I never tried that. 

As this is only a bug tracker such questions should be formulated as a feature request in a new issue or put on StackOverflow.

Falke-Design:
SSR will be implemented with V2 #9385

PR #6716, opened 2019-07-03
Update plugins.md

rightdroid:
Remove linebreaks that broke markdown

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -13,7 +13,6 @@ While Leaflet is meant to be as lightweight as possible, and focuses on a core s
 <div id="toc" class="clearfix">
 	<div class="toc-col">
 		<h4>Tile &amp; image layers</h4>
-
 		<ul>
 			<li> <a href='#basemap-providers'>Basemap providers</a></li>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm looking at https://leafletjs.com/plugins.html and it's not broken. I'm not clear about what you're trying to fix here.

rightdroid:
HTML displays fine, it's the markdown that is broken. Not a big deal, but a bit confusing when trying to update the .md file using any kind of markdown editor/viewer. You can see the borken markdown here: https://github.com/Leaflet/Leaflet/blob/master/docs/plugins.md. First formatting error is right at the top after "Tile & image layers" heading. Github renders the markdown language, but random linebreaks breaks the formatting, hence broken tables.

I was about to add a plugin to the plugins.md, but thought might as well format it correctly before the actual plugin PR.

IvanSanchez:
Superseded by #6813 / https://github.com/Leaflet/Leaflet/commit/3ce6e3040b80e7d37b62b74e0b7f531680ebd769

Issue #6257, opened 2018-07-27
"move" event has oldLatLng in Marker but not in Circle Marker


mmacquarrie:
- [X] I'm reporting a bug, not asking for help; support questions like "How can I do X with Leaflet?" will be closed (use [Stack Overflow](http://stackoverflow.com/) or [gis.stackexchange.com](http://gis.stackexchange.com/) for questions)
- [X] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behaviour is documented and expected
- [X] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)**
- [X] I've searched through the issues to make sure it's not yet reported

~~**The error is a result of a combination of react and leaflet and leaflet-markercluster, but the original source of the error is leaflet~~
UPDATE: it's reproducible with just leaflet

## How to reproduce

- Leaflet version I'm using: 1.3.1
- Browser (with version) I'm using: Chrome 67.0.3396.99
- OS/Platform (with version) I'm using: macOS 10.12.6

## What behaviour I'm expecting and which behaviour I'm seeing
Expect: setLatLng should work for CircleMarkers like it does for Markers
Reality: setLatLng leads to undefined errors

I originally experienced the error using the marker cluster plugin, but it's reproducible without it.
Related issue: https://github.com/YUzhva/react-leaflet-markercluster/issues/31

Error stack trace in leaflet-markercluster:
``` 
leaflet-src.js:1658 Uncaught TypeError: Cannot read property 'lat' of undefined
    at Object.project (leaflet-src.js:1658)
    at Object.latLngToPoint (leaflet-src.js:1497)
    at NewClass.project (leaflet-src.js:3931)
    at NewClass._removeFromGridUnclustered (leaflet.markercluster-src.js:690)
    at NewClass._removeLayer (leaflet.markercluster-src.js:738)
    at NewClass.removeLayer (leaflet.markercluster-src.js:181)
    at NewClass._moveChild (leaflet.markercluster-src.js:713)
    at NewClass._childMarkerMoved (leaflet.markercluster-src.js:703)
    at NewClass.fire (leaflet-src.js:592)
    at NewClass.setLatLng (leaflet-src.js:7843)
```

The problem is that e.oldLatLng is undefined here in leaflet-markercluster:
```
_childMarkerMoved: function (e) {
		if (!this._ignoreMove && !e.target.__dragStart) {
[...]

PR #6719, opened 2019-07-04
Add oldLatLng coordinates to CircleMarker move event

errnesto:
This caused a crash for me when using a CircleMarker with Leaflet.markercluster
because it was expecting the move event to have `oldLatLng`.

### DIFF
--- a/src/layer/vector/CircleMarker.js
+++ b/src/layer/vector/CircleMarker.js
@@ -38,7 +38,7 @@ export var CircleMarker = Path.extend({
 		this.redraw();
 
 		// @event move: Event
-		// Fired when the marker is moved via [`setLatLng`](#marker-setlatlng) or by [dragging](#marker-dragging). Old and new coordinates are included in event arguments as `oldLatLng`, `latlng`.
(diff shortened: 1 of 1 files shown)


src/layer/vector/CircleMarker.js
```diff
@@ -33,9 +33,13 @@ export var CircleMarker = Path.extend({
 	// @method setLatLng(latLng: LatLng): this
 	// Sets the position of a circle marker to a new location.
 	setLatLng: function (latlng) {
+		var oldLatLng = this._latlng;
 		this._latlng = toLatLng(latlng);
 		this.redraw();
-		return this.fire('move', {latlng: this._latlng});
+
+		// @event move: Event
+		// Fired when the marker is moved via [`setLatLng`](#marker-setlatlng) or by [dragging](#marker-dragging). Old and new coordinates are included in event arguments as `oldLatLng`, `latlng`.
```
IvanSanchez: This should be `#circlemarker-setlatlng` instead of `#marker-setlatlng`. Also, `CircleMarker`s do not have a `draggable` option as `Marker`s do.

errnesto: Oh I did't see that sorry should be fixed now :-)

ghybs:
Fix #6257

ghybs:
A test like in PR #2412 would be nice though...

PR #6721, opened 2019-07-04
Increased retina detection threshold (readability on non-retina screens)

vpithart:
## What
This very simple PR raises the threshold for retina detection. 

Before: `devicePixelRatio > 1.00`
After: `devicePixelRatio >= 2.00`

Tested on three different screens: 1.00, 1.25, 2.00 and default "mapnik" OpenStreetMap tiles. The 1.25 screen gets readable with this change.

## Why
The detection of hi-res displays is awesome and yields a brilliant picture on MacBooks or 4K displays where `devicePixelRatio == 2.00`.
On the other hand, it gives unreadable output on middle-resolution displays where `devicePixelRatio == 1.25`. The text labels on downsampled tiles aren't readable (screenshot on the right). Such a display doesn't seem to be a candidate.

Motivation: visual experiance on DPR=1.25 screen with `detectRetina:false` (left) or `detectRetina:true`:
egaplnaaepcodopg
Note the unreadable street names on the right (download the image and display it 1:1 - still unreadable)

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -111,8 +111,8 @@ export var mobileOpera = mobile && opera;
 export var mobileGecko = mobile && gecko;
 
 // @property retina: Boolean
-// `true` for browsers on a high-resolution "retina" screen or on any screen when browser's display zoom is more than 100%.
-export var retina = (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1;
+// `true` for browsers on a high-resolution "retina" screen or on any screen when browser's display zoom 200% or more.
+export var retina = (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) >= 2;
 
 // @property passiveEvents: Boolean
 // `true` for browsers that support passive events.


wasinger:
I'm not sure about the ">= 2" threshold. Wouldn't a display with 1.75 devicePixelRatio look better with the "retina" images? So I suggest using "> 1.5".

vpithart:
Perhaps `1.5` is good. I had no displays to test between 1.50-1.99. So would you please change it to
1.5 ?

IvanSanchez:
This looks good, but I'm intrigued about which devices have a `devicePixelRatio` between 1.25 and 1.75

wasinger:
> This looks good, but I'm intrigued about which devices have a `devicePixelRatio` between 1.25 and 1.75

Both my desktop monitor (4k 27") and my laptop screen (FHD 13") are set to 1.5, so this seems to be a very common value to me.

johnd0e:
So what value should we adopt as retina threshold, `1.5.` or `2`?

vpithart:
Let's agree on the fact that `> 1.00` is problematic. My original proposal to go to `>= 2.00` might be too much.

Please vote 👍 is you agree to change to `>= 1.50` or 👎 to keep 1.00 (no change)

PR #6723, opened 2019-07-09
Add worldCopyJumpCenter option to Map

acarapetis:
## Summary:
Adds an option to Map.Drag to make worldCopyJump prefer a central longitude different to 0.

## Rationale:
One of the simpler options to handle the issue of longitude wrapping is worldCopyJump, which avoids the need to duplicate or move markers. 

For maps concerned principally with the Americas, Europe and the Atlantic between them, it works well: simply place all your markers with longitudes between -180 and 180 as usual, and worldCopyJump will keep things fairly sensible.

Unfortunately, for maps where the region of interest is the Pacific (i.e. crosses the antimeridian), it does not work as well: if we don't want to duplicate layers, we should place our markers in the longitude range [0,360] so that they can all be visible at once; but then worldCopyJump in its current incarnation is no longer useful, since it will jump us over to -180 if we pan too far east. 

With this new option set to 180 and markers placed in the [0,360] range, everything works nicely.


tobske:
We are using leaflet to draw routes of aircraft and for airlines from east Asia the center of their operation is the pacific.

Falke-Design (approved):

Malvoz (approved):

Falke-Design (changes requested):
Tests are missing. Maybe I find some time in the next weeks to add one

PR #6728, opened 2019-07-15
Updated path to check for style being passed

TheRealTorreySmith:
Errors were thrown if the style object was not being passed so I added an extra check for the style object before checking if it "hasOwnProperty('weight').


strimbob:
this would fix my bug 
https://github.com/Leaflet/Leaflet/issues/6825
Can I just approve it?

strimbob:
no worries, thank you!

destus90:
Very important PR. This fixes some issues with `Leaflet.Draw` that seems to be died at all https://github.com/Leaflet/Leaflet.draw/issues/943

PR #6731, opened 2019-07-17
Plugin addition: Leaflet.Heightgraph

TimMcCauley:
We have created this interactive heightgraph plugin which provides the simple but very handy ability to visualize track/route segments with arbitrary colors and labels.


ghybs:
Hi and thank you for your contribution! :+1:

PR #6745, opened 2019-08-02
removed throttle for Canvas 'mousemove'

dnepromell:
Due to throttle some events of 'mousemove' not fired which leads to incorrect behavior.

Reproduce for the issue: https://codepen.io/anon/pen/ZgJPzx
Check console, there are a lot of outputs: mouseover and mouseout. With the fix mouseover and mouseout have to be consistently.

The bug only in the Canvas. If change Canvas to SVG everything will good. Just remove rendererFactory: L.canvas.tile from the example.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -60,7 +60,7 @@ export var Canvas = Renderer.extend({
 	_initContainer: function () {
 		var container = this._container = document.createElement('canvas');
 
-		DomEvent.on(container, 'mousemove', Util.throttle(this._onMouseMove, 32, this), this);
+		DomEvent.on(container, 'mousemove', this._onMouseMove, this);
 		DomEvent.on(container, 'click dblclick mousedown mouseup contextmenu', this._onClick, this);
 		DomEvent.on(container, 'mouseout', this._handleMouseOut, this);
 


IvanSanchez:
And what's the bug, exactly? What's the correct/incorrect behaviour?

Please note that the throttle was specifically tweaked at https://github.com/Leaflet/Leaflet/commit/64333d2dc32e3be605c647e991caa2c3451f20d4 and https://github.com/Leaflet/Leaflet/pull/3547 . And if

dnepromell:
@IvanSanchez 
Move mouse over the map quickly
correct and expected: 
image

incorrect (now for Canvas):
image

IvanSanchez:
I think you're trying to solve the wrong problem here.

You're using Leaflet.VectorGrid, so each polygon is split on a per-tile basis. See that by adding a `canvas {border: 2px solid black}` CSS rule. Whenever the mouse pointer crosses the boundaries between canvases, events are fired. *This is behaviour as expected*.

See for yourself at https://next.plnkr.co/edit/AitPRHoci2y7pEmI?preview

Is the bug reproducible **without VectorGrid**?

vohtaski:
The problem is that sometimes events are not fired because of throttle.
Imagine, you have square A and square B.
If you listen on events in square A and you quickly move a mouse from square A to square B and back to square A, you have the following sequence.
```
mouse over
# inside A
mouse out
 # inside B
(!!!missed!!!) mouse over
# inside A
```
Even though you are inside square A now, the system still thinks you are inside B as an event `mouse over` was missed because of throttle.

For svg tiles it works, for canvas tiles it does not.

Hope the explanation is clearer now.

dnepromell:
additional screenshot
image
I am inside the country but the last event was OUT. Due throttle we lost OVER event.

IvanSanchez:
I see now, that explanation makes understanding the race condition much easier.

Still, I'm reluctant to remove a throttling that was added for performance reasons. I think a better approach would be to keep throttling the event handler logic, but reset the throttling on the DOM  `mouseout` event - in other words, I'd appreciate a review of #6749.

dnepromell:
tried to apply your changes locally, but noticed that VectorGrid plugin doesn't work as uses removed function
image
image

dnepromell:
@IvanSanchez, tested locally, and it works as expected.

cherniavskii:
Closed by #6749

PR #6748, opened 2019-08-05
Plugins: add Leaflet.Maidenhead

IvanSanchez:


ghybs:
Just for reference:
https://stackoverflow.com/questions/57336432/how-to-add-a-maidengrid-overlay-to-leaflet

PR #6749, opened 2019-08-06
Cancel the canvas mousehover throttle on mouseout

IvanSanchez:
This fixes the same as #6745: a race condition when moving the mouse *very* quickly in and
out of a L.Canvas, which affects the 'mouseover' and 'mouseout' that such a L.Canvas fires.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -60,7 +60,7 @@ export var Canvas = Renderer.extend({
 	_initContainer: function () {
 		var container = this._container = document.createElement('canvas');
 
-		DomEvent.on(container, 'mousemove', Util.throttle(this._onMouseMove, 32, this), this);
+		DomEvent.on(container, 'mousemove', this._onMouseMove, this);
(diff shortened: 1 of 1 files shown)


src/layer/vector/Canvas.js
```diff
@@ -355,26 +355,24 @@ export var Canvas = Renderer.extend({
...
 		if (layer) {
 			// if we're leaving the layer, fire mouseout
 			DomUtil.removeClass(this._container, 'leaflet-interactive');
 			this._fireEvent([layer], e, 'mouseout');
 			this._hoveredLayer = null;
+			this._mouseHoverThrottled = false;
 		}
 	},
 
-	_handleMouseHover: function (e, point) {
```
dnepromell: this change will break external plugins like VectorGrid
image

IvanSanchez: Damn, I wasn't familiar with those hacks in VectorGrid.

Will revert.

dnepromell:
any updates? Please merge the fix.

PR #6759, opened 2019-08-08
docs/reference: fix duplicate footer, invalid HTML

simon04:
The `docs/reference*` files duplicate parts of `docs/_layouts/v2.html`. As a consequence, the footer is displayed twice, `docs/docs/js/docs.js` is run twice (resulting in two `#toc-copy` menus), and the served HTML files are invalid (containing markup after `</html>`).

This PR:
- removes the trailing content after `</html>`
- includes "CUT HERE" markers in `build/leafdoc-templates/html.hbs` to hopefully avoid this problem in the future
- moves reference css+js to separate files
- runs eslint on `docs/docs/js` files


mourner:
@simon04 thank you so much for all your documentation fixes! Much appreciated. ❤️

PR #6766, opened 2019-08-13
Avoid modifying source props object in Class.extend

johnd0e:
`extend` method had side effect: it replaced `props.options`
that behavior was undesirable, as it prevented to reuse props.

### DIFF
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -117,6 +117,22 @@ describe("Class", function () {
 			expect(K2.prototype.options.foo).to.eql('bar');
 		});
 
+		it("does not reuse original props.options", function () {
+			var props = {options: {}};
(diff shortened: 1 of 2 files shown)


johnd0e:
Sample:
```
var Mixin = {
  // ...
  // some useful props and methods
}

var MyLine = L.Polyline.extend(Mixin);

var MyGon = L.Polygon.extend(Mixin);

var line = new MyLine([]);

var gon = new MyGon([]);

console.log(line.options.fill, gon.options.fill);
```
Output expected: `false`, `true`
Output before PR: `false`, `false`

johnd0e:
The fix is updated considering one more edge case.
Both cases are now included into `ClassSpec.js`.

mourner (approved):

Falke-Design (approved):

PR #6770, opened 2019-08-17
[docs] fix broken accordions on leafletjs.com + offline docs

mbachner:
added missing reference.js to html layout (see #6759)
all live docs since 1.0.0 affected  (tested on chrome / safari /firefox)


IvanSanchez:
Thanks for noticing this!

Issue #6705, opened 2019-06-25
Click events don’t work in the shadow DOM on iOS


remcohaszing:
- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**

- Create a shadow root.
- Append the leaflet CSS to the shadow root.
- Create a node for Leaflet.
- Attach a map to the node.
- Add a click handler to the map.
- Append the node to the shadow root.

**Expected behavior**

The click handler is triggered

**Current behavior**

The click handler isn’t triggered when using Safari (or Chrome) on iOS.

**Environment**
- Leaflet version: 1.5.1
- Browser (with version): Safari
- OS/Platform (with version): iOS 10.3.1

**Additional context**

It works as expected in Chrome on desktop, Safari on desktop, and Chrome on Android.

**Minimal example reproducing the issue**

https://next.plnkr.co/edit/9WB2YBaVIWv4nQgv

- [x] this example is as simple as possible
- [x] this example does not rely on any third party code

IvanSanchez:
And I assume that other DOM events are fired by the shadow DOM elements? Is `L.DomEvent.on` behaviour as expected, or failing?

remcohaszing:
The following code works as expected:

```js
const root = document.createElement('div');
document.body.appendChild(root);
const shadowRoot = root.attachShadow();
const node = document.createElement('div');
node.style.width = '300px';
node.style.height = '300px';
node.style.backgroundColor = 'blue';
L.DomEvent.on(node, 'click', () => {
  node.style.backgroundColor = node.style.backgroundColor === 'blue' ? 'red' : 'blue';
});
shadowRoot.appendChild(root);
```

zmullett:
I've also observed this. It works as expected on desktop Chrome/Safari, as well as Android Chrome, but not working as expected on iOS 12.4 using Chrome 75.0.3770.103 and Leaflet 1.5.1. I've observed that the event is indeed dispatched from within map/handler/Map.Tap.js, but does not arrive when observed using L.Map.on().

@remcohaszing did you find the root cause or a workaround?

remcohaszing:
No, unfortunately not. If I do, I’ll report back here.

zmullett:
[...]

PR #6788, opened 2019-09-01
iOS L.Map.Tap shadow DOM click event fix

zmullett:
iOS (or more generally platforms without [PointerEvent](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) support) uses Map.Tap to handle touch events. This change provides support for the case where the map exists within a shadow DOM. In such a case the PointerEvent.target references the outer web component, which means events are not dispatched from the expected element.

### DIFF
--- a/src/map/handler/Map.Tap.js
+++ b/src/map/handler/Map.Tap.js
@@ -124,7 +124,10 @@ export var Tap = Handler.extend({
 		        e.clientX, e.clientY,
 		        false, false, false, false, 0, null);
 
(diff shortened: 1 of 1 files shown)


zmullett:
@IvanSanchez please take a look - this solves [issues/6705](https://github.com/Leaflet/Leaflet/issues/6705).

IvanSanchez:
I'm worrying about possible side effects - in my experience, event code has proven to be particularly fiddly. Would it be possible to check if the map is inside a Shadow DOM, and trigger the event in one way or the other that way?

PR #6795, opened 2019-09-10
Check if e.touches exists

richardhinkamp:
Saw a lot of errors in out logging on this line. Should check if it exists, like it does on line 33.

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -22,7 +22,7 @@ export function addDoubleTapListener(obj, handler, id) {
 			if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
 			count = _pointersCount;
 		} else {
-			count = e.touches.length;
+			count = e.touches ? e.touches.length : 1;
 		}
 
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Before merging, I'd like to know the platform (OS + browser + input device) that you're experiencing this bug in. Given my past experiences with touch events, I feel it'd be safer to investigate a bit.

src/dom/DomEvent.DoubleTap.js
```diff
@@ -22,7 +22,7 @@ export function addDoubleTapListener(obj, handler, id) {
 			if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
 			count = _pointersCount;
 		} else {
-			count = e.touches.length;
+			count = e.touches ? e.touches.length : 1;
```
johnd0e: This fix is terribly wrong, it could just mask the true root of real issue.

Here we get either `pointer` event, or `touch` event.
And in 'else' path we expect exactly `touch` event, which MUST contain `touches` property.

If there is no such field in your case - it means that something wrong in other parts of code.

PR #6800, opened 2019-09-13
Adds Leaflet Copy Coordinates Control

tinjaw:
I have forked the four year old Leaflet Coordinates Control to make it available via npm and to maintain it.


ghybs:
:+1:

PR #6813, opened 2019-09-20
fixed tables

0xflotus:


IvanSanchez:
What does this fix, exactly? All I see is some whitespace removal.

0xflotus:
> What does this fix, exactly? All I see is some whitespace removal.

Look at the preview. The tables are broken and the removed whitespace fixes this.

IvanSanchez:
The important thing is the end result at https://leafletjs.com/plugins.html , not the previews in the git repo browser. Merging anyway since it's not too conflictive.

PR #6815, opened 2019-09-23
Fix doubleTap events on touch devices

sashadev-sky:
This builds on the work done in https://github.com/Leaflet/Leaflet/pull/5881, but adds an extra check that resolves all of the issues mentioned by users in the comments section. 

It also introduces a level of stability. For ex, if a newer version of Chrome does start to fire native dblclick events, the functionality will not break

Tested on Mobile: Chrome, Firefox, Safari, Opera, Samsung Internet, Edge
Desktop Chrome, IE 9 - IE 11, Firefox, Safari, Opera, Edge

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -9,6 +9,13 @@ var _touchstart = Browser.msPointer ? 'MSPointerDown' : Browser.pointer ? 'point
 var _touchend = Browser.msPointer ? 'MSPointerUp' : Browser.pointer ? 'pointerup' : 'touchend';
 var _pre = '_leaflet_';
 
+function browserFiresNativeDblClick(e) {
+	// See https://github.com/w3c/pointerevents/issues/171
+	// Note however, that Chrome stopped firing native dblclick for
+	// touch since that issue was written.
+	return (Browser.ie || e.pointerType === 'mouse');
+}
+
 // inspired by Zepto touch code by Thomas Fuchs
 export function addDoubleTapListener(obj, handler, id) {
 	var last, touch,
@@ -17,9 +24,10 @@ export function addDoubleTapListener(obj, handler, id) {
 
 	function onTouchStart(e) {
 		var count;
+		var oe = e.originalEvent;
 
 		if (Browser.pointer) {
-			if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
+			if (browserFiresNativeDblClick(e) || (oe && !oe.sourceCapabilities.firesTouchEvents)) { return; }
 			count = _pointersCount;
 		} else {
 			count = e.touches.length;
@@ -38,7 +46,7 @@ export function addDoubleTapListener(obj, handler, id) {
 	function onTouchEnd(e) {
 		if (doubleTap && !touch.cancelBubble) {
 			if (Browser.pointer) {
-				if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
+				if (browserFiresNativeDblClick(e)) { return; }
 				// work around .type being readonly with MSPointer* events
 				var newTouch = {},
 				    prop, i;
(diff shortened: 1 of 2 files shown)


sashadev-sky:
@perliedman let me know your thoughts please! For now planning to use it in my own plugin https://github.com/publiclab/Leaflet.DistortableImage so would love to ensure it passes standards

perliedman:
Hi @sashadev-sky! Happy to see someone taking this on where I left off.

I am sorry to say that I have long since lost track of the issue as well as the possible fix, it looks like I opened the PR like two years ago. Also, I have stepped away a bit from Leaflet maintenance since then, so unfortunately I have little feedback to give at this point, and my current situation makes it hard to find the time to do a proper review.

I can just hope that maybe @cherniavskii (or someone else who is an active maintainer?) can have a look. Given that #5881 passed review I guess it should not be too much work to get this approved.

Sorry to not be more helpful!

ilblog:
I have a humble request. Can you add fixing #6817 to this PR? You are fixing basically the same stuff. Whole #6817 thread is pretty self explanatory and the solution is easy.

src/dom/DomEvent.DoubleTap.js
```diff
@@ -9,6 +9,19 @@ var _touchstart = Browser.msPointer ? 'MSPointerDown' : Browser.pointer ? 'point
 var _touchend = Browser.msPointer ? 'MSPointerUp' : Browser.pointer ? 'pointerup' : 'touchend';
 var _pre = '_leaflet_';
 
+function browserFiresNativeDblClick(e) {
+	// See https://github.com/w3c/pointerevents/issues/171
+	// Note however, that Chrome stopped firing native dblclick for
+	// touch since that issue was written.
```
johnd0e: Just tested on android.
Chrome 80 and Firefox 68.5 and Firefox Preview 3.2.1 - all fire dblclick.

Checked with this page: https://rbyers.github.io/eventTest.html

sashadev-sky: @johnd0e you're saying the PR isn't necessary anymore? I figured it's been fixed by now. We can close it if that's the case!

johnd0e: No, I'm only saying some facts verified on my android phone.
And I have no desktop with touchscreen to make similar tests.

Also, Safari (incl. iOS) is known for issues with dblclick.

As of this PR in general...
It's obvious that Leaflet code was written long time ago, and many parts should be revised.
But I have not enough data to decide whether proposed here changes can makes things better or not.

sashadev-sky: I don't know. I've been meaning to update the Leaflet version I currently use so i'll loop back here when I do

IvanSanchez: > It's obvious that Leaflet code was written long time ago, and many parts should be revised.

In particular, it's time to revisit the `pointerify` branch and make pointer events first-class citizens, instead of all the current hacks and workarounds.

> I don't know. I've been meaning to update the Leaflet version I currently use so i'll loop back here when I do

Please **do not** submit bugs or comment on existing ones if you haven't checked if they apply to the latest release. It's a waste of time for everybody involved.

src/dom/DomEvent.DoubleTap.js
```diff
@@ -9,6 +9,19 @@ var _touchstart = Browser.msPointer ? 'MSPointerDown' : Browser.pointer ? 'point
...
+	// See https://github.com/w3c/pointerevents/issues/171
+	// Note however, that Chrome stopped firing native dblclick for
+	// touch since that issue was written.
+	return (Browser.ie || e.pointerType === 'mouse');
+}
+
+function browserNeedsAManualDblClick(e) {
+	var oe = e.originalEvent;
+
+	return (L.Browser.Touch && (oe && oe.sourceCapabilities.firesTouchEvents));
```
johnd0e: @sashadev-sky, I've failed to understand purpose of `sourceCapabilities.firesTouchEvents` here.

[AFAIK](https://developer.mozilla.org/en-US/docs/Web/API/InputDeviceCapabilities_API), this can be useful when we listen to both touch and mouse events.
But for our code it is not in case, as there pointer events only (and we have exact `pointerType` value for them).

And what is mentioned 'level of stability'?
> It also introduces a level of stability. For ex, if a newer version of Chrome does start to fire native dblclick events, the functionality will not break

PR #6827, opened 2019-10-01
Fixes iOS 13 touch event issues

### LINKED ISSUE #6817: Contextmenu is not triggered on iOS13

ilblog:
Fixes #6817 at least on iPhone.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -93,7 +93,8 @@ export var msPointer = !window.PointerEvent && window.MSPointerEvent;
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
-export var pointer = !!(window.PointerEvent || msPointer) && !userAgentContains('iphone');
+export var pointer = !!(window.PointerEvent || msPointer) && !userAgentContains('iphone') && 
+    		!(navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);
 
 // @property touch: Boolean
 // `true` for all browsers supporting [touch events](https://developer.mozilla.org/docs/Web/API/Touch_events).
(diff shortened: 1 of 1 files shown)


ilblog:
The fix above detects only iPhone with iOS 13, since iPad iOS3 has same UA string as macOS. 

There is also way how to detect iPad with iOS 13 here https://stackoverflow.com/questions/57765958/how-to-detect-ipad-and-ipad-os-version-in-ios-13-and-up

If a reviewer considers this small change can you please add `&& !(navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1)` to modified line?

slimakovec:
any updates on if / when this will be merged? Also if / when the iPad will be fixed?

ilblog:
We have solution that covers iPad also. Uet waiting for reaction of Leaflet guys.

eliboni:
Many thanks for your effort - where is the solution covering iPad? Is it something I can pull myself? This has really a  huge impact on my user basis.

matt-e-king:
To help elevate and convince the Leaflet contributors on the severity of this, even the maps on the [Leaflet Docs](https://leafletjs.com/examples/quick-start/) pages do not work properly on iOS13:

- Pinch zoom also triggers single click
- Double tap doesn't work
- If you press and hold on the pin icon, the native OS contextual menu appears, and then the panning on the map breaks (this will occur if your map has "draggable" pins and the users presses and holds too long)

@eliboni I believe the fix for the iPad and iPhone that @ilblog is suggesting is replacing this line https://github.com/Leaflet/Leaflet/blob/master/src/core/Browser.js#L96 with the following:

```
export var pointer = !!(window.PointerEvent || msPointer) && !userAgentContains('iphone') && !(navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1)
```

I will also attempt this fix and test it - hopefully Leaflet guys come to the rescue!

mourner:
Fixed by #6855 (should be a more robust fix than nav agent sniffing). Please check if it works for you!

ilblog:
Vladimir thanks a lot

PR #6828, opened 2019-10-01
Leaflet.PM was renamed to Leaflet-Geoman

codeofsumit:
I moved my geometry mangement plugin for leaflet, leaflet.pm to a github org and renamed it.
NPM Modules are already live so all users should have an easy migration path. Old version stays up under the old name of course.


ghybs:
Hi and thank you for the amazing plugin and your long commitment! ;)

PR #6829, opened 2019-10-01
Fix for zoom control and double click on map if L_DISABLE_3D = true

### LINKED ISSUE #6820: Zoom control on map do not work if L_DISABLE_3D = true
Zoom control on map do not work if L_DISABLE_3D = true.

```javascript
<script>L_DISABLE_3D = true;</script>
<script src="leaflet.js"></script>
```
[https://plnkr.co/edit/S6LVFcI4fsPpSVnP5V5h?p=preview](url)

My LeafletJS version - 1.5.1. 
OS - Windows 10 1903 (64-bit).
Browser - Mozilla Firefox 69.0.1 (64-bit).
_________________________________________________________________
With best regards, IMMSPgisgroup

life777:
Fixes #6820

When `L_DISABLE_3D` flag is set to `true` Map class ignores `zoomDelta` property from map options and set it to 1. Zoom level rounding happens in `_limitZoom` method of Map class. If zoomDelta is smaller then 0.5 and L_DISABLE_3D set to true than `_limitZoom` method will return the same zoom level and map won't be zoomed.

I found two places where this issue happens:
1. Control.Zoom, if user clicks on zoom in/out buttons
2. Double click on map.

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -1,6 +1,7 @@
 
 import {Control} from './Control';
 import {Map} from '../map/Map';
+import {any3d} from '../core/Browser';
 import * as DomUtil from '../dom/DomUtil';
 import * as DomEvent from '../dom/DomEvent';
 
@@ -68,14 +69,16 @@ export var Zoom = Control.extend({
 	},
 
 	_zoomIn: function (e) {
+		var delta = any3d ? this._map.options.zoomDelta : 1;
 		if (!this._disabled && this._map._zoom < this._map.getMaxZoom()) {
-			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomIn(delta * (e.shiftKey ? 3 : 1));
 		}
 	},
 
 	_zoomOut: function (e) {
+		var delta = any3d ? this._map.options.zoomDelta : 1;
 		if (!this._disabled && this._map._zoom > this._map.getMinZoom()) {
-			this._map.zoomOut(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomOut(delta * (e.shiftKey ? 3 : 1));
 		}
 	},
 
--- a/src/map/handler/Map.DoubleClickZoom.js
+++ b/src/map/handler/Map.DoubleClickZoom.js
@@ -1,5 +1,6 @@
 import {Map} from '../Map';
 import {Handler} from '../../core/Handler';
+import {any3d} from '../../core/Browser';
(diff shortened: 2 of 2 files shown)


life777:
Hi @IvanSanchez 
Could you take a look at my fix? It's quite old. 
Thanks a lot

life777:
@IvanSanchez @mourner Should I change something in my PR? Thanks a lot

src/control/Control.Zoom.js
```diff
@@ -68,14 +69,16 @@ export var Zoom = Control.extend({
 	},
 
 	_zoomIn: function (e) {
+		var delta = any3d ? this._map.options.zoomDelta : 1;
```
IvanSanchez: ...but what if `zoomDelta` is **greater** than 1? There's an edge case of a map with only even (2,4,6,8,10,etc) zoom levels, and this would break that.

Maybe something like `delta = any3d ? zoomDelta : Math.ceil(zoomDelta)`??

life777: @IvanSanchez Yes, you are right if zoomDelta is bigger than 1 my fix has bug because delte will always be 1. I fixed it in new commit.
Also there will be no problems if zoomDelta will be bigger than 1 because map _limitZoom method just round it zoom buttons and double click will work.

src/control/Control.Zoom.js
```diff
@@ -68,14 +69,18 @@ export var Zoom = Control.extend({
 	},
 
 	_zoomIn: function (e) {
+		var zoomDelta = this._map.options.zoomDelta;
+		var delta = any3d ? zoomDelta : Math.max(1, zoomDelta);
```
IvanSanchez: Once again, this will misbehave in an edge case. Think about what happens when `zoomDelta = 1.5`.

life777: @IvanSanchez 
The main problem was with zoomDelta less then 0.5 with disabled 3D because if you take a look at [_limitZoom](https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js#L1576) method it rounds new zoom level. It means if zoomDelta equals 0.25 buttons for zoom in/out and double click won't work because with zoom level 1 will be increased to zoom level 1.25 and rounded in _limitZoom method to 1. It means nothing will be changed. But if zoomDelta will be more then 0.5 it will work but rounded to closes integer. 

in your case if zoomDelta = 1.5 with zoom level 1 when user  clicks on zoom in button in controls zoom will be increased to 2.5, but _limitZoom method will round it to 3.

johnd0e: Makes sense to me

PR #6838, opened 2019-10-06
Solves error: "Argument 1 of Node.appendChild does not implement interface Node"

sagarpreet-chadha:
Hey,
So the popup was not opening when i have multiple leaflet maps on my screen each with markers.

Screenshot 2019-10-06 at 2 17 14 PM

I tried adding/appending String in `bindPopup` function but still same error. This PR resolves that bug and popups are now opening, see screenshot. 

Thanks!

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -317,7 +317,7 @@ Map.include({
 	// Creates a popup with the specified content and options and opens it in the given point on a map.
 	openPopup: function (popup, latlng, options) {
 		if (!(popup instanceof Popup)) {
-			popup = new Popup(options).setContent(popup._content);
+			popup = new Popup(options).setContent(popup._content ? popup._content : popup);
 		}
 
 		if (latlng) {
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -317,7 +317,7 @@ Map.include({
 	// Creates a popup with the specified content and options and opens it in the given point on a map.
 	openPopup: function (popup, latlng, options) {
 		if (!(popup instanceof Popup)) {
-			popup = new Popup(options).setContent(popup);
+			popup = new Popup(options).setContent(popup._content);
 		}
 
 		if (latlng) {


ghybs:
Hi,

Thank you for the PR!

Please can you open a corresponding issue with a reproduction example of the failure you had, so that it is crystal clear the use case that this PR addresses?

Furthermore, adding a test case would be beneficial for this fix.

Thank you!

sagarpreet-chadha:
Fixes #6840 ,

So on debugging i found that in this line:
https://github.com/Leaflet/Leaflet/blob/3d392c7b33fc27110398c70e3f011a2de70722da/src/layer/Popup.js#L319
the popup variable was a class object having `_content` attribute as the string i passed to `bindpopup`,
So due to which this line : 
https://github.com/Leaflet/Leaflet/blob/3d392c7b33fc27110398c70e3f011a2de70722da/src/layer/Popup.js#L385
gave the following error: `"Argument 1 of Node.appendChild does not implement interface Node"`.

Hence this PR solves this and my maps works great now :) Thanks!

ghybs:
Hi,

> the popup variable was a class object having `_content` attribute as the string i passed to `bindpopup`

well this is clearly against the public API signature for [`bindPopup`](https://leafletjs.com/reference-1.5.0.html#layer-bindpopup):

> bindPopup(\<String|HTMLElement|Function|Popup> content, \<Popup options> options?)

Therefore it seems to me that the bug you are trying to solve is specific to your app and caused by incorrect usage of Leaflet API.

As such, I have to reject this PR.

For what reason cannot you directly pass your `popup._content` string into `bindPopup` method, instead of passing your custom class?

sagarpreet-chadha:
Hey @ghybs , 
I investigated and found that this problem was due to different versions of leaflet - the one on which my plugin is made and the one where i was integrating this plugin.
Thanks for the help :)

Also can you please take a look at PR #6842 :smiley:

Issue #6660, opened 2019-05-12
Duplicated anchors in documentation


johnd0e:
E.g. map 'zoom' event is anchored as https://leafletjs.com/reference-1.5.0.html#map-zoom
But it is also map `zoom` option.

PR #6851, opened 2019-10-10
Distinguishes zoom event and zoom option in the docs (#6660).

saricden:
Just a quick PR to fix the duplicate IDs for 'zoom' event & option.

:wave:


IvanSanchez:
This might be indicative of a deeper problem with LeafDoc. Is this specific of the 1.5.0 docs, or does it happen when building the docs from source?

PR #6855, opened 2019-10-10
Fix dblclick event when both Pointer Events and Touch Events are available

### LINKED ISSUE #6817: Contextmenu is not triggered on iOS13

filcab:
This happens on iOS13, at least.

Tested the fix on an iPhone and an iPad, both with iOS 13, and with and
without "Pointer Events" feature (enabled by default, can be disabled in
Safari Settings). There is still an issue with the iPad.

I tried keeping all other (non-"Safari with Pointer Events") code paths
the same.

Fixes #6817 for non-"iPad on iOS13"

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -93,7 +93,7 @@ export var msPointer = !window.PointerEvent && window.MSPointerEvent;
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
-export var pointer = !!(window.PointerEvent || msPointer);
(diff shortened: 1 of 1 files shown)


filcab:
Updated with the fix suggested by @mourner .

sveine:
> 
> 
> Updated with the fix suggested by @mourner .

After this fix for "var pointer = ..." Windows 10 touch does not work i Chrome for zoom/pan. 

By changing the statement below "var pointer = ..." it's working again:

`	var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window || navigator.maxTouchPoints > 0 || navigator.msMaxTouchPoints > 0 ||
		(window.DocumentTouch && document instanceof window.DocumentTouch));`

PR #6862, opened 2019-10-14
Fix toLatLngBounds

stereobooster:
Fix for https://github.com/Leaflet/Leaflet.markercluster/issues/740

execution gets to this point:

```js
function toLatLngBounds(a, b) {
	if (a instanceof LatLngBounds) {
		return a;
	}
	return new LatLngBounds(a, b);
}
```

where `a` is `LatLngBounds` (it has all the same methods), but it is not instance of `LatLngBounds` (it identifies itself as `B`).

Screenshot 2019-10-14 at 11 44 24

### DIFF
--- a/src/geo/LatLngBounds.js
+++ b/src/geo/LatLngBounds.js
@@ -248,7 +248,7 @@ export function toLatLngBounds(a, b) {
 		return a;
 	}
 	if (a.getSouthWest && a.getNorthEast) {
-		return new LatLngBounds(a.getSouthWest(), a.getNorthEast());	
+		return new LatLngBounds(a.getSouthWest(), a.getNorthEast());
 	}
 	return new LatLngBounds(a, b);
 }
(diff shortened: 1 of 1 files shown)


src/geo/LatLngBounds.js
```diff
@@ -247,5 +247,8 @@ export function toLatLngBounds(a, b) {
 	if (a instanceof LatLngBounds) {
 		return a;
 	}
+	if (a.getSouthWest && a.getNorthEast) {
+		return new LatLngBounds(a.getSouthWest(), a.getNorthEast());
```
stereobooster: I can do `return a;` instead

perliedman:
Hi there,
can you explain why you have a class that has everything `LatLngBounds` has but is not an instance of `LatLngBounds`? It sounds like you should really use the correct object.

stereobooster:
Hi, my guess that this has something to do to Webpack/Minification. I'm pretty sure this is `LatLngBounds` object it's just `instance of` does not recognise as one. This is the best solution I have so far, it's hard to understand what is happening there

stereobooster:
True. My thoughts as well.

PR #6869, opened 2019-10-22
Add aria label to the layer control to resolve WCAG issues

### LINKED ISSUE #6821: Layers Control has a WCAG Level A Empty Link Error

**Steps to reproduce**
Open a leaflet map with a layers control that is collapsed by default, e.g. https://leafletjs.com/examples/layers-control/example.html
Check it with a WCAG checker such as WebAim's WAVE tool: http://wave.webaim.org/report#/https://leafletjs.com/examples/layers-control/example.html

**Expected behavior**
The Layer Controls button needs text or context. Success Criterion 2.4.4 Link Purpose (In Context): https://www.w3.org/TR/WCAG21/#link-purpose-in-context

**Current behavior**
The hamburger button has no context on the link to expand the Layer Controls, and throws an error when people are surface-level checking for accessibility issues.

**Environment**
- Leaflet version: All
- Browser (with version): All
- OS/Platform (with version): All

**Solutions**
This can easily be solved with an aria-label.  https://www.w3.org/WAI/GL/wiki/Using_aria-label_for_link_purpose
```
<div class="leaflet-top leaflet-right">
<div class="leaflet-control-layers leaflet-control" aria-haspopup="true">
<a class="leaflet-control-layers-toggle" href="#" title="Layers"></a>
<section class="leaflet-control-layers-list"><div class="leaflet-control-layers-base">
```

to something like:
```
<div class="leaflet-top leaflet-right">
<div class="leaflet-control-layers leaflet-control" aria-haspopup="true">
<a class="leaflet-control-layers-toggle" href="#" title="Layers" aria-label="Toggle Layer Contol"></a>
<section class="leaflet-control-layers-list"><div class="leaflet-control-layers-base">
```

I just don't know how to do that in GitHub yet...

### LINKED ISSUE #7498: Layer control should have `role="button"`
## Motivation

The layer control is intended as (and behaves similarly to) a button, not a link, and should be conveyed as such to screen reader users.

## Proposed solution

Set attribute [`role="button"`](https://www.w3.org/TR/wai-aria-1.2/#button) on the `<a>` element:

https://github.com/Leaflet/Leaflet/blob/b52961aa469dc17b3710ef7a9c0f13b6f4dba550/src/control/Control.Layers.js#L201-L203

## Additional context

- MDN Documentation: [`role="button"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)
- Related PR: https://github.com/Leaflet/Leaflet/pull/6869/ (https://github.com/Leaflet/Leaflet/pull/6869/commits/71bc3781f4725bd8d21dd015806b8d1a299ba6ff)

rowanwins:
Adds an `aria-label` to the layer control to resolve WCAG issues

Resolves #6821
Close #7498

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -198,7 +198,7 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
-		link.setAttribute('aria-label', 'toggle layer control');
+		link.setAttribute('aria-label', 'Layers');
 		link.setAttribute('role', 'button');
 
 		if (Browser.touch) {
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -198,6 +198,7 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
+		link.setAttribute('aria-label', 'Toggle Layer Contol');
 
 		if (Browser.touch) {
 			DomEvent.on(link, 'click', DomEvent.stop);
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -198,7 +198,8 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
-		link.setAttribute('aria-label', 'Toggle Layer Contol');
+		link.setAttribute('aria-label', 'toggle layer control');
+		link.setAttribute('role', 'button');
 
 		if (Browser.touch) {
 			DomEvent.on(link, 'click', DomEvent.stop);


src/control/Control.Layers.js
```diff
@@ -198,6 +198,7 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
+		link.setAttribute('aria-label', 'Toggle Layer Contol');
```
Malvoz: - [ ] Typo: missing "r" in "control". Additionally I think aria-labels and titles in other places aren't capitalized.
- [ ] Please add `role="button"` such that screen readers don't announce this as a regular link.

❤️

rowanwins: Thanks for the suggestions @Malvoz  - changes made

src/control/Control.Layers.js
```diff
@@ -198,6 +198,8 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
+		link.setAttribute('aria-label', 'toggle layer control');
```
Malvoz, line 201: @rowanwins revisiting this, I think the naming of `aria-label` is a bit verbose. In comparison, zoom controls don't say "Zoom in/out control" or "Press to zoom in/out", they're just "Zoom in/out" (that they're "pressable" or "controls" is implied by the fact that they're buttons, well, `role=button`).

So IMO "Layers" should be perfectly fine (same as `title`). It'd also be more consistent with other controls, since almost all of Leaflet's controls duplicate the name in both `aria-label` and `title`.

Malvoz (changes requested):

Malvoz (approved):

src/control/Control.Layers.js
```diff
@@ -198,6 +198,8 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
+		link.setAttribute('aria-label', 'Layers');
```
Malvoz, line 201: This should ideally not have both `aria-label` and `title`. I haven't raised a Leaflet issue on this yet... but see motivation in https://github.com/mapbox/mapbox-gl-js/issues/11036.

Falke-Design: So you suggest to remove `aria-label` from the PR? But will it then resolve the WCAG issues?

Malvoz:
> Resolves #6821

Can someone please modify the original comment to say that this also closes https://github.com/Leaflet/Leaflet/issues/7498? Thanks.

PR #6885, opened 2019-11-09
doc: run hljs on all documentation pages

### LINKED ISSUE #6878: Code blocks are broken in the docs
https://leafletjs.com/examples/quick-start/

Just noticed that code blocks on the website are broken — they have a redundant indentation at the beginning and no syntax highlighting. I wonder what might have broken this. cc @simon04 maybe something in your changes?

simon04:
Run hljs not only on reference pages, but on all documentation pages.

Fixes #6878. Regression of 9a43cf72d768aad58a7a116ad426ace905c64817.

@mourner

### DIFF
--- a/docs/docs/js/docs.js
+++ b/docs/docs/js/docs.js
@@ -1,3 +1,6 @@
+/* global hljs */
+hljs.configure({tabReplace: '    '});
+hljs.initHighlighting();
 
 var tocCopy = document.createElement('div');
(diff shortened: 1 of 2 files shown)


mourner:
Thanks! It looks like it doesn't fix the code blocks though for some reason — I still see the problem on all pages (including the reference page). Front page:

image

Reference page:
image

@simon04 any idea what might be wrong?

PR #6891, opened 2019-11-16
Update plugin entry for Leaflet.Geodesic

henrythasler:
I have refactored the plug-in with Typescript and added new examples. I also think that it fits better into the Markers & renderers section because it's primary purpose is displaying geodesic geometry (lines, circles).


henrythasler:
See [Leaflet.Geodesic](https://github.com/henrythasler/Leaflet.Geodesic) repo for details.

mourner:
Thanks!

Issue #6896, opened 2019-11-18
Zoom/pan not working on Windows 10 Touch in Chrome and Edge


sveine:
The fix #6855 had some side effects on Windows 10 Touch in Chrome and Edge. In Chrome and Edge zoom scales page. Pan works in Edge, but does nothing in Chrome. Firefox is OK.

max-collomb:
I think this is related with `L.Browser.pointer` that is now `false` where it used to be `true` on my chrome on windows (with no touch).

sveine:
On my Windows 10 Surface PC - Chrome, Firefox and Edge are working on the touch screen with this fix, except when I detach the screen it is very hard to zoom in Firefox. I think this is a minor problem. I think most people are using Edge or Chrome on touch screen. I could not find out how to fix this.

plepe:
Same problem, Chromium 78 on Ubuntu Linux 18.04. Zooming via touch gestures works with Leaflet 1.5.1, but not with 1.6.0. `L.Browser.pointer` is `false` on 1.6.

Touch gestures work with Firefox 70 with Leaflet 1.5.1 and 1.6.0.

ivangermanov:
Touch events on Chrome desktop (79.0.3945.79) not working here (Leaflet 1.6.0). I think it's something to do with how Leaflet internally checks if the browser supports touch. L.Browser.touch=undefined for me on a Windows 10 Surface Pro.

Peetz0r:
I can confirm this on Linux (Chromium 79 on Fedora 31) too. Touch on my convertible works on 1.5.1, broken on 1.6.0.

Firefox seems to not be affected, both 1.5.1 and 1.6.0 behave the same and touch works fine.

ETJ9914:
Still affected by this in Chrome using Leaflet 1.6.0, any idea when will it be fixed? Or other workarounds since L.Browser.touch is readonly so cannot be set manually.

johnd0e:
Dirty hack: set `window.ontouchstart="true"` before loading Leaflet.

But better go and up-vote #6977

sveine:
I had manually patch Leaflet 1.7.1 with my fix. Other fix did not solve this problem.

johnd0e:
> I had manually patch Leaflet 1.7.1 with my fix. Other fix did not solve this problem. 

What is 'your fix'?

sveine:
> 
> 
[...]

PR #6897, opened 2019-11-18
Zoom/pan not working on Windows 10 Touch in Chrome and Edge #6896

sveine:
I think the change for `export var touch =` is safe.

Not sure about the change in src/dom/DomEvent.js. Edge is based on webkit and `var pointer =` is not set in 1.6.0.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -100,7 +100,7 @@ export var pointer = !webkit && !!(window.PointerEvent || msPointer);
 // This does not necessarily mean that the browser is running in a computer with
 // a touchscreen, it only means that the browser is capable of understanding
 // touch events.
-export var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window ||
+export var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window || navigator.maxTouchPoints > 0 || navigator.msMaxTouchPoints > 0 ||
 		(window.DocumentTouch && document instanceof window.DocumentTouch));
 
 // @property mobileOpera: Boolean; `true` for the Opera browser in a mobile device.
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -81,7 +81,7 @@ function addOne(obj, type, fn, context) {
 
 	var originalHandler = handler;
 
-	if (Browser.pointer && type.indexOf('touch') === 0) {
+	if ((Browser.pointer || Browser.edge) && type.indexOf('touch') === 0) {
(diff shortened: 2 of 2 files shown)


src/dom/DomEvent.js
```diff
@@ -81,7 +81,7 @@ function addOne(obj, type, fn, context) {
 
 	var originalHandler = handler;
 
-	if (Browser.pointer && type.indexOf('touch') === 0) {
+	if ((Browser.pointer || Browser.edge) && type.indexOf('touch') === 0) {
```
ggcaponetto: I would say that `if (Browser.pointer && type.indexOf('touch') === 0) {`
is correct and should not be changed. 

The line i think we need to change is the line 92 of [https://github.com/Leaflet/Leaflet/blob/master/src/core/Browser.js](https://github.com/Leaflet/Leaflet/blob/master/src/core/Browser.js).
The line `export var msPointer = !window.PointerEvent && window.MSPointerEvent;` should be changed to
`export var msPointer = window.PointerEvent || window.MSPointerEvent;` The reason for this is that both Edge and IE11 have `window.PointerEvent` but only IE11 `window.MSPointerEvent`

You can test it out here: [https://github.com/ggcaponetto/Leaflet/tree/pull-6897-zoom-pan-win10-chrome](https://github.com/ggcaponetto/Leaflet/tree/pull-6897-zoom-pan-win10-chrome).

ggcaponetto:
Due to inactivity i am submitting my own PR based on your solution. Thanks @sveine .
[https://github.com/Leaflet/Leaflet/pull/6908](https://github.com/Leaflet/Leaflet/pull/6908)

src/core/Browser.js
```diff
@@ -100,7 +100,7 @@ export var pointer = !webkit && !!(window.PointerEvent || msPointer);
 // This does not necessarily mean that the browser is running in a computer with
 // a touchscreen, it only means that the browser is capable of understanding
 // touch events.
-export var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window ||
+export var touch = !window.L_NO_TOUCH && (pointer || 'ontouchstart' in window || navigator.maxTouchPoints > 0 || navigator.msMaxTouchPoints > 0 ||
```
johnd0e: Proposed change is not correct, as [`Browser.touch`](https://leafletjs.com/reference-1.6.0.html#browser-touch) has different meaning, and should not depend on `navigator.maxTouchPoints`.
Compare with #6978.

johnd0e:
Real cause of issue explained here: #6977 (fix iincluded).

PR #6899, opened 2019-11-19
Added link to uqm_map_full.png download

Lucas-Kohorst:


IvanSanchez:
I don't think this is needed. The link to the source of that image at http://www.highprogrammer.com/alan/games/video/uqm/index.html is better, and https://github.com/Leaflet/Leaflet/commit/214cd3e368e05274e9f2dba6516d30363d9168a4 should prevent any confusion.

PR #6905, opened 2019-11-20
[docs] Update Doc Examples to Use Modern Endpoint

riastrad:
Currently, a good portion of the examples on https://leafletjs.com rely on a Mapbox legacy static tiles endpoint:

image

This PR updates the code and references in these documentation examples to use the newer [Static Tiles](https://docs.mapbox.com/api/maps/#static-tiles) endpoint:

image

This also has the added benefit of updating all of the example pages with slick looking new map styles:

**before**
image

**after**
image

This branch is still a work in progress, and I'm planning to tag team the update with @rmrice this week. 

cc/ @mourner


rmrice:
@mourner sounds good. Just updated the all legacy static tile urls with the modern static api url, using `{id}` for style id. 

When I ran this locally I noticed the example.html for the choropleth example was missing, so I added it back. Not sure what happened since the map renders on the website.

All the maps render for the urls we changed, so this should be good to go 👍 

Please re-review when you get a chance! ❤️

rmrice:
Since I didn't author this PR, I can't mark it as ready for review or remove `[WIP]` from the title. But it's ready.

riastrad:
@mourner looks like @rmrice took care of this final thing! Should be good for a final review whenever you have a moment.

PR #6908, opened 2019-11-21
fix setting of touch capability flag of browsers - closes #6896 #6897

### LINKED ISSUE #6896: Zoom/pan not working on Windows 10 Touch in Chrome and Edge

ggcaponetto:
With this small changes we try to flag the Browser as being able to understand touch events.
With this fix its is now able to open the [leaflet quickstart](https://leafletjs.com/examples/quick-start/) and interact with touch events using chrome 78 and Edge.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -89,7 +89,7 @@ export var mobileWebkit3d = mobile && webkit3d;
 
 // @property msPointer: Boolean
 // `true` for browsers implementing the Microsoft touch events model (notably IE10).
-export var msPointer = !window.PointerEvent && window.MSPointerEvent;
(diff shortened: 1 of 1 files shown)


sveine:
This change made Firefox now stop working, and did not fix zoom in Edge. Edge has webkit and I think pointer/msPointer should not be used.

mourner:
Yes, we want to avoid touch events capable browsers using the pointer events route. See #6817

PR #6913, opened 2019-11-26
Updating use of UMD to include L.Class.Extend and error message if no L

mmnoo:


PLUGIN-GUIDE.md
```diff
@@ -194,7 +194,11 @@ You can add support for AMD/CommonJS loaders to your Leaflet plugin by following
 
     // attach your plugin to the global 'L' variable
     if (typeof window !== 'undefined' && window.L) {
-        window.L.YourPlugin = factory(L);
+        var leafletInstance = window.L;
+        leafletInstance.YourPlugin = leafletInstance.Class.Extend(factory(L));
```
johnd0e, line 198: Why do you need this?

mmnoo: I cant remember is great detail, but I think the original code didnt work. 

Was using it for this: https://github.com/mmnoo/leaflet-declarative-layers/blob/master/src/index.ts

johnd0e: Your code is basically wrong (for most cases), because plugins rarely are extended from `L.Class`. 
The purpose of other cited changes is also not clear (no sense to me).

Feel free to reopen the PR when you get some real use cases.

Issue #6051, opened 2018-02-12
Preventing change of map center upon `invalidateSize`


SylvainCorlay:
(With the latest leaflet.js and earlier versions, all browsers and platforms)

`invalidateSize` causes the `center` to change even when passing `{pan: false, animate: false}` as options. This happens upon the firing of the `moveend` event in the end of the implementation of `invalidateSize`.

Is there a means to get the tiles to be downloaded accordingly when the map container is resized without changing the `center`?

The motivation is for the integration of leafletjs with JupyterLab. Multiple views of the same map model can be displayed. Model data includes zoom level and center, while the views may have different sizes. Invalidating the size of one view causes the center to change and events to bounce from one view to another until the call stack becomes too big. Ideally, one should be able to `invalidateSize` without changing the center.

![views](https://user-images.githubusercontent.com/2397974/36124266-b67206be-104f-11e8-9efe-c99b5dd731c2.gif)

SylvainCorlay:
This is related to https://github.com/Leaflet/Leaflet/issues/1919

In short, is there a means to invalidate the size without changing the center?

Malvoz:
See also https://github.com/Leaflet/Leaflet/issues/5008.

Falke-Design:
Doesn't happen in v1.9.2 anymore.

Issue #5030, opened 2016-10-21
Unable to disable pan while calling the invalidateSize method on the map object with options.pan = false


vaishaks:
I'm trying to disable panning while calling invalidateSize on the map by doing the following.
`map.leafletElement.invalidateSize({pan: false})`
The map however, still continues to pan.

IvanSanchez:
Can you publish a fiddle/codepen/[playground](https://playground-leaflet.rhcloud.com/) that displays this behaviour?

PR #6917, opened 2019-11-29
Fix invalidateSize re-centering pan logic

### LINKED ISSUE #5008: map.invalidateSize() changes map center
## How to reproduce
- Leaflet version I'm using: 1.0.1
- Browser (with version) I'm using: Chrome  51.0.2704.63 (64-bit) 
- OS/Platform (with version) I'm using: Ubuntu 16.04
### Step 1

Create html file with this content:

``` html
<!DOCTYPE html>
<html>
<head>
    <title>Leaflet</title>
    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.0.1/dist/leaflet.css" />
    <script src="https://unpkg.com/leaflet@1.0.1/dist/leaflet.js"></script>
</head>
<body>
<div id="mapid" style="position: absolute; top: 0; left: 0; width: 100%; height: 100%;"></div>
<script>
    window.map = L.map('mapid').setView([54.98, 82.89], 13);
    L.tileLayer('http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
        attribution: ''
    }).addTo(map);
</script>
</body>
</html>
```
### Step 2

In browser console run `map.setView(map.getCenter())`
##### This depends on the window size. If not repeat resize the browser window.

Problems in file `Map.js:1482`.
Maybe we can compare new center with current in function `setView` like this

``` js
if (this._loaded) {
            var currentCenter = this.getCenter();
            var newCenter = L.latLng(center);
            if (newCenter.lat === currentCenter.lat
                && newCenter.lng === currentCenter.lng
                && zoom === this.getZoom()) {
                this.fire('movestart');
                this.fire('moveend');
                return this;
            }
        }
```

But in events will be wrong center.

natevw:
The map currently pans in the wrong direction when its size is updated, shifting the center farther away from where it should be, rather than back to where it was originally. This inverts the direction of the `offset` passed to the `panBy`/`_rawPanBy` methods.

Might fix #5008 (and/or #6051, #5030).

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -584,7 +584,7 @@ export var Map = Evented.extend({
 		var newSize = this.getSize(),
 		    oldCenter = oldSize.divideBy(2).round(),
 		    newCenter = newSize.divideBy(2).round(),
-		    offset = oldCenter.subtract(newCenter);
+		    offset = newCenter.subtract(oldCenter);
 
 		if (!offset.x && !offset.y) { return this; }
 


natevw:
Bad fix, works for the bug I'm chasing down when map is coming from `display: none` but not in most other cases.

PR #6923, opened 2019-11-29
Improve Evented.off performance by using the Map class (fixes #6914)

### LINKED ISSUE #6914: Bad performance of Evented.off()
Problem:
Evented.off() is relatively expensive as it has to go thru all subscription to find the one that has to be removed. For me this is a problem, because we have up to 10k markers in the viewpoint at a time. As you zoom out the layer the markers are on gets removed which also removes the markers. In my case this can take up to 10s. A huge proportion of scripting time is taken up by Evented.off() (around 90%)

Solution:
I already changed the implementation of Evented to use a Dictionary for the events instead of an Array. This boosts the performance of Off() from O(n) to O(1). To this a few adjustments to the other functions of Evented had to be done so that all requirements are still met and there performance is affected as little as possible.

sidney-pauly:
Added the Dictionary Class:
Added a new class called Dictionary (name borrowed from c#). The dictionary adds the ability to store key/value pairs with high performance ([Spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size))
To make the class compatible with older java script versions  it evaluates if Map gets implemented and if not uses a fallback version which I created (The fallback version currently only contains the methods I needed for my implementation and not the full spec, which could be added later)  

Changed the implementation of Evented to use the Dictionary Class
Evented now uses the new Dictionary class. As a result the performance of Off() is greatly improved (from O(n) to O(1))

Considerations:
- I'm unsure how the Comment/Doumentation system works if somone can provide an explenation I will add the documentation for the Dictionary class.

- Eventhough all tests for the Evented Class where successfull a few tests failed, but they seem unrelated to my changes (Maybe my local testing enviroment is setup wrong?):

> Chrome 78.0.3904 (Windows 10.0.0) CRS #zoom && #scale convert zoom to scale and viceversa and return the same values FAILED
>         Error: expected 2.4999999999999996 to sort of equal 2.5
>             at Assertion.assert (node_modules/expect.js/index.js:96:13)
>             at Assertion.eql (node_modules/expect.js/index.js:230:10)
>             at Context.<anonymous> (spec/suites/geo/CRSSpec.js:233:31)
> ................................................................................
> ................................................................................
> ........................................................................
> Chrome 78.0.3904 (Windows 10.0.0) Tooltip is opened when tapping on touch FAILED
>         Error: expected false to equal true
>             at Assertion.assert (node_modules/expect.js/index.js:96:13)
>             at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
>             at Assertion.<computed> [as be] (node_modules/expect.js/index.js:69:24)
>             at Context.<anonymous> (spec/suites/layer/TooltipSpec.js:262:43)
> Chrome 78.0.3904 (Windows 10.0.0) Tooltip is closed if not permanent when clicking on the map elsewhere on touch FAILED
>         Error: expected false to equal true
>             at Assertion.assert (node_modules/expect.js/index.js:96:13)
>             at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
>             at Assertion.<computed> [as be] (node_modules/expect.js/index.js:69:24)
>             at Context.<anonymous> (spec/suites/layer/TooltipSpec.js:270:43)
> ......
> Chrome 78.0.3904 (Windows 10.0.0) Marker.Drag drag in CSS scaled container drags a marker with mouse, compensating for CSS scale FAILED
>         Error: Uncaught TypeError: Cannot read property 'dispatchEvent' of null (node_modules/prosthetic-hand/dist/prosthetic-hand.js:1071)
> .............................................
> Chrome 78.0.3904 (Windows 10.0.0) GridLayer number of 256px tiles loaded in synchronous animated grid @800x600px Loads 32, unloads 16 tiles zooming out 11-10 FAILED
>         Error: Uncaught Error: expected 20 to equal 32 (node_modules/expect.js/index.js:102)
> ...
> Chrome 78.0.3904 (Windows 10.0.0) GridLayer configurable tile pruning Loads map, moves forth and back by 512 px, keepBuffer = 0 FAILED
>         Error: Uncaught Error: expected 16 to equal 28 (node_modules/expect.js/index.js:102)
> ............................
> Chrome 78.0.3904 (Windows 10.0.0) Canvas #events should not fire click when dragging the map on top of it FAILED
>         Error: Uncaught Error: expected true to equal false (node_modules/expect.js/index.js:102)
> ............................................................
> WARN: 'Deprecated use of _flat, please use L.LineUtil.isFlat instead.'
> .
> WARN: 'Deprecated use of _flat, please use L.LineUtil.isFlat instead.'
> ................................................................................
> ...................................................
> Chrome 78.0.3904 (Windows 10.0.0) Map.Drag mouse events change the center of the map FAILED
>         Error: Uncaught Error: expected 0 to be within 21.943..21.9431 (node_modules/expect.js/index.js:102)
> ......
> Chrome 78.0.3904 (Windows 10.0.0) Map.Drag touch events change the center of the map FAILED
>         Error: Uncaught Error: expected 0 to be within 21.943..21.9431 (node_modules/expect.js/index.js:102)
> .
> Chrome 78.0.3904 (Windows 10.0.0) Map.Drag touch events reset itself after touchend FAILED
>         Error: Uncaught Error: expected { lat: 0, lng: 0 } to sort of not equal { lat: 0, lng: 0 } (node_modules/expect.js/index.js:102)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1782,6 +1782,15 @@ These plugins provide new markers or news ways of converting abstract data into 
 	</tr>
 	<tr>
 		<td>
+			<a href="https://github.com/henrythasler/Leaflet.Geodesic">Leaflet.Geodesic</a>
+		</td><td>
+			Draw geodesic lines and circles. A geodesic line is the shortest path between two given points on the earth surface. It uses Vincenty's formulae for highest precision and distance calculation. Written in Typescript and available via CDN. <a href="https://blog.cyclemap.link/Leaflet.Geodesic/complex-interactive.html">Demo</a>
+		</td><td>
+			<a href="https://github.com/henrythasler">Henry Thasler</a>
+		</td>
+	</tr>
+	<tr>
+		<td>
 			<a href="https://github.com/nuclearsecrecy/Leaflet.greatCircle">Leaflet.greatCircle</a>
 		</td>
 		<td>
@@ -4094,15 +4103,6 @@ The following plugins perform several sorts of geoprocessing (mathematical and t
 		</td>
 		<td>
 			<a href="https://github.com/w8r">Alexander Milevski</a>
-		</td>
-	</tr>
-	<tr>
-		<td>
-			<a href="https://github.com/henrythasler/Leaflet.Geodesic">Leaflet.Geodesic</a>
-		</td><td>
-			Draw geodesic (poly)lines. A geodesic line is the shortest path between two given positions on the earth surface. and You can also calculate the exact distance between two given points on the map.
-		</td><td>
-			<a href="https://github.com/henrythasler">Henry Thasler</a>
 		</td>
 	</tr>
 	<tr>
--- a/src/core/Dictionary.js
+++ b/src/core/Dictionary.js
@@ -1,0 +1,50 @@
+import {Class} from './Class';
+
+var _Dictionary = Class.extend({
+	initialize: function () {
+		this.map = {};
+		this.ordering = [];
+		this.size = 0;
+	},
+
+	get: function (key) {
+		return this.map[key];
+	},
+
+	set: function (key, value) {
+		if (!this.map[key]) {
+			this.map[key] = value;
+			this.size = this.ordering.push(key);
+		}
+	},
+
+	delete: function (key) {
+		delete this.map[key];
+		var indexToRemove = this.ordering.indexOf(key);
+		this.ordering.splice(indexToRemove, 1);
+		this.size = this.ordering.length;
+	},
+
+	forEach: function (delegate, ctx) {
+
+		if (!ctx) {
+			ctx = this;
+		}
+
+		this.ordering.forEach(function (e) {
+			var element = this.map[e];
+			delegate.call(ctx, element);
+		}, this);
+
+	}
+});
+
+// Check if map gets implemented nativly (ecma6 implementation)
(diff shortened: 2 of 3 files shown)


IvanSanchez:
What's the problem that this intends to fix? In which scenarios is the `O(n)` complexity of `Evented.off()` an issue?

sidney-pauly:
It becomes a problem if you call off on a event that has a lot of listeners. Each time you call Evented.off() the code iterates over all subscriptions to the event to find the one that has to be removed. If you have a lot of subscriptions to one event and want to remove a lot of them the performance gets quite poor. In our case we had to remove around 10.000 listeners from an event with around 50.000 listeners. This equates to roughly 500.000 iterations over the listener array. With the improvement only 10.000 iterations are needed. For us that meant an improvement from 10s waiting time to around 0.2s waiting time.

muhamedkarajic:
Would you merge it if we make a new PR for the current version?

jonkoops:
Sorry for not getting back to this PR in a while, I think we certainly can see if we can merge this in. The current implementation has a lot of fallback logic, as `Map` was not supported at the time. For version 2.x of Leaflet, we will be using modern JavaScript, so I think this PR can be trimmed quite a bit.

If there is still interest in landing this let us know, I will try to be available to review any work done.

PR #6928, opened 2019-12-03
Add crs property to LatLng

Schleuse:
`LatLng` currently assumes the Earth CRS for functions like: `distanceTo`, `wrap`, `toBounds` which is not always correct (See #4978 for example).

This pull request adds a `crs`-property to `LatLng` which enables offloading these functions to the CRS instance. The default remains the Earth CRS.

### DIFF
--- a/spec/suites/geo/CRSSpec.js
+++ b/spec/suites/geo/CRSSpec.js
@@ -221,6 +221,16 @@ describe("CRS.Simple", function () {
 			expect(crs.wrapLatLng(new L.LatLng(300, -250))).nearLatLng(new L.LatLng(-100, 150));
 		});
 	});
+
+	describe("extendLatLng", function () {
+		it("should extend a given LatLng by the number of units in all directions", function () {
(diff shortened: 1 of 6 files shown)


Schleuse:
I'm currently adding the crs as a constructor argument which I'm not particularly fond of. Another idea I've had, is to create subsclasses of `LatLng` to accommodate the CRS (eg. `LatLng.Earth`, `LatLng.Simple`)

johnd0e:
@Schleuse
I see here a lot of random commits. Could you clean up your branch?

Schleuse:
@johnd0e I've cleaned everything up :+1:

johnd0e:
I'm not sure that this is useful addition to Leaflet api, in proposed form.
1. What if some feature (e.g. Circle) is created using LatLng with one CRS but then added to map with another CRS?
2. What if some feature (e.g. Polyline) is created form several LatLngs with different CRS?
3. Is it ever useful to have different LatLngs with different CRS properties on the same map?

Issue #2934, opened 2014-10-06
User-initiated zoom events?


geleto:
Hello,
It seems that leaflet has no specific event, which is triggered when the user zooms the map, but not when the zoom level was changed programmatically.
Maybe dragstart should be triggered when pinch-zooming because this usually changes the map position too.
I use the leaflet-locatecontrol and I need to stop tracking the user movements when he drags or pinch-zooms the map.

Grsmto:
While creating a custom layer I found myself having to hack the lib to have access to zoom events.
I needed to know to which zoom level the map was animating, zoom property is updated only on `zoomend` event, which is too late if you want to animate anything custom at the same time as the map animation.
If anyone is interested by my solution, what I did is firing a `zoom` event:
`this.fire('zoom', {zoom: zoom});` in `setZoom()` and `setZoomAround()` functions from `L.Map`.

Can be related to https://github.com/Leaflet/Leaflet/issues/3269

dmitry:
I have the same situation, when I need to prevent from invoking `fitBounds` while user interacts with the map. Looks like it could be fixed while https://github.com/Leaflet/Leaflet/pull/3278, but it's already was merged. I believe it could be named something like `triggerSource` with values `user` and `program`, or just a boolean `triggeredByUser`.

dmitry:
Additional information, but related to the google maps: http://stackoverflow.com/questions/10097647/how-to-listen-for-user-generated-zoom-in-google-maps

dmitry:
Just for anyone, who stuck with the same issue, here are the approach I've used in the project:

```javascript
  onDragEnd() {
    this.onUserChange();
  },
  onZoomEnd() {
    if (this.isUser) {
      this.onUserChange();
    }
  },
  onMoveEnd() {
    this.isUser = true;
  },
```

bindings:

```javascript
    this.map.on('dragend', this.onDragEnd);
    this.map.on('zoomend', this.onZoomEnd);
    this.map.on('moveend', this.onMoveEnd);
```

Would be nice to replace `onMoveEnd` to `idle` (https://github.com/Leaflet/Leaflet/issues/3178), but for my case that's enough.

bpinney:
Any update on when this might be added?

bpinney:
#2267 and #2532 are open issues making the same request

yohanboniface:
The fastest way of getting this into core is a pull request ;)
[...]

PR #6929, opened 2019-12-04
add source to movestart, move, moveend, zoomstart, zoom and zoomend events

Athorcis:
It would be very useful to know when the map is moved or zoomed due to user interaction.

So here‘s a pull request which adds a 'source' property to movestart, move, moveend, zoomstart, zoom, and zoomend events.
The source is either `user(.[a-z]+)?` or `imperative(.[a-z]+)?`, depending on how was triggered the event.

This pull request is mostly based on this one #5247 which is three years old and was never merged for some unknown reason.

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -69,13 +69,13 @@ export var Zoom = Control.extend({
 
 	_zoomIn: function (e) {
 		if (!this._disabled && this._map._zoom < this._map.getMaxZoom()) {
-			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomIn(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1), 'user.zoomin');
 		}
 	},
 
 	_zoomOut: function (e) {
 		if (!this._disabled && this._map._zoom > this._map.getMinZoom()) {
-			this._map.zoomOut(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1));
+			this._map.zoomOut(this._map.options.zoomDelta * (e.shiftKey ? 3 : 1), 'user.zoomout');
 		}
 	},
 
--- a/src/dom/PosAnimation.js
+++ b/src/dom/PosAnimation.js
@@ -22,12 +22,12 @@ import * as DomUtil from '../dom/DomUtil';
 
 export var PosAnimation = Evented.extend({
 
-	// @method run(el: HTMLElement, newPos: Point, duration?: Number, easeLinearity?: Number)
+	// @method run(el: HTMLElement, newPos: Point, duration?: Number, easeLinearity?: Number, source?: String)
 	// Run an animation of a given element to a new position, optionally setting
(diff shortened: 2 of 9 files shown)


IvanSanchez:
As [I said before](https://github.com/Leaflet/Leaflet/issues/2934#issuecomment-512766397), I'm not convinced by this approach. Something in the back of my brain tells me that there's gonna be some edge case where an event handler calls another event handler and things go inconsistent.

This would be better with some unit tests, specifically in the parts where we test user interaction via touch and mouse drag, e.g. https://github.com/Leaflet/Leaflet/issues/2934#issuecomment-512766397. The code in the PR looks good, but unfortunately doesn't give me any assurance that it'll work.

Athorcis:
Wait before merging this pull request, I want to refactor these changes, instead of adding a parameter to a lot of methods, I'm going to use the parameter options. This way all the plugins which redefine methods of Leaflet will continue to work.

rdev-software:
Any update on this ?

rdev-software:
I've merged it on my fork, and it works fine.
If there is any other approach how to detect user, please let me know

wieringen:
Any update? Would be nice to have this feature.

AnthonyZink:
Any updates on this ? Would be great !

johnd0e (commented):
Just random thoughts:

johnd0e:
And all this should be covered by tests..

nasainsbury:
Any update on this? Very interested and happy to investigate, wasn't sure if this is just being disregarded now in favour of other methods of tracking event source?

lhoosconciso:
I need this feature as well. I've seen posts more than 10 years old from people asking for this, and I feel this is a rather essential.

fcFn (commented):
Oops, finishing my review from 2019. I didn't realize pending comments aren't seen.

Falke-Design (changes requested):
To merge this PR or a new one from someone else, a few things are needed:

1. tests for every case
2. Check if every function call has the source included correctly
3. Check if every event is included
4. Discussion why not on every event including the source? would it be possible?
5. Are there situations where the `user` is the source but because of some reason, for example event chaining, the source gets lost?

What are the usecases / why do you need the source?

gilest:
> What are the usecases / why do you need the source?

For my purpose – some map moves occur due to Leaflet features, like ensuring that a marker popup is in view. Others occur due to user input.

Being able to determine which is which is very helpful. For example in my use-case we only want to load new map pins when the user moved the map, not due to an automated move.

BenoitBouillard:
I’m working with a similar use case.

In GPS follow-up mode, I use setView to update the map location based on the current GPS position. However, calling setView triggers map movement events (movestart, move, moveend).

To avoid interfering with user interactions, I temporarily disable GPS follow-up when the user manually moves the map. I rely on those movement events to detect user actions and reset a reactivation timer.

The challenge is to distinguish between map movements triggered by the GPS (setView) and those initiated by the user. I need to deactivate the GPS follow-up only when the movement is user-driven, not system-driven.

FYI, in d3js, they attached the sourceEvent to the event in order to distinguish them: https://d3js.org/d3-zoom#zoom-events

Issue #662, opened 2012-05-03
Printing and pagebreaks in IE


DennisSchiefer:
Trying to print a page that contains a Leaflet map can yield unexpected results.

Consider following example:
A Leaflet map is placed into a div. Before the map, there some other elements in the html code, among those is another div with the style attribute "page-break:after;" set. When printing from IE, the map is correctly rendered on the second page, but on the bottom of the first page a part of the map is rendered, too. That part a) should normally not be shown at all - it is outside of the map bounds and b) does not respect any settings of the div (height, width, possible border,...).

Issue #6664, opened 2019-05-14
Polygon with no points causes javascript error


captain-igloo:
- [*] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [*] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [*] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
1. create a polygon with no points
2. add it to the map
3. call setStyle({weight: 1})

See https://jsfiddle.net/q5x3bsy0/

**Expected behavior**
I expect no javascript error.  There is no error with leaflet 1.4.0.

**Current behavior**
This results in a javascript error: "Cannot read property 'min' of undefined"

**Environment**
- Leaflet version:1.5.1
- Browser (with version): Chromium 73
- OS/Platform (with version): Ubuntu Linux 18.04

**Additional context**
Possibly a duplicate of https://github.com/Leaflet/Leaflet/issues/6662

**Minimal example reproducing the issue**
```
var mymap = L.map('map');
var polygon = L.polygon([]);
polygon.addTo(mymap);
polygon.setStyle({
	weight: 1
});
```
Please create an example using https://leafletjs.com/edit.html or any other jsfiddle-like site.
https://jsfiddle.net/q5x3bsy0/
- [*] this example is as simple as possible
- [*] this example does not rely on any third party code

opoto:
Same issue here.
I confirm the proposed fix from #6662 solves it on my side.

PR #6941, opened 2019-12-14
Fix issue #6662 and #6664

NielsHolt:
This is a new PR replacing #6671

### DIFF
--- a/src/layer/vector/Path.js
+++ b/src/layer/vector/Path.js
@@ -106,7 +106,7 @@ export var Path = Layer.extend({
 		if (this._renderer) {
 			this._renderer._updateStyle(this);
 			if (this.options.stroke && style && style.hasOwnProperty('weight')) {
-				this._updateBounds();
+				this._project();
 			}
 		}
 		return this;


NielsHolt:
Hi @mourner and @IvanSanchez 
Do you know when you expect this PR to be included in a patch?
Kind regards
Niels

naedx:
> Hi @mourner and @IvanSanchez
> Do you know when you expect this PR to be included in a patch?
> Kind regards
> Niels

Hi guys, are there any updates on merging this?

NielsHolt:
Hi @mourner and @IvanSanchez 
I see that this PR was not included in 1.7.1
Do you know when you expect this PR to be included in a patch?
It fixes a bug that was introduced in version 1.5 so I would urge you to please include it in the next patch.

Kind regards
Niels

johnd0e:
@NielsHolt 
1. Resolve conflicts
2. Add unit test

spec/suites/layer/vector/PolylineSpec.js
```diff
@@ -212,4 +212,15 @@ describe('Polyline', function () {
...
 		});
 	});
+
+	describe("#style", function () {
+		it("can set weight after being added to the map", function () {
+			var polyline = new L.Polyline([]).addTo(map);
+			polyline.setStyle({
+				weight: 3,
+			});
+			expect(map.hasLayer(polyline)).to.be(true);
```
johnd0e, line 222: No need to check this

spec/suites/layer/vector/PolylineSpec.js
```diff
@@ -212,4 +212,15 @@ describe('Polyline', function () {
 			expect(polyline._latlngs).to.eql([L.latLng([1, 2])]);
 		});
 	});
+
+	describe("#style", function () {
+		it("can set weight after being added to the map", function () {
+			var polyline = new L.Polyline([]).addTo(map);
```
johnd0e, line 218: In #6662 you were trying to use `setStyle` before adding to map.
But here you do it after adding.
Does it actually matter?

spec/suites/layer/vector/PolylineSpec.js
```diff
@@ -212,4 +212,15 @@ describe('Polyline', function () {
 			expect(polyline._latlngs).to.eql([L.latLng([1, 2])]);
 		});
 	});
+
+	describe("#style", function () {
+		it("can set weight after being added to the map", function () {
```
johnd0e, line 217: It is still not completely clear to me.
1. You've stated that this PR fixes #662 (_Error when calling Path.setStyle **before** added to the map_)
2. But here you are testing it **after** adding to map.

NielsHolt: Hi @johnd0e 
I am not sure what you mean. I have forked leaflet to fix a bug (one line) in Path.setStyle that was introduced in version 1.5. All the test etc. are unchanged

johnd0e: > All the test etc. are unchanged

I refer to 66ed2a88155699896205cb695ff508abcb8cc602 by @fodor0205

fodor0205:
The error only occurs with the following order:

1. Polyline added to the map
2. `setStyle()` called with weight

The description in #6662 does not seem to be correct, because no error occurs if `setStyle` is called before `addTo()`.

johnd0e:
> The description in #6662 does not seem to be correct, because no error occurs if `setStyle` is called before `addTo()`.

But it is not about description only, as there is code in **steps to reproduce** section:
```js
var polyline = L.polyline([[45.51, -122.68],[37.77, -122.43],[34.04, -118.2]]);
polyline.setStyle({color: 'red', weight: 4});
polyline.addTo(map);
```
And I do not see any problem even if I call `setStyle` after adding to map.

johnd0e:
So I am able to see that only #6664 issue is relevant, and there described slightly different case:

- polygon/polyline is empty
- `setStyle` is called after feature added to map.

Ok, then I have more questions:
1) As test is not specific to Polyline, shouldn't we place to place into `PathSpec.js`?
2) Doesn't other shapes suffer from the issue?
3) Isn't there some performance loss on replacing `this._updateBounds` with `_project` call?
I'm afraid that here we are fixing only **empty** polygon/line - very particular and unlikely case - at the cost of overall performance of every `L.Path`'s descendant.

fodor0205:
> As test is not specific to Polyline, shouldn't we place to place into PathSpec.js?

No, because other descendants of `Path` are not affected, as described below.

> Doesn't other shapes suffer from the issue?

No. 2 vector overlays extend Path:

- [Polyline](src/layer/vector/Polyline.js)

  It has the unit test which perfectly triggers the bug.

- [CircleMarker](src/layer/vector/CircleMarker.js)

  Is not affected, as this reference to undefined variable `_this._rawPxBounds` is located in Polyline's [_updateBounds()](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/layer/vector/Polyline.js#L216) function, not inside the Path class.

There is another possible fix without touching `Path.js`:
https://github.com/Leaflet/Leaflet/compare/master...fodor0205:master

> Isn't there some performance loss on replacing this._updateBounds with _project call? I'm afraid that here we are fixing only empty polygon/line - very particular and unlikely case - at the cost of overall performance of every L.Path's descendant.

The alternative fix linked above only changes the offending code-path, making less impact in other use cases.

@johnd0e @NielsHolt Could you please review this alternative fix?

johnd0e:
> The alternative fix linked above only changes the offending code-path, making less impact in other use cases.

I still doubt that it makes sense to project empty polyline.
How about simpler fix:
```js
		if (!this._rawPxBounds) { return; }
```

fodor0205:
This solution makes more sense @johnd0e, I have updated my branch, waiting for @NielsHolt to [pull it into his branch](https://github.com/FCOO/Leaflet/pull/3), so it will be available in this PR.

NielsHolt:
Hi 
Thanks! The fix from @fodor0205 seems to fix the issue.
I have merged the PR into my master, but I am not sure what to do next except wait for the leaflet guys to include the PR?

fodor0205:
@johnd0e do you think this is ready to be merged?

johnd0e:
Waiting for https://github.com/FCOO/Leaflet/pull/4

fodor0205:
What's the next step? @johnd0e

johnd0e (approved):
It is ready for merge, so waiting @mourner or @IvanSanchez.

contibru:
Is there any due date for this issue?

PR #6948, opened 2019-12-20
Remove tooltips outside screen bounds. Resolves #6734

### LINKED ISSUE #6734: Tooltips are not culled when off screen causing performance losses

**Steps to reproduce**
- Add and open several permanent tooltips to e.g. geoJSON features on map
- Zoom in far enough that most tooltips are out of view
- Move map to put other geoJSON polys in view and observe that it takes a moment for the vector to pop in but the tooltip was already present
- ...

**Expected behavior**

Tooltips, like overlays and rasters, should be culled like vectors and rasters when off-screen for performance reasons
**Current behavior**

Tooltips are always 'drawn' and in the DOM even when well out of view in the map
**Environment**
- Leaflet version: 1.5.1
- Browser (with version): Chrome 75.0.3770.142
- OS/Platform (with version): MacOS 10.14.5

**Additional context**
Currently using Leaflet for displaying a lot of data simultaneously on 300+ open permanent tooltips bound to geoJSON features. I currently manually switch permanent to false on these tooltips when it is zoomed out far enough to become unperformant (having more than, say 30, on screen at once), but when zoomed in all are open and permanent, most totally out of view. This causes a major performance hit (+20-50% CPU usage spike on a scary fast iMac) because they aren't culled when out of the viewport like the raster and overlay layers. My intuition says that when these tooltips are bound to an overlay like a geoJSON feature and that geoJSON feature is temporarily culled, so too should the tooltip.

rowanwins:
Removes the tooltip of a polygon layer when the polygon bounds does not intersect with the renderer bounds.

Resolves #6734

Screen_20

### DIFF
--- a/src/layer/vector/Polygon.js
+++ b/src/layer/vector/Polygon.js
@@ -131,14 +131,18 @@ export var Polygon = Polyline.extend({
 		this._parts = [];
 		if (!this._pxBounds || !this._pxBounds.intersects(bounds)) {
 			for (const lyr in this._eventParents) {
-				this._eventParents[lyr].closeTooltip();
+				if (this._eventParents[lyr].getTooltip()) {
+					this._eventParents[lyr].closeTooltip();
+				}
 			}
 			return;
 		}
 
 		for (const lyr in this._eventParents) {
-			if (!this._eventParents[lyr].isTooltipOpen()) {
-				this._eventParents[lyr].openTooltip();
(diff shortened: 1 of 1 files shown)


johnd0e:
Why polygons only are handled?

Malvoz:
@rowanwins Do you intend to continue working on this, to address the suggested changes?

See also https://github.com/Leaflet/Leaflet/issues/6734#issuecomment-1133961647.

PR #6951, opened 2019-12-25
CSS: ref marker-icon-2x.png and marker-shadow.png

simon04:
This enables automatic asset handling using webpack and its css-loader.

Relates to #4849, #4979

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -396,6 +396,12 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-default-icon-path {
 	background-image: url(images/marker-icon.png);
 	}
+.leaflet-default-icon-path-2x {
+	background-image: url(images/marker-icon-2x.png);
+	}
+.leaflet-default-icon-path-shadow {
+	background-image: url(images/marker-shadow.png);
+	}
 
 
 /* attribution and scale controls */
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -31,30 +31,32 @@ export var IconDefault = Icon.extend({
 	},
 
 	_getIconUrl: function (name) {
-		if (!IconDefault.imagePath) {	// Deprecated, backwards-compatibility only
-			IconDefault.imagePath = this._detectIconPath();
+		 // IconDefault.imagePath is deprecated, backwards-compatibility only
+		if (!IconDefault.imagePath && !IconDefault._imagePathDetected) {
+			this._detectIconPath('iconUrl', 'leaflet-default-icon-path');
+			this._detectIconPath('iconRetinaUrl', 'leaflet-default-icon-path-2x');
+			this._detectIconPath('shadowUrl', 'leaflet-default-icon-path-shadow');
 		}
 
 		// @option imagePath: String
 		// `Icon.Default` will try to auto-detect the location of the
 		// blue icon images. If you are placing these images in a non-standard
 		// way, set this option to point to the right path.
-		return (this.options.imagePath || IconDefault.imagePath) + Icon.prototype._getIconUrl.call(this, name);
+		return (this.options.imagePath || IconDefault.imagePath || '') + Icon.prototype._getIconUrl.call(this, name);
 	},
 
-	_detectIconPath: function () {
-		var el = DomUtil.create('div',  'leaflet-default-icon-path', document.body);
+	_detectIconPath: function (property, className) {
+		var el = DomUtil.create('div',  className, document.body);
 		var path = DomUtil.getStyle(el, 'background-image') ||
 		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
 
 		document.body.removeChild(el);
 
 		if (path === null || path.indexOf('url') !== 0) {
-			path = '';
-		} else {
-			path = path.replace(/^url\(["']?/, '').replace(/marker-icon\.png["']?\)$/, '');
+			return;
 		}
-
-		return path;
+		path = path.replace(/^url\(["']?/, '').replace(/["']?\)$/, '');
+		IconDefault.prototype.options[property] = path;
+		IconDefault._imagePathDetected = true;
 	}
 });


johnd0e:
This is not needed, as `.leaflet-default-icon-path` class is used **only** in path detection heuristic.

simon04:
In fact some treatment is needed. Either the relevant files (including marker-icon-2x.png and marker-shadow.png ) are referenced in the stylesheet, or the consumers have to adapt their build system (such as webpack) to copy the dangling, but needed files marker-icon-2x.png and marker-shadow.png

johnd0e:
Ok. I have no webpack setup to check this. 
Cound do provide sample webpack setup to test?

simon04:
Here is a SSCCE: https://gist.github.com/simon04/3b89aa6ca858c72110bb80dd13b4f941

johnd0e:
@simon04 
With your PR all images are included, with random names.
But this is not enough for your sample to work properly.

johnd0e:
So I can conclude that such approach makes sense.
But we do not actually need separate styles for that, so it's possible to use this:
```css
/* Default icon URLs */
.leaflet-default-icon-path {
	/* used only in path-guessing heuristic, see L.Icon.Default */

	/* here are images that is accessed from js only */
	/* we need to list them explicitly to instruct webpack file-loader to include them */
	background-image: url(images/marker-shadow.png);
	background-image: url(images/marker-icon-2x.png);
	background-image: url(images/marker-icon.png);
	}
```

johnd0e:
Still, I have some concerns.
AFAIK, in some other setup webpack may include images inline, with data-url, like here: https://github.com/Leaflet/Leaflet/issues/6496#issuecomment-524736021.

With your approach all 3 icons will be inlined into css, absolutely pointlessly increasing bundle's size.

simon04:
Hey @johnd0e, I've updated this PR according to the discussion in https://github.com/Leaflet/Leaflet/pull/7092#issuecomment-724591326

To test bundling with esbuild, build leaflet, and run `echo 'require("./dist/leaflet-src.js"); require("./dist/leaflet.css")' | npx esbuild --bundle --outdir=dist.esbuild/ --loader:.png=file` 

You should obtain the following files

```
$ find dist.esbuild/
dist.esbuild/
dist.esbuild/layers-2x.CUVBMIZT.png
dist.esbuild/layers.ZHTVFDSJ.png
dist.esbuild/marker-icon-2x.Z45FGZMW.png
dist.esbuild/marker-icon.MCUQXS5S.png
dist.esbuild/marker-shadow.PNVI35RZ.png
dist.esbuild/stdin.css
dist.esbuild/stdin.js
```

The stylesheet should reference the correct .png files

```
$ grep png dist.esbuild/stdin.css
  background-image: url(layers.ZHTVFDSJ.png);
  background-image: url(layers-2x.CUVBMIZT.png);
  background-image: url(marker-icon.MCUQXS5S.png);
  background-image: url(marker-icon-2x.Z45FGZMW.png);
  background-image: url(marker-shadow.PNVI35RZ.png);
```

johnd0e:
@simon04
What if we have many markers with different icons?

simon04:
What do you mean with "different icons"? Leaflet only ships one according to https://unpkg.com/browse/leaflet@1.7.1/dist/images/.

johnd0e:
But `L.Marker` is designed to support any user icon. And `_getIconUrl` will be used for any user-supplied image path, so that should be tested too.
For example: it is common practice - to put extra marker icons into Leaflet `images` directory (or to some relative path).

simon04:
Is this a specified behaviour? https://leafletjs.com/reference-1.7.1.html#icon-iconurl defines `iconUrl` – "The URL to the icon image (absolute or relative to your script path)."

johnd0e:
Perhaps this is.
But better to check if precise wording completely corresponds to actual behavior.

simon04:
`Icon.Default.imagePath` only mentions the blue icon, "Icon.Default will try to auto-detect the location of the blue icon images.", see https://leafletjs.com/reference-1.7.1.html#icon-default-imagepath

You're suggesting to augment this in the following way: when the default marker-icon is in `foobar/marker-icon.MCUQXS5S.png`, then a different icon having `iconUrl=baz.png` should be taken from `foobar/baz.png` as well?

johnd0e:
> `Icon.Default.imagePath` only mentions the blue icon

This is how we can override image path manually, which may be needed if autodetection fails (for whatever reason).
But we still prefer to make autodetection to cover most cases without touching `.imagePath` manually.

> You're suggesting to augment this in the following way

Perhaps. I am not sure about exact code (and unable to check right now), that is why you better to check yourself.

simon04:
Before this change, on the markers test http://localhost:2015/debug/map/markers.html,
- `Icon.Default` loads http://localhost:2015/dist/images/marker-icon.png
- `L.icon({iconUrl: 'custom-marker.png'})` loads http://localhost:2015/debug/map/custom-marker.png

This PR does not affect the loaded URLs. This makes sense, since `imagePath` is only used inside `Icon.Default`.

simon04:
Let's celebrate the 1 year anniversary of this PR. :partying_face: 
Any update?

specious:
I just applied this PR as a patch to leaflet 1.7.1 installed with npm, and it works perfectly with bundling using browserify with the [browserify-css](https://github.com/cheton/browserify-css) transform if I don't inline the images.

```console
$ cd node_modules/leaflet

$ wget https://patch-diff.githubusercontent.com/raw/Leaflet/Leaflet/pull/6951.patch

$ patch -p1 < 6951.patch
patching file dist/leaflet.css
patching file src/layer/marker/Icon.Default.js
```

Here you see that my build script picks up all the files referenced by `url()` in the dependencies' CSS:

```console
$ rm -rf build/ && gulp build
[03:30:47] Using gulpfile ~/code/myapp/gulpfile.js
[03:30:47] Starting 'build'...
[03:30:47] Starting 'files'...
[03:30:47] Starting 'html'...
[03:30:47] Finished 'files' after 100 ms
Bundling with browserify...
Including external file referenced by a dependency: leaflet/dist/images/layers.png
Including external file referenced by a dependency: leaflet/dist/images/layers-2x.png
Including external file referenced by a dependency: leaflet/dist/images/marker-icon.png
Including external file referenced by a dependency: leaflet/dist/images/marker-icon-2x.png
Including external file referenced by a dependency: leaflet/dist/images/marker-shadow.png
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.eot
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.eot
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.woff2
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.woff
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.ttf
Including external file referenced by a dependency: font-awesome/fonts/fontawesome-webfont.svg
Minifying bundle...
[03:30:50] Finished 'html' after 2.94 s
[03:30:50] Finished 'build' after 2.95 s
Done in 3.53s.
```

Everything works as it should.

If I turn on image inlining, then it's the same as before. Leaflet improperly adds `")marker-icon-2x.png` to the marker icon data URI which makes an invalid URI:

```
data:image/png;base64,iVBORw0KGgoAAAAN...UVORK5CYII=")marker-icon-2x.png
```

specious:
My current solution is just to apply the patch post-install by having this in `package.json`:

```json
{
  "scripts": {
    "postinstall": "patch -d node_modules/leaflet/ -p1 --forward -r - < patches/leaflet-pr-6951.patch"
  },
  "dependencies": {
    "leaflet": "^1.7.1"
  }
}
```

specious:
I've just tested this PR applied as a patch to Leaflet 1.7.1 with esbuild, and it doesn't work. The source file looks like:

```css
.leaflet-default-icon-path {                                                                                                            
  background-image: url(./vendor/marker-icon-2V3QKKVC.png);                                                                             
}                                                                                                                                       
.leaflet-default-icon-path-2x {                                                                                                         
  background-image: url(./vendor/marker-icon-2x-SP63CLEO.png);                                                                          
}                                                                                                                                       
.leaflet-default-icon-path-shadow {                                                                                                     
  background-image: url(./vendor/marker-shadow-MEXDWUU7.png);                                                                           
}
```

But the browser tries to access:

```
marker-icon-2V3QKKVC.png%22)marker-icon-2x.png
```

If I override the asset naming rule in the esbuild options to omit the hash in the file name, then it works:

```js
{
  loader: {
    '.png': 'file'
  },
  assetNames: 'vendor/[name]-[hash]'
}
```

simon04:
Here's a SSCCE based on the [esbuild](https://esbuild.github.io/) bundler: https://github.com/simon04/leaflet-esbuild/tree/leaflet-pr-6951. After running `esbuild --bundle --outdir=dist/ --loader:.png=file index.js` (via `yarn build`), the directory `dist/` contains the following PNG files –

```
dist/layers-2x-TBM42ERR.png
dist/layers-55W3Q4RM.png
dist/marker-icon-2V3QKKVC.png
dist/marker-icon-2x-SP63CLEO.png
dist/marker-shadow-MEXDWUU7.png
```

The compiled CSS file correctly references the PNG files –

```
dist/index.css:  background-image: url(./layers-55W3Q4RM.png);
dist/index.css:  background-image: url(./layers-2x-TBM42ERR.png);
dist/index.css:  background-image: url(./marker-icon-2V3QKKVC.png);
dist/index.css:  background-image: url(./marker-icon-2x-SP63CLEO.png);
dist/index.css:  background-image: url(./marker-shadow-MEXDWUU7.png);
```

mourner:
Looks like this is no longer relevant since we switched to SVG?

Issue #5774, opened 2017-09-13
map is not removed completely on map.remove() from the DOM


jpatel3:
## How to reproduce

- Leaflet version I'm using: 1.2.0
- Browser (with version) I'm using: Chrome Version 60.0.3112.113
- It works fine in Firefox and Safari (havn't tested in IE, Edge)
- OS/Platform (with version) I'm using: macOS Sierra
- add map in div element and add layer
```
this.leafletMap = new L.Map( <element> , {
            zoomControl: true, 
            dragging: this.isInDragMode, 
            touchZoom: false,
            scrollWheelZoom: false,
            doubleClickZoom: false,
            tap: false,
}
L.tileLayer( 'http://{s}.tile.osm.org/{z}/{x}/{y}.png', {
            attribution: '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors'
                } ).addTo( this.leafletMap );
```
- Remove the map on some user action
```
if (this.leafletMap ){
        this.leafletMap.eachLayer(function(layer){
            layer.remove();
        });
        this.leafletMap.remove();
        this.leafletMap = null;
    }
```

## What behaviour I'm expecting and which behaviour I'm seeing
- Post removal of the map, it removes the map from the element, however, if I do double click on the div, it throws error - `Uncaught TypeError: Cannot read property '_leaflet_pos' of undefined`
It seems like DOM element is still holding the event listeners even though the map and layers are removed. 

## Minimal example reproducing the issue

- [ ] this example is as simple as possible
- [ ] this example does not rely on any third party code

Using http://playground-leaflet.rhcloud.com/ or any other jsfiddle-like site.

perliedman:
Hi and thanks for taking the time to report this bug.

However, it appears something is missing from the steps you describe to reproduce the issue. I set up a playground example as well as I could along the steps above: http://playground-leaflet.rhcloud.com/rezop/edit?html,output

In this example, I can not see any errors after the map is removed, so apparently there's something more going on that causes the problem you're seeing. Could you please provide more detail, so we can track this down.

jpatel3:
@perliedman Thanks for quick response and putting together playground example. I am not able to reproduce on that one. Trying few more scenarios. while I am still looking into it, a couple of questions -

Do I need explicit Layer removing, if I am doing `map.remove()`? my guess is, it will take care of layer removal as well, but can you confirm. 

[...]

PR #6958, opened 2020-01-05
fix(Map): rm moveend listener from setMaxBounds

simon04:
Calling `map.setMapBounds`, adds the `"moveend"` event listener `_panInsideMaxBounds`. Previously, this did not get cleaned up via `remove`.

See https://github.com/Leaflet/Leaflet/issues/5774#issuecomment-570953719 for a detailed analysis and a SSCCE.


mourner:
Thank you!

NikoFoX:
When can we expect a new release with this fix?

PR #6960, opened 2020-01-06
Remove unneeded zoom style

zachricha:
gave a warning with firefox

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -530,10 +530,6 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-bottom: 1px solid #ddd;
 	border-top: 1px solid #ddd;
 	}
-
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Seems to be a leftover from the IE7 times, as per https://github.com/Leaflet/Leaflet/commit/59961216bb6f2a005fcf00b31639922993762569 . Also, see https://developer.mozilla.org/en-US/docs/Web/CSS/zoom and https://caniuse.com/#feat=css-zoom

IvanSanchez:
Closing in favour of https://github.com/Leaflet/Leaflet/commit/0603741cf7322599a5bd4520eaf3ef90e00e9ada - the prefixed CSS property should still work in those super-rare-these-days IE8 browsers.

PR #6993, opened 2020-02-12
Update plugins with Leaflet.Traceroute

elmatou:
Just to say, I wrote a plugin for Leaflet which adds control & behavior to trace routes on map. 
It is a basic tool useful for sea air air navigation.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -3350,6 +3350,15 @@ Allow the user to measure distances or areas.
 <table class="plugins"><tr><th>Plugin</th><th>Description</th><th>Maintainer</th></tr>
 	<tr>
 		<td>
+			<a href="https://github.com/elmatou/Leaflet-traceroute">Leaflet.Traceroute</a>
+		</td><td>
+			 Draw routes on map, measure distances, incoming and outcoming bearings. (<a href="https://elmatou.github.io/Leaflet-traceroute/dist/demo.html">Demo</a>
+		</td><td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
The plugin looks great, but since the name `traceroute` already has [a well established meaning among computer people](https://en.wikipedia.org/wiki/Traceroute), I wonder if it could be given a different name more related to aerial navigation.

elmatou:
Hi,
It is, indeed, a reference to the unix command. 
But since leaflet plugins can not be taken for a "computer network diagnostic command" I though it will be acceptable.
Do you have a proposal for a tool useful to trace sea or air route ?

IvanSanchez:
> It is, indeed, a reference to the unix command.
> But since leaflet plugins can not be taken for a "computer network diagnostic command" I though it will be acceptable.

Well, it got *me* confused at least - mainly because I've used "visual traceroute" tools in the past, like e.g. https://stefansundin.github.io/traceroute-mapper/ , which look like:
imagen

> Do you have a proposal for a tool useful to trace sea or air route ?

:thinking: That would be an "vessel route planner" maybe?

Just thinking out loud about the name. I just would like people with my mindset to not be confused by this.

elmatou:
ok, you got my with this traceroute mapping project !

I'll go for Leaflet-Course, as in https://en.wikipedia.org/wiki/Course_(navigation)
"of course" there still can be some confusion, but still. https://en.wikipedia.org/wiki/Course

PR #6998, opened 2020-02-18
Fix exception when calling LayerGroup/hasLayer() with wrong layerId

### LINKED ISSUE #6472: Uncaught TypeError: Cannot create property '_leaflet_id' on number '1'
LayerGroup has a function `hasLayer` which expects either an object or a number documented here: https://leafletjs.com/reference-1.4.0.html#layergroup-haslayer but still querying the function with a number returns an error.

**Steps to reproduce**
1. Create a LayerGroup
2. Call hasLayer with a number which is not within layer ids

**Expected behavior**
A boolean result

**Current behavior**
An error:
```javascript
Uncaught TypeError: Cannot create property '_leaflet_id' on number '1'
    at n (leaflet.js:5)
    at e.hasLayer (leaflet.js:5)
    at <anonymous>:1:5
```
**Environment**
- Leaflet version: 1.4.0
- Browser (with version): Google Chrome Version 71.0.3578.98 (Official Build) (64-bit)
- OS/Platform (with version): Windows 10 Pro 64bit 1803 build 17134.472

https://plnkr.co/edit/Qj6MILR7GplmVaQV3yGl

johnd0e:
Now `false` returned, as per docs
fix #6472

### DIFF
--- a/src/layer/LayerGroup.js
+++ b/src/layer/LayerGroup.js
@@ -73,7 +73,9 @@ export var LayerGroup = Layer.extend({
 	// @method hasLayer(id: Number): Boolean
 	// Returns `true` if the given internal ID is currently added to the group.
 	hasLayer: function (layer) {
-		return !!layer && (layer in this._layers || this.getLayerId(layer) in this._layers);
+		if (!layer) { return false; }
+		var layerId = typeof layer === 'number' ? layer : this.getLayerId(layer);
+		return layerId in this._layers;
 	},
 
 	// @method clearLayers(): this


johnd0e:
Note:
undefined/false/null are tolerated as `layer` value.

There is even special code in tests to ensure it.

In my opinion it has little sense but can lead to bugs.
Could anyone explain source reason?

IvanSanchez:
> Note:
> Currently undefined/false/null are tolerated as `layer` value.
[...]
> In my opinion it has little sense but can lead to bugs.
> Could anyone explain source reason?

That dates back to... wow, #4. My guess is that at some point, the meaning (and documentation) of `hasLayer()` changed without anybody noticing.

johnd0e:
> The travis CI tests are failing with

Fixed

Issue #1282, opened 2013-01-15
hasLayer not checking for null


paulovieira:
L.Map.hasLayer does this:

``` javascript
var id = L.Util.stamp(layer);
```

However it might happen that the variable "layer" is null. For instance, there could be a variable like "currentActiveLayer", which might be null. Doing map.hasLayer(currentActiveLayer) results in "Uncaught TypeError: Cannot read property '_leaflet_id' of null".

In a related subject, I think it would be very natural if L.LayerGroup also had a "hasLayer" method. 

Thanks.

PR #6999, opened 2020-02-18
Change undefined argument handling in Layer[Group]/hasLayer()

johnd0e:
`undefined`/`null`/`false` are not considered as special case anymore,
`layer` argument is always required, as per docs.

Previous excessive tolerance may lead to real errors get masked.

I've tracked the issue down to it's origin: #1282, but found mentioned there reasons weak.


johnd0e:
Sorry, I've messed with commit, but will fix it asap.

johnd0e:
May be better leave corresponding tests, if it's possible to express exception as expected result.

mourner (approved):

mourner:
Looks there are still some conflicts?

johnd0e:
> Looks there are still some conflicts?

Fixed!

PR #7008, opened 2020-03-02
Use passive event listeners

yneet:
To fix the error "Does not use passive listeners to improve scrolling performance" on Lighthouse audit (https://web.dev/uses-passive-event-listeners/?utm_source=lighthouse&utm_medium=devtools)

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -111,7 +111,7 @@ function addOne(obj, type, fn, context) {
 					filterClick(e, originalHandler);
 				};
 			}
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I thought this was already done in #6694 ? I'm not sure if using passive event listeners for *every* DOM event is the right approach... which specific event is triggering the audit warning?

yneet:
@IvanSanchez, #6694 not resolve the problem completely. I test with this patch and it resolve the problem.
It's on the specific event touchstart and touchand. You want to create a specific condition like wheel, mouseenter and mouseleave?

Issue #6980, opened 2020-01-23
L.Map.Tap needs to be revised


johnd0e:
https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/handler/Map.Tap.js#L18-L20

So here listed only 2 purposes:
1. Fix click delay on mobile devices
2. Simulate `contextmenu` events (reason is not described)

So let's take it in order:
1. Good point, but since 2013 there is much more proper way to achieve it: https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away
  I admit that it still can be useful in some rare circumstances, but in any case **it should not be enabled by default**.
2. I know only 1 possible reason: iOS devices do not produce `contextmenu`.
   So instead of simulating that unconditionally we should check `L.Browser.safari` (or just `'iOS` substring in `navigator.userAgent`).
  See also #6977.

And more:
3. Handler is enabled by default in Map options, but also has additional not documented condition:  https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/handler/Map.Tap.js#L133-L136
    `Browser.touch` looks reasonable (though see #6978), but what the hell is `Browser.pointer`?
    Blame leads to [commit](https://github.com/Leaflet/Leaflet/commit/bf3b660bef7c64bc85c1dd218f3e07f7133a1bff#diff-c2e3d5ef35bf1c4a7c6b549ddf734725L105-L106) which addressed some IE10 issue.
   But I still cannot get it's cumbersome purpose, so that must be either documented or revised (which is more likely, as this condition is obviously affects not only IE10).
   Also, this may prevent to achieve (1).
4. Besides of 2 documented purposes I see in code something not clear:
   <details><summary>simulation of mousedown and mousemove</summary>

   https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/handler/Map.Tap.js#L71
   https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/handler/Map.Tap.js#L112

   </details>
   But nowadays browsers by their own generate needed mouse events on response on touch.
   (one exception is iOS where we have no `contextmenu`).

   So I believe that events are simulated to match simulated `click` event.
   But wait a minute, doesn't matter the correct sequence of events?
[...]

PR #7010, opened 2020-03-03
Fix some issues related to wrong Browser.pointer value

### LINKED ISSUE #5627: Double-tap with Chrome on Android does not work

### LINKED ISSUE #6817: Contextmenu is not triggered on iOS13

### LINKED ISSUE #6896: Zoom/pan not working on Windows 10 Touch in Chrome and Edge

### LINKED ISSUE #6973: Problem with map move
Steps to reproduce
 1. Open desktop Google Chrome wiht dev tools and turn on device format
 2. Try to move map

Expected
 Map move by dragging

Actual 
 Map don't move by dragging

Environment 

leaflet 1.6.0

Bug not reproduced in 1.5.1. 
Also I found reason of this problem - pointer event turn off for any browser on webkit (line 1905 in leaflet-src.js)

### LINKED ISSUE #6977: L.Browser.pointer value is wrong
Here are both docs and code:
https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/core/Browser.js#L94-L96

It is clearly seen that `pointer` value will be `false` in webkit despite docs statement.

The issue appeared after #6855.

I understand that PR purpose, but still insist that proposed fix was not proper.
1. `webkit` is not iOS-specific property.
   `safari` is more close, but still, that was not proper place to make that fix.
2. Perhaps this place is better to check for `safari`:
   https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/handler/Map.Tap.js#L131-L136
   At least change here would be more local.
3. Enabling `tap` handler indeed fixes issue with `contextmenu` event (though not completely: #6865), but it also has unwanted sideeffects.
   Sample to reproduce issues involves `Leaflet.draw`, which is old, unsupported,and need to be fixed itself.
But if it is relevant - I can setup a fiddle.

### LINKED ISSUE #6988: Different controls size after update 1.5.1 -> 1.6.0
Hi, I have noticed that controls are different in size; also it comes from lack of `leaflet-touch` class on map, which was present in previous version - always as I assume.

This change is not documented in doc, also those controls now (map button with size 25 px? - padding is = 18px for icon?) are little bit too small IMO.

johnd0e:
- Close #6977.
- ~~Introduce `Browser.iOS` property.~~
  <details>

  ```js
  // https://stackoverflow.com/questions/57765958/how-to-detect-ipad-and-ipad-os-version-in-ios-13-and-up
  // @property iOS: Boolean; `true` for iPhone/iPad.
  export var iOS = /iPad|iPhone|iPod/.test(navigator.platform) ||
  		(navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);
  ```
  </details>

- Revert #6855, and fix #6817 in more local way.

- Fix #6988, fix #6973, fix #6896, and may be some more similar issues.

Last commits are not strictly belongs here, and it'd made sense to separate them in another PR, but they depend on previous changes, so..
- Supersedes other related PRs (close #5881, close #6815) and make some other issues obsolete (close #5627)

---

Test page: https://gist.githack.com/johnd0e/c7b44648c5b365cd9b2a9662a23180e0/raw/test-fix-pointer.html

### DIFF
--- a/src/map/handler/Map.Tap.js
+++ b/src/map/handler/Map.Tap.js
@@ -131,6 +131,6 @@ export var Tap = Handler.extend({
 // @section Handlers
 // @property tap: Handler
 // Mobile touch hacks (quick tap and touch hold) handler.
-if (Browser.touch && !Browser.pointer) {
+if (Browser.touch && (!Browser.pointer || Browser.safari)) {
 	Map.addInitHook('addHandler', 'tap', Tap);
 }
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -93,7 +93,7 @@ export var msPointer = !window.PointerEvent && window.MSPointerEvent;
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
-export var pointer = !webkit && !!(window.PointerEvent || msPointer);
+export var pointer = !!(window.PointerEvent || msPointer);
 
 // @property touch: Boolean
 // `true` for all browsers supporting [touch events](https://developer.mozilla.org/docs/Web/API/Touch_events).
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -9,6 +9,14 @@ var _touchstart = Browser.msPointer ? 'MSPointerDown' : Browser.pointer ? 'point
 var _touchend = Browser.msPointer ? 'MSPointerUp' : Browser.pointer ? 'pointerup' : 'touchend';
 var _pre = '_leaflet_';
 
+function browserFiresNativeDblClick(e) {
+	// See https://github.com/w3c/pointerevents/issues/171
+	if (Browser.edge || Browser.safari) {
+		return e.pointerType === 'mouse'; // fire native dblclick for mouse
+	}
+	return true; // other browsers fire native dblclick for any pointer type
+}
+
 // inspired by Zepto touch code by Thomas Fuchs
 export function addDoubleTapListener(obj, handler, id) {
 	var last, touch,
@@ -19,7 +27,7 @@ export function addDoubleTapListener(obj, handler, id) {
 		var count;
 
 		if (Browser.pointer) {
-			if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
+			if (browserFiresNativeDblClick(e)) { return; }
 			count = _pointersCount;
 		} else {
 			count = e.touches.length;
@@ -38,7 +46,7 @@ export function addDoubleTapListener(obj, handler, id) {
 	function onTouchEnd(e) {
 		if (doubleTap && !touch.cancelBubble) {
 			if (Browser.pointer) {
-				if ((!Browser.edge) || e.pointerType === 'mouse') { return; }
+				if (browserFiresNativeDblClick(e)) { return; }
 				// work around .type being readonly with MSPointer* events
 				var newTouch = {},
 				    prop, i;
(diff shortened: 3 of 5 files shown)


src/map/handler/Map.Tap.js
```diff
@@ -15,10 +15,10 @@ import * as Browser from '../../core/Browser';
 // @section Interaction Options
 Map.mergeOptions({
 	// @section Touch interaction options
-	// @option tap: Boolean = true
+	// @option tap: Boolean
 	// Enables mobile hacks for supporting instant taps (fixing 200ms click
 	// delay on iOS/Android) and touch holds (fired as `contextmenu` events).
-	tap: true,
+	tap: Browser.touch && !Browser.pointer && !Browser.iOS,
```
johnd0e:
```suggestion
	tap: Browser.touch && (!Browser.pointer || Browser.iOS),
```

johnd0e:
**@ reviewers**

Before I proceed to thorough testing please give me a hint that I'm on right way with this:
- [ ] 345618e is not actually important here, should I drop it?
- [ ] 035314e is controversial, and I'm not completely sure about it yet.
- [ ] In af7d241 I am relying on `Browser.iOS`, but in c33c31c for same purpose I use `Browser.safari`.
  In fact I believe that both cases should use same property, but still haven't decided which.
  So do we really need special `Browser.iOS` (from 77d557c), or better stick to less specific `Browser.safari`?

mourner:
@johnd0e hey, thanks so much for working on this! Sorry for not getting back to you earlier.

> * [ ]  [345618e](https://github.com/Leaflet/Leaflet/commit/345618ec4b1f93ca304d496456296f43da84b60f) is not actually important here, should I drop it?

Yes, it's not a hot path so we don't have to cache the user agent.

> * [ ]  [035314e](https://github.com/Leaflet/Leaflet/commit/035314ef1cb749f7a993208f2f31cbd019044752) is controversial, and I'm not completely sure about it yet.

I would prefer for tap hacks to be added only when they are necessary — there seem to be no use case for forcing them on non-supported platforms, that should be a logic we handle on our side. It's only there to disable it just in case, although I don't really recall why this was introduced as an option and if anyone actually uses it.

> * [ ]  In [af7d241](https://github.com/Leaflet/Leaflet/commit/af7d2417cec5169c7b579b90e3973c2f803c180b) I am relying on `Browser.iOS`, but in [c33c31c](https://github.com/Leaflet/Leaflet/commit/c33c31c2aeb8c67ea4f4ed156705fde9dd7dd7ee) for same purpose I use `Browser.safari`.
>   In fact I believe that both cases should use same property, but still haven't decided which.
>   So do we really need special `Browser.iOS` (from [77d557c](https://github.com/Leaflet/Leaflet/commit/77d557c840cb0ad88177b59a94b3389712b2ad89)), or better stick to less specific `Browser.safari`? (I am tending to the latter)

If the browser behavior is the same on desktop and iOS Safari, I would use the more broad check and not introduce a new one.

> * [ ]  Last commit (0a42baf) should be discussed in #7012.
>   Perhaps later I will separate it in own PR, after #7011 get clearer.

Did it get clearer? Anyway, if it seems like a reasonable fix, I would trust you on this one — you obviously spent a lot more thought and research on pointer behavior than I am.

johnd0e:
> It's only there to disable it just in case

Well, good point. But then bad name and documentation.
Anyway I'm removing that change here (leaving it on #6980).

> If the browser behavior is the same on desktop and iOS Safari, I would use the more broad check and not introduce a new one.

AFAIK, there is no such device yet (desktop Safari + touchscreen).
I believe that we should start with `Browser.safari`, and change that later (if it will be needed).

> Did it get clearer?

Well, nobody commented neither #7011 nor #7012 yet.

johnd0e:
Test page: https://gist.githack.com/johnd0e/c7b44648c5b365cd9b2a9662a23180e0/raw/test-fix-pointer.html

PR #7013, opened 2020-03-04
Remove legacy click filter, targeted android 4.x

johnd0e:
Remove code originally introduced 7 years ago (by #1227, purposed to fix android WebView bug).

Android 4.x market share is low enough and continues to decline rapidly.
https://developer.android.com/about/dashboards

P.S.
TODO:

Perhaps it's time to retire `Browser.android23` as well. At least stop using it in own code.

Occurrences of `Browser.android` and `Browser.androidStock` also should be revised.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -106,12 +106,7 @@ function addOne(obj, type, fn, context) {
 			obj.addEventListener(type === 'mouseenter' ? 'mouseover' : 'mouseout', handler, false);
 
 		} else {
-			if (type === 'click' && Browser.android) {
-				handler = function (e) {
-					filterClick(e, originalHandler);
-				};
-			}
-			obj.addEventListener(type, handler, false);
+			obj.addEventListener(type, originalHandler, false);
 		}
 
 	} else if ('attachEvent' in obj) {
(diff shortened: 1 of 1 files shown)


johnd0e:
Unfortunately I don't know.

> E.g. I see that 4.4 (KitKat) is still pretty prominent (~7%).

1. If we count among all devices (not only android) - we'd get much lower ratio
2. Specified number is outdated (May 2019). Let's wait for more fresh data.

mourner:
Ah, you're right. Given that now [all Android phones use Chrome as the browser](https://gs.statcounter.com/browser-market-share/mobile/worldwide), and the table only concerns the old Android browser, we can probably drop the hacks.

mourner:
@johnd0e BTW, I just invited you to the Leaflet organization with push access, which will hopefully make working with the repo easier for you. Any changes still need to go through the PR review process, but now you can make branches in the repo directly instead of your fork. Thank you for all your recent contributions!

johnd0e:
> Given that now [all Android phones use Chrome as the browser](https://gs.statcounter.com/browser-market-share/mobile/worldwide),

I suppose main issue is with legacy Android WebView, I'm not sure that it is updatable in Android 4.

johnd0e:
And if we are about to remove lagacy hacks:

> Perhaps it's time to retire `Browser.android23` as well. At least stop using it in own code.
> 
> Occurrences of `Browser.android` and `Browser.androidStock` also should be revised.

Should we do it now?

mourner:
@johnd0e yes, at least those we have enough confidence to remove.

johnd0e:
> Thank you for all your recent contributions!

Thank you for wonderful library!

> now you can make branches in the repo directly instead of your fork

Is it preferred way?

PR #7020, opened 2020-03-12
Make use of PointerEvent.isPrimary

johnd0e:
https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/isPrimary
https://caniuse.com/#search=isPrimary

IE 10 or later, but it is safe to use here, as for IE we anyway need return from function.


mourner:
Was merged as a part of #7010 (no sure if accidental there but looks good to me)

Issue #2078, opened 2013-10-07
Tiles disappear after zooming in on Chrome on Android


thordursigurdsson:
Noticed this after updating to Chrome 30.0.1599.82 for Android (4.1.2 and 4.2.2).
When zooming in the tiles seem to disappear, but will reappear when you move the map.
Tested on 0.6.4. The map on mapbox.com/tour (0.6.2) is also showing this behaviour.

http://jsfiddle.net/YZE6e/

PR #7022, opened 2020-03-12
Get rid of android legacy hacks

johnd0e:
Continue #7013 

- [x] Remove internal usage of `Browser.android23`/`.androidStock`.
Keep properties for now, but mark as deprecated.
- [ ] check if `Browser.android` is still needed.
<details><summary>Currently we have only 3 occurrences</summary>

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/layer/tile/TileLayer.js#L109-L112

https://github.com/Leaflet/Leaflet/blob/983092a4975e5a4ab74e8665b38e31a17f38ddb5/src/control/Control.Layers.js#L190-L195

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/layer/tile/GridLayer.js#L798-L802
</details>

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -40,7 +40,7 @@ export var android23 = userAgentContains('android 2') || userAgentContains('andr
 
 /* See https://stackoverflow.com/a/17961266 for details on detecting stock Android */
 var webkitVer = parseInt(/WebKit\/([0-9]+)|$/.exec(navigator.userAgent)[1], 10); // also matches AppleWebKit
-// @property androidStock: Boolean; `true` for the Android stock browser (i.e. not Chrome)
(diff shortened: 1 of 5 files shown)


johnd0e:
- [ ] Perhaps this statement should be removed from doc:
https://github.com/Leaflet/Leaflet/blob/e213469a296ea00e6f9e6e6beb49d081e8d1c4fe/docs/index.html#L165

Falke-Design (approved):
I think that you have checked if #137 and #2078 is no problem anymore

johnd0e:
> I think that you have checked if #137 and #2078 is no problem anymore

Sure, as nobody really cares about Android 3 nowadays.
And since 2013 Chrome advanced in dozens of versions.

mourner (approved):

PR #7024, opened 2020-03-12
fix: Unwanted change of scale when moving horizontally/vertically

GermainJonathan:
# Unwanted change of scale when moving horizontally/vertically

By moving at a given scale (10km or 5km for example) vertically or horizontally you can observe a change in the display of the scale. (especially in metric)

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
       "consistent-return": 0,
       "no-unused-expressions": [
         "error",
-        { 
+        {
           "allowShortCircuit": true
         }
       ]
--- a/src/control/Control.Scale.js
+++ b/src/control/Control.Scale.js
@@ -114,7 +114,7 @@ export var Scale = Control.extend({
 	_getRoundNum: function (num) {
 		var pow10 = Math.pow(10, (Math.floor(num) + '').length - 1),
 		    d = num / pow10;
-
+		d = Math.round((d + Number.EPSILON) * 100) / 100;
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -89,7 +89,7 @@
       "consistent-return": 0,
       "no-unused-expressions": [
         "error",
-        {
```
johnd0e, line 92: Please exclude unrelated changes

package.json
```diff
@@ -89,7 +89,7 @@
       "consistent-return": 0,
       "no-unused-expressions": [
         "error",
-        { 
+        {
```
johnd0e, line 92: Please exclude unrelated changes

src/control/Control.Scale.js
```diff
@@ -114,7 +114,7 @@ export var Scale = Control.extend({
 	_getRoundNum: function (num) {
 		var pow10 = Math.pow(10, (Math.floor(num) + '').length - 1),
 		    d = num / pow10;
-
```
johnd0e, line 117: Please avoid white-space changes without reason

src/control/Control.Scale.js
```diff
@@ -114,7 +114,7 @@ export var Scale = Control.extend({
...
+		d = Math.round((d + Number.EPSILON) * 100) / 100;
```
johnd0e, line 117: Please give us a hint what exact values you are trying to fix.

PR #7026, opened 2020-03-13
New TapHold handler instead of legacy Tap

### LINKED ISSUE #5425: ios: dragging=false but scrolling on map not possible

### LINKED ISSUE #6705: Click events don’t work in the shadow DOM on iOS

### LINKED ISSUE #6916: Keyboard issue with Chrome and Firefox in mobile devices with an input element inside L.Control (version 1.6)
Hello, I've migrated from Leaflet 1.5.1 to 1.6. I'm using a custom html form with an input element within a Leaflet map control. In the previous version (1.5.1) it was possible to type anything in the input element. However, this is not working with version 1.6 when using an Android device with Chrome or Firefox => the issue is that the virtual keyboard is not showing when clicking the input element (it looks this element does not get the focus)

**Steps to reproduce**
- Create a map control in Leaflet
- Create a div with an input element and assign it to the map control container
- Use an Android device with Chrome or Firefox and click on the input element => the virtual keyboard is not showing

**Expected behavior**

The virtual keyboard should pop up, allowing the user to introduce some text

**Current behavior**

The virtual keyboard does not pop up

**Environment**
- Leaflet version: 1.6
- Browser (with version): Chrome 78.0.3904.108
- OS/Platform (with version): Android 9

**Additional context**

Check this fiddle:
https://next.plnkr.co/edit/8M2c7fGnacCpys2j?preview

### LINKED ISSUE #6959: Setting custom long press touch duration (now fixed to 1 second)
Currently long press touch duration ( firing `contextmenu` event on touch devices) is fixed to 1 second.

It would be very useful to be able to set the duration of long press touch duration, either by some touch interaction option like `longTouchDuration` or at least by some internal value in tape handler object (like `_longTouchDuration: 1000`) that could be easily changed by `L.Map.Tap.include`.

### LINKED ISSUE #6980: L.Map.Tap needs to be revised

### LINKED ISSUE #7255: Leaflet 1.7.1 causes 2 click events to be emitted by Leaflet Controls in Safari
**Steps to reproduce**
- set up a basic map, with a control that has a button in it, which listens for click events and posts the events to the console
- click the button in the control using Safari on a Mac
- see two click events in the console, one with `isTrusted:true` and another with `isTrusted:false` attributes

(Using 'force touch' to click the button on a trackpad results in only a single click event being emitted, but it also triggers the Safari preview popup.)

**Expected behavior**

There should only be one click event emitted, with the `isTrusted:true` attribute.

Would Leaflet be able to safely filter out the click events with `isTrusted:false`?

**Current behavior**

In version 1.6 the expected behaviour is observed. In version 1.7.1 Chrome exhibits the expected behaviour, but Safari doesn't.

The differences between the two events are included below.

MouseEvent 1:
```
MouseEvent

_simulated: true
composed: false
isTrusted: false
layerX: 15
layerY: 28
offsetX: 15
offsetY: 28
pageX: 410
pageY: 46
screenX: 1390
screenY: 189
timeStamp: 499337
webkitForce: 0
x: 410
y: 38
```

MouseEvent 2:
```
MouseEvent

composed: true
isTrusted: true
layerX: 15
layerY: 20
offsetX: 15
offsetY: 20
pageX: 410
pageY: 38
screenX: 1390
screenY: 189
timeStamp: 538265
webkitForce: 1
x: 410
y: 38
```

`_simulated`, `composed`, `isTrusted` and perhaps `webkitForce` seem to be the key differences, but it also seems odd that there are different timestamps and Y dimensions between the two events.

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Safari 13.1.2 
- OS/Platform (with version): OS X 10.13.6 and 10.15.6

**Additional context**

I noticed this after OpenStreetMap updated to the latest version of Leaflet, where the issue manifests itself by preventing some of the map tools to be used as the first event toggles a tool on and the second toggles it off. Related issue posted at https://github.com/openstreetmap/openstreetmap-website/issues/2811

Please see https://codepen.io/dankarran/pen/JjXMXzd

### LINKED ISSUE #7499: `.leaflet-active` may cause more than 1 component to be outlined simultaneously
## Motivation

It seems that `.leaflet-active` may sometimes cause more than 1 component to be outlined simultaneously, this can be confusing because it is not clear which one is currently focused (and subject to input):

## Additional context

In our `<mapml-viewer>` custom component we [fixed](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/cd14e28cee43a161579cb2f7d047cee40ec26300/src/mapml.css#L351-L355) this using the following CSS:

```css
.leaflet-active:not(:focus) {
  outline: unset !important;
}
```

johnd0e:
- Fix #6980, fix #5425 (close #5670 PR as obsolete), fix #6705 (close #6788 PR as obsolete), fix #6916, fix #7499, fix #7255 (and numerous similar, see discussion).
  Also lets avoid numerous issues with Leaflet.draw.
- Use lower delay value: `600` (instead of `1000`), close #6959.

Test page: https://gist.githack.com/johnd0e/a3647bebc3a65440d2f7ea470998816b/raw/test.html

- [ ] May be change name? Like `tapContext` or `touchContext`
- [ ] Should we leave (deprecated) `tap` alias for compatibility?
- [ ] Is common map `Handler` really best way to implement this feature?
Or better do that similar to `DomEvent.DoubleTap.js` (to handle every individual target separately)?
- [ ] May be expose `tapHoldDelay` option.


johnd0e:
This is still WIP.

Currently major blocker is #7058.

IvanSanchez (approved):
Haven't tested this personally, but feels clean enough (and restricting the handler to iOS gives assurance)

johnd0e:
@IvanSanchez 
You see last commit marked as "temporary" (because of https://github.com/Leaflet/Leaflet/pull/7026#issuecomment-606575241).
Of course we can revert it later...

And we should decide what to do with old `tap`. Currently it is just marked as "Deprecated".
Do you think that it still can be useful (under some conditions, or for test purposes)?
I doubt that we really need it for compatibility.
So may be better to remove it completely?

And one more consideration: currently both `tap` and `tapHold` add map option with same name - `tapTolerance`.
I suppose that it can lead to issues with documentation.

src/map/handler/Map.TapHold.js
```diff
@@ -0,0 +1,105 @@
...
+	_isTapValid: function () {
+		return this._newPos.distanceTo(this._startPos) <= this._map.options.tapTolerance;
+	},
+
+	_simulateEvent: function (type, e) {
+		var simulatedEvent = document.createEvent('MouseEvents');
+
+		simulatedEvent._simulated = true;
+
+		simulatedEvent.initMouseEvent(
```
johnd0e, line 92: This is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent), so perhaps better switch to [constructor](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent) form.

johnd0e:
- [ ] Cleanup `_simulated` checks

sheyda6970a:
لطفا ب من در بازیابی جیمیل های نام برده و گذرواژه هام و اطلاعاتم کمک کنید

src/map/handler/Map.TapHold.js
```diff
@@ -0,0 +1,108 @@
...
+import {Point} from '../../geometry/Point';
+import * as Util from '../../core/Util';
+import * as Browser from '../../core/Browser';
+
+/*
+ * L.Map.TapHold is used to simulate `contextmenu` event on long hold,
+ * which otherwise is not fired by mobile Safari.
+ */
+
+var tapHoldDelay = 600;
```
Falke-Design, line 13: I think we should add this as a option, there are feature request for this, so why not implementing?
Can be used from #7698

spec/suites/map/handler/Map.TapHoldSpec.js
```diff
@@ -0,0 +1,209 @@
...
+
+		clock.tick(100);
+
+		expect(spy.called).to.be.ok();
+		expect(spy.calledOnce).to.be.ok();
+		expect(spy.lastCall.args[0].originalEvent._simulated).not.to.be.ok();
+
+		// Note: depending on tapHoldDelay value it's also possible that native contextmenu may come after simulated one
+		//       and the only way to handle this gracefully - increase tapHoldDelay value.
+		//       Anyway that is edge case, as tapHold is meant for browsers where native contextmenu is not fired on touch.
```
Falke-Design, line 102: Should we add this somewhere in the code or in the docs?

spec/suites/map/handler/Map.TapHoldSpec.js
```diff
@@ -0,0 +1,209 @@
...
+		clock.tick(100);
+
+		expect(spy.notCalled).to.be.ok();
+	});
+
+	it('.originalEvent has expected properties', function () {
+		L.extend(posStart, {
+			screenX: 2,
+			screenY: 2,
+		});
```
Falke-Design, line 152: Should it not be `var posStartNew = L.extend(`?

src/map/handler/Map.TapHold.js
```diff
@@ -0,0 +1,108 @@
...
+
+// @namespace Map
+// @section Interaction Options
+Map.mergeOptions({
+	// @section Touch interaction options
+	// @option tapHold: Boolean
+	// Enables simulation of `contextmenu` event, default is `true` for mobile Safari.
+	tapHold: Browser.touch && Browser.safari && Browser.mobile,
```
Falke-Design, line 21: I didn't found a usage in the code or is it that one `Map.addInitHook('addHandler', 'tapHold', TapHold);`?

Falke-Design (changes requested):
Nice work! 👍

src/map/handler/Map.TapHold.js
```diff
@@ -0,0 +1,108 @@
...
+
+		DomEvent.on(document, 'touchend touchcancel contextmenu', this._cancel, this);
+		DomEvent.on(document, 'touchmove', this._onMove, this);
+	},
+
+	_cancelClickPrevent: function cancelClickPrevent() {
+		document.removeEventListener('touchend', DomEvent.preventDefault, false);
+		document.removeEventListener('touchend', cancelClickPrevent, false);
+		document.removeEventListener('touchcancel', cancelClickPrevent, false);
+	},
```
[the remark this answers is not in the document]
johnd0e: No, with your change we loose context.
So we had 2 options: either bind context, or use named function.

src/map/handler/Map.TapHold.js
```diff
@@ -0,0 +1,108 @@
...
+			cancelable: true,
+			view: window,
+			// detail: 1,
+			screenX: e.screenX,
+			screenY: e.screenY,
+			clientX: e.clientX,
+			clientY: e.clientY,
+			// button: 2,
+			// buttons: 2
+		});
```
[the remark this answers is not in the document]
johnd0e: We are trying to simulate `contextmenu` event, with all relevant properties.
Values (commented) are those I can see in Chrome.
But in FF I see different values, so i decided to exclude them altogether.

src/map/handler/Map.TapHold.js
johnd0e: There is a very little sense because we are unable to make such option work consistently across all the systems.
The best we can do - pick a value close to observable on real systems.

spec/suites/map/handler/Map.TapHoldSpec.js
johnd0e: It is about implementation details, and not useful in docs.

spec/suites/map/handler/Map.TapHoldSpec.js
johnd0e: Why?

src/map/handler/Map.TapHold.js
johnd0e: https://github.com/Leaflet/Leaflet/blob/5516f8ae5308142dc3fc75633f15abfc6dc28b2f/src/map/Map.js#L727

spec/suites/map/handler/Map.TapHoldSpec.js
Falke-Design: My fault readed the docs wrong

src/map/handler/Map.TapHold.js
Falke-Design: Ok interessting

src/map/handler/Map.TapHold.js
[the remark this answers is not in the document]
Falke-Design: Hmm ok. Why are you not using `DomEvent` here? Then it would be possible to use the same context right?

src/map/handler/Map.TapHold.js
Falke-Design: Yes but we can still add it as undocumented option so if there is a reason why it should not `600` it is possible to change.

src/map/handler/Map.TapHold.js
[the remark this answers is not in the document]
johnd0e: Because of #7077. See 3a90acf1536176f72e2d738940284e4668718f47

src/map/handler/Map.TapHold.js
johnd0e: Sure, but why expose it prematurely? We can do it exactly when we actualy need it.

src/map/handler/Map.TapHold.js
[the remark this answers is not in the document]
johnd0e: Well, #7077 is not in case anymore, but I still prefer to keep native add/removeEventListener here to emphasize that only native touch events are relevant here.
(Will update comments)

src/map/handler/Map.TapHold.js
Falke-Design: Because currently the releases happens not often, so when there would be a change needed the users can maybe wait over a year (I don't hope that this happens). If this is adedd now, the users can fix it self if Apple is changing something.

But I also understand it why it should not changeable. So it's ok for me.

Falke-Design (approved):

src/map/handler/Map.TapHold.js
Mortvola: Out of curiosity, why was the timeout set to 1 second to start with?

Falke-Design:
@johnd0e do you want to change something? I would force this PR so that the bug is fixed in the master

johnd0e:
> do you want to change something?

Yes

Falke-Design (approved):

Falke-Design:
@mourner if you can review this PR we would solve a lot of issues and get a big step further

Falke-Design:
Fixes #7499 too

mourner (approved):
This looks great, let's land this. The only reservations I have is whether we should add `map.tap` alias so that existing apps that reference it (e.g. `map.tap.disable()`) don't break, but maybe it's not that common to invoke it.

Also, should we add `touch-action: manipulation` in CSS as mentioned [in this article](https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away) to make sure there's no delay even if there's no proper meta tag? Looks like it's now universally supported across browsers.

johnd0e:
> to make sure there's no delay

Do you still observe delay in recent browsers (without meta tag)?

In either case, we can add `touch-action: manipulation` later on request.

Issue #7012, opened 2020-03-04
Synthetic dblclick code paths


johnd0e:
https://github.com/Leaflet/Leaflet/blob/0d06c4256f4312a7122c23626c8bd36e14269d01/src/dom/DomEvent.js#L88-L92

https://github.com/Leaflet/Leaflet/blob/0d06c4256f4312a7122c23626c8bd36e14269d01/src/dom/DomEvent.js#L135-L137

This code (following it's counterparts in `DomEvent.DoubleTap.js`) intended to ensure that `dblclick` events are fired consistently in all browsers, both for any input type (mouse/touch/pen).
Original issue: https://github.com/w3c/pointerevents/issues/171

But since original issue filed there are some changes in browsers support.
1. In my [tests](https://github.com/Leaflet/Leaflet/pull/6815/files#r380854503) all available android (6+) browsers fire native `dblclick`.
2. Unable to test myself, but in #5881 @perliedman said that `ie` also supports it natively.
3. Unable to test myself, but according recent bug reports, Safari (at least on iOS/iPad) **does not** support it natively.

So I need someone to confirm (or disprove) these observations.
I suppose it's time to drop some crutches out of Leaflet's codebase.

johnd0e:
P.S.
I do not quite understand purpose of `&& addDoubleTapListener` and `&& removeDoubleTapListener` conditions: they should be always `true`, as both variables are imported from `DomEvent.DoubleTap.js`.

johnd0e:
Proposed changes are here: https://github.com/Leaflet/Leaflet/commit/77afc140c58cd068ace4d9d66b870a2967884b22

johnd0e:
Table in https://github.com/w3c/pointerevents/issues/171 needs to be updated, as browsers are evolving.

- [x] in Firefox: pointer events can be disabled with special flag.
We need to check if touch itself do not fire `dblclick` in this case.
  - in Android FF 68.5 `dblclick` is fired **even when pointer events are disabled**.
- [x] touch emulation via dev tools
  - Chrome: no `dblclick`. And [UIEvent.detail] is always `1`.
    Well, I do not feel that we need special handling for this case.
  - Firefox: `dblclick` is fired, but event sequence is strange:
    <details><summary>it's fired immediately after first click</summary>

    ![image](https://user-images.githubusercontent.com/1838643/75959874-c08a1880-5ec7-11ea-9a73-93ea962d9c2e.png)
    </details>

[...]

Issue #6777, opened 2019-08-25
dblclick event not firing in desktop Chrome when in device simulation mode


CramericaIndustries:
When you have the viewport meta tag with a value of "user-scalable=no" present in your html file:
`<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">`
then the map.on("dblclick") event stops firing as soon as as you switch Chrome dev tools to mobile mode (CTRL+SHIFT+M).
![Untitled-1](https://user-images.githubusercontent.com/5860703/63647548-2fb9cf80-c723-11e9-903a-af4408d77906.png)

Minimal example: [Plunker](https://plnkr.co/edit/StAuNUkPuQznHman9H0g?p=preview)

As soon as you remove the `user-scaleable=no` string from the viewport meta tag the dblclick event is being fired again.

**Environment**
- Leaflet version: 1.5.1+build.2e3e0ffb
- Browser (with version): Chrome Version 76.0.3809.100 (Official Build) (64-Bit)
- OS/Platform (with version): Windows 10 1809

**Example code**
```
<!DOCTYPE html>
<html>
	<head>
		<title>Quick Start - Leaflet</title>

		<meta charset="utf-8" />
		<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">

		<link rel="shortcut icon" type="image/x-icon" href="docs/images/favicon.ico" />
	    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.5.1/dist/leaflet.css" integrity="sha512-xwE/Az9zrjBIphAcBb3F6JVqxf46+CDLwfLMHloNu6KEQCAWi6HcDUbeOfBIptF7tcCzusKFjFw2yuvEpDL9wQ==" crossorigin=""/>

	    
	   <script src="https://unpkg.com/leaflet@1.5.1/dist/leaflet.js" integrity="sha512-GffPMF3RvMeYyc1LWMHtK8EbPv0iNZ8/oTtHPx9/cc2ILxQ+u905qIwdpULaqDkyBKgOaB57QTMg7ztg8Jm2Og==" crossorigin=""></script>

	</head>


	<body>
		<div id="mapid" style="width: 600px; height: 400px;"></div>
		<script>

			var mymap = L.map('mapid',{
				// doubleClickZoom: false
			}).setView([51.505, -0.09], 13);

[...]

PR #7027, opened 2020-03-15
Reimplement DoubleTap simulation in more simple and reliable way

### LINKED ISSUE #7106: Double tap breaks zoom out on iOS

**Steps to reproduce**
- Open https://leafletjs.com/
- Double tap on map
- Continuously tap on zoom out control

**Expected behavior**
Pressing on zoom out should zoom out

**Current behavior**
In the GIF below I'm double tapping on the map to zoom in and then pressing on the zoom out control multiple times. Notice how the zoom level changes unexpectedly. It starts by zooming in, then at some point switches to zooming out and again to zooming in.

ezgif-2-f76aa56882eb

**Environment**
- Leaflet version: 1.5.1, 1.6.0 and whatever is used on https://leafletjs.com/
- Browser (with version): Safari
- OS/Platform (with version): iOS 13.4.1

### LINKED ISSUE #7419: Leaflet 1.7.1 + Ionic React / Capacitor: no double click on generated iOS app
Hello

I am using Ionic React / Capacitor with Leaflet 1.7.1 to build iOS and Android apps, but the double click is not working on the generated iOS app.
The map does not zoom in and the `dblclick` events are not triggered.

Here is the Browser object in iOS:
**"ie":false,
"ielt9":false,
"edge":false,
"webkit":true,
"android":false,
"android23":false,
"androidStock":false,
"opera":false,
"chrome":false,
"gecko":false,
"safari":false,
"phantom":false,
"opera12":false,
"win":false,
"ie3d":false,
"webkit3d":true,
"gecko3d":false,
"any3d":true,
"mobile":true,
"mobileWebkit":true,
"mobileWebkit3d":true,
"msPointer":false,
"pointer":true,
"touch":true,
"mobileOpera":false,
"mobileGecko":false,
"retina":true,
"passiveEvents":true,
"canvas":true,
"svg":true,
"vml":false}**

and the navigator.userAgent:
**Mozilla/5.0 (iPhone; CPU iPhone OS 14_3 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148**

I looked at the code and the function `browserFiresNativeDblClick` (in DomEvent.js) returns true, indicating a native double click support.
But it doesn't work as expected. Do you have any idea?

Just for testing, I locally and temporary changed browserFiresNativeDblClick to always return false and it starts working successfully on the iOS device.

I also precise it is working fine in the generated Android app.

Thanks.

### LINKED ISSUE #7422: disableClickPropagation fails after double tapping map on iOS
IMG_2DE51D162B81-1

**Steps to reproduce**
- Open [https://6v1ht.csb.app](https://6v1ht.csb.app) on iOS device
- **Double tap** on upper 50% of map to zoom
- **Single tap** on overlay element  in lower 50% of map

**Expected behavior**
- **Single tap** on overlay element should not do anything because it is a `<div>` with `disableClickPropagation`

**Current behavior**
- Map zooms in

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): iOS Safari
- OS/Platform (with version): iOS 14.3

**Additional context**
- Also broken on [https://shademap.app](https://shademap.app) when you attempt to use slider at bottom of map.
- Thank you for maintaining this amazing library. I am investigating the bug and hoping to narrow down the cause. Any guidance is greatly appreciated.

https://codesandbox.io/s/epic-stonebraker-6v1ht?file=/src/index.js

johnd0e:
Take advantage of observations from #7012:

Base handler on 'click' event instead of 'touch*' / 'pointer*'.
Ensure that simulated 'dblclick's are fired only when needed.

ATM it's safe to use in any known browser, with or without touch support,
and it is equally efficient with or without pointer events.

Additionally: reduce `delay` value to 200, as it is closer to actually observed.

Fix #7106, fix #7419, fix #7422.
Also new approach can easily fix issues like #6777.

---
Test page: https://gist.githack.com/johnd0e/12cb0d6c2aead30ea7f2e17c5f9991be/raw/test-doubletap.html

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -57,6 +57,7 @@
 
 	<!-- /dom -->
 	<script src="suites/dom/DomEventSpec.js"></script>
+	<script src="suites/dom/DomEvent.DoubleTapSpec.js"></script>
 	<script src="suites/dom/DomUtilSpec.js"></script>
 
 	<!-- /layer -->
--- a/spec/suites/dom/DomEvent.DoubleTapSpec.js
+++ b/spec/suites/dom/DomEvent.DoubleTapSpec.js
@@ -89,4 +89,28 @@ describe('DomEvent.DoubleTapSpec.js', function () {
 		}
 		expect(event.isTrusted).not.to.be.ok();
 	});
+
+	it('respects disableClickPropagation', function () {
+		var spyMap = sinon.spy();
+		var map = L.map(el).setView([51.505, -0.09], 13);
+		map.on('dblclick', spyMap);
+
+		var spyCtrl = sinon.spy();
+		var ctrl = L.DomUtil.create('div');
+		L.DomEvent.disableClickPropagation(ctrl);
+		var MyControl = L.Control.extend({
+			onAdd: function () {
+				return ctrl;
+			}
+		});
+		map.addControl(new MyControl());
+		L.DomEvent.on(ctrl, 'dblclick', spyCtrl);
+
+		happen.click(ctrl, {detail: 1});
+		clock.tick(100);
+		happen.click(ctrl, {detail: 1});
+
+		expect(spyCtrl.called).to.be.ok();
+		expect(spyMap.notCalled).to.be.ok();
+	});
 });
--- a/spec/suites/dom/DomEvent.DoubleTapSpec.js
+++ b/spec/suites/dom/DomEvent.DoubleTapSpec.js
@@ -1,0 +1,92 @@
+describe('DomEvent.DoubleTapSpec.js', function () {
+	var el, clock, spy;
+
+	beforeEach(function () {
+		el = document.createElement('div');
+		document.body.appendChild(el);
+
+		clock = sinon.useFakeTimers();
+		clock.tick(1000);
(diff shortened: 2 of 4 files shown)


johnd0e:
Well, I'm not quite sure about last commit.

As we can see (https://github.com/Leaflet/Leaflet/issues/7012#issuecomment-595087386), there are 2 known cases when `dblclick` is not produced on double-tapping in modern browsers.
But it's not clear if those cases are ever valid.

What if it's more like browsers bugs, that should be fixed in their respective sources?

johnd0e:
I believe that this PR is ready to be merged.

Possible future enhancements:
- Some way to disable simulation, like [global switch](https://leafletjs.com/reference-1.7.1.html#global-switches), or `L.Browser` property (after #7335).
- Add some tests.

johnd0e:
Tests added: f9a7edceee49c96ec215c23f187bfb17d068c2e3

Falke-Design (approved):
Nice refactoring!

mourner (approved):

Issue #5898, opened 2017-11-03
Touch event detection in Internet Explorer and Edge


geoloep:
An offshoot of the discussion in #5881 .

Commit 11cbfa809576c3c6e7f3bcb8631e6cf6fc06c685 broke double tap to zoom in both Microsoft browsers. It changed a pointerType check. I do not know the history behind the check but my interpretation is that it should catch all the pointerEvents that aren't of the type 'mouse'.

After the commit the check looks like:
```javascript
e.pointerType !== 'mouse' && e.MSPOINTER_TYPE_MOUSE && e.pointerType !== e.MSPOINTER_TYPE_MOUSE
```

e.MSPOINTER_TYPE_MOUSE only exists in Internet Explorer 10: [docs](https://msdn.microsoft.com/en-us/library/jj192113(v=vs.85).aspx)

In IE11 and Edge e.MSPOINTER_TYPE_MOUSE is undefined and therefore the check never passes.

pointerType is a number in IE10 and a string in all other browsers: [docs](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType)

For the check to support both IE10 and IE1, Edge, Chrome etc I think it should be:
```javascript
(typeof(e.pointerType) === 'string' && e.pointerType !== 'mouse') || (typeof(e.pointerType) === 'number' && e.pointerType != e.MSPOINTER_TYPE_MOUSE)
```

First it checks the type of pointerType we're dealing with and then peforms the correct comparison.

It seems to work in IE10 (VM), IE11, Edge and Firefox. 👍 The problem is that this change breaks the double tap to zoom in Chrome once again... 😫

I theorize that touch pointerEvents in Chrome should not be 'preventDefaulted', but i'm sure that would open up yet another can of worms. 

I can't dig further at this moment, so i thought i'd summarize my finding up untill now in this issue.

perliedman:
Thanks for your detailed thoughts, after a quick read it makes sense. Unfortunately I'm in the same situation and can't dig into this right now, but this documentation will be handy to have when we find more time.

johnd0e:
@ghost
Could you check if the issue can fixed by #7033?

PR #7033, opened 2020-03-18
Fix invalid condition in _addPointerStart

johnd0e:
https://github.com/Leaflet/Leaflet/blob/e213469a296ea00e6f9e6e6beb49d081e8d1c4fe/src/dom/DomEvent.Pointer.js#L58

The condition is invalid as it's never met in any browser newer that IE10.
To work as intended there must be brackets:
```js
if (e.pointerType !== 'mouse' && (e.MSPOINTER_TYPE_MOUSE && e.pointerType !== e.MSPOINTER_TYPE_MOUSE)) {
```

This PR even more simplifies the condition (and another one, just for consistency).

Some considerations:
1. The bug was introduced by 19d2bd027ec1616f05759224f3c53081b6fad6af more than 3 years ago, but I see only one bug report (#5898)
2. That issue is about IE11/Edge only, and that is strange, 'cause broken condition should affect other browsers too.
So I suspect that we may not really need that `preventDefault(e)` call for modern browsers.

Anyway, current code has no sense and should be fixed.

### DIFF
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -55,15 +55,14 @@ export function removePointerListener(obj, type, id) {
 
 function _addPointerStart(obj, handler, id) {
 	var onDown = Util.bind(function (e) {
-		if (e.pointerType !== 'mouse' && e.MSPOINTER_TYPE_MOUSE && e.pointerType !== e.MSPOINTER_TYPE_MOUSE) {
+		if (e.pointerType !== (e.MSPOINTER_TYPE_MOUSE || 'mouse')) {
 			// In IE11, some touch events needs to fire for form controls, or
 			// the controls will stop working. We keep a whitelist of tag names that
 			// need these events. For other target tags, we prevent default on the event.
-			if (TAG_WHITE_LIST.indexOf(e.target.tagName) < 0) {
-				DomEvent.preventDefault(e);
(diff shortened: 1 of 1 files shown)


src/dom/DomEvent.Pointer.js
```diff
@@ -55,15 +55,14 @@ export function removePointerListener(obj, type, id) {
...
 	var onDown = Util.bind(function (e) {
-		if (e.pointerType !== 'mouse' && e.MSPOINTER_TYPE_MOUSE && e.pointerType !== e.MSPOINTER_TYPE_MOUSE) {
+		if (e.pointerType !== (e.MSPOINTER_TYPE_MOUSE || 'mouse')) {
 			// In IE11, some touch events needs to fire for form controls, or
 			// the controls will stop working. We keep a whitelist of tag names that
 			// need these events. For other target tags, we prevent default on the event.
-			if (TAG_WHITE_LIST.indexOf(e.target.tagName) < 0) {
-				DomEvent.preventDefault(e);
-			} else {
+			if (Browser.ie && TAG_WHITE_LIST.indexOf(e.target.tagName)) {
```
mourner:
```suggestion
			if (Browser.ie && TAG_WHITE_LIST.indexOf(e.target.tagName) >= 0) {
```

PR #7042, opened 2020-03-23
iOS: prevent selection of controls text on drag/hold

### LINKED ISSUE #6865: iOS 13.1 - contextmenu/longtap issue on Markers
This is linked to #6817 fixed by  #6855.

How to reproduce: http://www.mappite.org/tap/testTap.html
Leaflet Version: Leaflet master as of yesterday (with #6855 in)

**Long tap on map:** works thanks to #6855, alert appears highlighting a contextmenu/longtap has been detected
**Long tap on marker:** user gets screen to save/share Marker icon, after closing it the alert appears

Expected behavior: when long tap on Marker OS menu to save image should not appear

`var map = L.map('map', { center: [44.22253, 11.77083], zoom: 4, maxZoom: 18, layers: [ L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png') ], attributionControl: false, tap: true});
map.on('contextmenu', function(e) {alert("MAP: contextmenu/longtap detected"); });
var m = L.marker([44.22253, 11.77083]);
m.on('contextmenu', function(e) {alert("MARKER***: contextmenu/longtap detected"); } );
map.addLayer(m);`

### LINKED ISSUE #6872: Controls' texts catch selection (focus?) while dragging on Safari iOS 13.1.3

**Steps to reproduce**
- open leafletjs.com on mobile
- put your finger on the map and wait for a moment
Note: holding finger on the is *not* dragging per se, yet this also happens to me while dragging *slowly*; users may also want to pause while dragging, as they explore the map. 

**Expected behavior**
the '+' in the zoom control or text in attribution control are not selected

**Current behavior**
see screenshot
image

**Environment**
- Leaflet version: 1.5.1
- Browser (with version): Safari mobile
- OS/Platform (with version): iOS 13.1.3

**Additional context**
I got a React app, using react-leaflet.js.org, but I can also reproduce this on leafletjs.com, as mentioned above 

johnd0e:
Workaround to fix #6872, and also fix #6865.

Test page: https://output.jsbin.com/zagowek

Note: I propose to add only prefixed (-webkit) version of the property, as this issue is specific to iOS / Safari.

https://caniuse.com/#feat=user-select-none

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -24,6 +24,7 @@
 	   -moz-user-select: none;
 	        user-select: none;
 	  -webkit-user-drag: none;
+	  -webkit-touch-callout: none; /* #6865 */
 	}
 /* Prevents IE11 from highlighting tiles in blue */
 .leaflet-tile::selection {
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -166,7 +166,7 @@
 
 /* iOS-specific: prevent selection on drag/hold (#6872) */
 
-.leaflet-control a {
+.leaflet-control:first-child {
 	-webkit-user-select: none;
 	}
 
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -163,6 +163,11 @@
 	margin-right: 10px;
 	}
 
+/* iOS-specific: prevent selection on drag/hold (#6872) */
+
+.leaflet-control a {
+	-webkit-user-select: none;
+	}
 
 /* zoom and fade animations */
 


Malvoz:
The issue described in https://github.com/Leaflet/Leaflet/issues/6872#issuecomment-598861924 is not UA/OS specific though.

johnd0e:
> used Chrome on Windows

Have you tried it on provided test page?

Malvoz:
@johnd0e The issue that I described was observed at https://geogratis.gc.ca/mapml/en/osmtile/osm/, however I just now tested on the [provided test page](https://output.jsbin.com/zagowek), where the resulting selected text from double clicking the map area is the attribution, as opposed to the zoom controls:

leaflet-doubleclick

johnd0e:
> resulting selected text from double clicking [...]

You mean 'double tapping'? 

> [...] is the attribution

But this is not in case of iOS, so I suspect that it's may be different issue.

Malvoz:
> You mean 'double tapping'?

Refering to https://github.com/Leaflet/Leaflet/issues/6872#issuecomment-598861924 (again): I reported having the issue on "desktop after double-clicking".

Essentially, the original issue https://github.com/Leaflet/Leaflet/issues/6872 by @fosterer, I suppose is mainly about the "touch callout" part which is disabled using `-webkit-touch-callout: none` - and is indeed specific to Safari (there's no standardized `touch-callout` property) - however since this PR _also_ adds `-webkit-user-select: none` (**preventing text selection**, which without it would be presented to the user as the element receiving a colored background by the browser, as shown in the screenshots), it makes sense to also hide the unintended text selection in non-webkit browsers aswell. Thus I propose adding the non-vendor specific property `user-select: none` (whereas adding the `-ms-` and `-moz-` prefixed variants is probably less important judging by the market share from the Caniuse link you posted).

johnd0e:
#6872 is about dragging. So no surprise that double-clicking is different.
I suggest you to open separate issue, to avoid misunderstanding.

johnd0e:
And I am not sure that it's good idea - prevent attribution text selection.

Malvoz:
> not sure that it's good idea - prevent attribution text selection.

I agree with that, though this PR (as is) will prevent text selection of attribution links in WebKit browsers.

So perhaps `-webkit-user-select` is out of scope in solving #6872 to begin with?

johnd0e:
> Perhaps `-webkit-user-select` is out of scope in solving #6872

It's for #6865.

Malvoz:
Yes I re-read the 6872 issue description, I guess I don't understand why this shouldn't include the unprefixed variant.

johnd0e:
> I don't understand why this shouldn't include the unprefixed variant.

Are you thinking that unprefixed `user-select` will fix your Chrome issue?

Malvoz:
I'm thinking it's the standardized property. And the issue isn't specific to iOS which the [comment suggests](https://github.com/Leaflet/Leaflet/pull/7042/files#diff-e3d63482ae6ce17380258fdb6cb730deR167).

johnd0e:
> And the issue isn't specific to iOS

Well, AFAIK it is.
Your `dblclick` issue is not related, and behaves differently.

Malvoz:
AFAIK I'm using Chrome on a Windows desktop, and I'm having the same text selection issue, whether double clicking is causing the issue vs someone dragging on iOS; it's the same problem percieved, with the same solution...

johnd0e:
1. I don't think that it is same problem
   - different actions (touch-hold vs mouse `dblclick`)
   - different consequences (on iOS only 1st-child get selected)
   - on iOS it is 100% reproducible, but on Chrome I observe this in rare case.
2. For your Chrome either `-webkit-user-select` or `user-select` have the same effect anyway.

johnd0e:
I am about to discard this PR.
- #6872: It seems not good to disable user-selection unconditionally.
- #6865: There is little sense in fixing this for iOS, but leaving as is for Android (both Chrome and latest Firefox also show 'callout' menu on hold).

PR #7047, opened 2020-03-25
Implement touchpad based pan and zoom gestures

### LINKED ISSUE #7046: Add touchpad gesture support for pan and zoom
A leaflet maps currently handle touchpad-based navigation only by dragging and mouse wheel scrolling. This can get unintuitive on laptops with large touchpads where the paradigm has shifted towards gesture based workflow.

Dragging and zooming by two finger touchpad gestures ain to touchscreen devices, like phones and tablets, would be preferred.

jonnyrobbie:
Some devices, like  MacBooks allow two finger touchpad gestures simulating multidimensional pixel-precision scroll wheels.
This patch implements pixel-precision zoom and pan for such devices.
Tested on MacBook Pro with Firefox.

Closes #7046

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ Gemfile.lock
 package-lock.json
 yarn.lock
 *.log
+*.swp
+
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -30,18 +30,23 @@ Map.mergeOptions({
 
 export var ScrollWheelZoom = Handler.extend({
 	addHooks: function () {
-		DomEvent.on(this._map._container, 'mousewheel', this._onWheelScroll, this);
+		DomEvent.on(this._map._container, 'wheel', this._onWheelScroll, this);
 
 		this._delta = 0;
 	},
 
 	removeHooks: function () {
-		DomEvent.off(this._map._container, 'mousewheel', this._onWheelScroll, this);
+		DomEvent.off(this._map._container, 'wheel', this._onWheelScroll, this);
 	},
 
 	_onWheelScroll: function (e) {
-		var delta = DomEvent.getWheelDelta(e);
 
+		// detect gesture
+		if (e.deltaMode === e.DOM_DELTA_PIXEL) {
+			return;
+		}
+
+		var delta = e.deltaY;
 		var debounce = this._map.options.wheelDebounceTime;
 
 		this._delta += delta;
--- a/src/map/index.js
+++ b/src/map/index.js
@@ -9,6 +9,8 @@ import {Keyboard} from './handler/Map.Keyboard';
 Map.Keyboard = Keyboard;
 import {ScrollWheelZoom} from './handler/Map.ScrollWheelZoom';
 Map.ScrollWheelZoom = ScrollWheelZoom;
+import {Gesture} from './handler/Map.Gesture';
+Map.Gesture = Gesture;
 import {Tap} from './handler/Map.Tap';
 Map.Tap = Tap;
 import {TouchZoom} from './handler/Map.TouchZoom';


src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+import {Handler} from '../../core/Handler';
+import * as DomEvent from '../../dom/DomEvent';
+import * as Util from '../../core/Util';
+import * as DomUtil from '../../dom/DomUtil';
+/*
+ *import * as Browser from '../../core/Browser';
+ */
+
+/*
+ * L.Handler.TouchZoom is used by L.Map to add pinch zoom on supported mobile browsers.
```
johnd0e: The copy-paste should be replaced with correct description

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+	resetTimeout: 300,
+
+	// @option zoomMultiplier: Float
+	// Sets the zoom multiplier when doing the zoom gesture.
+	zoomMultiplier: -0.05,
+
+	// @option bounceAtZoomLimits: Boolean = true
+	// Set it to false if you don't want the map to zoom beyond min/max zoom
+	// and then bounce back when pinch-zooming.
+	bounceAtZoomLimits: true
```
johnd0e: This belongs to Map.TouchZoom.js

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+// @namespace Map
+// @section Interaction Options
+Map.mergeOptions({
+	// @section Touch interaction options
+	// @option gesture: Boolean|String = *
+	// Whether the map can be zoomed and panned by two finger gesture. If
+	// passed `'center'`, it will zoom to the center of the view regardless of
+	// where the touch events (fingers) were. Enabled for touch-capable web
+	// browsers except for old Androids.
+	gesture: true,
```
johnd0e: Wrong copy-paste description, so I'm not sure about default value.

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+		if (this._map.options.zoomAnimation) {
+			this._map._animateZoom(this._center, this._map._limitZoom(this._zoom), true, this._map.options.zoomSnap);
+		} else {
+			this._map._resetView(this._center, this._map._limitZoom(this._zoom));
+		}
+	}
+});
+
+// @section Handlers
+// @property touchZoom: Handler
```
johnd0e: Wrong name

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+	},
+
+	removeHooks: function () {
+		console.log('removing gesture hook');
+		DomUtil.removeClass(this._map._container, 'leaflet-touch-zoom');
+		DomEvent.off(this._map._container, 'wheel', this._onGestureStart, this);
+	},
+
+	_onGestureStart: function (e) {
+		DomEvent.preventDefault(e);
```
johnd0e: Prevent default before even check condition?

.gitignore
```diff
@@ -20,3 +20,5 @@ Gemfile.lock
 package-lock.json
 yarn.lock
 *.log
+*.swp
+
```
johnd0e: Definitely not intended change

jonnyrobbie: I will try to fix those and do another commit.

johnd0e: > However I wonder about the .gitignore chage. *.swp files are part of vim temporary swap files and might be beneficial to include those in .gitignore

Feel free to propose it in separate PR.

> I will try to fix those and do another commit.

Please note that I haven't got any hardware to test new function, and I didn't review your implementation.
May be better to include this gesture support into `Map.ScrollWheelZoom.js` or `Map.TouchZoom.js` directly (I cannot say now).

In any case I do not feel proposed handler's name good, as term "gesture" has too broad meaning.

jonnyrobbie: I though about extending ScrollWheelZoom because touchpad gestures effectively emulate a scroll wheel at the macos level, but this is not about zooming anymore but panning too and at the user level, it is a touchpad gesture.

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,105 @@
...
+	// Set it to false if you don't want the map to zoom beyond min/max zoom
+	// and then bounce back when pinch-zooming.
+	bounceAtZoomLimits: true
+});
+
+export var Gesture = Handler.extend({
+	addHooks: function () {
+		console.log('adding gesture hook');
+		DomUtil.addClass(this._map._container, 'leaflet-touch-zoom');
+		DomEvent.on(this._map._container, 'wheel', this._onGestureStart, this);
```
IvanSanchez: This risks double-handling mousewheel events. Perhaps choosing between `mousewheel` and `wheel` based on the browser capabilities would be neater?

jonnyrobbie: As far as I know, `mousewheel` event is deprecated and `wheel` event is universally supported according to MDN matrix: https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event
But tbh, I'm willing to learn. I hacked this PR together, because I grew tired of the lack of support on my work mac when browsing OSM, but I'm not a JS proffesional by trade.

jonnyrobbie:
Is there anything I should look into for this to be merged?

src/map/handler/Map.Gesture.js
```diff
@@ -0,0 +1,102 @@
+import {Map} from '../Map';
+import {Handler} from '../../core/Handler';
+import * as DomEvent from '../../dom/DomEvent';
+import * as Util from '../../core/Util';
+import * as DomUtil from '../../dom/DomUtil';
+
+/*
+ * L.Handler.Gesture is used by L.Map to add gesture based pan and zoom on toupad based devices.
```
johnd0e: As I see you've renamed handler from `gesture` to `touchpad`.
That means that you should rename source file too, and all related identifiers.

And I still don't feel that name `touchpad` is specific enough.

src/map/handler/Map.Gesture.js
johnd0e: There is smart handling for `mousewheel` is our code:
https://github.com/Leaflet/Leaflet/blob/d843c3b88486713827d7e860b58bdba75bfbd5a2/src/dom/DomEvent.js#L100-L101

`mousewheel` is deprecated long time ago so `wheel` is mostly ok.
Except that currently only `mousewheel` will be passive (but that probably should be fixed in relevant part of code).

Issue #6978, opened 2020-01-23
L.Browser.touch should be retired


johnd0e:
https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/core/Browser.js#L98-L104

At the moment [~90%](https://caniuse.com/#feat=touch) of browsers support touch events, so this property is not really helpful anymore.
Even more: it is confusing and often mistreated (see #5407, #5266, #3944 and many other).

Currently we have only 9 occurrences in code.
<details><summary>Some are definitely not important</summary>

https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/dom/Draggable.js#L24

https://github.com/Leaflet/Leaflet/blob/d1a1e97b8290f642eb677284af53c2db64199a76/src/map/Map.js#L1117-L1118
</details>

Other occurrences should be revised, and most probably we'll find that there will be no harm to consider touch as unconditionally supported.

So I propose to retire `L.Browser.touch` completely removing it from leaflet codebase.
For compatibility purposes we have to leave `L.Browser.touch` as `true` (and describe that in documentation).

johnd0e:
Just observation:
- On desktop Chrome without touchscreen `'ontouchstart' in window` is `false`
- After enabling mobile device emulation (in dev tools) it is still `false`
- But after reloading page (keeping device emulation on) - it is `true`

Interesting...

johnd0e:
Compare: 'touchevents' test is marked deprecated in [modernizr](https://www.npmjs.com/package/modernizr#deprecation).

johnd0e:
Currently in most places of code `touch` property check is used to prevent attaching excessive listeners if touch is unsupported.
May be that makes sense for legacy devices (though I doubt that it has any impact on performance with modern ones).

PR #7057, opened 2020-03-30
Control.Layers: stop relying on Browser `android` and `touch` properties

### LINKED ISSUE #6579: Layer Control - touch clicks first layer
Using a touchscreen and tapping the layers icon opens the list of layers but clicks (or deselects) the first layer checkbox. I just want to touch (click) to expand. I was trying to see how to disable the expand on mouse hover feature but cant find anything about it. The docs say for the L.control.layers option of collapsed, its default is true and will expand on mouse hover or touch.

johnd0e:
Control.Layers: stop relying on Browser `android` and `touch` properties
<details>

- `android` is useless here, as there are a lot of non-android touch devices (including desktop ones)
- `touch` is useless, as mouse can be used on touch-screen too

Now these use cases are covered:
- `mouseenter/leave` are handled on touch devices too
   <details>

  With this PR mouse attached to device with touch support behaves as it should: reacts on `mouseover/out` events.
  But that is up on browser support, e.g. Chrome on android does not produce `mousemove/over/out` events.
  (But Firefox does)
   </details>
- If Layers control created with `{collapsed:false}`, and than it collapsed explicitly
  [using .collapse() method], it's now possible to expand it back with click.
</details>

Fix #6579.
Fix unexpected layer switching when expanding control on touch
<details>

Most browsers produce compatibility `mouseenter` for touch event, but not all,
so we listen for click also.

The problem is that these compatibility events come all in single batch, so
same touch induces expand and then immediate input switching.

To avoid this we temporarily prevent click on layers section.
Note: more straight-forward way would be relying on `e.sourceCapabilities.firesTouchEvents`,
      but atm it has rather low support: https://caniuse.com/#feat=mdn-api_inputdevicecapabilities_firestouchevents
</details>

---
Related discussions: #6978, #7022.

---
Test page: https://gist.githack.com/johnd0e/a854f95019bb1a4c1ae909fbc6769d35/raw/pr-7057.html

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -187,7 +187,13 @@ export var Layers = Control.extend({
 			this._map.on('click', this.collapse, this);
 
 			DomEvent.on(container, {
-				mouseenter: this.expand,
+				mouseenter: function () {
+					DomEvent.on(section, 'click', DomEvent.preventDefault);
+					this.expand();
+					setTimeout(function () {
+						DomEvent.off(section, 'click', DomEvent.preventDefault);
+					});
+				},
 				mouseleave: this.collapse
 			}, this);
 		}
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -409,16 +409,6 @@ export var Layers = Control.extend({
 			this.expand();
 		}
 		return this;
-	},
-
-	_expand: function () {
-		// Backward compatibility, remove me in 1.1.
-		return this.expand();
-	},
-
-	_collapse: function () {
-		// Backward compatibility, remove me in 1.1.
-		return this.collapse();
 	}
 
 });
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -1,7 +1,6 @@
 
 import {Control} from './Control';
 import * as Util from '../core/Util';
-import * as Browser from '../core/Browser';
 import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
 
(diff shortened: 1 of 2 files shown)


johnd0e:
Control.LayersSpec.js: adapt tests to changes in code

Action executed on mouseover is not synchronous now,
so we need to wait until it get completed.

**Note**: In fact I am not familiar with our tests framework, an blindly applied technique seen in other test files.

johnd0e:
> Note: more straight-forward way would be relying on `e.sourceCapabilities.firesTouchEvents`,

Another alternative: use `pointerover`, where check `e.pointerType`, and if it is not from mouse, then do nothing (and call `preventDefault` in order to suppress `mouseover`).

In year or two pointer events support should be wide enough to adopt this approach.

src/control/Control.Layers.js
```diff
@@ -187,24 +188,22 @@ export var Layers = Control.extend({
...
-		if (Browser.touch) {
-			DomEvent.on(link, 'click', DomEvent.stop);
-			DomEvent.on(link, 'click', this.expand, this);
-		} else {
-			DomEvent.on(link, 'focus', this.expand, this);
-		}
+		DomEvent.on(link, 'click', function (e) { // necessary for touch
+			DomEvent.preventDefault(e); // prevent link function
+			this.expand();
+		}, this);
```
johnd0e, line 206:
```suggestion
		DomEvent.on(link, 'focus', this.expand, this); // necessary for touch
		DomEvent.on(link, 'click', DomEvent.preventDefault); // prevent link function
```

I'm not sure if we really need react to `focus` event though, but that was in original code.
Thoughts?

johnd0e:
PR improved.

mourner (approved):

PR #7059, opened 2020-03-31
DomEvent.Pointer: process touch events only

### LINKED ISSUE #7058: touch* listeners get events from mouse
`DomEvent.Pointer.js` currently has major flow: it does not distinguish between input sources.

See following sample: https://gist.githack.com/johnd0e/39784605658089e4cf1bf66530c0d1e5/raw/test-touch-pointer.html

```js
L.DomEvent.on(mymap.getContainer(), 'touchstart touchmove touchend touchcancel', logFn)
```
When using on desktop Chrome with mouse I see log like this:
```
pointerup (mouse) 
pointermove (mouse) 
pointermove (mouse) 
pointermove (mouse) 
pointermove (mouse) 
pointermove (mouse) 
pointerdown (mouse) 
```

I suppose that:
1. Mouse events should be ignored (should be easy to check if `pointerType==='touch'`).
See #7059.
2. Listener may expect originally specified event type.
   Perhaps we need something like this: https://github.com/Leaflet/Leaflet/pull/7044/files
   (Instead of additional handler argument, may be better put it into event as property, or override in some way)

---
See also #7077

johnd0e:
Fix #7058

Todo:
- [x] Should we reject `mouse` only? Or opposite: process `touch`only?
- [ ] Pass original event type (#7064)?

### DIFF
--- a/spec/suites/dom/DomEvent.PointerSpec.js
+++ b/spec/suites/dom/DomEvent.PointerSpec.js
@@ -46,6 +46,15 @@ describe('DomEvent.Pointer', function () {
 			});
 		});
 
+		it('ignores events from mouse', function () {
+			pointerEvents.forEach(function (type) {
+				happen.once(el, {type: type, pointerType: 'mouse'});
+			});
+			touchEvents.forEach(function (type) {
+				expect(listeners[type].notCalled).to.be.ok();
+			});
+		});
+
 		it('ignores native touch events', function () {
 			touchEvents.forEach(function (type) {
 				happen.once(el, {type: type});
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -17,7 +17,7 @@ var pEvent = {
 };
 var handle = {
 	touchstart  : _onPointerStart,
-	touchmove   : _onPointerMove,
+	touchmove   : _handlePointer,
 	touchend    : _handlePointer,
 	touchcancel : _handlePointer
 };
@@ -68,6 +68,8 @@ function _addPointerDocListener() {
 }
 
 function _handlePointer(handler, e) {
+	if (e.pointerType === (e.MSPOINTER_TYPE_MOUSE || 'mouse')) { return; }
+
 	e.touches = [];
 	for (var i in _pointers) {
 		e.touches.push(_pointers[i]);
@@ -81,14 +83,6 @@ function _onPointerStart(handler, e) {
 	// IE10 specific: MsTouch needs preventDefault. See #2000
 	if (e.MSPOINTER_TYPE_TOUCH && e.pointerType === e.MSPOINTER_TYPE_TOUCH) {
 		DomEvent.preventDefault(e);
-	}
-	_handlePointer(handler, e);
-}
-
-function _onPointerMove(handler, e) {
-	// don't fire touch moves when mouse isn't down
-	if ((e.pointerType === (e.MSPOINTER_TYPE_MOUSE || 'mouse')) && e.buttons === 0) {
-		return;
 	}
 	_handlePointer(handler, e);
 }


johnd0e:
For debug purposes we might want to keep previous behavior under some [global flag](https://leafletjs.com/reference-1.7.1.html#global-switches), or other `L.Browser.*` flag (depends on #7335).

Falke-Design (approved):
Looks good.
General question: I mean (but I'm not 100% sure) that it was currently possible to listen on `mousedown` event and then it was fired by `touchdown` and `mousedown` is this still possible? Or was it never possible?

johnd0e:
> I mean (but I'm not 100% sure) that it was currently possible to listen on `mousedown` event and then it was fired by `touchdown` and `mousedown` is this still possible? Or was it never possible?

Do you mean case when native `mousedown` produces simulated `touchstart`? It is not helpful at all.

Falke-Design:
> It is not helpful at all.

I see that helpful because if a we want to listen if the users holds down on a layer, we need to define `touchstart` and `mousedown`, but it would have the same effect so it can simplified to one event. For example the dragging in Geoman, we needed to define always `touch` and `mouse`  events ([see here](https://github.com/geoman-io/leaflet-geoman/blob/59e281464308ea1e2f5fad3e4355e9ea01cca13d/src/js/Mixins/Dragging.js#L58) ) but if there would be a "global" one like `down`, `move`, `up` then developers can be sure that Leaflet will handle it (touch & mouse)

johnd0e:
> > It is not helpful at all.
> 
> I see that helpful because if a we want to listen if the users holds down on a layer, we need to define `touchstart` and `mousedown`, but it would have the same effect so it can simplified to one event.

It is a matter of adding one word to `.on('...`

And imagine all other cases, when you need react to touch but not to mouse.

So better be a bit more expressive here, and specify what you need explicitly.

IvanSanchez:
> but if there would be a "global" one like down, move, up then developers can be sure that Leaflet will handle it (touch & mouse)

That would be `pointerdown` and `pointerup`.

Falke-Design:
> It is a matter of adding one word to .on('...

Yes sure but I'm thinking of that usecase where developers forget to add the touch event and then it is not working on touch devices.

> And imagine all other cases, when you need react to touch but not to mouse.

So a new event like `pointerdown` would be useful, then all cases are available. Only touch, only mouse and listen on both

johnd0e:
> So a new event like `pointerdown` would be useful, then all cases are available. Only touch, only mouse and listen on both

Right, but to make this consistently we need a huge refactoring (@IvanSanchez has some proof-of-concept).
And it will be not compatible with current API.

Falke-Design:
Okay thanks

IvanSanchez:
> > So a new event like `pointerdown` would be useful, then all cases are available. Only touch, only mouse and listen on both
> 
> Right, but to make this consistently we need a huge refactoring (@IvanSanchez has some proof-of-concept). And it will be not compatible with current API.

That'd be https://github.com/Leaflet/Leaflet/commit/3151aadb17a64131ecaa533a1f328c6a41339ff1 . That set of changes **is** meant to be backwards-compatible.

Falke-Design:
With a PR like that we will lose the difference between mouse and touch, or have I overseen something? 

Would it not possible to create a internal listener on `touchstart` and `mousedown` and then fire `pointerdown`?

johnd0e:
> With a PR like that we will lose the difference between mouse and touch

Where??
It's opposite - we get full control choosing what event do we want to process.
(Previously our `touch*` handlers reacted on mouse events, which is really unexpected).
And look - all the tests are passed.

> Would it not possible to create a internal listener on `touchstart` and `mousedown` and then fire `pointerdown`?

Out of scope of this small PR.

IvanSanchez:
> With a PR like that we will lose the difference between mouse and touch, or have I overseen something?

I guess you should read https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType

Falke-Design:
> I guess you should read

Yep, I didn't saw this. Sorry I thought PointerEvent is create by Leaflet.

> Out of scope of this small PR.

Sure was a general question. For me is now all clear, thank you.

Falke-Design (approved):
Approve it again, to show it below the off-topic messages

mourner (approved):

PR #7060, opened 2020-03-31
DomEvent: use passive for touchmove (instead of touchend)

### LINKED ISSUE #7052: Shouldn't 'touchmove' be passive?
https://github.com/Leaflet/Leaflet/blob/d843c3b88486713827d7e860b58bdba75bfbd5a2/src/dom/DomEvent.js#L103-L104

See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners

johnd0e:
Close #7052.

1. Original code (added in #7008) does not follow docs: https://developers.google.com/web/tools/lighthouse/audits/passive-event-listeners

2. We should not make `touchend` passive, as it does not help with performance: `preventDefault` is used there does not affect touch - it only useful to [prevent simulation of `click`](https://w3c.github.io/touch-events/#event-touchend) (which is also valuable feature).

---
Also:
- simplify some conditions and add passive option for `wheel` event listeners
  (previously it was supported for `mousewheel` only)
- prefer `wheel` over `mousewheel`

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -100,10 +100,10 @@ function addOne(obj, type, fn, context) {
 		if (type === 'mousewheel') {
 			obj.addEventListener('onwheel' in obj ? 'wheel' : 'mousewheel', handler, Browser.passiveEvents ? {passive: false} : false);
 
-		} else if ((type === 'touchstart') || (type === 'touchmove')) {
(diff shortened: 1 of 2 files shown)


johnd0e:
I've made some test with Lighthouse and found that it is happy now.
Sample page: https://jsbin.com/lacekosoje/1/edit?html,console

PR #7064, opened 2020-04-02
DomEvent: keep original value of event type for transformed events

johnd0e:
Currently for compatibility purposes Leaflet substitutes 'touch*' listeners with 'pointer*' ones (~~where available~~ if touch is not available).
And similarly with 'mouseenter' / 'mouseleave'.

It may lead to unexpected consequences, if handler relies on original value of `type` in own code.
That's why it'd better to keep original `type`.

In proposed modification I pass 'original' type as 2nd argument of handler, and this is enough for [events](https://leafletjs.com/reference-1.6.0.html#map-click) dispatched with `map._fireDOMEvent`: handler then has access both to 'original' event type, and to real type (`originalEvent.type`).

But for listeners attached directly with `DomEvent.on(...` it is not so simple, as `type` property is read-only (as a rule). I believe that we should keep things simple: `type` as second argument is smart enough.
Perhaps, that should be documented in https://leafletjs.com/reference-1.6.0.html#domevent.

### DIFF
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -15,6 +15,12 @@ var pEvent = {
 	touchend    : POINTER_UP,
 	touchcancel : POINTER_CANCEL
 };
+var tEvent = {
+	POINTER_DOWN  : 'touchstart',
+	POINTER_MOVE  : 'touchmove',
+	POINTER_UP    : 'touchend',
+	POINTER_CANCEL:	'touchcancel'
+};
 var handle = {
 	touchstart  : _onPointerStart,
 	touchmove   : _handlePointer,
@@ -76,7 +82,7 @@ function _handlePointer(handler, e) {
 	}
 	e.changedTouches = [e];
 
-	handler(e);
+	handler(e, tEvent[e.type]);
 }
 
 function _onPointerStart(handler, e) {
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -81,8 +81,8 @@ function addOne(obj, type, fn, context) {
 
 	if (obj[eventsKey] && obj[eventsKey][id]) { return this; }
 
-	var handler = function (e) {
-		return fn.call(context || obj, e || window.event);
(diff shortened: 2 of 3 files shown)


johnd0e:
> In proposed modification I pass 'original' type as 2nd argument of handler[...]

Alternatively original value could be passed as event's property `_originalType`.

johnd0e:
Todo: make use in `Draggable.js`

Falke-Design (changes requested):
Passing the new type to `_handleDOMEvent` is missing in Canvas:
https://github.com/Leaflet/Leaflet/blob/e3719a317ab325ed091fbbc6bfd34640fe19e13e/src/layer/vector/Canvas.js#L72-L73

For me is the name `originalType` not fitting, because it let me think that is the type of the `pointer` event and not the new `fake` type

johnd0e:
> Passing the new type to `_handleDOMEvent` is missing in Canvas:

Why do we need it there?

Falke-Design:
This was very bad described from me sorry ...

You have changed the function arguments of ` _handleDOMEvent` so that the `originalType` is used for `this._fireDOMEvent(e, type);`:
```
_handleDOMEvent: function (e, originalType) {
	// ...
	var type = originalType || e.type;
	// ...
	this._fireDOMEvent(e, type);
```

If Canvas is enabled `_fireDOMEvent` will be called too and then it should also use the `originalType`. Else `pointerevents` will **not** converted to `touchevents`. If SVG is used, it will be converted -> should be the same behaviour for both

------------

Besides from above, the `_handleDOMEvent` is not listening on `touchevents`, so there is currently no need to pass the originalType there, because it will be never fired by a `pointerevent`.
`_handleDOMEvent` is listening only on the following events:
https://github.com/Leaflet/Leaflet/blob/e3719a317ab325ed091fbbc6bfd34640fe19e13e/src/map/Map.js#L1314-L1315

**But** I currently working on the PR to adding touchevents and there I will need the `originalType` in the `_handleDOMEvent`, but a future PR should not be a reason to keep it in this PR

PR #7066, opened 2020-04-02
Draggable: reset internal property _moved on drag end (instead of start)

### LINKED ISSUE #7007: The first click will not trigger the mouse down event in chrome
I make a little modified on the example: [crs-simple](https://leafletjs.com/examples/crs-simple/crs-simple-example1.html). 

I add a rectangle on a map. And there is a mousedown event listening on the rectangle. I add the options to map: renderer: L.canvas() and preferCanvas: true.

After I drag and drop the map, when I first click the Rectangle nothing happens. When I click the rectangle second time, the mousedown event is fired. But the first click should fire mousedown event also.

After I comment out the following lines, the program works.
```
renderer: L.canvas()
preferCanvas: true
```

Chrome  73.0.3683.86 x64
Leaflet: 1.6

Here is the code: 

```
<script>
    var map = L.map('map', {
        crs: L.CRS.Simple,
        renderer: L.canvas(),                                                      
        preferCanvas: true,  
    });

    var bounds = [[0,0], [1000,1000]];
    var image = L.imageOverlay('https://leafletjs.com/examples/crs-simple/uqm_map_full.png', bounds).addTo(map);

    map.fitBounds(bounds);

    // define rectangle geographical bounds
    var bounds = [[500, 500], [750, 750]];
    // create an orange rectangle
    var device = L.rectangle(bounds, {color: "#ff7800", weight: 1, fillOpacity:1}).addTo(map);

    function onDeviceMouseDown(e) {
        alert("You clicked the map at " + e.latlng);
    }

    device.on('mousedown', onDeviceMouseDown);
</script>

```

johnd0e:
Prevents race conditions when canvas processes events earlier than map.

Fix #7007.

Note: Tests seem passed, but I have not reviewed all possible code paths of Draggable.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -92,8 +92,6 @@ export var Draggable = Evented.extend({
 		// under some circumstances, see #3666.
 		if (e._simulated || !this._enabled) { return; }
 
-		this._moved = false;
-
 		if (DomUtil.hasClass(this._element, 'leaflet-zoom-anim')) { return; }
 
 		if (Draggable._dragging || e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
@@ -230,6 +228,7 @@ export var Draggable = Evented.extend({
 		}
 
 		this._moving = false;
+		this._moved = false;
 		Draggable._dragging = false;
 	}
 


johnd0e:
> Note: Tests seem passed

But it means only that we have no test for this case.

Unfortunately this change brakes proper click detection:
```js
	var mymap = L.map('mapid', {tap:false}).setView([51.505, -0.09], 13);
	var marker = L.marker([51.5, -0.09],{draggable:true}).addTo(mymap);
	mymap.on('click', console.log);
```
After marker drag `click` MUST NOT occur .

PR #7079, opened 2020-04-10
Improve popup and attribution accessibility

nkreer:
This PR introduces minor changes to the way popups' close buttons and the map attribution look like to screen reader users. See the commit messages for further details. This greatly affects screen reader users when they come across Leaflet maps as their primary items are now less confusingly annotated - e.g. "times" is now "close" as it should be.


nkreer:
As Travis CI's log is itself not accessible, I cannot get any information from it to fix the failing tests myself. Running `npm test` locally results in an `ELIFECYCLE` error and nothing really descriptive about them is returned. I would appreciate being guided towards fixing these myself.

johnd0e:
It is not your fault, see #7080.

skorasaurus:
ooh, thank you for the Pull requests! I was about to make a similar issue on this! 

I'll test this on NVDA later today to ensure that works as expected.

skorasaurus:
Thank you for your work. 

I applied PR 7079 commits as of https://github.com/Leaflet/Leaflet/pull/7079/commits/45f04d3bc7a357ea6fe8b90e90ac09364a37d92c to leaflet 1.6.0; and compared it to leaflet 1.6.0 that didn't have your commits. 

Tested on NVDA, 2019.10. edition, on windows 10, with firefox and chrome; the behaviors are applicable for both firefox and chrome browsers unless specified. 

The test page can be found at http://skorasaurus.github.io/maptemplates/nkreer/nkreer.html

what's announced when user focuses on marker using the keyboard: 
on leaflet 1.6.0: nothing is announced so user doesn't know that a marker is there! 
while on leaflet with PR 7079: "marker graphic" (on chrome); "marker graphic clickable" (on firefox);
A remarkable improvement for the default behavior! 

Scenario: when focus is on the close button of an open pop-up, what is announced by screen-reader:
on leaflet 1.6.0: "link times"
on leaflet with PR 7079: "close button link"

This is also definitely an improvement! 

I am less familiar with the attribution issue but will try to better understand it to help.

skorasaurus:
@nkreer: 

In my setup of windows 10, leaflet 1.6.0 without your pull request, using the default settings of NVDA 2019.10 and chrome and edge; the `| ` symbol is already ignored by NVDA when I tab between links and move between words that are separated by | in the attribution. What unexpected behavior did you find with `|` ? 
My test case of leaflet 1.6.0 without your changes applied is available at https://deadpan-misty-agreement.glitch.me/


Regarding the fixing of the tests, there's two issues that I notice: 

One is that there are several failures that all reside in spec/suites/control/Control.AttributionSpec.js of leaflet, 
```
Firefox 75.0.0 (Ubuntu 0.0.0) Control.Attribution #addAttribution adds one attribution correctly FAILED
	expected 'prefix <span aria-hidden="true">|</span> foo' to sort of equal 'prefix | foo'
	Assertion.prototype.assert@node_modules/expect.js/index.js:96:13
	Assertion.prototype.eql@node_modules/expect.js/index.js:230:10
	@spec/suites/control/Control.AttributionSpec.js:28:35
	
Firefox 75.0.0 (Ubuntu 0.0.0) Control.Attribution #addAttribution adds no duplicate attributions FAILED
	expected 'prefix <span aria-hidden="true">|</span> foo' to sort of equal 'prefix | foo'
	Assertion.prototype.assert@node_modules/expect.js/index.js:96:13
	Assertion.prototype.eql@node_modules/expect.js/index.js:230:10
	@spec/suites/control/Control.AttributionSpec.js:34:35

```

The other failure is the following and is much easier to fix in spec/suites/layer/marker/MarkerSpec

It's setting the default Marker alt value from empty to be marker. 

```
Firefox 75.0.0 (Ubuntu 0.0.0) Marker #setIcon sets the alt attribute to an empty string when no alt text is passed FAILED
	expected 'Marker' to equal ''
	Assertion.prototype.assert@node_modules/expect.js/index.js:96:13
	Assertion.prototype.equal@node_modules/expect.js/index.js:216:10
	Assertion/this[name]@node_modules/expect.js/index.js:69:24
	@spec/suites/layer/marker/MarkerSpec.js:199:24
```

nkreer:
I hid the `|` separators because, especially mobile screenreaders, do not ignore them by default. I decided to make the close-button a `<span>`-element to not mess with styling as much. Setting the appropriate role is sufficient with modern browsers.

src/layer/Popup.js
```diff
@@ -191,7 +191,8 @@ export var Popup = DivOverlay.extend({
 		if (this.options.closeButton) {
 			var closeButton = this._closeButton = DomUtil.create('a', prefix + '-close-button', container);
 			closeButton.href = '#close';
-			closeButton.innerHTML = '&#215;';
+			// This is actually a button labelled "Close" for accessibility
+			closeButton.innerHTML = '<span role="button" aria-label="Close">&#215;</span>';
```
Malvoz, line 195: The [HTML validator](https://validator.w3.org/) reports:

> **Error: An element with the attribute `role=button` must not appear as a descendant of the `a` element.**

I suggest setting `role="button"` on the `<a>` element instead.

Malvoz: The above suggestion would also be consistent with how it's done for zoom controls:

https://github.com/Leaflet/Leaflet/blob/71f73785c91f50821cee4f590c41b14eeddf394c/src/control/Control.Zoom.js#L91-L92

Malvoz: Summary:
```diff
 closeButton.href = '#close';
+closeButton.setAttribute('role', 'button');
+closeButton.setAttribute('aria-label', 'Close popup');
-closeButton.innerHTML = '&#215;';
+closeButton.innerHTML = '<span aria-hidden="true">&#215;</span>';
```


<details><summary>Before & After comparison of computed ARIA properties</summary>
popup
</details>

@nkreer want to move forward with this? 🙏

src/layer/marker/Marker.js
```diff
@@ -42,7 +42,7 @@ export var Marker = Layer.extend({
 
 		// @option alt: String = ''
 		// Text for the `alt` attribute of the icon image (useful for accessibility).
-		alt: '',
+		alt: 'Marker',
```
Malvoz, line 45: Would it feel more "natural" to users if this said "Object" instead?

src/layer/Popup.js
Malvoz, line 195: I think changing from `aria-label="Close"` to `aria-label="Close popup"` is an improvement, see https://github.com/Leaflet/Leaflet/issues/7399.

nkreer: agreed, but that would only really make sense if you can make sure that there is only one popup visible at a time - otherwise including a title of some sort with the label would be necessary to avoid confusion I think.

mourner (approved):

Malvoz (approved):

mourner (changes requested):
Sorry for getting back to you so late! I've synced this PR with master after merging #7794, now it only needs a few test updates so that the CI build is green.

src/control/Control.Attribution.js
```diff
@@ -119,7 +119,7 @@ export var Attribution = Control.extend({
 			prefixAndAttribs.push(attribs.join(', '));
 		}
 
-		this._container.innerHTML = prefixAndAttribs.join(' | ');
+		this._container.innerHTML = prefixAndAttribs.join(' <span aria-hidden="true">|</span> ');
```
Malvoz, line 122: Attaching a video with audio capture (you may have to unmute the video) showing the problem on Android TalkBack:

https://user-images.githubusercontent.com/26493779/149366955-0d5acf0a-f7f8-4740-91db-b5739d45186c.mp4

`aria-hidden="true"` should prevent TalkBack from "focusing" and announcing the vertical line separator.

Falke-Design:
@nkreer please update the tests for `Control.Attribution`:

Simply replace in each test `|` with `<span aria-hidden="true">|</span>`

https://github.com/Leaflet/Leaflet/blob/2e205bf5ac80603936c6bc5e414aae3a56e6ce40/spec/suites/control/Control.AttributionSpec.js#L27

@mourner maybe you want to do this, because then this PR would be a valid addition and can be merged

Malvoz:
This PR was partly fixed in https://github.com/Leaflet/Leaflet/pull/7895 (for the default "marker" `alt`) and the other part is https://github.com/Leaflet/Leaflet/pull/7969 to hide the vertical line separator in the attribution from screen readers. Closing as superseded.

PR #6804, opened 2019-09-14
Better Icon.Default imagePath detection using regex

oleg-andreyev:
Regex covers:
- url(http://localhost:8000/base/dist/images/marker-icon.png)
- url(http://localhost:8000/base/dist/images/marker-icon-2x.png)
- url(http://localhost:8000/base/dist/images/marker-icon.hash.png)
- url(http://localhost:8000/base/dist/images/marker-icon-2x.hash.png)
- url("http://localhost:8000/base/dist/images/marker-icon.png")
- url('http://localhost:8000/base/dist/images/marker-icon.png')

PR #7092, opened 2020-04-14
Better icon image path detection

### LINKED ISSUE #6496: data-urls in L.Icon.Default

johnd0e:
Icon.Default/imagePath: more strict detect rules

Reject paths not ending on 'marker-icon.png'.
This should solve issues with images bundled into data-uri (webpack etc).

Add tests.

Close #6804 (superseded), fix #6496, fix (presumably) #7424.

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -31,7 +31,7 @@ export var IconDefault = Icon.extend({
 	},
 
 	_getIconUrl: function (name) {
-		if (!IconDefault.imagePath) {	// Deprecated, backwards-compatibility only
+		if (typeof IconDefault.imagePath !== 'string') {	// Deprecated, backwards-compatibility only
 			IconDefault.imagePath = this._detectIconPath();
 		}
 
@@ -42,19 +42,21 @@ export var IconDefault = Icon.extend({
 		return (this.options.imagePath || IconDefault.imagePath) + Icon.prototype._getIconUrl.call(this, name);
 	},
 
+	_stripUrl: function (path) {	// separate function to use in tests
+		var strip = function (str, re, idx) {
+			var match = re.exec(str);
+			return match && match[idx];
+		};
+		path = strip(path, /^url\((['"])?(.+)\1\)$/, 2);
+		return path && strip(path, /^(.*)marker-icon\.png$/, 1);
+	},
+
 	_detectIconPath: function () {
 		var el = DomUtil.create('div',  'leaflet-default-icon-path', document.body);
 		var path = DomUtil.getStyle(el, 'background-image') ||
 		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
 
 		document.body.removeChild(el);
-
-		if (path === null || path.indexOf('url') !== 0) {
-			path = '';
-		} else {
-			path = path.replace(/^url\(["']?/, '').replace(/marker-icon\.png["']?\)$/, '');
-		}
-
-		return path;
+		return this._stripUrl(path) || '';
 	}
 });
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -57,6 +57,10 @@ export var IconDefault = Icon.extend({
 		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
 
 		document.body.removeChild(el);
-		return this._stripUrl(path) || '';
+		path = this._stripUrl(path);
+		if (path) { return path; }
+		var link = document.querySelector('link[href$="leaflet.css"]');
+		if (!link) { return ''; }
+		return link.href.substring(0, link.href.length - 'leaflet.css'.length - 1);
 	}
 });
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -396,7 +396,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 
 /* Default icon URLs */
-.leaflet-default-icon-path {
+.leaflet-default-icon-path { /* used only in path-guessing heuristic, see L.Icon.Default */
 	background-image: url(images/marker-icon.png);
 	}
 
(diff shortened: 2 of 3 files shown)


src/layer/marker/Icon.Default.js
```diff
@@ -57,6 +57,10 @@ export var IconDefault = Icon.extend({
 		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
 
 		document.body.removeChild(el);
-		return this._stripUrl(path) || '';
+		path = this._stripUrl(path);
+		if (path) { return path; }
+		var link = document.querySelector('link[href$="leaflet.css"]');
```
IvanSanchez, line 62: I guess this should be hoisted up, so if this runs successfully the fallback method doesn't run?

johnd0e: I'm not sure: css-method can be used not only to detect path, it also allows to *override* path.

simon04:
Various build systems (often by default)
- include asset files (only) when they are referenced in JS/CSS
- rename the asset files and include a hash in the filename. This facilitates caching (via `Cache-Control:immutable`)

For instance, bundling with [esbuild](https://github.com/evanw/esbuild) yields:
```bash
$ npm install leaflet esbuild
$ echo 'require("leaflet/dist/leaflet.css")' | node_modules/.bin/esbuild --bundle --outdir=dist/ --loader:.png=file
$ ls -1 dist/
layers-2x.CUVBMIZT.png
layers.ZHTVFDSJ.png
marker-icon.MCUQXS5S.png
stdin.css
stdin.js
```

`images/marker-icon.png` is renamed to `marker-icon.MCUQXS5S.png`, `images/marker-shadow.png` is missing.

The bundler webpack also includes a files hash by default, see [file-loader#options](https://webpack.js.org/loaders/file-loader/#options).

Does this PR take care of this de-facto standard behaviour of bundlers?

johnd0e:
> Does this PR take care of this de-facto standard behaviour of bundlers?

It's impossible to guess random filenames, but you should be able to disable renaming in your bundler's config.
It would be wonderful if someone write small tutorial for such use case.

simon04:
There's no need to guess the filenames. They are correctly referenced in the source files –

```css
$ grep -C2 png dist/stdin.css

.leaflet-control-layers-toggle {
  background-image: url(layers.ZHTVFDSJ.png);
  width: 36px;
  height: 36px;
}
.leaflet-retina .leaflet-control-layers-toggle {
  background-image: url(layers-2x.CUVBMIZT.png);
  background-size: 26px 26px;
}
.leaflet-default-icon-path {
  background-image: url(marker-icon.MCUQXS5S.png);
}
```

johnd0e:
> They are correctly referenced in the source files –

That may be enough for layers, but javascript parts of ode still relies on original filenames - e.g. L.Marker, L.ImageOverlay.

simon04:
Thus, the path-guessing heuristic should deal with `marker-icon-2x.png` and `marker-shadow.png` separately:
- `div.leaflet-default-icon-path` → `background-image` url → `L.Icon.Default.iconUrl`
- `div.leaflet-default-icon-path.leaflet-retina` → `background-image` url → `L.Icon.Default.iconRetinaUrl`
- `div.leaflet-default-icon-path.leaflet-marker-shadow` → `background-image` url → `L.Icon.Default.shadowUrl`

Related: https://github.com/Leaflet/Leaflet/pull/6951

simon04:
@johnd0e, I'm happy to enhance my PR #6951 as suggested in https://github.com/Leaflet/Leaflet/pull/7092#issuecomment-724591326. What do you think, does this change have any chance to get merged?

johnd0e:
> What do you think, does this change have any chance to get merged?

Sure, there are chances (but no promises).

DimitrisRK:
Any news regarding this PR? I face trouble using default marker icons because of this issue and I prefer keeping webpack hash naming convention.

src/layer/marker/Icon.Default.js
```diff
@@ -42,19 +42,25 @@ export var IconDefault = Icon.extend({
 		return (this.options.imagePath || IconDefault.imagePath) + Icon.prototype._getIconUrl.call(this, name);
 	},
 
+	_stripUrl: function (path) {	// separate function to use in tests
```
specious, line 45: Might be more clear to say `// standalone function so it can be tested independently`

specious:
It looks like a combination of #6951 and this PR might do the trick, however this is still a hack.

What exactly is the reason Leaflet needs to perform path manipulation to switch the marker image?

mourner (approved):

mourner:
@Falke-Design nice work tracking all the relevant issues! Can you close those that no longer apply after merging this?

simon04:
Still does not work 😞 

Here's a SSCCE based on the [esbuild](https://esbuild.github.io/) bundler: https://github.com/simon04/leaflet-esbuild/tree/leaflet-bug-7092 using the latest https://github.com/Leaflet/Leaflet/commit/80b268525e842fe087f0d058fd16efa75423557b. After running `esbuild --bundle --outdir=dist/ --loader:.png=file index.js` (via `yarn build`), the directory `dist/` contains the following PNG files, but Leaflet tries to load the non-existing `marker-icon-2x.png` –

```
dist/layers-55W3Q4RM.png
dist/layers-2x-TBM42ERR.png
dist/marker-icon-2V3QKKVC.png
```

Falke-Design:
Like I understand this PR, it was not made to work with images names they contain hashs, it still needs `marker-icon.png`

PR #7125, opened 2020-05-05
Restore DomEvent.off(el) functionality when calling with one arg

johnd0e:
Currently we have code (unsuccessfully) trying to handle this case, see https://github.com/Leaflet/Leaflet/issues/6013#issuecomment-623954835

This PR (originally based on #6032) fixes this case.

1. Two more variations of `off` added:

```js
// @alternative
// @function off(el: HTMLElement, types: String): this
// Removes all previously added listeners of given types.
```
```js
// @alternative
// @function off(el: HTMLElement): this
// Removes all previously added listeners from given HTMLElement
```

2. More strict arguments check to catch common bugs

3. All cases are covered by tests.

---
This change is [Reviewable](https://reviewable.io/reviews/leaflet/leaflet/7125)

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -23,7 +23,7 @@ import {getScale} from './DomUtil';
 // Adds a set of type/listener pairs, e.g. `{click: onClick, mousemove: onMouseMove}`
 export function on(obj, types, fn, context) {
 
-	if (typeof types === 'object') {
+	if (types && typeof types === 'object') {
 		for (var type in types) {
 			addOne(obj, type, types[type], fn);
 		}
@@ -54,22 +54,24 @@ var eventsKey = '_leaflet_events';
 // Removes all previously added listeners from given HTMLElement
 export function off(obj, types, fn, context) {
 	var type;
-	if (typeof types === 'object') {
-		for (type in types) {
-			removeOne(obj, type, types[type], fn);
-		}
-	} else if (types) {
-		types = Util.splitWords(types);
-
-		for (var i = 0, len = types.length; i < len; i++) {
-			removeOne(obj, types[i], fn, context);
-		}
-	} else {
+	if (arguments.length === 1) {
 		for (var id in obj[eventsKey]) {
 			type = id.split(/\d/)[0];
 			removeOne(obj, type, null, null, id);
 		}
 		delete obj[eventsKey];
+
+	} else if (types && typeof types === 'object') {
+		for (type in types) {
+			removeOne(obj, type, types[type], fn);
+		}
+
+	} else {
+		types = Util.splitWords(types);
+
+		for (var i = 0, len = types.length; i < len; i++) {
+			removeOne(obj, types[i], fn, context);
+		}
 	}
 
 	return this;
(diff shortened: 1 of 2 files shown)


src/dom/DomEvent.js
```diff
@@ -23,7 +23,7 @@ import {getScale} from './DomUtil';
 // Adds a set of type/listener pairs, e.g. `{click: onClick, mousemove: onMouseMove}`
 export function on(obj, types, fn, context) {
 
-	if (typeof types === 'object') {
+	if (types && typeof types === 'object') {
```
Falke-Design, line 26: Should we add a check `if(!types)` because now it will go into `Util.splitWords(types);` and throws an error (if types == `undefined`)

src/dom/DomEvent.js
```diff
@@ -48,23 +48,46 @@ var eventsKey = '_leaflet_events';
...
+		if (!filterFn || filterFn(type)) {
+			removeOne(obj, type, null, null, id);
+		}
+	}
+}
+
 export function off(obj, types, fn, context) {
 
-	if (typeof types === 'object') {
+	if (arguments.length === 1) {
```
Falke-Design, line 70: Same as here; https://github.com/Leaflet/Leaflet/pull/7518#pullrequestreview-794001157.
I think we should use `if(!types)` instead

src/dom/DomEvent.js
```diff
@@ -48,23 +48,46 @@ var eventsKey = '_leaflet_events';
...
 
-		for (var i = 0, len = types.length; i < len; i++) {
-			removeOne(obj, types[i], fn, context);
-		}
 	} else {
-		for (var j in obj[eventsKey]) {
-			removeOne(obj, j, obj[eventsKey][j]);
+		types = Util.splitWords(types);
+
+		if (arguments.length === 2) {
```
Falke-Design, line 82: Same as here; https://github.com/Leaflet/Leaflet/pull/7518#pullrequestreview-794001157.
I think we should use `if(!fn)` instead

Falke-Design (changes requested):
This will be very helpful!

mourner:
Same comment as in another PR — should we avoid using `arguments` to be on the safe side for performance?

src/dom/DomEvent.js
```diff
@@ -48,23 +48,46 @@ var eventsKey = '_leaflet_events';
...
+// @function off(el: HTMLElement): this
+// Removes all previously added listeners from given HTMLElement
+function batchRemove(obj, filterFn) {
+	for (var id in obj[eventsKey]) {
+		var type = id.split(/\d/)[0];
+		if (!filterFn || filterFn(type)) {
+			removeOne(obj, type, null, null, id);
+		}
+	}
+}
```
Falke-Design, line 66: I suggest to move this function below `off()` because else the documentation can be irritating

src/dom/DomEvent.js
johnd0e: > and throws an error (if types == `undefined`)

Because that is exactly what is desired - to catch user bug.

src/dom/DomEvent.js
johnd0e: Done

johnd0e:
> Same comment as in another PR — should we avoid using `arguments` to be on the safe side for performance?

@mourner 
Ref: https://github.com/Leaflet/Leaflet/pull/7518#pullrequestreview-798713147

mourner (approved):

PR #7138, opened 2020-05-10
Fix permanent opened tooltips for layers inside a LayerGroup.

### LINKED ISSUE #6938: FeatureGroup do not support tooltip with permanent: true (as advertised)

**Steps to reproduce**
Add a featureGroup with `bindToolip` to the map
Add a marker to the featureGroup

**Expected behavior**
every Marker should own (or share in this case) a tooltip which is permanently shown.
as documented "bindPopup binds a popup to all of the layers at once (likewise with bindTooltip)"

**Current behavior**
no Marker is shown
no Tooltip is shown 
in console I get "Error: Unable to get source layer LatLng."

**Environment**
- Leaflet version:
- Browser (with version):
- OS/Platform (with version):

**Additional context**

when removing permanent: true is removed everything works fine.

https://next.plnkr.co/edit/S2W7zcKTkVIrFLur

mondeja:
As was reported in #6938, when `bindTooltip` is called with `permanent: true` option on a `FeatureGroup` that doesn't have layers inside it, the tooltip tries to open itself and produces an `Unable to get source layer LatLng` error.

This pull fix the bug adding checks that prevents the tooltips open for layers that are instance of `LayerGroup` when `permanent: true` has been defined. The tooltips, in this case, are binded to `Layer` instances on `_layerAdd` methods. I think that this isn't the correct place to handle this exceptions. If so, consider this pull as a POC to fix this issue. 

Closes #6938.

### DIFF
--- a/spec/suites/layer/TooltipSpec.js
+++ b/spec/suites/layer/TooltipSpec.js
@@ -184,6 +184,55 @@ describe('Tooltip', function () {
 		expect(group._tooltip._container.innerHTML).to.be("I'm marker 2.");
 	});
 
+	it("it should use a tooltip with a string as content with a FeatureGroup", function () {
+		var marker1 = new L.Marker([55.8, 37.6]);
+		var marker2 = new L.Marker([54.6, 38.2]);
+		var group = new L.FeatureGroup([marker1, marker2]).addTo(map);
+
+		group.bindTooltip("hello");
+
+		// toggle popup on marker1
+		happen.mouseover(marker1._icon, {relatedTarget: map._container});
+		expect(map.hasLayer(group._tooltip)).to.be(true);
+		expect(group._tooltip._container.innerHTML).to.be("hello");
+
+		// toggle popup on marker2
+		happen.mouseover(marker2._icon, {relatedTarget: map._container});
+		expect(map.hasLayer(group._tooltip)).to.be(true);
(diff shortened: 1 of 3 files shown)


spec/suites/layer/TooltipSpec.js
```diff
@@ -184,6 +184,107 @@ describe('Tooltip', function () {
...
+
+		group.bindTooltip(function (layer) {
+			return layer.options.description;
+		}, {permanent: true});
+
+		expect(map.hasLayer(marker1._tooltip)).to.be(true);
+		expect(marker1._tooltip._container.innerHTML).to.be("I'm marker 1.");
+
+		expect(map.hasLayer(marker2._tooltip)).to.be(true);
+		expect(marker2._tooltip._container.innerHTML).to.be("I'm marker 2.");
```
johnd0e, line 218:
```suggestion
		var group = new L.FeatureGroup([marker1, marker2]);
		group.bindTooltip(function (layer) {
			return layer.options.description;
		}, {permanent: true})

		group.addTo(map);

		expect(map.hasLayer(group._tooltip)).to.be(true);
		expect(group._tooltip._container.innerHTML).to.be("I'm marker 1.");
```
That's how it is untended to be: all FeatureGroup's members share single tooltip.

Not much sense to use with `permanent: true`, though I can imagine a couple of useful use cases.

PR #7140, opened 2020-05-12
Added tooltip rotate option

ryanlovett-au:
Provide the option for a user to define a rotate angle in degrees as part of the standard tooltip options object and, using the existing transform function, will apply a CSS rotate that will provide for vertical or angled tooltip labels.


ryanlovett-au:
This is my first pull request.. be gentle :)

ryanlovett-au:
Examples albeit with multiple stacked labels, but illustrates the point.

image

ronikar:
Hi @ryanlovett-au,
I like the idea :+1: . But, in my opinion, it should be a plugin.

johnd0e:
Related: #6183

Issue #7099, opened 2020-04-18
Tests for GeoJSON functions required


johnd0e:
None of following [GeoJSON functions](https://leafletjs.com/reference-1.6.0.html#geojson-geometrytolayer) are covered with tests yet:
- [ ] `geometryToLayer`
- [ ] `coordsToLatLng`
- [ ] `coordsToLatLngs`
- [ ] `latLngToCoords`
- [ ] `latLngsToCoords`
- [ ] `asFeature`

Sample test (for `latLngToCoords`):
```js
describe("L.GeoJSON functions", function () {
	it("L.GeoJSON.latLngToCoords", function () {
		// todo: check alt too
		var arrLatLng = [-1.4837191022531273, 43.49222084042808];
		var arrLngLat = [arrLatLng[1], arrLatLng[0]];
		var latLng = L.latLng(arrLatLng);

		expect(L.GeoJSON.latLngToCoords(latLng)).to.eql([43.492221, -1.483719]);
		expect(L.GeoJSON.latLngToCoords(latLng, 0)).to.eql([43, -1]);
		// expect(L.GeoJSON.latLngToCoords(latLng, false)).to.eql(arrLngLat); // after #7100
	});

	// etc
});
```

adriancerejo:
I am pretty new to open-source contributions but could I try adding one or few of these tests?

phloose:
@johnd0e i opened a ready-for-review pull request that adds the respective tests.

devadeka:
If this is still open could I have a crack at it for hacktoberfest?
Ta

phloose:
i proposed a pull request (#7147) in may, which after an initial discussion did not get further attention. So maybe someone could review it?! @johnd0e @mourner ?

PR #7147, opened 2020-05-14
Tests for L.GeoJSON static functions

phloose:
This adds the required tests for the L.GeoJSON static functions mentioned in #7099.

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -684,4 +684,23 @@ describe("L.GeoJSON functions", function () {
 			}
 		});
 	});
+
+	describe("#coordsToLatLng", function () {
+		it("creates a LatLng object given longitude and latitude", function () {
+			const latlng = L.GeoJSON.coordsToLatLng([0, 0]);
+			expect(latlng instanceof L.LatLng).to.be(true);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.alt).to.be(undefined);
+		});
+
+		it("creates a LatLng object given longitude, latitude and altitude", function () {
+			const latlng = L.GeoJSON.coordsToLatLng([0, 0, 0]);
+			expect(latlng instanceof L.LatLng).to.be(true);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.alt).to.be(0);
+		});
+
+	});
 });
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -577,6 +577,10 @@ describe("L.GeoJSON functions", function () {
 			});
 		}
 
+		function customCoordstoLatLng(coords) {
+			return new L.LatLng(coords[1] + 1, coords[0] + 1, coords[2] + 1);
+		}
+
 		[
 			point, line, polygon,
 			multiPoint, multiLine, multiPolygon,
@@ -643,6 +647,41 @@ describe("L.GeoJSON functions", function () {
 				expect(layer._layers[id] instanceof L.Circle).to.be(true);
 				expect(layer._layers[id].options.radius).to.be(100);
 			});
+		});
+
+		it("creates a Layer using coordsToLatLng option (Point)", function () {
+			const layer = L.GeoJSON.geometryToLayer({
+				type: "Feature",
+				geometry: {
+					type: "Point",
+					coordinates: [1, 2, 3]
+				}
+			}, {
+				coordsToLatLng: customCoordstoLatLng
+			});
+			expect(layer._latlng).to.eql({lat: 3, lng: 2, alt: 4});
+		});
+
+		it("creates a Layer using coordsToLatLng option (MultiPoint)", function () {
+			const layer = L.GeoJSON.geometryToLayer({
+				type: "Feature",
+				geometry: {
+					type: "MultiPoint",
+					coordinates: [
+						[1, 2, 3], [4, 5, 6]
+					]
+				}
+			}, {
+				coordsToLatLng: customCoordstoLatLng
+			});
+			const expected = [
+				{lat: 3, lng: 2, alt: 4},
+				{lat: 6, lng: 5, alt: 7}
+			];
+			const ids = Object.keys(layer._layers);
+			for (var i = 0; i < ids.length; i++) {
+				expect(layer._layers[ids[i]]._latlng).to.eql(expected[i]);
+			}
 		});
 	});
 });
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -748,24 +748,31 @@ describe("L.GeoJSON functions", function () {
 	});
 
 	describe("#latLngToCoords", function () {
-		it("returns an array of coordinates and altitude (longitude, latitude, altitude)", function () {
-			var coords = L.GeoJSON.latLngToCoords(L.latLng(2, 1, 3));
-			expect(coords).to.eql([1, 2, 3]);
+		it("returns an array of coordinates and altitude", function () {
+			var coords = L.GeoJSON.latLngToCoords(L.latLng(2, 1));
+			var coordsWithAlt = L.GeoJSON.latLngToCoords(L.latLng(2, 1, 3));
+			expect(coords).to.eql([1, 2]);
+			expect(coordsWithAlt).to.eql([1, 2, 3]);
 		});
 
-		it("returns an array of coordinates with given precision (longitude, latitude, altitude)", function () {
+		it("returns an array of coordinates with given precision", function () {
 			var coords = L.GeoJSON.latLngToCoords(L.latLng(
+				2.123456, 1.123456
+			), 3);
+			var coordsWithAlt = L.GeoJSON.latLngToCoords(L.latLng(
 				2.123456, 1.123456, 3.123456
 			), 3);
-			expect(coords).to.eql([1.123, 2.123, 3.123]);
+			expect(coords).to.eql([1.123, 2.123]);
+			expect(coordsWithAlt).to.eql([1.123, 2.123, 3.123]);
 		});
 	});
 
 	describe("#latLngsToCoords", function () {
-		it("returns a multidimensional array of coordinates (latitude, longitude, altitude)", function () {
-			var latLngs = [L.latLng(2, 1, 3), L.latLng(5, 4, 6)];
-			var coords = L.GeoJSON.latLngsToCoords(latLngs);
-			expect(coords).to.eql([[1, 2, 3], [4, 5, 6]]);
+		it("returns a multidimensional array of coordinates", function () {
+			var coords = L.GeoJSON.latLngsToCoords([L.latLng(2, 1), L.latLng(4, 3)]);
+			var coordWithAlt = L.GeoJSON.latLngsToCoords([L.latLng(2, 1, 3), L.latLng(5, 4, 6)]);
+			expect(coords).to.eql([[1, 2], [3, 4]]);
+			expect(coordWithAlt).to.eql([[1, 2, 3], [4, 5, 6]]);
 		});
 
 		it("returns a multidimensional array of coordinates (levelDeep=1)", function () {
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -842,6 +842,57 @@ describe("L.GeoJSON functions", function () {
 			const coords = L.GeoJSON.latLngsToCoords(latLngs);
 			expect(coords).to.eql([[0, 0, 0], [1, 1, 1]]);
 		});
+	});
 
+	describe("#asFeature", function () {
+		const geometry1 = {
+			type: "Point",
+			coordinates: [0, 0]
+		};
+
+		const geometry2 = {
+			type: "Point",
+			coordinates: [1, 1]
+		};
+
+		const feature1 = {
+			type: "Feature",
+			geometry: geometry1,
+			properties: {a: 1}
+		};
+
+		const feature2 = {
+			type: "Feature",
+			geometry: geometry2,
+			properties: {b: 2}
+		};
+
+		const featureCollection = {
+			type: "FeatureCollection",
+			features: [
+				feature1,
+				feature2
+			]
+		};
+
+		it("given a bare geometry returns a GeoJSON-like feature", function () {
+			const ret = L.GeoJSON.asFeature(geometry1);
+			const expected = {
+				type: "Feature",
+				properties: {},
+				geometry: geometry1
+			};
+			expect(ret).to.eql(expected);
+		});
+
+		it("given a GeoJSON feature directly returns it", function () {
+			const ret = L.GeoJSON.asFeature(feature1);
+			expect(ret).to.eql(feature1);
+		});
+
+		it("given a GeoJSON feature collection directly returns it", function () {
+			const ret = L.GeoJSON.asFeature(featureCollection);
+			expect(ret).to.eql(featureCollection);
+		});
 	});
 });
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -515,36 +515,36 @@ describe("L.LayerGroup#toGeoJSON", function () {
 
 describe("L.GeoJSON functions", function () {
 	describe("#geometryToLayer", function () {
-		const point = {
+		var point = {
 			type: "Point",
 			coordinates: [0, 0]
 		};
-		const multiPoint  = {
+		var multiPoint  = {
 			type: "MultiPoint",
 			coordinates: [
 				[0, 0], [10, 10]
 			]
 		};
-		const line =  {
+		var line =  {
 			type: "LineString",
 			coordinates: [
 				[0, 0], [10, 10], [20, 20]
 			]
 		};
-		const multiLine = {
+		var multiLine = {
 			type: "MultiLineString",
 			coordinates: [
 				[[10, 10], [20, 20], [30, 30]],
 				[[50, 50], [60, 60], [70, 70]]
 			]
 		};
-		const polygon = {
+		var polygon = {
 			type: "Polygon",
 			coordinates: [
 				[[30, 10], [40, 40], [20, 40], [10, 20], [30, 10]]
 			]
 		};
-		const multiPolygon = {
+		var multiPolygon = {
 			type: "MultiPolygon",
 			coordinates: [
 				[
@@ -555,7 +555,7 @@ describe("L.GeoJSON functions", function () {
 				]
 			]
 		};
-		const geometryCollection  = {
+		var geometryCollection  = {
 			type: "GeometryCollection",
 			geometries: [
 				{
@@ -572,13 +572,13 @@ describe("L.GeoJSON functions", function () {
 		};
 
 		function customPointToLayer(geojsonPoint, latLng) {
-			return new L.Circle(latLng, {
+			return L.circle(latLng, {
 				radius: geojsonPoint.properties.radius
 			});
 		}
 
 		function customCoordstoLatLng(coords) {
-			return new L.LatLng(coords[1] + 1, coords[0] + 1, coords[2] + 1);
+			return L.latLng(coords[1] + 1, coords[0] + 1, coords[2] + 1);
 		}
 
 		[
@@ -587,11 +587,11 @@ describe("L.GeoJSON functions", function () {
 			geometryCollection
 		].forEach(function (geometry) {
 			it("creates a Layer from a GeoJSON feature (type='" + geometry.type + "')", function () {
-				const layer = L.GeoJSON.geometryToLayer({
+				var layer = L.GeoJSON.geometryToLayer({
 					type: "Feature",
 					geometry: geometry
 				});
-				expect(layer instanceof L.Layer).to.be(true);
+				expect(layer).to.be.a(L.Layer);
 			});
 		});
 
@@ -601,56 +601,54 @@ describe("L.GeoJSON functions", function () {
 			geometryCollection
 		].forEach(function (geometry) {
 			it("creates a Layer from a GeoJSON geometry (type='" + geometry.type + "')", function () {
-				const layer = L.GeoJSON.geometryToLayer(geometry);
-				expect(layer instanceof L.Layer).to.be(true);
+				var layer = L.GeoJSON.geometryToLayer(geometry);
+				expect(layer).to.be.a(L.Layer);
 			});
 		});
 
 		it("throws an error if feature is an invalid GeoJSON object", function () {
-			expect(function () {
-				L.GeoJSON.geometryToLayer({
-					type: "Feature",
-					geometry: {
-						type: "invalid",
-						coordinates: [0, 0]
-					}
-				});
+			expect(L.GeoJSON.geometryToLayer).withArgs({
+				type: "Feature",
+				geometry: {
+					type: "invalid",
+					coordinates: [0, 0]
+				}
 			}).to.throwError("Invalid GeoJSON object.");
 		});
 
-		it("returns null if feature does not have a geometry property", function () {
-			const ret = L.GeoJSON.geometryToLayer({type: "Feature"});
-			expect(ret).to.be(null);
+		it("returns nothing if feature does not have a geometry property", function () {
+			var ret = L.GeoJSON.geometryToLayer({type: "Feature"});
+			expect(ret).not.to.be.ok();
 		});
 
 		it("creates a Layer using pointToLayer option (Point)", function () {
-			const layer = L.GeoJSON.geometryToLayer({
+			var layer = L.GeoJSON.geometryToLayer({
 				type: "Feature",
 				geometry: point,
 				properties: {radius: 100}
 			}, {
 				pointToLayer: customPointToLayer
 			});
-			expect(layer instanceof L.Circle).to.be(true);
+			expect(layer).to.be.a(L.Circle);
 			expect(layer.options.radius).to.be(100);
 		});
 
 		it("creates a Layer using pointToLayer option (MultiPoint)", function () {
-			const layer = L.GeoJSON.geometryToLayer({
+			var layer = L.GeoJSON.geometryToLayer({
 				type: "Feature",
 				geometry: multiPoint,
 				properties: {radius: 100}
 			}, {
 				pointToLayer: customPointToLayer
 			});
-			Object.keys(layer._layers).forEach(function (id) {
-				expect(layer._layers[id] instanceof L.Circle).to.be(true);
-				expect(layer._layers[id].options.radius).to.be(100);
+			layer.eachLayer(function (lyr) {
+				expect(lyr).to.be.a(L.Circle);
+				expect(lyr.options.radius).to.be(100);
 			});
 		});
 
 		it("creates a Layer using coordsToLatLng option (Point)", function () {
-			const layer = L.GeoJSON.geometryToLayer({
+			var layer = L.GeoJSON.geometryToLayer({
 				type: "Feature",
 				geometry: {
 					type: "Point",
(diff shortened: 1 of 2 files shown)


phloose:
the pipeline seems to still be running but when clicking on details its shows up as finished...

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			});
+		});
+
+		it("creates a multidimensional array of LatLngs (levelsDeep=1)", function () {
+			const latLngs = L.GeoJSON.coordsToLatLngs([
+				[[0, 0], [1, 1], [2, 2]],
+				[[4, 4], [5, 5], [6, 6]]
+			], 1);
+			expect(latLngs.length).to.be(2);
+			for (var i = 0; i < latLngs.length; i++) {
```
johnd0e, line 726: Keep it simpler, avoid `for` where you can use `forEach`.
May be even https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			const expected = [
+				{lat: 2, lng: 3, alt: 4},
+				{lat: 5, lng: 6, alt: 7},
+				{lat: 8, lng: 9, alt: 10}
+			];
+			for (var i = 0; i < latLngs.length; i++) {
+				expect(latLngs[i]).to.eql(expected[i]);
+			}
+		});
+
```
johnd0e, line 745: Same here

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+				[
+					{lat: 13, lng: 14, alt: 15},
+					{lat: 16, lng: 17, alt: 18}
+				]
+			];
+			for (var i = 0; i < latLngs.length; i++) {
+				for (var j = 0; j < latLngs[i].length; j++) {
+					expect(latLngs[i][j]).to.eql(expected[i][j]);
+				}
+			}
```
johnd0e, line 766: Perhaps you can check it without explicit loop, like this:
```
			expect(latLngs).to.be.eql([
				[
					{lat: 2, lng: 3, alt: 4},
					{lat: 5, lng: 6, alt: 7}
				],
				[
					{lat: 13, lng: 14, alt: 15},
					{lat: 16, lng: 17, alt: 18}
				]
			]);
```

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+	});
+
+	describe("#latLngToCoords", function () {
+		it("returns an array of coordinates (longitude, latitude)", function () {
+			const coords = L.GeoJSON.latLngToCoords(new L.LatLng(0, 0));
+			expect(coords).to.eql([0, 0]);
+		});
+
+		it("returns an array of coordinates and altitude (longitude, latitude, altitude)", function () {
+			const coords = L.GeoJSON.latLngToCoords(new L.LatLng(0, 0, 0));
```
johnd0e, line 777: You are using [1,1], `[0,0]` and `[0,0,0]` everywhere, and it is bad because
- `0` is rather special value. Consider `0*10===0*100`
- when testing tuples better use different values (like `[1,2,3]`) to be able to catch more types of errors.

Please revise all the tests with this consideration

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			expect(coords).to.eql([1.123, 1.123]);
+		});
+
+		it("returns an array of coordinates with given precision (longitude, latitude, altitude)", function () {
+			const coords = L.GeoJSON.latLngToCoords(new L.LatLng(
+				1.123456, 1.123456, 1.123456
+			), 3);
+			expect(coords).to.eql([1.123, 1.123, 1.123]);
+		});
+	});
```
johnd0e, line 794: Better combine these two tests into one.

(Perhaps there are more such tests that should be combined)

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			const latLngs = [
+				[L.latLng(0, 0), L.latLng(1, 1)],
+				[L.latLng(2, 2), L.latLng(3, 3)]
+			];
+			const coords = L.GeoJSON.latLngsToCoords(latLngs, 1);
+			const expected = [
+				[[0, 0], [1, 1]],
+				[[2, 2], [3, 3]]
+			];
+			expect(coords).to.eql(expected);
```
johnd0e, line 812: May be matter of taste, but I feel this more expressive
```suggestion
			expect(coords).to.eql([
				[[0, 0], [1, 1]],
				[[2, 2], [3, 3]]
			]);
```

There are many such places in your code, so I will not comment each of them.

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+		};
+
+		function customPointToLayer(geojsonPoint, latLng) {
+			return new L.Circle(latLng, {
+				radius: geojsonPoint.properties.radius
+			});
+		}
+
+		function customCoordstoLatLng(coords) {
+			return new L.LatLng(coords[1] + 1, coords[0] + 1, coords[2] + 1);
```
johnd0e, line 581: But in other part you prefer factory form (`L.latLng()`).
You should be consistent everywhere - either prefer constructors (with `new`), either factories.

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			point, line, polygon,
+			multiPoint, multiLine, multiPolygon,
+			geometryCollection
+		].forEach(function (geometry) {
+			it("creates a Layer from a GeoJSON feature (type='" + geometry.type + "')", function () {
+				const layer = L.GeoJSON.geometryToLayer({
+					type: "Feature",
+					geometry: geometry
+				});
+				expect(layer instanceof L.Layer).to.be(true);
```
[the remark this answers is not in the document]
phloose: why is that a better way? i am just curious

spec/suites/layer/GeoJSONSpec.js
phloose: I just thought it would be more readable and since it doesnt matter if these are real wgs84 coordinates or just cartesian i used the most simple ones. Will change that!

spec/suites/layer/GeoJSONSpec.js
phloose: Yeah true! Just saw that too late and then it was pushed. I wanted to wait for a review first...

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+					type: "Feature",
+					geometry: {
+						type: "invalid",
+						coordinates: [0, 0]
+					}
+				});
+			}).to.throwError("Invalid GeoJSON object.");
+		});
+
+		it("returns null if feature does not have a geometry property", function () {
```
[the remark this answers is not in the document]
phloose: ok, i just looked at the methods implementation and there it returns null, even though it is not mentioned in the doc string. Or what do you mean with specs? Strictly speaking i would also have to remove this one https://github.com/Leaflet/Leaflet/pull/7147#discussion_r429548120 as it is also not mentioned in the doc string.

spec/suites/layer/GeoJSONSpec.js
phloose: I didn't know that you can check nested structures this way (which is really cool btw....)

spec/suites/layer/GeoJSONSpec.js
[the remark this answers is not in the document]
johnd0e: It's more expressive.
Besides it is shorter

spec/suites/layer/GeoJSONSpec.js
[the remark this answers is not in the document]
johnd0e: By 'specs' I mean explicitly documented features. Only on them we can rely.
All internal implementation details are subject to change.

spec/suites/layer/GeoJSONSpec.js
johnd0e: Sure, pick simple. But not toooo simple)
E.g. have you mentioned that geojson order is [lng,lat], while leaflet's is opposite: [lat,lng]. Order is important, that's why we should pick tuples where lat value differs from lng.

phloose: sure, that makes sense!

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,387 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+		});
+
+		it("creates a LatLng object given longitude, latitude and altitude", function () {
+			const latlng = L.GeoJSON.coordsToLatLng([0, 0, 0]);
+			expect(latlng instanceof L.LatLng).to.be(true);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.lat).to.be(0);
+			expect(latlng.alt).to.be(0);
+		});
+	});
```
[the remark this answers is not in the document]
phloose: yeah a stupid copy/paste failure... but this tests is for the case when you use lon, lat and alt

spec/suites/layer/GeoJSONSpec.js
phloose: But why combine them? They are two different cases.

spec/suites/layer/GeoJSONSpec.js
phloose: tried flat, doesnt work. Using forEach two times now.

spec/suites/layer/GeoJSONSpec.js
johnd0e: Well, I agree that this is rather matter of taste.
As for this case, I feel that `[lat, lng]` is kind of special case of general `[lat, lng, alt]`.
Sure, there is nothing wrong to put it into separate test, but see:
- having too many discrete tests can be a bit confusing
- we do not lose anything if we put two similar tests together, as when test fails - we see where it fails (exact line).

phloose: > As for this case, I feel that `[lat, lng]` is kind of special case of general `[lat, lng, alt]`.

I think its more the general case to only have `[lat, lng]`, as until now i have never had any GeoJSON dataset that also had altitude values.  

> having too many discrete tests can be a bit confusing

Well, in this case the test description clearly tells what is tested, so i don't see any confusion here to be honest :)

> we do not lose anything if we put two similar tests together, as when test fails - we see where it fails (exact line).

Sure you are right, but with two different cases you instantly know which of the test failed by description. You wouldn't have to read line numbers. But yeah, you are right this is a matter of taste. Personally i prefer to put such cases into seperate tests, as for me DRY doesn't necessarily count in tests. I will put those two tests into one.

phloose:
Should i squash all commits when you approve this or is it done automatically?

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -782,9 +762,10 @@ describe("L.GeoJSON functions", function () {
 	});
 
 	describe("#latLngsToCoords", function () {
-		it("returns a multidimensional array of coordinates", function () {
-			var coords = L.GeoJSON.latLngsToCoords([L.latLng(2, 1), L.latLng(4, 3)]);
-			expect(coords).to.eql([[1, 2], [3, 4]]);
+		it("returns a multidimensional array of coordinates (latitude, longitude, altitude)", function () {
+			var latLngs = [L.latLng(2, 1, 3), L.latLng(5, 4, 6)];
+			var coords = L.GeoJSON.latLngsToCoords(latLngs);
+			expect(coords).to.eql([[1, 2, 3], [4, 5, 6]]);
```
johnd0e: In fact I'd prefer to keep both cases.
```js
		it("returns a multidimensional array of coordinates", function () {
			var lnglats = L.GeoJSON.latLngsToCoords([L.latLng(2, 1), L.latLng(4, 3)]);
			var lnglatalts = L.GeoJSON.latLngsToCoords([L.latLng(2, 1, 3), L.latLng(5, 4, 6)]);
			expect(lnglats).to.eql([[1, 2], [3, 4]]);
			expect(lnglatalts).to.eql([[1, 2, 3], [4, 5, 6]]);
```
or
```js
		it("returns a multidimensional array of coordinates", function () {
			var latlngs = [L.latLng(2, 1), L.latLng(4, 3)];
			var latlngalts = [L.latLng(2, 1, 3), L.latLng(5, 4, 6)];
			expect(L.GeoJSON.latLngsToCoords(latlngs)).to.eql([[1, 2], [3, 4]]);
			expect(L.GeoJSON.latLngsToCoords(latlngalts)).to.eql([[1, 2, 3], [4, 5, 6]]);
```
or
```js
		it("returns a multidimensional array of coordinates", function () {
			expect(L.GeoJSON.latLngsToCoords)
				.withArgs([L.latLng(2, 1), L.latLng(4, 3)])
				.to.eql([[1, 2], [3, 4]]);
			expect(L.GeoJSON.latLngsToCoords)
				.withArgs([L.latLng(2, 1, 3), L.latLng(5, 4, 6)])
				.to.eql([[1, 2, 3], [4, 5, 6]]);
```

johnd0e: (and same for other cases)

spec/suites/layer/GeoJSONSpec.js
johnd0e: > you instantly know which of the test failed by description. You wouldn't have to read line numbers

But to fix broken case you still need to inspect exact line.
Pay attention, some other your tests already have several `expect` checks (btw, test may fail not only because of `expect`).

phloose: > btw, test may fail not only because of expect 

huh? what do you mean exactly with that? programming errors? just curious...

spec/suites/layer/GeoJSONSpec.js
phloose: Ok will change that. Unfortunatly the more shorter form you proposed (which i would prefer)
```javascript
it("returns a multidimensional array of coordinates", function () {
			expect(L.GeoJSON.latLngsToCoords)
				.withArgs([L.latLng(2, 1), L.latLng(4, 3)])
				.to.eql([[1, 2], [3, 4]]);
			expect(L.GeoJSON.latLngsToCoords)
				.withArgs([L.latLng(2, 1, 3), L.latLng(5, 4, 6)])
				.to.eql([[1, 2, 3], [4, 5, 6]]);
```

 seems to only work for catching exceptions. It cannot test against a returned value.

johnd0e: Indeed..

johnd0e:
> Should i squash all commits when you approve this or is it done automatically?

Automatically

spec/suites/layer/GeoJSONSpec.js
johnd0e: `expect` is just special way to throw exception when some expectation not met.
Obviously that if any part of code throws when it shouldn't - test also fails.

phloose: ok yeah thats true... test code is just code that can have logical or syntax errors

phloose:
So is there anything left to do? ;)

phloose:
Any news on this? @mourner maybe?

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+
+		function customCoordsToLatLng(coords) {
+			return L.latLng(coords[1] + 1, coords[0] + 1, coords[2] + 1);
+		}
+
+		it("creates a multidimensional array of LatLngs", function () {
+			var latLngs = L.GeoJSON.coordsToLatLngs([[1, 2], [3, 4], [5, 6]]);
+			expect(latLngs.length).to.be(3);
+			latLngs.forEach(function (latLng) {
+				expect(latLng).to.be.a(L.LatLng);
```
johnd0e, line 714: Better compare exact result with `eql`.

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+
+		it("creates a multidimensional array of LatLngs (levelsDeep=1)", function () {
+			var latLngs = L.GeoJSON.coordsToLatLngs([
+				[[1, 2], [3, 4], [5, 6]],
+				[[5, 6], [7, 8], [9, 10]]
+			], 1);
+			expect(latLngs.length).to.be(2);
+			latLngs.forEach(function (arr) {
+				arr.forEach(function (latlng) {
+					expect(latlng).to.be.a(L.LatLng);
```
johnd0e, line 726: Same here

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+					{lat: 14, lng: 13, alt: 15},
+					{lat: 17, lng: 16, alt: 18}
+				]
+			]);
+		});
+	});
+
+	describe("#latLngToCoords", function () {
+		it("returns an array of coordinates and altitude", function () {
+			var coords = L.GeoJSON.latLngToCoords(L.latLng(2, 1));
```
johnd0e, line 762: Here and in many places below: better replace occurrences `L.latLng(2, 1)` with arrays `[2,1]`.

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+
+	describe("#coordsToLatLng", function () {
+		it("creates a LatLng object", function () {
+			var latLng = L.GeoJSON.coordsToLatLng([1, 2]);
+			var latLngWithAlt = L.GeoJSON.coordsToLatLng([3, 4, 5]);
+			expect(latLng).to.be.a(L.LatLng);
+			expect(latLngWithAlt).to.be.a(L.LatLng);
+		});
+
+		it("creates a LatLng object with given coordinates", function () {
```
johnd0e, line 692: I'd suggest to unite this with prev. test

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			var latLng = L.GeoJSON.coordsToLatLng([1, 2]);
+			var latLngWithAlt = L.GeoJSON.coordsToLatLng([3, 4, 5]);
+			expect(latLng.lat).to.be(2);
+			expect(latLng.lng).to.be(1);
+			expect(latLng.alt).not.to.be.ok();
+			expect(latLngWithAlt.lat).to.be(4);
+			expect(latLngWithAlt.lng).to.be(3);
+			expect(latLngWithAlt.alt).to.be(5);
```
johnd0e, line 700:
```suggestion
expect( L.GeoJSON.coordsToLatLng([1, 2])).to.eql({lng:1, lat:2});
expect(L.GeoJSON.coordsToLatLng([3, 4, 5])).to.eql({lng:3, lat:4, alt:5});
```

johnd0e (changes requested):

spec/suites/layer/GeoJSONSpec.js
```diff
@@ -512,3 +512,364 @@ describe("L.LayerGroup#toGeoJSON", function () {
...
+			point, line, polygon,
+			multiPoint, multiLine, multiPolygon,
+			geometryCollection
+		].forEach(function (geometry) {
+			it("creates a Layer from a GeoJSON feature (type='" + geometry.type + "')", function () {
+				var layer = L.GeoJSON.geometryToLayer({
+					type: "Feature",
+					geometry: geometry
+				});
+				expect(layer).to.be.a(L.Layer);
```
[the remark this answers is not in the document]
phloose: I explicitly check for L.Layer because the documentation states that this function returns a L.Layer

spec/suites/layer/GeoJSONSpec.js
phloose: The [last review](https://github.com/Leaflet/Leaflet/pull/7147#discussion_r429552996) you told me, i should use `to.be.a` in favor of `expect(latLng instanceof L.LatLng).to.be(true)` which i find reasonable, but i don't get why i should use `eql` because i am asserting against a type and find the test with `.to.be.a` more readable.

spec/suites/layer/GeoJSONSpec.js
phloose: See https://github.com/Leaflet/Leaflet/pull/7147#discussion_r559741035

spec/suites/layer/GeoJSONSpec.js
phloose: The docs for this function tell that a `L.LatLng` is used a input: https://leafletjs.com/reference-1.7.1.html#geojson-latlngtocoords.

spec/suites/layer/GeoJSONSpec.js
phloose: The previous test should only test that the `L.GeoJSON.coordsToLatLng` function returns a L.LatLng object. This test only makes assertions against coordinates. For me these are two different cases and i learned unit testing to be more fine grained and to have in the best case only one assertion per test.

spec/suites/layer/GeoJSONSpec.js
phloose: Done with 44e8e70

spec/suites/layer/GeoJSONSpec.js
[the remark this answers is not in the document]
johnd0e: > and the docs reflect that in the way that it "only" returns a `L.Layer`.

The docs are brief, that is true, and many things are implied.
Atm every GeoJSON geometry has unambiguous Leaflet's match, except Point (that's why we have `pointToLayer` option).
As for the tests: there is no harm if they are slightly stricter that docs, as everyone can adapt them in case of need.
But believe me: you will want to know if someone accidently change returned layer type.

spec/suites/layer/GeoJSONSpec.js
johnd0e: > `.to.be.a` more readable.

It is more readable when you want to test a type of value only.
But here I request you not just test against a type, but also check exact values.
Well, you could do it in separate test, I suppose that it would be overkill.

spec/suites/layer/GeoJSONSpec.js
johnd0e: > i learned unit testing to be more fine grained and to have in the best case only one assertion per test.

I suppose there can be different approaches for same thing.
For this case: I prefer to keep some balance between verbosity and maintainability.

spec/suites/layer/GeoJSONSpec.js
johnd0e: Sure, that is in specs, and your tests are correct. But it is also documented that you can pass plain array in every place where LatLng is expected.
Definitely you can double the number of tests and check both cases, but in my opinion that would be very excessive.
So I suggest you to keep just array form, because:
1) It is shorter and more expressive
2) You already prefer array in all the previous cases, so it make sense to keep single style.

johnd0e (changes requested):
I've added some comments.
Please do not mark them as resolved until I actually can read your answers.

spec/suites/layer/GeoJSONSpec.js
phloose: So i should not assert against the type and instead against the single values (coordinates)?

spec/suites/layer/GeoJSONSpec.js
phloose: I can change that, but like i said in this case i tried to stick to the documentation. The other cases like `coordsToLatLng` or `coordsToLatLng` accept, according to the docs, an array of coordinates and thats the reason i chose the array over `LatLng`.

spec/suites/layer/GeoJSONSpec.js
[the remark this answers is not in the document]
phloose: Done with b57c7da

spec/suites/layer/GeoJSONSpec.js
phloose: Done with cc74c2c

spec/suites/layer/GeoJSONSpec.js
phloose: Done with cc74c2c

spec/suites/layer/GeoJSONSpec.js
phloose: Done with 32ef194

spec/suites/layer/GeoJSONSpec.js
johnd0e: As I said your tests are correct but too verbose (as for my taste).

And to be consistent you also should consider [this](https://leafletjs.com/reference-1.7.1.html#latlng) part of docs:
> All Leaflet methods that accept LatLng objects also accept them in a simple Array form and simple object form (unless noted otherwise).

P.S.
I've also find these 2 cases where you are not checking resulting types:

https://github.com/Leaflet/Leaflet/blob/cc74c2c8a11dcc4574afa1d746cb2dc0696a31ea/spec/suites/layer/GeoJSONSpec.js#L729

https://github.com/Leaflet/Leaflet/blob/cc74c2c8a11dcc4574afa1d746cb2dc0696a31ea/spec/suites/layer/GeoJSONSpec.js#L739

It's not so important (for my taste), because such type of error hard to make by accident, and it will manifest itself in some other part anyway.
But it also makes sense to keep all your tests consistent, so...

phloose: > All Leaflet methods that accept LatLng objects also accept them in a simple Array form and simple object form (unless noted otherwise).

That may be true but for this specific case the function to be used is _explicitly_ called `latLngsToCoords`. It does not make sense to me to use an array when the function is named after this specific type and also used for converting a `LatLng` to an array of coordinates. Furthermore i tested your suggestion and **it does not work, the tests fail**. The code for `latLngToCoords` and `latLngsToCoords` uses properties of the latLng object (`lat`, `lng` and maybe `alt`) which obviously do not exist when you instead pass an array with just numbers.

> It's not so important (for my taste), because such type of error hard to make by accident, and it will manifest itself in some other part anyway.
> But it also makes sense to keep all your tests consistent, so...

Ok i can also change that, but i did not add this type check here because an array is not kind of a self invented object type like `LatLng`. But if you prefer i can add another `forEach` loop to these two lines and check for the array type...

johnd0e: > It does not make sense to me to use an array when the function is named after this specific type

There may be some use cases.

> Furthermore i tested your suggestion and **it does not work, the tests fail**.

Good to know, perhaps this should fall into "unless noted otherwise", so we need to update docs.

> and check for the array type...

No need to check for Array, but you may want to check that it contains proper LatLngs.

phloose: > No need to check for Array, but you may want to check that it contains proper LatLngs.

Sorry mate, i don't get it... What is the point of this? Have you looked at the [respective lines in the source code](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/layer/GeoJSON.js#L256)? The function `latLngToCoords` or `latLngsToCoords` returns an array of coordinates/numbers not `LatLng`s. In the tests you mention i already check if the returned array contains the correct numbers that were given to the `LatLng` factory function. Where should i test then for proper `LatLng`s

johnd0e: > The function `latLngToCoords` or `latLngsToCoords` returns an array of coordinates/numbers not `LatLng`s.

Again:
https://github.com/Leaflet/Leaflet/blob/cc74c2c8a11dcc4574afa1d746cb2dc0696a31ea/spec/suites/layer/GeoJSONSpec.js#L729

https://github.com/Leaflet/Leaflet/blob/cc74c2c8a11dcc4574afa1d746cb2dc0696a31ea/spec/suites/layer/GeoJSONSpec.js#L739

But you are right, no need to check types here, because they depend on `customCoordsToLatLng`.
So no objections left.

johnd0e (approved):
Good work!

spec/suites/layer/GeoJSONSpec.js
johnd0e: > Good to know, perhaps this should fall into "unless noted otherwise", so we need to update docs.

#7425

phloose:
Any chance this will ever be merged 😄?

Issue #2235, opened 2013-11-25
Add 'change' event for paths


Zverik:
L.Polyline.setLatLngs() and other similar methods affect object parameters, but there is no way to catch those events. I propose to add `change` event to L.Path subclasses that get called when a layer (in this case — an array of coordinates) is changed.

johnd0e:
`change` is too wide term, so either chose some more appropriate, either stick to `move`.
https://github.com/Leaflet/Leaflet/issues/6492#issuecomment-813499342

Issue #6492, opened 2019-01-25
Inconsistent `move` event on setLatLng


codeofsumit:
Hi leaflet team,

by tackling [this issue](https://github.com/codeofsumit/leaflet.pm/issues/357) in the leaflet.pm plugin we have encountered that the [CircleMarker class is firing the `move` event when `setLatLng` is called](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/CircleMarker.js#L38). No other class is doing that to my knowledge as it is a [long standing feature request](https://github.com/Leaflet/Leaflet/issues/2235) and [a recent one from last year](https://github.com/Leaflet/Leaflet/issues/6214).
This results in inconsistent when relying on setLatLng and the `move` event for leaflet classes.

- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
1. Listen to the `move` event on a CircleMarker and a Polygon
2. Use `setLatLng` and `setLatLngs` on the layers to move them
3. CircleMarker move event is fired. Polygon isn't.

**Expected behavior**
Either no `move` event is called. Or it's called in all setLatLng instances.

**Current behavior**
The `move` event is called inconsistently 

**Environment**
- Leaflet version: 1.4.0

johnd0e:
I believe that current behavior is consistent, as `move` meaning is completely clear for point-like features, but it is not defined for any other Path.
E.g. what when we just add one more point to polyline - is it `move`?

So I'd rather support `change` event for such cases.

Falke-Design:
I think `change` would be a good solution.

Should it also called for point layers? 

When should it fired:
- LatLng(s) changes
- Radius changes (CircleMarker & Circle)
- ImageOverlay `setBounds`

What should maybe also fire the event:
- Style change over `setStyle`
- Icon change `setIcon`
- `setOpacity`
- `bringToFront/Back()`?
- DivOverlay, Popup, Tooltip: `setContent` but this fires already `contentupdate`
- GridLayer: `setOpacity`, `setZIndex`
- TileLayer `setUrl`
- WMS: `setParams`
- ImageOverlay: `setOpacity`, `setUrl`, `setZIndex`

johnd0e:
[...]

PR #7150, opened 2020-05-16
Add move event to setLatLngs and addLatLng

Falke-Design:
Adding the `move` event to the `setLatLngs` and `addLatLng` function.
Added because of inconsistent `move` event on `setLatLng`. For more information read #6492

Fixed Issues:
#2235
#6214 
#6492

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -167,7 +167,7 @@ export var Polyline = Path.extend({
 	// the polyline in case of a multi-polyline, but can be overridden by passing
 	// a specific ring as a LatLng array (that you can earlier access with [`getLatLngs`](#polyline-getlatlngs)).
 	addLatLng: function (latlng, latlngs) {
-		var oldLatLngs = this._latlngs;
+		var oldLatLngs = this._convertLatLngs(this._latlngs);
 		latlngs = latlngs || this._defaultShape();
 		latlng = toLatLng(latlng);
 		latlngs.push(latlng);
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -74,7 +74,9 @@ export var Polyline = Path.extend({
 	// @method setLatLngs(latlngs: LatLng[]): this
 	// Replaces all the points in the polyline with the given array of geographical points.
 	setLatLngs: function (latlngs) {
+		var oldLatLngs = this._latlngs;
 		this._setLatLngs(latlngs);
+		this.fire('move', {oldLatLngs: oldLatLngs, latlngs: this._latlngs});
 		return this.redraw();
 	},
 
@@ -165,10 +167,12 @@ export var Polyline = Path.extend({
 	// the polyline in case of a multi-polyline, but can be overridden by passing
 	// a specific ring as a LatLng array (that you can earlier access with [`getLatLngs`](#polyline-getlatlngs)).
 	addLatLng: function (latlng, latlngs) {
+		var oldLatLngs = this._latlngs;
 		latlngs = latlngs || this._defaultShape();
 		latlng = toLatLng(latlng);
 		latlngs.push(latlng);
 		this._bounds.extend(latlng);
+		this.fire('move', {oldLatLngs: oldLatLngs, latlngs: this._latlngs});
 		return this.redraw();
 	},
 
--- a/debug/tests/move_event.html
+++ b/debug/tests/move_event.html
@@ -1,0 +1,80 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script src="../leaflet-include.js"></script>
+	<script src="../../dist/leaflet-src.js"></script>
+</head>
+<body>
+
+	<div id="map"></div>
+
+	<script>
+
+		var osmUrl = 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = L.tileLayer(osmUrl, {attribution: osmAttrib});
+
+		var map = L.map('map')
+				.setView([51.505, -0.09], 12)
+				.addLayer(osm);
+
+		var marker = L.marker([51.5, -0.09]).addTo(map);
+
+		var circle = L.circle([51.508, -0.11], 500).addTo(map);
+
+		var circleMarker = L.circleMarker([51.518, -0.05], 10).addTo(map);
+
+		var polygon = L.polygon([
+			[51.509, -0.08],
+			[51.503, -0.06],
+			[51.51, -0.047]
+		]).addTo(map);
+
+		var polyline = L.polyline([
+			[51.5095, -0.085],
+			[51.5035, -0.075]
+		]).addTo(map);
+
+		var rectangle = L.rectangle([
+			[51.519, -0.18],
+			[51.513, -0.16],
+		]).addTo(map);
+
+
+		function logEvent(e) { console.log(e);
+
+		marker.on('move', logEvent);
+		circle.on('move', logEvent);
+		circleMarker.on('move', logEvent);
+		polygon.on('move', logEvent);
+		polyline.on('move', logEvent);
+		rectangle.on('move', logEvent);
+
+		console.log('marker');
+		marker.setLatLng(marker.getLatLng());
+		console.log('circle');
+		circle.setLatLng(circle.getLatLng());
+		console.log('circleMarker');
+		circleMarker.setLatLng(circleMarker.getLatLng());
+		console.log('polyline setLatLngs');
+		polyline.setLatLngs(polyline.getLatLngs());
+		console.log('polyline addLatLng');
+		polyline.addLatLng([51.5035, -0.077]);
+		console.log('polygon');
+		polygon.setLatLngs(polygon.getLatLngs());
+		console.log('rectangle setLatLngs');
+		rectangle.setLatLngs(rectangle.getLatLngs());
+		console.log('rectangle setBounds');
+		rectangle.setBounds(rectangle.getBounds());
+
+	</script>
+</body>
+</html>


Falke-Design:
@mourner What is needed that this will be merged? The issue #6492 describes good that today the event / function of `move` is incositend

johnd0e:
See https://github.com/Leaflet/Leaflet/issues/6492#issuecomment-813419600

And it's not clear why not consider Circle radius change as `move` too. 

Besides, this PR lacks:
- events documentation
- events tests

Falke-Design:
@johnd0e PR is ready for review

src/layer/DivOverlay.js
```diff
@@ -79,12 +79,16 @@ export var DivOverlay = Layer.extend({
...
+		var oldLatLng = this._latlng;
 		this._latlng = toLatLng(latlng);
 		if (this._map) {
 			this._updatePosition();
 			this._adjustPan();
 		}
-		return this;
+
+		// @event move: Event
+		// Fired when the divoverlay is moved via [`setLatLng`](#divoverlay-setlatlng). Old and new coordinates are included in event arguments as `oldLatLng`, `latlng`.
```
johnd0e, line 90: Camel case for object (divOverlay, etc).
And better to wrap long line.

src/layer/vector/Circle.js
```diff
@@ -43,8 +43,13 @@ export var Circle = CircleMarker.extend({
...
 	// Sets the radius of a circle. Units are in meters.
 	setRadius: function (radius) {
+		var oldRadius = this._mRadius;
 		this._mRadius = radius;
-		return this.redraw();
+		this.redraw();
+
+		// @event move: Event
+		// Fired when the circle radius is changed via [`setRadius`](#circle-setradius). Old and new radius are included in event arguments as `oldRadius`, `radius`.
+		return this.fire('move', {oldRadius: oldRadius, radius: this._mRadius});
```
johnd0e, line 52: For sake of clarity it may be better to duplicate `setLatLngs` here too.
Not sure though.

johnd0e: Also, this may be a (a bit of) breaking change, as some code may expect latlng properties to always be included..

src/layer/vector/CircleMarker.js
```diff
@@ -51,8 +51,13 @@ export var CircleMarker = Path.extend({
...
 	// Sets the radius of a circle marker. Units are in pixels.
 	setRadius: function (radius) {
+		var oldRadius = this._radius;
 		this.options.radius = this._radius = radius;
-		return this.redraw();
+		this.redraw();
+
+		// @event move: Event
+		// Fired when the circle marker radius is changed via [`setRadius`](#circlemarker-setradius). Old and new radius are included in event arguments as `oldRadius`, `radius`.
+		return this.fire('move', {oldRadius: oldRadius, radius: this._radius});
```
johnd0e, line 60: Objection. For circle marker radius is style property, and no move events needed.

src/layer/vector/Polyline.js
```diff
@@ -165,11 +170,16 @@ export var Polyline = Path.extend({
...
 		latlngs = latlngs || this._defaultShape();
 		latlng = toLatLng(latlng);
 		latlngs.push(latlng);
 		this._bounds.extend(latlng);
-		return this.redraw();
+		this.redraw();
+
+		// @event move: Event
+		// Fired when the polyline is moved via [`addLatLng`](#polyline-addlatlng). Old and new coordinates are included in event arguments as `oldLatLngs`, `latlngs`.
+		return this.fire('move', {oldLatLngs: oldLatLngs, latlngs: this._latlngs});
```
johnd0e, line 182: We may need to add more properties in this case - function arguments.

johnd0e (changes requested):

src/layer/vector/Circle.js
Falke-Design: It is already different between marker / polyline because of `latlng` and `latlngs`, so I think it needs to be handled for each shape different, then the user can do by him self `getLatLng()`. But I added the property `latlng` (it doesn't hurt if it is there too)

src/layer/vector/CircleMarker.js
Falke-Design: The "problem" is that `setRadius` changes the shape and that would be good to catch. Every draw / edit plugin has troubles to update the handlers when a change is made in the code. 
Looks like that (same for radius of circle markers):


> For circle marker radius is style property

`setStyle({radius: XYZ})` calls `setRadius()` so it will fired too if the radius is changed over styles -> Which is good

src/layer/vector/Polyline.js
```diff
@@ -165,11 +171,17 @@ export var Polyline = Path.extend({
 	// the polyline in case of a multi-polyline, but can be overridden by passing
 	// a specific ring as a LatLng array (that you can earlier access with [`getLatLngs`](#polyline-getlatlngs)).
 	addLatLng: function (latlng, latlngs) {
```
Falke-Design, line 173: Suggestion to rename the second argument to `ring` because of the doc `passing a specific ring`

src/layer/vector/Polyline.js
```diff
@@ -165,11 +171,17 @@ export var Polyline = Path.extend({
...
+		var oldLatLngs = this._convertLatLngs(this._latlngs);
```
Falke-Design, line 174: This is not good for performance, because it loops through all to copy it (and break reference). Is there a better solution?

src/layer/vector/CircleMarker.js
johnd0e: > `setRadius` changes the shape and that would be good to catch

May be good, may be not.
Someone else might want to catch change in color.
I have no objections, except it is not `move` in this case.

Falke-Design: Would you say it is `move` for a circle? What do you think about `radiuschange`?

src/layer/vector/Polyline.js
johnd0e: `ring` itself is `latlngs` array

src/layer/vector/Polyline.js
johnd0e: The purpose of `_convertLanlngs` is validating user-provided data.
We just do not need it here, I suppose.

src/layer/vector/Polyline.js
Falke-Design: Yes I know, but every time I read `addLatLng: function (latlng, latlngs)` I think first that I need to pass the complete latlngs of the polyline. It would be much clearer if `latlngs` is renamed to `ring`.

src/layer/vector/CircleMarker.js
johnd0e: > Would you say it is `move` for a circle?

Yes, for circle it's radius defines it's coverage.
No need for `radiuschange`.

src/layer/vector/Polyline.js
Falke-Design: I used it to break the reference. If I use `this._latlngs` for `oldLatLngs` the added LatLng is added too, because it has the same reference.

src/layer/vector/Polyline.js
johnd0e: > It would be much clearer if `latlngs` is renamed to `ring`.

I have no opinion.
@mourner @IvanSanchez @jonkoops

src/layer/vector/Polyline.js
johnd0e: > I used it to break the reference.

Make separate light-weight function, based on `L.extend`.

src/layer/vector/CircleMarker.js
Falke-Design: I don't understand why `move` is ok for Circle but not for CircleMarker? They are "same" only the projection and fixed size is different.
According to the docs:
CircleMarker
> A circle of a fixed size with radius specified in pixels

Circle:
> A class for drawing circle overlays on a map

Can you please explain what is the difference for you

johnd0e: CircleMarker is just point on map. It's radius is style feature.
Circle is shape, it's radius defines it's bounds.

src/layer/vector/Polyline.js
jonkoops: Seems fine to me

src/layer/vector/CircleMarker.js
Falke-Design: @mourner what do you think? For plugins it can be very useful to catch when the radius and latlng is changed especially if circle and circlemarkers are handled equal, like [here](https://jsfiddle.net/falkedesign/ygefu7s9/)

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,24 @@ describe('LineUtil', function () {
...
 	});
+
+	describe('#cloneLatLngs', function () {
+		it('clone a flat latlng array', function () {
+			var latLngs = [
+				L.latLng([1, 2]),
+				L.latLng([3, 4])
+			];
+			var latLngsClone = L.LineUtil.cloneLatLngs(latLngs);
+			expect(latLngsClone).to.not.be(latLngs);
```
johnd0e, line 94:
```
			expect(latLngsClone).to.eql(latLngs);
```

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,24 @@ describe('LineUtil', function () {
...
+		});
+
+		it('clone a latlng array', function () {
+			var latLngs = [[
+				L.latLng([1, 2]),
+				L.latLng([3, 4])
+			]];
+			var latLngsClone = L.LineUtil.cloneLatLngs(latLngs);
+			expect(latLngsClone).to.not.be(latLngs);
```
johnd0e, line 103:
```
			expect(latLngsClone).to.eql(latLngs);
```

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,19 @@ export function _flat(latlngs) {
 	console.warn('Deprecated use of _flat, please use L.LineUtil.isFlat instead.');
 	return isFlat(latlngs);
 }
+
+// @function cloneLatLngs(latlngs: LatLng[]): LatLng[]
+// Recursively clone latlngs to a new array and breaks the reference to the original one.
+export function cloneLatLngs(latlngs) {
```
johnd0e, line 247: Are you sure that we need to make this function public?
@mourner 
@IvanSanchez

johnd0e (changes requested):

src/layer/vector/Polyline.js
```diff
@@ -164,12 +170,18 @@ export var Polyline = Path.extend({
 	// Adds a given point to the polyline. By default, adds to the first ring of
 	// the polyline in case of a multi-polyline, but can be overridden by passing
 	// a specific ring as a LatLng array (that you can earlier access with [`getLatLngs`](#polyline-getlatlngs)).
-	addLatLng: function (latlng, latlngs) {
-		latlngs = latlngs || this._defaultShape();
+	addLatLng: function (latlng, ring) {
+		var oldLatLngs = LineUtil.cloneLatLngs(this._latlngs);
```
mourner, line 174: Cloning the whole array on every `addLatLng` is potentially a significant performance problem. Basically, any time you build out a route point by point, there will be extraordinary amount of redundant memory allocations and GC churn. 

I'm not sure what's the best approach here but maybe it's just not worth including `oldLatLngs` in the event in principle, and avoid any cloning.

mourner (changes requested):

src/layer/vector/Polyline.js
Falke-Design: Is ok for me, we can add it later if it is needed. I will update the PR today

Falke-Design:
Todo: Revert of removing `oldLatLng` from marker and maybe add it to ImageOverlay too. But for Polylines the old LatLngs should not return

src/geometry/LineUtil.js
mourner: It's no longer used as per review below, so let's remove it for now.

Falke-Design: Ok sad, because I would like to have it the Utils function. Already needed it in some projects. But before this blocks this longer I will remove it

Falke-Design:
@mourner is the `updateradius` event ok for you? And the `move` event has now 3 different types `LatLngMoveEvent`, `LatLngsMoveEvent`, `BoundsMoveEvent`

src/geometry/LineUtil.js
mourner: Let's readd it in another PR then!

Issue #6764, opened 2019-08-11
tooltipAnchor has limited or no effect for some Tooltip directions


Istador:
- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
> The coordinates of the point from which tooltips will "open", relative to the icon anchor.
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
- Define an Icon with the `tooltipAnchor` option with a non-zero value for x.
- Create a Marker using that Icon, that has a Tooltip bound with an `top`, `bottom` or `center` direction.

**Expected behavior**
The anchor of the Tooltip should move relative to the `iconAnchor` according to the values defined in the `tooltipAnchor`. It should have the same behavior as `popupAnchor` according to the documentation, e.g. if given the same values it should position itself at the same position as the Popup.

**Current behavior**
`tooltipAnchor.x` has no effect for the directions `top`, `bottom` and `center`.
`tooltipAnchor.y` has no effect for the `bottom` direction.

**Environment**
- Leaflet version: 1.5.1
- Browser (with version): Chrome 76.0.3809.100, Waterfox 56.2.12
- OS/Platform (with version): Windows 7 (6.1 Build 7601), Ubuntu 18.04

**Additional context**
Source code responsible for the issue:
https://github.com/Leaflet/Leaflet/blob/1e0d43b11d477e0c89de4dace33123aa871c76b8/src/layer/Tooltip.js#L144-L156

**Minimal example reproducing the issue**
https://jsfiddle.net/590jt8fx/
```js
const map = L.map('map').setView([0.0, 0.0], 12)

const icon = L.icon({
  iconUrl: 'https://leafletjs.com/docs/images/logo.png',
  iconSize: [120, 31.8], // (600 x 159)/5
  iconAnchor: [60, 15.9], // center
  popupAnchor: [50, -15.9], // top right
  tooltipAnchor: [50, -15.9], // top right
})

L.marker([0, 0], { icon } )
  .bindPopup("popup is working correctly")
  .bindTooltip("tooltipAnchor.x has no effect", { direction: 'top' })
  .addTo(map)
  
L.marker([-0.02, 0], { icon } )
  .bindTooltip("offset is needed", { direction: 'top', offset: [50, 0] })
  .addTo(map)
```

[...]

Issue #5101, opened 2016-11-16
Leaflet tooltip does not use correct direction if set to 'auto' on repeat mouse over


skalb:
## How to reproduce

- Leaflet version I'm using: development
- Browser (with version) I'm using: Chrome 54
- OS/Platform (with version) I'm using: Mac OSX 10.11

1. Mouse over marker (tooltip shows with correct arrow)
2. Pan map so auto direction changes to left from right, or vice versa
3. Mouse over marker

## What behaviour I'm expecting and which behaviour I'm seeing

I'd expect the arrow to be on the correct side of the tooltip (the side closer to the marker), but it appears on the opposite side. See screenshot.
![screen shot 2016-11-15 at 4 17 58 pm](https://cloud.githubusercontent.com/assets/262745/20329777/29758ef4-ab4f-11e6-82aa-797ccd6b8a1a.png)

If you add an offset, it's also calculated incorrectly.

If you mouse out and mouse over the marker it seems to fix itself.

## Minimal example reproducing the issue

https://playground-leaflet.rhcloud.com/bol/edit?html,output

perliedman:
Interesting, thanks for reporting, it looks like some update is done out of sync, since the problem corrects itself as you say, when the tooltip is shown a second time.

Istador:
I can't reproduce 4.5 years later with any of these versions (Chrome 89 on Windows):
- 1.0.0-rc.2 (first version that added `direction`)
- 1.0.0
- 1.6.0
- 1.7.0
- 1.7.1

Is this issue still present?
If so, please add a new minimal example (domain doesn't exist anymore).

@johnd0e

PR #7155, opened 2020-05-21
fix tooltipAnchor behavior for different directions

Istador:
This commit fixes issue #6764, see the discussion and examples there.
It's an alternative implementation for pull request #6116.

It makes `tooltipAnchor` behave the same like `popupAnchor` regardless of the tooltip's direction. With one exception: the `auto` direction flips the `x` axis of `tooltipAnchor` and the tooltip's `offset` when it switches to `left` depending on the position on the screen. `auto` therefore assumes that the icon is somewhat symmetrical to the left and the right of the y axis, which is the case for the default icon.

### Example with this pull request applied

https://jsfiddle.net/emwbsr9j/

#### `auto` to the right
NEW: auto to the right

#### `auto` to the left
NEW: auto to the left

### Old behavior

https://jsfiddle.net/5nycvLez/

#### `auto` to the right
OLD: auto to the right

#### `auto` to the left
OLD: auto to the left

### Example Code

```js
const map = L.map('map').setView([0.0, 0.0], 12)

const iconBase = {
  iconUrl: 'https://leafletjs.com/docs/images/logo.png',
  iconSize: [120, 31.8], // (600 x 159)/5
  iconAnchor: [60, 15.9], // center
}

const icons = {
  top: L.icon(Object.assign({
    popupAnchor   : [50, -15.9], // top right
    tooltipAnchor : [50, -15.9], // top right
  }, iconBase)),
  bottom: L.icon(Object.assign({
    popupAnchor   : [-3, 15.9], // bottom center
    tooltipAnchor : [-3, 15.9], // bottom center
  }, iconBase)),
  left: L.icon(Object.assign({
    popupAnchor   : [-60, 8], // bottom left
    tooltipAnchor : [-60, 8], // bottom left
  }, iconBase)),
  right: L.icon(Object.assign({
    popupAnchor   : [60, 0], // right
    tooltipAnchor : [60, 0], // right
  }, iconBase)),
  center: L.icon(Object.assign({
    popupAnchor   : [50, -15.9], // top right
    tooltipAnchor : [50, -15.9], // top right
  }, iconBase)),
  auto: L.icon(Object.assign({
    popupAnchor   : [60, 0], // right
    tooltipAnchor : [60, 0], // right
  }, iconBase)),
}

let offset = 0.06
L.marker([offset, 0])
  .bindPopup("default")
  .bindTooltip("default")
  .addTo(map)
  .openTooltip()
for (const k in icons) {
  L.marker([offset -= 0.02, 0], { icon: icons[k] } )
    .bindPopup("popup " + k)
    .bindTooltip("tooltip " + k, { direction: k })
    .addTo(map)
    .openTooltip()
}
```

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -131,7 +131,8 @@ export var Tooltip = DivOverlay.extend({
 	_adjustPan: function () {},
 
 	_setPosition: function (pos) {
-		var map = this._map,
+		var subX, subY,
+		    map = this._map,
 		    container = this._container,
 		    centerPoint = map.latLngToContainerPoint(map.getCenter()),
 		    tooltipPoint = map.layerPointToContainerPoint(pos),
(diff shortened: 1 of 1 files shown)


mourner:
Great, thank you!

LeOSW42:
Just upgraded to 1.7.1 and my marker's tooltips have moved.

I cannot figure out why the "top" position is now top right by default: https://jsfiddle.net/2kw3bdgs/1/

Istador:
The tooltip opens up at the `tooltipAnchor` position, unaffected by the direction (with an exception for `auto`).

The default marker has its `tooltipAnchor` right of the upper circle:
https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/layer/marker/Icon.Default.js#L29

From the docs:
> Note about tooltip offset. Leaflet takes two options in consideration for computing tooltip offsetting:
>
> - the `offset` Tooltip option: it defaults to [0, 0], and it's specific to one tooltip. Add a positive x offset to move the tooltip to the right, and a positive y offset to move it to the bottom. Negatives will move to the left and top.
> - the `tooltipAnchor` Icon option: this will only be considered for Marker. You should adapt this value if you use a custom icon.

So, you can change the point where you want the tooltip to be positioned by these two options. Once generally for all markers of an icon via `tooltipAnchor` or for specific tooltips individually via `offset`.

Before this PR, the `direction` to which the tooltips open was mixed up with the positioning/offsetting of the anchor point, making it impossible or harder to change the position with some of the `direction` values (it ignored one or both axes of `tooltipAnchor` completely and - judging by looking at the [old code](https://github.com/Leaflet/Leaflet/blob/1e0d43b11d477e0c89de4dace33123aa871c76b8/src/layer/Tooltip.js#L144-L156) now - I think it inverted both `offset` axes when using the `center` direction).

johnd0e:
@Istador 
Could you check #5101?

PR #7170, opened 2020-06-09
Update Events.js

donvan:
context


johnd0e:
https://leafletjs.com/reference-1.6.0.html#evented-off
https://github.com/Leaflet/Leaflet/blob/25f7bbb93d73ecbca7ae47de5e86009e7568576e/src/core/Events.js#L62-L64

If fact this syntax already supports `context`, so docs should be updated:
```
@method off(eventMap: Object, context?: Object): this
```

PR #7194, opened 2020-07-02
Docs of LatLngBounds.overlaps are wrong

### LINKED ISSUE #7190: Docs of LatLngBounds.overlaps are wrong
The docs say, that the first and only parameter of `LatLngBounds.overlaps` is a Object of type `Bounds` ([see here](https://leafletjs.com/reference-1.6.0.html#latlngbounds-overlaps)) although the code expects a `LatLngBounds` Object ([see here](https://github.com/Leaflet/Leaflet/blob/e8c28b996179373447dde1df479898576a508579/src/geo/LatLngBounds.js#L201)), which makes a lot more sense. So this is - at least in my opinion - not a bug but wrong documentation.

**Steps to reproduce**
- Create `LatLngBounds`-Object
- Create `Bounds`-Object
- Call `overlaps` of the `LatLngBounds`-Object with the `Bounds`-Object as parameter

**Expected behavior**
No error (at least according to the docs)

**Current behavior**
- `LatLngBounds.overlaps` fails when called with a `Bounds` Object:
   img
- `LatLngBounds.overlaps` works fine when called with another `LatLngBounds`-Object as the parameter

**Environment**
- Leaflet version: 1.60
- Browser (with version): Chrome Desktop Version 83.0.4103.116
- OS/Platform (with version): Windows 10 Version 2004 (Build 19041.330)

```js
const latLngBounds = map.getBounds();
const bounds = map.getPixelBounds();

console.log(latLngBounds.overlaps(bounds))
```
[Example Link](https://plnkr.co/edit/hbeBKMzayRscrLBi)

DerZade:
This PR will fix #7190

### DIFF
--- a/src/geo/LatLngBounds.js
+++ b/src/geo/LatLngBounds.js
@@ -195,7 +195,7 @@ LatLngBounds.prototype = {
 		return latIntersects && lngIntersects;
 	},
 
-	// @method overlaps(otherBounds: Bounds): Boolean
+	// @method overlaps(otherBounds: LatLngBounds): Boolean
 	// Returns `true` if the rectangle overlaps the given bounds. Two bounds overlap if their intersection is an area.
 	overlaps: function (bounds) {
 		bounds = toLatLngBounds(bounds);


IvanSanchez:
Thanks!

DerZade:
Always happy to help :)

PR #7197, opened 2020-07-07
Move worldCopyJump to L.Map, make it work when zooming.

IvanSanchez:
As prompted by https://stackoverflow.com/questions/62764699/can-leaflet-jump-to-new-world-copy-on-zoom-instead-of-just-on-panning . 

I have tested it with mouse only - at the moment I do lack a tablet (or any other big multipointer input device). I would appreciate a bit of manual testing of this:

- Check out the branch
- Do the build (`npm run rollup`)
- Manually open `debug/tests/dragging_and_copyworldjump.html` in a browser
- Do a bunch of pinch-zoom operations (including one-finger dragging before and after the second finger for the pinch is down)

### DIFF
--- a/debug/tests/dragging_and_copyworldjump.html
+++ b/debug/tests/dragging_and_copyworldjump.html
@@ -19,8 +19,8 @@
 		<button id="foo">
 			Click to enable dragging on the right map, then dragging around and watch copying
 		</button><br>
-		<div id="map1" style="height: 300px;width: 400px; float:left;"></div>
-		<div id="map2" style="height: 300px;width: 400px; float:left; margin-left: 10px;"></div>
+		<div id="map1" style="height: 300px;width: 40vw; float:left;"></div>
+		<div id="map2" style="height: 300px;width: 40vw; float:left; margin-left: 10px;"></div>
 		<div style="clear:both"></div>
 
 
--- a/debug/leaflet-src.js.map
+++ b/debug/leaflet-src.js.map
@@ -1,0 +1,1 @@
+../dist/leaflet-src.js.map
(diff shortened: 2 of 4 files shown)


joshiraez:
Hi, I have got a tablet and I think I can take on this to get started with leaflet. Ill be working on this if its okay.

joshiraez:
Tests results, according to the original issue where markers where dissapearing.

All tests didn't remove any of the markers in any of the maps. Did it with only one map unlocked first, then with both.

Dragging with only one touch - OK
Dragging, and then zooming pinching with the other finger - OK
Dragging, and then zooming out expanding with the other finger - OK
Zooming in - OK
Zooming out - OK

I always reverted the operation just to make sure that the marker was there after returning back to screen, so it seems like it's being recalculated correctly now. Is ok to merge

joshiraez:
I created a set of tests that checks the transition between zooming and dragging

I created a PR to this branch with the 4 new tests, with the 2 that are expected to fail after the manual testing failing.

https://github.com/Leaflet/Leaflet/pull/7226

misfitonie:
Hi, if this ticket is still open, i'll be happy to work on it.

ardada2468 (approved):
This should fix the issue and make it zoom

ardada2468 (approved):

debug/tests/dragging_and_copyworldjump.html
```diff
@@ -19,8 +19,8 @@
 		<button id="foo">
 			Click to enable dragging on the right map, then dragging around and watch copying
 		</button><br>
-		<div id="map1" style="height: 300px;width: 400px; float:left;"></div>
-		<div id="map2" style="height: 300px;width: 400px; float:left; margin-left: 10px;"></div>
+		<div id="map1" style="height: 300px;width: 40vw; float:left;"></div>
+		<div id="map2" style="height: 300px;width: 40vw; float:left; margin-left: 10px;"></div>
```
Falke-Design, line 23: Please revert this changes

Falke-Design (changes requested):
Please remove the file `debug/leaflet-src.js.map`

PR #7203, opened 2020-07-08
Inline default icons as data:svg+xml URLs, remove def. icon path detection

### LINKED ISSUE #7202: Default marker icon doesn't load (because of race condition?)
Lately I've been hitting a [heisenbug](https://en.wikipedia.org/wiki/Heisenbug) regarding the dreaded autodetection of the path for  `L.Icon.Default` (see #7092, #6496, and others).

In order to reproduce:
- Clear your browser cache (or go into private mode)
- Create a map with (at least) one default `L.Marker`
- Load the map and ***wish*** that your browser evaluates the Leaflet JS code *before* it evaluates the CSS code

One out of every 20 times, I get something like

imagen

Refreshing the page (with the Leaflet JS/CSS code cached) will make the default marker appear (with the right L.Icon.Default URL):

imagen

I'm able to reproduce this in a fairly consistent manner using Firefox 77 (on Debian), private mode, network console open, cache disabled.

IvanSanchez:
A proposal to fix #7202 and end all the woes associated with the default icon when the Leaflet code is bundled. This inlines the SVG icon for the marker, and a SVG icon I made for the shadow, as `data:svg+xml` URIs.

Feels incomplete because the image for the layers control still is done via CSS backgrounds, which has similar woes.

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -1,6 +1,13 @@
 import {Icon} from './Icon';
 import * as DomUtil from '../../dom/DomUtil';
 
+// The contents of src/image/marker-clean.svg, minus spaces, replacing '#' with '%23'.
+var svgIcon = '<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="25" height="41"><defs><linearGradient id="c" gradientUnits="userSpaceOnUse" x1="0" y1="38.71" x2="0" y2="1.144"><stop offset="0" stop-color="%23126fc6"/><stop offset="1" stop-color="%234c9cd1"/></linearGradient><linearGradient id="d" gradientUnits="userSpaceOnUse" x1="0" y1="20.197" x2="0" y2="1.144"><stop offset="0" stop-color="%232e6c97"/><stop offset="1" stop-color="%233883b7"/></linearGradient></defs><path fill="%23fff" d="M6.329 5.286h12.625v14.5H6.329z"/><path d="M12.594 1.323c-6.573 0-12.044 5.691-12.044 11.866 0 2.778 1.564 6.308 2.694 8.746l9.306 17.872 9.262-17.872c1.13-2.438 2.738-5.791 2.738-8.746 0-6.175-5.383-11.866-11.956-11.866zm0 7.155c2.584.017 4.679 2.122 4.679 4.71s-2.095 4.663-4.679 4.679c-2.584-.017-4.679-2.09-4.679-4.679 0-2.588 2.095-4.693 4.679-4.71z" fill="url(%23c)" stroke="url(%23d)" stroke-width="1.1" stroke-linecap="round"/><path d="M12.581 2.43c-5.944 0-10.938 5.219-10.938 10.75 0 2.359 1.443 5.832 2.563 8.25l.031.031 8.313 15.969 8.25-15.969.031-.031c1.135-2.448 2.625-5.706 2.625-8.25 0-5.538-4.931-10.75-10.875-10.75zm0 4.969c3.168.021 5.781 2.601 5.781 5.781 0 3.18-2.613 5.761-5.781 5.781-3.168-.02-5.75-2.61-5.75-5.781 0-3.172 2.582-5.761 5.75-5.781z" fill="none" stroke="%23fff" stroke-width="1.1" stroke-linecap="round" stroke-opacity=".122"/></svg>';
+
+// The contents of src/image/shadow-clean.svg, minus spaces, replacing '#' with '%23'.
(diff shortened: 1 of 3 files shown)


johnd0e:
> A proposal to fix #7202 and end all the woes associated with the default icon when the Leaflet code is bundled.

Haven't checked the code, but here are some general thoughts:
- inlining standard icon is not good idea, because each of 9999 markers have to use own copy (despite that all copies are equivalent)
- when dealing with `svg`, it's common to avoid redundancy with [`use`] element.
- I'm not sure that we should get rid of `imagePath` completely.

[`use`]: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use

src/layer/marker/Icon.Default.js
```diff
@@ -1,6 +1,13 @@
...
..."1.144"><stop offset="0" stop-color="%23126fc6"/><stop offset="1" stop-color="%234c9cd1"/></linearGradient><linearGradient id="d" gradientUnits="userSpaceOnUse" x1="0" y1="20.197" x2="0" y2="1.144"><stop offset="0" stop-color="%232e6c97"/><stop offset="1" stop-color="%233883b7"/></linearGradient></defs><path fill="%23fff" d="M6.329 5.286h12.625v14.5H6.329z"/><path d="M12.594 1.323c-6.573 0-12.044 5.691-12.044 11.866 0 2.778 1.564 6.308 2.694 8.746l9.306 17.872 9.262-17.872c1.13-2.438 2.738-5.791 2.738-8.746 0-6.175-5.383-11.866-11.956-11.866zm0 7.155c2.584.017 4.679 2.122 4.679 4.71s-2.095 4.663-4.679 4.679c-2.584-.017-4.679-2.09-4.679-4.679 0-2.588 2.095-4.693 4.679-4.71z" fill="url(%23c)" stroke="url(%23d)" stroke-width="1.1" stroke-linecap="round"/><path d="M12.581 2.43c-5.944 0-10.938 5.219-10.938 10.75 0 2.359 1.443 5.832 2.563 8.25l.031.031 8.313 15.969 8.25-15.969.031-.031c1.135-2.448 2.625-5.706 2.625-8.25 0-5.538-4.931-10.75-10.875-10.75zm0 4.969c3.168.021 5.781 2.601 5.781 5.781 0 3.18-2.613 5.761-5.781 5.781-3.168-.02-5.75-2.61-5.75-5.781 0-3.172 2.582-5.761 5.75-5.781z" fill="none" stroke="%23fff" stroke-width="1.1" stroke-linecap="round" stroke-opacity=".122"/></svg>';
```
johnd0e, line 5: Should be url-encoded for better compatibility.

More info: https://css-tricks.com/probably-dont-base64-svg/

IvanSanchez: If by "url-encode" you mean "replace `#` with `%23`", then that's already done.

IvanSanchez:
> * inlining standard icon is not good idea, because each of 9999 markers have to use own copy (despite that all copies are equivalent)

My thought about that is to use `blob:` URLs (when supported), using the trick from https://gitlab.com/IvanSanchez/rollup-plugin-file-as-blob/blob/master/src/index.js#L15

> * when dealing with `svg`, it's common to avoid redundancy with [`use`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use) element.

The SVG icons have undergone several `svgo` passes and manual edits to check that everything is as simple as it can be, so I'm unsure what you mean.

> * I'm not sure that we should get rid of `imagePath` completely.

I'm ambivalent about that right now :-|

mourner:
I think that's a really good idea! And by this we'd be officially dropping IE8 support (which no one checks anyway), so we could remove things like https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/SVG.VML.js too :)

One option to consider here is instead of inling SVG in JS, do all of the icons (controls, markers, etc.) with CSS backgrounds inlining SVG.

src/layer/marker/Icon.Default.js
johnd0e: Depending on browser it also may require to escape `<`, `>` (and may be more, I'm not sure)

johnd0e:
> The SVG icons have undergone several `svgo` passes and manual edits to check that everything is as simple as it can be, so I'm unsure what you mean

I mean that in your current implementation every new marker use new copy of icon.
Instead, it's possible to define svg symbol once, and reuse it everywhere.

> My thought about that is to use `blob:` URLs

I agree, this approach is also good.
But it is [IE10+](https://caniuse.com/#feat=bloburls).

> And by this we'd be officially dropping IE8 support (which no one checks anyway),

Well, when we rise requirements up to IE10, we can drop PNG too. And we can use the same SVG for retina too.
And one more advantage - SVGs are styleable!

PR #7204, opened 2020-07-08
Rely on CSS backgrounds for the default icon images

IvanSanchez:
Another approach for #7202, taking inspiration in how the image for the layers control is loaded only via CSS.

This has the advantage that smartypants bundlers (webpack and the like) might be able to pick references to external assets from the `background-image: url(foo)` CSS rules and automagically copy them around.

This was a quick test, so I haven't given much thought on the compatibility issues of (now missing) `L.Icon.Default.imagePath` :confused:

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -1,6 +1,8 @@
 import {Icon} from './Icon';
 import * as DomUtil from '../../dom/DomUtil';
 
+var emptyImageUrl = 'data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==';
+
 /*
  * @miniclass Icon.Default (Icon)
  * @aka L.Icon.Default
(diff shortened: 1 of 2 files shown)


src/layer/marker/Icon.Default.js
```diff
@@ -20,41 +22,14 @@ import * as DomUtil from '../../dom/DomUtil';
 export var IconDefault = Icon.extend({
 
 	options: {
-		iconUrl:       'marker-icon.png',
-		iconRetinaUrl: 'marker-icon-2x.png',
-		shadowUrl:     'marker-shadow.png',
+		iconUrl:     emptyImageUrl,
+		shadowUrl:   emptyImageUrl,
```
johnd0e, line 26: Why not DivIcon?

src/layer/marker/Icon.Default.js
```diff
@@ -20,41 +22,14 @@ import * as DomUtil from '../../dom/DomUtil';
...
-		if (!IconDefault.imagePath) {	// Deprecated, backwards-compatibility only
-			IconDefault.imagePath = this._detectIconPath();
-		}
-
-		// @option imagePath: String
-		// `Icon.Default` will try to auto-detect the location of the
-		// blue icon images. If you are placing these images in a non-standard
-		// way, set this option to point to the right path.
-		return (this.options.imagePath || IconDefault.imagePath) + Icon.prototype._getIconUrl.call(this, name);
+		shadowSize:  [41, 41],
```
johnd0e, line 31: Why not in `options`?

src/layer/marker/Icon.Default.js
```diff
@@ -20,41 +22,14 @@ import * as DomUtil from '../../dom/DomUtil';
...
 		iconSize:    [25, 41],
 		iconAnchor:  [12, 41],
 		popupAnchor: [1, -34],
 		tooltipAnchor: [16, -28],
-		shadowSize:  [41, 41]
-	},
-
-	_getIconUrl: function (name) {
-		if (!IconDefault.imagePath) {	// Deprecated, backwards-compatibility only
-			IconDefault.imagePath = this._detectIconPath();
```
johnd0e, line 35: We should not remove that, otherwise we lose ability to specify just icon filenames.
This would break everything.

src/layer/marker/Icon.Default.js
IvanSanchez: No reason, really. I made this as a quick hack and didn't think about making it a `DivIcon`.

src/layer/marker/Icon.Default.js
IvanSanchez: I guess I could add this back, and put in some logic so that if `L.Icon.Default.imagePath` is specified, the CSS class is removed. Adding a `className` option was the quickest to implement, but it would be possible to a it conditionally before the code from `L.Icon` needs it.

PR #7216, opened 2020-07-21
add space to character regex for template

lubojr:
In order to allow substitution of Object keys containing spaces (can be external JSON data outside of control) to be used for templating of URLs


rheh (commented):
LGMT

rheh (approved):

IvanSanchez (approved):

PR #7226, opened 2020-07-28
Added tests for zooming after dragging, and dragging after zooming

joshiraez:
Added 4 tests to check the transition between zooming and dragging when using touch events.


johnd0e:
@joshiraez 
1. Tests are failed.
2. PR is created against wrong branch

Issue #7211, opened 2020-07-15
Make the disabled state of zoom controls available to ATs


Malvoz:
The `leaflet-disabled` class is set on zoom controls:
https://github.com/Leaflet/Leaflet/blob/71f73785c91f50821cee4f590c41b14eeddf394c/src/control/Control.Zoom.js#L102-L104
and styles are applied to convey a "disabled" state:
https://github.com/Leaflet/Leaflet/blob/71f73785c91f50821cee4f590c41b14eeddf394c/dist/leaflet.css#L310-L314

the zoom controls remain operable however, and the [state](https://www.w3.org/TR/WCAG21/#dfn-states) is not available to assistive technologies such as screen readers. There are at least 3 ways to convey the disabled state of buttons:

- `tabindex="-1"` - [Bing Maps](https://maps4html.org/HTML-Map-Element-UseCases-Requirements/examples/create-map.html#bing-maps-embed) for example, uses this approach. Though it may be confusing to keyboard-only users, as the button is then removed from the tab order.
- The [`disabled`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) attribute - Not applicable (Leaflet uses `<a role="button">` as opposed to `<button>`).
- [`aria-disabled="true"`](https://www.w3.org/TR/wai-aria/#aria-disabled) - May or may not be as well supported as a negative `tabindex` or the `disabled` attribute, it does seem to be [well supported](https://a11ysupport.io/tech/aria/aria-disabled_attribute#support-table-1) though.

Whether `tabindex="-1"` or `aria-disabled="true|false"` is used (I personally vote for the latter), it could be set _in addition_ to the `leaflet-disabled` class for backwards compatibility.

<hr>

Implementing this may satisfy (parts of) the WCAG 2.1 Success Criterion: [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships) and [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) for the zoom controls.

mourner:
@Malvoz great suggestion! Let's go with `aria-disabled` in addition to the current class. Would like to submit a PR for this?

Augani:
Hello @mourner , can i take this?

Malvoz:
@mourner Sorry for the delay! I'm happy to defer this to @Augani.

AkshataJ96:
If the issue is still available, can i start looking into this? @mourner

[...]

PR #7229, opened 2020-08-01
Made the zoom controls more accessible

Augani:
As mentioned in issue #7211 i made the controls more accessible by adding the aria-disabled attribute when it is disabled

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -105,12 +105,16 @@ export var Zoom = Control.extend({
 
 		DomUtil.removeClass(this._zoomInButton, className);
 		DomUtil.removeClass(this._zoomOutButton, className);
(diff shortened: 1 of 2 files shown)


jaeh:
i am not allowed to review code, but i read through the changes and thought about the implications.

this commits:
* add `DomUtil.setAttribute(element, attribute, value)` in src/dom/DomUtil.js
* seem to correctly set 'aria-disabled' to the Zoom control in src/control/Control.js 

i think it's clear that this code needs to add a DomUtil.setAttribute function,
which makes this a lot more impactful than obvious at first glance.

this means that 
* a `DomUtil.removeAttribute(element, attribute)` function should be added too
* the documentation has to be updated to include the DomUtil methods
* if other controls need aria attributes, those should be added as well.

@Augani you read into this a bit, are there more controls that need aria attributes?

Falke-Design:
Already fixed in #7280

PR #7232, opened 2020-08-10
Fix exception on FeatureGroup getBounds() with LayerGroup/FeatureGroup as child

### LINKED ISSUE #7228: getBounds() of FeatureGroup throw error when it has a LayerGroup as child
When a FeatureGroup has a LayerGroup as child and `getBounds()` is called, it throws the error **"e.getLatLng is not a function"** 

**Steps to reproduce**
```
var fg = L.featureGroup();
console.log(fg.getBounds());
var lg = L.layerGroup().addTo(fg)
console.log(fg.getBounds()); // throws error
```

**Current code**
```
getBounds: function() {
    var bounds = new LatLngBounds();

        var layer = this._layers[id];
        bounds.extend(layer.getBounds ? layer.getBounds() : layer.getLatLng());
    }
    return bounds;
}
```

**Expected behavior**
I think it should go through all layers of a LayerGroup. Like:
```
getBounds: function() {
    var bounds = new LatLngBounds();

        var layer = this._layers[id];
        if (layer instanceof L.LayerGroup && !layer.getBounds) {
            for (var lgid in layer._layers) {
                var lglayer = layer._layers[lgid];
                bounds.extend(lglayer.getBounds ? lglayer.getBounds() : lglayer.getLatLng());
            }
        } else {
            bounds.extend(layer.getBounds ? layer.getBounds() : layer.getLatLng());
        }
    }
    return bounds;
}
```
But now is the problem what if the LayerGroup contains also a LayerGroup?

The [Doc](https://leafletjs.com/reference-1.6.0.html#featuregroup-getbounds) says "Returns the LatLngBounds of the Feature Group (created from bounds and coordinates of its children)" this means for me, that it returns the bounds of **all** childrens also for LayerGroups and LayerGroups with LayerGroups childrens.

The easiest way is to add the `getBounds()` function to the LayerGroup, but I think this is not what you want.

rafaelgalani:
Added logic to handle groups (LayerGroup/FeatureGroup) as children of a FeatureGroup. It will also work with nested groups.
Unit tests were also added to check the new logic. There was no need to modify the documentation, since the current one already expected such behaviour.

This PR fixes #7228.

### DIFF
--- a/spec/suites/layer/FeatureGroupSpec.js
+++ b/spec/suites/layer/FeatureGroupSpec.js
@@ -84,7 +84,7 @@
 			expect(fg.hasLayer(marker)).to.be(false);
 		});
 	});
-	
+
 	describe('getBounds', function () {
 		it('returns the bounds (latlng) of the group', function () {
 			var fg = L.featureGroup([
@@ -96,12 +96,12 @@
 			]);
 
 			var southWest = new L.LatLng(0, 0),
-				northEast = new L.LatLng(4, 4);
-			
-			var bounds = new L.LatLngBounds(southWest, northEast);			
+			    northEast = new L.LatLng(4, 4);
+
+			var bounds = new L.LatLngBounds(southWest, northEast);
 			expect(fg.getBounds()).to.eql(bounds);
 		});
-		
+
 		it('returns the bounds (LatLng) of the group', function () {
 			var fg = L.featureGroup([
 				L.marker([0, 0]),
@@ -112,43 +112,43 @@
 			]);
 
 			var southWest = new L.LatLng(0, 0),
-				northEast = new L.LatLng(4, 4);
-			
-			var bounds = new L.LatLngBounds(southWest, northEast);			
+			    northEast = new L.LatLng(4, 4);
+
+			var bounds = new L.LatLngBounds(southWest, northEast);
 			expect(fg.getBounds()).to.eql(bounds);
 		});
 
-		describe('when a FeatureGroup contains a LayerGroup as children', function(){
-			it("returns the bounds (LatLng) of the group, including group member's child layers", function(){
-				var parentFeatureGroup = L.featureGroup([ L.marker([-23, -102]) ]);
+		describe('when a FeatureGroup contains a LayerGroup as children', function () {
+			it("returns the bounds (LatLng) of the group, including group member's child layers", function () {
+				var parentFeatureGroup = L.featureGroup([L.marker([-23, -102])]);
 
-				var layerGroupWithChildren = L.layerGroup([
+				L.layerGroup([
 					L.marker([39.61, -105.02]),
 					L.marker([39.74, -104.99]),
 					L.marker([39.73, -104.8]),
 					L.marker([39.77, -105.23]),
-				]).addTo(parentFeatureGroup)
+				]).addTo(parentFeatureGroup);
 
 				var bounds = new L.LatLngBounds(
-					new L.LatLng(-23, -105.23), 
+					new L.LatLng(-23, -105.23),
 					new L.LatLng(39.77, -102)
-				);			
+				);
 				expect(parentFeatureGroup.getBounds()).to.eql(bounds);
-			})
+			});
 		});
-		
-		describe('when a FeatureGroup contains nested LayerGroups/FeatureGroups as children', function(){
-			it("returns the bounds (LatLng) of the group, including bounds of nested groups' child layers", function(){
+
+		describe('when a FeatureGroup contains nested LayerGroups/FeatureGroups as children', function () {
+			it("returns the bounds (LatLng) of the group, including bounds of nested groups' child layers", function () {
 				var parentFeatureGroup = L.featureGroup();
 
-				var layerGroupWithChildren = L.layerGroup([
+				L.layerGroup([
 					L.marker([39.61, -105.02]),
 					L.marker([39.74, -104.99]),
 					L.marker([39.73, -104.8]),
 					L.marker([39.77, -105.23]),
-				]).addTo(parentFeatureGroup)
-				
-				var layerGroupWithNestedGroups = L.layerGroup([
+				]).addTo(parentFeatureGroup);
+
+				L.layerGroup([
 					L.marker([39.72, -103.31]),
 					L.layerGroup([
 						L.marker([39.51, -104]),
@@ -160,9 +160,9 @@
 						L.marker([42, -104]),
 					]),
 					L.marker([39.77, -105.32]),
-				]).addTo(parentFeatureGroup)
-				
-				var featureGroupWithNestedGroups = L.layerGroup([
+				]).addTo(parentFeatureGroup);
+
+				L.layerGroup([
 					L.marker([39.72, -103.31]),
 					L.layerGroup([
 						L.marker([39.51, -104]),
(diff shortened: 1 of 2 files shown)


Falke-Design:
Nice PR, maybe you can also add something what break the loop if a Group has already called `getBounds()`.

```
var fg = L.featureGroup();
var lg = L.featureGroup().addTo(fg);
fg.addTo(lg);
console.log(fg.getBounds());
```

This ends in a infinity loop, because both FeatureGroups has the other FeatureGroup as child

rafaelgalani:
@Falke-Design Yup, it loops forever. I wasn't expecting this scenario - to be honest, I didn't even thought that someone could do this. I'll add it to the unit tests and fix it. Thanks!

rafaelgalani:
@Falke-Design again, thanks for the feedback. The issue you pointed was fixed, I guess it's all OK now.

rafaelgalani:
Oops. 
Accidentally closed the pull request. Just reopened it now... Sorry about it 😅

rafaelgalani:
@IvanSanchez I'm not sure whether you saw the PR, but it's been some time since I opened it.. Since it hasn't been assigned a reviewer, I thought of mentioning you. 
Please let me know if you think that there's anything that could be improved 😸

johnd0e (changes requested):

johnd0e:
And question: do we have recursive processing for `getBounds` only?

Falke-Design:
@rafaelgalani can you please resolve the suggestions

Falke-Design:
> And question: do we have recursive processing for getBounds only?

Same here: #7273

rafaelgalani:
Sorry y'all for leaving this one behind. I'll rebase on `main` and work on it asap.

rafaelgalani:
I really dislike what I did on this one. I don't really know what I had in mind 😬 
I updated and refactored most of it, and took a different approach to handling recursion:

I think we should prevent it (as added by @johnd0e) instead of adding workarounds - also, this was a very hacky and _suboptimal_ workaround.

Please let me know if that looks better.

src/layer/LayerGroup.js
```diff
@@ -50,6 +52,26 @@ export const LayerGroup = Layer.extend({
...
+	// @method hasDescendant(layer: Layer): Boolean
+	// Returns `true` if the given layer is already a child of this instance, directly or nested.
+	hasDescendant(layer) {
+		let hasDescendant = false;
+
+		this.eachLayer((childLayer) => {
+			if (childLayer === layer || (childLayer instanceof LayerGroup && childLayer.hasDescendant(layer))) {
+				hasDescendant = true;
+			}
+		});
```
Falke-Design, line 70: Changing this to a for-loop would make it possible to break it if a descendant has been found.

src/layer/LayerGroup.js
```diff
@@ -50,6 +52,26 @@ export const LayerGroup = Layer.extend({
 		return this;
 	},
 
+	_preventLayerGroupRecursion(layer) {
+		if (layer instanceof LayerGroup && layer.hasDescendant(this)) {
+			throw new Error('Cannot add a LayerGroup as a child if it is already included, directly or nested.');
+		}
+	},
```
Falke-Design, line 59: This is a huge breaking-change, I don't know what I should thing about it

Falke-Design (changes requested):

src/layer/LayerGroup.js
rafaelgalani: Ummm, you're right. Sorry, I didn't think about this specifically...

So, in this case, I think we'd want to go with the recursive solution. It would be a non-breaking change, and it would also solve both issues.

rafaelgalani:
@Falke-Design Howdy! 
I updated the PR to go with the recursive behavior. LMK what you think. 

I still would like to understand a little more about the recursive use cases, but that's for another PR/issue 😬

src/layer/LayerGroup.js
```diff
@@ -147,6 +148,23 @@ export const LayerGroup = Layer.extend({
...
+		const bounds = toLatLngBounds();
+
+		for (const layerId in this._layers) {
+			const layer = this._layers[layerId];
+			if (visitedIds.has(layerId)) {
+				continue;
+			}
+
+			visitedIds.add(layerId);
+			bounds.extend(layer instanceof LayerGroup ? layer.getBounds(visitedIds) : layer.getLatLng());
```
Falke-Design, line 165: Not all layers have `.getLatLng()`, so we need to check for `getBounds()` (ImageOverlay) and maybe `getLatLngs()` too. I would add also a check for `getLatLng`, because maybe a layer is included with no coordinates.

```suggestion
if(layer instanceof LayerGroup || layer.getBounds){
			bounds.extend(layer.getBounds(visitedIds));
} else if(layer.getLatLngs){
			bounds.extend(layer.getLatLngs());
} else if(layer.getLatLng){
			bounds.extend(layer.getLatLng());
}
```
bounds.extend(layer.getBounds ? layer.getBounds() : layer.getLatLng());

src/layer/LayerGroup.js
```diff
@@ -147,6 +148,23 @@ export const LayerGroup = Layer.extend({
 	// Returns the internal ID for a layer
 	getLayerId(layer) {
 		return Util.stamp(layer);
+	},
+
+	// @method getBounds(): LatLngBounds
+	// Returns the LatLngBounds of the Layer Group (created from bounds and coordinates of its children).
+	getBounds(visitedIds = new Set()) {
```
Falke-Design, line 155: Let us make the variable `visitedIds` "private":
```suggestion
	getBounds(_visitedIds = new Set()) {

```

Falke-Design (changes requested):
You included the recursive logic for `getBounds()` but no test.

Do something like this:
```
const childGroup = new FeatureGroup();
const parentGroup = new FeatureGroup([childGroup]);
childGroup.addLayer(parentGroup);

parentGroup.getBounds(); --> No error
```

--------------

I still would like to get this recursive solution but I don't know if it is the best. Maybe we should prevent it like your solution but this will be a breaking change. 
I think for now we should go with the recursive solution, but maybe another maintainer can let us know what he thinks

rafaelgalani:
@Falke-Design Thanks for the review! I just updated the PR with your suggestions, lmk what you think.

spec/suites/layer/LayerGroupSpec.js
```diff
@@ -112,4 +112,64 @@
...
+			const bounds = L.latLngBounds(southWest, northEast);
+
+			let result = [];
+			const callback = () => {
+				result = parentGroup.getBounds();
+			};
+			expect(callback).to.not.throwError();
+			expect(result).to.eql(bounds);
+		});
+	});
```
Falke-Design, line 153: This works with the current solution too, you need to add to the `childGroup` the `parenGroup`: `childGroup.addLayer(parentGroup)`

Falke-Design (changes requested):

spec/suites/layer/LayerGroupSpec.js
rafaelgalani: Oops, forgot to add it. My bad.

rafaelgalani:
☝🏻 Sorry for that, clicked it by accident

Falke-Design (approved):

rafaelgalani:
Rebase on `main` ☝🏻

src/layer/LayerGroup.js
```diff
@@ -147,6 +148,29 @@ export const LayerGroup = Layer.extend({
...
+		const bounds = toLatLngBounds();
+
+		for (const layerId in this._layers) {
+			const layer = this._layers[layerId];
+			if (_visitedIds.has(layerId)) {
+				continue;
+			}
+
+			_visitedIds.add(layerId);
+			if (layer instanceof LayerGroup || layer.getBounds) {
```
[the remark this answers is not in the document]
jonkoops: I think a `typeof` check would suffice here. Just thinking out loud if all entities passed here have some kind of base class with a `getBounds()` method (in which case we could `instanceof` that base class).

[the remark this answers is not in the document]
Falke-Design: Let us start with a typeof check
```suggestion
			if (layer instanceof LayerGroup || typeof layer.getBounds === 'function') {
```

PR #7257, opened 2020-09-07
New plugin: Add Leaflet.Legend

ptma:
Leaflet.Legend is a plugin for Leaflet that display legend symbols and toggle overlays.


IvanSanchez (approved):
Looks good!

PR #7259, opened 2020-09-08
fix (Keyboard): Allow WCAG21 focus visible criteria for keyboard navigation - fixes #6986

### LINKED ISSUE #6986: When Tabbing through map, focus outline flashes but does not stay lit up
Try to use Tab to advance through focusable elements in a Leaflet map. For example, try with the example map on the Leaflet homepage (https://leafletjs.com/). Elements including the "x" to close the popup, the zoom in ("+") and zoom out ("-") controls, and the bottom-right corner links for "Leaflet" and "OpenStreetMap", will briefly flash the focus ring, but it will then disappear. The element is actually still focused, there is no visible indication. Among other things, this is a nightmare for accessibility.

This seems to be occurring because the `Map` `_handleDOMEvent` function calls `DomUtil.preventOutline` for all `mousedown`, `keypress`, `keyup`, and `keydown` events. `DomUtil.preventOutline` is injecting `outline: none` into the `element.style`. With the caveat that I am not familiar with this code base, I don't understand this approach. If the desire is to remove the focus ring outline, why wouldn't this be done with CSS on `:focus`? This would prevent the brief focus ring flash, and it would allow for it to be overridden in CSS (without resorting to `!important`).

jafin:
Allow WCAG21 focus visible criteria for keyboard navigation by not stripping outline on focus for keyboard events, only mouse events.

For issue #6986 

https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1389,7 +1389,7 @@ export var Map = Evented.extend({
 
 		var type = e.type;
 
-		if (type === 'mousedown' || type === 'keypress' || type === 'keyup' || type === 'keydown') {
+		if (type === 'mousedown') {
 			// prevents outline when clicking on keyboard-focusable element
 			DomUtil.preventOutline(e.target || e.srcElement);
 		}


IvanSanchez (approved):
LGTM

clairesunstudio:
@jafin @IvanSanchez  thank you for the fix! When will this be released?

jafin:
@clairesunstudio I'm not part of the release team so I can't answer sorry.  If you need it urgently you could use https://www.npmjs.com/package/patch-package to patch the existing release.

PR #7260, opened 2020-09-08
Fix excessive click in desktop Safari

### LINKED ISSUE #7255: Leaflet 1.7.1 causes 2 click events to be emitted by Leaflet Controls in Safari

johnd0e:
Tap handler is useful only in iOS

Fix #7255

### DIFF
--- a/src/map/handler/Map.Tap.js
+++ b/src/map/handler/Map.Tap.js
@@ -131,6 +131,7 @@ export var Tap = Handler.extend({
 // @section Handlers
 // @property tap: Handler
 // Mobile touch hacks (quick tap and touch hold) handler.
-if (Browser.touch && (!Browser.pointer || Browser.safari)) {
+var iOS = Browser.safari && Browser.mobile;
+if (Browser.touch && (!Browser.pointer || iOS)) {
 	Map.addInitHook('addHandler', 'tap', Tap);
 }


IvanSanchez (approved):

clairesunstudio:
thanks for the fix! When will this be released?

Issue #7254, opened 2020-09-05
Examples Should Relate to Tutorials - Using GeoJSON with Leaflet


flywire:
These comments specifically refer to [Using GeoJSON with Leaflet](https://leafletjs.com/examples/geojson/) but apply more generally.

A tutorial should work towards a solution step by step allowing opportunities for people to try what they are learning. The tutorial is frustrating, the examples aren't used in the demo example at the end and the files where the content occurs aren't referenced.

This tutorial could easily follow the [Leaflet Quick Start Guide](https://leafletjs.com/examples/quick-start/) tutorial by adding geojson data step by step. This would ensure the preparation of the map is fully explained even though the basics are out of scope for this topic.

flywire:
The [Leaflet Quick Start Guide](https://leafletjs.com/examples/quick-start/), the most basic tutorial, doesn't even build the example step by step with code the users can post straight into the code development environment. Code in the demo example is marked with `|` in the margin.

```html
    <!DOCTYPE html>
    <html>
    <head>
        
        <title>Quick Start - Leaflet</title>
    
        <meta charset="utf-8" />
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        
        <link rel="shortcut icon" type="image/x-icon" href="docs/images/favicon.ico" />
    
|       <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" integrity="sha512-xodZBNTC5n17Xt2atTPuE1HxjVMSvLVW9ocqUKLsCC5CXdbqCmblAshOMAS6/keqq/sMZMZ19scR4PsZChSR7A==" crossorigin=""/>
|       <script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js" integrity="sha512-XQoYMqMTK8LvdxXYG3nZ448hOEQiglfqkJs1NOQV44cWnUrBc8PkAOcXy20w0vlaXaVUearIOBhiXZ5V3ynxwA==" crossorigin=""></script>
    
    
        
    </head>
    <body>
    
    
    
?   <div id="mapid" style="width: 600px; height: 400px;"></div>
    <script>
    
|       var mymap = L.map('mapid').setView([51.505, -0.09], 13);
    
[...]

PR #7269, opened 2020-09-11
Revised for tutorial and example consistency

flywire:
Revised for tutorial and example consistency:

Change tutorial text to address any issues:
- no metadata
- source integrity check causing issues and was removed
- source changed to osm to remove compexities of using registered data
- consolidating `var` statements not described or included in example
- consolidating `.bindPopup` command statements not described or included in example


mourner (changes requested):
I might address other changes later, but one thing to note is that OSM tile server is not used in the tutorials on purpose — the team behind OSM didn't want us to encourage using it on production websites, since it's not suited to that kind of load and it's expensive to keep. So I can't accept this particular change for sure. E.g. see here https://twitter.com/OSM_Tech/status/1308465290263629825

flywire:
https://github.com/flywire/Leaflet/commits/patch-1 changed, I expected it would show here.

(Discussed at https://github.com/Leaflet/Leaflet/issues/7254#issuecomment-687609542.)

flywire:
@mourner requested changes done with #92c8d25

example-basic.md
```diff
@@ -0,0 +1,16 @@
+---
```
IvanSanchez, line 1: I think you meant to edit files in `docs/`, but the changes here are *adding* files to the root (`/`) directory of the repo instead.

IvanSanchez (changes requested):

mourner:
-> #7816

PR #7270, opened 2020-09-12
Remove circular dependencies on build

### LINKED ISSUE #5968: Circular dependency
## How to reproduce

- Leaflet version I'm using: v1.2.0
- Browser (with version) I'm using: chrome 63.0.3239.84
- OS/Platform (with version) I'm using: os x

1. Add a log to `CRS.Earth.js`
```
console.log(CRS);
```
1. import crs index
```
<script type="module">
import * as index from '/node_modules/leaflet/src/geo/crs/index.js';
</script>
```

CRS should exist. However I got `undefined`;

----
Just shift my project to ES Modules. I intend to import modules separately from Leaflet. But got some strange results. After struggled with the error, I finally found it's due to circular dependencies.

For the example above, moving `CRS.Earth` in crs/index.js to the first item solves the problem.
While since the doc page lists 'Modular build system' as a feature, I am not sure if that is an ideal solution.

Btw, I found this tool is useful: https://github.com/pahen/madge
The result:
image

Will you deem it a problem? Or should I import the `Leaflet.js` directly?

jafin:
Remove Circular dependencies on build. Fixes #5968.

```
(!) Circular dependency: src\geo\crs\CRS.Earth.js -> src\geo\crs\CRS.js -> src\geo\LatLng.js -> src\geo\crs\CRS.Earth.js
(!) Circular dependency: src\geo\LatLng.js -> src\geo\LatLngBounds.js -> src\geo\LatLng.js
(!) Circular dependency: src\core\Browser.js -> src\layer\vector\SVG.Util.js -> src\core\Browser.js
(!) Circular dependency: src\dom\DomEvent.js -> src\dom\DomEvent.Pointer.js -> src\dom\DomEvent.js
(!) Circular dependency: src\dom\DomEvent.js -> src\dom\DomUtil.js -> src\dom\DomEvent.js
```
API compatibility should be maintained, I've tried to keep the exports in the same modules to avoid changing.

## Challenges
I could not easily separate `LatLng` and `LatLngBounds`, they seem tightly coupled, so they were merged. 
If anyone has any peer advice on a better solution, I'm more than happy to amend the code. 

## Note
This came about as I was doing some work on Rollup and Terser, and when running the Karma build it would spiral into an endless loop, related to the above circular dependencies.


dist/leaflet-src.js.bak
```diff
@@ -0,0 +1,11680 @@
+/* @preserve
```
IvanSanchez, line 1: Please remove this file.

src/dom/getScale.js
```diff
@@ -0,0 +1,17 @@
+// @function getScale(el: HTMLElement): Object
```
IvanSanchez, line 1: Docstrings are missing the namespace. Please add `// @namespace DomEvent` before.

src/dom/preventDefault.js
```diff
@@ -0,0 +1,16 @@
+// @function preventDefault(ev: DOMEvent): this
```
IvanSanchez, line 1: Docstrings are missing the namespace. Please add // @namespace DomUtil before.

src/geo/LatLng.js
```diff
@@ -135,3 +134,302 @@ export function toLatLng(a, b, c) {
...
+export function wrapLatLng(latlng) {
+	var lng = this.wrapLng ? Util.wrapNum(latlng.lng, this.wrapLng, true) : latlng.lng,
+	lat = this.wrapLat ? Util.wrapNum(latlng.lat, this.wrapLat, true) : latlng.lat,
+	alt = latlng.alt;
+
+	return new LatLng(lat, lng, alt);
+}
+
+/*
+ * @class LatLngBounds
```
IvanSanchez, line 150: Why merge `LatLngBounds` into `LatLng`? They should be separate.

src/geo/crs/CRS.Earth.distance.js
```diff
@@ -0,0 +1,18 @@
+export var wrapLng = [-180, 180];
+
+export var R = 6371000;
+
+/**
+ * @param {{ lat: number; lng: number; }} latlng1
```
IvanSanchez, line 6: Docstrings not needed - doc shall be inherited from the `distance` method of the parent `L.CRS` class in the end.

src/geo/crs/CRS.Earth.js
```diff
@@ -12,22 +13,14 @@ import * as Util from '../../core/Util';
  */
 
 export var Earth = Util.extend({}, CRS, {
-	wrapLng: [-180, 180],
+	wrapLng: wrapLng,
 
 	// Mean Earth Radius, as recommended for use by
```
IvanSanchez, line 18: Move non-docstring comments about the mean earth radius to `CRS.Earth.distance`.

src/layer/vector/svgCreate.js
```diff
@@ -0,0 +1,7 @@
+// @function create(name: String): SVGElement
```
IvanSanchez, line 1: Missing docstrings namespace; add `// @namespace SVG`

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
IvanSanchez, line 1: I don't think this is needed on this PR. It might be appropriate for a *different* PR about improving QoL, but not for a refactor. File should be removed.

IvanSanchez (changes requested):
Overall looks very nice (removing circular deps for bundling is good!!), but I have issues regarding some extraneous files, and you should double check that the documentation builds correctly for all the methods that have changed places.

src/geo/LatLng.js
jafin: LatLng toBound uses toLatLngBounds, which references LatLng.  I could not discover how to escape this circular import unless I merged them. Do you have a suggestion to fix?

jafin: Also verified documentation built, note I did have issue with build on my platform but raised in different issue. (#7271)

IvanSanchez: API-breaking changes are frowned upon. I'd go with the `LatLngBounds` file adding functionality to `LatLng`, e.g. `LatLng.include({ toLatLngBounds: function(...){...} })`. Look at how the code for `GeoJSON` adds stuff to markers and polylines.

This approach would mean that a `LatLng` imported stand-alone would not have the `toLatLngBounds()` method, and that  users would need to import `LatLngBounds` as well. I think this is a lesser problem than always bundling both `LatLng` and `LatLngBounds` together.

jafin: @IvanSanchez , I have separated LatLng from LatLngBounds.  As you advised, leverage the Class.Extend to isolate toBounds(...) from LatLng and avoid circular dependency.  geo/index will export toBounds extended as LatLng to hopefully avoid api change.

plainheart:
Hi, I'm sorry but I'd like to get clear if this progress could go ahead? It seems the circular dependencies will invalidate the tree-shaking to some extent.

Malvoz:
@jafin Would you like to resolve the merge conflicts please?

jafin:
@Malvoz,  merge conflicts resolved.

Note:  LatLng and LatLngBounds are merged again.  I don't know of a clean option to keep them separated without keeping the circular dependency.  As noted from @IvanSanchez he raised a comment about these being merged.  I had originally responded with an approach to extend from Class.  But the testsuite had errors for this change, so it has been reverted.

If someone can provide guidance on a better way to maintain LatLng and LatLngBounds separately and still avoid the circular dependency I'm happy to attempt to fix

Removing `toBounds` from `LatLng` would solve it, but its a breaking change.

src/geo/crs/CRS.js
```diff
@@ -83,22 +84,31 @@ export var CRS = {
...
 	// axis wraps around a given range and how. Defaults to `[-180, 180]` in most
 	// geographical CRSs. If `undefined`, the longitude axis does not wrap around.
-	//
+	set wrapLng(wrapLng) {
+		Wrap.lng = wrapLng;
+	},
+	get wrapLng() {
+		return Wrap.lng;
+	},
+
```
mourner, line 100: - We can't use getters/setters yet because we support old versions of IE — this would be a breaking change.
- I'm really confused by this global state setup with `Wrap`, can you explain what's going on here?

mourner (changes requested):
Generally I'm in doubt about whether we should merge this — it definitely makes the code more confusing and organized in a more complicated way for no apparent gain. Is there any reason to avoid circular dependencies (which is a 100% valid ES use case) other than silencing Rollup warnings?

src/geo/crs/CRS.js
jafin: Apologies the intent was not to make the code more confusing.  Especially if there are incompatibilities Introduced.  
Re wrap. Ithe two wraped values were externalised to break the circular dep between LatLng and CRS.

Feel free to close the PR if it serves no value.

mourner: Sorry, I may have sounded harsh but I'm just trying to understand, not criticizing. If there's a legitimate case where circular deps cause problems without an easier workaround, we should definitely continue looking into this.

jafin: @mourner, originally I recall having errors with the Karma breaking because of the circular dependencies when calling the test runner with tests we have in a project (react/typescript/leaflet).  That was over 12 months ago, this project uses karma without error, and the former project has moved to jest.  So I do not have a hard requirement to fix this anymore and cannot produce a repro now.  

Further, as you mention, breaking changes for IE because of the use of the setters/getters, this was originally meant to be a refactor with minimal surface changes, I could not achieve this.   So I believe this should be closed to avoid introducing more pain that it is trying to solve.

PR #7274, opened 2020-09-14
Apply `:hover` styles to `:focus` as well

### LINKED ISSUE #6742: Leaflet / openstreet map links only identified by colour
The maps have 'leaflet' link which is identified only by color.

Criteria 1.4.1
Compliance success level A

0f49f0d4-0e1d-11e7-8e12-1a83beefc03c

**Steps to reproduce**
- View a map
- Look at the link text in the bottom right
- No visual indication that text is a link other than the colour difference

**Expected behavior**
When the link has focus or hover it should show an underline or some other visual indicator to pass WCAG compliance.
See https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-without-color.html

**Current behavior**
Only identified by a colour.
See https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F73 for an example failure.

**Environment**
- Leaflet version: all
- Browser (with version): all
- OS/Platform (with version): all

Malvoz:
The related outline fix in https://github.com/Leaflet/Leaflet/pull/7259 greatly improves usability for keyboard-only users, however it is common for developers to remove outlines, not considering the impact on users who do not use a mouse pointer. 

By applying the current `:hover` styles to `:focus` as well we can make sure there's _some_ feedback to users when they tab to a component, even if developers choose to remove outlines.

<hr>

WebAIM too [recommends](https://webaim.org/techniques/hypertext/link_text#hover) applying hover styles on focus:

> ### Important
> Nearly any time `a:hover` (or other hover effects) are defined in CSS, it should be modified to be `a:hover, a:focus` to ensure the same visual presentation is available when keyboard users navigate or 'tab' to the link.


dist/leaflet.css
```diff
@@ -277,8 +277,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	box-shadow: 0 1px 5px rgba(0,0,0,0.65);
 	border-radius: 4px;
 	}
-.leaflet-bar a,
-.leaflet-bar a:hover {
```
Malvoz, line 281: Any particular reason these styles were applied both to links in the default state and `:hover`?

Malvoz: If this edit is causing hesitancy by any means I'll gladly revert it to focus on the main issue.

Falke-Design (approved):

PR #7277, opened 2020-09-15
docs: add accessibility section to plugins guide

Malvoz:
Increase awareness and advocate for accessible content from plugin authors.

### DIFF
--- a/PLUGIN-GUIDE.md
+++ b/PLUGIN-GUIDE.md
@@ -15,9 +15,11 @@ This guide lists a number of best practices for publishing a Leaflet plugin that
 	- [File Structure](#file-structure)
 	- [Code Conventions](#code-conventions)
 	- [Plugin API](#plugin-api)
-3. [Publishing on NPM](#publishing-on-npm)
-4. [Module Loaders](#module-loaders)
-5. [Adding to the plugins list](#adding-to-the-plugins-list)
+3. [Content Accessibility](#content-accessibility)
+  - [Accessibility Testing Tools](#accessibility-testing-tools)
+4. [Publishing on NPM](#publishing-on-npm)
+5. [Module Loaders](#module-loaders)
+6. [Adding to the plugins list](#adding-to-the-plugins-list)
 
 ## Presentation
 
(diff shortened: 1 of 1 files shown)


Malvoz:
@IvanSanchez WDYT?

Also, should perhaps CONTRIBUTING.md possibly have a short sentence in the likes of "Try to write code with accessibility in mind, if you are not sure, ask the community for guidance." or some such.

hereweare666 (approved):

hereweare666 (approved):

hereweare666 (approved):

Malvoz:
May I ask for a review of this PR?

Falke-Design (approved):
I see no problems with adding this to the documentation.
For me personally:  The links are very "hard to read" and I don't think that a new plugin contributer will read through it. It would be good if there would be a list of common mistakes (specially for Leaflet) and / or maybe a link to a docu which explains the first steps and the important information about this topic. Do you understand what I mean?

Malvoz:
Thanks for the feedback @Falke-Design.

> For me personally: The links are very "hard to read" and I don't think that a new plugin contributer will read through it. 

Those documents are authoritative and are often referred to in the accessibility dev world, but they may take some time getting used to. I don't expect developers to read it all, but more so for them to know that they're there, and that they mean something. And they can _choose_ to use them (and learn how to use them) if they wish. Perhaps they should be replaced with- or accompanied by e.g. https://developer.mozilla.org/en-US/docs/Learn/Accessibility or https://github.com/mfairchild365/a11y-resources?

That second resource even covers what we're discussing here 😋:  

> Standards can be complex and hard to read, and therefore are often only referenced as a last resort. I'd argue that they should be consulted first. They are the most authorative source of how things **should** work.

(https://github.com/mfairchild365/a11y-resources#standards)

> It would be good if there would be a list of common mistakes (specially for Leaflet)

I did consider that, but how many would be enough, or too few? Generally the issues in Leaflet and plugins are issues that are common to all web sites. I.e. issues with: focus delegation, focus visibility, missing [accessible names](https://www.w3.org/TR/accname-1.1/#dfn-accessible-name) (and then I suppose I'd have to explain what an accessible name is and that it may be derived from: text from element/`aria-label`/`title`, which those resources would otherwise explain), etc. But these things are generally covered by the testing tools (which I do expect developers to use) or easily discovered during manual testing with a keyboard, and a screen reader (thus I recommended manual testing, and added a list of tools).

Malvoz:
> Apart from this PR, would you have the time to create PRs for each [accessibility issue](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility)? If not it would be very helpful if you can write into the issues what should be changed (from this into this), then we create PRs

I believe each issue should have a proposed solution already. I will try to find time to look over the issues tomorrow to clarify them, perhaps by using headings, such as "Motivation" and "Solution". 👍

Falke-Design:
> [...] to know that they're there, and that they mean something. And they can choose to use them (and learn how to use them) if they wish

Good point 👍 

> how many would be enough, or too few? [...] are common to all web sites

Yes I understand. I looked into your links and they have so much informations, that I closed the sites fast again 😉 But you are right there are enough tools out there. And you added already a good list to start.

> to look over the issues tomorrow to clarify them

Thanks. I think you have the most experience so it would be good to have clear suggestions and then we can change this in Leaflet 👍

Malvoz:
(off-topic, along with other comments here, maybe we should open a general Accessibility Discussion?)

@Falke-Design I've updated [the issues](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility+author%3AMalvoz), they should much clearer now. I obviously didn't use the Bug/Feature issue templates (those have makeshift headings), I guess I wasn't sure how to classify the issues when I opened them, I feel like they're kinda both. 😜

Falke-Design:
> maybe we should open a general Accessibility Discussion?

I hope the next questions discussion are below theier PR or issue. I don't think that we have much general discussion. Or do you think that it would be good?

Thanks for updating the issue they are now much clearer 👍 I think some of them will be no problem to merge but I also think that some of them will be rejected, but let us look what the future brings.

Malvoz:
I think this is a good addition as is, it could be tweaked later if necessary.

PR #7279, opened 2020-09-16
Refactor vmlCreate() so that it does not leak closure to `TypeError`

darcyparker:
When `document.namespaces.add()` throws `TypeError: Cannot read property 'add' of undefined`, the `TypeError` persists in the memory heap snapshot.  

I am not sure why Google Chrome does not garbage collect this `TypeError` because clearly it is not used in the returned fn. But if you look at memory Heap snapshot, you can see it persists after garbage collection.
* Try it with any leaflet map such as https://leafletjs.com/ and search heap snapshot for 'TypeError' and you will see it was never garbage collected after this simple feature test was performed.
image

With this change, the `TypeError` is garbage collected and not in memory heap snapshot.

I don't think it's serious... but the change is simple and with the change I can easily see the `TypeError` is removed from the heap snapshot in my application.

### DIFF
--- a/src/layer/vector/SVG.VML.js
+++ b/src/layer/vector/SVG.VML.js
@@ -14,10 +14,12 @@ export var vmlCreate = (function () {
 			return document.createElement('<lvml:' + name + ' class="lvml">');
 		};
 	} catch (e) {
-		return function (name) {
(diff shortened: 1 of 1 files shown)


IvanSanchez (approved):
The change is simple enough - but we should remove VML support together as part of #6136 anyway.

IvanSanchez:
And thanks for the detailed investigation, @darcyparker !

PR #7280, opened 2020-09-19
Introduced accessibility in zoom and Layer control

AkshataJ96:
With reference to #7211 ,
This pull requests addressed 2 things:

- Added 'aria-disabled = true | false' attribute in zoom control
- Modified the usage of 'label' tag in Layer control

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -307,7 +307,7 @@ export var Layers = Control.extend({
 	// IE7 bugs out if you create a radio dynamically, so you have to do it this hacky way (see http://bit.ly/PqYLBe)
 	_createRadioElement: function (name, checked) {
 
-		var radioHtml = '<input type="radio" class="leaflet-control-layers-selector" name="' +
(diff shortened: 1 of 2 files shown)


src/control/Control.Layers.js
```diff
@@ -342,6 +343,7 @@ export var Layers = Control.extend({
 		// https://github.com/Leaflet/Leaflet/issues/2771
 		var holder = document.createElement('div');
 
+		label.setAttribute('for', 'layer-select');
```
Malvoz, line 346: The `for` attribute is specifically for when you need to associate an `input` that is not a child element of `label`. I suggest reverting changes in Control.Layers.js. Other than that LGTM!

AkshataJ96: Oh okay. Thanks for the feedback! @Malvoz 
I have reverted these changes as per your suggestions.
Let me know if anything else needs to be changed.
Thanks!

mourner (approved):
Thank you!

PR #7313, opened 2020-10-18
Allow pass custom function in locate to get coordinates

### LINKED ISSUE #7296: Allow to pass the coordinates in .locate()
I'm making a web app with Capacitor and https didn't work in iOS so can't use `navigator.geolocation.getCurrentPosition()` but Capacitor exposes that in his modules, so I can get the coordinates, but not with the function leaflet uses

A new option in locate config to overwrite the function to read the coordinates in https://github.com/Leaflet/Leaflet/blob/bdd7ebf836201550f6d173943b728098b83f40d5/src/map/Map.js#L660-L665

I try to use https in iOS Capacitor but in the issues the devs say that didn't work anyways

**Additional context**
I think that the code would be something like this:
```js
this.map.locate({
    getCoordinates: async (onResponse, onError, options) => {
        return await Geolocation.getCurrentPosition(); // Return the same as navigator.geolocation.getCurrentPosition()
    }
})

if (options.getCoordinates) {
    if (options.watch) {
        this._locationWatchId = options.getCoordinates(options);

    } else {
        try {
            var pos = await options.getCoordinates(options);
            onResponse(pos);

        } catch (error) {
            onError(error);

        }
    }
} else {
    if (options.watch) {
        this._locationWatchId =
            navigator.geolocation.watchPosition(onResponse, onError, options);

    } else {
        navigator.geolocation.getCurrentPosition(onResponse, onError, options);

    }
}
```

Nisgrak:
Allow to pass a custom function to get coordinates, this allow leaflet to work with Capacitor like this.

The getCoordinates must return a Promise.

```js
this.map.locate({
    getCoordinates: async (options, callback) => {
        return Geolocation.getCurrentPosition(options); // Return the same as navigator.geolocation.getCurrentPosition();
        // return Geolocation.watchPosition(options, callback); // If you want to watch
    },
   // watch: true
})
```

Resolves #7296.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -663,24 +663,19 @@ export var Map = Evented.extend({
 					.then(function (locationWatchId) {
 						this._locationWatchId = locationWatchId;
 					});
-
 			} else {
 				options.getCoordinates(options)
 					.then(function (pos) {
 						onResponse(pos);
-
 					}).catch(function (error) {
 						onError(error);
-
 					});
 			}
 		} else if (options.watch) {
 			this._locationWatchId =
 				navigator.geolocation.watchPosition(onResponse, onError, options);
-
 		} else {
 			navigator.geolocation.getCurrentPosition(onResponse, onError, options);
-
 		}
 		return this;
 	},
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -655,13 +655,32 @@ export var Map = Evented.extend({
 		}
 
 		var onResponse = Util.bind(this._handleGeolocationResponse, this),
-		    onError = Util.bind(this._handleGeolocationError, this);
+			onError = Util.bind(this._handleGeolocationError, this);
 
-		if (options.watch) {
+		if (options.getCoordinates) {
+			if (options.watch) {
+				options.getCoordinates(options)
+					.then(function (locationWatchId) {
+						this._locationWatchId = locationWatchId;
+					});
+
+			} else {
+				options.getCoordinates(options)
+					.then(function (pos) {
+						onResponse(pos);
+
+					}).catch(function (error) {
+						onError(error);
+
+					});
+			}
+		} else if (options.watch) {
 			this._locationWatchId =
-			        navigator.geolocation.watchPosition(onResponse, onError, options);
+				navigator.geolocation.watchPosition(onResponse, onError, options);
+
 		} else {
 			navigator.geolocation.getCurrentPosition(onResponse, onError, options);
+
 		}
 		return this;
 	},
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -655,7 +655,7 @@ export var Map = Evented.extend({
 		}
 
 		var onResponse = Util.bind(this._handleGeolocationResponse, this),
-			onError = Util.bind(this._handleGeolocationError, this);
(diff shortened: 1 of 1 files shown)


src/map/Map.js
```diff
@@ -657,9 +657,23 @@ export var Map = Evented.extend({
...
-		if (options.watch) {
+		if (options.getCoordinates) {
+			if (options.watch) {
+				options.getCoordinates(options)
+					.then(function (locationWatchId) {
+						this._locationWatchId = locationWatchId;
+					});
+			} else {
+				options.getCoordinates(options)
+					.then(function (pos) {
```
rheh, line 668: anonymous function not required,  I think you can just do this:
`options.getCoordinates(options).then(onResponse)`

src/map/Map.js
```diff
@@ -657,9 +657,23 @@ export var Map = Evented.extend({
...
+						onResponse(pos);
+					}).catch(function (error) {
```
rheh, line 670: Similair here: I think you can just do this:
`options.getCoordinates(options)
    .then(onResponse)
    .catch(onError)
`

src/map/Map.js
```diff
@@ -657,9 +657,23 @@ export var Map = Evented.extend({
 		var onResponse = Util.bind(this._handleGeolocationResponse, this),
 		    onError = Util.bind(this._handleGeolocationError, this);
 
-		if (options.watch) {
+		if (options.getCoordinates) {
+			if (options.watch) {
+				options.getCoordinates(options)
+					.then(function (locationWatchId) {
+						this._locationWatchId = locationWatchId;
```
rheh, line 664: Are we sure 'this' Is referring to the context now it is wrapped in an anonymous function.  If the intent is to set _locationWatchId on the map object I think this needs to be an arrow function, or, bind the function to control the context of this?

rheh (approved):
LGMT

Nisgrak:
Any changes?

src/map/Map.js
```diff
@@ -657,9 +657,17 @@ export var Map = Evented.extend({
 		var onResponse = Util.bind(this._handleGeolocationResponse, this),
 		    onError = Util.bind(this._handleGeolocationError, this);
 
-		if (options.watch) {
+		if (options.getCoordinates) {
```
Falke-Design, line 660: Missing documentation

src/map/Map.js
```diff
@@ -657,9 +657,17 @@ export var Map = Evented.extend({
 		var onResponse = Util.bind(this._handleGeolocationResponse, this),
 		    onError = Util.bind(this._handleGeolocationError, this);
 
-		if (options.watch) {
+		if (options.getCoordinates) {
+			if (options.watch) {
+				this._locationWatchId = options.getCoordinates(options, onResponse);
```
Falke-Design, line 662: onError is missing

Falke-Design (changes requested):

johnd0e:
> Allow to pass a custom function to get coordinates

It's not about single function. There is `watch` option. There is `stopLocate` method. And all is tied to `navigator.geolocation.*`.

So.. I don't think that we can consistently support custom functions here.
If you need - rewrite `locate` with own method.

PR #7318, opened 2020-10-25
[docs] fixed issue with minus sign not showing

twei0407:
'&#x2212 ' The text set on the 'zoom out' button.
->
'−' The text set on the 'zoom out' button.

### DIFF
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -26,7 +26,7 @@ export var Zoom = Control.extend({
 		// The title set on the 'zoom in' button.
 		zoomInTitle: 'Zoom in',
 
-		// @option zoomOutText: String = '&#x2212;'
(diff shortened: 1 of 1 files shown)


johnd0e:
Wrong fix.
Plus sign is not expected to be shown here.

But there is still bug with apostrophe, which should be solved in [Leafdoc](https://github.com/Leaflet/Leafdoc/issues).

P.S.
I remember something similar was already reported...

IvanSanchez:
Related to #7317

I agree with @johnd0e , this is a Leafdoc upstream issue. The default value in question is a string containing a HTML escape sequence and *not* the related unicode glyph; the documentation should reflect that.

johnd0e:
> The default value in question is a string containing a HTML escape sequence and _not_ the related unicode glyph;
> the documentation should reflect that.

Well, that is pretty obvious from the value itself

PR #7320, opened 2020-10-26
docs: control-attribution-prefix may be `false`

simon04:


johnd0e:
Wrong place to fix: these *.html are autogenerated from source files.

simon04:
I know. That's why I've updated  the source file `src/control/Control.Attribution.js` as well.

johnd0e:
@simon04

simon04:
Sorry, but no. I'm not interested in revising and worsen this PR after 13 months.

PR #7325, opened 2020-10-28
Icon.DefaultSpec: test leaflet-default-icon-path

simon04:
Adds test for #4849.


johnd0e:
@simon04
You've probably noticed that many tests are failed?

Also consider #7092.

simon04:
Closing in favour of https://github.com/Leaflet/Leaflet/pull/7092…

PR #6539, opened 2019-03-02
Enable to define how GeoJSON polygons and lineString spawn leaflet layers

ronikar:
Today, L.GeoJson enables to define how GeoJSON points spawn Leaflet layers, but no for polygons and lineStrings.
I added to GeoJSON options two fields in order to enable users to define how polygons and lineStrings spawn.

PR #7327, opened 2020-11-02
Add lineToLayer and polygonToLayer options to GeoJSON layer

MasterEric:
# Summary

This change adds two new options to the GeoJSON layer. These provide similar functionality to the `pointToLayer` function. However, where `pointToLayer` allows an override in behavior in the creation of Markers, `lineToLayer` and `polygonToLayer` override the creation of Polylines/Multipolylines and Polygons/Multipolygons, respectively:

> **lineToLayer**
A Function defining how GeoJSON LineString and MultiLineString geometry types spawn Leaflet layers. It is internally called when data is added, passing the GeoJSON line feature and its LatLngs. The default is to spawn a default Polyline: 
> ```
> function(geoJsonLine, latlngs) {
>     return L.polyline(latlngs);
> }
> ```

> **polygonToLayer**
A Function defining how GeoJSON Polygon and MultiPolygon geometry types spawn Leaflet layers. It is internally called when data is added, passing the GeoJSON polygon feature and its LatLngs. The default is to spawn a default Polygon: 
> ```
> function(geoJsonPolygon, latlngs) {
>     return L.polygon(latlngs);
> }
> ```

# Use Case

This code allows users to use their own objects to define how polylines and polygons are rendered from GeoJSON data onto a Leaflet map.

As an example: If you wanted the GeoJSON line to be rendered as a [L.multiOptionsPolyline](https://github.com/hgoebl/Leaflet.MultiOptionsPolyline) instead of an `L.polyline`, the existing code base did not provide this option to developers.

I encountered this case, and found it odd that only point markers had the option to override in this manner.

# Changes Made

I implemented the relevant code, modified the relevant documentation comments to reflect the new properties, and ensured that code spacing/linting matched that of the surrounding code. I also tested the code functions as intended by adding an example of `lineToLayer` to the documentation, by changing the rendered lines in `docs/examples/geojson` to create polylines with a magenta color.

### DIFF
--- a/docs/examples/geojson/example.md
+++ b/docs/examples/geojson/example.md
@@ -65,7 +65,16 @@ title: GeoJSON tutorial
 			return false;
 		},
 
-		onEachFeature: onEachFeature
+		onEachFeature: onEachFeature,
+
+		lineToLayer: function (feature, latlngs) {
+			console.log('lineToLayer');
+	    	return L.polyline(latlngs, {
+	    		color: "#FF00FF",
+	    		weight: 4,
+				opacity: 1
+		    })
+		}
 	}).addTo(map);
 
 	var coorsLayer = L.geoJSON(coorsField, {
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -37,12 +37,35 @@ export var GeoJSON = FeatureGroup.extend({
 	 * @aka GeoJSON options
 	 *
 	 * @option pointToLayer: Function = *
-	 * A `Function` defining how GeoJSON points spawn Leaflet layers. It is internally
-	 * called when data is added, passing the GeoJSON point feature and its `LatLng`.
+	 * A `Function` defining how GeoJSON Point and MultiPoint geometry types
+	 * spawn Leaflet layers. It is internally called when data is added,
+	 * passing the GeoJSON point feature and its `LatLng`.
 	 * The default is to spawn a default `Marker`:
 	 * ```js
 	 * function(geoJsonPoint, latlng) {
 	 * 	return L.marker(latlng);
+	 * }
+	 * ```
+	 * 
+	 * @option lineToLayer: Function = *
+	 * A `Function` defining how GeoJSON LineString and MultiLineString geometry types
+	 * spawn Leaflet layers. It is internally called when data is added,
+	 * passing the GeoJSON line feature and its `LatLngs`.
+	 * The default is to spawn a default `Marker`:
+	 * ```js
+	 * function(geoJsonLine, latlngs) {
+	 * 	return L.marker(latlng);
+	 * }
+	 * ```
+	 * 
+	 * @option polygonToLayer: Function = *
+	 * A `Function` defining how GeoJSON Polygon and MultiPolygon geometry types
+	 * spawn Leaflet layers. It is internally called when data is added,
+	 * passing the GeoJSON polygon feature and its `LatLngs`.
+	 * The default is to spawn a default `Marker`:
+	 * ```js
+	 * function(geoJsonPolygon, latlngs) {
+	 * 	return L.marker(latlngs);
 	 * }
 	 * ```
 	 *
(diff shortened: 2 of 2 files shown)


johnd0e:
See similar rejected #6539.
And pay attention to failed tests.

MasterEric:
Attempted resolution of failed test using a squashed commit.

As stated under the Use Case section of the pull request, there are multiple plugins, such as MultiOptionsPolyline and TextPath, which override how lines are created.

I made this pull request in part such that I could use the Github branch as a dependency for my own project, and I am currently utilizing this feature to replace the rendering of a solid line with a TextLine with arrows, to create a directed path.

MasterEric:
The test suit error is occurring because of the option `markersInheritOptions`; the current code states that by default, the individual lines of the GeoJSON do not inherit the options of the GeoJSON.

The question is how ought the code be revised from here. I see several paths forward:

* Leave the library as is, and the test modified to set `markersInheritOptions` to true. As seen, this is a breaking change so I don't think this is suitable.
* Add new options for `polyLinesInheritOptions` and `polygonsInheritOptions`. Either these have a default value of true (which may be confusing for developers as there are inconsistent default values for similar parameters) or these have a default value of false, which is again a code change which modifies how maps are rendered.
* Remove polylines and polygons for consideration for the `markersInheritOptions` option and have it always inherit. This would result in users maps looking the same, but would leave no ability to allow users to turn inheritance off.

What is the best fix?

MasterEric:
Apologies for the delay on fixing this PR.

I have opted for the second option, adding the `linesInheritOptions` and `polygonsInheritOptions` parameters, which default to true (as previous version of leaflet inherited options as well).

docs/examples/geojson/example.md
```diff
@@ -65,7 +65,16 @@ title: GeoJSON tutorial
 			return false;
 		},
 
-		onEachFeature: onEachFeature
+		onEachFeature: onEachFeature,
+
+		lineToLayer: function (feature, latlngs) {
+			console.log('lineToLayer');
```
johnd0e, line 71:
```suggestion
```

docs/examples/geojson/example.md
```diff
@@ -65,7 +65,16 @@ title: GeoJSON tutorial
...
+		onEachFeature: onEachFeature,
+
+		lineToLayer: function (feature, latlngs) {
+			console.log('lineToLayer');
+	    	return L.polyline(latlngs, {
+	    		color: "#FF00FF",
+	    		weight: 4,
+				opacity: 1
+		    })
+		}
```
johnd0e, line 77: Something wrong with whitespaces here.

johnd0e:
> As stated under the Use Case section of the pull request, there are multiple plugins, such as MultiOptionsPolyline and TextPath, which override how lines are created.

Well, I see now that this can be useful.
Still I am not convinced yet that we need to put it into Leaflet core.
How about to make this by plugin extending `L.GeoJSON` with all extra functions?

src/layer/GeoJSON.js
```diff
@@ -81,6 +104,12 @@ export var GeoJSON = FeatureGroup.extend({
 	 *
 	 * @option markersInheritOptions: Boolean = false
 	 * Whether default Markers for "Point" type Features inherit from group options.
+   *
+   * @option linesInheritOptions: Boolean = true
+	 * Whether default Markers for "Polyline" type Features inherit from group options.
```
Falke-Design, line 109:
```suggestion
	 * Whether default Polyline for "LineString" type Features inherit from group options.

```

src/layer/GeoJSON.js
```diff
@@ -81,6 +104,12 @@ export var GeoJSON = FeatureGroup.extend({
 	 *
 	 * @option markersInheritOptions: Boolean = false
 	 * Whether default Markers for "Point" type Features inherit from group options.
+   *
+   * @option linesInheritOptions: Boolean = true
+	 * Whether default Markers for "Polyline" type Features inherit from group options.
+   *
+   * @option polygonsInheritOptions: Boolean = true
+	 * Whether default Markers for "Polygon" type Features inherit from group options.
```
Falke-Design, line 112:
```suggestion
	 * Whether default Polygon for "Polygon" type Features inherit from group options.

```

src/layer/GeoJSON.js
```diff
@@ -37,15 +37,38 @@ export var GeoJSON = FeatureGroup.extend({
...
 	 * ```js
 	 * function(geoJsonPoint, latlng) {
 	 * 	return L.marker(latlng);
 	 * }
 	 * ```
 	 *
+	 * @option lineToLayer: Function = *
+	 * A `Function` defining how GeoJSON LineString and MultiLineString geometry types
+	 * spawn Leaflet layers. It is internally called when data is added,
+	 * passing the GeoJSON line feature and its `LatLngs`.
```
Falke-Design, line 53:
```suggestion
	 * passing the GeoJSON lineString feature and its `LatLngs`.

```

src/layer/GeoJSON.js
```diff
@@ -37,15 +37,38 @@ export var GeoJSON = FeatureGroup.extend({
...
+	 * The default is to spawn a default `Polyline`:
+	 * ```js
+	 * function(geoJsonLine, latlngs) {
+	 * 	return L.polyline(latlng);
```
Falke-Design, line 57:
```suggestion
	 * 	return L.polyline(latlngs);

```

Falke-Design (changes requested):
Please update your PR.

I think this features can be useful in Leaflet core, because it makes it a step more easier customizeable which is one of the greate things in Leafet.

PR #7328, opened 2020-11-03
Bugfix min/maxZoom & detectRetina

bozdoz:
if detectRetina is true, maxZoom was being set to a value less than minZoom.  This PR adds a test, and fixes the logic for setting min/maxZoom in L.TileLayer.

### DIFF
--- a/spec/before.js
+++ b/spec/before.js
@@ -1,2 +1,5 @@
 // Trick Leaflet into believing we have a touchscreen (for desktop)
-if (window.TouchEvent) { window.ontouchstart = function(){} };
+if (window.TouchEvent) { window.ontouchstart = function () {}; }
+
+// allow tests for `detectRetina`
+window.devicePixelRatio = 2;
--- a/spec/suites/layer/tile/TileLayerSpec.js
+++ b/spec/suites/layer/tile/TileLayerSpec.js
@@ -430,6 +430,28 @@ describe('TileLayer', function () {
 				});
 			});
 		}
+
+		it('loads tiles with matching min/maxZoom and detectRetina', function (done) {
+			const kittenLayer = kittenLayerFactory({
+				minZoom: 1,
+				maxZoom: 1,
+				detectRetina: true,
+			});
+
+			var tilesLoaded = 0;
+
+			kittenLayer.on('tileload', function () {
+				tilesLoaded += 1;
+			});
+
+			kittenLayer.on('load', function () {
+				// copied from loads 8 kittens zoom 1 above
+				expect(tilesLoaded).to.be(8 * 4);
+				done();
+			});
+
+			map.addLayer(kittenLayer).setView([0, 0], 1);
+		});
 	});
 
 	describe('#setUrl', function () {
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -93,10 +93,10 @@ export var TileLayer = GridLayer.extend({
 
 			if (!options.zoomReverse) {
 				options.zoomOffset++;
-				options.maxZoom--;
+				options.maxZoom = Math.max(options.minZoom, options.maxZoom - 1);
 			} else {
 				options.zoomOffset--;
-				options.minZoom++;
+				options.minZoom = Math.min(options.maxZoom, options.minZoom + 1);
 			}
 
 			options.minZoom = Math.max(0, options.minZoom);


spec/before.js
```diff
@@ -1,2 +1,5 @@
 // Trick Leaflet into believing we have a touchscreen (for desktop)
-if (window.TouchEvent) { window.ontouchstart = function(){} };
+if (window.TouchEvent) { window.ontouchstart = function () {}; }
```
johnd0e, line 2: Please avoid whitespace-only changes

spec/before.js
```diff
@@ -1,2 +1,5 @@
...
+
+// allow tests for `detectRetina`
+window.devicePixelRatio = 2;
```
johnd0e, line 5: You should not change this globally.

spec/suites/layer/tile/TileLayerSpec.js
```diff
@@ -430,6 +430,28 @@ describe('TileLayer', function () {
 				});
 			});
 		}
+
+		it('loads tiles with matching min/maxZoom and detectRetina', function (done) {
```
johnd0e, line 434: I'd prefer more straightforward tests just ensuring that min/maxZoom is changed in expected way.

johnd0e (changes requested):
Thank you for this fix.
But please pay attention to review comments.

spec/suites/layer/tile/TileLayerSpec.js
bozdoz: I agree. I didn't know what the issue was until I wrote this test.

spec/before.js
bozdoz: Do you have any recommendations as to how to set retina to true at runtime for the tilelayer?

bozdoz: I don't really see a good way to mock L.Browser.retina, since it's just an exported var/const, and I can't import anything in the spec file so that I can mock it with sinon.

bozdoz: Looks like I might be able to install karma-requirejs in order to mock L.Browser: https://medium.com/@BoweiHan/setup-for-testing-amd-modules-with-karma-requirejs-and-mocha-2be3931c6a72

Is that something we would want to do?

bozdoz: Just tried installing requirejs and karma-requirejs and it is a pain to configure.  Unless someone else knows better, I won't be able to add that for mocking L.Browser.

johnd0e: > Do you have any recommendations as to how to set retina to true at runtime for the tilelayer?

That is good question. And I see now that our suite lacks `detectRetina` tests completely(!).
ATM I can propose such approach:
1) introduce `skipIfNoRetina`/`skipIfRetina` helpers, similar to that: https://github.com/Leaflet/Leaflet/blob/bc918d4bdc2ba189807bc207c77080fb41ecc196/spec/suites/SpecHelper.js#L57-L60
2) Setup separate browser config with proper retina params, similar to that: https://github.com/Leaflet/Leaflet/blob/bc918d4bdc2ba189807bc207c77080fb41ecc196/spec/karma.conf.js#L108-L114
(Actually I do not know if FF has such option, need to inspect it)

johnd0e: > Looks like I might be able to install karma-requirejs in order to mock L.Browser:

I am not sure that it is possible to achieve in this way.
But you are right: that can be really useful, and not only for this test, but for many other.
I believe that we should refactor `L.Browser` to be mutable (and then just set arbitrary `L.Browser.retina` in runtime).

bozdoz: I can definitely look into making L.Browser mutable.  At first glance that does not seem difficult.

johnd0e: @IvanSanchez @mourner 
What do you think, is it good idea to make `L.Browser` mutable?

bozdoz:
@johnd0e I made L.Browser mutable, for your consideration.

johnd0e:
@bozdoz 
Looks good.
But as it goes beyond of current subject - I'd suggest to post it as separate PR.

bozdoz:
Reverted, but tests will fail now

spec/suites/layer/tile/TileLayerSpec.js
```diff
@@ -431,6 +431,33 @@ describe('TileLayer', function () {
 				});
 			});
 		}
+
+		it('sets min/maxZoom appropriately with detectRetina', function (done) {
+			const maxZoom = 1;
```
Falke-Design, line 436:
```suggestion
			var maxZoom = 1;
```

if you replace `const` with `var` the tests should be running fine in IE10

Falke-Design (changes requested):

spec/suites/layer/tile/TileLayerSpec.js
bozdoz: Right, simple mistake.

bozdoz:
A timeout issue on that last failed IE build

Falke-Design (changes requested):
Can you please add the same (without -1) for none retina displays and add a test for it (f.ex. minZoom: 10, maxZoom:9) :

```diff
if (options.detectRetina && Browser.retina && options.maxZoom > 0) {

			options.tileSize = Math.floor(options.tileSize / 2);
			if (!options.zoomReverse) {
				options.zoomOffset++;
				options.maxZoom = Math.max(options.minZoom, options.maxZoom - 1);
			} else {
				options.zoomOffset--;
				options.minZoom = Math.min(options.maxZoom, options.minZoom + 1);
			}
-		}
+		} else {
+			if (!options.zoomReverse) {
+				options.maxZoom = Math.max(options.minZoom, options.maxZoom);
+			} else {
+				options.minZoom = Math.min(options.maxZoom, options.minZoom);
+			}
+		}
```


Usecase:
If `minZoom` is greater then `maxZoom` we set the `minZoom` value to `maxZoom` and then tiles are loading for only one zoom level.

bozdoz:
@Falke-Design Not a fan of that at all.  It makes sense to adjust min/max zoom for retina, but adjusting min/max zoom just because of garbage inputs seems like too much hand-holding.  Is there a reason to make this change?  Bug fix?

Falke-Design:
Yes you are right we should not care to much about the input and if it is valid. 
But if a user is developing on a retina screen with the options `{maxZoom: 9, minZoom: 10, detectRetina: true}` everything looks good and is working and then a other user comes without a retina screen, the tiles are not loading anymore. So I think we should prevent from erros like that.

I don't know what is the best way but in my eyes we need to throw an error or overwrite the `maxZoom` with the `minZoom` value.


Same code with retina screen:
grafik

and without retina:
grafik

bozdoz:
@Falke-Design Done.

Falke-Design (approved):

PR #7335, opened 2020-11-10
makes L.Browser mutable

bozdoz:
As suggested in https://github.com/Leaflet/Leaflet/pull/7328

tests can't test for detectRetina without mutable L.Browser.


johnd0e (dismissed):

src/layer/tile/TileLayer.js
```diff
@@ -93,10 +93,10 @@ export var TileLayer = GridLayer.extend({
 
 			if (!options.zoomReverse) {
 				options.zoomOffset++;
-				options.maxZoom--;
+				options.maxZoom = Math.max(options.minZoom, options.maxZoom - 1);
```
bozdoz, line 96: shoot; I accidentally included the fix to #7328. I will revert this @johnd0e

bozdoz: For some reason my last commit isn't being added to this PR; perhaps because you've approved it @johnd0e

johnd0e: > perhaps because you've approved it @johnd0e

I don't think that there can be any connection. But whatever..

bozdoz: Looks like it worked! The reverts commit has been added! @johnd0e

johnd0e (approved):

Falke-Design:
@bozdoz please solve the conflicts

PR #7344, opened 2020-11-20
Fix closed coord's reference in latLngsToCoords

marlo22:
Hi :wink: 

I've noticed that "closed coord" in `latLngsToCoords` method is just a reference to the first coords. Sometimes it can make troubles, for example when we return GeoJSON by `toGeoJSON` method and then we mutate the coordinates directly.

An example is available here:
* live - https://marlo22.github.io/leaflet-coords/
* repo - https://github.com/marlo22/leaflet-coords/blob/main/index.html

I get a GeoJSON object by `toGeoJSON` method, then iterate through each coord and reverse all of them.
As you can see (in the browser's console) in the example above first and last coords aren't inverted correctly (they're inverted two times and as a result they have initial value).


johnd0e:
You are using `turf.coordEach` to iterate over coordinates, and I suppose that is exact problem.

Otherwise please point to specs where we can find that geojson coords **must not** be closed.

Falke-Design (approved):
I think this is a valid suggestion and makes for me totally sense. I would like to merge this but want the opinion of another maintainer

jonkoops (changes requested):
This is a perfectly valid case. Could you add a test to ensure this code does not accidentally regress?

mourner (approved):

PR #7345, opened 2020-11-21
fix outdated OpenStreetMap attribution texts + url

saerdnaer:
OpenStreetMap changed it's data licence from CC BY-SA 2.0 to ODBL – c.f. https://www.openstreetmap.org/copyright/en

> How to credit OpenStreetMap
> We require that you use the credit “© OpenStreetMap contributors”.
> You must also make it clear that the data is available under the Open Database License. You may do this by linking to this copyright page. […]


mourner (approved):
Good call, thank you!

PR #7346, opened 2020-11-22
Use a legible `font-size` in the attribution

Malvoz:
Fix Lighthouse audit failure "Document doesn't use legible font sizes" by increasing the font-size by 1px.

See [the report](https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fleafletjs.com%2Fexamples%2Fquick-start%2Fexample.html#seo) from auditing the URL: https://leafletjs.com/examples/quick-start/example.html (generated from https://web.dev/measure).

This is an accessibility improvement, although Lighthouse categorizes it as an SEO audit.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -418,7 +418,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-container .leaflet-control-attribution {
 	font-size: 12px;
-  }
+	}
 .leaflet-container .leaflet-control-scale {
 	font-size: 11px;
 	}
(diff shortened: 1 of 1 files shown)


mourner:
I don't agree that we should fix all Lighthouse warnings without considering context. The current font size is suitable for the purpose of the scale control, it's intended to be subtle.

Malvoz:
@mourner so keep `font-size: 11px` for both `.leaflet-control-scale` and `.leaflet-control-scale-line`?

PR #7349, opened 2020-11-24
Align overlay and shadow panes order

elfalem:
The descriptions for the two panes in the docs were switched. Also flipped the pane creation order to align with the z-index order.


IvanSanchez (approved):
Wow, nice catch, that was a subtle mistake in the docstrings.

PR #7352, opened 2020-11-26
Update reference-0.7.7.html

Dev-Steven:
Just thought 'options' should be capitalized like the other titles.


mourner (changes requested):
This should be fixed in the source [here](https://github.com/Leaflet/Leaflet/blob/master/src/control/Control.js#L18). The generated files shouldn't be updated on its own.

PR #7355, opened 2020-12-01
Added clarification on class properties

### LINKED ISSUE #7354: Using objects as class properties results in shared references to these objects between all instances of the class
Hello, so problem is basically described at the title. Probably, this is not a bug, but it should be documented. To illustrate this better, please, take a look at examples below.

This is how I expect my code to look like (explanations in comments):
```JS
let MyClass1 = L.Class.extend({
	// I want to have an object property, so I simply instantiate my object here as I do in any object-oriented language
	myObject: { myValue: 1 }, // myObject: new Object() causes this behavior too

	// And later I want to modify this object. Object should be modified only in one particular instance of MyClass1 when I call setObjectValue().
	setObjectValue: function (value) {
		this.myObject.myValue = value;
	}
});

let instance1 = new MyClass1();
let instance2 = new MyClass1();

instance1.setObjectValue(2); // Here I expect myObject to be modified only in instance1
console.log(instance1.myObject.myValue, instance2.myObject.myValue); // Expected: "2 1", outputs: "2 2"
// However, as you can see, myObject has been modified in both instances

instance2.setObjectValue(3); // Same goes for any other instance of myClass
console.log(instance1.myObject.myValue, instance2.myObject.myValue); // Outputs "3 3"
```

At this point we understand that `myObject` is being shared among our instances instead of being recreated at each instantiation. Let's instantiate `myObject` in constructor to ensure that it will be recreated.

```JS
let MyClass2 = L.Class.extend({
	initialize: function () {
		this.myObject = { myValue: 1 };
	},

	setObjectValue: function (value) {
		this.myObject.myValue = value;
	}
});

instance1 = new MyClass2();
instance2 = new MyClass2();

instance1.setObjectValue(2);
console.log(instance1.myObject.myValue, instance2.myObject.myValue); // Outputs "2 1" as expected

instance2.setObjectValue(3);
console.log(instance1.myObject.myValue, instance2.myObject.myValue); // Outputs "2 3" as expected
```

And it works just fine!

I don't know if this is intended behavior or not, it doesn't seem fine in terms of OOP *(but fine in terms of prototype-based programming if you'll happen to have insight on how Leaflet classes internally work)*, but I think it should be documented.

In my case, this simple misunderstanding led to hours of debugging my fairly complex application.

matafokka:
Instantiating an object in the class body will result in all instances of the class having a reference to this object instead. This commit clarifies this behavior.
fixes #7354

### DIFF
--- a/docs/examples/extending/extending-1-classes.md
+++ b/docs/examples/extending/extending-1-classes.md
@@ -70,6 +70,9 @@ When naming classes, methods and properties, adhere to the following conventions
 * Class names should be in [`UpperCamelCase`](https://en.wikipedia.org/wiki/CamelCase).
 * Private properties and methods start with an underscore (`_`). This doesn't make them private, just recommends developers not to use them directly.
 
+Usage notes:
+* Don't instantiate objects in the class body, do it in constructor. `L.Class.extend({ myProperty: new SomeClass() })` will result in `myProperty` being a reference to `new SomeClass()` in all instances because `L.Class.extend()` treats passed objects as-is and doesn't re-instantiate it's properties. Instead, write following in the constructor: `this.myProperty = new SomeClass();`.
+
 ### `L.Class.include()`    
     
 If a class is already defined, existing properties/methods can be redefined, or new ones can be added by using `.include()`:


mourner:
Thanks for the PR! However I feel like this does not need clarification in the example because it's not a behavior specific to `L.Class`, this is just how JavaScript works.

matafokka:
Hello, @mourner, thank you for reviewing my PR! However, I still think that clarification is needed. The subject has been discussed in my issue #7354 in which we established that described behavior needs to be clarified because:

> ... [Leaflet] classes are called classes and not prototypes.

At least, clarification will be very convenient because, as I said in that issue, one might expect that Leaflet will re-instantiate class properties.

PR #7364, opened 2020-12-04
Add `playsInline` attribute on the video element

### LINKED ISSUE #7357: Video overlay: `playsinline` attribute required for phones on iOS
https://webkit.org/blog/6784/new-video-policies-for-ios/:

> - On iPhone, `<video playsinline>` elements will now be allowed to play inline, and will not automatically enter fullscreen mode when playback begins.
> `<video>` elements without `playsinline` attributes will continue to require fullscreen mode for playback on iPhone.

Might we worth to add the attribute by default. See the HTML [`playsinline` attribute](https://html.spec.whatwg.org/multipage/media.html#attr-video-playsinline) definition.

donnisnoni:
Fixes #7357

### DIFF
--- a/src/layer/VideoOverlay.js
+++ b/src/layer/VideoOverlay.js
@@ -78,6 +78,7 @@ export var VideoOverlay = ImageOverlay.extend({
 		vid.autoplay = !!this.options.autoplay;
 		vid.loop = !!this.options.loop;
 		vid.muted = !!this.options.muted;
+		vid.playsInline = true;
 		for (var i = 0; i < this._url.length; i++) {
 			var source = DomUtil.create('source');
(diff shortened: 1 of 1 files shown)


IvanSanchez:
Just checking - has this been tested on an actual iOS device?

donnisnoni:
Not yet... but maybe @Malvoz can help to test it...

Malvoz:
@donnisnoni I don't own an iOS device (I just made the observation and reported on the potential requirement), sorry.

IvanSanchez:
Well, even though the code looks good (thanks @donnisnoni for sending the PR), I'd rather not merge until it has been tested (and I don't have a iOS device either)

PR #7366, opened 2020-12-05
Added supercluster

KristjanESPERANTO:
Added 'supercluster' before 'Leaflet.markercluster' because it seems that  'Leaflet.markercluster' isn't an active project


mourner:
Leaflet.markercluster is currently being resurrected: https://github.com/Leaflet/Leaflet.markercluster/issues/979
Since Supercluster isn't really a Leaflet plugin, and not a fully equivalent replacement, let's wait and have faith in the old plugin.

PR #7381, opened 2020-12-14
Replace shortened url with the destination

BakuCity:
As there's no need to econom characters, hide the destination or introduce extra analytics in the comments.


IvanSanchez (approved):
LGTM

Issue #7385, opened 2020-12-18
any3d misuse and other browser poking features


c-smile:
Name of this variable:
```var any3d = !window.L_DISABLE_3D && (ie3d || webkit3d || gecko3d) && !opera12 && !phantom;```
is misleading.

Lefalet is not using any 3D transformations actually as far as I can tell.

Instead ```any3d``` is used rather as ```hwaccelerated``` flag. But even in this case its use (as is-hw-accelerated flag) is not correct. In all browsers `translate(x,y)` and `translate3d(x,y,0)` use the same mechanism that is either H/W accelerated or is good enough for Leaflet purposes.

Please consider removal of browser feature testing using userAgent as [it is bad](https://css-tricks.com/browser-detection-is-bad/).

As far as I can tell Leaflet don't need anything more than [CSS.supports()](https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports) can provide.

IvanSanchez:
> Name of `var any3d` is misleading.

Not a reason good enough to introduce an API-breaking change. Plugins depends on this. Plus, it would impact fractional zoom support, iOS hacks, and a bunch of unit tests.

> Please consider removal of browser feature testing using userAgent as it is bad.

Won't do. Leaflet does feature detection, not user agent sniffing, as much as possible. Like, Leaflet is doing exactly what the article you linked to prompts devs to do.

(And yes we might consider dropping the remaining browser sniffing tests for Android v2-v3 at the same time we might drop IE8 support)

> As far as I can tell Leaflet don't need anything more than CSS.supports() can provide.

IE11 support, plus a bunch of weird engine-dependent stuff. My "favourite" example is https://github.com/Leaflet/Leaflet/blob/7db94fd1ec23f0967dd8b07a5e2e2b1554b6d8db/src/dom/DomEvent.DoubleTap.js#L62-L66

c-smile:
I am running Leaflet on [Sciter.JS](https://sciter.com/sciter-js-runs-leafletjs-as-it-is/) - Sciter.JS is an ElectronJS replacement and uses its own HTML/CSS/JS implementations.

Problems:

1. Sciter.JS does not support ```translate3d()```  but does support ```translate()```.    
2. Sciter.JS uses H/W acceleration on all platforms.
3. It has its own ```navigator.userAgent``` that is different from browsers . 

In any case Leaflet fails to detect ```translate3d()``` support properly. That should be done

in CSS as : 
```css
[...]

PR #7387, opened 2020-12-19
Future-proof L.Browser.any3d via CSS.supports(translate3d)

IvanSanchez:
As per #7385 - `L.Browser.any3d` should be `true` for any browser that supports `transform3D` CSS rules (except PhantomJS - we depend on that for the unit tests).

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -75,7 +75,7 @@ export var gecko3d = 'MozPerspective' in style;
 
 // @property any3d: Boolean
 // `true` for all browsers supporting CSS transforms.
-export var any3d = !window.L_DISABLE_3D && (ie3d || webkit3d || gecko3d) && !opera12 && !phantom;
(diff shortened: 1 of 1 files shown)


src/core/Browser.js
```diff
@@ -75,7 +75,7 @@ export var gecko3d = 'MozPerspective' in style;
 
 // @property any3d: Boolean
 // `true` for all browsers supporting CSS transforms.
-export var any3d = !window.L_DISABLE_3D && (ie3d || webkit3d || gecko3d) && !opera12 && !phantom;
+export var any3d = !window.L_DISABLE_3D && (ie3d || webkit3d || gecko3d || (CSS && CSS.supports('transform', 'translate3d(0,0,0)'))) && !opera12 && !phantom;
```
rheh, line 78: Is is worth removing phantom and operator from here as these are end-of-life (I think)

rheh (approved):

Falke-Design:
We will no longer suppert `any3d` with Leaflet v2

PR #7388, opened 2020-12-21
add method setPane(name) to layers

Falke-Design:
I've added a new method `setPane(name)`. It moves a layer to another pane without removing it and adding it to the map again. So no events are fired and the layer is not reinitialized.

Additionally the layer is now also moved to another pane when the pane is changed in the options over `setStyle({pane: 'xyzPane'})`

I tried a lot around also with canvas and there were no problems, I hope there is no "hidden" feature that I forgot.


src/layer/Layer.js
```diff
@@ -78,6 +79,27 @@ export var Layer = Evented.extend({
...
+			DomUtil.remove(this._icon);
+			this.options.pane = name;
+			this.getPane().appendChild(this._icon);
+		} else {
+			this._map.getRenderer(this)._removePath(this);
+			this.options.pane = name;
+			this._renderer = this._map.getRenderer(this);
+			this.onAdd();
+		}
+		return this;
```
johnd0e, line 100: Shouldn't we first check if the pane is actually differs from current?
Extra DOM operations may appear costly, e.g. on mass `setStyle`.

Falke-Design: I had this implemented but there was a problem with `setStyle`. In the `setStyle` function it overrides the options (and the new pane) with `Util.setOptions(this, style);` and then I was not able to check anymore if the options pane is different to the new pane.

Maybe I can check over the html parent which is the current pane, if you want this.

johnd0e: > it overrides the options

You may choose to set pane before overwriting, for example.

> if you want this.

Well, you better do what you want.

In general: I suppose that this PR makes sense, but I cannot review it thoroughly atm.

src/layer/Layer.js
```diff
@@ -78,6 +79,27 @@ export var Layer = Evented.extend({
...
+			return this;
+		}
+
+		if (this instanceof L.Marker) {
+			DomUtil.remove(this._icon);
+			this.options.pane = name;
+			this.getPane().appendChild(this._icon);
+		} else {
+			this._map.getRenderer(this)._removePath(this);
+			this.options.pane = name;
```
rheh, line 96: nit-pick:  The above line is included in every branch of the function.  Could you call it once and remove duplication? e.g.

```
setPane: function (name) {
		this.options.pane = name;

		if (!this._map) {
			return this;
		}

		if (this instanceof L.Marker) {
			DomUtil.remove(this._icon);
			this.getPane().appendChild(this._icon);
		} else {
			this._map.getRenderer(this)._removePath(this);
			this._renderer = this._map.getRenderer(this);
			this.onAdd();
		}
		return this;
	},
```

Falke-Design: Yes this is not looking well, but i have not found another solution as this, because `options.pane` have to be the old before `this._map.getRenderer(this)._removePath(this);` is called, else it is not removed clean.

johnd0e:
> Additionally the layer is now also moved to another pane when the pane is changed in the options over `setStyle({pane: 'xyzPane'})`

`pane` in not style option, imho.
@mourner ?

johnd0e:
> It moves a layer to another pane without removing it and adding it to the map again.

Sounds good, but could you give me more context about presumable use cases?

Falke-Design:
> pane in not style option, imho: https://leafletjs.com/reference.html#path-setstyle

I added it to the `setStyle` function because while creation it is passed to the options too. And I would read the docu like `pane` options can passed to the `setStyle` too, because of the describtion `Changes the appearance of a Path based on the options in the Path options object.` If you expand the `Options inherited from Layer` part you see there is the option `pane`. 

> give me more context about presumable use cases

When a plugin like a draw plugin creates a layer, users can listen on the create event and then change the layer to the wanted pane.

johnd0e:
> added it to the `setStyle` function because while creation it is passed to the options too.

You can call `setPane` in constructor explicitly.


> If you expand the `Options inherited from Layer` part you see there is the option `pane`.

Does it mean that it is related to _appearance_? I don't think so.
And Layer class still does not have `setStyle` method.

Falke-Design:
I have no problem with removing the `setStyle` part, this was only the explanation why I added it

Falke-Design:
@johnd0e I removed the pane from `setStyle` and added a check if the pane differs from curent https://github.com/Leaflet/Leaflet/pull/7388#discussion_r547132179

src/layer/Layer.js
```diff
@@ -78,6 +79,29 @@ export var Layer = Evented.extend({
...
+			this._renderer = this._map.getRenderer(this);
+			this.onAdd();
+		}
```
johnd0e, line 101: What about L.DivIcon?
What about zillions of custom layers?

PR #7390, opened 2020-12-24
New plugin: Add Leaflet.Mask

ptma:
Hi,
    Leaflet.Mask is a plugin that loads polygons from geojson to mask the rest of the map.

   THX.


johnd0e:
@ptma Have you closed this intentionally?

Issue #7372, opened 2020-12-08
map.panInside() does not pan along y axis when "padding" option is set


t3539:
- [ ] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [ ] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [ ] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
Move the view across and down (to somewhere like India or Saudi Arabia in my example below). Click on the map. It only pans in the x direction.

**Expected behavior**
Pan along both x axis and y axis

**Current behavior**
Only pans along x axis

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Firefox 83.0
- OS/Platform (with version): Ubuntu

**Minimal example reproducing the issue**

https://plnkr.co/edit/yBVCWWHuPqRywn9z

- [ ] this example is as simple as possible
- [ ] this example does not rely on any third party code

daverayment:
Thanks for reporting this, @t3539 . I'm the contributor who originally wrote this function and the associated tests. I've confirmed that when the padding option is provided, the map does not correctly pan the y axis. I will attempt to look at this soon and submit a fix.

daverayment:
Hi @t3539. The parameter is a `Point`, so you need to provide both x and y values. For example:

`map.panInside(latlng, {padding: [500, 500]});`

Supplying only one will set the x padding value and leave the y not set, which is why you're experiencing issues.

Padding for `panInside` behaves like the `FitBounds` option, which is described here (this is also linked to from the `panInside` documentation):

https://leafletjs.com/reference-1.7.1.html#fitbounds-options

daverayment:
@t3539 This isn't a Leaflet bug, so could you close this issue, please?

johnd0e:
@t3539
I have objections as current implementation is error prone.
- In some (though unrelated) parts of documentation we have `padding` as not point: https://leafletjs.com/reference-1.7.1.html#renderer-padding
- [`panInside`](https://leafletjs.com/reference-1.7.1.html#map-paninside) docs mention just `options` without any link.
Perhaps we should link [`FitBounds options`](https://leafletjs.com/reference-1.7.1.html#fitbounds-options) (not sure as `maxZoom` seems not applicable).

So I propose:
1) Make documentation easier
2) Secure the function making sure that passed arguments are properly typed.

[...]

Issue #7128, opened 2020-05-06
Functions should throw exception on invalid args


johnd0e:
In general I believe that it's better when functions called with invalid argument throw early, because such approach leaves less space for bugs.

Question is how far should we go implementing this, considering that we have no builtin facilities like languages like TypeScript have.

For example see #7125.
- [x] I'm sure that more strict conditions (introduced there) are good.
There is no explicit type checking, but code is composed with type-checking in mind, and intentionally fails on many cases of invalid args.
- [ ] But should we put that to the test (like done there)? I have some doubts, because that "fails" are not in specs, so they are specific to current implementation.
Anyway, it should not be a problem to fix (or even remove) tests when we will need that.
- [ ] Still things are not consistent now: not all cases of invalid args are covered. E.g. `DomEvent.off` doesn't fail if pass invalid object instead of `el`. And (in other functions) `event` argument is not validated at all.
All such cases could be covered with explicit type checking, but I suppose that would be overkill.

Thoughts?

johnd0e:
After discussion #7397 we've found that [`toPoint`](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/geometry/Point.js#L208) is error prone, as it allows to create [Point](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/geometry/Point.js#L27) with `undefined` coordinates.
Particularly important case is when point is create indirectly, by some option, which can be wrongly specified as plain value like `50`, instead of array `[50, 50]`.

daverayment:
When a single non-point and non-object argument is passed to `toPoint`, it will end up on [the last line of that function](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/geometry/Point.js#L221), which calls the `Point` constructor, which is expecting a `y` parameter (and optionally a `round` value).

	return new Point(x, y, round);

In our problem case of a single number being passed, the `Point` [constructor](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/geometry/Point.js#L27) will therefore end up setting `y` to `undefined`:

	this.y = (round ? Math.round(y) : y);

A check in `toPoint` for the existence of a `y` value would seem a good first step then. What do you think? For example:

```
if (y === undefined || y === null) {
[...]

PR #7397, opened 2021-01-03
Improve panInside documentation

daverayment:
This is a small fix to resolve issue #7372. It includes an improvement to the documentation to ensure users are aware the padding parameters are `Point` types and it directly links to the `fitBounds` options section which describes each option in more detail.

In addition, there is a new check at the start of the function which sets both X and Y padding correctly if only one value has been supplied; this is a friendlier alternative to raising an error in the console if a `Point` was mistakenly not given.


johnd0e:
> this is a friendlier alternative to raising an error

I believe this is error prone approach.
Better either allow and document that officially, or forbid that and throw error.

daverayment:
@johnd0e Thanks. I'm now throwing an error if the padding is invalid, and have added a test to exercise this.

johnd0e:
Looks good.
But may be we should also enhance docs, introducing dedicated "Padding options" section (instead of [FitBounds options](https://leafletjs.com/reference-1.7.1.html#fitbounds-options))?

daverayment:
@johnd0e I agree. Having `fitBounds options` inherit from a new `padding options` is cleaner.

I have made these changes to the documentation in the latest commit, and also added error checking and tests for `topLeftPadding` and `bottomRightPadding` params.

src/map/Map.js
```diff
@@ -511,17 +511,18 @@ export var Map = Evented.extend({
...
 	},
 
-	// @method panInside(latlng: LatLng, options?: options): this
+	// @method panInside(latlng: LatLng, options?: padding options): this
 	// Pans the map the minimum amount to make the `latlng` visible. Use
-	// `padding`, `paddingTopLeft` and `paddingTopRight` options
-	// ([Point](#point) types) to fit the display to more restricted bounds,
-	// like the [`padding options for fitBounds`](#fitbounds-options).
+	// `padding`, `paddingTopLeft` and `paddingBottomRight` options
+	// ([Point](#point) types) to fit the display to more restricted bounds.
```
johnd0e, line 517:
```suggestion
	// padding options to fit the display to more restricted bounds.
```

johnd0e:
> Having `fitBounds options` inherit from a new `padding options` is cleaner.

I like it!
But that also makes clear that `fitBounds` also requires similar handling (and tests), right?

daverayment:
> But that also makes clear that `fitBounds` also requires similar handling (and tests), right?

Good point! Since `fitBounds` and `flyToBounds` have optional padding, too, I've added a check to the `_getBoundsCenterZoom` function, which they both call.

I've also added tests for both methods to exercise this check. There were no pre-existing tests for `flyToBounds`, so I started a new `describe` section for that in MapSpec.js. I hope that's OK.

One potential concern: this could throw errors for existing projects because we're being stricter with the parameter checking for these 3 functions. (However, affected projects would not have been behaving correctly.) What are your thoughts?

johnd0e:
> There were no pre-existing tests for `flyToBounds`, so I started a new `describe` section for that in MapSpec.js. I hope that's OK.

Sure that's ok.
But! This is not enough in general, as `flyToBounds` needs more specific tests.
Perhaps this is out of scope of this PR, but you should at least add some TODO note somewhere.

It's also possible to gather all related tests in one `describe` block.
And maybe we should be more specific about error, checking `.to.throw('Padding must be a point type.');`
<details>

```js
	describe('#padding options', function () {
		beforeEach(function () {
			//...
		});
		var paddingError = 'Padding must be a point type.';

		['padding', 'paddingTopLeft', 'paddingBottomRight'].forEach(function (prop) {
			var paddingOptions = {[prop]: 50};

			it('throws an error with invalid ' + prop, function () {
				expect(function () {
					map.flyToBounds(bounds, paddingOptions);
				}).to.throwError(paddingError);

				expect(function () {
					map.fitBounds//...
				}).to.throwError(paddingError);

				expect(function () {
					map.panInside//...
				}).to.throwError(paddingError);
			});
		});
	});
```
</details>


> One potential concern: this could throw errors for existing projects because we're being stricter with the parameter checking

Yeah, after upgrading maintainers should fix bugs in their code, it's easy because they will know exact cause.

daverayment:
Thanks, @johnd0e. I like your approach - it keeps all the checks in one place in MapSpec.js and can easily be expanded upon if other methods have padding options in the future. I've implemented this suggestion and removed the previous tests from elsewhere (including the new `flyToBounds` block). If further work on `flyToBounds` testing is required in the future, I agree that that's probably a separate piece of work.

spec/suites/map/MapSpec.js
```diff
@@ -791,6 +791,36 @@ describe("Map", function () {
 		});
 	});
 
+	describe('#padding options types', function () {
+		beforeEach(function () {
```
johnd0e, line 795: Perhaps we may use `before`, to execute this block just once before all included tests.

spec/suites/map/MapSpec.js
```diff
@@ -791,6 +791,36 @@ describe("Map", function () {
...
+			container.style.width = container.style.height = '100px';
+			map.setView([50.5, 30.51], 15);
+		});
+
+		var bounds = L.latLngBounds(L.latLng(1, 102), L.latLng(11, 122));
+		var paddingError = 'Padding must be a point type.';
+
+		['padding', 'paddingTopLeft', 'paddingBottomRight'].forEach(function (prop) {
+			var paddingOptions = {};
+			paddingOptions[prop] = 50; // invalid
```
johnd0e, line 806: In fact there is a lot of other invalid values possible, like `[50]` or `"[50]"`. Not sure that we should check over all of them.
But if we not - then better refine the comment to something like "// most common misuse"...

src/map/Map.js
```diff
@@ -252,6 +252,13 @@ export var Map = Evented.extend({
 	_getBoundsCenterZoom: function (bounds, options) {
 
 		options = options || {};
+
+		if ((options.padding && options.padding.length !== 2) ||
```
johnd0e, line 256: Invalid value like '50' will pass this test too.

src/map/Map.js
```diff
@@ -511,15 +518,20 @@ export var Map = Evented.extend({
...
 	// Pans the map the minimum amount to make the `latlng` visible. Use
-	// `padding`, `paddingTopLeft` and `paddingTopRight` options to fit
-	// the display to more restricted bounds, like [`fitBounds`](#map-fitbounds).
+	// padding options to fit the display to more restricted bounds.
 	// If `latlng` is already within the (optionally padded) display bounds,
 	// the map will not be panned.
 	panInside: function (latlng, options) {
 		options = options || {};
 
+		if ((options.padding && options.padding.length !== 2) ||
```
johnd0e, line 529: Another occurrence of same code. Not good...

src/map/Map.js
```diff
@@ -511,15 +518,20 @@ export var Map = Evented.extend({
...
+			(options.paddingTopLeft && options.paddingTopLeft.length !== 2) ||
+			(options.paddingBottomRight && options.paddingBottomRight.length !== 2)) {
+			throw new Error('Padding must be a point type.');
+		}
+
 		var paddingTL = toPoint(options.paddingTopLeft || options.padding || [0, 0]),
```
johnd0e, line 535: The best solution might be including those checks into `toPoint` function.
What do you think?

johnd0e: As for me - I believe that it would be most clean solution, with much more positive impact, because it is able to catch Point misuse in every place, not only in padding options.
But I have some concerns about (1) compatibility and (2) performance.
1. Perhaps not an issue. Who would want to call `toPoint(50)` intentionally?
2. We need to make sure that `toPoint` is not used in any hot path.

spec/suites/map/MapSpec.js
daverayment: I think this needs to be a `beforeEach`, like the other tests, because of the scope of the map object. It doesn't have any measurable performance impact, so I think we should keep it like this to be consistent.

spec/suites/map/MapSpec.js
daverayment: The intention is not to test for all possible invalid values, so I'm happy to update the comment.

src/map/Map.js
daverayment: This is just to catch cases where a single numeric value is passed in, as it was to deal with the original #7372 issue. Yes, a two-character string will be classed as valid, but I think that's beyond the scope of this simple fix.

src/map/Map.js
daverayment: Is the issue that I haven't factored out the error check into a separate fn? I can certainly do that if we decide these checks are enough for now.

src/map/Map.js
daverayment: I thought about this a few days ago, but I discounted it because I've used `toPoint` in the past on quite large datasets and I was worried the additional checks could affect performance in such cases. I think putting validation on padding options and improving the documentation as we've done here is a good start, and maybe `toPoint` can be revisited in the future.

src/map/Map.js
johnd0e: > if we decide these checks are enough for now.

I tend to think that current checks definitely not enough.
Why should we care padding options more that others?
And even if we do, why we do not care about similar occurrences:
- https://leafletjs.com/reference-1.7.1.html#marker-autopanpadding
- https://leafletjs.com/reference-1.7.1.html#popup-autopanpaddingtopleft (there are 3)
- etc

So I propose following:
1. Exclude validation code (and corresponding testes) from this PR, leave docs enhancement only (which alone solves most of original #7372 issue).
2. Make another PR adding point object validation code to `toPoint`.

daverayment: The other padding parameters you mention were explicitly documented as taking `Point`s. The `panInside` documentation originally lacked this explicit mention, which contributed to the original misunderstanding by the OP and the #7397 issue being raised. I suspect adding `toPoint` validation may end up catching fewer genuine issues while potentially adding a performance penalty for larger projects.

However, we can open that up for discussion on a separate PR! I'm happy to remove the options checks here and submit just the documentation updates for now.

johnd0e (approved):

src/map/Map.js
johnd0e: > I discounted it because I've used `toPoint` in the past on quite large datasets and I was worried the additional checks could affect performance in such cases.

Still, I believe that it is very important - to catch bugs early, so I've added this case to this discussion https://github.com/Leaflet/Leaflet/issues/7128#issuecomment-760765116

johnd0e (changes requested):
@daverayment 
Could you please resolve merge conflicts?

johnd0e:
@daverayment ?

mourner (approved):

PR #7405, opened 2021-01-11
Added 'FeatureCollection' case to geoJSON decoder

EbashuOnHolidays:
Hi everyone. I'm trying to convert the featureGroup object to geoJSON and get the layer again from geoJSON, but it doesn't work:

```js
this.map.on('draw:created', geometryCreate)

geometryCreate(event) {
  this.featureOverlay.addLayer(event.layer)

  let geoJSON = this.featureOverlay.toGeoJSON()

  console.log('geoJSON', geoJSON)

  let layer = L.GeoJSON.geometryToLayer(geoJSON)
}
```
I get this error:

image

The problem is that the FeatureCollection object is not supported.

```js
case 'FeatureCollection':
    for (i = 0, len = geometry.features.length; i < len; i++) {
	    var layer = geometryToLayer({
		    geometry: geometry.features[i].geometry,
		    type: geometry.features[i].type,
		    properties: geojson.properties
	    }, options);
    
	    if (layer) {
		    layers.push(layer);
	    }
    }
    return new FeatureGroup(layers)
```

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -202,6 +202,20 @@ export function geometryToLayer(geojson, options) {
 	case 'MultiPolygon':
 		latlngs = coordsToLatLngs(coords, geometry.type === 'Polygon' ? 1 : 2, _coordsToLatLng);
 		return new Polygon(latlngs, options);
+			
+	case 'FeatureCollection':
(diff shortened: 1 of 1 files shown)


IvanSanchez:
You're confusing GeoJSON *features* with GeoJSON *geometries*. This is behaviour as expected.

PR #7406, opened 2021-01-11
A few improvements in the Docs and on DivOverlay, Tooltip and Popup

Falke-Design:
**Improvements in the Docs: [Demo](http://falke-design.bplaced.net/leaflet/dist/#popup)**
- DivOverlay is now on the right place in the section `UI Layers`, before it was not added in the menu and placed at the bottom below `Events` [DivOverlay](https://leafletjs.com/reference-1.7.1.html#divoverlay)
- Popup and Tooltip references now to the functions of DivOverlay. Changed also the description Fixes: #6954, #6597
- Now it is possible to create anchor links from headers, when clicking on the link icon
- Anchor links they are in a collapsed accordion are now expanding the accordion if they are requested (tested in Firefox, Chrome, Edge, IE) [#divoverlay-bindpopup](http://falke-design.bplaced.net/leaflet/dist/#divoverlay-bindpopup)

**Added Features:**
- Fires on the popup self the events `popupopen` and `popupclose`
- DivOverlay (Popup & Tooltip) supports now  `content` in the options and as second parameter `latlng`. Very useful for standalone Popups or Tooltips, because the expected behaviour is to pass them on initializing.
Before:
```
var popup = L.popup()
    .setLatLng(latlng)
    .setContent('<p>Hello world!<br />This is a nice popup.</p>')
    .openOn(map);
```
Now:
```
var popup = L.popup({content: '<p>Hello world!<br />This is a nice popup.</p>'},latlng).addTo(map);
```

**Fixed Bugs:**
- Tooltip RTL support: was possible over css changes, explained [here](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-713112492). New function to the map is added `map.setRTLDirection(boolean)`, it adds the CSS-Class `leaflet-rtl` to the map container . During initialization the style `direction: rtl` is searched on the map container and if is found `setRTLDirection(true)` is called. Fix: #7201
- ~~If Tooltip is Sticky and Permanent it is now following mouse [comment in #7201](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-655746354)~~ PR: #7563 

**Additional thoughts:**
- DivOverlay has a default `offset: [0,7]` I think this would be better if this is moved to Popup and changed to [0,0] but this is a breaking change ...
- `.setTooltipContent("")` / `.setPopupContent("");` and  `.setContent("")` confuses developers, maybe this can explained better but I don't know how.

Restructured commits of PR: #7398

Issue #995, opened 2012-09-13
Leaflet line ending style is mixed


lapo-luchini:
OK, this is really really minor.
But still… ;-)
Different Leaflet sources use a different line ending style (src/map/Map.js uses CRLF, src/map/handler/Map.TouchZoom.js LF), so the generated -src.js script is mixed.
(Of course the minified isn't anways.)

12finger:
it's 2021 and i am experiencing the same still.
i forked and wanted to work on some code amends for my own project and later maybe roll them up to a PR 

Popup.js is CRLF
while 
SVGOverlay is LF

PR #7398, opened 2021-01-05
A few improvements in the Docs and on DivOverlay, Tooltip and Popup

Falke-Design:
**Improvements in the Docs: [Demo](http://falke-design.bplaced.net/leaflet/dist/#popup)**
- DivOverlay is now on the right place in the section `UI Layers`, before it was not added in the menu and placed at the bottom below `Events` [DivOverlay](https://leafletjs.com/reference-1.7.1.html#divoverlay)
- Popup and Tooltip references now to the functions of DivOverlay. Changed also the description Fixes: #6954, #6597
- Now it is possible to create anchor links from headers, when clicking on the link icon
- Anchor links they are in a collapsed accordion are now expanding the accordion if they are requested (tested in Firefox, Chrome, Edge, IE) [#divoverlay-bindpopup](http://falke-design.bplaced.net/leaflet/dist/#divoverlay-bindpopup)

**Added Features:**
- Fires on the popup self the events `popupopen` and `popupclose`
- DivOverlay (Popup & Tooltip) supports now  `content` in the options and as second parameter `latlng`. Very useful for standalone Popups or Tooltips, because the expected behaviour is to pass them on initializing.
Before:
```
var popup = L.popup()
    .setLatLng(latlng)
    .setContent('<p>Hello world!<br />This is a nice popup.</p>')
    .openOn(map);
```
Now:
```
var popup = L.popup({content: '<p>Hello world!<br />This is a nice popup.</p>'},latlng).addTo(map);
```

**Fixed Bugs:**
- Tooltip RTL support: was possible over css changes, explained [here](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-713112492). New function to the map is added `map.setRTLDirection(boolean)`, it adds the CSS-Class `leaflet-rtl` to the map container . During initialization the style `direction: rtl` is searched on the map container and if is found `setRTLDirection(true)` is called. Fix: #7201
- If Tooltip is Sticky and Permanent it is now following mouse [comment in #7201](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-655746354)

**Additional thoughts:**
- DivOverlay has a default `offset: [0,7]` I think this would be better if this is moved to Popup and changed to [0,0] but this is a breaking change ...
- `.setTooltipContent("")` / `.setPopupContent("");` and  `.setContent("")` confuses developers, maybe this can explained better but I don't know how.

[...]

PR #7409, opened 2021-01-14
Change all files to LF line break

Falke-Design:
A Issue from a long time ago: #995
And I had same problem #7406

With this PR all files are changed to the `LF` line break / ending

**How:**
Enable auto set line break
```
git config core.autocrlf true
```
Creating a clean branch:
```
git checkout --orphan master_line_break
```
Remove all existing files:
```
git reset --hard
```
Get a fresh pull of the master:
```
git pull https://github.com/Leaflet/Leaflet.git master
```
And then renormalize all files:
```
git add --renormalize .
```

And then after a refresh in my IDE I was able to commit all files with the line break `LF`

**Problems:**
- I think a problem would be that all existing PRs showing the complete file as difference
- If this is not merged before a new commit is made in the master, maybe the new lines has the wrong line break. I suggest that a Admin creates a own PR and commit them immediately


johnd0e:
> Problems:

Also this may interfere with `git blame` functionality.

mourner:
I think we shouldn't do this — the minor benefit of having consistent line endings doesn't justify breaking all PR diffs, `git blame`, introducing possible merge conflicts etc.

Falke-Design:
Yes I see it like you, but then I recommand to create an introduction how the files can be commited with the correct line ending. I'm not an expert but it worked for me with `git config core.autocrlf false`.

Also you shouldn't be too strict with PRs when they have struggles with the endings ... 
For example my PR #7406. After a lot of trying around because of the endings I created the PR #7398 with only one summarized commit, that was bad, I understand this. But then I created a new one #7406 and now you converted it to a draft because I had again a problem with the endings with a single file at one commit.

I understand that clean commits are important to understand the changes but the line ending problems are comming from Leaflet self and a contributor needs to invest time to find out which file has which line ending. I have to say after 2 hours of fighting with the line endings I don't want to invest more time again to create a new PR or update the draft PR.

mourner:
@Falke-Design hey, I'm very sorry about this — we'll try to improve this going forward! Also, don't take converting to a draft as criticism — it's just a maintainer tool to remind us when a PR needs updates or decisions to be made, whether on the author's side or on ours. Additionally, it's very stressful to maintain popular projects like this, especially in our free time and during the pandemic, so being strict helps maintainers avoid burn out.

PR #7418, opened 2021-01-20
Add toPoint validation

daverayment:
Before creating a new `Point`, check the Y coordinate has been provided. This catches issues where callers only provide a single value by mistake, e.g. when setting a padding value for one of the map functions (see #7372).


johnd0e:
I have to refer to more general issue: https://github.com/Leaflet/Leaflet/issues/7128#issuecomment-761544732.

@daverayment 
Your PR fixes some particular cases, leaving all others unaffected.
For example see #7128.

So some more general solution would be preferable.

PR #7432, opened 2021-01-25
L.Point parse string to number, because polygon disapears sometimes

Falke-Design:
This fixes a bug that is very hard to reproduce.

Polygons disappearing sometimes while zooming and moving.

The OP from SO found out that it only happens when the weight of the polygon is set as a string: `{weight: "10"}` instead of `{weight: 10}`

Then I looked in the code and I think I found the problem in the function `_clipPoints` at line 126 and 129.
The weight is a string and the point is create out of the weight value --> `L.Point("10","10")`
And then it is added and subtracted to the bounds. And we all know that in JS when we add a string to a number it retuns a string: `3 + "10" = "310"` and not `13`. You can see this while debugging, look into the bounds value before and after adding the "weight" point.
https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/layer/vector/Polygon.js#L121-L130

SO: https://stackoverflow.com/q/65829340/8283938
Video from SO: 

https://user-images.githubusercontent.com/19800037/105766896-99fb2300-5f5a-11eb-9f57-5b74f8c18b20.mp4

### DIFF
--- a/src/geometry/Point.js
+++ b/src/geometry/Point.js
@@ -26,9 +26,10 @@ import {isArray, formatNum} from '../core/Util';
 
 export function Point(x, y, round) {
 	// @property x: Number; The `x` coordinate of the point
-	this.x = (round ? Math.round(x) : x);
+	this.x = (round ? Math.round(x) : parseFloat(x));
 	// @property y: Number; The `y` coordinate of the point
-	this.y = (round ? Math.round(y) : y);
+	this.y = (round ? Math.round(y) : parseFloat(y));
+
 }
 
 var trunc = Math.trunc || function (v) {
@@ -163,7 +164,7 @@ Point.prototype = {
 		point = toPoint(point);
 
 		var x = point.x - this.x,
-		    y = point.y - this.y;
+		y = point.y - this.y;
 
 		return Math.sqrt(x * x + y * y);
 	},
@@ -174,7 +175,7 @@ Point.prototype = {
 		point = toPoint(point);
 
 		return point.x === this.x &&
-		       point.y === this.y;
+			point.y === this.y;
 	},
 
 	// @method contains(otherPoint: Point): Boolean
@@ -183,15 +184,15 @@ Point.prototype = {
 		point = toPoint(point);
 
 		return Math.abs(point.x) <= Math.abs(this.x) &&
-		       Math.abs(point.y) <= Math.abs(this.y);
+			Math.abs(point.y) <= Math.abs(this.y);
 	},
 
 	// @method toString(): String
 	// Returns a string representation of the point for debugging purposes.
 	toString: function () {
 		return 'Point(' +
-		        formatNum(this.x) + ', ' +
-		        formatNum(this.y) + ')';
+			formatNum(this.x) + ', ' +
+			formatNum(this.y) + ')';
 	}
 };
 
@@ -210,13 +211,13 @@ export function toPoint(x, y, round) {
 		return x;
 	}
 	if (isArray(x)) {
-		return new Point(x[0], x[1]);
+		return new Point(parseFloat(x[0]), parseFloat(x[1]));
 	}
 	if (x === undefined || x === null) {
 		return x;
 	}
 	if (typeof x === 'object' && 'x' in x && 'y' in x) {
-		return new Point(x.x, x.y);
+		return new Point(parseFloat(x.x), parseFloat(x.y));
 	}
-	return new Point(x, y, round);
+	return new Point(parseFloat(x), parseFloat(y), round);
 }


src/geometry/Point.js
```diff
@@ -26,9 +26,10 @@ import {isArray, formatNum} from '../core/Util';
 
 export function Point(x, y, round) {
 	// @property x: Number; The `x` coordinate of the point
-	this.x = (round ? Math.round(x) : x);
+	this.x = (round ? Math.round(x) : parseFloat(x));
 	// @property y: Number; The `y` coordinate of the point
-	this.y = (round ? Math.round(y) : y);
+	this.y = (round ? Math.round(y) : parseFloat(y));
+
```
johnd0e, line 32: Avoid whitespace-only changes.

src/geometry/Point.js
```diff
@@ -163,7 +164,7 @@ Point.prototype = {
 		point = toPoint(point);
 
 		var x = point.x - this.x,
-		    y = point.y - this.y;
+		y = point.y - this.y;
```
johnd0e, line 167: Wrong

src/geometry/Point.js
```diff
@@ -174,7 +175,7 @@ Point.prototype = {
 		point = toPoint(point);
 
 		return point.x === this.x &&
-		       point.y === this.y;
+			point.y === this.y;
```
johnd0e, line 178: Wrong

src/geometry/Point.js
```diff
@@ -183,15 +184,15 @@ Point.prototype = {
 		point = toPoint(point);
 
 		return Math.abs(point.x) <= Math.abs(this.x) &&
-		       Math.abs(point.y) <= Math.abs(this.y);
+			Math.abs(point.y) <= Math.abs(this.y);
```
johnd0e, line 187: Wrong

src/geometry/Point.js
```diff
@@ -183,15 +184,15 @@ Point.prototype = {
...
 	},
 
 	// @method toString(): String
 	// Returns a string representation of the point for debugging purposes.
 	toString: function () {
 		return 'Point(' +
-		        formatNum(this.x) + ', ' +
-		        formatNum(this.y) + ')';
+			formatNum(this.x) + ', ' +
+			formatNum(this.y) + ')';
```
johnd0e, line 195: Wrong

src/geometry/Point.js
```diff
@@ -210,13 +211,13 @@ export function toPoint(x, y, round) {
 		return x;
 	}
 	if (isArray(x)) {
-		return new Point(x[0], x[1]);
+		return new Point(parseFloat(x[0]), parseFloat(x[1]));
```
johnd0e, line 214: Excessive, as you anyway will parse it in constructor.

src/geometry/Point.js
```diff
@@ -210,13 +211,13 @@ export function toPoint(x, y, round) {
...
 	}
 	if (x === undefined || x === null) {
 		return x;
 	}
 	if (typeof x === 'object' && 'x' in x && 'y' in x) {
-		return new Point(x.x, x.y);
+		return new Point(parseFloat(x.x), parseFloat(x.y));
 	}
-	return new Point(x, y, round);
+	return new Point(parseFloat(x), parseFloat(y), round);
```
johnd0e, line 222: Same

src/geometry/Point.js
```diff
@@ -26,9 +26,10 @@ import {isArray, formatNum} from '../core/Util';
 
 export function Point(x, y, round) {
 	// @property x: Number; The `x` coordinate of the point
-	this.x = (round ? Math.round(x) : x);
+	this.x = (round ? Math.round(x) : parseFloat(x));
```
johnd0e, line 29: Good!
But I worry about this case: `parseFloat("1x") === 1`.

johnd0e (changes requested):

src/geometry/Point.js
Falke-Design: What should it do, when `"1x"` is passed? An error?

johnd0e: Well, maybe just `NaN`, I am not sure.
See more general issue: https://github.com/Leaflet/Leaflet/issues/7128

Falke-Design: It looks like that Leaflet today don't validate inputs / values, maybe in future but I suggest that this should not be a part of this PR. 
In my eyes the value "1x" is invalid so why not parse it to "1"? Input is not a number as in the doku defined, so it can happens that an unexpected result comes out.

johnd0e: > In my eyes the value "1x" is invalid so why not parse it to "1"?

Invalid values should be rejected, otherwise it may lead to hard-to-find bugs.

> but I suggest that this should not be a part of this PR.

Well, may be.

Falke-Design: > Invalid values should be rejected, otherwise it may lead to hard-to-find bugs.

Ok yeah this can be a problem

What do you think about:
```js
export function Point(x, y, round) {
        x = isNaN(x) ? x : parseFloat(x);
        y = isNaN(y) ? y : parseFloat(y);
	// @property x: Number; The `x` coordinate of the point
	this.x = (round ? Math.round(x) : x);
	// @property y: Number; The `y` coordinate of the point
	this.y = (round ? Math.round(y) : y);
}
```

johnd0e: `new Point('five', 'six').add([0,0]) === {x: 'five0', y: 'six0'}`?

Falke-Design: Ok let it change to `x = isNaN(x) ? NaN : parseFloat(x);` then it always returns `NaN` also after called `add([0,0])`

johnd0e: May be better to throw explicitly, like here:
https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/geo/LatLng.js#L30-L33

Falke-Design: Yes this sounds better, I updated it

src/geometry/Point.js
```diff
@@ -25,10 +25,13 @@ import {isArray, formatNum} from '../core/Util';
...
 export function Point(x, y, round) {
+	if (isNaN(x) || isNaN(y)) {
+		throw new Error('Invalid Point object: (' + x + ', ' + y + ')');
+	}
 	// @property x: Number; The `x` coordinate of the point
-	this.x = (round ? Math.round(x) : x);
+	this.x = (round ? Math.round(x) : parseFloat(x));
 	// @property y: Number; The `y` coordinate of the point
-	this.y = (round ? Math.round(y) : y);
+	this.y = (round ? Math.round(y) : parseFloat(y));
```
johnd0e, line 34: I also worry about values like `'0x10'` that will be converted to `0`.
And it's not clear what should we do with "Infinity".

See this table:
image
(from https://stackoverflow.com/questions/17106681/parseint-vs-unary-plus-when-to-use-which)

johnd0e (changes requested):

daverayment:
Wouldn't it be better instead to raise an error if a string is passed in for one or more of the arguments? That would seem to be the root cause of the issue.

The current version of the documentation has the Polygon's weight parameter as a Number type (inherited from `Path`), so I don't think there's inconsistency there. See: https://leafletjs.com/reference-1.7.1.html#polygon-weight

I think it's more robust to be strict about accepted types and raising errors rather than trying to coerce values.

johnd0e:
> I think it's more robust to be strict about accepted types

That may break a lot of 3rd-party code that was working without any problem before...
So I'm not sure.

Falke-Design:
@johnd0e I think with the `isNaN()` we have a good start for validation but I think also that the input validation of L.Point and other Classes should be a part of another issue / PR. 

Today there is a bug in leaflet and this PR will fix the bug but when we discuss here what we can / should do but we don't find out what is the right way, the bug will not be fixed.

So my suggestion is that we don't implement validation in this PR but we create a new issue "todo" for this.
I understand that you want that when something like this is edited it should be a clean way. But is it better to don't fix a bug because we don't know, how we should do something, what is not a part of this PR?

johnd0e:
> but we don't find out what is the right way, 

I was in believe that it wouldn't be a big issue to find proper way..

> the bug will not be fixed.

Well, Leaflet contain a lot of bugs (I have my own PRs unmerged for years).
From my side - I will try to look again to this issue.

But you are definitely are able to fix it locally not waiting this issue to get merged.

johnd0e:
Added reference to #6072

PR #7439, opened 2021-01-29
More reliable disableClickPropagation

### LINKED ISSUE #6546: disableClickPropagation include right click / contextmenu
When creating a custom container on top of a Leaflet map and I call `disableClickPropagation` on that custom container my regular clicks are not propagated to the map (good). If I right click on the container, I see my map contextmenu opening under the custom container (not so good).

So I am wondering if it would make sense to add the `contextmenu` event to `disableClickPropagation` or maybe to create another method that disables the right click propagation?

This can be solved individually like this:
```
L.DomEvent.on(this._container, 'contextmenu', L.DomEvent.stopPropagation);
```

So I am not sure if it's worth the overhead? But then on the other hand, why have a helper method to disable the click propagation :).

### LINKED ISSUE #7431: Simplify fakeStop/skipped code paths
Currently `disableClickPropagation` is rather complicated.

For [some](https://github.com/Leaflet/Leaflet/commit/5a7420dd1a43474cccaa8cdefa4f324452d18f36) cases it uses [`fakeStop`](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/dom/DomEvent.js#L187):
https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/dom/DomEvent.js#L257-L260
The purpose of this function - to solve #301. Whether it is really best solution - is out of scope of this issue, but here I want to point the fact, that it is implemented in such way, that not easy to understand. There is also related [`skipped`](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/src/dom/DomEvent.js#L262-L267) function with unobvious behavior, called from couple of places, and so on.

1. Originally it was rather simple: 5a7420dd1a43474cccaa8cdefa4f324452d18f36.
2. It was more complex after 5c1a34979ecb34e596abeffc4342d7926f6434fa, which purpose was fix for IE<9.

I suppose we should not care about legacy IE anymore, and propose to revert latter commit.

johnd0e:
ATM `DomEvent.disableClickPropagation` is not reliable enough as it relies on shared state (stored in private `skipEvents`).
I can imagine many cases where this approach may fail. E.g. if some external component will use native `stopPropagation` (instead of Leaflet's `DomEvent.stopPropagation`) - the state left set, and next click will be ignored.

This PR makes things simple and straight-forward.

Close #7431, fix #6546 (prevent `contextmenu` as well).

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -63,6 +63,7 @@ export var Canvas = Renderer.extend({
 		DomEvent.on(container, 'mousemove', this._onMouseMove, this);
 		DomEvent.on(container, 'click dblclick mousedown mouseup contextmenu', this._onClick, this);
 		DomEvent.on(container, 'mouseout', this._handleMouseOut, this);
+		container['_leaflet_disable_events'] = true;
 
 		this._ctx = container.getContext('2d');
 	},
@@ -355,7 +356,6 @@ export var Canvas = Renderer.extend({
 			}
 		}
 		if (clickedLayer)  {
-			DomEvent.fakeStop(e);
 			this._fireEvent([clickedLayer], e);
 		}
 	},
--- a/spec/suites/dom/DomEventSpec.js
+++ b/spec/suites/dom/DomEventSpec.js
@@ -142,25 +142,6 @@ describe('DomEvent', function () {
 
 			map.remove();
 		});
-
-		it('does not interfere with stopPropagation', function () { // test for #1925
-			var child = document.createElement('div');
-			el.appendChild(child);
-			L.DomEvent.disableClickPropagation(child);
-			L.DomEvent.on(child, 'click', L.DomEvent.stopPropagation);
-			var map = L.map(el).setView([0, 0], 0);
-			map.on('click', listener);
-
-			happen.once(child, {type: 'click'});
-
-			expect(listener.notCalled).to.be.ok();
-
-			happen.once(map.getContainer(), {type: 'click'});
-
-			expect(listener.called).to.be.ok();
-
-			map.remove();
-		});
 	});
 
 	describe('#preventDefault', function () {
--- a/spec/suites/dom/DomEventSpec.js
+++ b/spec/suites/dom/DomEventSpec.js
@@ -104,9 +104,10 @@ describe('DomEvent', function () {
 			var child = document.createElement('div');
 			el.appendChild(child);
 			L.DomEvent.disableClickPropagation(child);
-			L.DomEvent.on(el, 'dblclick mousedown touchstart', listener);
+			L.DomEvent.on(el, 'dblclick contextmenu mousedown touchstart', listener);
 
 			happen.once(child, {type: 'dblclick'});
+			happen.once(child, {type: 'contextmenu'});
 			happen.once(child, {type: 'mousedown'});
 			happen.once(child, {type: 'touchstart', touches: []});
 
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -184,7 +184,7 @@ export function disableScrollPropagation(el) {
 // `'mousedown'` and `'touchstart'` events (plus browser variants).
 export function disableClickPropagation(el) {
 	on(el, 'mousedown touchstart dblclick', stopPropagation);
-	addOne(el, 'click', fakeStop);
+	el['_leaflet_disable_click'] = true;
 	return this;
 }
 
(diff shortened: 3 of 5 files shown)


mourner (approved):

mourner:
Nice approach!

Falke-Design:
Because of this PR the tests are failing

src/layer/vector/Canvas.js
```diff
@@ -63,6 +63,7 @@ export var Canvas = Renderer.extend({
 		DomEvent.on(container, 'mousemove', this._onMouseMove, this);
 		DomEvent.on(container, 'click dblclick mousedown mouseup contextmenu', this._onClick, this);
 		DomEvent.on(container, 'mouseout', this._handleMouseOut, this);
+		container['_leaflet_disable_events'] = true;
```
Falke-Design, line 66: if this line is moved from here to line 358 the tests are working

johnd0e:
> Because of this PR the tests are failing

Investigating..

src/layer/vector/Canvas.js
```diff
@@ -355,7 +356,6 @@ export var Canvas = Renderer.extend({
 			}
 		}
 		if (clickedLayer)  {
-			DomEvent.fakeStop(e);
 			this._fireEvent([clickedLayer], e);
 		}
```
johnd0e, line 360: Perhaps we need `else` here to pass event to map.

mustafaradheyyan:
While using React with Leaflet v1.8.0, I've noticed an error "Uncaught TypeError: Cannot read properties of null (reading '_leaflet_disable_click')" whenever the popup state is reset upon firing a "click" type event.

This error did not occur with Leaflet v1.7.1, and I'm presuming it's because I'm using https://github.com/slutske22/react-leaflet-editable-popup which has some setState functions inside of the extended Popup class.

I was able to remedy the problem somewhat using onMouseDown instead of onClick within the popup buttons, but when hiding and unhiding markers from the map, through changing an overlay layer using the layer control, the error appears once again, probably because the popup was triggered again through a "click" event on the layer control radio buttons.

This might not be applicable since I'm using React and especially because I'm using a modified Popup class, but I thought it was worth mentioning to see if there is anyway to adjust these changes to accommodate for different use cases, if that would keep the same reliable functionality of the disableClickPropagation.

~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Specific error message:

Uncaught TypeError: Cannot read properties of null (reading '_leaflet_disable_click')
    at NewClass._isClickDisabled (Map.js:1381:1)
    at NewClass._handleDOMEvent (Map.js:1388:1)
    at HTMLDivElement.handler (DomEvent.js:108:1)
_isClickDisabled @ Map.js:1381
_handleDOMEvent @ Map.js:1388
handler @ DomEvent.js:108

AngelaCandela:
Since I upgraded leaflet to version 1.8.0, two components in my app have stopped working. They are a [react-search-box](https://www.npmjs.com/package/react-search-box) and a [react-select](https://react-select.com/home) that I placed inside my own custom control. I have applied `DomEvent.disableClickPropagation()` to the control container, like so:
`L.DomEvent.disableClickPropagation(container);`
and it used to work without any problem before the upgrade. Since the upgrade, whenever I select one of the options available from the search (react-search-box) or the select (react-select) dropdown menu, I get this error:

image

image (1)

@mourner and @Falke-Design do you guys have any idea what I could do to fix this? Or is there anything that you guys can do on your side? Thanks for your precious time!

Falke-Design:
Looks like react has removed already the DOM elememt before the listener / code is executed. I haven't digged in the code but I think this should be resolveable with checking if the element is existing.

@AngelaCandela can you please set up a demo in codesandbox or in a other online editor

AngelaCandela:
@Falke-Design Yes, that's exactly the same conclusion I came to after some debugging: once the option is selected, it no longer shows in the dropdown, hence the corresponding DOM element is removed which causes the error.

Yes, I will try and set up a demo in codesandbox for you! Thanks for the quick response!

AngelaCandela:
Thank you, @mustafaradheyyan! I did try and use your approach on both the [react-search-box](https://www.npmjs.com/package/react-search-box) and the [react-select](https://react-select.com/home) components. However, my problem was that I could not capture the event:
* the `react-search-box` event handler's type is: `(record: Record) => void`, being `Record`'s type the following:
```
interface Record {
  item: {
    key: string;
    value: string;
  };
}
```
* the `react-select` event handler's type is: `(newValue: unknown) => void`.

All this to say that neither of them exposes the `event` so, unfortunately, I wasn't able to set `L.DomEvent.stopPropagation(event)`. Thank you so much for the suggestion though!

Issue #7429, opened 2021-01-23
Error on clearLayers - this._icon is undefined


luizalexandrew:
I'm having a problem when I call the Leaflet's **clearLayers** function. In 99% of the use of the application this function is executed correctly, but in some moments an exception is launched from the clearLayers function

Call that triggers the error
![image](https://user-images.githubusercontent.com/10445220/105580900-f1886b80-5d6d-11eb-859a-d466c1611aba.png)

getMapLayersLeaflet is just a function that searches for layers or creates if it doesn't exist (singleton idea)
![image](https://user-images.githubusercontent.com/10445220/105580824-74f58d00-5d6d-11eb-825d-3bcb7cc7ac77.png)

**Error**
![image](https://user-images.githubusercontent.com/10445220/105580021-a750bb80-5d68-11eb-9076-3efbdb027c7f.png)

Original Code: **leaflet-src.js** in function **_removeIcon**. The error is coming from the call  **remove(this._icon)**, because it's coming in this function **this._icon** with **undefined**

![image](https://user-images.githubusercontent.com/10445220/105580051-d0714c00-5d68-11eb-8b3e-1bf760018279.png)

**Debug Code**: Showing that it is coming undefined in the _removeIcon function

![image](https://user-images.githubusercontent.com/10445220/105580115-3eb60e80-5d69-11eb-9850-fefbe56895d7.png)

**Result 1**

![image](https://user-images.githubusercontent.com/10445220/105580136-6016fa80-5d69-11eb-9674-897aad15ef80.png)

**Result 2**

![image](https://user-images.githubusercontent.com/10445220/105580208-c734af00-5d69-11eb-8566-20d5a583f44f.png)

I believe that a race condition is occurring, at some point that it has 2 expenses at the same time, this is an assumption

**Initial Solution**:  The solution was to check if this._icon exist on **leaflet-src.js** in function **_removeIcon**.
![image](https://user-images.githubusercontent.com/10445220/105580249-0400a600-5d6a-11eb-9441-9b1f68449344.png)
I don't know if this solution was correct

I'm running the latest version of Leaflet v1.7.1, the error also occurs in previous versions
The browser that is running the code is the Google Chrome Version 81.0.4044.92 (Official Build) (64-bit)

![image](https://user-images.githubusercontent.com/10445220/105580523-a1100e80-5d6b-11eb-877f-ca20655c2a58.png)

johnd0e:
[...]

PR #7440, opened 2021-01-29
Bugfixed #7429 - Error on clearLayers - this._icon is undefined

luizalexandrew:


johnd0e (changes requested):
Please exclude `dist` directory and `package.json`.

PR #7441, opened 2021-01-30
Fix `this._icon` undefined on `clearLayers()`

### LINKED ISSUE #7429: Error on clearLayers - this._icon is undefined

luizalexandrew:
Bugfixed - Error on clearLayers - this._icon is undefined fixes #7429

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -272,9 +272,11 @@ export var Marker = Layer.extend({
 				mouseout: this._resetZIndex
 			});
 		}
-
-		DomUtil.remove(this._icon);
-		this.removeInteractiveTarget(this._icon);
+		
+		if (this._icon) {
+			DomUtil.remove(this._icon);
+			this.removeInteractiveTarget(this._icon);
+		}
 
 		this._icon = null;
 	},


luizalexandrew:
The error occurs because in specific situations a race condition occurs in this._icon, only one check solved the problem

johnd0e:
> The error occurs because in specific situations a race condition occurs in this._icon,

Original issue description lacks details required to reproduce the issue.

Falke-Design:
I close this PR until we have a reproduceable example

Issue #7435, opened 2021-01-28
DomUtil functions not covered by tests


johnd0e:
We have some [DomUtil](https://github.com/Leaflet/Leaflet/blob/master/src/dom/DomUtil.js) tests here: https://github.com/Leaflet/Leaflet/blob/master/spec/suites/dom/DomUtilSpec.js, but not all functions are covered.

Here the list of lacking tests:
1. `getStyle(<HTMLElement> el, <String> styleAttrib)`
2. `create(<String> tagName, <String> className?, <HTMLElement> container?)`
3. `remove(<HTMLElement> el)`
4. `empty(<HTMLElement> el)`
5. `toFront(<HTMLElement> el)`
6. `toBack(<HTMLElement> el)`
7. `setClass(<HTMLElement> el, <String> name)`
8. `getClass(<HTMLElement> el)`
9. `setOpacity(<HTMLElement> el, <Number> opacity)`
0. `testProp(<String[]> props)`
0. `setTransform(<HTMLElement> el, <Point> offset, <Number> scale?)`
0. `setPosition(<HTMLElement> el, <Point> position)`
0. `getPosition(<HTMLElement> el)`
0. `disableTextSelection()`
0. `enableTextSelection()`
0. `disableImageDrag()`
0. `enableImageDrag()`
0. `preventOutline(<HTMLElement> el)`
0. `restoreOutline()`
0. `getSizedParentNode(<HTMLElement> el)`
0. `getScale(<HTMLElement> el)`

Ref: https://leafletjs.com/reference-1.7.1.html#domutil

Relkfaw:
Hello, this issue is still available ? Can i work on it ?

johnd0e:
Sure

Relkfaw:
1. I'm not sure how to test `disableTextSelection() enableTextSelection() disableImageDrag() enableImageDrag()`, but when i look at them, it seems these functions just called DomEvent.off and DomEvent.on. So maybe they should be tested into DomEventSpec (related to #7434 ?)
2. `restoreOutline()` is just called inside preventOutline and like above, I'm not really sure of its effect. Should it be necessarily tested ?
3. Some tests failed, but I'm not sure if what I expect is wrong or not, for example:
    1. `preventOutline(element)` should set element.style.outline to "none" but nothing change.
    2. `getSizedParentNode(element)` crash when the element isn't added to the document yet, should I remove this test ? Maybe this case will never appear

johnd0e:
[...]

PR #7443, opened 2021-01-31
Add lacking test into dom util spec

Relkfaw:
#7435


Falke-Design:
I suggested the good tests in #7547 and we will keep going with that one

PR #7446, opened 2021-02-01
Prevent false map click after fast dragging of marker

### LINKED ISSUE #4457: Leaflet map draggable marker event bug
Everytime I click on the map I create a draggable marker with a `drag` function. It works great, however everytime I release the mouse after I dragged some marker, the `click` event of the map is triggered. How can I stop that ?

```
map.on('click',
    function (e) {
        var pos = e.latlng;
        console.log('map click event');

        var marker = L.marker(
            pos,
            {
                draggable: true
            }
        );
        marker.on('drag', function (e) {
            console.log('marker drag event');
        });
        marker.on('click', L.DomEvent.stopPropagation);
        marker.addTo(map);
    }
);
```

**Edit:** `marker.on('click', L.DomEvent.stopPropagation);` This partially fixes the bug. Now if I drag the marker VERY quickly and release it, the bug still occurs, because the marker "follows" the mouse cursor in a speed slower than the speed with which I move my cursor, and so when the mouse is released, the mouse is not on the marker anymore, so the map `click` event is triggered.

**Edit 2** For the record I am using `Leaflet 1.0.0-rc1` and chrome `49.0.2623.112`

johnd0e:
Fix #4457 (and duplicates).

**Fix final position after dragging**
For unclear reason `_startPos` was set wrong (increasing it for offset of first move event), so we fix that,
like it was in initial implementation (no idea why it was ever changed).
(https://github.com/Leaflet/Leaflet/commit/2d2fc74110f2fe7fbc06d7f6e0a2e90528ce46c6#diff-d8f00e9d1726fd6483110dc44c93d9f8R84)

**Prevent false map click after fast dragging of marker**
To solve this we have to ensure that draggable final position is already set on the time
of mouseup/touchend. Thus we have only option - reject using of `requestAnimationFrame`,
which was originally introduced to fix performance issues of map panning, see
https://github.com/Leaflet/Leaflet/commit/c93d91bfad73483990fd5340ed0db8908ba5a4b6#diff-46eca63ac2c5237cbd2d45c8a60f6ee28ce34231aef53669247ead03f012dff2

Perhaps browsers evolved since then.

**Adapt test to expect exact position values**
(as we do not depend on timings anymore)

**Todo:**
- [ ] Check panning performance.
In case if it is not sufficient we could use `requestAnimationFrame` but for map only.

---
Demo pages:
- Original bug: https://gist.githack.com/johnd0e/279b2c5bd985027fd3f959d8b1cec9ea/raw/bug.html
- Fixed with this PR: https://gist.githack.com/johnd0e/279b2c5bd985027fd3f959d8b1cec9ea/raw/fixed.html

### DIFF
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -41,11 +41,8 @@ describe("Marker.Drag", function () {
 				onStop: function () {
 					expect(marker.getOffset()).to.eql(offset);
 
-					expect(map.getCenter()).to.eql({lat: 0, lng: 0});
-					expect(marker.getLatLng().equals({ // small margin of error allowed
-						lat: -40.979898069620134,
-						lng: 360
-					})).to.be.ok();
+					expect(map.getCenter()).to.be.nearLatLng([0, 0]);
+					expect(marker.getLatLng()).to.be.nearLatLng([-40.979898069620134, 360]);
 
 					done();
 				}
@@ -76,11 +73,8 @@ describe("Marker.Drag", function () {
 					onStop: function () {
 						expect(marker.getOffset()).to.eql(offset);
 
-						expect(map.getCenter()).to.eql({lat: 0, lng: 0});
-						expect(marker.getLatLng().equals({ // small margin of error allowed
-							lat: -40.979898069620134,
-							lng: 360
-						})).to.be.ok();
+						expect(map.getCenter()).to.be.nearLatLng([0, 0]);
+						expect(marker.getLatLng()).to.be.nearLatLng([-40.979898069620134, 360]);
 
 						done();
 					}
@@ -110,11 +104,8 @@ describe("Marker.Drag", function () {
 					expect(marker.getOffset()).to.eql(offset);
 
 					// small margin of error allowed
-					expect(map.getCenter().equals({lat: 0, lng: 11.25})).to.be.ok();
-					expect(marker.getLatLng().equals({
-						lat: -40.979898069620134,
-						lng: 419.0625
-					})).to.be.ok();
+					expect(map.getCenter()).to.be.nearLatLng([0, 11.25]);
+					expect(marker.getLatLng()).to.be.nearLatLng(-40.979898069620134, 419.0625);
 
 					done();
 				}
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -105,7 +105,7 @@ describe("Marker.Drag", function () {
 
 					// small margin of error allowed
 					expect(map.getCenter()).to.be.nearLatLng([0, 11.25]);
-					expect(marker.getLatLng()).to.be.nearLatLng(-40.979898069620134, 419.0625);
+					expect(marker.getLatLng()).to.be.nearLatLng([-40.979898069620134, 419.0625]);
 
 					done();
 				}
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -17,101 +17,112 @@ describe("Marker.Drag", function () {
 		document.body.removeChild(div);
 	});
 
+	var MyMarker = L.Marker.extend({
+		_getPosition: function () {
+			return L.DomUtil.getPosition(this.dragging._draggable._element);
+		},
+		getOffset: function () {
+			return this._getPosition().subtract(this._initialPos);
+		}
+	}).addInitHook('on', 'add', function () {
+		this._initialPos = this._getPosition();
+	});
+
 	describe("drag", function () {
 		it("drags a marker with mouse", function (done) {
-			var marker = new L.Marker([0, 0], {
-				draggable: true
-			}).addTo(map);
+			var marker = new MyMarker([0, 0], {draggable: true}).addTo(map);
+
+			var start = L.point(300, 280);
+			var offset = L.point(256, 32);
+			var finish = start.add(offset);
 
 			var hand = new Hand({
 				timing: 'fastframe',
 				onStop: function () {
-					var center = map.getCenter();
-					expect(center.lat).to.be(0);
-					expect(center.lng).to.be(0);
+					expect(marker.getOffset()).to.eql(offset);
 
-					var markerPos = marker.getLatLng();
-					// Marker drag is very timing sensitive, so we can't check
-					// exact values here, just verify that the drag is in the
-					// right ballpark
-					expect(markerPos.lat).to.be.within(-50, -30);
-					expect(markerPos.lng).to.be.within(340, 380);
+					expect(map.getCenter()).to.eql({lat: 0, lng: 0});
+					expect(marker.getLatLng().equals({ // small margin of error allowed
+						lat: -40.979898069620134,
+						lng: 360
+					})).to.be.ok();
 
 					done();
 				}
 			});
 			var toucher = hand.growFinger('mouse');
 
-			toucher.wait(100).moveTo(300, 280, 0)
-				.down().moveBy(5, 0, 20).moveBy(256, 32, 1000).wait(100).up().wait(100);
+			toucher.moveTo(start.x, start.y, 0)
+				.down().moveBy(5, 0, 20).moveTo(finish.x, finish.y, 1000).up();
 		});
 
 		describe("in CSS scaled container", function () {
-			var scaleX = 2;
-			var scaleY = 1.5;
+			var scale = L.point(2, 1.5);
 
 			beforeEach(function () {
 				div.style.webkitTransformOrigin = 'top left';
-				div.style.webkitTransform = 'scale(' + scaleX + ', ' + scaleY + ')';
+				div.style.webkitTransform = 'scale(' + scale.x + ', ' + scale.y + ')';
 			});
 
-			// fixme IE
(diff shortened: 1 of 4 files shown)


nrenner:
Looks good to me, I can confirm it solves both the offset and the fast dragging issue. I have only tested DivIcon Marker, not other elements, nor performance.

The [Marker.DragSpec](https://github.com/Leaflet/Leaflet/blob/master/spec/suites/layer/marker/Marker.DragSpec.js) should have detected the offset bug, so I think it should be fixed as well.

Either by also removing the `moveBy(5, 0, 20)`, whose omission by the wrong offset fell within the tested tolerance. Or by testing exact values instead of a `within` range. I don't know if the timing issue mentioned in the comment still exists or if it was just assumed because the offset bug produced an unexpected result. Repeated tests with exact values do succeed, yet I do see some sporadic failures when calling a reduced `spec/index.html` (only Marker.Drag and Map.Drag spec) in Chromium and Firefox, but those also occur on current master and the failing results are far outside the tested tolerance.

Testing for pixel values as used for moving instead of coordinates would also make the test better understandable, so my proposal would be to replace the current [expect statements](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/spec/suites/layer/marker/Marker.DragSpec.js#L33-L38) with this: 

```js
var markerPos = marker._icon._leaflet_pos;
expect(markerPos.x).to.be(300 + 5 + 256);
expect(markerPos.y).to.be(300 + 32);
```

johnd0e:
> nor performance.

I hope that you'll find some time to check that too

> Either by also removing the moveBy(5, 0, 20),

I'm thinking about returning that move (but replace following `moveBy` with `moveTo`).

> Testing for pixel values as used for moving instead of coordinates would also make the test better understandable

Sounds good, but it can be not so obvious for scaled container test.

johnd0e:
> The [Marker.DragSpec](https://github.com/Leaflet/Leaflet/blob/master/spec/suites/layer/marker/Marker.DragSpec.js) should have detected the offset bug, so I think it should be fixed as well.

How to reproduce that?
There is no problems on travis: https://travis-ci.org/github/Leaflet/Leaflet/builds/757065282

Do you mean "in CSS scaled container" tests? Make sure that you have browser window is large enough (like 1280*1024).

nrenner:
> There is no problems on travis: https://travis-ci.org/github/Leaflet/Leaflet/builds/757065282

What I mean is it should fail on master, because there is a bug.

But I see you already implemented what I meant.

johnd0e:
> What I mean is it should fail on master, because there is a bug.

```
PhantomJS 2.1.1 (Windows 8) Map.Drag mouse events change the center of the map FAILED
        Error: expected { x: 251, y: 32 } to sort of equal { x: 256, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8) Map.Drag mouse events in CSS scaled container change the center of the map, compensating for CSS scale FAILED
        Error: expected { x: 253.5, y: 32 } to sort of equal { x: 256, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8) Map.Drag touch events change the center of the map FAILED
        Error: expected { x: 251, y: 32 } to sort of equal { x: 256, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8) Marker.Drag drag drags a marker with mouse FAILED
        Error: expected { x: 251, y: 32 } to sort of equal { x: 256, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8) Marker.Drag drag pans map when autoPan is enabled FAILED
        Error: expected { x: 285, y: 32 } to sort of equal { x: 290, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8) Marker.Drag drag in CSS scaled container drags a marker with mouse, compensating for CSS scale FAILED
        Error: expected { x: 253.5, y: 32 } to sort of equal { x: 256, y: 32 } (node_modules/expect.js/index.js:102)
PhantomJS 2.1.1 (Windows 8): Executed 16 of 704 (6 FAILED) (1.307 secs / 1.254 secs)
```

Falke-Design (approved):

johnd0e:
@Falke-Design 
Could you confirm that there is no performance loss (see first message)?

Falke-Design:
I dragged the map, the marker, the DivIcon, clicked and has done some monkey tests but there was never a performance loss.

johnd0e:
> ...but there was never a performance loss.

Perhaps to see it we need hard conditions, like large map with many objects.

spec/suites/layer/marker/Marker.DragSpec.js
```diff
@@ -17,101 +17,112 @@ describe("Marker.Drag", function () {
...
+					expect(marker.getOffset()).to.eql(offset);
 
-					var markerPos = marker.getLatLng();
-					// Marker drag is very timing sensitive, so we can't check
-					// exact values here, just verify that the drag is in the
-					// right ballpark
-					expect(markerPos.lat).to.be.within(-50, -30);
-					expect(markerPos.lng).to.be.within(340, 380);
+					expect(map.getCenter()).to.eql({lat: 0, lng: 0});
+					expect(marker.getLatLng().equals({ // small margin of error allowed
```
Falke-Design, line 45: Why not using `nearLatLng` or the percision argument?

spec/suites/layer/vector/CanvasSpec.js
```diff
@@ -159,9 +161,9 @@ describe('Canvas', function () {
 			var mouse = hand.growFinger('mouse');
 
 			mouse.wait(100)
-				.moveTo(300, 300, 0).down().moveBy(5, 0, 20).up()
-				.moveTo(100, 100, 0).down().moveBy(5, 0, 20).up()
-				.moveTo(300, 300, 0).down().moveBy(5, 0, 20).up();
+				.moveTo(300, 300, 0).down().moveBy(5, 0, 20).up()  // control case
+				.moveTo(100, 100, 0).down().moveBy(5, 0, 20).up()  // drag the map (outside of circle)
+				.moveTo(300, 300, 0).down().moveBy(5, 0, 20).up(); // expect mousedown ok
```
Falke-Design, line 166: Please check this, the test checked before if the circle is at the exact position and now the second click works because the radius is big enough.

After moving the circle 5px and then the map 5px, the second circle click should happen at `.moveTo(310, 300, 0)`and not at `.moveTo(300, 300, 0)`

spec/suites/map/handler/Map.DragSpec.js
```diff
@@ -185,10 +203,8 @@ describe("Map.Drag", function () {
 			});
 			var mouse = hand.growFinger('mouse');
 
-			// We move 5 pixels first to overcome the 3-pixel threshold of
-			// L.Draggable.
-			mouse.wait(100).moveTo(200, 200, 0)
```
Falke-Design, line 190: Are you sure that removing `wait(100)` is good? I take the experience that it can sometimes results in errors if it is not set.

spec/suites/map/handler/Map.DragSpec.js
```diff
@@ -324,30 +338,34 @@ describe("Map.Drag", function () {
...
-					var zoom = map.getZoom();
-					expect(center.lat).to.be.within(21.9430, 21.9431);
-					expect(center.lng).to.be(-180);
-					expect(zoom).to.be(1);
+					expect(map.getOffset()).to.eql(offset);
+
+					expect(map.getZoom()).to.be(1);
+					expect(map.getCenter().equals({ // small margin of error allowed
+						lat: 21.943045533, // 21.94304553343818
+						lng: -180
```
Falke-Design, line 359: Interesseting that it is still exact `-180`, in my tests the lng was not any more exact at -180, because the first move is not ignored anymore. But maybe I was wrong or you have changed something that I don't see.

spec/suites/layer/vector/CanvasSpec.js
johnd0e: But do we really need to preserve exact position?

spec/suites/map/handler/Map.DragSpec.js
johnd0e: I see no problem without wait, that's why I removed it.

spec/suites/map/handler/Map.DragSpec.js
johnd0e: May be because of removing anim frame.

Falke-Design (changes requested):

spec/suites/layer/vector/CanvasSpec.js
Falke-Design: No

Falke-Design:
@mourner I don't know how fast we get response from @johnd0e, he was not available since 2022. If you want you can give me write access to update the PR. The other option is that you change it or we remove it from the 1.8 Release. But the PR would be ready after the changes

mourner:
> If you want you can give me write access

Just did so! Thanks for being such a valuable contributor to the project. Just make sure you don't merge potentially breaking or performance-impacting PRs without other maintainer's review.

Falke-Design:
Thank you, I will be careful!

Falke-Design:
There is a open todo:
> Todo:
>    - [ ] Check panning performance.
>    In case if it is not sufficient we could use requestAnimationFrame but for map only.

mourner (approved):

PR #7451, opened 2021-02-03
Fix map spec

johnd0e:
Minor fixes
- tests: prevent warnings from webserver
  Like this: [`03 02 2021 08:16:25.848:WARN [web-server]: 404: /000`](https://travis-ci.org/github/Leaflet/Leaflet/builds/757360973#L284)
  (also seen as unexpected newlines in progress log)
  In fact those warnings was caused by these tests:
  <details>

  https://github.com/Leaflet/Leaflet/blob/master/spec/suites/map/MapSpec.js#L558-L602
  </details>
  But better to fix all just in case.
- make quotation marks consistent
- minor simplification
- fix tests containing expectations in callbacks

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -286,7 +286,7 @@ describe("Map", function () {
 			expect(bounds.contains(map.getBounds())).to.be(true);
 		});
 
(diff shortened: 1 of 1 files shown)


mourner (approved):
So good to see all the simplification / cleanup PRs!

PR #7459, opened 2021-02-08
Break the reference to the options of the Class prototype

### LINKED ISSUE #7458: Default Class options are changing
1. I create a rectangle
2. While creating, I call initHook and creating a new instance of a custom class (`L.Class.extend`)
3. I change the options of the class on the rectangle
4. The options of the prototype of the class are changed

**Expected behavior**
The options of the extended L.Class should never changed from outside

**Environment**
- Leaflet version: 1.7.1

**Test**
```
L.PM = L.Class.extend({
   options: {
  	test: 1,
        something: true
  }
})

function initRectangle() {
  this.pm = new L.PM(this);
}
L.Rectangle.addInitHook(initRectangle);


console.log(L.PM.prototype.options)

var rect = L.rectangle(map.getBounds()).addTo(map);
rect.pm.options.x = 'option changed'

console.log(L.PM.prototype.options)
```
https://jsfiddle.net/falkedesign/gwj6dphf/

When I add following the connection / reference between the options breaks:
```
var rect = L.rectangle(map.getBounds()).addTo(map);
L.setOptions(rect.pm,{a: 'xxx'})
rect.pm.options.x = 'option changed'
```

Falke-Design:
Fix #7458


spec/suites/core/ClassSpec.js
```diff
@@ -117,6 +117,15 @@ describe("Class", function () {
 			expect(K2.prototype.options.foo).to.eql('bar');
 		});
 
+		it("prevents change of prototype options", function () {
+			var K1 = L.Class.extend({options: {}});
+			var rect = L.rectangle([[0, 0], [0, 0]]);
+			rect.klass = new K1(rect);
+			rect.klass.options.foo = 'bar';
+			expect(K1.prototype.options.foo).to.not.eql('bar');
+			expect(rect.klass.options.foo).to.eql('bar');
```
johnd0e, line 126:
```suggestion
			var Klass = L.Class.extend({options: {}});
			var instance = new Klass();
			expect(Klass.prototype.options).to.not.be(instance.options);
```

src/core/Class.js
```diff
@@ -17,6 +17,8 @@ Class.extend = function (props) {
 	// Returns a Javascript function that is a class constructor (to be called with `new`).
 	var NewClass = function () {
 
+		Util.setOptions(this, {});
```
johnd0e, line 20:
```suggestion
		Util.setOptions(this);
```

johnd0e: And even more: as now it is guaranteed that every object has distinct `options` we could refactor the code, adding this to constructor function:
```js
		this.options = this.options ? Util.create(this.options) : {};
```
And removing this:
https://github.com/Leaflet/Leaflet/blob/620bc3e9c5d282090f429ac21d28d00fbffdb3db/src/core/Util.js#L133-L135

johnd0e (changes requested):
In general I found this PR useful.

Atm Leaflet class system implies that we call `setOptions` somewhere in `initialize`, and if we don't - we get very undesirable side effect (all instances will share same options).

To prevent that we can either ensure that every instance has own distinct options (as in this PR), or document that user MUST call `setOptions` in `initialize` (we even can check if this demand is really satisfied).

src/core/Class.js
Falke-Design: Where do you want me to add the code `this.options = this.options ? Util.create(this.options) : {};`? Instead of `Util.setOptions(this):`?

Falke-Design:
I think it would be more usefull when `setOptions` is called internally, else users will be always confused why the options are not correct and have to search what is wrong.

I searched around 30 minutes why my code was not working until I found this issue.

src/core/Class.js
johnd0e: > Where do you want me to add the code

I suppose this PR is finished, and we will wait for other maintainers to review it.
And if it will be merged - then we will continue from that part.

johnd0e (approved):

johnd0e (approved):

mourner (approved):

PR #7460, opened 2021-02-08
Clarify zoomend event

xeruf:


johnd0e:
Documentation is autogenerated from sources, so I suppose you should `Map.js`.

mourner (approved):

PR #7463, opened 2021-02-10
Replace uglify-js with terser

ronikar:
In order to use `ECMAScript 6` features in leaflet code, I replaced `uglify-js` package :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "rollup-plugin-json": "^4.0.0",
     "sinon": "^7.5.0",
     "ssri": "^8.0.1",
-    "uglify-js": "^3.12.6"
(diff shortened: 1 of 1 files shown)


johnd0e:
> In order to use `ECMAScript 6`

But Leaflet release also includes `leaflet-src.js`, so...

ronikar:
But 

> > In order to use `ECMAScript 6`
> 
> But Leaflet release also includes `leaflet-src.js`, so...

But `npm run build` throws an exception when I use `ES6` features

mourner:
Leaflet still uses ES5 for its code base so we can't upgrade to `terser`. Upgrading the codebase for ES6 is a separate discussion and we can't do it abruptly.

Issue #7445, opened 2021-02-01
Unexpected behavior of panInside


user073:
- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**

Please see the unexpected behavior of [panInside](https://leafletjs.com/reference-1.7.1.html#map-paninside) the CodeSandbox below. The bottom padding passed to panInside in map2 is only 1px greater than the bottom padding passed to panInside in map1 but map2 pans the marker all the way to the top while map1 does not pan at all.

**Expected behavior**

map2 should only pan up one pixel, as this is the minimum amount to make marker2 visible within the padding.

**Current behavior**

map2 pans so that marker2 is at the top of the map.

**Environment**
- Leaflet version: ^1.7.1
- Browser (with version): Chrome 88.0.4324.96 
- OS/Platform (with version): macOS High Sierra 10.13.6

**Additional context**

**Minimal example reproducing the issue**

[![Edit quirky-frost-946z7](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/quirky-frost-946z7?fontsize=14&hidenavigation=1&theme=dark)

Please see this screen recording: 

https://user-images.githubusercontent.com/63527968/106411334-e8e10680-63f9-11eb-8671-7a7ddd3fb899.mov

- [x] this example is as simple as possible
- [x] this example does not rely on any third party code

user073:
@daverayment Is this a bug?

daverayment:
@user073 Thanks for reporting this and for the clear example. This is definitely not intended behaviour. The map should pan the smallest amount so the marker is within the padded bounds.

I have a lot on at the moment, but I'll try to track down the root cause soon.

daverayment:
The calculations currently depend on the padding taking up less than half the bounds, otherwise the movement offset created is reversed, i.e. a Y scroll of 1 pixel down from the top of the bounds instead of 1 pixel up from the bottom. I'll attempt to correct this soon.

user073:
[...]

PR #7469, opened 2021-02-13
Update panInside

daverayment:
Rewrite of panInside to correct issue #7445. Now correctly deals with cases where the padding exceeds half of the display bounds.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1093,21 +1093,15 @@ describe("Map", function () {
 
 		it("pans correctly when padding takes up more than half the display bounds", function () {
 			var oldCenter = map.project(center);
-			// Create target inside the padding area, but above the centrepoint
-			// (replicates issue #7445)
-			var panTarget = oldCenter.subtract([0, 5]);
-			var halfSize = map.getPixelBounds().getSize().divideBy(2);
-			// Make padding take up the bottom half of the display bounds plus 15 pixels
-			var paddingAmount = halfSize.y + 15;
-			// Execute the pan to the target point, taking into account the padding
-			map.panInside(map.unproject(panTarget), {paddingBottomRight: [0, paddingAmount], animate: false});
-			// Calculate the distance moved during the pan
-			var newCenter = map.project(map.getCenter());
-			var offset = newCenter.subtract(oldCenter);
-			// The target was placed 5 pixels above center and the padding was 15 pixels
-			// above center, so the Y distance should be 15 - 5 = 10 pixels
-			expect(offset.y - 10).to.be.lessThan(1);
-			expect(offset.x).to.be.lessThan(1);
+			var targetOffset = L.point(0, -5); // arbitrary point above center
+			var target = oldCenter.add(targetOffset);
+			var paddingOffset = L.point(0, 15);
+			var padding = map.getSize().divideBy(2) // half size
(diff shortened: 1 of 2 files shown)


spec/suites/map/MapSpec.js
```diff
@@ -1090,6 +1090,17 @@ describe("Map", function () {
...
 		});
+
+		it("pans correctly when padding takes up more than half the display bounds", function () {
+			var c = map.project(center);
+			var p = c.add([0, 5]);
+			var halfSize = map.getPixelBounds().getSize().divideBy(2);
+			map.panInside(map.unproject(p), {paddingBottomRight: [0, halfSize.y + 15], animate: false});
+			var newCenter = map.project(map.getCenter());
+			var offset = newCenter.subtract(c);
+			expect(Math.abs(offset.y - 20)).to.be.lessThan(1);
```
johnd0e, line 1101: Please try to make these calculations more obvious.
With brief view it's hard to say why we take some arbitrary values, process them randomly, and compare with other values.

daverayment: Updated with some contextual comments and clearer variable naming in new commit.

spec/suites/map/MapSpec.js
```diff
@@ -1090,6 +1090,25 @@ describe("Map", function () {
...
+			var paddingAmount = halfSize.y + 15;
+			// Execute the pan to the target point, taking into account the padding
+			map.panInside(map.unproject(panTarget), {paddingBottomRight: [0, paddingAmount], animate: false});
+			// Calculate the distance moved during the pan
+			var newCenter = map.project(map.getCenter());
+			var offset = newCenter.subtract(oldCenter);
+			// The target was placed 5 pixels above center and the padding was 15 pixels
+			// above center, so the Y distance should be 15 - 5 = 10 pixels
+			expect(offset.y - 10).to.be.lessThan(1);
+			expect(offset.x).to.be.lessThan(1);
```
johnd0e, line 1110:
```suggestion
			var targetOffset = L.point(0, -5); // arbitrary point above center
			var target = oldCenter.add(targetOffset);
			var paddingOffset = L.point(0, 15);
			var padding = map.getSize().divideBy(2) // half size
			  .add(paddingOffset); // padding more than a half the display bounds (replicates issue #7445)
			map.panInside(map.unproject(target), {paddingBottomRight: padding, animate: false});
			var offset = map.project(map.getCenter()).subtract(oldCenter); // distance moved during the pan
			var result = paddingOffset.add(targetOffset).subtract(offset);
			expect(result.trunc()).to.be.eql(L.point(0,0));
```
I'd prefer less comments, but the code that will speak for itself.
(I'm sure that my code also can be improved)

daverayment: Thanks for the review and suggestions. I didn't know about `trunc()`!

It looks like your code changes the `paddingBottomRight` value from my original intention, however. This is supposed to have an X value of 0, but it will end up being half the display bounds' width in your case. The test still passes because the target point sits in the horizontal centre, too, but it may be clearer to set X and Y separately, e.g.:

`map.panInside(map.unproject(target), {paddingBottomRight: [0, padding.y], animate: false});`

I'm happy to accept the changes as-is, too. Just let me know.

johnd0e: > This is supposed to have an X value of 0

Feel free to modify your test as you want, as you know your code (and edge cases) better.

johnd0e (approved):

PR #7477, opened 2021-02-23
fixed mouse scroll wheel zooming map two steps on Chrome on Linux with HiDPI displays

### LINKED ISSUE #7403: Mouse wheel scrolling results in 2 step zoom

**Steps to reproduce**
- setup 4k (UHD) monitor on your system (in Xfce settings / Apperance / set Window scaling to 2x)
- open Chrome and go to https://leafletjs.com/examples/quick-start/
- move scroll wheel just a single step

**Expected behavior**

Zoom should be changed by value of 1

**Current behavior**

Zoom is changed by value of 2

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Google Chrome 87.0.4280.141
- OS/Platform (with version): Debial Bullseye Linux, Xfce 4.16

**Additional context**

- window.devicePixelRatio is 1.875
- no problem in Firefox (where window.devicePixelRatio is 1.2)
- zooming page to 50% (where Chrome then reports devicePixelRatio to be 0.9375) doesn't help
- see also https://github.com/Leaflet/Leaflet/issues/4538#issuecomment-720090197
- also reproducible in MS Edge 89.0.752.1 

https://leafletjs.com/examples/quick-start/example-basic.html

Please create an example using https://leafletjs.com/edit.html or any other jsfiddle-like site.

zdila:
fixes #7403

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -64,6 +64,9 @@ export var opera12 = 'OTransition' in style;
 // @property win: Boolean; `true` when the browser is running in a Windows platform
 export var win = navigator.platform.indexOf('Win') === 0;
 
+// @property linux: Boolean; `true` when the browser is running in a Linux platform
+export var linux = navigator.platform.indexOf('Linux') === 0;
+
 // @property ie3d: Boolean; `true` for all Internet Explorer versions supporting CSS transforms.
 export var ie3d = ie && ('transition' in style);
 
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -233,7 +233,8 @@ export function getMousePosition(e, container) {
 // and Firefox scrolls device pixels, not CSS pixels
 var wheelPxFactor =
 	(Browser.win && Browser.chrome) ? 2 * window.devicePixelRatio :
-	Browser.gecko ? window.devicePixelRatio : 1;
+	(Browser.linux && Browser.chrome || Browser.gecko) ? window.devicePixelRatio :
+	1;
 
 // @function getWheelDelta(ev: DOMEvent): Number
 // Gets normalized wheel delta from a wheel DOM event, in vertical


johnd0e:
I do not like idea about introducing new `Browser.linux` property.

What about other cases?
- [ ] Browser.android && Browser.chrome
- [ ] Browser.safari 
- [ ] Browser.edge 
- [ ] Browser.opera

zdila:
This issue fixes the problem on Linux Chrome. If somebody finds issue on other platform, he may add another pull request. I don't have means to test all existing platforms.

If `Browser.linux` is a problem then OS check can be inlined directly to the condition.

johnd0e:
> . I don't have means to test all existing platforms.

I understand that you just want quick fix for your issue.
Still, here is proper place for discussion, and someone else may contribute too.

> If `Browser.linux` is a problem

It is.

> then OS check can be inlined directly to the condition.

Sure. Anyway, before it get merged I'd like to see opinions of other maintainers.

zdila:
BTW may I ask why `Browser.linux` is a problem but `Browser.win` is not?

johnd0e:
> `Browser.win` is not?

Many properties such as `Browser.win` were essential in the past.
But all the browsers are must more standardized since then.
And in general we should prefer feature detection approach (over legacy UA sniffing etc).

Answering to your question: I suppose that we do not need `Browser.win` either.
But since it is already documented - we have to support it (at least till Leaflet v2), just for compatibility reasons.

zdila:
Patch needs to be updated, please see https://github.com/Leaflet/Leaflet/issues/7403#issuecomment-931490795

johnd0e:
> Patch needs to be updated, please see [#7403 (comment)](https://github.com/Leaflet/Leaflet/issues/7403#issuecomment-931490795)

Are you going to update it?

zdila:
And will you merge it then? ;-)

zdila:
Anyway, maybe it deserves testing all common platforms (OS, browser, HiDPI vs not HiDPI) first. But I don't have them all.

jinchun-wang:
Is it possible to check zoomDelta in the following method in leaflet.js?
	
_performZoom: function () {
                ...

               if (!delta) { return; }

               // Fix the issue - https://github.com/Leaflet/Leaflet/issues/7403
               if (Math.abs(delta) > map.options.zoomDelta) {
                    delta = delta > 0 ? map.options.zoomDelta : -map.options.zoomDelta;
                }
  		if (map.options.scrollWheelZoom === 'center') {
  			map.setZoom(zoom + delta);
  		} else {
  			map.setZoomAround(this._lastMousePos, zoom + delta);
  		}
  	}

marconiho:
// Maybe it can be solved


let snap = 0.5;
if (L.Browser.win) snap = 0.1;
mapRef.current.map = L.map('lmap', {
            //...,
            zoomSnap: snap, 
 });

PR #7480, opened 2021-02-25
License BSD-2 Missing Fix #7426

### LINKED ISSUE #7426: License missing from download and web page is not clear about which license is used
Please consider improving how the license is communicated.

I can't find any mention on leafletjs.com about the license. Maybe there is one, but my point is that it should be easy to find.

I did eventually find the LICENSE file in github, which is good, but it doesn't actually say which open-source license it is. Not knowing the text of all licenses exactly, I had to research and compare the text at https://opensource.org/licenses.

For example, the front page starts with this line currently:

> Leaflet is the leading open-source JavaScript library for mobile-friendly interactive maps. Weighing just about 39 KB of JS, it has all the mapping features most developers ever need.

This could be changed to insert this text after "open-source":

> (BSD 2-Clause license)

The text within parentheses would be a link to the LICENSE file in github.

Similar change could be made to the Readme in github.

Furthermore, the license stipulates that:

> Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

However, the download of leaflet.zip from leafletjs.com does not itself include the license. I think this is unfortunate, and that the download should include the LICENSE file to make it easier to comply with the terms of the license.

HRSArgyropoulos:
Added BSD-2 Clause licence reference in the docs (index.html) as well as in the github README.md linking to the LICENSE github markdown file. fixes #7426

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <img width="600" src="https://rawgit.com/Leaflet/Leaflet/master/src/images/logo.svg" alt="Leaflet" />
 
-Leaflet is the leading open-source JavaScript library for **mobile-friendly interactive maps**.
+Leaflet is the leading open-source ([BSD 2-Clause license][]) JavaScript library for **mobile-friendly interactive maps**.
 Weighing just about 39 KB of gzipped JS plus 4 KB of gzipped CSS code, it has all the mapping [features][] most developers ever need.
 
 Leaflet is designed with *simplicity*, *performance* and *usability* in mind.
@@ -20,6 +20,7 @@ and push the limits of what's possible with online maps!
 
 [![Build Status](https://travis-ci.org/Leaflet/Leaflet.svg?branch=master)](https://travis-ci.org/Leaflet/Leaflet)
 
+ [BSD 2-Clause license]: https://github.com/Leaflet/Leaflet/blob/master/LICENSE
  [contributors]: https://github.com/Leaflet/Leaflet/graphs/contributors
  [features]: http://leafletjs.com/#features
  [plugins]: http://leafletjs.com/plugins.html
--- a/docs/index.html
+++ b/docs/index.html
@@ -4,7 +4,7 @@ layout: v2
 
 <div class="announcement">Sep 4, 2020 — <a href="/2020/09/04/leaflet-1.7.1.html">Leaflet 1.7.1</a> has been released!</div>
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 <img width="600" src="https://rawgit.com/Leaflet/Leaflet/master/src/images/logo.svg" alt="Leaflet" />
 
-Leaflet is the leading open-source JavaScript library for **mobile-friendly interactive maps**.
+Leaflet is the leading open-source ([BSD 2-Clause license][]) JavaScript library for **mobile-friendly interactive maps**.
```
johnd0e, line 3:
```suggestion
Leaflet is the leading open-source ([BSD 2-Clause license]) JavaScript library for **mobile-friendly interactive maps**.
```

(not sure that it is best place to link license, but you do not need those empty brackets anyway.

HRSArgyropoulos: I am aware about the brackets, i was just trying to stick to the way the rest of the links were also written in the md.

cherniavskii (commented):
Hey there,
It looks good to me!

Although it would be great to leverage Github license feature!

I thought Github has trouble detecting license in leaflet repo because of file content/format or something, but turns out it gets recognised properly in newly created repo - see https://github.com/cherniavskii/github-license-test and screenshots below:

Screenshot 2021-03-03 at 23 40 04
Screenshot 2021-03-03 at 23 40 23

I'll probably get in touch with GitHub support later to clarify this - looks like a bug to me :)

docs/index.html
```diff
@@ -4,7 +4,7 @@
 
 <div class="announcement">Sep 4, 2020 — <a href="/2020/09/04/leaflet-1.7.1.html">Leaflet 1.7.1</a> has been released!</div>
 
-<p>Leaflet is the leading open-source JavaScript library for mobile-friendly interactive maps.
+<p>Leaflet is the leading open-source (<a title="Leaflet's open-source license, BSD 2-Clause" href="https://github.com/Leaflet/Leaflet/blob/master/LICENSE">BSD 2-Clause license</a>) JavaScript library for mobile-friendly interactive maps.
```
mourner, line 7: Please revert this change — we don't need license indication in the first sentence of the website.

mourner (changes requested):

docs/index.html
HRSArgyropoulos: Removed as requested. 

If you don't like it as is, could you propose where the license in the docs to be referenced? Like at the end of Get Involved maybe like this "Leaflet uses the BSD-2 Clause license" as before, with a link to the GitHub license?

mourner:
Not necessary because GitHub exposes License information prominently:

image

mourner:
After adding a tag at the top, it detects correctly now:

image

PR #7482, opened 2021-02-26
Map Examples Responsive Fix (#7420)

### LINKED ISSUE #7420: Examples are not mobile-friendly

_(Edited by @Malvoz)_

HRSArgyropoulos:
fixes #7420 where maps were expanding outside the container.
Had to remove fixed values in cloropleth and quick-start example.
Examples tested on localhost (ruby/jekyll/bundler).
Please let me if/what you do not like so i can come back to this if needed.

### DIFF
--- a/docs/_layouts/tutorial_frame.html
+++ b/docs/_layouts/tutorial_frame.html
@@ -19,7 +19,7 @@
 			margin: 0;
 		}
 		#map {
-			width: 600px;
+			width: 100%;
 			height: 400px;
 		}
 	</style>
--- a/docs/_layouts/tutorial_frame.html
+++ b/docs/_layouts/tutorial_frame.html
@@ -19,7 +19,8 @@
 			margin: 0;
 		}
 		#map {
-			width: 100%;
+			width: 600px;
+			max-width: 100%;
 			height: 400px;
 		}
 	</style>
--- a/docs/examples/layers-control/index.md
+++ b/docs/examples/layers-control/index.md
@@ -7,7 +7,7 @@ title: Layer Groups and Layers Control
 
 This tutorial will show you how to group several layers into one, and how to use the layers control to allow users to easily switch different layers on your map.
 
-{% include frame.html width="100%" height="400" url="example.html" %}
+{% include frame.html url="example.html" %}
 
 ### Layer Groups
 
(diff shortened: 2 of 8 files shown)


johnd0e:
Using term "Responsive Fix" do you mean that samples now adapts to screen width? Or anything else?

HRSArgyropoulos:
> Using term "Responsive Fix" do you mean that samples now adapts to screen width? Or anything else?

Yes, if you take a look at the examples now on the website (tutorials) it forces the page to be scrollable. So we restrain the maps with a max-width and give them 100% width of the container. And also for the maps to not be scrollable we also need to change the iframe width. The height will remain as is.

p.s. the pov/zoom height should probably taken into consideration here as the map size becomes smaller.

docs/_layouts/tutorial_frame.html
```diff
@@ -19,7 +19,7 @@
 			margin: 0;
 		}
 		#map {
-			width: 600px;
+			width: 100%;
```
Malvoz, line 22: If an initial width of 600px is desired (keeping `width: 600px` as currently set) then `max-width: 100%` could be used additionally to accomplish responsiveness of the maps.

docs/examples/layers-control/index.md
```diff
@@ -7,7 +7,7 @@ title: Layer Groups and Layers Control
 
 This tutorial will show you how to group several layers into one, and how to use the layers control to allow users to easily switch different layers on your map.
 
-{% include frame.html url="example.html" %}
+{% include frame.html width="100%" height="400" url="example.html" %}
```
Malvoz, line 10: In HTML it's invalid to use % in `width` and `height` attributes, although some browsers seem to respect it. Just an FYI.

docs/_layouts/tutorial_frame.html
HRSArgyropoulos: The default width was 600px so i thought we wanted to keep it that way and from then making it smaller. I changed the width to your proposal, the desired result is still the same so its cool :)

docs/examples/layers-control/index.md
HRSArgyropoulos: The was left from a test that i was doing with setting value to attributes of the map. Thanks for letting me know. Removed them and the example is working as the rest.

johnd0e:
@HRSArgyropoulos 
Please resolve conflicts

Falke-Design:
@Malvoz is this PR completly superseded by #7950? (Ignore the `mapid` changes)

Malvoz:
@Falke-Design Yes, I think https://github.com/Leaflet/Leaflet/issues/7420 involves a bit more than just responsive _maps_, so keeping that open however. 👍🏼

PR #7483, opened 2021-02-26
add check for map in _onclick handler in canvas

Jobbalob:
Really simple PR, adding a check to make sure this._map exists in the _onClick handler in Canvas, similar to the check in _onMouseMove below it.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -344,6 +344,7 @@ export var Canvas = Renderer.extend({
 	// so we emulate that by calculating what's under the mouse on mousemove/click manually
 
 	_onClick: function (e) {
+		if (!this._map) { return; }
 		var point = this._map.mouseEventToLayerPoint(e), layer, clickedLayer;
 
 		for (var order = this._drawFirst; order; order = order.next) {


johnd0e:
This PR does not make sense unless you will describe what exact issue are you trying to solve with it.
Please provide minimal sample code to let us to reproduce your problem.

Jobbalob:
Agree it needs more info, think I got a bit overexcited to have it working! 

The issue is this:
When map.remove() is called while there is a canvas layer present on the map, there is a brief period before the canvas' destruction where the canvas' click handler can still be called, even though its reference to the map in this._map has become null. This causes an error because the click handler then tries to call this._map.mouseEventToLayerPoint which is not a function of null.

I ran into this issue in my project because the map is frequently re-rendered in response to data changes caused by click events.

JSFiddle here, not yet been able to automate the clicking. Double clicking the rendered rectangle causes the error:
https://jsfiddle.net/JobbleFrobble/xs2kn95o/129/

johnd0e:
I see now.
But I'm still not convinced that this is proper way to solve the issue, because I do not know what exactly causes it.

Why this code is not called in time?
https://github.com/Leaflet/Leaflet/blob/32ea41baaa93adcefd34006a12ebe3b4ee7fda7d/src/layer/vector/Canvas.js#L70-L76

PR #7489, opened 2021-03-02
Update Map.Tap.js

markusuntera:
Simulated events originalEvent doesn't carry ctrlKey, altKey, shiftKey original values.


johnd0e:
1. Consider CI errors
2. Consider #7255 caused by Tap handler. It is about to be deprecated.
3. But your fix still can be useful for #7026 (haven't inspected that).

PR #7493, opened 2021-03-06
GeoImageViewer

hdersch:
Tutorial example and documentation about linking maps to photographs.


johnd0e:
This is about 3rd-party tool, and I really doubt that Leaflet should host related tutorial.
Proper place to mention this would be https://github.com/Leaflet/Leaflet/blob/master/docs/plugins.md

PR #7505, opened 2021-03-12
Fix DomUtil getPosition when provided EL is undefined

fouad-j:
Hello, 

I faced an issue with map.flyTo(position, zoom), after a quick investigation I found that getPosition try to call an attribute en undefined HtmlElement.

An extract from my code
```ts
  onMapReady(map: Map): void {
    this.map = map;
   map.flyTo([countryLocation.lat, countryLocation.lng], 7);
  }
```

Logs
```
TypeError: Cannot read property '_leaflet_pos' of undefined
    at getPosition (leaflet-src.js:2450)
    at NewClass._getMapPanePos (leaflet-src.js:4439)
    at NewClass._moved (leaflet-src.js:4443)
    at NewClass.getCenter (leaflet-src.js:3798)
    at NewClass.flyTo (leaflet-src.js:3330)
    at SafeSubscriber._next (map-shipment.component.ts:85)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:183)
    at SafeSubscriber.next (Subscriber.js:122)
    at Subscriber._next (Subscriber.js:72)
    at Subscriber.next (Subscriber.js:49)
```

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -23,6 +23,16 @@ describe('DomUtil', function () {
 		});
 	});
 
+	describe('#getPosition', function () {
+		it('returns x0 y0 as position when provided el is undefined', function () {
+			expect(L.DomUtil.getPosition(undefined)).to.eql({x:0, y:0});
+		});
+
+		it('returns x0 y0 as position when provided el doesnt have position', function () {
+			expect(L.DomUtil.getPosition(el)).to.eql({x:0, y:0});
+		});
+	});
+
 	describe('#addClass, #removeClass, #hasClass', function () {
 		it('has defined class for test element', function () {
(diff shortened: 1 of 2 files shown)


johnd0e:
Hi.
It's not clear what issue are you trying to fix.
I'm pretty sure that `L.DomUtil.getPosition(undefined)` should throw, that is expected behavior, and no need to fix it.

So you better find exact conditions when this issue occurs, perhaps there is a bug in some other place.

fouad-j:
Hello,

In the existing code, we have

```ts
export function getPosition(el) {
    return el._leaflet_pos || new Point(0, 0);
}
```

`el` could be null non ?

in any case that's what I got as an error

johnd0e:
> `el` could be null non ?

`el` cannot be `null`.

Issue #1813, opened 2013-06-30
BoxZoom should prevent hover events on layers


lotrmj:
Suppose I have some markers on the map. When I initialize boxzoom and my cursor moves over the marker, marker's mousover() event fires... Besides, boxzoom rectangle seems to be under the markers (which is not so great). I think it would be better if mouse cursor wouldn't interact with other elements on the map while drawing boxzoom rectangle. Or at least rectangle should be over other markers...

P.S. (related stuff): I used to create invisible overlay on dragstart(), which got removed on dragend() and it prevented mouse cursor from interacting with markers and so on while dragging map (and it worked flawlessly), but recently it stopped working on Firefox. Is it because of new FF 22 or did you make any core adjustments to map mouse drag event? :/
I see a lot of people overlook it and let mouse cursor interact with map while dragging map (which is usually useless, quite annoying and stupid)... Especially it's annoying when using markerclusterer: suppose you begin dragging map when cursor is near the circle and suddenly mouse cursor hovers the circle, which causes convex hull of circle points to render... not so great... user wanted just to drag a map!

PR #7507, opened 2021-03-14
Disable pointer-events of relevant descendants on drag and box zoom

Malvoz:
Close https://github.com/Leaflet/Leaflet/issues/1813.

The issue as described in https://github.com/Leaflet/Leaflet/issues/1813:

<table><td width="400"><video src="https://user-images.githubusercontent.com/26493779/111072383-ac192c80-84da-11eb-94fd-828acebcabba.mp4"></video></td></table>

## PR Preview

<table><td width="400"><video src="https://user-images.githubusercontent.com/26493779/111072403-c3f0b080-84da-11eb-98d2-7658297ac90e.mp4"></video></td></table>

(side note: we're using the same fix for our [`<mapml-viewer>`](https://github.com/Maps4HTML/Web-Map-Custom-Element) custom component.)


johnd0e:
Nice solution!

But your videos shown box-zoom only, but not drag.
Also I'm thinking about other potential issue - dragging marker over some feature with pointer listeners attached (e.g. with tooltip).

It'd be better if instead of video you setup some live sample, using plnkr/jsfiddle/whatever sandbox.
And it'd be perfect if we have unit test for all affected cases to prevent regression in future.

Malvoz:
@johnd0e I hadn't thought of draggable markers. To be fair I had never worked with maps before joining the [Maps for HTML CG](https://www.w3.org/community/maps4html/). 😋 I'm mostly a web standard enthusiast and accessibility advocate.

Due to your feedback I've added `leaflet-dragging` (which is active when dragging a marker) as a condition for disabling pointer-events of descendants for more robust handling in https://github.com/Leaflet/Leaflet/pull/7507/commits/3cbdd5c307cb06a082ea89d2ba8fcf15097d5a01.

But you're right, there are probably cases that this solution doesn't cover. I kind of have my plate full right now, but if anyone wishes to continue working on this and find this inspirational then that's great.

PR #7510, opened 2021-03-15
Make getPosition more robust so that it can also handle undefined

### LINKED ISSUE #7508: _handleGeolocationResponse: exception when map is already removed
See #2001

Seeing this error at 

[./src/dom/DomUtil.js:247](https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/dom/DomUtil.js#L247):	`return el._leaflet_pos || new Point(0, 0);`

googol7:
Fix #7508


johnd0e:
Duplicate of #7505

PR #7515, opened 2021-03-16
Update Renderer documentation to clarify tolerance option Canvas only

### LINKED ISSUE #7494: SVG renderer tolerance does not work for paths
- [x ] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [ x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [ x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
use custom renderer:
```
var renderer=L.svg({ padding: 0.5, tolerance: 10 });
 var map = L.map('map', {
            center: core.options.center,
            zoom: core.options.zoom,
            zoomSnap: 0.5,
            zoomDelta: 0.5,
            tap: false, // fixes Safari issues with popups
            renderer: renderer,
...
});

add renderer to polylines:
```
var polyline_options = {
            renderer: renderer
        };
polyline = L.polyline([nodes], polyline_options);
```

**Expected behavior**
<!-- A clear and concise description of what you expected to happen. -->
Touch events working on the set tolerance, eg. +10 pixels

**Current behavior**
<!-- A clear and concise description of what is actually happening. -->
no tolerance, path only clickable dirctly on their `width`

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Firefox 86
- OS/Platform (with version): Ubuntu 20.04

### PR DESCRIPTION
**Pull request by Hippl-Eric**
Fixes #7494 

- Renderer documentation does not clarify the [tolerance ](https://leafletjs.com/reference-1.7.1.html#renderer-tolerance ) option is only supported by Canvas (not SVG) renderer.
- Update documentation to reflect this.
- Refer to similar instance in documentation: Path [className](https://leafletjs.com/reference-1.7.1.html#path-classname) option

Close #7494.

### DIFF
--- a/src/layer/vector/Path.js
+++ b/src/layer/vector/Path.js
@@ -142,6 +142,7 @@ export var Path = Layer.extend({
 
 	_clickTolerance: function () {
 		// used when doing hit detection for Canvas layers
-		return (this.options.stroke ? this.options.weight / 2 : 0) + this._renderer.options.tolerance;
+		return (this.options.stroke ? this.options.weight / 2 : 0) +
+		(this._renderer.options.tolerance ? this._renderer.options.tolerance : 0);
 	}
 });
--- a/src/layer/vector/Path.js
+++ b/src/layer/vector/Path.js
@@ -143,6 +143,6 @@ export var Path = Layer.extend({
 	_clickTolerance: function () {
 		// used when doing hit detection for Canvas layers
 		return (this.options.stroke ? this.options.weight / 2 : 0) +
-		(this._renderer.options.tolerance ? this._renderer.options.tolerance : 0);
+		  (this._renderer.options.tolerance || 0);
 	}
 });
--- a/docs/reference-1.7.1.html
+++ b/docs/reference-1.7.1.html
@@ -22831,7 +22831,7 @@ e.g. 0.1 would be 10% of map view in each direction</td>
 		<td><code><b>tolerance</b></code></td>
 		<td><code>Number</code></td>
 		<td><code>0</code></td>
-		<td>How much to extend click tolerance round a path/object on the map</td>
+		<td>How much to extend click tolerance around a path/object on the map. Only for <a href="#canvas"><code>Canvas</code></a>.</td>
 	</tr>
 </tbody></table>
 
--- a/docs/reference-1.7.1.html
+++ b/docs/reference-1.7.1.html
@@ -22831,7 +22831,7 @@ e.g. 0.1 would be 10% of map view in each direction</td>
 		<td><code><b>tolerance</b></code></td>
 		<td><code>Number</code></td>
 		<td><code>0</code></td>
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by johnd0e** [comment]
Our documentation is autogenerated from source comments, so the fix should be in `Renderer.js`.

> * Refer to similar instance in documentation: Path [className](https://leafletjs.com/reference-1.7.1.html#path-classname) option

Not exactly a good example for this case.

I have in mind 2 alternative ways:
1. Note that `tolerance` is defined not in `Canvas.js`. So simplest fix would be just move it there.
2. Or we can consider this as generic renderer option, which is subject to be (or not) implemented in respective renderers.
   This case requires more general wording, saying that the option requires support from renderer, and SVG lacks one.

In fact I'm not sure what way should be chosen.
May be let's start from 1 and wait for other opinions.

**Comment by Hippl-Eric** [comment]
> Our documentation is autogenerated from source comments, so the fix should be in `Renderer.js`.

I reverted commit to incorrect file `reference-1.7.1.html`.  Awaiting direction before modifying `Renderer.js`.

> In fact I'm not sure what way should be chosen.
> May be let's start from 1 and wait for other opinions.

I can work through implementing 1, but can you please clarify what you are expecting?  As you point out, `tolerance` is not defined in `Canvas.js` - it is defined in `Renderer.js`:
https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/layer/vector/Renderer.js#L39-L41
`Canvas` does not have an `options` property (only inherited).  Are you suggesting to add an `options` property to `Canvas` with one nested property `tolerance : 0`?  Then removing the `tolerance` from `Renderer`?

**Comment by johnd0e** [comment]
> Canvas does not have an options property (only inherited). Are you suggesting to add an options property to Canvas with one nested property tolerance : 0? Then removing the tolerance from Renderer?

That is exactly what I mean, right.

**Comment by Hippl-Eric** [comment]
@johnd0e I completed the following changes:

1. Moved tolerance option from `Render` to `Canvas`
2. Updated `Path_clickTolerance` to handle possible renderers that do include a tolerance option (all `SVG` renderers now)

File: src/layer/vector/Path.js [hunk 1]
```diff
@@ -142,6 +142,7 @@ export var Path = Layer.extend({
 
 	_clickTolerance: function () {
 		// used when doing hit detection for Canvas layers
-		return (this.options.stroke ? this.options.weight / 2 : 0) + this._renderer.options.tolerance;
+		return (this.options.stroke ? this.options.weight / 2 : 0) +
+		(this._renderer.options.tolerance ? this._renderer.options.tolerance : 0);
```
johnd0e, line 146:
```suggestion
		  (this._renderer.options.tolerance || 0);
```

johnd0e (changes requested):

johnd0e (approved):

PR #7518, opened 2021-03-18
Evented: protect from common errors (+tests)

johnd0e:
Such as:
1. wrong value passed as listener `fn`
   - in `on/once`: errors will throw too late (in fire), and it may be tricky to find exact source
   - in `off`: wrong (false) value may lead to removing all the listeners, this case is also not easy to debug
2. wrong value passed as `types`
    - in `off`: wrong (false) value may lead to removing all the listeners at once too

All such cases now handled gracefully and without much overhead: either exception is thrown, or a warning is printed in console.
Warnings approach was chosen for compatibility reasons, but may be we should prefer stricter handling in some cases - throw exceptions (this is approach that I preferred in similar #7125 about `DomEvent`).

So I'm waiting for maintainers' advice.

---
This change is [Reviewable](https://reviewable.io/reviews/leaflet/leaflet/7518)

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -387,6 +387,16 @@ describe('Events', function () {
 			expect(obj.listens('test')).to.be(true);
 		});
 
+		it('ignores non-function listeners passed', function () {
+			var obj = new L.Evented();
+			var off = obj.off.bind(obj);
+			['string', {}, [], true, false, undefined].forEach(function (fn) {
+				obj.on('test', fn);
+				expect(obj.listens('test')).to.be(false);
+				expect(off).withArgs('test', fn).to.not.throwException();
+			});
+		});
+
 		it('works like #addEventListener && #removeEventListener', function () {
 			var obj = new L.Evented(),
 			    spy = sinon.spy();
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -397,6 +397,19 @@ describe('Events', function () {
 			});
 		});
 
+		it('throws with wrong types passed', function () {
+			var obj = new L.Evented();
+			var on = obj.on.bind(obj);
+			var off = obj.off.bind(obj);
+			// todo? make it throw  with []
+			[true, false, undefined, 1].forEach(function (type) {
+				expect(on).withArgs(type, L.Util.falseFn).to.throwException();
+				expect(off).withArgs(type, L.Util.falseFn).to.throwException();
+			});
+
+			// todo? make `fire` and `listen` to throw with wrong type
+		});
+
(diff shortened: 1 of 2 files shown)


johnd0e:
There are some unchecked cases left:
- `on`/`off` still accept array (as it is `object` type)
- `fire`/`listen` do not throw on wrong 'type'

mourner (changes requested):
As far as I remember, I explicitly avoided using `arguments` throughout the code because it disabled optimizations in many JS engines (definitely V8 but possibly others), and would impact performance. It may not be the case anymore, but since Leaflet supports so many legacy browsers, maybe we should be on the safe side and continue avoiding it?

Falke-Design:
From a blog post in 2021: https://v8.dev/blog/adaptor-frame#conclusion
> The arguments adaptor frame was an ad-hoc solution to calls with a mismatch number of arguments and formal parameters. It was a straightforward solution, but it came with high performance cost and added complexity to the codebase. The performance cost is nowadays exacerbated by many Web frameworks using this feature to create a more flexible API. The simple idea of reversing the arguments in the stack allowed a significant reduction in implementation complexity and removed almost the entire overhead for such calls.

Because it is always possible to check if the next passed variable is undefined, we can avoid using `arguments`

johnd0e:
> As far as I remember, I explicitly avoided using `arguments` throughout the code because it disabled optimizations in many JS engines (definitely V8 but possibly others), and would impact performance

Any proof?
AFAIK that was fixed by `strict` mode, in [this way](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode/Transitioning_to_strict_mode):
> **`arguments` doesn't alias named function arguments**
In non-strict mode, modifying a value in the arguments object modifies the corresponding named argument. This made optimizations complicated for JavaScript engine and made code harder to read/understand. In strict mode, the arguments object is created and initialized with the same values than the named arguments, but changes to either the arguments object or the named arguments aren't reflected in one another.

> From a blog post in 2021

Not relevant at all here.

> Because it is always possible to check if the next passed variable is `undefined`

Not really helpful, as there may be a lot of reasons why some arg has `undefined` value. E.g. bug in user code.

johnd0e:
- I believe that it is very important to validate data in event processing, because it is the source of common mistakes, which are hard to debug in any other place.
- It is important to check if some arg was omitted. And there is a big difference between cases where it was not passes at all (we can check with `arguments.length`), or `undefined` was passed (which most likely is a bug). **I do not know other way to do this, except checking `arguments.length`.**
- I failed to find any reference that using `arguments` has any performance loss.

mourner (approved):
👍 `on` and `off` don't seem to be on a hot path anyway (not called often) unlike something like `fire`, so even it is a perf hit (searching for it seems difficult, although I remember it being the case in the past), it doesn't matter much.

Issue #7302, opened 2020-10-04
Tooltips with direction option not longer working correct


Falke-Design:
In the newest Version of Leaflet 1.7.1 the Tooltips are not displayed correct, if a direction is set.

**To reproduce:**
Go to: https://leafletjs.com/examples/quick-start/example.html
And add this in the console: 
```
var marker = L.marker([51.52,-0.092]).addTo(mymap)
marker.bindTooltip('TEST', {
  direction: 'left'
}).openTooltip()
```

The Tooltip should completly displayed on the left side of the marker but it is starting on the right side and then goes to the left side.

I think it has something to do with: #7155

![tooltip](https://user-images.githubusercontent.com/19800037/95018006-0628e280-065d-11eb-94d2-42976a14ba45.PNG)

johnd0e:
@Istador
@mourner

Istador:
That's the expected behavior.

The `tooltipAnchor` for the [Icon of the default marker](https://github.com/schloerke/Leaflet_js/blob/9084c4a8a3c6869df8fbb1ad15e23c1fad2a4193/src/layer/marker/Icon.Default.js#L29) that is used is at `[16, -28]`, so horizontally to the right of it and vertically in the middle. `{direction: 'left'}` means that the tooltip should open to the left of the anchor, which it does.

The default marker uses the `auto` direction and not `left`, where it automatically inverts the x axis to the left of the marker if needed.

If you want to use another direction for the default marker, you should provide it with an `offset`:
```js
var marker = L.marker([51.52,-0.092]).addTo(mymap)
marker.bindTooltip('TEST', {
  direction: 'left',
  offset: [ -32, 0 ]
}).openTooltip()
```

Alternatively, you could also create another (default) Icon with a different default value for its `tooltipAnchor`

Falke-Design:
Why you think this is the right way? This is a big change for so many projects! And now every project have to edit their code for something what is for the most users so trivial.
I can understand that this is a wanted feature but that all project now have to change their icons / options of the tooltip can't be right.

Istador:
I'd argue that it was working incorrectly and inconsistently before.
And that it's totally normal, with every dependency that you update that you might need to migrate/adopt your application to it.

But I do see that there's a potential need for this kind of behavior of inverted axes.

Maybe something like new directions: `invertedLeft`, `invertedRight`, `invertedTop` and `invertedBottom`, which invert the x- or y-axis for the anchor and the offset.
[...]

PR #7519, opened 2021-03-20
Fix tooltip position

ronikar:
Hi guys,
I fixed tooltip position when the direction of the tooltip is top or bottom and anchor isn't ` {x : 0, y: 0} ` 
 
**Before**
image

**After**
image

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -143,11 +143,11 @@ export var Tooltip = DivOverlay.extend({
 		    anchor = this._getAnchor();
 
 		if (direction === 'top') {
-			subX = tooltipWidth / 2;
+			subX = tooltipWidth / 2 + anchor.x;
 			subY = tooltipHeight;
 		} else if (direction === 'bottom') {
-			subX = tooltipWidth / 2;
-			subY = 0;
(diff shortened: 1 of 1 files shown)


Falke-Design:
Linking #7302

src/layer/Tooltip.js
```diff
@@ -143,11 +143,11 @@ export var Tooltip = DivOverlay.extend({
...
-			subX = tooltipWidth / 2;
+			subX = tooltipWidth / 2 + anchor.x;
 			subY = tooltipHeight;
 		} else if (direction === 'bottom') {
-			subX = tooltipWidth / 2;
-			subY = 0;
+			subX = tooltipWidth / 2 + anchor.x;
+			subY = anchor.y;
 		} else if (direction === 'center') {
 			subX = tooltipWidth / 2;
```
itamair, line 152: We need to also change this into:

`subX = tooltipWidth / 2 + anchor.x;`

to properly x align the tooltip also for the "center" use case

itamair: With the above amend this looks good to me for merging into main branch.

itamair (changes requested):

PR #7532, opened 2021-03-28
Make tooltips interactive themselves

johnd0e:
`mytooltip.on('click', ...`

---
_More ideas:_
<details>

Close on `preclick` is not good..
Perhaps better something like this:
```js
		if (!this.options.permanent) {
			events.tooltipopen = L.Util.bind(function (e) {
				if (e.tooltip !== this) { this._close(); }
			}, (this));
		}
```
</details>


src/layer/Tooltip.js
```diff
@@ -70,13 +70,20 @@ export var Tooltip = DivOverlay.extend({
 		DivOverlay.prototype.onAdd.call(this, map);
 		this.setOpacity(this.options.opacity);
 
+		if (this.options.interactive) {
+			DomUtil.addClass(this._container, 'leaflet-clickable');
```
Falke-Design, line 74: New class `leaflet-interactive` should be used

src/layer/Tooltip.js
```diff
@@ -88,13 +95,20 @@ export var Tooltip = DivOverlay.extend({
 	onRemove: function (map) {
 		DivOverlay.prototype.onRemove.call(this, map);
 
+		if (this.options.interactive) {
+			DomUtil.removeClass(this._container, 'leaflet-clickable');
```
Falke-Design, line 99: New class `leaflet-interactive` should be used

Falke-Design (changes requested):
Because of https://github.com/Leaflet/Leaflet/pull/7531#issuecomment-955696600:
The events on the tooltip is not working because of:
https://github.com/Leaflet/Leaflet/blob/f018e2de2ce507ab9b070f382e6bded9b9311263/src/layer/Tooltip.js#L71-L76

it adds the interactiveTarget to the `source` and not to the layer self. This is fixed with this PR additionally it adds the event parent `source` which makes the changes not to a breaking change.

I came after my research to the exact same changes

Falke-Design:
And maybe `this._source.fire('tooltipclose', {tooltip: this}, true);` should be then fired on the tooltip self and with the event parent it is also fired on the source layer

src/layer/Tooltip.js
```diff
@@ -69,10 +69,8 @@ export var Tooltip = DivOverlay.extend({
 		this.setOpacity(this.options.opacity);
 
 		if (this.options.interactive) {
+			DomUtil.addClass(this._container, 'leaflet-clickable');
 			DomUtil.addClass(this._container, 'leaflet-interactive');
```
Falke-Design, line 73: `leaflet-clickable` is still added

Falke-Design (changes requested):

Falke-Design (approved):
Looks good! Next step to mark PR "ready for review"

johnd0e:
> And maybe `this._source.fire('tooltipclose', {tooltip: this}, true);` should be then fired on the tooltip self and with the event parent it is also fired on the source layer

Please explain what you mean

Falke-Design:
With `this._source.fire('tooltipclose', {tooltip: this}, true);` the event `tooltipclose` is only fired on the source layer, but it would make sense to fire this on the tooltip too. So it should changed to `this.fire('tooltipclose', {tooltip: this}, true);`

johnd0e:
> So it should changed to `this.fire`

No, should fire on source layer - it is really intended so.

johnd0e:
> Next step to mark PR "ready for review"

Next step is decide how to be with closing on `preclick`.

P.S.
And final step - describe possible use cases. Do you have some?

Falke-Design:
> No, it should fire on source layer - it is really intended so.

But why? A tooltip can also be a own layer on the map, so why not fireing events on it?

```
		var tooltip = L.tooltip().setContent('Tooltip');
		map.openTooltip(tooltip, map.getCenter());
		
		tooltip.on('tooltipclose',(e)=>console.log('Tooltip',e));
		map.on('tooltipclose',(e)=>console.log('Map',e));
```

johnd0e:
> But why? A tooltip can also be a own layer on the map, [...]

So it already has all related events ('remove', not 'tooltipclose')

Falke-Design:
> So it already has all related events ('remove', not 'tooltipclose')

Ok yes sorry, you are right didn't thought about that one.

> Next step is decide how to be with closing on preclick.

Ah sorry, didn't thought that you want to add this in this PR.
I don't think that new function instead of  `preclick` will be good, because it breaks current behaviour. It doesn't close anymore on click on the map or on the tooltip. It closes only if a new tooltip is open or `closeTooltip()` is called.

> And final step - describe possible use cases. Do you have some?

I know that had some usecase where I wanted to add a click listener on the tooltip. But I don't know for what exactly anymore. 
Other usecases would be to make Tooltips dragable, specially if they are added as a own layer without "source". Or to make a "Click me to revert" Tooltip after removing a layer, ...

johnd0e:
> It doesn't close anymore on click on the map or on the tooltip.

It will be easy to add too, but with clear map `click`.

> Ah sorry, didn't thought that you want to add this in this PR.

But you are right, it is not necessary in this PR.

> I know that I had some usecase

Me too)

Falke-Design (approved):

mourner (approved):
Nice! How do we surface this in the docs?

Falke-Design:
Good point.

This section is missing:
image

johnd0e:
May be this way: 9251b807f70ca68c41217df8201ac7938b6c9743

Falke-Design:
But then DivOverlay and Popup would be "marked" as interactive too, but the code is missing. `popup.on('click',(e)=>console.log(e))` is not working.
```
var layer = new L.Marker(map.getCenter()).bindPopup('TEST').addTo(map);
layer.getPopup().on('click',(e)=>console.log(e))
```
https://jsfiddle.net/falkedesign/xqc9mbzd/

johnd0e:
Perhaps not every descendant of `DivOverlay` must to be `interactive`.
But I do not know if **LeafDoc** has ability to attach `Interactive layer` class 'in the middle', so..

@mourner @IvanSanchez 
Ideas?

johnd0e:
@Falke-Design 
Something like this?

Falke-Design (approved):
Nice!

Good to know: `click` event will not fire on the Popup because of:
https://github.com/Leaflet/Leaflet/blob/cc9f327b41d5a283af40beae195d196ca4089599/src/layer/Popup.js#L184

johnd0e:
> Good to know: `click` event will not fire on the Popup

Because of #7750.

Falke-Design:
@johnd0e can you please resolve the conflicts and mark the PR as ready for review. I would say we keep this PR in the 1.8.0 milestone, because it was already approved by @mourner and then it is a small task to check the changes since them and then merge it.

Falke-Design (approved):
@mourner valid for me, the tests are covering all changes

PR #7535, opened 2021-03-29
Tooltip: stop relying on Browser `touch` property

johnd0e:
Rationale: #6978

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -1,5 +1,3 @@
-
-import * as Browser from '../core/Browser';
 import {DivOverlay} from './DivOverlay';
 import {toPoint} from '../geometry/Point';
 import {Map} from '../map/Map';
@@ -106,7 +104,7 @@ export var Tooltip = DivOverlay.extend({
 	getEvents: function () {
 		var events = DivOverlay.prototype.getEvents.call(this);
 
-		if (Browser.touch && !this.options.permanent) {
+		if (!this.options.permanent) {
 			events.preclick = this._close;
 		}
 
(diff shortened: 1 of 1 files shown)


johnd0e:
<details>

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/layer/Tooltip.js#L106-L115

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/src/layer/Tooltip.js#L304-L320
</details>

In fact for modern browsers we could avoid handling clicks, and rely just on compatibility `mouseover`/`mouseout` even for touch.

johnd0e:
Going beyond subject: I'd prefer tooltips to behave in slightly different way: if opened on `click` - do not close on `mouseout`.

Reason: while such behavior is quite similar to current - it is more flexible, as it would also allow to keep tooltip opened, for convenience.

In this way we could use non-permanent tooltips with some interactive content (e.g. html link).

Falke-Design:
Do you think `Browser.touch` was used here because of performance or because of a other reason?

> Currently in most places of code `touch` property check is used to prevent attaching excessive listeners if touch is unsupported.
> May be that makes sense for legacy devices (though I doubt that it has any impact on performance with modern ones).

_Originally posted by @johnd0e in https://github.com/Leaflet/Leaflet/issues/6978#issuecomment-607261210_

johnd0e:
> Do you think `Browser.touch` was used here because of performance or because of a other reason?

I think that mentioned code was written with some assumptions in mind, which used to be true in that times.
But things are changing.

No, I do not think that it's about performance here.
It's rather about clear code logic - "if we use touch device than we need additional handling".
It's convenient logic but nowadays we are unable to say for true if touch device be used or not.

Falke-Design (approved):
I didn't tested it but looks good

Falke-Design:

mourner (approved):

PR #7536, opened 2021-03-29
Tooltip/Popup: remove redundant checks

johnd0e:
Removing all this over-protection we demand user code to be stricter, but with benefit instead - bugs will manifest themselves earlier.

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -416,12 +416,10 @@ Layer.include({
 	// @method openPopup(latlng?: LatLng): this
 	// Opens the bound popup at the specified `latlng` or at the default popup anchor if no `latlng` is passed.
 	openPopup: function (layer, latlng) {
-		if (this._map) {
-			latlng = this._popup._prepareOpen(this, layer, latlng);
+		latlng = this._popup._prepareOpen(this, layer, latlng);
 
-			// open the popup on the map
-			this._map.openPopup(this._popup, latlng);
-		}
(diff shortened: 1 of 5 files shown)


src/layer/Popup.js
```diff
@@ -401,67 +401,56 @@ Layer.include({
...
 	},
 
 	// @method closePopup(): this
 	// Closes the popup bound to this layer if it is open.
 	closePopup: function () {
-		if (this._popup) {
-			this._popup._close();
-		}
-		return this;
+		return this._popup._close();
```
Falke-Design, line 430: I think we should keep checking `if (this._popup) {` because logics which want to make sure that popups get closed are breaking now.

src/layer/Popup.js
```diff
@@ -401,67 +401,56 @@ Layer.include({
...
+			this.openPopup(target);
 		}
 		return this;
 	},
 
 	// @method isPopupOpen(): boolean
 	// Returns `true` if the popup bound to this layer is currently open.
 	isPopupOpen: function () {
-		return (this._popup ? this._popup.isOpen() : false);
+		return this._popup.isOpen();
```
Falke-Design, line 447: Should be possible to call if no popup is existing

Falke-Design, line 447: Should be possible to call if no popup is existing

Falke-Design (changes requested):

PR #7540, opened 2021-04-03
DivOverlay/Popup/Tooltip refactoring and fixes

johnd0e:
This is continuation of #6613.

- Fix a couple of edge cases
- Make `_prepareOpen`-related code logic more clear
- Allow to prevent overlay opening gracefully on some conditions (like empty `FeatureGroup`)
- Refactor to avoid of undocumented arguments for public methods

### DIFF
--- a/src/layer/DivOverlay.js
+++ b/src/layer/DivOverlay.js
@@ -160,13 +160,8 @@ export var DivOverlay = Layer.extend({
 	},
 
 	// prepare bound overlay to open: update latlng pos / content source (for FeatureGroup)
-	_prepareOpen: function (source, latlng) {
-		if (source instanceof Layer) {
-			this._source = source;
-		} else {
-			latlng = source;
-			source = this._source;
-		}
+	_prepareOpen: function (latlng) {
+		var source = this._source;
 		if (!source._map) { return; }
 
 		if (source instanceof FeatureGroup) {
--- a/src/layer/DivOverlay.js
+++ b/src/layer/DivOverlay.js
@@ -159,38 +159,50 @@ export var DivOverlay = Layer.extend({
 		return this;
 	},
 
-	_prepareOpen: function (parent, layer, latlng) {
-		if (!(layer instanceof Layer)) {
-			latlng = layer;
-			layer = parent;
+	// prepare bound overlay to open: update latlng pos / content source (for FeatureGroup)
+	_prepareOpen: function (source, latlng) {
+		if (source instanceof Layer) {
+			this._source = source;
+		} else {
+			latlng = source;
+			source = this._source;
 		}
+		if (!source._map) { return; }
 
-		if (layer instanceof FeatureGroup) {
-			for (var id in parent._layers) {
-				layer = parent._layers[id];
-				break;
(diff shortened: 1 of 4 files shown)


src/layer/DivOverlay.js
```diff
@@ -159,38 +159,45 @@ export var DivOverlay = Layer.extend({
...
-		}
-
-		if (layer instanceof FeatureGroup) {
-			for (var id in parent._layers) {
-				layer = parent._layers[id];
-				break;
+	// prepare bound overlay to open: update latlng pos / content source (for FeatureGroup)
+	_prepareOpen: function (latlng) {
+		var source = this._source;
+		if (!source._map) { return; }
```
Falke-Design, line 165: Should we return `false`?

src/layer/DivOverlay.js
```diff
@@ -159,38 +159,45 @@ export var DivOverlay = Layer.extend({
...
+		if (source instanceof FeatureGroup) {
+			source = null;
+			var layers = this._source._layers;
+			for (var id in layers) {
+				if (layers[id]._map) {
+					source = layers[id];
+					break;
+				}
 			}
+			if (!source) { return; } // Unable to get source layer.
```
Falke-Design, line 176: Same `return false`?

src/layer/Popup.js
```diff
@@ -438,12 +436,12 @@ Layer.include({
 
 	// @method togglePopup(): this
 	// Opens or closes the popup bound to this layer depending on its current state.
-	togglePopup: function (target) {
+	togglePopup: function () {
 		if (this._popup) {
 			if (this._popup._map) {
```
Falke-Design, line 441: Here is the check if the popup has the map property `this._popup._map` and in `_openPopup` the check is `if (this._map.hasLayer(this._popup))`

Falke-Design (commented):
A few things but looks good for me

src/layer/DivOverlay.js
```diff
@@ -159,38 +159,45 @@ export var DivOverlay = Layer.extend({
...
+				latlng = source.getCenter();
+			} else if (source.getLatLng) {
+				latlng = source.getLatLng();
+			} else if (source.getBounds) {
+				latlng = source.getBounds().getCenter();
 			} else {
 				throw new Error('Unable to get source layer LatLng.');
 			}
 		}
+		this.setLatLng(latlng);
```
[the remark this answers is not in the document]
johnd0e: Explain please

src/layer/Popup.js
johnd0e: Yes, but that is necessary in both places

src/layer/DivOverlay.js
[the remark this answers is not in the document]
Falke-Design: `setLatLng` calls `_adjustPan` and in the Popup a event is fired: 
https://github.com/Leaflet/Leaflet/blob/f018e2de2ce507ab9b070f382e6bded9b9311263/src/layer/Popup.js#L273-L277

But I don't think that this will make problems

Falke-Design (approved):

mourner (approved):

Issue #5720, opened 2017-08-18
Wrong relatedTarget on mouseout event (to close tooltip except if mouseout towards tooltip)


robertomlsoares:
## How to reproduce

- Leaflet version I'm using: 1.1.0
- Browser (with version) I'm using: IE 11 or Firefox 54
- OS/Platform (with version) I'm using: Windows 7

Open the fiddles below and hover over the markers while checking the console. The fiddle with Leaflet 0.5.1 works as expected but the one with Leaflet 1.1.0 doesn't,

## What behaviour I'm expecting and which behaviour I'm seeing

The `relatedTarget` should be the popup but it's another element (the map container).

## Minimal example reproducing the issue

This fiddle uses Leaflet 0.5.1 and works as expected (check the console while hovering over markers or simply notice that the popups are not closed when going with the mouse from the marker to the popup): http://jsfiddle.net/3JbNY/106/

This one is exactly the same as the one above but it uses Leaflet 1.1.0 and as you can see, it doesn't work anymore; and if you check the console, the `relatedTarget` element is not the popup on 90% of the times (very rarely it actually is the popup): http://jsfiddle.net/3JbNY/105/

What is the correct behavior here? Why was that changed? Shouldn't the `relatedTarget` on `mouseout` always be the element you're exiting into?

PS.: I asked a question on SO about this, check it out if you want more information or if you know a solution that doesn't rely on `relatedTarget`. https://stackoverflow.com/questions/45755984/why-doesnt-leaflet-provide-the-correct-relatedtarget-element-on-mouseover-event

robertomlsoares:
Ok, so since 0.5.1 is quite an old version I decided to go check which one is the latest working version for this. It still behaves correctly in the v.1.0.0-rc.1 version but starts breaking in the v.1.0.0-rc.2 version.

I'll try to find what caused the change between those two versions.
Does anyone agree that this behavior is odd? I can help find and/or fix the problem but if the maintainers think that this is not a bug then I can move on to trying to find another solution.

robertomlsoares:
So, I think this is caused by this change: #4599

On version 1.1.0, I changed the value of `pointer-events` to `auto` on the CSS class `leaflet-tooltip` (**not** the pane) and it works as I expected.
Now remains the question of what should be the desired outcome here.

perliedman:
[...]

PR #7541, opened 2021-04-05
Prevent click on popup tip from firing on map.

### LINKED ISSUE #4596: Click on leaflet-popup-tip-container propagates on map and not on underlying object

johnd0e:
Fix #4596 / #5720

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -495,6 +495,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	padding: 1px;
 
 	margin: -10px auto 0;
+	pointer-events: auto;
 
 	-webkit-transform: rotate(45deg);
 	   -moz-transform: rotate(45deg);


Falke-Design (changes requested):
I suggest to add this test:

```
	it("Don't close Popup when clicking on tip", function () {
		var popup = L.popup().setLatLng(map.getCenter())
			.setContent('<p>Hello world!<br />This is a nice popup.</p>')
			.openOn(map);

		expect(popup.isOpen()).to.be(true);

		var point = map.latLngToContainerPoint(map.getCenter());
		point.y -= 2; // move mouse into the popup-tip

		const ev = new MouseEvent('click', {
			'view': window,
			'bubbles': true,
			'cancelable': true,
			'clientX': point.x,
			'clientY': point.y
		});
		const el = document.elementFromPoint(point.x, point.y);
		el.dispatchEvent(ev);

		expect(popup.isOpen()).to.be(true);
	});
```
and change the `setup()` function:
```
function setup() {
	c = document.createElement('div');
	c.style.width = '400px';
	c.style.height = '400px';
	c.style.position = 'absolute';
	c.style.top = '0';
	c.style.left = '0';
	document.body.appendChild(c);
	map = new L.Map(c);
	map.setView(center, 6);
}
```

But after that change I can confirm that this fixes the bug and can be merged

johnd0e:
@Falke-Design Thank you for great suggestion!
I've added your test (with some mods).

The only flaw - test case does not work in IE.

Falke-Design (approved):
Nice change!

Falke-Design:

mourner (approved):

PR #7556, opened 2021-04-22
Quote strings in url()

jrief:
I'm having problems with a CSS postprocessor. It unwraps the CSS file and adds a hash to each of the referred images to assure that those images will be reloaded after a change. That hash value is appended using a `?somehashvalue` which causes trouble when loaded by the browser.

By quoting the 3 `url()` statements in CSS, this problem can easily be fixed.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -348,12 +348,12 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-radius: 5px;
 	}
 .leaflet-control-layers-toggle {
-	background-image: url(images/layers.png);
+	background-image: url("images/layers.png");
 	width: 36px;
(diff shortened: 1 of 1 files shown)


johnd0e:
> By quoting the 3 `url()` statements in CSS, this problem can easily be fixed.

But that may broke some others workflow, so..

For example searching here by "webpack" keyword you may found that there are too many possible ways to setup stuff, and when making any change we should be sure that it will not break more than fix.

jonkoops:
I am closing this as the syntax used here is perfectly [valid CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/url()). If this is not supported by a specific tool which needs to process these files, I would recommend to open an issue there.

Issue #7201, opened 2020-07-07
Direction RTL Tooltip


Falke-Design:
When I add the css `direction: rtl` to the html and body element, tooltips not showing correctly.

**Steps to reproduce**
- Add `direction: rtl` to html and body
- Add Marker at latlng
- Add Tooltip at latlng
- Add Popup at latlng

**Expected behavior**
Same positions

**Current behavior**
Tooltip is not on the correct position

**Environment**
- Leaflet version: 1.6.0
- Browser (with version): Firefox 78.0.1, Chrome 83.0.4103.116
- OS/Platform (with version): Windows 10

**Minimal example reproducing the issue**
https://jsfiddle.net/falkedesign/sctpj4m0/

![grafik](https://user-images.githubusercontent.com/19800037/86816421-1512ef80-c084-11ea-8398-31316ac18c82.png)

IvanSanchez:
@Falke-Design I did a bit of work on https://github.com/Leaflet/Leaflet/commit/25083057e7434136e72ecd18f05f24fbb4f7d4f2 - that covers the use case of your jsfiddle.

I don't want to merge that before checking that permanent and non-permanent tooltips *bound to markers* (with any combination of [tooltip direction](https://leafletjs.com/reference-1.6.0.html#tooltip-direction) and [CSS direction](https://developer.mozilla.org/en-US/docs/Web/CSS/direction)) work as expected as well. Would you be able to give it a shot?

Falke-Design:
@IvanSanchez I created a test file and it is not working really well 😕
Here the test file: [map-tooltip-direction-rtl.html](https://gist.github.com/Falke-Design/86c72cba5849549b92faaf772beee4d4)

Also I found out that when a tooltip is bound to a layer and permanent and sticky is true, it don't follow the mouse. Is this a Bug?
```
var poly = L.polygon([[40.02531407536202,32.804961204528816],[40.024853993628774, 32.8178358078003],[40.01966142753885, 32.81315803527833]]).addTo(map)
poly.bindTooltip('Sticky Permanent Bound --> Bug, not following mouse',{sticky: true,permanent: true,direction:'auto'}).openTooltip();
```

Falke-Design:
@IvanSanchez I think I had found a easier solution.

Set on the tooltip pane the direction `ltr` and on all childs / the tooltip self `rtl`:

```
.leaflet-tooltip-pane > * {
    direction: rtl;
}
.leaflet-tooltip-pane {
[...]

PR #7563, opened 2021-04-30
Fix Bug: permanent & sticky tooltip

Falke-Design:
Fixes bug: if tooltip is sticky and permanent it was not following the mouse https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-655746354

Demo: https://plnkr.co/edit/dPPR4QQ4rBwVwQH2

Removed from the PR #7406


johnd0e:
> Fixes bug: if tooltip is sticky and permanent it was not following the mouse [#7201 (comment)](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-655746354)

Seems provided link is not relevant here

johnd0e:
Please add plunkr to illustrate the issue.
Also, we need unit tests for this.

Falke-Design:
@johnd0e plunkr and tests added

johnd0e:
> Fixes bug: if tooltip is sticky and permanent it was not following the mouse

Just thought: the bug might be fixed easier, by documenting that such option combination is not allowed.

Is there useful case for sticky+permanent, which not looks ugly?

mourner:
> Just thought: the bug might be fixed easier, by documenting that such option combination is not allowed.

Easy enough in the current form, so I think it's not a big deal to allow it. Never know how people might make use of it.

mourner (approved):

Issue #7548, opened 2021-04-10
Relative `font-size` units for resizable text


Malvoz:
## Motivation

Text should be resizable to meet the requirements of <span title="Web Content Accessibility Guidelines">WCAG</span> 2.1 <span title="Success Criterion">SC</span> [1.4.4 Resize text](https://www.w3.org/TR/WCAG21/#resize-text).

This would require changing to _relative_ `font-size` units for _text_ content.

## Proposed solution

Change the following to use relative `font-size` units (e.g. `rem`, `em`, `%`, `smaller`, `larger`):

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L270

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L421

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L434

## Additional context

PR: https://github.com/Leaflet/Leaflet/pull/7571

Chandu-4444:
Hey, I'm a new one here... I think I can work on this issue. Would you allow me to do this?

Malvoz:
@Chandu-4444 Fine with me! Sorry for the delayed response.

Falke-Design:
Closed by #7800

PR #7571, opened 2021-05-14
Change absolute units(px) to relative units(rem)

Chandu-4444:
Hey, I've changed the units mentioned here #7548 . I'm a beginner, please let me know if anything's wrong.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -267,7 +267,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 /* general typography */
 .leaflet-container {
-	font: 12px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
+	font: 0.75rem/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
 	}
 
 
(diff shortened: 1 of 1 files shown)


ivanetra (approved):
These changes can be merged!

jonkoops:
I believe this might actually be a breaking change as the `rem` unit is relative to the font-size of the root element of the page. Since this can be different for each page Leaflet is embedded in this might cause unexpected side-effects.

jonkoops:
I don't believe compatibility with Internet Explorer needs to be a problem as we can provide a [fallback value](https://w3bits.com/rem-fallback/).

Malvoz:
Agree with @jonkoops, please consider reopening this.

dist/leaflet.css
```diff
@@ -431,7 +431,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-top: none;
 	line-height: 1.1;
 	padding: 2px 5px 1px;
-	font-size: 11px;
+	font-size: 0.69rem;
```
Malvoz, line 434: @Chandu-4444 would you like to change this to also include the `px` unit declarations as a fallback per https://github.com/Leaflet/Leaflet/pull/7571#issuecomment-961001947?

Chandu-4444: Yeah sure!

mourner:
I tried reopening but it doesn't allow me to, saying that the repo of the PR has been deleted — want to update @Chandu-4444, or submit a new PR if it's not easily possible?

Chandu-4444:
> I tried reopening but it doesn't allow me to, saying that the repo of the PR has been deleted — want to update @Chandu-4444, or submit a new PR if it's not easily possible?

I'll submit a new PR then.

Issue #3210, opened 2015-02-06
Accessibility improvements


stevevance:
I was testing [my website](http://chicagocityscape.com) with a person who is completely blind and uses the VoiceOver (VO) function on an iPad. I was observing her as she browsed the entire webpage (through a tapping function in VO that mimics tabbing between elements*) and came across the Leaflet map. I don't remember if she came across the map by tabbing through the page or by randomly tapping on it. What comes next is the unexpected behavior. 
## Problems while using VoiceOver

VO spoke the filename of the images in the map's base layer, speaking, "6078 dot p n g, link", "6079 dot p n g, link", and so forth. When the user tabbed and landed on the zoom controls VO read "plus sign, link" and "hyphen, link". I can't remember if it spoke the default `title` attributes on these buttons

The ESRI search button had no attribute to read, but VO said that it was a search input that could be opened. This is its own issue. 

When she landed on the layer control (where layers are switched on and off) the outcome was as I expected, and useful. VO spoke the name of the layer, indicated it was a checkbox or radio button and said whether it was selected. By double tapping she could control the layer's visibility. 

I feel at this point that it must be explained that this user likes maps. She can use maps that are designed for the blind. She highlighted the "Blindsquare" app as extremely usable, as it speaks the names of businesses she's walking past or that are nearby in a different direction. 

I think that Leaflet can be designed to be usable by the blind because it's still possible to describe what features are being overlaid on the map. Through continuing to tab through the page VO came across the one marker on my map that has a popup that's popped up when the map loads. The information in the popup window could be read, and any links within it to other pages could be clicked/tapped. 

What she or VO couldn't do was tab to the next few hundred markers on the map that were inside MarkerCluster (the plugin) clusters. 
## Expectations

My ideas apply only to using VO on an iPad. I think that when tabbing through elements on the screen that VO should come across the <div id='map'> tag and read its `title` attribute which should tell the user that this is a map, what the map shows, and that it's been designed for their screen-reading environment. 

The next tab should land on the zoom controls, layer controls, and any other controls, so the user can understand what functions are available to control the map. These should all have `title` attributes. 

The next tab should land on any marker that has a popup and the contents of the popup should be read. The next tabs should land and activate other markers depending on how the map designer activates them (does a click on the marker open a popup or does it open a link to another page?). 

[...]

PR #7584, opened 2021-05-26
Buttons and inputs usability and accessibility improvements

keul:
This PR try to solve some usability and accessibility issues for both buttons (like zoom controls) and inputs (like layers controls).

It contains a set of changes I'm applying again and again to every Leaflet project so I was wondering if this can be useful to be integrated in the core.

# Make zoom buttons to be more like a button

Currently there's a wrong semantic use of links instead of real buttons for zoom-in/zoom-out controls, but this change would be more complex and I'm not sure about how many issue this will generate in plugins.

Using a simple link had some accessibility drawbacks while using keyboard navigation and focus on elements as some browsers default outline is difficult to be seen:

Firefox default outline | IE 11 default outline
---------------------- | --------------------
firefox-default-outline | IE11-default-outline

Plus: often the CSS of the host page remove the default outline on links (maybe adding another kind of effect) but this can be a major issue for zoom control.

The introduced change: a visible CSS outline triggered on focus that works down to IE 11.

ie11-new-outline

Another common cosmetic CSS effect applied by modern framework on buttons is the hover animation.
Leaflet already applies a background color while hovering on zoom controls, but the background is changed immediately. In this change there's a 0.3s transition applied.

# Port same accessibility changes to input too

Modern browsers are good enough on using keyboard navigation on input elements but:

- IE 11 uses an hard to see dotted outline on focus
- Relying on default input outline with changes discussed above makes tab navigation to looks different when focusing input and zoom controls (different color, not animation).

IE 11 input outline | Chrome tabbing from zoom to inputs
------------------ | -------------------------------------
ie11-input-outine | chrome-tabbing

For this reason the tabbing effect has been extended to input keyboard navigation too.

chrome-tabbing-new

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -295,9 +295,19 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	background-repeat: no-repeat;
 	display: block;
 	}
+.leaflet-bar a {
+	transition: 0.3s ease-out;
+	}
 .leaflet-bar a:hover {
 	background-color: #f4f4f4;
 	}
+.leaflet-bar a:focus {
+	border-radius: 4px;
+	box-shadow: 0 0 0 3px #0078A8;
+	outline: none;
+	position: relative;
+	z-index: 1;
+  }
 .leaflet-bar a:first-child {
 	border-top-left-radius: 4px;
 	border-top-right-radius: 4px;
@@ -326,6 +336,9 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-bottom-left-radius: 2px;
 	border-bottom-right-radius: 2px;
 	}
+.leaflet-touch .leaflet-bar a:focus {
+	border-radius: 4px;
+  }
 
 /* zoom control */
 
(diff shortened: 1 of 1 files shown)


Malvoz:
> Currently there's a wrong semantic use of links instead of real buttons for zoom-in/zoom-out controls, but this change would be more complex and I'm not sure about how many issue this will generate in plugins.

I agree, I've noted this in https://github.com/Leaflet/LEaflet/issues/3210#issue-4.

> some browsers default outline is difficult to be seen

Yes that's unfortunate, but I'm not sure Leaflet should fix the outline-width for these UAs. They're really browser issues (2 in the case of Firefox):
- https://bugzilla.mozilla.org/show_bug.cgi?id=251198
- https://bugzilla.mozilla.org/show_bug.cgi?id=53927

If you want to enhance the visibility of outlines in these browsers in your own code you could achieve that using [`outline-width: thick`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-width) for example (perhaps conditionally using `@supports (-moz-appearance: none)`, as thick outlines look super thick in Chrome in comparison).

I personally don't agree with replacing outlines with box-shadows as this PR suggests. Some UAs ensure that outlines remain visible regardless the background-color, that doesn't happen for `box-shadow`.

Malvoz:
I appreciate the accessibility concern, though I'll close this per https://github.com/Leaflet/Leaflet/pull/7584#issuecomment-870844062. Feel free to continue discussing.

PR #7585, opened 2021-05-26
Fix default icon path detection for Rails asset pipelines

dbforge:
In Rails applications you would usually use the asset pipeline or the webpacker gem  for bundling assets together. As all assets are fingerprinted with a SHA hash to enable effective caching, which results in output files that are named differently than in the project structure. This also applies to image files. When importing/requiring leaflet over yarn (npm), the bundling process also affects the default icon files in `dist/images` as these also need to be imported due to the reference in the stylesheet. Webpacker then adds the fingerprint to the filename, resulting in e.g. `marker-icon-2273e3d8.png` instead of `marker-icon`. It also replaces the corresponding URL occurences in the stylesheet which define the default icon paths over styles. This breaks line 55 of `src/layer/marker/Icon.Default.js` because the RegExp replaces the hardcoded filename with an empty string. In case of a fingerprinted file, it consequently doesn't match, leading to a broken path which errors in the console.

The fix introduces support for Webpacker and the Rails asset pipeline (Sprockets) without breaking the detection of the initial filename (without any fingerprints).

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -52,7 +52,7 @@ export var IconDefault = Icon.extend({
 		if (path === null || path.indexOf('url') !== 0) {
 			path = '';
 		} else {
-			path = path.replace(/^url\(["']?/, '').replace(/marker-icon\.png["']?\)$/, '');
+			path = path.replace(/^url\(["']?/, '').replace(/marker-icon(-[A-Za-z0-9]+)?\.png["']?\)$/, '');
 		}
 
 		return path;


IvanSanchez:
Ok, so this is an attempt to fix a specific instance of the default icon path detection (#4968 / #7092 / #6496).

I'm... unsure about this. Generally, Leaflet tries to focus on vanilla JS *only*, so a fix for a specific build toolchain seems out of place. On the other hand, this change shouldn't break anything. So I'm conflicted on merging this.

(I wish we'd rather spend some time on #7203)

dbforge:
#7203 would also be a good alternative, I think.
In any case some change would be useful. Basically everybody who tries to use it in Rails (which is kind of comfortable due to the NPM package and yarn) must provide a custom module or else it will break the icons (unless you specify custom ones which I haven't explored any further).

I understand this fix is very specifically related to Rails' builde pipeline. Even though it doesn't break existing code, it might not come in very handy. Nevertheless, we'd somehow need a solution that doesn't break any build toolchains by default.

PR #7593, opened 2021-06-08
added polygon utility example

jblake09:
Added a visual example of the polygon utility function clipPolygon to the documentation.


Falke-Design:
You need to change the Doc here: https://github.com/Leaflet/Leaflet/blob/81dc24892a3ffd4b53158bb0b5533ae55c1931c1/src/geometry/PolyUtil.js#L7-L14

Falke-Design:
@mourner do we want this in the docu? If yes I will update the docs

PR #7594, opened 2021-06-10
Leaflet 1.7.1 with typescript and snyk

luismendes070:


IvanSanchez:
This PR seems to have no rationale to it. From my point of view, this just adds more tooling for no benefit.

PR #7597, opened 2021-06-13
BoxZoom: Reset _moved flag after cancel with ESC

Falke-Design:
After cancel the BoxZoom with ESC the `_moved` flag is not reseted and destorys the `click` event on the map. 
The part 
```
this._clearDeferredResetState();
this._resetStateTimeout = setTimeout(Util.bind(this._resetState, this), 0);
``` 
is missing in the [_onKeyDown](https://github.com/Leaflet/Leaflet/blob/81dc24892a3ffd4b53158bb0b5533ae55c1931c1/src/map/handler/Map.BoxZoom.js#L140) function. [_onMouseUp](https://github.com/Leaflet/Leaflet/blob/master/src/map/handler/Map.BoxZoom.js#L128) Reference

Usage of the `_moved` flag over `boxZoom.moved()` which breaks some functions outside of `boxZoom`:
https://github.com/Leaflet/Leaflet/blob/81dc24892a3ffd4b53158bb0b5533ae55c1931c1/src/map/Map.js#L1430-L1433

Can simply tested with `map.on('click', (e)=>console.log(e));`, it will not be triggered after cancel BoxZoom:

boxzoom_esc_break
After reuse and finish the boxZoom the `click` event is fired again.

### DIFF
--- a/spec/suites/map/handler/Map.BoxZoom.js
+++ b/spec/suites/map/handler/Map.BoxZoom.js
@@ -1,0 +1,65 @@
+describe("Map.BoxZoom", function () {
+	var container, map;
+
+	beforeEach(function () {
+		container = document.createElement('div');
+		container.style.width = container.style.height = '600px';
+		container.style.top = container.style.left = 0;
+		container.style.top = container.style.top = 0;
+		container.style.position = 'absolute';
+		document.body.appendChild(container);
+	});
+
+	afterEach(function () {
+		map.remove();
+		document.body.removeChild(container);
+	});
+
+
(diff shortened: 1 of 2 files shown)


mourner (changes requested):
Really nice fix, thank you! Can we also add a unit test for this to make sure this doesn't accidentally regress in future?

Falke-Design:
@mourner I added a test. I hope it is good enough, it was not better possible for me. I didn't used prosthetic-hand because it was not possible to use it together with the shiftKey flag.

mourner (approved):

PR #7600, opened 2021-06-18
cleanup marker.svg

rala72:
Cleaning up the `marker.svg`.


rala72:
Not directly part of PR but connected: Is there a vector image of the shadow?

mourner:
> Is there a vector image of the shadow?

Not that I know of — might have been lost to time, but I'll look for it.

Issue #7564, opened 2021-05-02
Documentation error in Map Panes section


mz8i:
In the documentation section about Map Panes: 
https://leafletjs.com/reference-1.7.1.html#map-overlaypane

There seems to be an error in the descriptions of `shadowPane` and `overlayPane` - something seem to be mixed up. Comparing to a much older version of the documentation (https://leafletjs.com/reference-1.0.3.html#map-pane) it seems that perhaps there was a modification of the order of the panes, but the documentation was not updated correctly.

My guess is that the fix is to switch the values in the Pane column of the docs table - swapping `overlayPane` with `shadowPane` between rows

answerquest:
I noticed the same! Anybody has an idea of where the source is of this documentation?
I see this: https://github.com/Leaflet/Leaflet/blob/master/docs/reference-1.7.1.html#L1711 
but is it a built output from somewhere else, or can a PR with the change done do the needful?

Falke-Design:
@johnd0e this is already fixed and can be closed: #7349

PR #7601, opened 2021-06-19
switch overlayPane and shadowPane in doc

answerquest:
switching mistakenly places overlayPane and shadowPane in documentation. #7564


Falke-Design:
@answerquest the correct place to change the Doc would be here: 
https://github.com/Leaflet/Leaflet/blob/81dc24892a3ffd4b53158bb0b5533ae55c1931c1/src/map/Map.js#L1143-L1148

But this is already fixed: #7349

PR #7603, opened 2021-06-19
Update `getCenter()` calculation and move it to PolyUtil / LineUtil

### LINKED ISSUE #5216: Tooltip binded to polygon displacement on zoom in
## How to reproduce

Leaflet version I'm using: 1.0.2
Browser (with version) I'm using: Chrome & FF
OS/Platform (with version) I'm using: Windows 10

Binded tooltip to polygon changes its place on zoom in. I'm attaching jsfiddle examples. 

In first example problem occurs once map is zoomed to polygon.

In second example problem does not occur if we initialize map with higher zoom.

https://jsfiddle.net/cco2v6vq/

https://jsfiddle.net/cco2v6vq/1/

Falke-Design:
Changed the `getCenter()` function for Polygon & Polyline. Now it uses `map.un-/project()` of the latlngs instead of the pre-generated `_rings` with `latLngToLayerPoint()`.

Also added the center functions to `PolyUtil` and `LineUtil`, because then the function can be used without creating a Polyline/-gon and adding it to the map:
grafik

## Visible changes:

**Polyline**:
 - if `getCenter()` is called on a low zoom level (0), then the returned point is not anymore the first latlng:
grafik
- `getCenter()` returns the same point no matter which zoom level (2 & 16)
grafik

**Polygon**:
 - if `getCenter()` is called on a low zoom level (0), then the returned point is not anymore the first latlng:
grafik
- `getCenter()` returns the same point no matter which zoom level (8 & 16)
grafik

Fix #5216


johnd0e:
Your solution is correct.
The only unclear thing is performance difference.

Falke-Design:
I have done some tests, it looks like the new calculation is faster:
Old one:
```
Count: 100.000
performance: 4440ms //Rectangle
performance: 8931ms //Polygon
performance: 12969ms //Polyline

performance: 4443ms
performance: 8935ms
performance: 13094ms

performance: 4136ms
performance: 8532ms
performance: 12389ms

Count: 1.000.000
performance: 47880ms
performance: 97666ms
performance: 145388ms

performance: 47784ms
performance: 98649ms
performance: 144880ms
```
New:
```
Count: 100.000
performance: 4634ms
performance: 8484ms
performance: 12012ms

performance: 3850ms
performance: 7774ms
performance: 11199ms

performance: 4365ms
performance: 8241ms
performance: 11629ms

Count: 1.000.000
performance: 39544ms
performance: 84632ms
performance: 126625ms

performance: 39555ms
performance: 84180ms
performance: 125914ms
```

Code:
```
window.onload=()=> {
        setTimeout(() => {

          var COUNT = 1000*1000;

          console.log('start')
          console.time('performance');
          for (var i = 0; i < COUNT; i++) {
            var rect = L.rectangle([[48.78232, 9.17702], [48.8, 9.2]]).addTo(map);
            rect.getCenter();
            rect.removeFrom(map);
          }
          console.timeLog('performance');
          for (var i = 0; i < COUNT; i++) {
            var rect = L.polygon([[48.78232, 9.17702], [48.8, 9.2], [48.7, 9.19], [48.7, 9.13]]).addTo(map);
            rect.getCenter();
            rect.removeFrom(map);
          }
          console.timeLog('performance');
          for (var i = 0; i < COUNT; i++) {
            var rect = L.polyline([[48.78232, 9.17702], [48.8, 9.2], [48.7, 9.19], [48.7, 9.13]]).addTo(map);
            rect.getCenter();
            rect.removeFrom(map);
          }
          console.timeEnd('performance')

        }, 2000)
      }
```

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,59 @@ export function _flat(latlngs) {
...
+	if (!isFlat(latlngs)) {
+		throw new Error('latlngs are not flat!');
+	}
+
+	// throws error when map is not passed because calculation requires projected coordinates
+	if (!map) {
+		throw new Error('map not passed');
+	}
+
+	var zoom = map.getMaxZoom() === Infinity ? map.getZoom() : map.getMaxZoom();
```
johnd0e, line 265: 1. Excessive call of `map.getMaxZoom()`
2. Do you have tests with `Infinity` zoom?
3. Do you have tests with `undefined` zoom?

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,59 @@ export function _flat(latlngs) {
 	console.warn('Deprecated use of _flat, please use L.LineUtil.isFlat instead.');
 	return isFlat(latlngs);
 }
+
+/* @function polylineCenter(latlngs: LatLng[], map: Map): LatLng
+ * Returns the center of the `Polyline` of the passed LatLngs.
+ * The returned LatLng will be on the `Polyline`.
+ */
+export function polylineCenter(latlngs, map) {
```
johnd0e, line 249: Not sure that we are ready to expose that functions, without any discussion.
- As it is (relatively) low-level function, what if we pass `crs` and `zoom` directly, instead of map?
- What if we give reasonable defaults for `crs` and `zoom`?

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,59 @@ export function _flat(latlngs) {
...
+	if (!map) {
+		throw new Error('map not passed');
+	}
+
+	var zoom = map.getMaxZoom() === Infinity ? map.getZoom() : map.getMaxZoom();
+	var points = [];
+	for (var j in latlngs) {
+		points.push(map.project(toLatLng(latlngs[j]), zoom));
+	}
+	var len = points.length;
```
johnd0e, line 270: Isn't that the same value as `latlngs.length`?

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,59 @@ export function _flat(latlngs) {
...
+	}
+
+	if (!isFlat(latlngs)) {
+		throw new Error('latlngs are not flat!');
+	}
+
+	// throws error when map is not passed because calculation requires projected coordinates
+	if (!map) {
+		throw new Error('map not passed');
+	}
```
johnd0e, line 263: Why do we need explicit throw here?
It will throw anyway on next line.

src/layer/vector/Polygon.js
```diff
@@ -61,39 +61,14 @@ export var Polygon = Polyline.extend({
 		return !this._latlngs.length || !this._latlngs[0].length;
 	},
 
+	// @method getCenter(): LatLng
+	// Returns the center ([centroid](http://en.wikipedia.org/wiki/Centroid)) of the Polygon.
 	getCenter: function () {
 		// throws error when not yet added to map as this center calculation requires projected coordinates
 		if (!this._map) {
 			throw new Error('Must add layer to map before using getCenter()');
 		}
```
johnd0e, line 70: Again, it will throw naturally anyway.
You can put that on the test, by the way.

johnd0e (commented):
It seems impossible, but I can confirm that new code is faster.
I have no idea why that could be, considering that new code uses more `project` calls.

src/geometry/LineUtil.js
Falke-Design: I copied this from the original code. But I think it is done because `latlngs.length` returns not the correct length if it is a multipolyline.

src/geometry/LineUtil.js
```diff
@@ -240,3 +241,59 @@ export function _flat(latlngs) {
...
+ * The returned LatLng will be on the `Polyline`.
+ */
+export function polylineCenter(latlngs, map) {
+	var i, halfDist, segDist, dist, p1, p2, ratio, center;
+
+	if (!latlngs || latlngs.length === 0) {
+		throw new Error('latlngs not passed');
+	}
+
+	if (!isFlat(latlngs)) {
```
Falke-Design, line 256: I need to check this, in the original function was allowed to pass a multiployline

src/geometry/LineUtil.js
Falke-Design: I think this is a good idea

src/geometry/PolyUtil.js
```diff
@@ -53,3 +53,54 @@ export function clipPolygon(points, bounds, round) {
...
+ * Returns the center ([centroid](http://en.wikipedia.org/wiki/Centroid)) of the `Polygon` of the passed LatLngs.
+ */
+export function polygonCenter(latlngs, map) {
+	var i, j, p1, p2, f, area, x, y, center;
+
+	if (!latlngs || latlngs.length === 0) {
+		throw new Error('latlngs not passed');
+	}
+
+	if (!LineUtil.isFlat(latlngs)) {
```
Falke-Design, line 67: Should be possible to use for a multipolygon too

src/geometry/LineUtil.js
Falke-Design: Copied from original `getCenter()`. It is clearer but it is not necessary

src/layer/vector/Polygon.js
Falke-Design: Clearer message, if it throw only `this._map is undefined` it is not clear why this error happens

johnd0e: > it is not clear why this error happens

But it gets crystal clear when you see traceback.

There are A LOT of such places in leaflet, and if we age going to explicitly check every one - better switch to typescript instead.

src/geometry/LineUtil.js
johnd0e: > I copied this from the original code.

But it is redundant for the new code, right?

src/layer/vector/Polygon.js
Falke-Design: I don't totally agree with you. This check here makes sense because it would not be clear why `getCenter()` is not working. It is a common problem that users want to use `getCenter()` before adding to the map.

src/geometry/LineUtil.js
Falke-Design: Yep if 
> But I think it was / is used because `latlngs.length` returns not the correct length if it is a multipolyline.

makes no problems

src/layer/vector/Polygon.js
johnd0e: > It is a common problem that users want to use `getCenter()` before adding to the map.

So? It fails, you see (short) call stack, you see args, and all gets clear. No?

src/geometry/LineUtil.js
johnd0e: But here latlngs mapped to points directly, so I do not see any difference.

Falke-Design: I need to check this mapping because this is not working for mult-lines. Currently `latlngs.length` work because I check with `if (!isFlat(latlngs)) {` if it is a multi-line. If we say that multipolylines should supported too, it will not work.

Can it be that the old `getCenter()` is also only working for **no**t mulit-lines? Because it uses only the first array `points = this._rings[0],`

src/geometry/LineUtil.js
johnd0e: Instead we could document that only first ring is processed for multi-features.

src/geometry/LineUtil.js
Falke-Design: 1. replaced - reading out once and then check in the if
2. The most tests are with maxZoom Infinity, or do you mean the actual zoom `getZoom()`? When does this happen?
3. When does this happen?

Falke-Design:
Updated
- Change arguments to `crs` and `zoom` instead of `map`.
- Allow non flat latlngs but only takeing the first ring

src/layer/vector/Polygon.js
Falke-Design: @mourner what do you think? Should we remove this error message from Polygon & Polyline in the `getCenter()` method?

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,62 @@ describe('LineUtil', function () {
 			]);
 		});
 	});
+
+	describe('#polylineCenter', function () {
+		var map;
+		before(function () {
+			map = new L.Map(document.createElement('div'), {center: [55.8, 37.6], zoom: 6});
```
johnd0e, line 90:
```suggestion
			map = L.map(document.createElement('div'), {center: [55.8, 37.6], zoom: 6});
```
(Same for polygon)

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,62 @@ describe('LineUtil', function () {
...
+
+		after(function () {
+			map.remove();
+		});
+
+		// More tests in PolylineSpec
+
+		it('compute center of line', function () {
+			var latlngs = [[80, 0], [80, 90]];
+			expect(L.LineUtil.polylineCenter(latlngs, map.options.crs, map.getZoom())).to.be.nearLatLng(L.latLng([80, 45]), 1e-2);
```
johnd0e, line 101: Too long line (here and below)

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,62 @@ describe('LineUtil', function () {
...
+		});
+
+		it('compute center of line with maxZoom', function () {
+			L.gridLayer({maxZoom: 18}).addTo(map);
+			var latlngs = [[80, 0], [80, 90]];
+			expect(L.LineUtil.polylineCenter(latlngs, map.options.crs, map.getMaxZoom())).to.be.nearLatLng(L.latLng([80, 45]), 1e-2);
```
johnd0e, line 107: What if define `zoom` and `crs` variables in `before` hook?

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -83,4 +83,62 @@ describe('LineUtil', function () {
...
+			var latlngs = [
+				[[80, 0], [80, 90]]
+			];
+			var spy = sinon.spy(console, 'warn');
+
+			var center = L.LineUtil.polylineCenter(latlngs, map.options.crs, zoom);
+			expect(spy.calledOnce).to.eql(true);
+			expect(center).to.be.nearLatLng(L.latLng([80, 45]), 1e-2);
+
+			console.warn.restore();
```
johnd0e, line 141: Should be before `expect`

spec/suites/geometry/PolyUtilSpec.js
```diff
@@ -38,4 +38,66 @@ describe('PolyUtil', function () {
...
+			var latlngs = [
+				[[0, 0], [10, 0], [10, 10], [0, 10]]
+			];
+			var spy = sinon.spy(console, 'warn');
+
+			var center = L.PolyUtil.polygonCenter(latlngs, map.options.crs, zoom);
+			expect(spy.calledOnce).to.eql(true);
+			expect(center).to.be.nearLatLng(L.latLng([5, 5]), 1e-1);
+
+			console.warn.restore();
```
johnd0e, line 93: Should be before `expect`

spec/suites/layer/vector/PolygonSpec.js
```diff
@@ -156,13 +156,22 @@ describe('Polygon', function () {
...
+			var layer = new L.Polygon(latlngs).addTo(map);
+			expect(layer.getCenter()).to.be.nearLatLng(L.latLng([5, 5]), 1e-1);
+		});
+
 		it('should compute center of a small simple polygon', function () {
 			var latlngs = [
 				[[0, 0], [0.010, 0], [0.010, 0.010], [0, 0.010]]
 			];
 			var layer = new L.Polygon(latlngs).addTo(map);
 			map.setZoom(0);  // Make the polygon disappear in screen.
```
johnd0e, line 173: You should then restore the zoom, here or in `afterEach` hook.

spec/suites/layer/vector/PolygonSpec.js
```diff
@@ -175,6 +184,25 @@ describe('Polygon', function () {
...
 		});
 
+		it('should compute same center for low and high zoom', function () {
+			var latlngs = [
+				[[0, 0], [0.010, 0], [0.010, 0.010], [0, 0.010]]
+			];
+			var layer = new L.Polygon(latlngs).addTo(map);
+			map.setZoom(0);
+			var center = layer.getCenter();
+			map.setZoom(18);
```
johnd0e, line 194: Same here, and in all other places.

spec/suites/layer/vector/PolygonSpec.js
```diff
@@ -175,6 +184,25 @@ describe('Polygon', function () {
...
+			expect(layer.getCenter()).to.be.nearLatLng(center);
+		});
+
+		it("should compute center for multi-polygon including hole", function () {
+			var latlngs = [
+				[[[10, 20], [30, 40], [50, 60]]],
+				[[[0, 10], [10, 10], [10, 0]], [[2, 3], [2, 4], [3, 4]]]
+			];
+			var layer = new L.Polygon(latlngs).addTo(map);
+			expect(layer.getCenter()).to.be.nearLatLng(L.latLng([31.43, 39.99]), 1e-1);
```
johnd0e, line 204: Why are you using different precision values?
What if you stick to default?

johnd0e (changes requested):

johnd0e:
Just for info: there are other ways to find center of polygon:
https://blog.mapbox.com/a-new-algorithm-for-finding-a-visual-center-of-a-polygon-7c77e6492fbc

spec/suites/layer/vector/PolygonSpec.js
Falke-Design: The latlng values are long numbers but I replaced it now with the exact number, it is cleaner

spec/suites/layer/vector/PolygonSpec.js
Falke-Design: I changed it to `beaforeEach` so for each test the map is new created and the zoom & center will be set

spec/suites/layer/vector/PolygonSpec.js
Falke-Design: I changed it to `beaforeEach` so for each test the map is new created and the zoom & center will be set

Falke-Design:
> Just for info: there are other ways to find center of polygon:
https://blog.mapbox.com/a-new-algorithm-for-finding-a-visual-center-of-a-polygon-7c77e6492fbc

Very interessting. It would be nice to have this but if we use the code from Mapbox we need to add the copyright. I would create it from the explanation but I'm not good in algorithm / mathematical things.
I suggest we merge this PR and then we can look if we want to use the "visual center algorithm".

spec/suites/geometry/PolyUtilSpec.js
```diff
@@ -40,57 +40,53 @@ describe('PolyUtil', function () {
...
-			var zoom = maxZoom === Infinity ? map.getZoom() : maxZoom;
 			var latlngs = [
 				[[0, 0], [10, 0], [10, 10], [0, 10]]
 			];
 			var spy = sinon.spy(console, 'warn');
-
-			var center = L.PolyUtil.polygonCenter(latlngs, map.options.crs, zoom);
+			var center = L.PolyUtil.polygonCenter(latlngs, crs, zoom);
+			console.warn.restore();
 			expect(spy.calledOnce).to.eql(true);
```
johnd0e, line 88:
```suggestion
			expect(spy.calledOnce).to.be.ok();
```

spec/suites/geometry/LineUtilSpec.js
```diff
@@ -127,18 +131,14 @@ describe('LineUtil', function () {
...
-			var zoom = maxZoom === Infinity ? map.getZoom() : maxZoom;
 			var latlngs = [
 				[[80, 0], [80, 90]]
 			];
 			var spy = sinon.spy(console, 'warn');
-
-			var center = L.LineUtil.polylineCenter(latlngs, map.options.crs, zoom);
+			var center = L.LineUtil.polylineCenter(latlngs, crs, zoom);
+			console.warn.restore();
 			expect(spy.calledOnce).to.eql(true);
```
johnd0e, line 140:
```suggestion
			expect(spy.calledOnce).to.be.ok();
```

src/layer/vector/Polyline.js
```diff
@@ -114,44 +114,15 @@ export var Polyline = Path.extend({
 	},
 
 	// @method getCenter(): LatLng
-	// Returns the center ([centroid](http://en.wikipedia.org/wiki/Centroid)) of the polyline.
+	// Returns the center of the Polyline.
```
johnd0e, line 117: Why remove that?

johnd0e (dismissed):

src/layer/vector/Polyline.js
Falke-Design: Reverted

johnd0e:
In my opinion it is ready to merge, except 1 unresolved question: do we need explicit throws (see above).
@mourner @IvanSanchez @jonkoops

johnd0e:
@Falke-Design 
What about numerous `.nearLatLng(L.latLng...` occurrences?

jonkoops (approved):

jonkoops:
This change is looking good to me. @johnd0e do you feel comfortable merging in this change?

Falke-Design:
@jonkoops I don't think that @johnd0e will response, he was offline since january. I would like to merge it now, because we have now 2 approvals with only one question "should we keep the already existing exceptions?". I think they don't hurt and else we can still remove them.

jonkoops:
Yes I agree. Let's go ahead and merge this in 👍

PR #7605, opened 2021-06-23
Remove plugin leaflet-echarts

AlexanderMisel:
`leaflet-echarts` isn't an open source project and is impossible to act as a plugin, as the repo only provides minified source, which packed all echarts, leaflet and some others stuffs we don't know.


johnd0e:
Isn't this plugin?
https://github.com/wandergis/leaflet-echarts/blob/master/src/leaflet-echarts.js

PR #7609, opened 2021-06-30
Correct condition to add zoom limits for Layer

vcoppe:
Corrected a small mistake which attempts to add zoom limits when maxZoom is NaN while it should be the opposite (when maxZoom is not NaN, as done for minZoom).

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -235,7 +235,7 @@ Map.include({
 	},
 
 	_addZoomLimit: function (layer) {
-		if (isNaN(layer.options.maxZoom) || !isNaN(layer.options.minZoom)) {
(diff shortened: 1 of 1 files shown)


johnd0e:
The regression was introduced here: https://github.com/Leaflet/Leaflet/commit/36b8c7cf4dff080d989d86983082ac799958bdc0
The fix is proper.

> local tests passed

The problems are:
- the first part of expression is not covered by tests at all.
- not clear why we need `isNaN` check - that is not covered by tests too.

johnd0e (approved):

mourner:
> The regression was introduced here: [36b8c7c](https://github.com/Leaflet/Leaflet/commit/36b8c7cf4dff080d989d86983082ac799958bdc0)

Wow, 8 years ago. And no one noticed until now.

Issue #7577, opened 2021-05-21
Support dark mode


vanillajonathan:
Add a [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) CSS media query for dark mode.

Something like:

```css
.leaflet-bar a,
.leaflet-bar a:hover {
	width: 26px;
	height: 26px;
	line-height: 26px;
	display: block;
	text-align: center;
	text-decoration: none;
}

@media (prefers-color-scheme: light) {
.leaflet-bar a,
.leaflet-bar a:hover {
	background-color: #fff;
	border-bottom: 1px solid #ccc;
	color: black;
        }
}

@media (prefers-color-scheme: dark) {
.leaflet-bar a,
.leaflet-bar a:hover {
	background-color: #000;
	border-bottom: 1px solid #333;
	color: white;
	}
}
```

Falke-Design:
@mourner @johnd0e @IvanSanchez is this a wanted feature?

Malvoz:
A note on backwards compatibility:

The "light" styles (the current styles applied by default) should remain outside a `prefers-color-scheme` media query, otherwise nothing will apply in browsers that do not [support](https://caniuse.com/?search=prefers-color-scheme) that media feature. I.e. only a `prefers-color-scheme: dark` media query is necessary (to override the default light colors).

vanillajonathan:
Something like this:

```css
.leaflet-bar a,
.leaflet-bar a:hover {
	background-color: #fff;
	border-bottom: 1px solid #ccc;
	color: black;
	width: 26px;
	height: 26px;
	line-height: 26px;
	display: block;
	text-align: center;
	text-decoration: none;
}

@media (prefers-color-scheme: dark) {
.leaflet-bar a,
.leaflet-bar a:hover {
	background-color: #000;
	border-bottom: 1px solid #333;
	color: white;
	}
}
```

Malvoz:
Would it be appropriate to add a basic tutorial for dark mode? (whether part of PR https://github.com/Leaflet/Leaflet/pull/7612 or separately)

In that case, it'd be pretty neat to add a section on how developers also can use custom CSS to modify tiles to fit the dark color scheme, see: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/551#issuecomment-955412511.

Falke-Design:
I think there is not much to explain. What do you think should be in the tutorial?

[...]

Issue #7539, opened 2021-04-01
Popup's close button does not have sufficient contrast


Malvoz:
## Motivation

The popup's close button has a contrast ratio of 1.76, and **fails** all of the following <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> Success Criteria:
- [1.4.3 Contrast (Minimum)](https://www.w3.org/TR/WCAG21/#contrast-minimum)
- [1.4.6 Contrast (Enhanced)](https://www.w3.org/TR/WCAG21/#contrast-enhanced)
- [1.4.11 Non-text Contrast](https://www.w3.org/TR/WCAG21/#non-text-contrast)

<img width="300" src="https://user-images.githubusercontent.com/26493779/113319315-9779c880-9311-11eb-90da-2e90f66c785b.png" alt="Popup contrast issue">

## Proposed solution

Increase the contrast of the popup's close button by defaulting to a darker color:

https://github.com/Leaflet/Leaflet/blob/b52961aa469dc17b3710ef7a9c0f13b6f4dba550/dist/leaflet.css#L521

## Additional context

Some browser's dev tools make suggestions for colors to increase the contrast to acceptable levels (Chrome in this case):

<img width="325" src="https://user-images.githubusercontent.com/26493779/113319359-a19bc700-9311-11eb-9c51-3d61dfe758c2.png" alt>

Malvoz:
The color of the close button is a design choice, if it's decided not to change it for that reason then it could be added as a known limitation in the [accessibility statement](https://github.com/Leaflet/Leaflet/issues/7537).

Issue #7538, opened 2021-04-01
Increase contrast of the attribution container to ensure readability


Malvoz:
Attribution can be hard to read depending on the underlying graphic, this is especially true for people with visual impairments.

Here's an example:

![attrib-contrast](https://user-images.githubusercontent.com/26493779/113310847-f38c1f00-9308-11eb-9568-41cd4d4e971c.png)

The applied Alpha transparency could be reduced to meet <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> <span title="Success Criteria">SC</span> [1.4.3 Contrast (Minimum)](https://www.w3.org/TR/WCAG21/#contrast-minimum), or [1.4.6 Contrast (Enhanced)](https://www.w3.org/TR/WCAG21/#contrast-enhanced):

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L405

Malvoz:
The applied transparency is a design choice, if it's decided not to change it for that reason then it could be added as a known limitation in the [accessibility statement](https://github.com/Leaflet/Leaflet/issues/7537).

ppKrauss:
Ideal is to apply contrast only into a layer, not all map. Using the example of https://leafletjs.com/ home, at its `<div id="map">`.

## Wrong full mode
Is **easy to add a CSS filter** to all map, for example
```html
<div id="map" class="map map-home" style="filter: brightness(200%); height: 300px; margin-top: 50px;"></div>
```

## Correct, by layer
Layer by layer, for example
```js
L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
    attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
}).addTo(map).filter_brightness('150%');  // low contrast on background

L.marker([51.5, -0.09]).addTo(map);  // no filter at marker, best contrast!
```

ppKrauss:
**Solution using CSS class with a filter**. Example:   apply CSS a "contrast filter" in the  https://leafletjs.com/ home, <br/> at its `<div id="map">`.  Suppose:
```html
<style>
 .brightness{
    filter: brightness(130%);
   -webkit-filter: brightness(130%); /* Safari need it in 2020? */
}
</style>
```

## Apply filter in the full map
Add more one class:
```html
[...]

PR #7612, opened 2021-07-01
Feature Dark Mode

Falke-Design:
# Adds Dark Mode to Leaflet

If the Browser settings of the user is in Dark Mode, Leaflet will automatically change it too. The information comes from [prefers-color-scheme](prefers-color-scheme). This automation can also disabled while map creation with `useUserColorScheme: false` -> `L.map("map",{useUserColorScheme: false});`

Dark Mode can also be forced with the map option `darkMode: true`. 

@Malvoz maybe you can take a look into it about accessibility / contrast

Fix: #7577 

grafik
grafik

### DIFF
--- a/debug/map/darkmode.html
+++ b/debug/map/darkmode.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script src="../dist/leaflet-src.js"></script>
+
+	<style>
+		html, body{
+			background: #333333;
+		}
+
+		#map{
+			border: none;
+		}
+	</style>
+</head>
+<body>
+
+	<div id="map"></div>
+
+	<script>
+
+		var osmUrl = 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+
+		var map = L.map('map')
+				.setView([50.5, 30.51], 15)
+				.addLayer(osm);
+
+		var marker = L.marker(map.getCenter()).addTo(map);
+
+		marker.bindPopup("<p>Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Donec odio. Quisque volutpat mattis eros. Nullam malesuada erat ut turpis. Suspendisse urna nibh, viverra non, semper suscipit, posuere a, pede.</p><p>Donec nec justo eget felis facilisis fermentum. Aliquam porttitor mauris sit amet orci. Aenean dignissim pellentesque.</p>").addTo(map).openPopup();
+		marker.bindTooltip("This is a Tooltip",{permanent: true}).addTo(map).openTooltip();
+
+		L.control.scale().addTo(map);
+
+
+		var grayscale = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+			streets   = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+
+		var baseMaps = {
+			"Grayscale": grayscale,
+			"Streets": streets
+		};
+
+		var overlayMaps = {
+			"Cities": null
+		};
+
+		L.control.layers(baseMaps, baseMaps, {collapsed: true}).addTo(map);
+		L.control.layers(baseMaps, baseMaps, {collapsed: false}).addTo(map);
+
+		var myIcon = L.divIcon({});
+		L.marker(map.getCenter(), {icon: myIcon}).addTo(map);
+
+	</script>
+</body>
+</html>
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -328,22 +328,21 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 
 
-@media (prefers-color-scheme: dark) {
-	.leaflet-bar a,
-	.leaflet-bar a:hover{
-		background-color: #333;
-		border-bottom: 1px solid #ddd;
-		color: #fff;
-	}
+/* Dark Mode */
+.leaflet-darkmode .leaflet-bar a,
+.leaflet-darkmode .leaflet-bar a:hover{
+  background-color: #333;
+  border-bottom: 1px solid #ddd;
+  color: #fff;
+}
 
-	.leaflet-bar a:hover {
-		background-color: #444;
-	}
+.leaflet-darkmode .leaflet-bar a:hover {
+  background-color: #444;
+}
 
-	.leaflet-bar a.leaflet-disabled {
-		background-color: #444;
-		color: #bbb;
-	}
+.leaflet-darkmode .leaflet-bar a.leaflet-disabled {
+  background-color: #444;
+  color: #bbb;
 }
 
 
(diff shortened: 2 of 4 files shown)


debug/map/darkmode.html
```diff
@@ -0,0 +1,66 @@
...
+
+	<link rel="stylesheet" href="../css/screen.css" />
+
+	<script src="../dist/leaflet-src.js"></script>
+
+	<style>
+		html, body{
+			background: #333333;
+		}
+
```
Malvoz, line 18: Should this example have a fixed dark background?

I think authors are more likely to use either of the following methods:

1. Change background based on preference:

    ```css
    @media (prefers-color-scheme: dark) {
      :root {
        background-color: #333333;
      }
    }
    ```

2. Indicate to UAs it is compatible with dark mode and allow the UA to set the background:

    - using HTML:

      ```html
      <meta name="color-scheme" content="dark light">
      ```
    -  alternatively in CSS:

       ```css
       :root {
         color-scheme: dark light;
       }
       ```

References:

- https://html.spec.whatwg.org/multipage/semantics.html#meta-color-scheme
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name#:~:text=color-scheme
- https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
- https://web.dev/color-scheme/
- https://web.dev/prefers-color-scheme/

Falke-Design: Good catch! I will change

Falke-Design:
@Malvoz changed. What do you think about the contrast and co?

dist/leaflet.css
```diff
@@ -327,6 +327,25 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-bottom-right-radius: 2px;
 	}
 
+
+/* Dark Mode */
```
Malvoz, line 331: Having the dark mode styles in the main leaflet.css means users will always have to download these extra KB even if the `darkMode` option is set to false. That's something that should be considered. Maybe there should be a separate stylesheet for dark mode styles? I don't have a particularly strong opinion about this, perhaps maintainers do.

Malvoz:
> What do you think about the contrast and co?

There are contrast issues in light mode too (https://github.com/Leaflet/Leaflet/issues/7539, https://github.com/Leaflet/Leaflet/issues/7538), perhaps that should be dealt with separately from this PR?

Malvoz (commented):
Relating to:

https://github.com/Leaflet/Leaflet/blob/f2f7868d07dd734b8b50cc249090fc956c755005/dist/leaflet.css#L252-L253

I think a dark background-color for the leaflet container is desired in dark mode.

debug/map/darkmode.html
Malvoz: Just noting that the _commit message_ of https://github.com/Leaflet/Leaflet/pull/7612/commits/f2f7868d07dd734b8b50cc249090fc956c755005 indicates option 2 was used, though option 1 was used. 😋

Falke-Design: Option 2, option 1? I don't get it 😟

Malvoz: @Falke-Design I was referring to https://github.com/Leaflet/Leaflet/pull/7612#discussion_r739652955.

> 1. Change background based on preference

> 2. Indicate to UAs it is compatible with dark mode and allow the UA to set the background

The commit description is "Indicate to UAs that the example htmls are compatible with dark mode", but that's not what https://github.com/Leaflet/Leaflet/commit/f2f7868d07dd734b8b50cc249090fc956c755005 does, it changes the background based on preferences. It's not a big deal, just a mismatch in the commit description.

Falke-Design: But I added in the `screen.css` following:
```
:root {
	color-scheme: light dark;
	}
```

Malvoz: Ah, I somehow missed that. Sorry for all the fuss.

dist/leaflet.css
```diff
@@ -330,6 +330,25 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
```
Malvoz, line 334: @Falke-Design Should we move all dark mode styles to the bottom of this file (but before `@media print` styles) for clarity?

dist/leaflet.css
```diff
@@ -330,6 +330,25 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-bottom-right-radius: 2px;
 	}
 
+
+/* Dark Mode */
+.leaflet-darkmode .leaflet-bar a,
+.leaflet-darkmode .leaflet-bar a:hover{
```
Malvoz, line 336: These styles apply to both the default state and on `:hover`, should probably remove `.leaflet-darkmode .leaflet-bar a:hover` here?

dist/leaflet.css
```diff
@@ -330,6 +330,25 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
+/* Dark Mode */
+.leaflet-darkmode .leaflet-bar a,
+.leaflet-darkmode .leaflet-bar a:hover{
+	background-color: #333;
+	border-bottom: 1px solid #ddd;
+	color: #fff;
+	}
+
+.leaflet-darkmode .leaflet-bar a:hover {
+	background-color: #444;
```
Malvoz, line 343: Probably add a selector for `:focus` per https://github.com/Leaflet/Leaflet/pull/7274.

dist/leaflet.css
```diff
@@ -557,6 +612,30 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
+.leaflet-darkmode .leaflet-popup-tip {
+	background: #333;
+	color: #fff;
+	box-shadow: 0 3px 14px rgba(0,0,0,0.4);
+	}
+.leaflet-darkmode.leaflet-container a.leaflet-popup-close-button{
+	color: #ddd;
+	}
+.leaflet-darkmode.leaflet-container a.leaflet-popup-close-button:hover {
+	color: #999;
```
Malvoz, line 626: Probably add a selector for `:focus` per https://github.com/Leaflet/Leaflet/pull/7274.

dist/leaflet.css
```diff
@@ -557,6 +612,30 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
 
+/* Dark Mode */
+.leaflet-darkmode .leaflet-popup-content-wrapper,
+.leaflet-darkmode .leaflet-popup-tip {
+	background: #333;
+	color: #fff;
+	box-shadow: 0 3px 14px rgba(0,0,0,0.4);
+	}
+.leaflet-darkmode.leaflet-container a.leaflet-popup-close-button{
+	color: #ddd;
```
Malvoz, line 623: Unless needed for specificity, I suggest leaving out `.leaflet-container` in `.leaflet-darkmode.leaflet-container [...]`.

dist/leaflet.css
Malvoz, line 626: Unless needed for specificity, I suggest leaving out `.leaflet-container` in `.leaflet-darkmode.leaflet-container [...]`.

dist/leaflet.css
```diff
@@ -557,6 +612,30 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
+.leaflet-darkmode .leaflet-popup-scrolled {
+	border-bottom: 1px solid #ddd;
+	border-top: 1px solid #ddd;
+	}
+.leaflet-darkmode.leaflet-oldie .leaflet-control-zoom,
+.leaflet-darkmode.leaflet-oldie .leaflet-control-layers,
+.leaflet-darkmode.leaflet-oldie .leaflet-popup-content-wrapper,
+.leaflet-darkmode.leaflet-oldie .leaflet-popup-tip {
+	border: 1px solid #999;
+	}
```
Malvoz, line 637: I think we're doing away with `leaflet-oldie` soon, should we exclude it from this PR?

src/map/Map.js
```diff
@@ -120,7 +120,15 @@ export var Map = Evented.extend({
...
-		trackResize: true
+		trackResize: true,
+
+		// @option darkMode: Boolean = false
+		// Force Dark Mode of Leaflet
+		darkMode: false,
+
+		// @option useUserColorScheme: Boolean = true
+		// Use the user color scheme to switch between Light and Dark Mode
+		useUserColorScheme: true
```
jonkoops, line 131: Perhaps we could do multiple behavior here instead? For example, `'dark' | 'light' | 'auto'`?

dist/leaflet.css
Malvoz: `leaflet-oldie` was removed in https://github.com/Leaflet/Leaflet/pull/8559.

TurtIeSocks:
Anything I can help out with in this PR to move it along?

dist/leaflet.css
vanillajonathan: This could lead to two separate HTTP requests with additional round-trip delay and overhead, so I think it would be preferable to have the both light and dark in the same CSS file.

src/map/Map.js
vanillajonathan: Good idea!

```javascript
const prefersDarkTheme = window.matchMedia('(prefers-color-scheme: dark)');
```

Issue #7591, opened 2021-06-04
Gradually migrating to TypeScript


chrisvanmook:
**Is your feature request related to a problem? Please describe.**
As I don’t encounter any bugs at the moment, I am slightly worried about future maintenance of Leaflet, as I want to use Leaflet in my applications for at least a couple of years.  I’ve noticed that the code in some files are still written in old JS syntax and types need to be maintained separately. A mismatch between the types and the actual code could introduce some issues.

**Describe the solution you’d like**
Gradually migrating to TypeScript would be a solution to start with. Why?
- Able to use modern syntax while still supporting older browsers
- All the benefits of TypeScript
- If everything is converted to TS, the `@types/leaflet` package is no longer needed
- Overal better development experience

**Describe alternatives you’ve considered**
A solution with minimal impact on the build process and code. Being able to actually use TypeScript and converting existing JS files to TS would be a good start. I’ve quickly tried something out, seems that the build output is almost the same https://github.com/chrisvanmook/Leaflet/pull/1/files. 

I would like to know if you, the maintainers / contributors of leaflet, would be open for a similar change like this? In that case I will work on a pull-request, I might need some help with getting karma and leafdoc to work with TS.

IvanSanchez:
> I would like to know if you, the maintainers / contributors of leaflet, would be open for a similar change like this?

I'm *very* wary of TS. Specially when some of the Leaflet code still works around browser quirks, and I fear that TS will try to enforce a strict definition of properties available in `document`, `window` and so on. The little I've had to work with TS, it got in my way and felt like a burden.

Ideally I'd like work to be done on browser-loadable ESM modules (developing without needing to run a bundling toolchain feels *so* good), move the examples to `<script type=`module`>` and deprecate IE10 functionality (along with `L.SVG.VML`). I'm personally against complicating the toolchain.

chrisvanmook:
@IvanSanchez great to hear you want to move towards ES Modules! Do you have a roadmap / plan for this? Im curious. Regarding your worries about TypeScript, I'm not sure I fully understand. TypeScript indeed enforces the developer to use type definitions, that is it's purpose and that's a good thing, right? In some rare cases where this could be an issue and blocks you, you could always use `ts-ignore` or cast it as `any` or `unknown` and fix it later.  Perhaps in the current situation, where the `Leaflet` object is exported on `window`, it might be a little bit of an issue. 
[...]

Issue #7055, opened 2020-03-28
Add entrypoint for ES modules to package


rjackson:
Hi folks,

I've been digging around today trying to figure out why Webpack's tree shaking magic did not appear to be working on Leaflet's code. Digging back through the history here, I see the following historic PRs:

* February 2018: #6021 adds ES6 Module support, and unlocking all the goodies that come alongside
* March 2018: Leaflet/Leaflet.markercluster#874 is broken due to the lack of a global `L`
* July 2018: #6239 was a hotfix removing support for ES6 modules due to the above

In the removal hotfix, there was a comment that "people who want to take advantage of tree shaking can opt-in by adding custom rules resolving Leaflet to `dist/leaflet-src.esm.js`." I'll freely admit that I'm struggling to figure out how to do that (it may be pretty clear to be fair; I'm really not up to date with the modern Javascript ecosystem.)

Ever since that hotfix, [there's been a few PRs attempting to reintroduce module support](https://github.com/Leaflet/Leaflet/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aunmerged+esm+) but they've all been closed with a link back to the removal hotfix, and no further discussion.

The Leaflet/Leaflet.markercluster#874 issue remains open, with recent (January this year) comments floating the ideas of:

* A new major version release of Leaflet to enforce the removal of the `L` global variable
* Updating plugins to automatically resolve `L` to the `leaflet` module, rather than requiring Leaflet to expose `L` as a global variable.

Beyond this chatter on the Markercluster repo, I haven't seen any other discussion on how the Leaflet community is feeling about the lack of ES6 module support and the widespread (?) dependence on a global `L`. If I may, I'd like to be the one to bring these up for discussion.

Do we know what the remaining blockers are for:

* Removing the global `L`
* Bringing back ES6 module support?

And is there anything we can do make progress on these?

Thanks all,
Rob

bennypowers:
It would be great to just add the "module" field to package.json

ogdenstudios:
Hey everyone, I'm new to the project. Working with a team, considering migrating from Mapbox to Leaflet to save on performance. Would really really the ability to import at an ES6 module. Threw me for a loop this morning that I could not.

bennypowers:
You can import from `/dist/leaflet-src.esm.js`, but if your toolchain relies on the `module` field in package.json, as most tools do at the moment, you will be out of luck.

[...]

PR #7614, opened 2021-07-03
WIP: Rewrite classes to es6 syntax

mathe42:
This is a WIP rewrite to use the "new" es6 class syntax.

Goal is to intodruce no (or as little as possible) breaking changes. I try to keep the diff as small as posible for this PR so near 0 code changes. 

I create this PR now so there can be some feedback (by the leaflet-Team) about the change. I will stop working on this until some Team members had a look on this PR!

## Why?
ES6 was released in 2015 so 6 years ago. Since then the JS syntax got a loot more features that makes code easier and simpler to read. So I think it is a good thing if you use "modern" JS-Syntax.

## What is bad about the leaflet class strukture?
Nothing. 
But there are some advantages about the new syntax:

1. Typescript support can be added with simple JSDoc / TSDoc comments + maybe 1-2 .d.ts files.
2. IDE can infer a loot with the new syntax.
3. Better DX for new developers.

## What about old Browsers?
I transpile the code with esbuild back to es6 syntax and use `buble` or `babel` to create es5 syntax. This is important to remove the new class syntax and check if the prototype chain and all works properly.

I currently create 3 Output files:

* Modern (es2017 wich is supported by 95% of Browser user)
* es6 wich is support by many old versions of Browsers.
* legacy wich supports IE (currently not shure wich versions)

My plan is that the legacy build supports all browsers that leaflet currently supports. Output files have will be renamed so that there is no breaking change when finishing this PR.

But if a developer don't need IE support they can use the modern or es6 build. Or transpile the es2017 build to some other target.

Note to support IE <= 9 we need to transpile to es3 see https://caniuse.com/es5. Will propably use https://github.com/google/closure-compiler to do that.

## What about Plugins?
Plugins can use the old class syntax or can use the es6 class syntax. If they get transpiled to es5 / es3 there should be no problems to support even old IE. If useing es6 class syntax out of the box intellisense without adding `@types/leaflet`

## Current breaking changes
List of breaking changes. all of them I will try to fix. 

* Docs?
* Events Mixin is removed
* initHooks in parent class get called
* dist file names

## Summary
1. No breaking changes (if possible)
2. IE support
3. Moders JS sytax (not only classes)
4. Much smaller build if you only support modern browsers.
5. Needed for proper Treeshaking (there are some more problems before this (Mixins) works but this is needed! But that are breaking changes)

## Related
see #7591, #7055


mathe42:
Disccusion if this is a good efford or not should happen in #7615

src/layer/FeatureGroup.js
```diff
@@ -22,7 +22,7 @@ import {LatLngBounds} from '../geo/LatLngBounds';
  * 	.addTo(map);
  * ```
  */
-
+console.log(LayerGroup);
```
Falke-Design, line 25: console.log(LayerGroup);

mathe42: Oh Thanks! Will remove that. But as said above still WIP. So there might be some console.log or debugger calls in the code...

IvanSanchez (changes requested):

mathe42:
see #7651 for new pr

Issue #7615, opened 2021-07-03
Modern JavaScript


mathe42:
I think it is time to move to modern JavaScript in the source code.

> IMPORTANT: This is not a proposal to drop support for some Browser!

JavaScript evolved a lot in the last years. This Repro is (except for esModules that are handled by rollup) written in es3 way. This is JavaScript of 1999.

good:
* No transpiler needed
* Runs in all Browsers

bad:
* Bundle is bigger for all browsers with code they will never use
* Development is more complex as you have to check if you are allowed to use specific features.

## Proposal
Write source code in a modern syntax and transpile it down to differnet levels. I suggest transpiling for es3 and es2017. es2017 is supported by all evergreen Browsers (95% combined useage) and can be transpiled down as needed by the developer of a webpage.

We can provide more tragets as needed. The idea is not to add any breaking changes. So default entry file would still be the es3 version but developers can manualy change that. (Docs should be added how to to this in webpack, rollup, esbuild, ...)

## What changes in the source should happen
1. Better build with transpile to es3
1. Class Syntax to es6 classes (see #7614 - includes transpiling to es5)
2. Instead of most Util functions use the native ones (transpiler handles the rest)
3. Use modern syntax for example rest Arguments: `(a, b, ...rest) => {/* ... */}`

## What about typescript?
I think typescript is great but I think we should do one step at a time so if this issue is resolved in a positive way we can talk about typescript. If the above is done you can rewrite all code to ts in no time.

See also #7591

Atharva-Shinde:
I would like to work on this issue! #7615.

mathe42:
@Atharva-Shinde I want to wait for a Leaflet-Team member to say something before I continue working on it. - I don't have the time to do something like this and in 3 Month they say: "Nope modern JS will never happen in leaflet". I'm happy to put efford into it but not for nothing...


But great if you want to help.



If you (or anyone else) want to help you can:
1. Review my PR #7614 
2. Fork my Fork of Leaflet and modernize some code (I have a long list of todo's) - other than es6 classes as I have done near all of them
3. Check the docs - leafdocs has to be changed!?
4. I want to write some codemods that can be run on a Plugin Repro to modernize the code. Can be near / completely automatic. Why?
    * I want to try write some codemods - never have done that.
    * If the ecosystem can not modernize leaflet will also be behind. So as easy update methods as posible are needed!
[...]

PR #7618, opened 2021-07-09
changed functions to arrow functions

Atharva-Shinde:
All the functions inside src file are changed to arrow functions except export's functions.


mathe42:
This breaks support for some Browsers as they are not supported.

Not shure but arrow functions are different with this. That might be a problem did you run the tests?

mathe42:
And if my PR #7615 is merged this PR has near no changes. (or is a big Merge Conflikt with few places where this change helps.)

IvanSanchez:
Arrow functions just for the sake of arrow functions is a "no". There's *nothing* wrong with using the `function` keyword.

PR #7622, opened 2021-07-17
Check maxBounds, which may be undefined

utilmind:
options.maxBounds could be initially set, but removed by the moment when the event is fired (with l.fn.call(l.ctx || this, event), inside of fire method). If maxBounds already cleared, calling the _panInsideMaxBounds may produce chain of calls with undefined parameter and "Uncaught TypeError: latlng is undefined" at the end, in "project" method.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1261,7 +1261,7 @@ export var Map = Evented.extend({
 	},
 
 	_panInsideMaxBounds: function () {
-		if (!this._enforcingBounds) {
+		if (!this._enforcingBounds && this.options.maxBounds) {
 			this.panInsideBounds(this.options.maxBounds);
 		}
 	},


utilmind:
I wrote workaround inside of my app, now I'm calling .panTo() only after unsetting the maxBounds, but anyway proposed fix may be useful for someone.

utilmind:
Updating... The proposed fix required in any case, even with various workarounds the "TypeError: latlng is undefined" may occur when user zoom in/out map with maximum bounds too quickly, BUT simultaneously something unlock map maximum bounds.

utilmind:
Hi @Falke-Design! This was several months ago, I have already completed that project and switched my focus to other tasks (I'm freelancer, this was not long term project). I'm not sure that I can remember how to reproduce it now :(
I checked out the code, maybe the problem was because I used `setMaxBounds(false)` instead of `setMaxBounds(null)`, but I'm really not certain now.

So you may close this pull request, no problem. If I will work with Leaflet again and will have the same problem in the future, I will reopen this pull request with test/demo code.

Thanks!

PR #7626, opened 2021-07-23
leaflet-geojson-vt plugin added

iamtekson:
The leaflet plugin helps to render vector tiles for the GeoJSON data on the fly.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -794,6 +794,15 @@ Plugins to display [vector tiles](https://github.com/mapbox/vector-tile-spec).
 	</tr>
 	<tr>
 		<td>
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This looks like it overlaps with Leaflet.VectorGrid.Slicer. How is this different from VectorGrid functionality?

iamtekson:
Yes, it is a little bit similar to leaflet vectorgrid. But it is very lightweight and very easy to use. Just by adding few lines of code, the geojson vector grid can be generated. It extends the geojson-vt and geojson options.

jonkoops (approved):

PR #7631, opened 2021-08-04
add new in plugin lists

stefanocudini:
and updated old links


stefanocudini:
i can't figure out where the block is

IvanSanchez (approved):

PR #7633, opened 2021-08-07
bindTooltip should unbind existent tooltip

Falke-Design:
if a tooltip is bound to a layer with the option `permantent: true` and then a new tooltip is bound to the layer, the open permanent tooltip is not removed

grafik

```
var layer = new L.Marker(center).addTo(map);
layer.bindTooltip('Tooltip Test text', {permanent: true});
layer.bindTooltip('Other text', {permanent: false}).openTooltip();
```
https://jsfiddle.net/falkedesign/h6mogur9/

I think this can be added in the [Refactor DivOverlay/Popup/Tooltip](https://github.com/Leaflet/Leaflet/projects/2) project.


johnd0e:
"Overlapping" is not important
"permanent" is not important

What is really important: bindTooltip should close existent tooltip.

spec/suites/layer/TooltipSpec.js
```diff
@@ -283,4 +283,11 @@ describe('Tooltip', function () {
 		layer.bindTooltip('Tooltip', {interactive: true});
 		layer.closeTooltip();
 	});
+
+	it("closes already open permanent-tooltip on bindTooltip", function () {
+		var layer = new L.Marker(center).addTo(map);
+		layer.bindTooltip('Tooltip Test text', {permanent: true});
+		layer.bindTooltip('Other text', {permanent: false}).openTooltip();
+		expect(Object.keys(map._layers).length).to.be(2);
```
johnd0e, line 291:
```suggestion
		layer.bindTooltip('Tooltip1'}).openTooltip();
		local tooltip1 = layer.getTooltip();
		layer.bindTooltip('Tooltip2',);
		expect(map.hasLayer(tooltip1)).to.not.be.ok();
```
(not tested)

johnd0e (changes requested):

spec/suites/layer/TooltipSpec.js
```diff
@@ -283,4 +283,11 @@ describe('Tooltip', function () {
 		layer.bindTooltip('Tooltip', {interactive: true});
 		layer.closeTooltip();
 	});
+
+	it("closes already open permanent-tooltip on bindTooltip", function () {
```
johnd0e, line 287:
```suggestion
	it("closes existent tooltip on new bindTooltip call", function () {
```

johnd0e:
> What is really important: bindTooltip should close existent tooltip.

Correction: should **unbind** existent tooltip.

spec/suites/layer/TooltipSpec.js
johnd0e, line 291: layer.unbindTooltip = sinon.spy();
		layer.bindTooltip('Tooltip1'}).openTooltip();
		layer.bindTooltip('Tooltip2',);
		expect(spy.calledOnce).to.be.ok();

spec/suites/layer/TooltipSpec.js
```diff
@@ -303,4 +303,18 @@ describe('Tooltip', function () {
...
 		layer.closeTooltip();
 	});
+
+	it("closes existent tooltip on new bindTooltip call", function () {
+		var layer = new L.Marker(center).addTo(map);
+		var eventSpy = sinon.spy();
+		layer.unbindTooltip = function () {
+			eventSpy();
+			L.Marker.prototype.unbindTooltip.call(layer);
+		};
```
johnd0e, line 313: https://sinonjs.org/releases/v7.5.0/spies/#using-a-spy-to-wrap-an-existing-method

spec/suites/layer/TooltipSpec.js
```diff
@@ -303,4 +303,14 @@ describe('Tooltip', function () {
 		layer.bindTooltip('Tooltip', {interactive: true});
 		layer.closeTooltip();
 	});
+
+	it("closes existent tooltip on new bindTooltip call", function () {
+		var layer = new L.Marker(center).addTo(map);
+		var eventSpy = sinon.spy(layer, "unbindTooltip");
+		layer.bindTooltip('Tooltip1', {permanent: true});
+		var tooltip1 = layer.getTooltip();
+		layer.bindTooltip('Tooltip2').openTooltip();
```
johnd0e, line 312:
```suggestion
		layer.bindTooltip('Tooltip2').openTooltip();
		layer.unbindTooltip.restore(); // unwrap the spy
```

Falke-Design: I will change it 👍 Only for understanding: is this not automatically removed / restored when the layer is removed / deleted by the test?

johnd0e: Not sure, depends on implementation.
You could check by inspecting `L.Layer.prototype.unbindTooltip`

Falke-Design: `L.Layer.prototype.unbindTooltip` don't keeps a reference, so it would be in this case no problem without calling restore. But it is a cleaner code if it is called, so I added it.

spec/suites/layer/TooltipSpec.js
```diff
@@ -303,4 +303,15 @@ describe('Tooltip', function () {
...
+		expect(map.hasLayer(tooltip1)).to.not.be.ok();
+		expect(eventSpy.calledOnce).to.be.ok();
+		layer.unbindTooltip.restore(); // unwrap the spy
```
johnd0e, line 315: Wrong place. Must be before `expect` (because it throws)

johnd0e (approved):

PR #7651, opened 2021-09-01
feat(es6 class syntax): add new class syntax

mathe42:
see #7615

This PR replaces the base Class.

This should be near 100% compatibale but note the following changes that could break plugins (but I don't think they will):

3 internal fileds added: `_options`, `_getProto`, `_defaultOptions` additional you can not set `_initHooks`manualy it only has a getter.

This PR includes #7649 for build process.

~I removed `checkDeprecatedMixinEvents` for now but  I will add it back in later.~

### Known problems

There are some cases where the class syntax breaks but they are so rare that I think they don't exists:

```js
class A extends L.Class {}
const B = A.extend({})

const instance = new B() // creates error
```

Assuming leaflet was imported as the es3 build. If you go from the es3 syntax to es6 and back it breaks.


mourner:
Hey @mathe42, sorry for not reviewing these PRs yet — it's a lot to process, and big, potentially breaking changes like this should go through a lot of scrutiny and careful consideration.

mathe42:
@mourner Thanks for the update! I currently work on some test cases to debug / test where possible breaking changes are and eliminate them.

Would it help if I would test some plugins (from https://leafletjs.com/plugins.html ) with my replacement? - Maybe some of the big ones and some random ones? To get a feeling if I test 20 and no one is failing this is a good thing... And maybe some small incossistencies can be fixed with that.

mathe42:
I close this as this is bad code by me. - sorry. Found a better solution (but that only works in IE 9 or higher [I need `__proto__` or `classInstance.constructor.prototype`]) currently looking for a solution.

The problem is with calling initHooks and properly define options. But I work on solutions.

mathe42:
But (for me) more important is #7649. If this (or a similar) build setup is established there are lots of things that will be better...

PR #7653, opened 2021-09-03
Add a Dependabot configuration to keep dependencies up-to-date

jonkoops:
Also adds a package-lock.json file to ensure broken dependencies are not included by accident, as [recommended](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json) by the NPM documentation:

> This file is intended to be committed into source repositories, and serves various purposes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: /
+    schedule:
+      interval: weekly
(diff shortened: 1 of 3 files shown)


mourner (approved):
All right, let's try this! Previously I didn't include `package-lock.json` because I used Yarn but didn't want to force it on other contributors, but it seems we're good to use NPM again these days.

Weekly Dependabot alerts might be a little too nagging, and we might change that to monthly later (it only concerns dev dependencies anyway, since Leaflet does not have production ones), but let's see how it goes.

Issue #7611, opened 2021-07-01
Switch to GitHub Actions for CI


mourner:
Travis CI suddenly stopped running builds for Leaflet and many of my other open source repos for unknown reasons. Rather than try to figure out what's wrong, we should just switch to GitHub Actions — it's part of GitHub, it's more flexible and runs faster.

Example config on one of my projects: https://github.com/mourner/flatbush/blob/master/.github/workflows/node.yml

mathe42:
I worked a little bit with GH-Actions if you are interessted I could set something up...

mourner:
@jonkoops thanks again for landing this! 

Before I can close this, to finish fully migrating to GH Actions, we also need to port over the [artifacts bit](https://github.com/Leaflet/Leaflet/blob/master/.travis.yml#L13-L32) in the Travis config which was responsible for uploading a distribution of Leaflet (for download) for every commit on the `master` branch.

jonkoops:
@mourner Sure, I can take a look at that as well. Where exactly were those files being uploaded before?

FYI I also opened #7660 which allows the tests to run on all the different browsers properly.

mourner:
The S3 upload is working now, all good!

PR #7654, opened 2021-09-03
Run CI on Github actions

jonkoops:
Depends on #7653 as a `package-lock.json` file is needed to run `npm ci`. Allows running the CI on Github actions instead of Travis, as it looks like the Travis builds are not running. Added benefit is that forks can use the same configuration to run CI without having to set up Travis and change the source.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,5 +11,6 @@ jobs:
           check-latest: true
           cache: npm
       - run: npm ci
+      - run: npm run lint
       - run: npm test -- -- --browsers PhantomJSCustom,Chrome1280x1024,FirefoxPointer,FirefoxTouch,FirefoxPointerTouch --reporters
       - run: npm run build
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,9 +1,5 @@
 name: CI
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+on: push
 jobs:
   ci:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,19 @@
+name: CI
+on:
+  push:
+    branches: [master]
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,19 @@
+name: CI
+on:
+  push:
+    branches: [master]
```
mourner, line 4: I think this should be any branch on `push` since we want CI to run in PRs before they are merged.

```suggestion
    branches: ['*']
```

mourner (changes requested):
Thanks for looking into this! Closes #7611. I merged #7653 so let's rebase again.

.github/workflows/main.yml
jonkoops: @mourner I've changed this up to run on any `push` event regardless of branch.

jonkoops:
> Thanks for looking into this! Closes #7611. I merged #7653 so let's rebase again.

Interesting, I hadn't even spotted this issue yet. PR is rebased and review comments have been processed.

mourner (approved):

mourner:
Not sure why it doesn't run now — maybe because it's a third party PR, but let's merge and see what happens.

mourner:
Running great — took just 49s for the uncached job! https://github.com/Leaflet/Leaflet/runs/3525418592

jonkoops:
@mourner Nice, that is a lot faster than I expected!

jonkoops:
It looks like the tests only ran on PhantomJS though, I guess that is not entirely correct considering the browser parameters passed in. This might be related to my changes to the scripts in the package.

I'll open up a seperate PR to debug this, hopefully that will just run the task now that an initial version was merged.

Psoffie009:
> Depends on #7653 as a `package-lock.json` file is needed to run `npm ci`. Allows running the CI on Github actions instead of Travis, as it looks like the Travis builds are not running. Added benefit is that forks can use the same configuration to run CI without having to set up Travis and change the source.

PR #7660, opened 2021-09-06
Run tests on browsers instead of PhantomJS

jonkoops:
Follow up on #7654 to allow tests to run on actual browsers instead of on PhantomJS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: CI
-on: push
+on: [push, pull_request]
 jobs:
   ci:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jonkoops:
@mourner This change allows the tests to run properly on the different browsers. I've also added a change to the `on` section to run the tasks on a PR as simply providing `push` was not enough to actually start the build.

mourner (approved):

Psoffie009:
> Follow up on #7654 to allow tests to run on actual browsers instead of on PhantomJS.

Psoffie009:
> Follow up on #7654 to allow tests to run on actual browsers instead of on PhantomJS.

PR #7662, opened 2021-09-07
TileLayer feature: Add failover for subdomains

miallo:
If one of the subdomains fails, currently the tiles for this will not be rendered. With this PR it now tries to get the tile from the next subdomain until all servers were tried.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -136,8 +136,10 @@ export var TileLayer = GridLayer.extend({
 	createTile: function (coords, done) {
 		var tile = document.createElement('img');
 
+		tile.dataset.failoverOffset = 0;
+
 		DomEvent.on(tile, 'load', Util.bind(this._tileOnLoad, this, done, tile));
-		DomEvent.on(tile, 'error', Util.bind(this._tileOnError, this, done, tile));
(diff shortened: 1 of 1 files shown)


miallo:
I could __not__ run the tests, because I could not get PhantomJS to work, so hopefully this is done in CI, or maybe a reviewer could help me out on this one :)

src/layer/tile/TileLayer.js
```diff
@@ -219,8 +227,8 @@ export var TileLayer = GridLayer.extend({
 		return zoom + zoomOffset;
 	},
 
-	_getSubdomain: function (tilePoint) {
-		var index = Math.abs(tilePoint.x + tilePoint.y) % this.options.subdomains.length;
+	_getSubdomain: function (tilePoint, failoverOffset = 0) {
```
mathe42, line 230: This "new" JS syntax is (currently) not supported.

Replace it with something like

```js
_getSubdomain: function (tilePoint, failoverOffset) {
  if(failoverOffset === undefined) failoverOffset = 0
```

I have #7649 pending to support this and other syntax features

jonkoops:
I am not quite sure I like the solution or if it should even be included in the source. It seems to me that `TileLayer` could be extended in a project or plugin to provide this functionality outside of core Leaflet.

johnd0e:
Agree. Good idea for plugin.

PR #7663, opened 2021-09-08
Added leaflet-route-print to plugin list

hersle:
I am still working on tuning the plugin to make it completely finished, but I think it has reached a releasable and usable state now.
Do let me know if I should make any changes before it gets added to the list!


jonkoops (approved):

hersle:
Thank you! :)

Issue #7528, opened 2021-03-27
404 in Mobile Example


arduino-man:
- [✔️] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [✔️] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [✔️] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
-step1:
    Visit: https://leafletjs.com/examples/mobile/example.html
-step2:
    Open your browser developer tools and look at the Javascript console.
-step3:
    Locate the 404 message saying: GEThttps://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/-1/0/0?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw
[HTTP/1.1 404 Not Found 26ms]

**Expected behavior**
The expected behavior would be no 404 errors with the Mapbox API such as in the quickstart example found at: https://leafletjs.com/examples/quick-start/example.html

**Current behavior**
The current behavior displays a 404 error complaining that the API URL cannot be found which I also believe is causing a significant delay on the map loading while it somehow recovers from the crash/404.

**Environment**
- Leaflet version: leaflet@1.7.1
- Browser (with version): Firefox 87.0
- OS/Platform (with version): Ubuntu 20.04

**Additional context**
No further context

Example:
https://leafletjs.com/examples/mobile/example.html

- [✔️] this example is as simple as possible
- [✔️] this example does not rely on any third party code

johnd0e:
I can confirm the bug, with such prerequisites:
- geolocation forbidden
- size of map<512px (either x or y)
   If both dimensions > 512 - different tile is requested: https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/0/0/0?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw

IvanSanchez:
At first sight, this seems that that example is gonna need a `minZoom: 1` setting.

arduino-man:
I've confirmed that adding "minZoom: 1" fixed the issue. Thanks a lot.

[...]

PR #7664, opened 2021-09-09
TyleLayer _getZoomForUrl avoid returning negative zoom

shlomi-lachmish:
TyleLayer _getZoomForUrl avoid returning negative zoom (Leaflet#7528)


IvanSanchez:
Nope. See https://github.com/Leaflet/Leaflet/issues/7528#issuecomment-808752337

PR #7677, opened 2021-09-19
Upgrade to latest version of Rollup

jonkoops:
Also cleans up the build configuration itself a little so it's a bit more DRY and conforms to the linting standards of the project.


mourner (approved):
After studying https://github.com/rollup/rollup/pull/2141 very carefully, I discovered that we can keep compatibility with IE8 if we use both `freeze: false` and `esModule: false`. 🎉 Haven't actually tested on IE8 (lol), but pretty sure we can proceed with the upgrade.

mourner:
cc @IvanSanchez

jonkoops:
Awesome @mourner! Should we also update the supported browsers on the site? It would seem that Internet Explorer 7 isn't supported anymore if I read this correctly.

image

PR #7685, opened 2021-10-01
add tests setClass and getClass for DomUtil

micro-chipset:
issues: https://github.com/Leaflet/Leaflet/issues/7435

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
 		});
 	});
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
+			el.className = 'bar foo baz';
+			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
+		});
+		it('should return empty string if no className', function () {
+			el.id = 'testId';
+			expect(L.DomUtil.getClass(el)).to.eql('');
+		});
+	});
+
+	describe('setClass', function () {
+		it('should set className', function () {
+			el.className = '';
+			L.DomUtil.setClass(el, 'bar');
+			expect(el.className).to.eql('bar');
+		});
+
+		it('should set classNames', function () {
+			el.className = '';
+			L.DomUtil.setClass(el, 'bar foo baz');
+			expect(el.className).to.eql('bar foo baz');
+		});
+	});
+
 	// describe('#setPosition', noSpecs);
 
 	// describe('#getStyle', noSpecs);
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -54,24 +54,24 @@ describe('DomUtil', function () {
 	});
 
 	describe('getClass', function () {
-		it('should return the element\'s class', function () {
+		it('return the element\'s class', function () {
 			el.className = 'bar foo baz';
 			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
 		});
-		it('should return empty string if no className', function () {
+		it('return empty string if no className', function () {
 			el.id = 'testId';
 			expect(L.DomUtil.getClass(el)).to.eql('');
 		});
 	});
 
 	describe('setClass', function () {
-		it('should set className', function () {
+		it('set className', function () {
 			el.className = '';
 			L.DomUtil.setClass(el, 'bar');
 			expect(el.className).to.eql('bar');
 		});
 
-		it('should set classNames', function () {
+		it('set classNames', function () {
 			el.className = '';
 			L.DomUtil.setClass(el, 'bar foo baz');
 			expect(el.className).to.eql('bar foo baz');


micro-chipset:
Please add he label 'hacktoberfest', waiting for feedback against the PR.

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
 		});
 	});
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
```
jonkoops, line 57: I think it's better to keep the test descriptions consistent and omit the 'should' here.
```suggestion
		it('returns the element\'s class', function () {

```

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
...
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
+			el.className = 'bar foo baz';
+			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
+		});
+		it('should return empty string if no className', function () {
+			el.id = 'testId';
+			expect(L.DomUtil.getClass(el)).to.eql('');
+		});
```
jonkoops, line 64: We can omit the test case for en empty class name, this doesn't change anything about the tested implementation compared to the previous test.
```suggestion
```

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
...
+	});
+
+	describe('setClass', function () {
+		it('should set className', function () {
+			el.className = '';
+			L.DomUtil.setClass(el, 'bar');
+			expect(el.className).to.eql('bar');
+		});
+
```
jonkoops, line 73: This test case can be removed as there is no difference between the code paths hit between this test and the next.
```suggestion
```

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
...
+		it('should set classNames', function () {
```
jonkoops, line 74: Also here it would be preferred if 'should' is omitted from the description. Maybe also we can keep these descriptions consistent with the other one you added.
```suggestion
		it('sets the element\'s class', function () {

```

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
...
+			el.className = '';
```
jonkoops, line 75: There is no need to reset the className here as it is never assigned another value before.
```suggestion
```

jonkoops (changes requested):

jonkoops:
@micro-chipset Can you also add test cases for SVG elements? These have a different path in the code if you check the implementation of `setClass()` and `getClass()`.

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
 		});
 	});
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
+			el.className = 'bar foo baz';
+			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
```
johnd0e, line 59:
```suggestion
			expect(L.DomUtil.getClass(el)).to.be('bar foo baz');
```

spec/suites/dom/DomUtilSpec.js
johnd0e: Objection: we expect empty string here. Not `undefined` or so.
I suppose the test make sense.

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
...
 	});
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
+			el.className = 'bar foo baz';
+			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
+		});
+		it('should return empty string if no className', function () {
+			el.id = 'testId';
+			expect(L.DomUtil.getClass(el)).to.eql('');
```
johnd0e, line 63:
```suggestion
			expect(L.DomUtil.getClass(el)).to.be('');
```

spec/suites/dom/DomUtilSpec.js
```diff
@@ -53,6 +53,31 @@ describe('DomUtil', function () {
 		});
 	});
 
+	describe('getClass', function () {
+		it('should return the element\'s class', function () {
+			el.className = 'bar foo baz';
+			expect(L.DomUtil.getClass(el)).to.eql('bar foo baz');
+		});
+		it('should return empty string if no className', function () {
+			el.id = 'testId';
```
johnd0e, line 62: Why should we care about `id`?

johnd0e (changes requested):

spec/suites/dom/DomUtilSpec.js
jonkoops: Under which circumstances would `className` be `undefined` here though? I honestly do not see any code paths that would justify this test.

johnd0e: > I honestly do not see any code paths that would justify this test.

We write tests not only for current code paths. Imagine tomorrow someone would like to refactor the function.

PR #7688, opened 2021-10-03
Fix map display on larger screens

### LINKED ISSUE #7420: Examples are not mobile-friendly

branko-stankovic:
Fixes map displaying on desktop screens, while remaining responsive on mobile devices.

Fix #7420

### DIFF
--- a/docs/_layouts/tutorial_frame.html
+++ b/docs/_layouts/tutorial_frame.html
@@ -19,8 +19,8 @@
 			margin: 0;
 		}
 		#map {
-			width: 600px;
-			height: 400px;
+			max-width: 100%;
+			height: 100vw;
 		}
 	</style>
 {% endunless %}


docs/_layouts/tutorial_frame.html
```diff
@@ -19,8 +19,8 @@
 			margin: 0;
 		}
 		#map {
-			width: 600px;
```
johnd0e, line 22: Should we make examples fullscreen?
@mourner 
@jonkoops

Falke-Design: I think we have no disadvantages with fullscreen. Additionally then there are no problem with mobile devices but this can be done with `@media` .

I personally don't like `height: 100vh`. I use in my projects `height: 100%` and add it to `html, body` too, but I don't know what is better

branko-stankovic:
I appreciate the feedback, thank you for your time to review it!

PR #7693, opened 2021-10-08
Tutorial to convert layer to a geoJSON object

### LINKED ISSUE #6904: Add style info in toGeoJSON result
The `toGeoJSON` method only returns position information. Things like marker icon and title are not included. It would be great to have style information included.

Usecase: I am using [Leaflet StyleEditor](https://github.com/dwilhelm89/Leaflet.StyleEditor) together with Leaflet Draw ([demo](https://dwilhelm89.github.io/Leaflet.StyleEditor/StyleEditorWithLeafletDraw.html)) to create a visual map editor inside of wiki pages. These wiki pages store GeoJson which is edited with the map.

Relevant code for markers: https://github.com/Leaflet/Leaflet/blob/f8e09f993292579a1af88261c9b461730f22e4e6/src/layer/GeoJSON.js#L305-L310 (this issue is about all layers, not just markers)

I am not sure if it is appropriate to add the style info in `toGeoJSON` and am not set on any specific solution. Going from GeoJson with style info to a layer on the map also seems to be missing, which might affect how this is best implemented. (Creating styled GeoJson via http://geojson.io shows up unstyled in my Leaflet maps)

I'm willing to contribute to the Leaflet library itself to make this happen, if that is the best place, and if my use case can not already be done in some other manner.

hamza-arshad:
Added the tutorial to convert the Leaflet layer to a geoJSON object.

Fixes #6904

### DIFF
--- a/docs/examples/geojson/index.md
+++ b/docs/examples/geojson/index.md
@@ -214,12 +214,24 @@ L.geoJSON(someFeatures, {
     radius: 500
 })</code></pre>
 
-<p> We can make use of the <code>toGeoJSON</code> method and <code>options</code> attribute of <code>layer</code> (<code>circle</code> in above case) to convert it to a GeoJSON object</p>
+<p> We can make use of the <code>toGeoJSON</code> method of <code>layer</code> (<code>circle</code> in above case) to convert it to a GeoJSON object</p>
+
+<pre><code>var geoJSONObject = circle.toGeoJSON()</code></pre>
+
+<h4>Coords Precision</h4>
+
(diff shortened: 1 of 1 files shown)


hamza-arshad:
would like to request you to add a `hacktoberfest-accepted` label if this PR is merged later on. Thanks.

johnd0e:
https://github.com/IITC-CE/ingress-intel-total-conversion/blob/master/plugins/overlay-kml.js
Here is a sample of GeoJSON styling, based on https://github.com/mapbox/simplestyle-spec.

PR #7694, opened 2021-10-09
1.7.2 with typescript and snyk

luismendes070:
eslint changes
typescript update


mourner:
No idea what's going on here but it surely must be an accidental PR?

PR #7696, opened 2021-10-13
Use consistent map HTML element ID everywhere

fulldecent:
This makes it consistent with the quick start guide so you can copy-paste between the two


jonkoops:
I would have to agree with @Falke-Design, I would expect all examples to use `map` as the identifier.

PR #7698, opened 2021-10-17
Adds a Map option to set the timeout for firing the contextmenu event.

Mortvola:
The current timeout in Map.Tap.js is set to 1000 ms. iOS provides tactile feedback at around 600ms on a long touch. Users are likely to end their long touch after receiving the tactile feedback and wonder why the context menu did not appear.

This PR does not change the current value but does allow application developers to change the timeout to something that is reasonable for their purposes.


johnd0e:
There is a very little sense because we are unable to make such option work consistently across all the systems.
The best we can do - pick a value close to observable on real systems.

PR #7703, opened 2021-10-20
Make docs headings sticky

avioli:
Every now and again, when I browse the documentation I find myself unsure which namespace I'm on or what options I'm looking at, so using a bit of CSS and `position: sticky` I've the headings stay comfortable put at the top of the screen, both on desktop and mobile.

https://user-images.githubusercontent.com/524259/138038696-a73b4531-8b8e-4825-9236-1ef914684e5c.mp4

Are there any test for docs or did I edit some auto-generated files?

### DIFF
--- a/docs/docs/css/main.css
+++ b/docs/docs/css/main.css
@@ -488,6 +488,10 @@ table.plugins td:first-child a {
 	color: #333;
 	border-bottom: 3px solid #555;
 	transition: border-color .25s ease;
+	position: sticky;
+	top: 0;
+	z-index: 1;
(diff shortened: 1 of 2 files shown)


avioli:
Actually a better showcase would have been to search for something off-screen, like say `getPane`, which lands me on the Popup namespace, but in the methods section:

https://user-images.githubusercontent.com/524259/138039448-d8d7a2f6-abeb-4af2-92fd-ec743233dead.mp4

mourner (approved):
Brilliant!

Issue #2669, opened 2014-05-05
Toggle Clickable


eborden:
There is no ability to toggle the clickable state of a marker/layer. You can toggle dragging on/off, but clickable has no such feature.

johnd0e:
Closed in favor of #5442

Issue #5442, opened 2017-04-07
Feature: new method setInteractive(interactive) for interactive layers


perliedman:
For interactive layers (layers with the `interactive` option, or extending the virtual `InteractiveLayer` class), we could provide a `setInteractive` method to change the state after the layer is created.

As it is now, there's no way to change the interactive state after a layer has been created, causing issues like #5441.

maxdeepfield:
any plans about this feature development?

perliedman:
@maxdeepfield no plans, as far as I know.

I'd be happy to guide if someone would be willing to take this on.

Ulrikop:
I would like to see a method like that. It feels false to re-create a marker/line just because I want to change the behavior of the interactivity

verbeeckjan:
I really need this functionality too, any updates or workarounds @perliedman?

jmayer86:
Thanks @Jadaw1n ! That worked like a charm, here it is in Typescript for anyone who wants it:

`export class InteractiveImageOverlay extends L.ImageOverlay {
    setInteractive(interactive: boolean) {
        this.options.interactive = interactive;

        if (interactive) {
            L.DomUtil.addClass(this.getElement(), 'leaflet-interactive');
            this._map.on('click', (event: L.LeafletMouseEvent) => {
                this.fireEvent('click', event);
            });
        } else {
            L.DomUtil.removeClass(this.getElement(), 'leaflet-interactive');
            this._map.removeEventListener('click');
        }
    }
}`

jktu2870:
@Jadaw1n it is a nice solution you've shared, unfortunately for me it doesn't work with canvas elements

DrPitLazarus:
@Jadaw1n Wanted to say thanks for your code. I ran into this issue changing my rendering from canvas to SVG. Surprised this isn't in leaflet by now.

PR #7708, opened 2021-10-27
Allow to change interactive option of layer

Falke-Design:
Makes it possible to toggle `interactive` options over `setStyle`.
New function `setInteractive` for ImageOverlay because it has no setStyle

Fix: #2669, #5442

### DIFF
--- a/spec/suites/layer/ImageOverlaySpec.js
+++ b/spec/suites/layer/ImageOverlaySpec.js
@@ -157,4 +157,28 @@ describe('ImageOverlay', function () {
 			});
 		}
 	});
+
+	describe("#interactive", function () {
+		var blankUrl = "data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw==";
+		it("should have class leaflet-interactive", function () {
+			var overlay = L.imageOverlay(blankUrl, imageBounds, {interactive: true}).addTo(map);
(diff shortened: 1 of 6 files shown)


johnd0e:
There is no reliable way to change `interactive` option, because there may be special actions in constructor.

Falke-Design:
@johnd0e in which constructor? Currently we don't have special actions in one of the constructors ... Only in the `onAdd` & `onRemove` and if it can added / removed, it can be toggled too.

This is a very helpful feature which makes it possible to disable all listeners on a layer without removing them.

I would like to reopen this PR and discuss about this. @mourner do you see it like @johnd0e?

johnd0e:
> in which constructor?

`initialize` function

> Only in the `onAdd` & `onRemove`

So the option can be safely changed only when feature is not on map (in the moment of `setStyle` call).
Pretty useless, as for me.

johnd0e:
Imho it is better to refactor leaflet to include common extension methods like `_initInteraction(true/false)`, and then we can implement changing of this option in any way (preferable dedicated, not `setStyle`).

Falke-Design:
@johnd0e I still think it would be useful (and working) if the interactive can changed while the layer is on the map. But it sounds like a good solution to add a general function `_initInteraction(true/false)` for layers. But in my eyes it would be better to make it a public function.

Btw. is it wanted that you named the function `_initInteraction` and not `_initInteractive`?

johnd0e:
`setInteractive(boolean)` for public method.
And _initInteraction()` for private (if we need it, which is not clear yet.)

Issue #5253, opened 2017-01-13
Attribution doesn't observe minZoom and maxZoom


kingrollo:
Layer attribution appears in the attribution box even when the layer is invisible in the map, if this is controlled by minZoom or MaxZoom.

Please see here: http://playground-leaflet.rhcloud.com/riv/edit?html,output - zoom out 1 level to see the alternative tile layer

I can see how to update the attribution myself using addAttribution and removeAttribution, but this seems strange default behaviour to me - shouldn't the attribution control handle this automatically?

Thanks!

IvanSanchez:
Interesting find. This will need a bit of investigation to see why it's happening.

PR #7716, opened 2021-10-30
Remove attribution if tiles not visible

Falke-Design:
Fix: #5253

Attributions are now removed if tiles not visible. This can be because of `minZoom`, `maxZoom` and `bounds`.

Also added the option `getAttributionText` to `Control.Attribution`.
attribution_tiles
attribution_tiles_move

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -1095,4 +1095,73 @@ describe('GridLayer', function () {
 		grid.redraw();
 		expect(wrapped.neverCalledWith(sinon.match.any, null)).to.be(true);
 	});
+
+	describe("Attribution adding / removing", function () {
+		var clock;
+		beforeEach(function () {
+			map.setView([0, 0], 1);
+			clock = sinon.useFakeTimers();
+		});
+
+		it("shows attribution only when layer is between min and maxZoom", function () {
+
+			L.gridLayer({
+				attribution: 'Grid 1',
+				maxZoom: 20,
+				minZoom: 1
+			}).addTo(map);
+
+			L.gridLayer({
+				attribution: 'Grid 2',
+				maxZoom: 10,
+				minZoom: 5
+			}).addTo(map);
+
+			// Grid 2 should be hidden on Zoom 1
+			expect(map.attributionControl.getAttributionText().indexOf('Grid 2') > -1).to.be(false);
+
+			// Grid 2 should be visible on Zoom 10
+			map.setZoom(10, {animate: false});
+			clock.tick(250);
+			expect(map.attributionControl.getAttributionText().indexOf('Grid 2') > -1).to.be(true);
+
+			// Grid 2 should be hidden on Zoom 11
+			map.setZoom(11, {animate: false});
+			clock.tick(250);
+			expect(map.attributionControl.getAttributionText().indexOf('Grid 2') > -1).to.be(false);
+
+			// Grid 2 should be visible on Zoom 5
+			map.setZoom(5, {animate: false});
+			clock.tick(250);
+			expect(map.attributionControl.getAttributionText().indexOf('Grid 2') > -1).to.be(true);
+
+			// Grid 2 should be hidden on Zoom 4
+			map.setZoom(4, {animate: false});
+			clock.tick(250);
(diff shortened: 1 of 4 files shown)


spec/suites/layer/tile/GridLayerSpec.js
```diff
@@ -1095,4 +1095,82 @@ describe('GridLayer', function () {
 		grid.redraw();
 		expect(wrapped.neverCalledWith(sinon.match.any, null)).to.be(true);
 	});
+
+	describe("Attribution adding / removing", function () {
+		var clock;
+		beforeEach(function () {
+			map.setView([0, 0], 1);
```
Falke-Design, line 1102: The error in "Loads 8 kittens zoom 1" happens because of `map.setView([0, 0], 1);` ... I didn't find out why this happens. In my Browser over the karma `index.html` the tests are working, only on `PhantomJS 2.1.1 (Windows 8)` it breaks.
@mourner @johnd0e @IvanSanchez does someone of you know how to fix this in tests? I searched for 30 minutes

Falke-Design: The error:
> PhantomJS 2.1.1 (Windows 8) TileLayer number of kittens loaded Loads 8 kittens zoom 1 FAILED
        TypeError: null is not an object (evaluating 'this._map._fadeAnimated') (src/Leaflet.js:11510)

src/layer/tile/GridLayer.js
```diff
@@ -552,6 +552,8 @@ export var GridLayer = Layer.extend({
 			tileZoom = this._clampZoom(tileZoom);
 		}
 
+		this._attributionCheck();
```
johnd0e, line 555: I doubt that `GridLayer` should know anything about attribution.
The things should be opposite: `GridLayer/TileLayer` fire events - `Control.Attribution` listens and makes changes.

johnd0e (changes requested):

src/layer/tile/GridLayer.js
Falke-Design: > I doubt that GridLayer should know anything about attribution

Do you mean that it should not possible to add a attribution for Gridlayers? Then the [doc](https://leafletjs.com/reference.html#gridlayer) is wrong.

What do you thing would be better? Removing / adding the complete attribution of the layer with the event or only the visiblity? Is the validation on the right place or should it moved to somewhere else?

johnd0e: > Do you mean that it should not possible to add a attribution for Gridlayers?

I do not mean that.
I mean that juggling with attributions should be implemented in `L.Control.Attribution`, reacting on events like `moveend`/`resetview`/etc.

Falke-Design: I already wanted to make this, but the problem is the complete `Control.Attribution` needs to be changed, because currently only text can be added / removed. And then I don't wanted to change the current usage of managing text.
If we change the code of the Control to manage layers instead of text, I think we should also think about to add the same "validation" for layers as for Gridlayers. If the layer is out of the view, the attribution of the layer is not displayed.

johnd0e: I'd suggested to implement additional events, like "getvisible"/"gethidden" (or similar).
May be it is needed to be done separately in:
- Layer (add/remove)
- GridLayer (bounds)
- TileLayer (minZoom/maxZoom)

johnd0e: Or in some other way. Let me think..

johnd0e:
Here is my considerations: This PR is useless while attribution stays static.

To make proposed features consistent we should:
1. Implement 'dynamic' option for `L.Control.Attribution` and listen for map's viewport changes. 
2. Every Layer (and it's ancestors) may implement sort of `isVisible` method (alternatively: 'visibilitychanges`).

Falke-Design:
I think events `visibilitychange` and a function `isVisible()` would make sense for plugins too. I like this suggestion

johnd0e:
BTW:
https://github.com/Leaflet/Leaflet/blob/93ebeb7786772c95c572737a4b026b15899e7826/src/layer/Layer.js#L116-L118
I do not feel that this is proper place for such code.
IMHO all this should be encapsulated in Control.Attribution implementation, listening on map's `layeradd`/`layerremove`.

PR #7723, opened 2021-11-01
Continue running tests even if one fails

jonkoops:
Sets the `fail-fast` option for the tests on the CI to `false`. This will prevent Github Actions from cancelling other test jobs if one of them fails, which allows us to see the result of all the tests runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,6 +60,7 @@ jobs:
   test:
     needs: setup
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
       matrix:
         browser:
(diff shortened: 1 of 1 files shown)


mourner (approved):

mourner (changes requested):
Hmm, according [to the docs](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), this should also make  the workflow pass even if the `test` job fails, which is definitely not something we want. Can we check to make sure?

mourner:
@jonkoops OK, looks what we want instead is [`fail-fast: false`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) (it's true by default).

jonkoops:
Ah yes, I always get these confused thanks @mourner

mourner (approved):

PR #7726, opened 2021-11-01
Add npm scripts `test-all` and `pre-commit`

Falke-Design:
New npm scripts which should it make easier for contributers to test and push valid code.

- With `npm run test-all` all tests can be run in all browsers (same as Github CI)
- With `npm run pre-commit` all scripts that should be made before commiting / pushing can be done with one call. First building, then linting and then run all tests in all browsers and check if there is no `.only` test

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,8 @@
   "scripts": {
     "docs": "node ./build/docs.js",
     "test": "karma start ./spec/karma.conf.js",
+    "test-all": "karma start ./spec/karma.conf.js --browsers Chrome1280x1024,FirefoxPointer,FirefoxTouch,FirefoxPointerTouch",
+    "pre-commit": "npm run build && npm run lint && npm run test-all -- --forbidOnly",
     "build": "npm run rollup && npm run uglify",
     "release": "./build/publish.sh",
     "lint": "eslint src spec/suites docs/docs/js build",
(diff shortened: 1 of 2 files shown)


jonkoops:
I think the tests you might want to run locally are highly subjective as some browsers may not even be installed or cannot be installed locally due to them being made for a different operating system. Since this is also covered by the CI I don't think it is useful to make this change.

I do see value in a pre-commit hook of sorts to automatically lint and fix changes, something like husky combined with lint-staged. However this further increases the project complexity as well and I don't know if that is the right move at this time.

Falke-Design:
Ah okay, I didn't know that the Browsers needs to be installed, I thought that they are installed anyway over package.json. Then it is clear that this will not work.

A common mistake of contributers are that they not run the tests or linting because they forgot it. This can **not** prevented with `pre-commit`, but it will help if someone want to run all necessary things but don't know (exactly) what he needs to do.  If it is only one command it is easier to not forget a step. (And the check --forbidOnly is currently not existing but I think this would be valuable too)

johnd0e:
https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md

Falke-Design:
> https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md

it is not a problem of not knowing what should be done (maybe sometimes), it is a thing of forgetting to run, because multiple commands needed to be executed.

And as so often the CONTRIBUTING.md is not easy / simple to read. It has not clear steps 1. run lint 2. run test 3. check for .only 4. ... So if many steps can be simplified by calling (optional) one script, I don't understand how this increases the complexity.

PR #7727, opened 2021-11-02
Simplify release process

### LINKED ISSUE #7718: Retire `publish.sh` and simplify release process
The [publish.sh](https://github.com/Leaflet/Leaflet/blob/master/build/publish.sh) is quite outdated — we should eliminate it completely, and make it possible to release Leaflet by just running e.g. `npm version patch`, and pushing the generated tag to GitHub.

The only special thing that it does (apart from running test, build, creating a tag etc.), is this line: 

https://github.com/Leaflet/Leaflet/blob/0ae68f4658cb1cd85e6e0c67a24b12e5e6282e22/build/publish.sh#L21

I traced it back to this ancient issue from 2013 #1903 about supporting Bower. We don't care about Bower anymore so we can just stop doing this.

After that and #7712, we can configure GitHub Actions to do the actual `npm publish` and a docs update automatically on every tag, removing laborious and error-prone manual process.

mourner:
Closes #7718.

- Remove the outdated `publish.sh`.
- Add `prepublishOnly` a script that runs automatically on `npm publish`, making a clean dependencies install, running tests, generating the final builds and aborting if any of these steps fails.
- Update `RELEASE.md`, removing all unnecessary steps

A follow-up simplification would be to automate as much of the "docs" section as possible, eliminating manual steps like copying files, editing content manually etc., but I'll leave that to a different PR.


Falke-Design:
I would suggest to use `reference.html` for the newest doc instead of redirecting to `reference-1.7.1.html`. 
If a link of a method / option / etc. is shared it has always the same link and displays the newest version. Very useful for answers in issues or in documation of plugins. And yes this is currently already possible by using `reference.html#marker` but the most users don't know this and use the link of the specific version to share.

mourner:
@Falke-Design good suggestion! Let's handle changes related to `reference.html` in a different PR though.

Issue #7712, opened 2021-10-29
Consider deprecating old Leaflet versions docs, and automate publishing


mourner:
Currently dealing with Leaflet docs causes quite a bit of confusion:

- Since `reference.html` is built locally and committed to the repo, we often get PRs on the generated files rather than sources for docs fixes.
- Updating docs for a release is a [quite involved](https://github.com/Leaflet/Leaflet/blob/master/RELEASE.md) manual process — ideally this would be completely automated.
- Old version docs might show up in Google search results instead of latest docs.
- Keeping generated files in the repo (and we have quite a few, including old versions) is a bad practice.

We introduced links to old Leaflet docs back in 2012, when Leaflet was v0.3. Now that it's way more stable and mature and breaking changes are extremely rare, I think we can drop the old versions and only have one reference docs reflecting the latest version.

Also, now that we're using GitHub Actions, automating the docs publishing when tagging a release seems quite doable. cc @jonkoops @IvanSanchez

Falke-Design:
So you want to remove the old docs from the website too? https://leafletjs.com/reference-versions.html
For me it was / is a good way to determine differences between versions and help others if they have problems with plugins and are not on the newest Version

mourner:
@Falke-Design yes. Can you expand on how exactly you'd determine differences this way? The reference pages look identical, with the same sections, and differences so tiny you'd have to know exactly where to look to spot them. IMO the only reliable way to determine differences between versions is the Changelog, and I see no purpose in hosting many nearly-identical docs pages at this stage of the project.

Falke-Design:
I don't know what I was searching but a user of my library was having a issue and then we found out that the function that I used was not existing in 1.6 which he used because of the IOS double tab problem. But yes I searched exactly for the function in the old reference

mourner:
@Falke-Design you could have found it by searching the Changelog as well (and quickly finding the first version where it was added). I don't say that older Leaflet reference docs can't be useful, just that their usefulness doesn't justify the issues I listed above.

jonkoops:
Yeah, I agree that this should be automated, let's consider the publishing of the package separate from the documentation for now so we can chunk up the work more easily. We could use the [release](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release) event and hook into Github releases to publish things to NPM at the same time.

mourner:
Yep, either a `release` event or a git tag that fits semver.

jonkoops:
[...]

PR #7730, opened 2021-11-02
Simplify docs update process on release

mourner:
This PR simplifies the process of updating docs when releasing a new version to minimize the potential for mistakes, and get Leaflet closer to being able to do this automatically in GitHub Actions. Ref #7712 

- Remove redirect from `reference.html` to `reference-1.7.1.html`. Instead, `reference.html` contains docs for the latest stable release directly. Updates links accordingly.
- Make `npm run docs` update `dist/reference.html` directly instead of generating a temporary file that you have to manually move / edit. I guess that we might have had an intention to have self-contained offline docs in the `dist` some time in the past but it never actually happened it seems, and it's doesn't feel necessary in the modern age.
- Update `RELEASE.md`, removing all unnecessary steps.


Falke-Design (approved):
Looks good to me. But I don't know about it that well.

What if we create directly the reference-1.7.1.html and the reference.html? Then this step can never forget

Falke-Design:
@mourner all current links are broke, because now it is `reference.html` and not https://leafletjs.com/reference-1.7.1.html anymore

mourner:
@Falke-Design ah, we should bring it back but as a redirect this time so that old links aren't broken, right?

Falke-Design:
Yes
My general suggestion would be that always the version reference (`reference-1.7.1.html`) is existing and the `reference.html`. If the newest version is `1.7.1` then `reference-1.7.1.html` redirects to `reference.html`. If then a new release is published  `reference-1.8.0.html` redirects to `reference.html` and `reference-1.7.1.html` is still existing. (Until all old release docs are removed)

mourner:
Why would we do it for newer releases if there are no links to it yet? I think we can just limit this to 1.7.1

Falke-Design:
Currently we provide the old docs too, until they are removed we should add the newest version to this list and keep it until all old docs are removed.

Falke-Design:
@mourner FYI: https://leafletjs.com/reference-1.7.1.html is still missing

PR #7733, opened 2021-11-03
add type definition (WIP)

mathe42:
* used types of https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/leaflet (last change is  3940eb2 )
* extracted types for the Browser.js file into a seprate file.
* extrated types for the Class.js file into seperate file and made it better so the `Class.extend` syntax gets proper types.

* see https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-1576931 for discussion.

For how much better the IDE supports get you can check it out here: https://vscode.dev/github/mathe42/Leaflet/tree/patch-2 

For example go to the map.js file and add 

```js
var exampleMap = new Map()
```
and type then `exampleMap.` this will give you intellisense for Map and Evented (what Map extends).

If this is something that will be merged (please say so) I will continue working on it. That would include
1. Continue splitting the file into smaller files
2. Optimize Class.d.ts with proper `this` context when useing `Class.extend` etc. for even better IDE support so that the user of Leaflet (and) Plugins get some Intellisense even if the Plugins don't provide it's own type definitions.
3. Optimize Evented with EventMap-Interface like in https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-1577078 
4. If this gets merged I will try to keep the types up to date with the master branch.


Falke-Design:
I had currently only working with the `.d.ts` files for a plugin and never with TypeScript self. Can you explain me why it is useful to split it in smaller files? 

I like this PR specially point 3.

mathe42:
Basicly there are 2 things:
1. Better maintainable if something changes we have the corresponding file where it has to be changed directly.
2. We get better IDE support while developing leaflet. (Try it out! it makes a lot more fun!)

We could add a build step where all .d.ts files are combined into a single file. (If this is wanted.)

But having more files will have the result that PR for new features by developers who know a little bit TS can change the PR more easy - doing that in a 2000 Line file is not so easy.

mathe42:
I have done 3. (I changed only Evented so Classes extending Evented lost some type information)

Example:
```ts
import { Evented, LeafletEvent } from "./Events";

interface X extends LeafletEvent {
  myProp: number
}

const e = new Evented<{
  'test': X
}>()

e.on('test', (ev) => {
  ev.myProp
})

e.on('jfdsksdfsldf', (e)=>{
  
})
```

Falke-Design:
> I changed only Evented so Classes extending Evented lost some type information

Which informations get lost?

src/core/Events.d.ts
```diff
@@ -0,0 +1,177 @@
...
+  [name: string]: LeafletEvent;
+};
+
+export interface LeafletEvent {
+  type: string;
+  target: any;
+  sourceTarget: any;
+  propagatedFrom: any;
+  /**
+   * @deprecated The same as {@link LeafletEvent.propagatedFrom propagatedFrom}.
```
Falke-Design, line 13: I personally prefer a clear describing action: `@deprecated Use insteaed {@link LeafletEvent.propagatedFrom propagatedFrom}`

src/core/test.ts
```diff
@@ -0,0 +1,17 @@
...
+  'test': X
+}>()
+
+e.on('test', (ev) => {
+  ev.myProp
+})
+
+e.on('jfdsksdfsldf', (e)=>{
+  
+})
```
Falke-Design, line 17: Remove this part or rename `jfdsksdfsldf``.

A real sample would be good to understand it better. Marker, Polygone or so (one where only the layer has the event and not the map)

mathe42:
> > I changed only Evented so Classes extending Evented lost some type information
> 
> Which informations get lost?

What events exists but when I handle Map, Layer,... I will create the EventMaps. 
So "lost some information" meant that the information wich events exists are currently not in the types anymore but will be added back later.

src/core/Events.d.ts
mathe42: Just copied it. I agree with you will change that.

mathe42:
Added Map as an example.

mathe42:
@Falke-Design Finished the "easy" parts. Will finish it later - but waiting for a leaflet-Team statement before finishing this. I'm happy to invest some time but if this has no chance getting merged I would spend that time on some other projects.

Falke-Design:
@mourner decission: will be TypeScript definition file merged?

mourner:
This is extremely valuable work, but please give the maintainers time to consider and think it through very carefully, because it will have major implications on maintainer burden and codebase complexity. Let's pause and evaluate before continuing.

This is why I wanted to focus on smaller and more immediate things first, like getting v1.8.0 with bugfixes out, modernizing tests & CI, streamlining release process, etc. Pressure to do major decisions urgently and rushing through PRs with thousands of lines is an easy way to maintainer burnout.

jonkoops:
Although I am huge fan of this proposal I have to agree with @mourner that we should take this nice and easy. Once we clean up the huge backlog of PRs and issues we can start focusing ion improving this.

Falke-Design:
@johnd0e @mathe42 please change this to a draft

PR #7029, opened 2020-03-16
Correct Browser.touch (and more related changes)

johnd0e:
https://github.com/Leaflet/Leaflet/blob/e213469a296ea00e6f9e6e6beb49d081e8d1c4fe/src/core/Browser.js#L98-L104

Leaflet's own code does not correspond to documentation: instead of detecting **touch** capabilities we detect also **pointer** capabilities. It's always bad when any property value deviates from it's 'physical' meaning, as eventually we get code that nobody can understand, and it's hard to fix anything without breaking something else.

So I propose:
1. Explicitly document **actual** meaning of this flag. 
Alternative: retire it completely (but not in this PR, related discussion: #6978)

2. Consider changes in modern browsers:
   - `window.TouchEvent` should trigger `Browser.touch` (as per recommendation https://www.chromestatus.com/feature/4764225348042752)
     E.g. desktop (with no touch screen) may be connected with aux touch panel (even in runtime).
   - Drop support for [legacy touch](https://developer.mozilla.org/en-US/docs/Web/API/DocumentTouch) Firefox versions (<25), because of their insignificant market share.

3. Stop simulating touch via `DomEvent.Pointer.js` when native touch events are available.
   Close #7077, fix #5745, fix #5969, fix #6632.
   May also fix* #6401 (not confirmed yet).

---

[test (tap:false)](https://gist.githack.com/johnd0e/3ad393c73b12b24af7e947711a87b7a4/raw/test.html)

[test (tap:true)](https://gist.githack.com/johnd0e/3ad393c73b12b24af7e947711a87b7a4/raw/test_tap_true.html)

PR #7736, opened 2021-11-04
Tests: remove non-functional Firefox setting

### LINKED ISSUE #7728: Failed to disable pointer events in recent Firefox versions
After some of last changes in dependencies this setup is not working anymore:
https://github.com/Leaflet/Leaflet/blob/e3fde981e25dda3f90ab33de433a05be40f406a5/spec/karma.conf.js#L105-L125

Look here: 
https://github.com/Leaflet/Leaflet/blob/e3fde981e25dda3f90ab33de433a05be40f406a5/spec/suites/SpecHelper.js#L69-L70

E.g. if we run `npm test -- --browsers FirefoxTouch` - we should get:
```
L.Browser.pointer, false
L.Browser.touch, true
```

But we see `true` for both cases instead.

johnd0e:
`dom.w3c_pointer_events.enabled` does not exist in recent Firefox
Anyway, it is not really useful after 0ae68f4658cb1cd85e6e0c67a24b12e5e6282e22

Fix #7728
Ref: #7029

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,6 @@ jobs:
         browser:
           [
             Chrome1280x1024,
-            FirefoxPointer,
             FirefoxTouch,
             FirefoxPointerTouch,
           ]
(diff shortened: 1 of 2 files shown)


mourner (approved):

jonkoops:
@mourner Looks like `FirefoxTouch` and `FirefoxPointerTouch` now both have the same `prefs` associated with them. Perhaps one of the launchers can be omitted?

johnd0e:
My fault. Will prepare new PR

Issue #7414, opened 2021-01-19
Tests failing on IE


johnd0e:
Tests failing on IE11:
`npm run test-nolint -- --browsers IE`
20 tests failed.
<details>

```
IE 11.0 (Windows 10) ImageOverlay _image when loaded should raise the load event FAILED
        TypeError: Argument not optional
           at raiseImageEvent (spec/suites/layer/ImageOverlaySpec.js:66:4)
           at Anonymous function (spec/suites/layer/ImageOverlaySpec.js:74:5)
IE 11.0 (Windows 10) ImageOverlay _image when load fails should raise the error event FAILED
        TypeError: Argument not optional
           at raiseImageEvent (spec/suites/layer/ImageOverlaySpec.js:66:4)
           at Anonymous function (spec/suites/layer/ImageOverlaySpec.js:83:5)
IE 11.0 (Windows 10) ImageOverlay _image when load fails should change the image to errorOverlayUrl FAILED
        TypeError: Argument not optional
           at raiseImageEvent (spec/suites/layer/ImageOverlaySpec.js:66:4)
           at Anonymous function (spec/suites/layer/ImageOverlaySpec.js:88:5)
IE 11.0 (Windows 10) ImageOverlay #setZIndex should update the z-index of the image if it has allready been added to the map FAILED
        Error: expected 1 to equal '1'
           at Assertion.prototype.assert (node_modules/expect.js/index.js:102:7)
           at Assertion.prototype.equal (node_modules/expect.js/index.js:216:5)
           at Anonymous function (node_modules/expect.js/index.js:69:13)
           at Anonymous function (spec/suites/layer/ImageOverlaySpec.js:111:4)
IE 11.0 (Windows 10) ImageOverlay #setZIndex should set the z-index of the image when it is added to the map FAILED
        Error: expected 10 to equal '10'
           at Assertion.prototype.assert (node_modules/expect.js/index.js:102:7)
           at Assertion.prototype.equal (node_modules/expect.js/index.js:216:5)
           at Anonymous function (node_modules/expect.js/index.js:69:13)
           at Anonymous function (spec/suites/layer/ImageOverlaySpec.js:121:4)
IE 11.0 (Windows 10) ImageOverlay #setZIndex should use the z-index specified in options FAILED
        Error: expected 20 to equal '20'
           at Assertion.prototype.assert (node_modules/expect.js/index.js:102:7)
           at Assertion.prototype.equal (node_modules/expect.js/index.js:216:5)
[...]

PR #7741, opened 2021-11-04
Run tests on Internet Explorer 11

### LINKED ISSUE #7734: Do we really need latest sinon?
After https://github.com/Leaflet/Leaflet/pull/7725#issuecomment-957769787 we lost the ability to run test suite in IE8-10.
Is it OK?

If it is - then those part of config should be removed.
If it is not - then `sinon` should be downgraded.

Opinions?

jonkoops:
Adds the ability to run the tests on different operating systems and enables test runs for Internet Explorer 11. This PR also downgrades Sinon to a compatible version with Internet Explorer and adds an ignore rule so Dependabot will not offer upgrades for this specific dependency (closes #7734).

The tests for Internet Explorer seem to fail, so we will have to open up another PR to fix these issues. @johnd0e perhaps you can provide some insights as to why these tests seem to be failing? It seems to be related to the `prosthetic-hand` dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prosthetic-hand": "^1.3.1",
     "rollup": "^2.59.0",
     "rollup-plugin-git-version": "^0.3.1",
-    "sinon": "^12.0.0",
+    "sinon": "^7.5.0",
     "ssri": "^8.0.1",
     "uglify-js": "^3.14.2"
   },
@@ -95,7 +95,9 @@
     },
     "overrides": [
       {
-        "files": ["build/**/*"],
+        "files": [
+          "build/**/*"
+        ],
         "env": {
           "node": true
         },
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,5 @@ updates:
     ignore:
       - dependency-name: eslint-config-mourner
         update-types: ["version-update:semver-major"]
+      - dependency-name: sinon
+        update-types: ["version-update:semver-major"]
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -95,7 +95,9 @@
     },
     "overrides": [
       {
-        "files": ["build/**/*"],
+        "files": [
+          "build/**/*"
+        ],
```
jonkoops, line 100: NPM keeps formatting the file like this when the dependencies change, so I included it here as well.

johnd0e:
> @johnd0e perhaps you can provide some insights as to why these tests seem to be failing?

Sure, will check.
(perhaps the tests requires some minor changes after #7029)

johnd0e:
Ref: #7414

- 2 tests ("in CSS scaled container") fail because Leaflet really fail them and should be fixed (should it?)
- 5 tests ("Object doesn't support this action") require `"prosthetic-hand": "^1.4.0"` (which is not published on npm)

@morner?

mourner:
@IvanSanchez I don't have permissions to publish to NPM — can you do `npm publish` in `prosthetic-hand`, and then add more owners?

mourner:
OK published prosthetic-hand v1.4.0 now, let's try that out!

jonkoops:
Bumped the version in #7745 and rebased the PR.

mourner:
@johnd0e looks like most failures are related to recent changes to `on` / `off` / etc., although not sure why only IE11 fails there.

johnd0e:
That is because `false._leaflet_id=1` is ok in IE, but expected to fail in tests

johnd0e (approved):

PR #7743, opened 2021-11-04
Fix indentation issues in Karma config

jonkoops:
Removes mixed spaces used for indentation in the Karma config file.

### DIFF
--- a/spec/karma.conf.js
+++ b/spec/karma.conf.js
@@ -104,15 +104,15 @@ module.exports = function (config) {
 			},
 			'FirefoxTouch': {
 				base: 'FirefoxHeadless',
-			        prefs: {
+				prefs: {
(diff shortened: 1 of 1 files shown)


Falke-Design:
Can this checked over eslint?

Falke-Design (approved):

jonkoops:
> Can this checked over eslint?

Yes, we can lint the entire `spec` directory, but that will throw errors due to usage of Node APIs such as `process`.

johnd0e (approved):

Falke-Design:
> to usage of Node APIs such as process.

Would it possible to add `// eslint-disable-next-lin` ?

mourner:
You can do a file-wide specific rules override in a eslint comment at the top of `karma.conf.js`. Let's follow-up in a different PR.

PR #7751, opened 2021-11-07
Better disableClickPropagation

### LINKED ISSUE #7750: disableClickPropagation should not prevent events on element itself (and child elements)
By definition [`disableClickPropagation`](https://github.com/Leaflet/Leaflet/blob/257ca098f1edfe091ec029bcead25f7f9ce2b676/src/dom/DomEvent.js#L196-L203) should stop propagation to parent elements.
But currently if we apply `disableClickPropagation` to some elements - we miss events of any feature bound to this element.
I suppose that is non-intuitive and limiting.

It is possible to implement more intuitive behavior - only child elements' events should be ignored.
I propose to place related code in [`_findEventTargets`](https://github.com/Leaflet/Leaflet/blob/257ca098f1edfe091ec029bcead25f7f9ce2b676/src/map/Map.js#L1350) (instead of [`_isClickDisabled`](https://github.com/Leaflet/Leaflet/blob/257ca098f1edfe091ec029bcead25f7f9ce2b676/src/map/Map.js#L1371))

**Additional context**
Proof of concept - [interactive Popup](https://github.com/Leaflet/Leaflet/pull/7532/commits/817a43ff7cdca1afd560dbceec311638906c1667) (for click only).

johnd0e:
Now `disableClickPropagation` does not prevent events on elements' feature

Fix #7750

### DIFF
--- a/spec/suites/dom/DomEventSpec.js
+++ b/spec/suites/dom/DomEventSpec.js
@@ -301,7 +301,7 @@ describe('DomEvent', function () {
 	});
 
 	describe('#disableClickPropagation', function () {
-		it('stops click events from propagation to parent elements', function () { // except 'click'
+		it('does not stop click events from propagation to parent HTML elements', function () {
 			var child = document.createElement('div');
 			el.appendChild(child);
 			L.DomEvent.disableClickPropagation(child);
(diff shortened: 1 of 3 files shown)


johnd0e:
Alternative solution in last commit (**actually I'm not sure that it is better**).

- Leave DomEvent.disableClickPropagation as is
  (but w/o of legacy click tricks)
- Introduce special method Evented.disableClickPropagation

Note: Control.* should use DomEvent.disableClickPropagation

PR #7755, opened 2021-11-08
No need in `preclick` event

johnd0e:


mourner:
Does it preserve the current `closeOnClick` behavior? As far as I remember, the purpose of `preclick` was to make sure that all popups are closed before any `click` listeners fire (possibly triggering opening of another popup), otherwise we can get inconsistent behavior where in some cases a popup is opened on click and then immediately closed.

johnd0e:
> Does it preserve the current closeOnClick behavior?

Yes it does, checked manually.
But I will make sure that relevant unit tests are present too.

johnd0e:
> make sure that all popups are closed before any click listeners fire (possibly triggering opening of another popup), otherwise we can get inconsistent behavior where in some cases a popup is opened on click and then immediately closed.

Could you provide any sample?

johnd0e:
> Does it preserve the current `closeOnClick` behavior?

We have unit tests for that. And just added one more.

> we recently introduced `preclick` event in GL JS — it shows an equivalent test case: [mapbox/mapbox-gl-js#10926](https://github.com/mapbox/mapbox-gl-js/pull/10926)

Leaflet atm does not suffer from this bug, and has all required unit tests to prevent regression in future: [Popup](https://github.com/Leaflet/Leaflet/blob/315e4d7bde7a4979485320b643a1990ac29bf3be/spec/suites/layer/PopupSpec.js#L21-L55), [Tooltip](https://github.com/Leaflet/Leaflet/blob/315e4d7bde7a4979485320b643a1990ac29bf3be/spec/suites/layer/TooltipSpec.js#L274-L282).

BTW, `preclick` event itself is still [available](https://github.com/Leaflet/Leaflet/search?q=%22preclick%22&type=code), just not used for Popup/Tooltip.

PR #7765, opened 2021-11-10
LatLng: stricter arguments validation

johnd0e:
Throw on any invalid value of lat/lng/alt
(alt is validated if is explicitly specified)

Ref: #7128

---
Question:
- [x] Should we throw instead of returning `null`?
- [ ] Should we be same strict with `alt`, or better allow `undefined` value for compatibility purposes?

### DIFF
--- a/spec/suites/geo/LatLngSpec.js
+++ b/spec/suites/geo/LatLngSpec.js
@@ -10,13 +10,25 @@ describe('LatLng', function () {
 			expect(b.lng).to.eql(-74);
 		});
 
-		it('throws an error if invalid lat or lng', function () {
-			expect(function () {
-				L.latLng(NaN, NaN);
-			}).to.throwError();
+		it('throws an error if invalid lat, lng, alt', function () {
+			expect(L.latLng).withArgs(0, 0, undefined).to.throwError();
+			expect(L.latLng).withArgs(0).to.throwError();
+			expect(L.latLng).withArgs(NaN, 0).to.throwError();
+			expect(L.latLng).withArgs(Infinity, 0).to.throwError();
+			expect(L.latLng).withArgs(-Infinity, 0).to.throwError();
+			expect(L.latLng).withArgs(true, 0).to.throwError();
+			expect(L.latLng).withArgs(false, 0).to.throwError();
+			expect(L.latLng).withArgs(undefined, 0).to.throwError();
+			expect(L.latLng).withArgs(null, 0).to.throwError();
+			expect(L.latLng).withArgs('', 0).to.throwError();
+			expect(L.latLng).withArgs('\n', 0).to.throwError();
+			expect(L.latLng).withArgs('1px', 0).to.throwError();
+			expect(L.latLng).withArgs([]).to.throwError();
+			expect(L.latLng).withArgs([50]).to.throwError();
+			// expect(L.latLng).withArgs('1e5', 0).to.throwError();
 		});
 
-		it('does not set altitude if undefined', function () {
+		it('does not set altitude if not specified', function () {
 			var a = new L.LatLng(25, 74);
 			expect(typeof a.alt).to.eql('undefined');
 		});
@@ -80,15 +92,8 @@ describe('LatLng', function () {
 		});
 
 		it('accepts an array of coordinates', function () {
-			expect(L.latLng([])).to.eql(null);
-			expect(L.latLng([50])).to.eql(null);
 			expect(L.latLng([50, 30])).to.eql(new L.LatLng(50, 30));
 			expect(L.latLng([50, 30, 100])).to.eql(new L.LatLng(50, 30, 100));
-		});
-
-		it('passes null or undefined as is', function () {
-			expect(L.latLng(undefined)).to.eql(undefined);
-			expect(L.latLng(null)).to.eql(null);
 		});
 
 		it('creates a LatLng object from two coordinates', function () {
@@ -101,10 +106,6 @@ describe('LatLng', function () {
 
 		it('accepts an object with lat/lon', function () {
 			expect(L.latLng({lat: 50, lon: 30})).to.eql(new L.LatLng(50, 30));
-		});
-
-		it('returns null if lng not specified', function () {
-			expect(L.latLng(50)).to.be(null);
 		});
 
 		it('accepts altitude as third parameter', function () {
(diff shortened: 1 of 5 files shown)


src/geo/LatLng.js
```diff
@@ -122,16 +133,17 @@ export function toLatLng(a, b, c) {
...
+		var lat = a.lat,
+		    lng = 'lng' in a ? a.lng : a.lon;
+		if ('alt' in a) {
+			return new LatLng(lat, lng, a.alt);
+		}
+		return new LatLng(a.lat, 'lng' in a ? a.lng : a.lon);
 	}
-	if (b === undefined) {
-		return null;
+	if (arguments.length === 3) {
```
Falke-Design, line 145: With the check `arguments.length == 3` it is a breaking change which can affect many projects. I think it is a valid change but we need to keep this in mind

src/layer/GeoJSON.js
```diff
@@ -232,7 +232,10 @@ function _pointToLayer(pointToLayerFn, geojson, latlng, options) {
 // Creates a `LatLng` object from an array of 2 numbers (longitude, latitude)
 // or 3 numbers (longitude, latitude, altitude) used in GeoJSON for points.
 export function coordsToLatLng(coords) {
-	return new LatLng(coords[1], coords[0], coords[2]);
+	if (coords.length === 3) {
+		return new LatLng(coords[1], coords[0], coords[2]);
+	}
+	return new LatLng(coords[1], coords[0]);
```
Falke-Design, line 238: If `coordsToLatLng` is overwritten like in many SO answers suggested, the check `coords.length === 3` will not be done.

For example a simple switch of the array if source is already latlng:
```
coordsToLatLng(coords) {
	return new LatLng(coords[0], coords[1], coords[2]);
```

src/geo/LatLng.js
```diff
@@ -28,23 +28,34 @@ import {toLatLngBounds} from './LatLngBounds';
...
 	// @property alt: Number
 	// Altitude in meters (optional)
-	if (alt !== undefined) {
-		this.alt = +alt;
+	if (arguments.length === 3) {
+		this.alt = checkNumber(alt);
+	}
+}
+
+function checkNumber(a) {
```
Falke-Design, line 46: I would suggest to move this into Utils, maybe we will have some other places where we want to use this. (#7432)

Falke-Design (commented):
I like this PR 👍 
Maybe you want to generalize this PR and add checks for L.Point to #7432

src/geo/LatLng.js
johnd0e: I do not agree, because the function implementation is heavily defined by it's particular application.
E.g. we may need integer numbers, or allow Infinity, etc.

johnd0e:
> Maybe you want to generalize this PR and add checks for L.Point to #7432

In separate PR.

src/geo/LatLng.js
Falke-Design: Hmm you are right but I still think it would be cleaner to have on general function that can used outside of Leaflet core too. Maybe a second optional parameter `allowedTypes` can be passed as array `['integer', 'double', 'Infinity']`.

For infinity it would be also possible to use `if(value === Infinity || checkNumber(value))`

johnd0e: > Maybe a second optional parameter `allowedTypes` can be passed as array

Too many parameters to take into account all possible cases. E.g. should it accept hex strings? Exponential form?
BTW, the function itself is not perfect too. I'd like to improve it, but in reasonable limits. May be you have ideas.

spec/suites/geo/LatLngSpec.js
```diff
@@ -2,21 +2,51 @@ describe('LatLng', function () {
...
+			expect(a.lat).to.be(25);
+			expect(a.lng).to.be(74);
+
+			var b = new L.LatLng("-25", "-74 ");
+			expect(b.lat).to.be(-25);
+			expect(b.lng).to.be(-74);
+
+			var c = new L.LatLng("010", "0x10"); // better'd throw here..
+			expect(c.lat).to.be(10);
+			expect(c.lng).to.be(0); // bad!
```
johnd0e, line 24: Here are the cases that I do not like

spec/suites/geo/LatLngSpec.js
```diff
@@ -2,21 +2,51 @@ describe('LatLng', function () {
...
+			expect(L.latLng).withArgs('', 0).to.throwError();
+			expect(L.latLng).withArgs('\n', 0).to.throwError();
+			expect(L.latLng).withArgs('1px', 0).to.throwError();
+			expect(L.latLng).withArgs('nonsense', 0).to.throwError();
+			expect(L.latLng).withArgs([]).to.throwError();
+			expect(L.latLng).withArgs([50]).to.throwError();
+			// do not throw atm:
+			// expect(L.latLng).withArgs('0010', 0).to.throwError();
+			// expect(L.latLng).withArgs('0x10', 0).to.throwError();
+			// expect(L.latLng).withArgs('1e5', 0).to.throwError();
```
johnd0e, line 46: And here.

Falke-Design: `1e5` = `100000` is a correct number, what is not vaild in my eyes is that the string `'1e5'` is converted to `1` (`parseInt('1e5')`)

src/geo/LatLng.js
Falke-Design: But we can define / document which cases / types are allowed. 

A way to get `0x10` managed is to check if the string is starting with `0`, length > 1 and the second char is not `.` for decimal numbers. With that `0010` is invalid too.

johnd0e: Too much actions. Can be performance penalties

spec/suites/geo/LatLngSpec.js
johnd0e: > is a correct number

Sure, but do we really expect such number in Leaflet?

johnd0e:
Found on stackoverflow: https://run.plnkr.co/plunks/93FPpacuIcXqqKMecLdk/

Falke-Design:
Infinity is now not valid anymore. And I would say if someone uses `0x10` or `1e5` then they are converted to theier decimal number and this is ok

spec/suites/geo/LatLngSpec.js
Falke-Design: Maybe in CRSSimple but I don't think so. I think we should only check if it is a valid number and not if the number makes sense

PR #7766, opened 2021-11-10
Summing the offset of chained movements

Falke-Design:
If multiple movements (`panBy`, `setView()`) are chained, the offset is summed.

```
map.setView([40.722036, -73.998599], map.zoom, {
						animate: true,
						duration: 8
					}).panBy([-150, 0], {animate: true, duration: 0.5}).panBy([0, 50], {animate: true, duration: 0.5});
```

With Timeout and "unchained"-chain call:

```
setTimeout(function () {
	map.setView([40.722036, -73.998599], map.zoom, {
		animate: true,
		duration: 8
	})
}, 1000);
setTimeout(function () {
	map.panBy([-150, 0], {animate: true, duration: 0.5}).panBy([0, 50], {animate: true, duration: 0.5});
}, 2000);
```

anim_chain

Fix: #2609

-------------

@johnd0e I don't get it work with `clock` to sleep x seconds, can you please help me to get it work?

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -144,6 +144,15 @@ describe("Map", function () {
 	});
 
 	describe("#setView", function () {
+		var clock;
+
+		beforeEach(function () {
+			clock = sinon.useFakeTimers();
+		});
+
+		afterEach(function () {
+			clock.restore();
+		});
 		it("sets the view of the map", function () {
 			expect(map.setView([51.505, -0.09], 13)).to.be(map);
 			expect(map.getZoom()).to.be(13);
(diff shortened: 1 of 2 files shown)


johnd0e:
> I don't get it work with `clock` to sleep x seconds, can you please help me to get it work?

Are you sure that your issue is with `sinon`? I do not see major issues there.
Except you better limit `useFakeTimers`/`restore` to single test (but it does not related to failed test).

Have you tried `setTimeout` instead?

Falke-Design:
> single test

I tried this first but was not working too, so I tried to use it with `before` / `after`.

I will try it with setTimeout and done() 👍

PR #7767, opened 2021-11-11
Point: arguments validation

johnd0e:
Throw on any invalid value of x/y.

Fix #6072, fix #6075, fix #7372.
Supersedes: close #7418, close #7432.

Issue #7761, opened 2021-11-09
isNaN usage not covered by tests


johnd0e:
There are some (~5) occurrences of `isNaN` usage in code, and no single test to understand their real purpose.
And that function is known to be not so obvious: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN
It's not straightforward to make sure that value is number, because boolean and empty string also coercing to numbers.

So I am not sure that our current usage is correct.
And more - instead of coercing numeric strings - I'd prefer throwing errors

Thus, we need to refactor that parts and cover them by tests.

johnd0e:
Ref: https://github.com/Leaflet/Leaflet/issues/7128#issuecomment-767523311

johnd0e:
1. `isNaN` itself is not reliable way to check if value can be converted to number. There a lot of discussions, and I just point to thes unit tests: https://run.plnkr.co/plunks/93FPpacuIcXqqKMecLdk/
So I propose to implement `isNumeric` like here: #7765
2. Just making sure that the value is convertable tu number is not enough. Instead we should actually convert the value to number (otherwise `'5'+5 = '55'`).
There are a lot of places in code vulnerable to this sort of bugs, and a number of open issues (too lazy to link them now).

itsyourOOP:
@johnd0e 

Is the issue still open? Because if it is, I'd like to try if that's fine.

Falke-Design:
@itsyourOOP I think this not a good issue to start, it is currently not clearly how we want to further with the number check in Leaflet.

A easy issue to start would be: #7787

PR #7768, opened 2021-11-11
New private function Util._checkNumber instead of checking over isNaN

johnd0e:
1. `isNaN` itself is not reliable way to check if value can be converted to number. There a lot of discussions, and I just point to these unit tests: https://run.plnkr.co/plunks/93FPpacuIcXqqKMecLdk/

2. Just making sure that the value is convertible to number is not enough. Instead we should actually convert the value to number as early as possible (otherwise `'5'+5 = '55'`).
There are some places in code vulnerable to this sort of bugs, and a number of open issues (too lazy to link them right now).

Ref: #7761, #7128.

---
Questions:
- [x] Should we have common method instead of copy-paste (#7765, #7767, this, and potentially a lot of other cases)?
- [ ] Better name (`validateNumber`)?
- [ ] Should it be exposed as public method?
- [x] Should it be applied widely (e.g. in all constructors taking number options)?

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -239,3 +239,19 @@ export function cancelAnimFrame(id) {
 		cancelFn.call(window, id);
 	}
 }
+
+function isNumeric(str) { // https://stackoverflow.com/a/175787/2520247
+	return !isNaN(str) && !isNaN(parseFloat(str));
+}
+
+// keep it private intentionally, because such function implementation
+// heavily depends on it's particular application and unable to suit all the cases
+export function _checkNumber(a) {
+	if (typeof a === 'string' && isNumeric(a)) {
+		a = +a;
+	}
+	if ((typeof a === 'number' || a instanceof Number) && isFinite(a)) {
+		return a;
+	}
+	throw new Error('Number expected');
+}
--- a/src/layer/vector/Circle.js
+++ b/src/layer/vector/Circle.js
@@ -31,8 +31,7 @@ export var Circle = CircleMarker.extend({
 		}
 		Util.setOptions(this, options);
 		this._latlng = toLatLng(latlng);
-
-		if (isNaN(this.options.radius)) { throw new Error('Circle radius cannot be NaN'); }
+		this.options.radius = Util._checkNumber(this.options.radius);
 
 		// @section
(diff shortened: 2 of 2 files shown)


IvanSanchez (approved):
LGTM

IvanSanchez:
> Should we have common method instead of copy-paste (LatLng: stricter arguments validation #7765, Point: arguments validation #7767, this, and potentially a lot of other cases)?

IMO, yes. But if it's going to be a common method/function, I'd rather name it *without* the leading underscore, and put some docstrings.

src/core/Util.js
```diff
@@ -239,3 +239,19 @@ export function cancelAnimFrame(id) {
...
 	}
 }
+
+function isNumeric(str) { // https://stackoverflow.com/a/175787/2520247
+	return !isNaN(str) && !isNaN(parseFloat(str));
+}
+
+// keep it private intentionally, because such function implementation
+// heavily depends on it's particular application and unable to suit all the cases
+export function _checkNumber(a) {
```
johnd0e, line 249:
```suggestion
// @function checkNumber(data: Number|String): Number
// Makes sure that passed `data` is finite number, and returns it.
// If string is passed - converts it to number trying to avoid coercion pitfalls.
// Throws error in case of any failure.
export function checkNumber(a) {
```

@IvanSanchez
But I'm not sure, because fixing it in public API
- .. makes impossible future internal changes. E.g. what if later we'll decide to throw on strings like '1e9'
- .. opens the door to misusing (remember using `Browser.touch`)

Falke-Design: @IvanSanchez what do you think about that points? I personally would like to have a public function so it can be easy used in plugins. Not everyone should reinvent the wheel.

johnd0e: @mourner ?

Falke-Design: @mourner what do you think is better? A public or private function for checking if the input is a number.

jonkoops:
I'd argue to make it private, if it creates added value for plugins devs we can always make it public in the future. Let it prove itself first.

Falke-Design:
@mourner what do you think?

PR #7776, opened 2021-11-13
Remove leaflet-include.js

Falke-Design:
I think the usage of `leaflet-include.js` is outdated and can be replaced with a direct link.
Or is there some usage that I don't know?


mourner (approved):
I think that's a leftover after switching to Rollup, we can remove it.

PR #7777, opened 2021-11-14
Add DivOverlay to menu

Falke-Design:
DivOverlay is missing in the menu and was placed at the bottom of the docs (below Event objects): https://leafletjs.com/reference.html#divoverlay

### DIFF
--- a/build/leafdoc-templates/html.hbs
+++ b/build/leafdoc-templates/html.hbs
@@ -35,7 +35,7 @@ bodyclass: api-page
 		<h4>UI Layers</h4>
 		<ul>
 			<li><a href="#marker">Marker</a></li>
-            <li><a href="#divoverlay">DivOverlay</a></li>
+			<li><a href="#divoverlay">DivOverlay</a></li>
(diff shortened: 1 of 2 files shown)


johnd0e:
Not sure that we need it in menu.

Falke-Design:
Because of the sentence "[Inherit from it for custom popup like plugins.](https://leafletjs.com/reference.html#divoverlay)" it would be good if it is documented and in the menu visible

build/leafdoc-templates/html.hbs
```diff
@@ -35,6 +35,7 @@ bodyclass: api-page
 		<h4>UI Layers</h4>
 		<ul>
 			<li><a href="#marker">Marker</a></li>
+            <li><a href="#divoverlay">DivOverlay</a></li>
```
mourner, line 38:
```suggestion
			<li><a href="#divoverlay">DivOverlay</a></li>
```

mourner (approved):

Issue #6597, opened 2019-04-14
DivOverlay methods are not documented


johnd0e:
[DivOverlay](https://leafletjs.com/reference-1.4.0.html#divoverlay) manual section documents options, events, but not methods.

- Some of methods are mentioned in [Popup](https://leafletjs.com/reference-1.4.0.html#popup-getlatlng) section
- Some documented wrong: https://github.com/Leaflet/Leaflet/blob/d5201acee0234ff4239cc8cbfa59c687286a2692/src/layer/DivOverlay.js#L104
- [Tooltip](https://leafletjs.com/reference-1.4.0.html#tooltip) section does not mention those methods, but from code it's obvious that L.Tooltip has the same methods

I suppose part of Popup methods documentation should be moved to DivOverlay, and then Popup and Tooltip linked to DivOverlay methods description.

wigsters52:
I would love to work on this. Would that be okay?

PR #7778, opened 2021-11-14
Update the doc of DivOverlay

Falke-Design:
Fix: #6597

- DivOverlay has a default offset: [0,7] I think this would be better if this is moved to Popup and changed to [0,0] but this is a breaking change
- DivOverlay has not factory function


src/layer/DivOverlay.js
```diff
@@ -19,16 +19,16 @@ export var DivOverlay = Layer.extend({
 	// @aka DivOverlay options
 	options: {
 		// @option offset: Point = Point(0, 7)
-		// The offset of the popup position. Useful to control the anchor
+		// The offset of the overlay position. Useful to control the anchor
 		// of the popup when opening it on some overlays.
```
johnd0e, line 23: > when opening it on some overlays

What does it ever mean?

src/layer/DivOverlay.js
```diff
@@ -19,16 +19,16 @@ export var DivOverlay = Layer.extend({
...
 
 		// @option className: String = ''
-		// A custom CSS class name to assign to the popup.
+		// A custom CSS class name to assign to the overlay.
 		className: '',
 
 		// @option pane: String = 'popupPane'
-		// `Map pane` where the popup will be added.
+		// `Map pane` where the overlay will be added.
 		pane: 'popupPane'
```
johnd0e, line 32: Shouldn't we change option default value? May be unset?

src/layer/DivOverlay.js
```diff
@@ -214,6 +215,11 @@ export var DivOverlay = Layer.extend({
 			}
 			node.appendChild(content);
 		}
+
+		// @namespace DivOverlay
+		// @section DivOverlay events
+		// @event contentupdate: Event
+		// Fired when the content of the overlay is updated
 		this.fire('contentupdate');
```
johnd0e, line 223: Interesting..

johnd0e:
> * DivOverlay has a default offset: [0,7] I think this would be better if this is moved to Popup and changed to [0,0] but this is a breaking change

Hm.. Still I'd prefer to make that change.

> * DivOverlay has not factory function

And that is right, because it is abstract class.

src/layer/DivOverlay.js
Falke-Design: I don't know. Maybe we can remove the second sentence

src/layer/DivOverlay.js
Falke-Design: if it is a abstract class (like you said), then I would say yes

johnd0e:
@Falke-Design

johnd0e (approved):

Falke-Design:
To keep it clear before merging: This is a breaking change! The default values for DivOverlay are chaged

mourner (approved):

PR #7779, opened 2021-11-14
Draggable: reset internal property _moved on drag end (instead of start)

### LINKED ISSUE #7775: In Canvas mousedown event not called after dragging map
While useing `preferCanvas: true` and adding `mousedown` listener on a layer, the event is not fired correctly.
Tested on Firefox and Chrome.

Hard to reproduce and a very special usecase:
1. mousedown and move on the layer -> `mousedown` on the layer fired & map is moveing / dragging
2. drag the map outside of the layer
3. mousedown and move on the layer -> `mousedown` on the layer is not fired

```
		var circle = L.circle(map.getCenter(), {radius: 500}).addTo(map);
		circle.on('mousedown',(e)=>{
			console.log('Circle mousedown');
		});
		L.DomEvent.on(map.getContainer(),'mousedown',(e)=>{
			console.log('Map mousedown','');
		})
```

After clicking somewhere on the map, it is working again and possible to fire `mousedown` on the layer.

I found out that this happens because `moved()` is `true`:
https://github.com/Leaflet/Leaflet/blob/a37f0297afd1e547e94d5ae4e217f7cef1d696c8/src/map/handler/Map.Drag.js#L88-L90

The stacktrace:
https://github.com/Leaflet/Leaflet/blob/a37f0297afd1e547e94d5ae4e217f7cef1d696c8/src/layer/vector/Canvas.js#L353
https://github.com/Leaflet/Leaflet/blob/a37f0297afd1e547e94d5ae4e217f7cef1d696c8/src/map/Map.js#L1448-L1450

Demo: https://plnkr.co/edit/RZBfMbG3UERW1FB2

@johnd0e the problem occured since the commit https://github.com/Leaflet/Leaflet/commit/f2d625fcad8b5bf01bd7e98293ea0607c9d4fff7 -> PR #7059

map_mousedown

johnd0e:
Prevents race conditions when canvas processes events earlier than map

Fix #7775

(Same fix was proposed earlier in #7066)

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -79,8 +79,6 @@ export var Draggable = Evented.extend({
 		// under some circumstances, see #3666.
 		if (e._simulated || !this._enabled) { return; }
 
-		this._moved = false;
-
 		if (DomUtil.hasClass(this._element, 'leaflet-zoom-anim')) { return; }
 
 		if (Draggable._dragging || e.shiftKey || ((e.which !== 1) && (e.button !== 1) && !e.touches)) { return; }
@@ -216,6 +214,9 @@ export var Draggable = Evented.extend({
 		}
 
 		this._moving = false;
+		setTimeout(L.Util.bind(function () { // defer resetting in order to be able to reject next click
+			this._moved = false;
+		}, this), 0);
 		Draggable._dragging = false;
 	}
 


Falke-Design (approved):
We are useing this workaround already for a long time in the Drag-End function of Leaflet-Geoman: [Code](https://github.com/geoman-io/leaflet-geoman/blob/master/src/js/Mixins/Dragging.js#L279)

PR #7782, opened 2021-11-14
Better support for RTL text (tooltips)

Falke-Design:
New function `map.setRTLDirection(boolean)`. It adds the CSS-Class `.leaflet-rtl` to the map container. 
During initialization the style `direction: rtl` is searched on the map container and if is found `setRTLDirection(true)` is called. 
Fix: #7201

RTL was not supported for Tooltips:
grafik
grafik

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -106,6 +106,9 @@
 .leaflet-map-pane canvas { z-index: 100; }
 .leaflet-map-pane svg    { z-index: 200; }
 
+
+.leaflet-rtl .leaflet-tooltip-pane   {direction: ltr;}
+
 .leaflet-vml-shape {
 	width: 1px;
 	height: 1px;
@@ -268,6 +271,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 /* general typography */
 .leaflet-container {
 	font: 12px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
+	}
+
+.leaflet-rtl{
+	direction: rtl;
 	}
 
 
@@ -638,4 +645,8 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	left: 0;
 	margin-left: -12px;
 	border-right-color: #fff;
+	}
+
+.leaflet-rtl .leaflet-tooltip-pane > * {
+	direction: rtl;
 	}
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -943,6 +943,17 @@ export var Map = Evented.extend({
 		return this._container;
 	},
 
+	// @method setRTLDirection(rtl: Boolean): this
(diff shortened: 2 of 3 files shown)


johnd0e:
> New function `map.setRTLDirection(boolean)`. It adds the CSS-Class `.leaflet-rtl` to the map container.
> During initialization the style `direction: rtl` is searched on the map container and if is found `setRTLDirection(true)` is called.

Sounds crazy - "search for one style and set another".
IMHO one style should be enough (`.leaflet-rtl`).

And may be `setRTLDirection` is not needed.

Falke-Design:
> "search for one style and set another"

Without this the user needs to set on the map container `.leaflet-rtl` by him self. With my solution it makes "more sense". The user sets on the `html, body` tag in CSS  the `direction: rtl` and leaflet detects if the class `.leaflet-rtl` is needed. 

> And may be setRTLDirection is not needed.

Is needed when the Webseite is designed to switch between languages then the initial set of  `.leaflet-rtl` will not work

johnd0e:
> Without this the user needs to set on the map container `.leaflet-rtl`

Right.

> Is needed when the Webseite is designed to switch between languages

Direct `add/removeClass` is even easier.
Anyway, leaflet lacks any multilanguage support in other places.

spec/karma.conf.js
```diff
@@ -93,7 +93,7 @@ module.exports = function (config) {
 		// - Opera
 		// - Safari (only Mac)
 		// - IE (only Windows)
-		browsers: ['Chrome1280x1024'],
+		browsers: ['IE'],
```
johnd0e, line 96: Really?

dist/leaflet.css
```diff
@@ -273,6 +276,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	font: 0.75rem/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
 	}
 
+.leaflet-rtl{
+	direction: rtl;
+	}
```
Malvoz, line 281: Is this at all needed?

Falke-Design: Yes. If the user decides that leaflet has direction `rtl` but not the parent of the map container, then it would not work. (visible in the attribution control)

src/map/Map.js
```diff
@@ -1108,6 +1119,8 @@ export var Map = Evented.extend({
 			(Browser.safari ? ' leaflet-safari' : '') +
 			(this._fadeAnimated ? ' leaflet-fade-anim' : ''));
 
+		this.setRTLDirection(DomUtil.getStyle(container, 'direction') === 'rtl');
```
Falke-Design, line 1122: @Malvoz what do you think? Should we keep the auto detection or is a option while creating the map instance better? Both is ok for me. `L.map({ direction: 'rtl' })`

Malvoz: Auto detection seems correct to me, it behaves like direction inheritance in HTML.

PR #7783, opened 2021-11-14
Add content and latlng while initializing of Popup / Tooltip

Falke-Design:
DivOverlay (Popup & Tooltip) supports now `content` in the `options`:
```
L.popup({content: 'Test'});
```
instead of
```
L.popup()
    .setContent('Test')
```

-------

And it is now possible to pass an `latlng` point and the `options` as second parameter while creation:
```
var popup = L.popup(latlng, options?).addTo(map);
```
instead of
```
var popup = L.popup()
    .setLatLng(latlng)
```

------
This two new things together makes the usage of standalone popup / tooltip easier:
```
L.popup(latlng, {content: 'Test'}).addTo(map);
```


ziaS26 (approved):

johnd0e:
I see here some unrelated (but useful additions).
Wouldn't it be better to open separate PR for them?

johnd0e:
> this is theier relation

I expect relation to subject, which is defined it PR's title.

Malvoz:
Going through the v1.8.1 milestone PRs; @Falke-Design Please resolve merge conflicts when you have time.

jonkoops:
@Falke-Design do we want to keep this in the 1.9 milestone or can we move this forward? I'd like to start working on a release.

Falke-Design:
@jonkoops it only needs a review from another maintainer like many other PRs. If you could do the review it would be great

mourner (approved):

Issue #5035, opened 2016-10-21
Calling fitBounds with a keepInView: true popup present can cause Leaflet to lock up


grantHarris:
## How to reproduce
- Leaflet version I'm using: 1.0.1
- Browser (with version) I'm using: Chrome Version 53.0.2785.143 (64-bit)
- OS/Platform (with version) I'm using: macOS Sierra
## What behaviour I'm expecting and which behaviour I'm seeing

If you have a popup present on your map with its keepInView option set to true and then call fitBounds on a range outside of the current view, Leaflet locks up. Effectively the two 'fight'. Pan operations called by the leaflet events cause the stack to blow up (Console displays Uncaught RangeError: Maximum call stack size exceeded). In my experience this often totally locks up my browser requiring me to kill the process.

I'm willing to submit a PR for this, but I'm unsure of the correct behaviour in this case. Which one should 'win'? KeepInView or fitBounds?  Some possible approaches:
- Have keepInView win. Make fitBounds do a no-op if popup would be outside bounds.
- Extend fitBounds bounds to include the popup.
- Ignore the keepInView flag in such a situation.
- Throw an error.

Thoughts?
## Minimal example reproducing the issue

http://jsfiddle.net/granth/L4du3L4j/

This example creates a popup with keepInView: true, then triggers a setTimeout after 5 seconds that calls fitBounds with a bounds far away from the popup location.

grantHarris:
Anyone?

omeysalvi:
I have the exact same issue. Any luck?

IvanSanchez:
@grantHarris I haven't looked at this problem, but it seems to me that the saner solution would be to expand the bounds as to include the popup somehow. I'd certainly like to see a PR for this :smile:

rbavery:
I have the same issue as well!

mikaelmeulle:
same issue too

forger:
I have a similar issue with Leaflet freezing up (due to an overflowing stack) when minimizing / reducing the map to a size that doesn't fit an open keepInView popup.

It seems to me that the priority of these features needs to be negotiated. In my (resizing) scenario for example the solution "Have keepInView win" that @grantHarris suggested could be interpreted as "don't allow the map size to be reduced beyond the size of an open keepInView popup" which is not acceptable widget behaviour IMHO.

HamidOsouli:
Same issue here, any found solution yet ?

Quanda:
+1 Same issue

[...]

PR #7792, opened 2021-11-20
Fix Popup keepInView if the map needs to panned over a long distance

Falke-Design:
if `keepInView` is enabled, a listener for `moveend` is set to the map:
https://github.com/Leaflet/Leaflet/blob/3466cbba34351ee0d7477a83eb0215cc0e6b454b/src/layer/Popup.js#L163

`_adjustPan` calls `panBy`:
https://github.com/Leaflet/Leaflet/blob/3466cbba34351ee0d7477a83eb0215cc0e6b454b/src/layer/Popup.js#L276

And then `_resetView` is called because the popup latlng/ offset is not in the map size (and `animate` is not set), in `_resetView` the `moveend` event is fired without panning the map and then `moveend` event ist fired from `keepInView` and it is starting again.
https://github.com/Leaflet/Leaflet/blob/3466cbba34351ee0d7477a83eb0215cc0e6b454b/src/map/Map.js#L322-L325

To prevent this, we can add to the `panBy` call in `_adjustPan` the option `{animate: true}`. This will force that `_resetView` can't be called and we don't end up in a recursive call. This will only happen if the function `_adjustPan` is called over the `moveend` event, so I add `animate: true` if the event object is set.

Maybe this is not the best way but would be a solution.

Fix: #5035


mourner (approved):

Falke-Design:
@mourner Can you please retrigger the tests. It failed while starting the tests

PR #7793, opened 2021-11-20
Fix current error in the maser (SVG width: auto in IE)

Falke-Design:
If in IE `width: auto` is added to a svg layers, it overwrites the predefined `width` from the tag / element self. So I remove `width: auto` from svg layer. Or better I excluded the svg style `.leaflet-overlay-pane svg` from the styles they has `width: auto`.
ie_svg_width

The error occurs because the svg layer (circle) is not anymore there, when the click at the expected position happens.

The error is caused from #6843

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -48,8 +48,7 @@
 .leaflet-container .leaflet-overlay-pane svg{
 	max-width: none !important;
 	max-height: none !important;
-
-}
+  }
 .leaflet-container .leaflet-marker-pane img,
 .leaflet-container .leaflet-shadow-pane img,
 .leaflet-container .leaflet-tile-pane img,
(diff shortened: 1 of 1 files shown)


Falke-Design:
@mourner this PR would fix the failures of the tests in the master

dist/leaflet.css
```diff
@@ -45,7 +45,10 @@
 	}
 /* .leaflet-container svg: reset svg max-width decleration shipped in Joomla! (joomla.org) 3.x */
 /* .leaflet-container img: map is broken in FF if you have max-width: 100% on tiles */
-.leaflet-container .leaflet-overlay-pane svg,
+.leaflet-container .leaflet-overlay-pane svg{
```
mourner, line 48:
```suggestion
.leaflet-container .leaflet-overlay-pane svg {

```

Falke-Design: Fixed, can we add eslint on `leaflet.css` too?

mourner (approved):

dist/leaflet.css
jonkoops: @Falke-Design perhaps you can log a separate issue for that?

Issue #7472, opened 2021-02-15
The × character of popup's close button should be explicitly hidden from ATs


Malvoz:
## Motivation

Similar to https://github.com/Leaflet/Leaflet/issues/7433, even if the popup's close button had an accessible name (https://github.com/Leaflet/Leaflet/issues/7399) screen readers would still announce "times" or "multiplication" (depending on screen reader and/or browser combination) if the user hovers the close button.

## Proposed solution

The × character (`&#215;`) should be wrapped in `<span aria-hidden="true">` to prevent screen readers from announcing it.

https://github.com/Leaflet/Leaflet/blob/b52961aa469dc17b3710ef7a9c0f13b6f4dba550/src/layer/Popup.js#L194

Falke-Design:
@Malvoz after integrating #7399 the code looks like: 
```
closeButton.innerHTML = '<span role="button" aria-label="Close popup">&#215;</span>';
```

and now you suggest to add a second `<span>` or does it work to have `aria-hidden` and `aria-label` in the same element?
```
closeButton.innerHTML = '<span aria-hidden="true" role="button" aria-label="Close popup">&#215;</span>';
```

Malvoz:
@Falke-Design Thanks for looking into accessibility issues, much appreciated! 🙏🏼

> after integrating #7399 the code looks like:
>
> `closeButton.innerHTML = '<span role="button" aria-label="Close popup">&#215;</span>';`

https://github.com/Leaflet/Leaflet/issues/7399 suggests to add `role` and `aria-label` on the `<a>` element. I [commented](https://github.com/Falke-Design/Leaflet/commit/efbe5c42aa47986bce0b6e43fb40f43d0ce1a0e3#r60616941) on the commit in your fork.

> does it work to have `aria-hidden` and `aria-label` in the same element

No, `aria-hidden="true"` hides the content of that element and everything inside from screen readers, that's why we need the `<span>` as described in this issue.

Issue #7399, opened 2021-01-06
Popup's close button should have an accessible name and `role="button"`


Malvoz:
## Motivation

The tables below provide screen capture of screen readers' output with the current markup of close buttons for popups, and also of improved markup.

The "improved" markup for comparison uses `aria-label="Close popup"` and `role="button"` on `leaflet-popup-close-button`.

<table>
  <tr>
    <th width="120"><a href="https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en">ChromeVox</a></th>
    <th>Screen capture</th>
    <th>Screen reader output</th>
  </tr>
  <tr>
    <td>Current</td>
    <td width="300"><video src="https://user-images.githubusercontent.com/26493779/103791768-2349c480-5043-11eb-80a3-3f9b9ae8238e.mp4"></video>
</td>
    <td><q>Multiplication, <i>internal link</i></q></td>
  </tr>
  <tr>
    <td>Improved</td>
    <td width="300"><video src="https://user-images.githubusercontent.com/26493779/103791869-41afc000-5043-11eb-8d91-526cd48cbfc2.mp4"></video></td>
    <td><q>Close popup, <i>button</i></q></td>
  </tr>
</table>

<table>
  <tr>
    <th width="120"><a href="https://www.nvaccess.org/">NVDA</a></th>
    <th>Screen capture</th>
    <th>Screen reader output</th>
  </tr>
  <tr>
    <td>Current</td>
    <td width="300"><video src="https://user-images.githubusercontent.com/26493779/103792111-8b98a600-5043-11eb-8bb9-05280a78fc39.mp4"></video></td>
    <td><q><i>Frame clickable</i>, times, <i>link</i></q></td>
  </tr>
  <tr>
    <td>Improved</td>
    <td width="300"><video src="https://user-images.githubusercontent.com/26493779/103792127-905d5a00-5043-11eb-9800-0e244d043aaf.mp4"></video></td>
    <td><q><i>Clickable</i>, close popup, <i>button</i></q></td>
  </tr>
</table>

## Proposed solution

Set `role="button"` and `aria-label="Close"` (or "Close popup") on the popup's close button:

https://github.com/Leaflet/Leaflet/blob/3466cbba34351ee0d7477a83eb0215cc0e6b454b/src/layer/Popup.js#L192

## Additional context

[...]

PR #7794, opened 2021-11-21
Accessibility popup close button

Falke-Design:
- Change the color of the close button for a better contrast #7539 
- Add `role=button` and the `aria-label='Close popup'` to the `<a>` element #7399
- Add `aria-hidden` to the `times` / `x` character to prevent the reading of screen readers #7472

Close Button old:
grafik
New:
grafik

Fixes: #7539, #7399, #7472

Related PR: #7079
@Malvoz fyi

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -520,13 +520,13 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	width: 18px;
 	height: 14px;
 	font: 16px/14px Tahoma, Verdana, sans-serif;
-	color: #c3c3c3;
+	color: #757575;
 	text-decoration: none;
 	font-weight: bold;
 	background: transparent;
 	}
 .leaflet-container a.leaflet-popup-close-button:hover {
-	color: #999;
+	color: #585858;
 	}
 .leaflet-popup-scrolled {
 	overflow: auto;
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -525,7 +525,6 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	font: 16px/14px Tahoma, Verdana, sans-serif;
 	color: #757575;
 	text-decoration: none;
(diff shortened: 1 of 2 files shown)


Malvoz (approved):
Nice, LGTM!

dist/leaflet.css
```diff
@@ -523,13 +523,13 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	width: 18px;
 	height: 14px;
 	font: 16px/14px Tahoma, Verdana, sans-serif;
-	color: #c3c3c3;
+	color: #757575;
 	text-decoration: none;
 	font-weight: bold;
```
mourner, line 528:
```suggestion
```

mourner (changes requested):
This looks great! I find the close button too jarring when simpling making the color darker, but not if we remove the bold font:

image

mourner (approved):

ahukkanen:
The close button is still not accessible for other languages than English because the label cannot be translated.

Malvoz:
@ahukkanen https://github.com/Leaflet/Leaflet/issues/3210#issuecomment-634066856:

> Regarding translations: [internationalization isn't built into leaflet core](https://github.com/Leaflet/Leaflet/pull/1232) but it is available as a plugin - https://github.com/yohanboniface/Leaflet.i18n

I suppose you could open a separate issue, or ask if the maintainers could reconsider https://github.com/Leaflet/Leaflet/pull/1232, but that's off topic from this PR?

ahukkanen:
@Malvoz Thanks for the context!

After reading that i18n discussion, I'm not too certain I could convince the maintainers otherwise than what was discussed there.

At https://github.com/Leaflet/Leaflet/pull/7079#issuecomment-634080439 I tried to suggest to make the hard coded texts configuration options instead of hard coded. This way it would be possible to change programmatically without the need of creating a separate i18n layer in the core.

Issue #7479, opened 2021-02-24
Improved UX for keyboard and screen reader users through change in DOM order


Malvoz:
The DOM order of elements affects the order in which screen readers announce content, and also the tab order.

I believe that the map controls should be available/presented to the user before other author provided content such as markers and popups. In other words: `.leaflet-control-container` should come before `.leaflet-map-pane` in the DOM order.

This would reduce confusion and allow blind users to form a consistent mental model of the content. It also creates a more consistent and natural tab order, even across different maps (perhaps by different authors) with varying content.

Relates to <span title="Web Content Accessibility Guidelines">WCAG</span> 2.1 Success Criterion [2.4.3 Focus Order](https://www.w3.org/TR/WCAG21/#focus-order) and [3.2.3 Consistent Navigation](https://www.w3.org/TR/WCAG21/#consistent-navigation).

<hr>

@stevevance raised a similar concern in https://github.com/Leaflet/Leaflet/issues/3210 (citing the relevant parts):
> I was testing my website with a person who is completely blind and uses the VoiceOver (VO) function on an iPad. I was observing her as she browsed the entire webpage (through a tapping function in VO that mimics tabbing between elements*) and came across the Leaflet map.

> ## Expectations

> The next tab should land on the zoom controls, layer controls, and any other controls, so the user can understand what functions are available to control the map.

> The next tab should land on any marker that has a popup and the contents of the popup should be read.

Malvoz:
> I believe that the map controls should be available/presented to the user before other author provided content such as markers and popups.

While I think this would be an improvement (especially when there are a lot of markers), the problem with this approach is that it causes a disruptive tab stop between panning and interacting with the markers.

Below is a description of 3 different methods that I think should be considered.

## A. Current scenario

The current scenario in a map with say 100 markers:

1. <kbd>Tab</kbd> to map
2.  <kbd>Tab</kbd> through 100 markers
3. <kbd>Tab</kbd> to controls

In this scenario, it would take more than 100 tab presses to reach the controls, to go into fullscreen for example. It's also harder for users relying on screen readers to detect the available controls.

## B. Changing DOM order (https://github.com/Leaflet/Leaflet/pull/7796)

If the DOM order of the map pane and controls is changed, the scenario would be:

1. <kbd>Tab</kbd> to map
2. <kbd>Tab</kbd> through a few controls
3. <kbd>Tab</kbd> to markers

[...]

PR #7796, opened 2021-11-21
Add first the `control-container` to the DOM and then the  `map-pane` (for tabbing) #7479

Falke-Design:
Add first the `control-container` to the DOM and then the  `map-pane`, to change the order while tabbing. Now first the controls are focused and then all the layers.

@stevevance raised a similar concern in #3210 (quoting the relevant parts):

> I was testing my website with a person who is completely blind and uses the VoiceOver (VO) function on an iPad. I was observing her as she browsed the entire webpage (through a tapping function in VO that mimics tabbing between elements*) and came across the Leaflet map.
> ### Expectations
> The next tab should land on the zoom controls, layer controls, and any other controls, so the user can understand what functions are available to control the map

Fix: #7479

@Malvoz fyi

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1111,11 +1111,11 @@ export var Map = Evented.extend({
 			container.style.position = 'relative';
 		}
 
-		this._initPanes();
-
 		if (this._initControlPos) {
 			this._initControlPos();
 		}
+
+		this._initPanes();
 	},
 
 	_initPanes: function () {


Malvoz:
Thanks @Falke-Design!

This may need some more consideration first, I commented on the issue: https://github.com/Leaflet/Leaflet/issues/7479#issuecomment-974918350.

Malvoz:
I think we can close this for now per https://github.com/Leaflet/Leaflet/issues/7479#issuecomment-975160765?

PR #7803, opened 2021-11-23
CSS improvements (responsive map embedding in examples/tutorials)

### LINKED ISSUE #7342: Mobile Example Outdated
https://leafletjs.com/examples/mobile/

```css
html, body, #map {
    height: 100%;
    width: 100vw;
}
```

> note: In this example we use percentage for height. While vh is arguably better, due to a bug with Google Chrome on mobile.

I guess this can be updated now, as I cannot see any issues with this on chrome.

### LINKED ISSUE #7420: Examples are not mobile-friendly

alvarolozano:
Fix #7342 
Fix #7420 

I described here my changes: https://github.com/Leaflet/Leaflet/issues/7342#issuecomment-976381354

### DIFF
--- a/docs/_includes/frame.html
+++ b/docs/_includes/frame.html
@@ -1,7 +1,7 @@
 <table>
-<tr><td style='text-align: center; border: none'>
+<tr><td style='text-align: center; border: none; padding: 0'>
 <iframe src='{{ include.url }}'
-	width='{% if include.width %}{{ include.width }}{% else %}616{% endif %}'
+	width='{% if include.width %}{{ include.width }}{% else %}100%{% endif %}'
 	height='{% if include.height %}{{ include.height }}{% else %}416{% endif %}'></iframe>
 </td></tr>
 <tr><td style='text-align: center; border: none'>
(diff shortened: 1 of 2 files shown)


docs/_includes/frame.html
```diff
@@ -1,7 +1,7 @@
 <table>
-<tr><td style='text-align: center; border: none'>
+<tr><td style='text-align: center; border: none; padding: 0'>
 <iframe src='{{ include.url }}'
-	width='{% if include.width %}{{ include.width }}{% else %}616{% endif %}'
+	width='{% if include.width %}{{ include.width }}{% else %}100%{% endif %}'
```
Malvoz, line 4: The [HTML validator](https://validator.w3.org/nu/#textarea) returns:

> **Error:** Bad value `100%` for attribute `width` on element `iframe`: Expected a digit but saw `%` instead

I know that this works in Chrome, but does it work in all browsers?

Falke-Design:
supersede by #7950

Issue #7801, opened 2021-11-23
Plugin.md: Merge plugins PRs, use `tabs` and update links


Falke-Design:
I think it would be good if we "update" the plugin page. Means that we replaces blanks with `tabs` and updating corrupt `urls`.
But first all open PRs should be merged, else we will have merge conflicts.

I created a script to get all [corrupt links](https://florianbischof.net/leaflet/links.html). Not every suggestion / redirect is correct but it is a start.

TODOs:
- [ ] Merge [PRs](https://github.com/Leaflet/Leaflet/pulls?q=is%3Aopen+is%3Apr+label%3Aplugins)
- [ ] Replace blanks with tabs
- [ ] Update urls. [corrupt links](https://florianbischof.net/leaflet/links.html)

Plugins PRs ready to merge:
- [x] #7786

Any many others with merge conflicts or with blanks. (Maybe we can merge them with blanks and then updated manually by me)

mourner:
We also want to do at least minimal quality control on the plugins that get merged into the list, so that it doesn't list things that are not maintained / broken / not making much sense.

Falke-Design:
I would say this also includes an English description, what do you think? Example: https://github.com/wandergis/leaflet-echarts

IvanSanchez:
Since the leaflet website uses github pages under the hood, and in turn that uses Jekyll, there's the option of leveraging [Jekyll collections](https://jekyllrb.com/docs/collections/).

This means a rebuild of the *whole* plugins page, and turning each line of the plugins table into a separate markdown document. The plus side is that each plugins is a separate small markdown file, so conflicts go down and it'd be possible to add more attributes (e.g. version compatibility) to each plugin.

Falke-Design:
Reference #5387

Falke-Design:
But then we would have around 500 seperate files. We can also split it by the topics into seperate files.

I like the suggestion because then the data and layout are different things. Also for future things like adding a search for tags, version, npm etc.

Falke-Design:
Fixed by moving the plugins into seperate files #7805

PR #7804, opened 2021-11-23
WIP: Minimal jekyll setup for handling the plugins webpage in a data-driven manner

IvanSanchez:
Prompted by #7801.

This PR includes a *minimal* amount of stuff to demo how a collections-based plugins page would look like.

This is WIP since:
* [ ] all entries in the current plugins table need to be converted into individual markdown files
* [ ] `plugins-database.md` needs to cover all categories (instead of just the two first ones)
* [ ] Documentation for how to add a plugin to the list needs to be updated

cc @Falke-Design @mourner @johnd0e

### DIFF
--- a/docs/_plugins/leaflet-providers.md
+++ b/docs/_plugins/leaflet-providers.md
@@ -1,0 +1,12 @@
+---
+name: leaflet-providers
+category: basemap-providers
+repo: https://github.com/leaflet-extras/leaflet-providers
+author: leaflet-extras members
+author-url: https://github.com/leaflet-extras
+demo:
+compatible-v0: false
+compatible-v1: true
+---
+
+Contains configurations for various free tile providers & OSM, OpenCycleMap, Stamen, Esri, etc.
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -13,3 +13,8 @@ latest_leaflet_version: 1.7.1
 integrity_hash_css: "sha512-xodZBNTC5n17Xt2atTPuE1HxjVMSvLVW9ocqUKLsCC5CXdbqCmblAshOMAS6/keqq/sMZMZ19scR4PsZChSR7A=="
 integrity_hash_source: "sha512-I5Hd7FcJ9rZkH7uD01G3AjsuzFy3gqz7HIJvzFZGFt2mrCS4Piw9bYZvCgUE0aiJuiZFYIJIwpbNnDIM6ohTrg=="
 integrity_hash_uglified: "sha512-XQoYMqMTK8LvdxXYG3nZ448hOEQiglfqkJs1NOQV44cWnUrBc8PkAOcXy20w0vlaXaVUearIOBhiXZ5V3ynxwA=="
+
+
+collections:
+  plugins:
+    output: false
--- a/docs/_includes/plugin_category_table.html
+++ b/docs/_includes/plugin_category_table.html
@@ -1,0 +1,19 @@
+{% assign filteredplugins = site.plugins | where: "category", include.category %}
+
+<table class="plugins"><tr><th>Plugin</th><th>Description</th><th>Maintainer</th></tr>
+{% for plugin in filteredplugins %}
+	<tr>
+		<td>
+			<a href="{{ plugin.repo }}">{{ plugin.name }}</a>
+		</td><td>
+			{{ plugin.content | markdownify }}
+		</td><td>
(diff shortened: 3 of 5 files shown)


Falke-Design:
@IvanSanchez is this [explaination](https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#improving-documentation) up to date? I don't get it work

docs/_plugins/leaflet-providers.md
```diff
@@ -0,0 +1,12 @@
...
+category: basemap-providers
+repo: https://github.com/leaflet-extras/leaflet-providers
+author: leaflet-extras members
+author-url: https://github.com/leaflet-extras
+demo:
+compatible-v0: false
+compatible-v1: true
+---
+
+Contains configurations for various free tile providers & OSM, OpenCycleMap, Stamen, Esri, etc.
```
Falke-Design, line 12: Why are you adding the describtion without key? And not like `content: Contains configurations for va ....`?

IvanSanchez:
> @IvanSanchez is this [explaination](https://github.com/Leaflet/Leaflet/blob/master/CONTRIBUTING.md#improving-documentation) up to date? I don't get it work

Should be, yes. I've had a tad of trouble myself with `gem install rexml -v 3.2.3`, and Works In My Machine™.

docs/_plugins/leaflet-providers.md
IvanSanchez: `content` is defined by the jekyll engine. See https://jekyllrb.com/docs/variables/

Falke-Design:
It worked (Windows) after installing Rubx with DevKit and this steps: https://jekyllrb.com/docs/

But the plugins are not inserted:
grafik

IvanSanchez:
@Falke-Design It might be because I forgot to commit changes to `_config.yml` :sweat_smile:

mourner (commented):
This is awesome, thanks so much for doing this! I've been meaning to explore something like this for a long time. The idea of having plugins as separate md files is pretty brilliant, I haven't thought of that — assumed it would be some kind of a `plugins.yaml` file, but this has the huge advantage of pretty much eliminating the chance of merge conflicts.

One thing we should probably do is add some kind of `order` or `pin` field to "bubble up" the most important plugins to the top of its category. I also thought about doing some kind of automatic fetching of github stars per repo with GitHub Actions and sort by that, but not sure how involved that would be so we can leave that for later.

Falke-Design:
Yes I think a order would be good too, besides I suggest to order it alphabetical.
@mourner if you approve this suggestion, I would be starting with converting. First I will try if I can automate it but else I will do it manually. For each section a own PR.

IvanSanchez:
> One thing we should probably do is add some kind of `order` or `pin` field to "bubble up" the most important plugins to the top of its category.

According to https://jekyllrb.com/docs/collections/#custom-sorting-of-documents , custom sorting is available in jekyll v4, and according to https://pages.github.com/versions/ , gh-pages uses jekyll v3.9. :disappointed: 

If I'm reading this correctly, the `_config.yaml` setup for jekyll v4 would look like:

```yaml
collections:
  plugins:
    output: false
    sort_by: name
    order:
        - very-important-plugin.md
```

IvanSanchez:
> if you approve this suggestion, I would be starting with converting.

I suggest giving a thought to any other fields that *might* be useful, first. Compatibility fields are obvious (have the data so that the tables can display icons or hide/show plugins and whatnot), but what about...

* Split the git repo and homepage. Maybe some plugins don't have a git repo.
* License
* NPM package name, or link to a `package.json` for metadata
* Whether the plugin loads via ESM or UMD.

I'm not saying that those are mandatory; I'm just doing a quick brain dump.

IvanSanchez:
Superseded by #7805

mourner:
> > One thing we should probably do is add some kind of `order` or `pin` field to "bubble up" the most important plugins to the top of its category.
> 
> According to https://jekyllrb.com/docs/collections/#custom-sorting-of-documents , custom sorting is available in jekyll v4, and according to https://pages.github.com/versions/ , gh-pages uses jekyll v3.9. 😞

@IvanSanchez one way we could theoretically do it with v3.9 now is add an optional `pin: true` field in individual plugin files, and then simply filter based on that when outputting a list of plugins in our templates (e.g. first show all with `pin === true`, then all with `pin !== true`).

PR #7808, opened 2021-11-25
add/removePointerListener: tolerate wrong event names

johnd0e:
Seen in Leaflet.draw (`touchleave`)

### DIFF
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -31,12 +31,20 @@ export function addPointerListener(obj, type, handler) {
 	if (type === 'touchstart') {
 		_addPointerDocListener();
 	}
+	if (!handle[type]) {
+		console.warn('wrong event specified:', type);
+		return L.Util.falseFn;
+	}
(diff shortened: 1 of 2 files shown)


src/dom/DomEvent.Pointer.js
```diff
@@ -31,7 +31,7 @@ export function addPointerListener(obj, type, handler) {
 	if (type === 'touchstart') {
 		_addPointerDocListener();
 	}
-	handler = handle[type].bind(this, handler);
+	handler = handler[type].bind(this, handler);
```
Falke-Design, line 34: Are you sure that this is correct? Because the `handle` object contains the mapping to the functions.

https://github.com/Leaflet/Leaflet/blob/99361e805654d01e4adf4e325e677a2870aa16cd/src/dom/DomEvent.Pointer.js#L18-L23

johnd0e:
@Falke-Design Updated

Falke-Design (approved):

mourner (approved):

PR #7813, opened 2021-11-26
Locate - locationfound: Add check if map container has leaflet_id / is existing

### LINKED ISSUE #7508: _handleGeolocationResponse: exception when map is already removed

Falke-Design:
If the map is removed before the location is found, it throwed an error. Now it checks if the map container has the leaflet_id else it returns.

Fix #7508

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1327,7 +1327,7 @@ describe("Map", function () {
 				fn({coords: {latitude: 40.415296, longitude: 10.7419264, accuracy: 1129.5646101470752}});
 			}).to.not.throwException();
 		});
-		it("don't throw error if location is found and map is not existing", function () {
+		it("don't throw error if location is not found and map is not existing", function () {
 			map._locateOptions = {setView: true};
 			var fn = L.Util.bind(map._handleGeolocationError, map);
 			map.remove();
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1318,4 +1318,14 @@ describe("Map", function () {
 			expect(map.getZoom()).to.be(undefined);
 		});
 	});
+
+	describe("#Geolocation", function () {
+		it("don't throw error if location is found and map is not existing", function () {
+			var fn = L.Util.bind(map._handleGeolocationResponse, map);
+			map.remove();
+			expect(function (){
+				fn({ coords: { latitude: 40.415296, longitude: 10.7419264, accuracy: 1129.5646101470752} })
+			}).to.not.throwException();
+		});
+	});
 });
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1323,8 +1323,8 @@ describe("Map", function () {
 		it("don't throw error if location is found and map is not existing", function () {
 			var fn = L.Util.bind(map._handleGeolocationResponse, map);
 			map.remove();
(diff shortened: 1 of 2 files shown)


src/map/Map.js
```diff
@@ -665,6 +665,8 @@ export var Map = Evented.extend({
 	},
 
 	_handleGeolocationError: function (error) {
+		if (!this.getContainer()._leaflet_id) { return; }
```
johnd0e, line 668: Perhaps here we can safely use private `_container` property.

spec/suites/map/MapSpec.js
```diff
@@ -1318,4 +1318,14 @@ describe("Map", function () {
 			expect(map.getZoom()).to.be(undefined);
 		});
 	});
+
+	describe("#Geolocation", function () {
+		it("don't throw error if location is found and map is not existing", function () {
+			var fn = L.Util.bind(map._handleGeolocationResponse, map);
```
johnd0e, line 1324: Perhaps you should ass location error test as well

Falke-Design: Have you an idea how to test? I don't throw an error so I can't catch them

Falke-Design: Maybe I can listen if the `locationerror` event is fired

johnd0e: > Have you an idea how to test?

In the same manner - call `_handleGeolocationError` and check it does not throw.

Falke-Design: `_handleGeolocationError` doesn't throw an error, so I can't catch it

johnd0e: Even if `setView` option is used?

Falke-Design: You are right. I have tested it wrong (while map initialisation I set center and zoom) and then setView was not called, because the map was already loaded.

spec/suites/map/MapSpec.js
```diff
@@ -1318,4 +1318,22 @@ describe("Map", function () {
 			expect(map.getZoom()).to.be(undefined);
 		});
 	});
+
+	describe("#Geolocation", function () {
+		it("don't throw error if location is found and map is not existing", function () {
```
johnd0e, line 1323:
```suggestion
		it("doesn't throw error if location is found and map is not existing", function () {
```

spec/suites/map/MapSpec.js
```diff
@@ -1318,4 +1318,22 @@ describe("Map", function () {
...
+			var fn = L.Util.bind(map._handleGeolocationResponse, map);
+			map.remove();
+			expect(function () {
+				fn({coords: {latitude: 40.415296, longitude: 10.7419264, accuracy: 1129.5646101470752}});
+			}).to.not.throwException();
+		});
+		it("don't throw error if location is not found and map is not existing", function () {
```
johnd0e, line 1330:
```suggestion
		it("doesn't throw error if location is not found and map is not existing", function () {
```

johnd0e (changes requested):

johnd0e (approved):

PR #7814, opened 2021-11-26
Add `false` to docu of Attribution prefix

Falke-Design:
Continued from #7320

Thanks to @simon04

### DIFF
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -19,7 +19,7 @@ export var Attribution = Control.extend({
 	options: {
 		position: 'bottomright',
 
-		// @option prefix: String = 'Leaflet'
+		// @option prefix: String|false = 'Leaflet'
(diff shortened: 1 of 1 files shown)


src/control/Control.Attribution.js
```diff
@@ -19,7 +19,7 @@ export var Attribution = Control.extend({
 	options: {
 		position: 'bottomright',
 
-		// @option prefix: String = 'Leaflet'
+		// @option prefix: String|false = 'Leaflet'
```
jonkoops, line 22: Technically it looks like any falsy value is will trigger this behavior so `''`, `undefined`, `null` or `0` as well. Is an explicit boolean what we want here?

johnd0e: All mentioned values coerce to `false`, so why not make it explicit.

mourner (approved):

PR #7823, opened 2021-11-28
Draggable: Remove offset subtract from _startPos

Falke-Design:
I will try to explain this, hopefully it works.

If the dragging is started (`_onDown`) and then the first `mousemove` event is fired (`_onMove`), the `_startPos` is calculated wrong. The element can't be moved until now, because it is the first `mousemove` event, so it makes no sense that for the `startPos` calculation the `offset` is removed from the current position. -> there is no offset because the element is not moved.
```
this._startPos = DomUtil.getPosition(this._element).subtract(offset);
```

Example: https://plnkr.co/edit/Fz9jPeP7JIrPleXo
1. Create draggable Marker
2. Hold down on the Marker
3. Call manual `_onMove` to visualize the wrong `_startPos` calculation.
```
var marker = new L.Marker([0, 0], { draggable: true}).addTo(map);
setTimeout(function () {
    marker.dragging._draggable._onMove({
        type: 'mousemove',
        buttons: 1,
        clientX: 100,
        clientY: 100,
        target: marker._icon
    });
}, 3000);
```
**Expectation:**
The marker "jumps" to the position of the `mousemove` event ([100, 100]). But instead it creates the `_startPos` (`currentElementPos - offset`) and then it calculates the new position of the element by calling `_startPos + offset` = `currentElementPos - offset + offset` = `currentElementPos` -> And now we are at exact same position where the marker already is. If the next `mousemove` is fired, the dragging starts working because now the `_startPos` is defined and the offset is different.

**Solution:**
This can be simply fixed by changing:
```
this._startPos = DomUtil.getPosition(this._element).subtract(offset);
```
to:
```
this._startPos = DomUtil.getPosition(this._element);
```

---------

We see this wrong usage in already existing tests. If `toucher.wait(100).moveTo(200, 200, 0)
.down().moveBy(5, 0, 20).moveBy(256, 32, 200).up();` is called, the first `mousemove` while dragging is ignored `moveBy(5, 0, 20)` and the marker is not moved.

The offset subtracting was added in this commit https://github.com/Leaflet/Leaflet/commit/2d2fc74110f2fe7fbc06d7f6e0a2e90528ce46c6

### DIFF
--- a/spec/suites/layer/marker/Marker.DragSpec.js
+++ b/spec/suites/layer/marker/Marker.DragSpec.js
@@ -114,4 +114,36 @@ describe("Marker.Drag", function () {
 				.down().moveBy(5, 0, 20).moveBy(290, 32, 1000).wait(100).up().wait(100);
 		});
 	});
+
+	it("checks if the offset of the first mousemove event is correct", function (done) {
+		map.setZoom(15);
+		var marker = new L.Marker([0, 0], {
+			draggable: true
+		}).addTo(map);
+		var offset = 20;
+
+		var hand = new Hand({
+			timing: 'fastframe',
+			onStop: function () {
+				marker.dragging._draggable._onMove({
+					type: 'mousemove',
+					buttons: 1,
+					clientX: 100,
+					clientY: 100,
+					target: marker._icon
+				});
+
+				setTimeout(function () {
+					// marker latlng is lower then the point where the mouse has clicked (offset)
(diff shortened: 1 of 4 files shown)


johnd0e:
Seems duplicate of some other pr

Falke-Design:
Ah yes I found #7446. 
@johnd0e Can you convert your PR from draft to "ready for review" ? Then I will close this one.

johnd0e:
Close in favor of #7446

PR #7827, opened 2021-11-29
Lint examples

mourner:
Adds linting to examples. There are 163 issues to fix now but they're mostly trivial (styling, spaces etc).


Falke-Design:
@mourner do you want to change this by your self or should I take this and fix it?

PR #7835, opened 2021-12-04
Add touchevents to Canvas

Falke-Design:
Touchevents are now passed to the layer through Canvas.

It was not possible to listen on mouse or touch events on a layer if `preferCanvas` is enabled on the map.

"No" events fired:
canvas_touch_wrong

Fixed: Touchevents fired
canvas_touch_working

I created a demo code for this, which catches all mouse and touch events on the layer. You will see in the "current demo" nothing is fired / a few times (very randomly) mousemove fired. 

```
circle.on('mousedown touchstart mousemove touchmove mouseup touchend', (e)=>{
			var obj = events[e.type];
			events[e.type] = obj ? obj + 1 : 1;

			var elm = document.getElementById('debug');

			var text = "";
			for(var type in events) {
				text += type+": "+events[type]+"<br>";
			}
			elm.innerHTML = text;
		});
```

[Here](https://plnkr.co/edit/EjJh2WRwiXIEZ10d) is a Plunkr, but touch simulation is not working there, so I added the demo on my server too:
- [Current](https://florianbischof.net/leaflet/canvas_touch/old.html)
- [New / fixed](https://florianbischof.net/leaflet/canvas_touch/)

### DIFF
--- a/spec/suites/SpecHelper.js
+++ b/spec/suites/SpecHelper.js
@@ -44,7 +44,7 @@ expect.Assertion.prototype.nearLatLng = function (expected, delta) {
 };
 
 happen.at = function (what, x, y, props) {
-	this.once(document.elementFromPoint(x, y), L.Util.extend({
+	var event = L.Util.extend({
 		type: what,
 		clientX: x,
 		clientY: y,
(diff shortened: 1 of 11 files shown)


Falke-Design:
It looks like if the browser has no nativ touch support, that then the `touch` listeners not working correct. `DomEvent.Pointer` should make this possible but it looks like it doesn't work. 

How can I fix this?

johnd0e:
> but it looks like it doesn't work.

Sure it does. We have unit tests for that.
Please explain your issue in details

Falke-Design:
Touchevents are missing overall, not only in Canvas. 

PR for all touchevents: #7842

Issue #7752, opened 2021-11-07
Touchevents missing?


Falke-Design:
Are here the touch events missing?
https://github.com/Leaflet/Leaflet/blob/257ca098f1edfe091ec029bcead25f7f9ce2b676/src/layer/vector/Canvas.js#L60-L69

And maybe here?
https://github.com/Leaflet/Leaflet/blob/257ca098f1edfe091ec029bcead25f7f9ce2b676/src/map/Map.js#L1311-L1312

mourner:
No, touch devices still fire mouse events. https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent

Falke-Design:
Okay thanks, then I was not so wrong with https://github.com/Leaflet/Leaflet/pull/7059#pullrequestreview-796457015 👍

Falke-Design:
@mourner After digging more into this, it looks like mouse and touchevents are **not** fired in the same way. The mouseevents are still there but not usable.
I created a very minimalistic example without any librarys and added touch & mouse events to a DOM element.
The touchevents fired much often than the mouseevents. Also mouseevents are only fired if we click on the element without moveing.

Tested on Firefox Touch Simulation, Firefox Android, Chrome Android

```
<div id="map" style="width: 400px; height: 400px;background: #f00"></div>
<div id="debug" style="position: absolute; top: 0;z-index: 10000000;background: #ffffff;"></div>


var map = document.getElementById('map');

    var events = {}
	function updateHtml(e){
		var obj = events[e.type];
		events[e.type] = obj ? obj + 1 : 1;

		var elm = document.getElementById('debug');

		var text = "";
		for(var type in events) {
			text += type+": "+events[type]+"<br>";
		}
		elm.innerHTML = text;
	}

	map.addEventListener('touchmove',updateHtml)
	map.addEventListener('mousemove',updateHtml)

	map.addEventListener('touchstart',updateHtml)
	map.addEventListener('mousedown',updateHtml)

	map.addEventListener('touchend',updateHtml)
	map.addEventListener('mouseup',updateHtml)
```

Demo: https://florianbischof.net/leaflet/testEvents.html

What do you think?

johnd0e:
> The touchevents fired much often than the mouseevents. Also mouseevents are only fired if we click on the element without moveing.

That is right. Only 'click' produces compatibility mouse events.

johnd0e:
[...]

PR #7842, opened 2021-12-08
Add touch support to Map, Layer, Canvas

Falke-Design:
Fix: #7752

Touchevents are now passed to the Map and Layers, it is also working through Canvas.

No touchevents fired:
canvas_touch_wrong

Fixed: Touchevents fired
canvas_touch_working


src/layer/vector/Canvas.js
```diff
@@ -354,7 +356,11 @@ export var Canvas = Renderer.extend({
 	// so we emulate that by calculating what's under the mouse on mousemove/click manually
 
 	_onClick: function (e) {
-		var point = this._map.mouseEventToLayerPoint(e), layer, clickedLayer;
+		var first = e.touches ? e.touches[0] : e;
+		if (['touchend', 'touchcancel', 'pointerup', 'pointercancel'].indexOf(e.type) > -1) {
+			first = e.changedTouches[0];
+		}
+		var point = this._map.mouseEventToLayerPoint(first), layer, clickedLayer;
```
Falke-Design, line 363: Instead of doing this for every `mouseEventToLayerPoint` call, we can update the `getMousePosition`:
https://github.com/Leaflet/Leaflet/blob/e3719a317ab325ed091fbbc6bfd34640fe19e13e/src/dom/DomEvent.js#L230-L244

PR #7843, opened 2021-12-08
Fix word

jeremypeters:


Falke-Design:
Thank you for your contribution.

In my eyes your change is not good. I think it is better to make (keep) clear that all layers can be added to the LayerGroup and not only Markers.

Issue #7821, opened 2021-11-28
Use `<button>` for controls


Malvoz:
**Is your feature request related to a problem? Please describe.**

`<button>` behaves a bit differently from `<a href="#" role="button">`, this would supersede https://github.com/Leaflet/Leaflet/issues/7368 and https://github.com/Leaflet/Leaflet/issues/7369 (these describe some of the problems with `<a>` controls).

**Describe the solution you'd like**

Use `<button type="button">` over `<a href="#">` for controls (and over `<img tabindex="0">` for markers?).

**Additional context**

  > >Use buttons, not links, for controls
  >
  > Agreed!

  _Originally posted by @mourner in https://github.com/Leaflet/Leaflet/issues/3210#issuecomment-73291846_

  > we should use the button element here

  > This would be a breaking change we can make in Leaflet 2.

  _Originally posted by @jonkoops in https://github.com/Leaflet/Leaflet/issues/7368#issuecomment-976742275_

- Related PR: https://github.com/Leaflet/Leaflet/pull/6248
- Reference: https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/button/button.html

Malvoz:
This change would require a CSS reset for buttons, something along the lines of:

```css
button.leaflet-control {
  background-color: transparent;
  border: none;
  border-radius: 0;
  color: inherit;
  font: inherit;
  line-height: inherit;
  margin: 0;
  padding: 0;
  overflow: hidden;
  text-align: inherit;
  text-transform: none;
  -webkit-appearance: none;
     -moz-appearance: none;
          appearance: none;
}
```

(this is just a generic class I've been using before, if Leaflet will use a default border-radius for all controls for example there's probably no reason to reset that.)

Falke-Design:
I only want to document this: We should **not** remove the existing css classes because they are used for many custom controls

jonkoops:
I've created a label to keep track of breaking changes like these.

Chi-teck:
Would be nice it those buttons had accessible name.
https://www.drupal.org/project/leaflet/issues/3232474

Malvoz:
@Chi-teck The buttons would have accessible names as they have now. This issue is about the elements themselves.

> https://www.drupal.org/project/leaflet/issues/3232474

[...]

PR #7850, opened 2021-12-09
Set `role=button` on the Layer control

### LINKED ISSUE #6821: Layers Control has a WCAG Level A Empty Link Error

### LINKED ISSUE #7498: Layer control should have `role="button"`

Malvoz:
Closes #6821 - This PR closes that issue because the WCAG error reported is only applicable to elements with `role=link` (implicit from `<a>` in the case of the layer control) and not buttons (which this PR changes)!

Closes #6869 - Unlike that PR, this doesn't set `aria-label` (in addition to the existing `title`) as that's not only redundant but would also cause some screen readers to announce "Layers" twice - which is an obvious annoyance to screen reader users.

Fix #7498.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -201,6 +201,7 @@ export var Layers = Control.extend({
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
 		link.href = '#';
 		link.title = 'Layers';
+		link.setAttribute('role', 'button');
 
 		DomEvent.on(link, 'click', DomEvent.preventDefault); // prevent link function
 		DomEvent.on(link, 'focus', this.expand, this);


Falke-Design (approved):

mourner (approved):

vanillajonathan:
Maybe using a `button` element ought to be considered instead of a `a` element?

Malvoz:
@vanillajonathan see https://github.com/Leaflet/Leaflet/issues/7821.

Issue #7490, opened 2021-03-02
Some controls don't display properly when printing


Malvoz:
Controls such as the layer control and the [fullscreen](https://github.com/Leaflet/Leaflet.fullscreen) control that use background-images don't display properly when printing.

See https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/294 for more information and proposed solutions.

PR #7816, opened 2021-11-26
Simplifiy examples

Falke-Design:
- I changed the name of the map variable `mymap` to `map`
- And added variable names to variables (easier for debugging / testing)

Very similar to #7269 but without updating the text and changing to osm

PR #7851, opened 2021-12-09
Add print styles to prevent printers from removing background-images in controls

Malvoz:
Fix https://github.com/Leaflet/Leaflet/issues/7490.

### After

See CSS [`color-adjust`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-adjust) for more info.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -650,6 +650,6 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	/* Prevent printers from removing background-images of controls. */
 	.leaflet-control {
 		-webkit-print-color-adjust: exact;
-		              color-adjust: exact;
+		color-adjust: exact;
 		}
 	}
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -643,3 +643,13 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	margin-left: -12px;
 	border-right-color: #fff;
 	}
+
+/* Printing */
+	
+@media print {
+	/* Prevent printers from removing background-images of controls. */
+	.leaflet-control {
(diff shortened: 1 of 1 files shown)


Malvoz:
I think it'd desirable to revisit the selector used here if https://github.com/Leaflet/Leaflet/pull/7204 gets merged, so that markers don't disappear when printing. 🙂

dist/leaflet.css
```diff
@@ -643,3 +643,13 @@ svg.leaflet-image-layer.leaflet-interactive path {
...
 	border-right-color: #fff;
 	}
+
+/* Printing */
+	
+@media print {
+	/* Prevent printers from removing background-images of controls. */
+	.leaflet-control {
+		-webkit-print-color-adjust: exact;
+		              color-adjust: exact;
```
Falke-Design, line 653: There a spaces used instead of tabs

Falke-Design (changes requested):

dist/leaflet.css
Malvoz: Where is it documented what to use, and the tab size?

Falke-Design: @mourner? I only know it because it was / is often request on my PRs
https://github.com/Leaflet/Leaflet/pull/7816#pullrequestreview-817002397

Falke-Design (approved):

dist/leaflet.css
mourner: It's fine in this case because spaces are used for alignment, not indentation. Generally I think we'll convert the whole project to spaces at some point (and fix CRLF at the same time) — it brings more problems than it solves.

mourner (approved):

PR #7852, opened 2021-12-09
Cleanup all tests

Falke-Design:
I generalized all tests so that all have the same "base":
- replaced `new L.XXXX` with the factory method `L.xXXX`
- Using everywhere `beforeEach` and `afterEach` to make sure that not the map / container / variables from the test before are used
- Using `container` and `map` as default variable name for the map-container and the map
- Removed `PolylineGeometrySpec.js` and moved the tests to `Polyline` and `LineUtils`


IvanSanchez (approved):
LGTM

spec/index.html
```diff
@@ -23,6 +23,12 @@
 
 	<script src="../dist/leaflet-src.js"></script>
 
+    <script>
+		// sets the default imagePath to the project path like "/Leaflet/dist/images/"
+        // same as in after.js but not static
+		L.Icon.Default.imagePath = '/' + window.location.pathname.split('/')[1] + '/dist/images/';
+    </script>
+
```
mourner, line 31: Nit: mixed tabs / spaces here

spec/suites/SpecHelper.js
```diff
@@ -79,5 +79,33 @@ var touchEventType = L.Browser.touchNative ? 'touch' : 'pointer'; // eslint-disa
...
 //       see https://github.com/Leaflet/prosthetic-hand/issues/14
 
+function createContainer() {
+	var container = document.createElement("div");
+	container.style.position = 'absolute';
+	container.style.top = '0px';
+	container.style.left = '0px';
+	container.style.height = '400px';
+	container.style.width = '400px';
+	container.style.opacity = '0.4';
```
mourner, line 89: Any particular reason to have 0.4 opacity here?

spec/suites/SpecHelper.js
```diff
@@ -79,5 +79,33 @@ var touchEventType = L.Browser.touchNative ? 'touch' : 'pointer'; // eslint-disa
...
+	document.body.appendChild(container);
+
+	return container;
+}
+
+function createMapContainer(mapOptions) { /* eslint-disable-line no-unused-vars */
+	var container = createContainer();
+	var map = L.map(container, mapOptions);
+	return {map: map, container: container};
+}
```
mourner, line 99: I find this particular function redundant — you return an object, then unroll it:

```js
var obj = createMapContainer(options);
map = obj.map;
container = obj.container;
```

An equivalent is both more explicit and simpler:

```js
container = createContainer();
map = L.map(container, options);
```

spec/suites/SpecHelper.js
```diff
@@ -79,5 +79,33 @@ var touchEventType = L.Browser.touchNative ? 'touch' : 'pointer'; // eslint-disa
 // Note: this override is needed to workaround prosthetic-hand fail,
 //       see https://github.com/Leaflet/prosthetic-hand/issues/14
 
+function createContainer() {
```
mourner, line 82: We could also add optional `width` and `height` parameters here to cut down on boilerplate.

mourner (changes requested):
Looks great apart from a few nits!

spec/suites/SpecHelper.js
Falke-Design: Yes, the map is displayed always in the top left corner and hides the passed tests behinde it. With opacity it is possible to see while running if the tests are worked or failed without waiting until all tests passed or there is a slot without a map

spec/suites/SpecHelper.js
Falke-Design: With keeping the map in an own option is makes it possible to adjust while developing only one place. For example when I want to implement touchevents and I debug failed tests, it can be very useful if there is only one function which creates the map and I need only to edit one place to add for example listeners to the map.
I think if we always do the same (like creating the map instance) we can do this i one central place.

Would it be better if I move the map creation into a own function without creating the container?

spec/suites/SpecHelper.js
Falke-Design: grafik

mourner: Can we just put the test results list over the map, e.g. with `position: relative` and a higher `z-index`?

spec/suites/SpecHelper.js
mourner: I would introduce new utility functions when they are actually needed, instead of trying to anticipate a chance of something being useful in the future. The more explicit the code, the better.

spec/suites/SpecHelper.js
Falke-Design: No because of the click events by coords

spec/suites/SpecHelper.js
Falke-Design: It was a feature that I needed to debug else I would suggest it 😄 but is ok for me, I will replace it

spec/suites/SpecHelper.js
mourner: Maybe position it to the right rather than left then?

Falke-Design: While merging the newest main branch into it, I saw again the advantages of opacity 0.4 . @mourner should I revert it back or can we keep it? Let us decide that and then merge it. It doesn't make fun to update this PR always after changes in the main 😄

mourner: Let's keep, not a strong opinion on this one, and we'll be able to quickly update it in the future now.

mourner (approved):

PR #7855, opened 2021-12-10
Fixing noConflict and merge rollup-config

Falke-Design:
In all tests that I made, I was not able to reset `L` variable to the value before initializing Leaflet with `L.noConflict()`. Can this please confirmed by some others?

https://plnkr.co/edit/N3qnFEscKsGinjoC
or go to https://leafletjs.com/examples/quick-start/example.html and add in the console:
```
console.log(L.noConflict()); // should return the leaflet instance
console.log(L); // should be undefined or returning the old value
```

Very strange is that it works in a project from me (Leaflet is loaded over npm V1.7.1): https://www.lagekarte.info/

**Changes:**
1. I changed in the rollup the `ouput.name` to `leaflet` because if we use `L` Leaflet was already attached to global `L` variable and then `noConflicts` don't worked because `L` was already overwritten by Leaflet.
2. I merged the two rollup files because if there will be changes in the config they will be applied to `watch` and `build` process. The reason for this changes was, that `noConflict` was not added in the `rollup-watch-config` and then it was not added while developing and running `watch`.
3. I changed in the **karma.conf.js** the `ouput.name` to `leaflet` same as point 1. I also wanted to import the config from the rollup  into karma, to have one central config, but it was not working because karma can't handle `import`: `SyntaxError: Cannot use import statement outside a module`


build/rollup-config.js
```diff
@@ -7,42 +7,48 @@ import pkg from '../package.json';
...
 			esModule: false
-		},
+		}
+	],
+	plugins: [
+		release ? json() : rollupGitVersion()
+	]
+};
+
+if (!watch) {
```
jonkoops, line 43: Why would we not want to build this during watch mode?

Falke-Design: Why not? Like already described I was developing in Leaflet and checked the function `noConflict` and it was not existing because of the watch mode. It would be good to keep the watch and release builds similar, else it can be confusing or something can be missed.

Falke-Design: Why should it not be in the watch mode? Is there a good reason? The build time can't be the reason, because there is not a real difference

Falke-Design: Sorry misunderstood the context behinde your question. I don't know why we don't want to build this. This was already configurated like that. Maybe a build-time reason

mourner (approved):

PR #7859, opened 2021-12-12
Add ignoreBounds option to GridLayer.

martinschneider:
This option allows ignoring tiles within a specified `LatLngBounds`, e.g. when tiles for that area will be provided by a different layer.

This prevents loading invisible tiles and solves the problem described in [this StackOverflow post](https://stackoverflow.com/questions/28551376/avoid-loading-of-invisible-tiles-when-using-multiple-tile-layers-in-leaflet).

Tested on Firefox 78.15.0esr (64-bit).


martinschneider:
Are there any updates on whether this feature can be accepted? What is the process like @Falke-Design? Thank you.

PR #7862, opened 2021-12-12
Fix opening / closing tooltip while dragging map

Falke-Design:
While dragging the map it can happen that the mouse is moved faster then the map and then on the marker is `mouseover` and `mouseout` fired. Then this flickering happens (opening / closing):

tooltip_flakering

To prevent this I added a check in `openTooltip` if the map is currently moving.

This means:
- If the tooltip is open and while dragging `mouseout` is fired, it closes and don't opening again if `mouseover` is fired later.
- If the tooltip is closed and `mouseover` is fired, it don't opens.

Fixed:
tooltip_flakering_fix
While the mouse is on the marker the tooltip keeps open, but if it goes away from the marker, `mouseout` is fired and the tooltip closes and is not opened again.

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -388,7 +388,7 @@ Layer.include({
 	},
 
 	_openTooltip: function (e) {
-		if (!this._tooltip || !this._map) {
+		if (!this._tooltip || !this._map || this._map.dragging.moving()) {
 			return;
 		}
 		this._tooltip._source = e.layer || e.target;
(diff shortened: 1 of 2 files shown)


src/layer/Tooltip.js
```diff
@@ -388,7 +388,7 @@ Layer.include({
 	},
 
 	_openTooltip: function (e) {
-		if (!this._tooltip || !this._map) {
+		if (!this._tooltip || !this._map || this._map.dragging.moving()) {
```
johnd0e, line 391:
```suggestion
		if (!this._tooltip || !this._map || (this._map.dragging && this._map.dragging.moving())) {
```

mourner (approved):

PR #7867, opened 2021-12-14
Support dashOffset on canvas layer

McBen:
the PathOption `dashOffset` is currently ignored on canvas layers.

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -340,7 +340,8 @@ export const Canvas = Renderer.extend({
 
 		if (options.stroke && options.weight !== 0) {
 			if (ctx.setLineDash) {
-				ctx.setLineDash(layer.options && layer.options._dashArray || []);
+				ctx.lineDashOffset = Number(options.dashOffset || 0);
+				ctx.setLineDash(options._dashArray || []);
 			}
 			ctx.globalAlpha = options.opacity;
 			ctx.lineWidth = options.weight;


src/layer/vector/Canvas.js
```diff
@@ -339,7 +339,8 @@ export var Canvas = Renderer.extend({
 
 		if (options.stroke && options.weight !== 0) {
 			if (ctx.setLineDash) {
-				ctx.setLineDash(layer.options && layer.options._dashArray || []);
+				ctx.lineDashOffset = Number(options.dashOffset || 0);
```
Falke-Design, line 342: Is `Number` really neaded? We don't normaly use this in Leaflet

```suggestion
				ctx.lineDashOffset = options.dashOffset || 0;
```

McBen: a) Leaflet docs say dashOffset is string; lineDashOffset should be float
b)  _dashArray is processed the same way. see [Canvas._updateDashArray](https://github.com/Leaflet/Leaflet/blob/e9fc60c022fab6557744576f324c82ae1958b16c/src/layer/vector/Canvas.js#L195)

Falke-Design: Interessting ... you are right 👍

src/layer/vector/Canvas.js
```diff
@@ -339,7 +339,8 @@ export var Canvas = Renderer.extend({
 
 		if (options.stroke && options.weight !== 0) {
 			if (ctx.setLineDash) {
```
Falke-Design, line 341: Please add the check `options._dashArray`

Falke-Design (changes requested):
Good catch!

Falke-Design (approved):

AnthorNet:
Any news on that one?

PR #7868, opened 2021-12-14
Workaround for DashArray with starting 0 on canvas layers

McBen:
On svg layers a `dashArray="0 8 16 8"` is fine.  
`(example result: "  ----    ----  ")`
On canvas layers the internal `setLineDash([0,8,16,8])` will create artefacts at the begining of the dash. 
`(example result: "-  ----  -  ----  ")`

So we got different results on svg and canvas layers.

This can be workaround by converting the first zero argument into the dashOffset.

( required PR #7867 to support dashOffset)

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -339,7 +339,8 @@ export var Canvas = Renderer.extend({
 
 		if (options.stroke && options.weight !== 0) {
 			if (ctx.setLineDash) {
-				ctx.setLineDash(layer.options && layer.options._dashArray || []);
+				ctx.lineDashOffset = Number(options.dashOffset || 0);
+				ctx.setLineDash(options._dashArray || []);
 			}
 			ctx.globalAlpha = options.opacity;
 			ctx.lineWidth = options.weight;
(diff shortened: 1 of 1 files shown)


Falke-Design:
> ( required PR #7867 to support dashOffset)

This PR will already contains the changes of #7867

src/layer/vector/Canvas.js
```diff
@@ -339,7 +339,13 @@ export var Canvas = Renderer.extend({
 
 		if (options.stroke && options.weight !== 0) {
 			if (ctx.setLineDash) {
-				ctx.setLineDash(layer.options && layer.options._dashArray || []);
+				ctx.lineDashOffset = Number(options.dashOffset || 0);
```
Falke-Design, line 342: Is `Number` really neaded? We don't normaly use this in Leaflet

```suggestion
				ctx.lineDashOffset = options.dashOffset || 0;
```

Falke-Design:
@McBen can you please provide a sample. I can't reproduce the artefact-differences from canvas and svg.
Here a sample: https://plnkr.co/edit/dwb7kGpUrwA0xwsH with 1.7.1

This `artefacts` are because of linecap: 
 - https://plnkr.co/edit/Wltt2AmC7eG0qN9p
 - https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap

McBen:
yes, you're right. It's about the linecap.

PR #7872, opened 2021-12-20
Remove will-change

janjaap:
Prevent browser (mostly Firefox) resource depletion by premature optimization usage of `will-change`. At the municipality of Amsterdam, we're heavily relying on Leaflet and having map instances open regularly leads to degraded performance in Firefox, because of the use of `will-change`.

See https://developer.mozilla.org/en-US/docs/Web/CSS/will-change for reference.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -171,9 +171,6 @@
 
 /* zoom and fade animations */
 
-.leaflet-fade-anim .leaflet-tile {
-	will-change: opacity;
-	}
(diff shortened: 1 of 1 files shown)


Falke-Design:
Comes from this commit: https://github.com/Leaflet/Leaflet/commit/6b67df023234186e70647334c381fd66e4838721

mourner (approved):
Yeah, looks like things perform well without it, and the recommendations on the MDN page are convincing.

Issue #7891, opened 2022-01-07
License yearly update


marcoschicote:
**Is your feature request related to a problem? Please describe.**
Every year we need to update the License file to reflect the new year.

**Describe the solution you'd like**
Implement Github action to create a PR updating the License year automatically every year. This is available at https://github.com/marketplace/actions/update-license-copyright-year-s

**Describe alternatives you've considered**
Update the year manually. This would make sense if the Github action did not exist, but why not use something that is already developed?

**Additional context**
None

marcoschicote:
Created the PR for this on https://github.com/Leaflet/Leaflet/pull/7892

PR #7893, opened 2022-01-07
Added Github action to update license year automatically

marcoschicote:
Fixes https://github.com/Leaflet/Leaflet/issues/7891

### DIFF
--- a/.github/workflows/update-copyright-years-in-license-file.yml
+++ b/.github/workflows/update-copyright-years-in-license-file.yml
@@ -1,0 +1,16 @@
+name: Update copyright year(s) in license file
+
+on:
+  schedule:
+    - cron: "0 3 1 1 *"
+
+jobs:
+  action-update-license-year:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: FantasticFiasco/action-update-license-year@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}


Falke-Design:
Have you checked if it works correctly? The `LICENSE.md` file contains two date ranges:

https://github.com/Leaflet/Leaflet/blob/e9bc85952d955347bbfd0a2d7909fdc51f08a48a/LICENSE#L3-L4

marcoschicote:
> Have you checked if it works correctly? The `LICENSE.md` file contains two date ranges:
> 
> https://github.com/Leaflet/Leaflet/blob/e9bc85952d955347bbfd0a2d7909fdc51f08a48a/LICENSE#L3-L4

Yes, this is a manual run of the Github action: https://github.com/marcoschicote/Leaflet/pull/2/files

Falke-Design (approved):

jonkoops (approved):
This seems like a good way to automate a trivial but annoying task. However I wonder if keeping this workflow file maintained might end up being more work than manually adding a commit for the copyright dates.

marcoschicote:
> This seems like a good way to automate a trivial but annoying task. However I wonder if keeping this workflow file maintained might end up being more work than manually adding a commit for the copyright dates.

If that were to be the case it's just a matter of deleting the Github action. Hopefully this will not happen :)

mourner:
Appreciate the contribution, but personally I don't think non-essential actions that need to happen once a year need automation — it adds complexity and a dependency on one more third-party GH action. Also, there are more places where the year could be updated (e.g. at the bottom of the website).

lordt4ever:
OK.man

marcoschicote:
> Appreciate the contribution, but personally I don't think non-essential actions that need to happen once a year need automation — it adds complexity and a dependency on one more third-party GH action. Also, there are more places where the year could be updated (e.g. at the bottom of the website).

Did not know the year was also included in the footer. Shouldn't we use `new Date().getFullYear()` there?

marcoschicote:
> I close this PR because of @mourner and @jonkoops concerns.
> 
> @marcoschicote would be very nice to see a PR from you which updates the `LICENSE.md` and the footer 😊

Will do! 
But what's your opinion on how to fix the footer? Should we use `new Date().getFullYear()` or hardcode the year?

jonkoops:
IMHO we could just add `<script>document.write(new Date().getFullYear())</script>` to the footer.

Falke-Design:
I created #7907 which uses a jekyll command instead of JavaScript

Issue #7428, opened 2021-01-23
Interactive markers should have an accessible name and `role="button"`


Malvoz:
## Motivation

Interactive markers (**markers that bind to popups**) currently have no alternative text for screen readers (not announced, or announced as _blank_), they're also conveyed as images (not actionable buttons):

<img width="450" src="https://user-images.githubusercontent.com/26493779/105568669-7d34d480-5d3b-11eb-8a87-8d2375fe6de5.png" alt="inaccessible marker">

## Proposed solution

Interactive markers should have an accessible name by default, for example `alt="Marker"` (may be fixed by https://github.com/Leaflet/Leaflet/pull/7079/commits/f7e1cd9f5b5433ffb03784cfcdd5ba4b9955ca90 in https://github.com/Leaflet/Leaflet/pull/7079), and `role="button"`:

<img width="450" src="https://user-images.githubusercontent.com/26493779/105568661-5eced900-5d3b-11eb-9255-1ad8488a45f0.png" alt="accessible marker">

## Additional context
- MDN: [Interactive elements must be labeled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#interactive_elements_must_be_labeled)
- Related PRs: https://github.com/Leaflet/Leaflet/pull/7079, https://github.com/Leaflet/Leaflet/pull/6248.

Malvoz:
It should be noted that due to the missing `role="button"` the map actually zooms in instead of opening the popup when using a screen reader (tested in NVDA) after pressing <kbd>Enter</kbd> when a marker is focused. This was a bit unexpected to me, other controls e.g. `<a>` that also are missing `role="button"` work fine in terms of keyboard operability, presumably because they're standard interactive elements.

PR #7895, opened 2022-01-09
Update marker: role=button, alt=marker

tmiaa:
Marker is set to alt=marker and role=button to Fix https://github.com/Leaflet/Leaflet/issues/7428 and now it's possible to open the pop-ups with my screen reader, it's not possible here: https://leafletjs.com/examples/quick-start/

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -40,9 +40,9 @@ export var Marker = Layer.extend({
 		// Text for the browser tooltip that appear on marker hover (no tooltip by default).
 		title: '',
 
-		// @option alt: String = ''
+		// @option alt: String = 'Marker'
 		// Text for the `alt` attribute of the icon image (useful for accessibility).
-		alt: '',
+		alt: 'Marker',
 
 		// @option zIndexOffset: Number = 0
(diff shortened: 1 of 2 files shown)


mourner (approved):
Thank you! cc @Malvoz

Malvoz (approved):
LGTM, a much better default. Thanks!

src/layer/marker/Marker.js
```diff
@@ -40,9 +40,9 @@ export var Marker = Layer.extend({
 		// Text for the browser tooltip that appear on marker hover (no tooltip by default).
 		title: '',
```
mrlessc, line 41: Here

src/layer/marker/Marker.js
```diff
@@ -225,6 +225,7 @@ export var Marker = Layer.extend({
 
 		if (options.keyboard) {
 			icon.tabIndex = '0';
+			icon.setAttribute('role', 'button');
```
mrlessc, line 228: Here

mrlessc, line 228: Canada

src/layer/marker/Marker.js
Falke-Design: What is here?

src/layer/marker/Marker.js
jonkoops: :canada:

Issue #7537, opened 2021-03-29
Should Leaflet have an Accessibility Statement?


Malvoz:
Originally suggested in <a href="https://github.com/Leaflet/Leaflet/issues/3210#issue-2.3">#3210 (comment)</a>.

Inaccessibility may be a legal liability under certain [international laws and policies](https://www.w3.org/WAI/policies/). End-users of Leaflet maps have the right to an accessible experience, and implementers (e.g. organizations, individual developers) need to know if they meet those requirements. Documentation should be provided, describing if - and to what extent - Leaflet is accessible.

The <span title="World Wide Web Consortium">W3C</span> Web Accessibility Initiative (WAI) provides a [tool to generate](https://www.w3.org/WAI/planning/statements/generator/#create) an [accessibility statement](https://www.w3.org/WAI/planning/statements/) that could be published to the Leaflet website (reachable through the Leaflet site link in the attribution).

Malvoz:
Please label this issue "docs" as well.

Issue #7193, opened 2020-07-01
Make the leaflet container a programmatically identifiable region


Malvoz:
It is easy to determine that a Leaflet map is an interactive map that is embedded on a web page when we see it. This may not be the case for visually impaired users who rely on their screen reader. Therefore I propose making the `.leaflet-container` a programmatically identifiable region by 1) adding a landmark role ([`role="region"`](https://www.w3.org/TR/wai-aria/#region)), and 2) adding an [accessible name](https://www.w3.org/TR/accname-1.1/#dfn-accessible-name) that describes the purpose of the element (e.g. `aria-label="Interactive map"`).

[Landmark roles](https://www.w3.org/TR/wai-aria/#landmark_roles) are special types of roles in that they allow assistive technologies to include the landmarks in a page summary, and also enables landmark-based keyboard navigation.

This is essentially implementing the WCAG 2.1 [technique ARIA11](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11) which satisfies Success Criterion [1.3.1: Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships), [1.3.6: Identify Purpose](https://www.w3.org/WAI/WCAG21/Understanding/identify-purpose) and [2.4.1: Bypass Blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks) for the leaflet container.

Related resources:
- https://www.scottohara.me/blog/2018/03/03/landmarks.html
- https://www.w3.org/TR/wai-aria-practices/examples/landmarks/
- https://www.w3.org/TR/wai-aria-practices/examples/landmarks/region.html

Falke-Design:
I would say `aria-label="Map"` would be make sense but like I understand it, with defining `role='region'` we would say that the Leaflet map is one of th main parts of the website, but this is not everytime correct. Leaflet is also often used to show an address in a contact form like [here](https://preview.colorlib.com/#meetup). And sometimes the map is not accessible and then it would be wrong to say it is `role='region'`. I think this needs to be done by the developers self.

Malvoz:
I think independently of whether the map is part of a form or other semantic element an appropriate default `role` would be useful to provide additional context. Currently the components of the map are seemingly not part of any particular structure (other than the root document, of course), see the current Accessibility Tree:

![zoom-button-context](https://user-images.githubusercontent.com/26493779/142779754-2b9990aa-db65-4291-b05d-182130711824.png)

[...]

PR #7900, opened 2022-01-10
Add an official Accessibility Statement

Malvoz:
Resolves https://github.com/Leaflet/Leaflet/issues/7537.

The accessibility statement, including _all_ headings and most paragraphs, were [generated](https://www.w3.org/WAI/planning/statements/generator/#create) from input. Though many of the paragraphs were modified, mostly to fit appropriate links in (the template wanted a list of known limitations a.k.a. accessibility issues, I chose to link to Github instead).

Adding this proposed accessibility statement has the implications:
- [x] Leaflet is officially concluded to **partially** support WCAG 2.1 [Level A](https://www.w3.org/TR/WCAG21/#cc1) - this is because Leaflet fails at least one Level A Success Criterion (and probably more). Naturally, the immediate goal would be to fully support Level A and then reflect that in the statement.
- [x] Developers and users (coming from the attribution link in the maps, for example) can find official information on accessibility in Leaflet.
- [ ] There should probably be a link either in the header or footer to the accessibility statement.
- [ ] Should most likely scope the Accessibility Statement to Leaflet _maps_ (i.e. excluding the website, we don't want to block WCAG Level conformance on site issues, although site issues should also be reported by users). TBD how to express this.
- [ ] Probably rename this file `accessibility-statement` for ambiguity now that https://leafletjs.com/examples/accessibility/ exists.

I'm hoping this is all that is required to add a new document. <s>I don't have Ruby installed.</s> 😛

/cc @mourner, @IvanSanchez, @johnd0e, @jonkoops, @Falke-Design

### DIFF
--- a/docs/accessibility.md
+++ b/docs/accessibility.md
@@ -1,0 +1,80 @@
+---
+layout: v2
+title: Accessibility
+bodyclass: a11y-statement-page
+---
+
+## Accessibility Statement for Leaflet
+
+This is an accessibility statement from Leaflet.
+
+<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="50px" viewBox="0 0 24 24" width="50px" style="fill:#199900;margin:-.75em 0 -1.5em 0;"><path d="M0 0h24v24H0z" fill="none"></path><path d="M20.5 6c-2.61.7-5.67 1-8.5 1s-5.89-.3-8.5-1L3 8c1.86.5 4 .83 6 1v13h2v-6h2v6h2V9c2-.17 4.14-.5 6-1l-.5-2zM12 6c1.1 0 2-.9 2-2s-.9-2-2-2-2 .9-2 2 .9 2 2 2z"></path></svg>
+
+### Conformance status
+
+The
+[Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/WAI/standards-guidelines/wcag/)
+defines requirements for designers and developers to improve accessibility
+for people with disabilities.
+It defines three levels of conformance:
+Level A, Level AA, and Level AAA.
+
+Leaflet is partially conformant with WCAG 2.1
+[Level A](https://www.w3.org/TR/WCAG21/#cc1).
+Partially conformant means that some parts of the content do not fully conform
+to the accessibility standard.
+
+### Measures to support accessibility
+
+Leaflet strives to provide accessible maps.
+[Accessibility issues to be addressed](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility) are tracked and discussed
+on Github.
+
+### Feedback
+
+We welcome your feedback on the accessibility of Leaflet.
+If you encounter accessibility barriers, please check to see if there's a
+relevant bug or
+[file a new bug report](https://github.com/Leaflet/Leaflet/issues/new/choose).
+Or tweet
+[@LeafletJS](https://twitter.com/LeafletJS).
+
+
+### Compatibility with browsers and assistive technology
+
+Leaflet is designed to be compatible with the following assistive technologies:
+
+- Screen readers (partial compatibility)
+
+### Technical specifications
+
+Accessibility of Leaflet relies on the following technologies to work with
+the particular combination of web browser and any assistive technologies or
+plugins installed on your computer:
+
+- ARIA
+- CSS
+- HTML
+- JavaScript
+
+These technologies are relied upon for conformance with the accessibility
+standards used.
+
+### Limitations
+
+Despite our best efforts to ensure accessibility of Leaflet,
+there may be some limitations.
+
+Known limitations for Leaflet exist as
+[Github issues with the "accessibility" label](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility).
+
+### Evaluation report
+
+An evaluation report for Leaflet JS is available at:
+
+- [Web map tools WCAG 2.1 evaluation](https://github.com/Malvoz/web-maps-wcag-evaluation/blob/master/README.md)
+
+---
+
+This statement was created on 10 January 2022 using the
+[W3C Accessibility Statement Generator Tool](https://www.w3.org/WAI/planning/statements/).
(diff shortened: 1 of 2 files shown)


docs/accessibility.md
```diff
@@ -0,0 +1,80 @@
...
+there may be some limitations.
+
+Known limitations for Leaflet exist as
+[Github issues with the "accessibility" label](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3Aaccessibility).
+
+### Evaluation report
+
+An evaluation report for Leaflet JS is available at:
+
+- [Web map tools WCAG 2.1 evaluation](https://github.com/Malvoz/web-maps-wcag-evaluation/blob/master/README.md)
```
Malvoz, line 75: I'm the author of the linked Evaluation report. The template asked for an evaluation report, it even had an additional "Other evidence" heading which I excluded.

I'm fine with removing this, but if you want to keep it I'm fine with that too.

docs/accessibility.md
```diff
@@ -0,0 +1,80 @@
...
+layout: v2
+title: Accessibility
+bodyclass: a11y-statement-page
+---
+
+## Accessibility Statement for Leaflet
+
+This is an accessibility statement from Leaflet.
+
+<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="50px" viewBox="0 0 24 24" width="50px" style="fill:#199900;margin:-.75em 0 -1.5em 0;"><path d="M0 0h24v24H0z" fill="none"></path><path d="M20.5 6c-2.61.7-5.67 1-8.5 1s-5.89-.3-8.5-1L3 8c1.86.5 4 .83 6 1v13h2v-6h2v6h2V9c2-.17 4.14-.5 6-1l-.5-2zM12 6c1.1 0 2-.9 2-2s-.9-2-2-2-2 .9-2 2 .9 2 2 2z"></path></svg>
```
Malvoz, line 11: This icon is https://fonts.google.com/icons?selected=Material+Icons&icon.query=accessibility+new. Let me know if I should remove it.

Malvoz:
Also noting that the proposed accessibility statement lies somewhere between the [Example of Minimal Accessibility Statement](https://www.w3.org/WAI/planning/statements/minimal-example/) and [Example of Complete Accessibility Statement](https://www.w3.org/WAI/planning/statements/complete-example/) in the amount of content, since any input to the generator tool is optional.

docs/accessibility.md
Malvoz: (The linked 'accessibility' Github issues should be sufficient in providing any form of "evidence" or evaluation.)

Malvoz:
> - Leaflet is officially concluded to partially support WCAG 2.1 Level A - this is because Leaflet fails at least one Level A Success Criterion

The "fails at least one Level A" SC refers to https://github.com/Leaflet/Leaflet/issues/7193, which is to address a failure of [Success Criterion 1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).

docs/accessibility.md
Falke-Design: In the Github `md-renderer` the icon is not showing, does it show on the website?

docs/accessibility.md
```diff
@@ -0,0 +1,80 @@
...
+on Github.
+
+### Feedback
+
+We welcome your feedback on the accessibility of Leaflet.
+If you encounter accessibility barriers, please check to see if there's a
+relevant bug or
+[file a new bug report](https://github.com/Leaflet/Leaflet/issues/new/choose).
+Or tweet
+[@LeafletJS](https://twitter.com/LeafletJS).
```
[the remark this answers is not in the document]
lordt4ever: Nice Sr..

docs/accessibility.md
Malvoz: I finally came around to install Ruby. Yes, it shows:

a11y-statement

Maybe it should be a bit oversized to be a bit more impactful? Perhaps some added opacity and different positioning? 

oversized-icon

I don't feel strongly about it, it can be removed. But it's one of the [icons that symbolizes accessibility](https://fonts.google.com/icons?selected=Material+Icons&icon.query=accessibility), and makes the doc less dull?

Falke-Design:
Where should this linked? As a own menu point or as a link in the Documentation or on Github?

docs/accessibility.md
Falke-Design: I like the second attempt. Maybe the Icon can be behinde the text. Anyway this needs to work on mobile devices too

Malvoz:
> Where should this linked? As a own menu point or as a link in the Documentation or on Github?

That's a good question. I'm not sure it can be justified as a menu link:



(or using the commonly used "a11y" numeronym which is shorter:)



Perhaps a link in the footer? I think it should be reachable directly from the site at least.

docs/accessibility.md
Malvoz: Improved styling a bit (IMO) to blend in more with the page in https://github.com/Leaflet/Leaflet/pull/7900/commits/d57a66c8f508a57f486a836840ab2f5db0f4db94.

a11y-icon

It'd require more CSS to make it look good on mobile, also not exactly sure what that'd be, so hiding the icon on smaller screens, for now at least.

docs/accessibility.md
```diff
@@ -0,0 +1,80 @@
...
+
+The
+[Web Content Accessibility Guidelines (WCAG)](https://www.w3.org/WAI/standards-guidelines/wcag/)
+defines requirements for designers and developers to improve accessibility
+for people with disabilities.
+It defines three levels of conformance:
+Level A, Level AA, and Level AAA.
+
+Leaflet is partially conformant with WCAG 2.1
+[Level A](https://www.w3.org/TR/WCAG21/#cc1).
```
Malvoz, line 23: This statement if the most important one, and is the main reason (https://github.com/Leaflet/Leaflet/issues/7537) for the accessibility statement.

Thinking out loud:

But it'd be nearly impossible to confidently conclude anything above Level A _for Leaflet as a whole_ at any given point. This is because there are many features that would have to be reviewed, and even if every feature of Leaflet passes Level AA (for example) in isolation, when used together they may fail an SC.

Perhaps - at least in the long run - it'd be better to have a table to state e.g.:

- Basic Leaflet map (no added features) passes Level X.
- Feature A passes level X.
- Feature B passes level X.
- etc.

And since any feature can change over time, they'd have to be re-reviewed, for each version of Leaflet? That's not really feasible...

Falke-Design:
@Malvoz maybe you have time to make this PR ready for review

Issue #7811, opened 2021-11-25
Add a bundle size check in GitHub Actions


mourner:
It would be nice to add a check to CI that shows how many bytes a PR adds to the JS bundle. Preferably as a text in GH checks rather than a comment bot. cc @jonkoops 

```
leaflet.js: 142kB (+35), 40kB gzipped (+15)
```

mathe42:
https://github.com/siddharthkp/bundlesize#build-status-for-github would be a option for that. Or write a simple script for that.

If interessted I would add it to the workflow. No matter in wich way. Custom script is not that difficult would create it in a week or so.


As I'm a big fan of treeshaking I would propose to add a size check for a minimal Map-Example (just Map with 1 Marker) so we can keep an eye on that too. If that is a thing you are interessted in I can add this too.

Falke-Design:
@mathe42 can you please setup the suggested changes?

PR #7904, opened 2022-01-12
(feat) add bundlesize check

mathe42:
WIP for #7811

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        command: [lint]
+        command: [lint, bundlesize]
     steps:
       - name: Restore setup
         uses: actions/cache@v2
@@ -91,8 +91,6 @@ jobs:
       - name: Run tests on ${{ matrix.browser }}
         run: npm test -- --browsers ${{ matrix.browser }}
       
-      - name: Run bundlesize
-        run: bundlesize
 
   publish-artifacts:
     needs: setup
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        command: [lint, bundlesize]
+        command: [lint]
     steps:
       - name: Restore setup
         uses: actions/cache@v2
@@ -90,6 +90,9 @@ jobs:
 
       - name: Run tests on ${{ matrix.browser }}
         run: npm test -- --browsers ${{ matrix.browser }}
+      
+      - name: Run bundlesize
+        run: bundlesize
 
   publish-artifacts:
     needs: setup
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,13 @@ jobs:
 
       - name: Run ${{ matrix.command }} task
         run: npm run ${{ matrix.command }}
+        env:
+          CI_REPO_OWNER: Leaflet
+          CI_REPO_NAME: Leaflet
+          CI_COMMIT_MESSAGE: N/A
+          CI_COMMIT_SHA: ${{ github.sha }}
+          # BUNDLESIZE_GITHUB_TOKEN 
+          
 
   test:
     needs: setup
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,12 +62,7 @@ jobs:
       - name: Run ${{ matrix.command }} task
         run: npm run ${{ matrix.command }}
         env:
-          CI_REPO_OWNER: Leaflet
-          CI_REPO_NAME: Leaflet
-          CI_COMMIT_MESSAGE: N/A
-          CI_COMMIT_SHA: ${{ github.sha }}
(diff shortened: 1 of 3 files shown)


mathe42:
@mourner @Falke-Design could one of you approve to run the workflows?

Falke-Design:
Sorry I don't have the permission.

@IvanSanchez @johnd0e @jonkoops

package.json
```diff
@@ -5,6 +5,7 @@
   "description": "JavaScript library for mobile-friendly interactive maps",
   "devDependencies": {
     "@rollup/plugin-json": "^4.1.0",
+    "bundlesize": "0.18.1",
```
jonkoops, line 8: Can you use a version range here instead of an exact version? We're using a lockfile, which takes care of locking the versions for us.

jonkoops (changes requested):

.github/workflows/main.yml
```diff
@@ -90,6 +90,9 @@ jobs:
 
       - name: Run tests on ${{ matrix.browser }}
         run: npm test -- --browsers ${{ matrix.browser }}
+      
+      - name: Run bundlesize
```
jonkoops, line 94: Adding this task to the `test` job will run the same thing on every test target. It might be better to put this `run` job, which can run the specified NPM scripts.

jonkoops (changes requested):

jonkoops:
I'd also love to see the [build status for Github](https://github.com/siddharthkp/bundlesize#build-status-for-github) so it's easy to spot the file size difference compared to `master`. @mourner WDYT, would you mind setting those environment variables up?

mathe42:
> I'd also love to see the [build status for Github](https://github.com/siddharthkp/bundlesize#build-status-for-github) so it's easy to spot the file size difference compared to `master`. @mourner WDYT, would you mind setting those environment variables up?

I set it up. Only BUNDLESIZE_GITHUB_TOKEN has to be set like in the README.

.github/workflows/main.yml
```diff
@@ -90,6 +97,7 @@ jobs:
 
       - name: Run tests on ${{ matrix.browser }}
         run: npm test -- --browsers ${{ matrix.browser }}
+
```
jonkoops, line 100: Some stray whitespace here that can be removed.
```suggestion
```

.github/workflows/main.yml
```diff
@@ -61,6 +61,13 @@ jobs:
 
       - name: Run ${{ matrix.command }} task
         run: npm run ${{ matrix.command }}
+        env:
+          CI_REPO_OWNER: Leaflet
+          CI_REPO_NAME: Leaflet
+          CI_COMMIT_MESSAGE: N/A
+          CI_COMMIT_SHA: ${{ github.sha }}
+          # BUNDLESIZE_GITHUB_TOKEN
```
jonkoops, line 69: What is the purpose of the commit related environment variables? Are we expecting that this tool will be making commits on the repo?

jonkoops (changes requested):

.github/workflows/main.yml
mathe42: I'm not shure with CI_COMMIT_MESSAGE but as this is CI undependent it has to get the information where to push the status. Therefore it needs Owner, Name and the SHA for wich commit to attach the status. Commits to the repo would be bad.

mathe42:
And I changed the bundlesite settings to a max size of 42KB.

.github/workflows/main.yml
mathe42: And I checked the link to generate the token only allowes write access to the status not any commit rights.

mathe42:
When the access token is created and setup this should report the status back to the PR.

FYI:

I will add a extra PR with a check for a minimal example (as mentioned in #7811) so that we can detect if a PR effects treeshaking.

mathe42:
Nevermind the CI env vars are not needed as it is handled by https://www.npmjs.com/package/ci-env . Will remove them.

mathe42:
Only required step is to set a secret for this repo with name BUNDLESIZE and value from https://github.com/siddharthkp/bundlesize#build-status-for-github

Falke-Design:
I don't understand much from workflows but is it possible (or maybe it is already so) that the bundlesize check runs always and needs no "approve" on frist contributions. And that it is still executing if tests are failing.

@jonkoops what do you think about this?

mathe42:
The thing is this:

> Anyone can fork a public repository, and then submit a pull request that proposes changes to the repository's GitHub Actions workflows. Although workflows from forks do not have access to sensitive data such as secrets, they can be an annoyance for maintainers if they are modified for abusive purposes.
> 
> To help prevent this, workflows on pull requests to public repositories from some outside contributors will not run automatically, and might need to be approved first. By default, all first-time contributors require approval to run workflows.

So if they run automaticly this is a secutity risk as any change to the repro could result in the secrets beeing transfered to a webserver to collect them etc.

See https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks for more info.

jonkoops:
I opened up another PR for this with a comparison of what I found between this and other solutions. For more info see #7905

PR #7905, opened 2022-01-13
Add bundlemon to watch bundle size

### LINKED ISSUE #7811: Add a bundle size check in GitHub Actions

jonkoops:
Adds [`bundlemon`](https://github.com/LironEr/bundlemon) to the CI allowing us to see the changes in bundle size between a PR and `master`. Closes #7811.

I ended up testing a couple of different tool to monitor changes in bundle size which had to meet several requirements:

1. No required file limits, as we just want to see the diff between a PR and `master`.
2. Show the uncompressed diff as well as the gzipped one.
3. Does not use modules dependent on `node-gyp`.
4. Actively maintained.

I've tested the following packages all which were missing something.

## [Bundlemon](https://github.com/LironEr/bundlemon)
:heavy_check_mark: Checks all the boxes.

## [Bundlesize](https://github.com/siddharthkp/bundlesize)
1. :x: Not supported.
2. :x: Not supported.
3. :x: Depends on [`iltorb`](https://www.npmjs.com/package/iltorb) which needs `node-gyp` and is deprecated.
4. :x: Not actively maintained, author started a new library (see below).

## [Bundlesize 2](https://github.com/siddharthkp/bundlesize2)
1. :x: Not supported.
3. :heavy_check_mark:  Supported.
3. :heavy_check_mark:  No dependencies on `node-gyp`.
4. :heavy_check_mark:  Actively maintained. (but still alpha and has some bugs)

## [BundleWatch](https://github.com/bundlewatch/bundlewatch)
1. :x: Not supported.
2. :x: Not supported.
3. :x: Depends on [`iltorb`](https://www.npmjs.com/package/iltorb) which needs `node-gyp` and is deprecated.
4. :x: Not actively maintained, no actual commits since a while and breaking [bugs](https://github.com/bundlewatch/bundlewatch/issues/404).

@mourner You'll have to [authorize](https://github.com/LironEr/bundlemon-github-actions#add-github-integration) Bundlemon for the Github integration. I've sent a request to do so, or you can enable it yourself.

I also have work in progress branch for Bundlesize 2 [here](https://github.com/jonkoops/Leaflet/tree/bundlesize2).

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -1,0 +1,18 @@
+{
+  "baseDir": "dist",
+  "files": [
+    {
+      "path": "leaflet.js"
+    }
+  ],
+  "reportOutput": [
+    [
+      "github",
+      {
+        "checkRun": true,
+        "commitStatus": true,
+        "prComment": false
+      }
+    ]
+  ]
+}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,7 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        command: [lint]
+        command: [lint, bundlemon]
     steps:
       - name: Restore setup
         uses: actions/cache@v2
@@ -61,6 +61,9 @@ jobs:
 
       - name: Run ${{ matrix.command }} task
         run: npm run ${{ matrix.command }}
+        env:
+          BUNDLEMON_PROJECT_ID: 61e0545915f6c3000980d0ed
+          CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
   test:
     needs: setup
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "BSD-2-Clause",
       "devDependencies": {
         "@rollup/plugin-json": "^4.1.0",
+        "bundlemon": "^1.3.0",
         "eslint": "^8.1.0",
         "eslint-config-mourner": "^2.0.3",
         "eslint-plugin-script-tags": "^0.5.0",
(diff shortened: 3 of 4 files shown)


mourner:
@jonkoops great! I was waiting for your summary before authorizing :) Done now — let's try running the job again.

Falke-Design:
@jonkoops why is every task runned twice?
grafik

jonkoops:
@Falke-Design this is due to the way we have Github actions [set up](https://github.com/Leaflet/Leaflet/blob/master/.github/workflows/main.yml#L2), it will run on every push of the branch and on a pull request. I believe since I am a collaborator the 'push' event is also triggered.

mourner:
@jonkoops do you think we should do it too? With the current set up, I had cases like long-term feature development where PRs target a feature branch, or developing an experimental branch without creating a PR yet. But maybe that's not as essential, or we could choose something in between (e.g. PRs always, push only on main).

mourner:
> Now we're on the subject anyways, how do we feel about renaming `master` to `main`?

We should do it 👍 Anyway, let's ticket any related follow up separately. This PR looks good to me.

mourner (approved):
Needs a package lock update to resolve conflicts, otherwise 👍

mourner:
@jonkoops Do you know why the checks don't compare against base versions, and instead always show absolute sizes on recent commits? Looks like we didn't fully configure this.

jonkoops:
Yeah I am seeing the same, it should show the relative size on PRs only.

jonkoops:
@LironEr Thanks for the info, I came to the same conclusion. There are a couple of other things I need to fix as well that I will put in a PR together with this. For example we're now comparing the development build against a production build, which will always report a different size.

Could you tell us why on the `main` branch the reported file size increase is always the same as the whole file size? Instead of a relative number compared to the previous commit.

LironEr:
> See #7934 for the fixes that I mentioned.

I think u can keep bundlemon in the `main.yml` file, instead of creating a new workflow.

> Could you tell us why on the main branch the reported file size increase is always the same as the whole file size? Instead of a relative number compared to the previous commit.

Currently, a comparison to the previous commit is not implemented, only when a base branch exists (PR).
But I will fix it during the weekend

jonkoops:
Great, I just go the changes landed in #7934

LironEr:
> Currently, a comparison to the previous commit is not implemented, only when a base branch exists (PR).
But I will fix it during the weekend

Fixed in https://github.com/LironEr/bundlemon/pull/83

You don't need to upgrade anything, the fix was on the server side

PR #7901, opened 2022-01-10
Improve site accesibility

MxDui:
Tasks completed as described in #7899 and I also formatted the  docs/docs/css/main.css file with prettier

PR #7906, opened 2022-01-13
Fix issues in the layout

Falke-Design:
There are some issues in the layout because of #7901

- "Trusted by" not filling the whole page
- Footer is not in the center

"Trusted by"
grafik

----------

Footer:
grafik

@mourner @jonkoops this is a important PR which should merged fast ⚠

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -34,7 +34,7 @@ or head straight to the <a href="reference.html">API documentation</a>.
 If you have any questions, take a look at the <a href="https://github.com/Leaflet/Leaflet/blob/master/FAQ.md">FAQ</a> first.</p>
 
 
-</div>
+</main>
 
 <h2 class="usedby-title">Trusted by the best</h2>
 <div class="usedby">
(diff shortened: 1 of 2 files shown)


docs/index.html
```diff
@@ -34,7 +34,7 @@
 If you have any questions, take a look at the <a href="https://github.com/Leaflet/Leaflet/blob/master/FAQ.md">FAQ</a> first.</p>
 
 
-</div>
+</main>
```
Malvoz, line 37: So the problem is that `class="container"` now wraps "Trusted by" per https://github.com/Leaflet/Leaflet/pull/7901/.

`<main>` should only be used once per document though, so another solution could be this structure:

```html
<!-- docs/_layouts/v2.html -->
<main>
  <div class="container">
    {{ content }}
  </div>
</main>
```

and in docs/index.html change these `<main>` back to `<div>` for `class="container"`. Does that make sense @Falke-Design?

Falke-Design: Thx changed

mourner (approved):
Thanks!

Issue #7549, opened 2021-04-10
Ensure accessible target size for commonly used controls


Malvoz:
The <span title="Web Content Accessibility Guidelines">WCAG</span> 2.1 <span title="Success Criterion">SC</span> [2.5.5 Target Size](https://www.w3.org/TR/WCAG21/#target-size) requires targets for [_pointer input_](https://www.w3.org/TR/WCAG21/#dfn-pointer-inputs) to be at least **44 by 44 in CSS pixels**. Meeting this requirement has [multiple benefits](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#benefits), it'd be great to pass this criteria for the commonly used controls:

- General controls:
  https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L284-L286<br>
- Markers:
  https://github.com/Leaflet/Leaflet/blob/bc918d4bdc2ba189807bc207c77080fb41ecc196/src/layer/marker/Icon.Default.js#L26
  (Changing the marker's `width` and `height` to `44px` would cause it to stretch. Either the image would need to be modified or `padding` (right and left) could be modified to achieve a 44 by 44 px target.)<br><br>
- Popup's Close button:
  https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L517-L518
  (Here too `padding` can be utilized if keeping a visually small icon is desired.)<br><br>

Increasing the default size of controls would also have the benefit of not needing extra CSS for conditional sizing (presumably, since the controls would be large enough by default). For example, the following lines could be removed:

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L316-L320

https://github.com/Leaflet/Leaflet/blob/0f904a515879fcd08f69b7f51799ee7f18f23fd8/dist/leaflet.css#L359-L362

Malvoz:
In any case 44 by 44 px cannot be achieved for these targets (for example, for markers it can get tricky if increasing the size of the image itself is not desired, but setting a `padding` is also not desired as it may cause markers to be unexpectedly interacted with since they're larger than percieved) it should be added to as a known limitation in the [accessibility statement](https://github.com/Leaflet/Leaflet/issues/7537).

PR #7908, opened 2022-01-14
Small enhancement to the popup's close button

Malvoz:
(The Yada yada section for background  - see screenshots below for tl;dr)

This relates to https://github.com/Leaflet/Leaflet/issues/7549. That issue points to the WCAG 2.1 Success Criterion (SC) for Target Size, and requires a Target Size of 44x44 px which I think may be a big leap to take, at least for now. The good news is that WCAG 2.2 (which notably wont reach Recommendation status until June IIRC) instead has 2 SCs:

- Target Size (Minimum) which requires target size of 24x24
- Target Size (Enhanced) which requires target size of 44x44

This PR change makes it so that the popup close buttons (will) pass WCAG 2.2 [SC 2.5.8 Target Size (Minimum)](https://www.w3.org/TR/WCAG22/#target-size-minimum). This PR also fixes a small visual bug with the outline for keyboard users.

### Proposed

<ins>See the latest changes in https://github.com/Leaflet/Leaflet/pull/7908#issuecomment-1012696490!</ins>


Malvoz:
@IvanSanchez In this case I'm only setting the close button to 24x24px. In either case it seems there is indeed a need to adjust the margin on `leaflet-popup-content`:

https://github.com/Leaflet/Leaflet/blob/142ffdee3b87a506f4cc2084209ec16f373eb731/dist/leaflet.css#L476-L477

If that line was changed to `margin: 24px` in this case, we can be sure that there's no overlap, the result would look like:

24px-margin

same change, with some highlighting:

24px-margin-highlighted

so that does have a visual impact on the popup (of course, there's no need for `margin-bottom` to be 24px as well, but for symmetry that might be warranted?).

Other than that there's not much room for negative margins on the close button itself before it hugs the edge of the popup.

Malvoz:
Actually, all we have to worry about is the margin-right (and left for symmetry) right? It's getting late here...

so if we change:

https://github.com/Leaflet/Leaflet/blob/142ffdee3b87a506f4cc2084209ec16f373eb731/dist/leaflet.css#L476-L477

to `margin: 13px 24px` we'd get:

24px-margin-inline

same screenshot, with highlighting:

24px-margin-inline-highlighted

Malvoz:
Changed `leaflet-popup-content`'s margin from `13px 19px` to `13px 24px` in https://github.com/Leaflet/Leaflet/pull/7908/commits/542a54a066fa789ccf5fa62f0709dd040eafb9ca per https://github.com/Leaflet/Leaflet/pull/7908#issuecomment-1012696490 and Ivan's feedback.

mourner:
@Malvoz This is very nice! I'll take a second look a few days later to make sure the design aesthetic isn't compromised too much with the wide side margins — perhaps it would be OK to have a little overlap.

Off topic, but on your screenshots, there's a 1pm gap between the popup and its tip — I wonder if that's a bug we should fix too? Is it easily reproducible?

lordt4ever:
Tnks all

Malvoz:
@mourner https://github.com/Leaflet/Leaflet/pull/7908#issuecomment-1013529489:

> Off topic, but on your screenshots, there's a 1pm gap between the popup and its tip — I wonder if that's a bug we should fix too? Is it easily reproducible?

The gap is often appearing, but inconsistently so.

Even the popup in Chrome's dev tools has the same issue (compare their popup in [this](https://user-images.githubusercontent.com/26493779/149426796-bd7e386c-0781-4c38-9a85-8ad560d1ed4c.png) screenshot to [this](https://user-images.githubusercontent.com/26493779/149426960-9620a9f1-a5d4-46be-8954-b8e4a4e9e439.png)), I think the root cause is the same as the 1px gap bug between tiles, probably a Chromium compositing bug.

In the Maps4HTML `<mapml-viewer>` I fixed this by setting [`.leaflet-popup-tip-container { margin-top: -1px }`](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/b44bc709d27ca242c1297cf57c96e3635be08113/src/mapml.css#L585-L587). The downside of that fix is that if a developer changes the background-color of the popup they'll see a 1px dent, but in reality they'll probably want to change the background-color of the popup's tip to the same color anyway so I think it's fine. Rather have a robust looking default, IMO.

lordt4ever:
Way is gap me I am the gap

Malvoz:
Re @mourner's https://github.com/Leaflet/Leaflet/pull/7908#issuecomment-1013529489, I've opened a separate PR: https://github.com/Leaflet/Leaflet/pull/7920.

Malvoz:
It'd be great to have a decision for this before the final v1.8.0 release, now that outlines are showing on keyboard focus (https://github.com/Leaflet/Leaflet/pull/7259), since the current outline on the close button looks pretty bad.

Falke-Design:
I had currently not the time to check, but I will do!

As reminder, we should check if the close button is now displayed better when the popup is empty:
grafik

Falke-Design (approved):
@Malvoz I see no problems with merging this. The margins are not much bigger and the difference is not really visible.

mourner (approved):
I reduced the margin on the left, making the popup slightly asymmetrical, which I think looks better overall considering the close button already shifts symmetrical balance.

image

image

PR #7912, opened 2022-01-15
Change `<section>` to `<fieldset>` in the layers control

### LINKED ISSUE #7530: `leaflet-control-layers-list` should be conveyed as a group of related controls
## Motivation

The `leaflet-control-layers-list` is currently a `<section>` element (per https://github.com/Leaflet/Leaflet/pull/6380/). The `<section>` element only has [implicit ARIA semantics](https://www.w3.org/TR/html-aria/#el-section) when it has an [accessible name](https://www.w3.org/TR/accname-1.1/#dfn-accessible-name) (see [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names) for more info) - which it doesn't (and even if it did the implicit `region` role is not appropriate). It currently isn't any more useful to users of assistive tech than a `<div>`.

## Proposed solution

Replace `<section>` with [`<fieldset>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset):

https://github.com/Leaflet/Leaflet/blob/b52961aa469dc17b3710ef7a9c0f13b6f4dba550/src/control/Control.Layers.js#L184

## Additional context
- Relates to <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> <span title="Success Criteria">SC</span> [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships)
-  Together with https://github.com/Leaflet/Leaflet/issues/7529 the layers list would be conveyed as a group of related controls to screen reader users.

tmiaa:
As explained by issue #7530 the Section element is not so useful in the leaflet-control-layers-list. changed to Fieldset.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -366,6 +366,11 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	display: block;
 	position: relative;
 	}
+.leaflet-control-layers-list {
+	border: 0;
+	margin: 0;
+	padding: 0;
+	}
 .leaflet-control-layers-expanded {
 	padding: 6px 10px 6px 6px;
 	color: #333;
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -181,7 +181,7 @@ export var Layers = Control.extend({
 		DomEvent.disableClickPropagation(container);
 		DomEvent.disableScrollPropagation(container);
 
-		var section = this._section = DomUtil.create('section', className + '-list');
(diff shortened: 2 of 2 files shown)


Falke-Design:
This is a **Breaking change**, many user add custom html into the layer control by fetching the html structure.

Malvoz (approved):

jonkoops (approved):

Issue #2771, opened 2014-07-01
Hover over disabled radio/checkbox in Layer Control closes control


knownasilya:
Here's a [jsfiddle](http://jsfiddle.net/4T25B/1/) reproducing the issue.

Basically it closes the layer control if hovering over the disabled item.

mourner:
Fixed.

knownasilya:
Yay!

PR #7914, opened 2022-01-15
Fix HTML error in Layer Control, Fix #6434

### LINKED ISSUE #6434: AODA issue caused by Issue #2771
The addition of the extra DIV that prevents the layer control flicker when checkboxes are disabled causes [AODA ](https://www.aoda.ca/) issues because it's not semantic HTML.

**Steps to reproduce**
- Inspect the page with dev tools
- From the Elements tab of the dev tools, right click on the <html> element and select Copy > Copy element
- Navigate to https://validator.w3.org/nu/#textarea
- Retain the doctype in the sample HTML in the text input; replace the rest of the sample HTML with your copied HTML from your page
- Click Check

**Expected behavior**
The expected behavior is that the page produces no errors

**Current behavior**
There are 6 errors related to the extra DIV placed in a LABEL as a fix for Issue #2771

**Environment**
- Leaflet version: 1.3.1

tmiaa:
Fix #6434: "Error: Element div not allowed as child of element label" message from HTML validator

### DIFF
--- a/spec/suites/control/Control.LayersSpec.js
+++ b/spec/suites/control/Control.LayersSpec.js
@@ -297,7 +297,7 @@ describe("Control.Layers", function () {
 				'Marker A': markerA
 			}).addTo(map);
 
-			var elems = map.getContainer().querySelectorAll('div.leaflet-control-layers label span');
+			var elems = map.getContainer().querySelectorAll('div.leaflet-control-layers label span span');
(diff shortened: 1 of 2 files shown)


mourner (approved):
Thanks for the fix, this looks good! Did you make sure #2771 is not reproducible with it?

Falke-Design (approved):
Checked. #2771 is not reproducible

@mourner @IvanSanchez ready to merge

PR #7567, opened 2021-05-07
Fix exception Invalid LatLng object: (NaN, NaN) with the firefox debu…

SuperPat45:
…gger in the mobile simluator mode

Fix a problem with the Firefox debugger in simulated mobile mode who pointerdown event received have two touches with the same coordinate which leaflet does not handle correctly.

Fix bug #7566

PR #6935, opened 2019-12-10
add 'topcenter', 'bottomcenter' control positions

vulkd:
Similar functionality has already been written / discussed in possibly stale PRs such as #4518, #5554, and #5264.

An existing plugin (https://github.com/FCOO/leaflet-control-topcenter) implements near-identical functionality but uses flexbox (reducing IE compatibility).

Example code:

```html
<link rel='stylesheet' href='./Leaflet/dist/leaflet.css' />
<script src='./Leaflet/dist/leaflet-src.js'></script>
<div id='Map' style="height: 50vh"></div>
<script>
	const map = L.map('Map', { zoom: 4, center: [0, 0] });

	const baseLayer = L.tileLayer();

	L.control.layers({a: baseLayer, b: baseLayer, c: baseLayer}, null, {
		collapsed: false,
		position: 'bottomcenter'
	}).addTo(map);

	L.control.layers({a: baseLayer, b: baseLayer, c: baseLayer}, null, {
		collapsed: false,
		position: 'topcenter'
	}).addTo(map);
</script>
```

PR #7921, opened 2022-01-19
Rename master branch to main

jonkoops:
As part of the broader initiative to use more inclusive language we are renaming the `master` branch to `main`.


.github/workflows/main.yml
```diff
@@ -113,7 +113,7 @@ jobs:
         id: artifacts-directory
         run: |
           VERSION=$(git tag --points-at HEAD)
-          echo "::set-output name=path::content/leaflet/${VERSION:-master}"
+          echo "::set-output name=path::content/leaflet/${VERSION:-main}"
```
jonkoops, line 116: I've also updated the path where previously the `master` version would be uploaded, it seems that all these references have been updated. It might however be necessary for this directory to be removed from the Amazon bucket and set up redirects (if possible).

If this is too much effort (or breaks links out in the wild) we might consider leaving this set to `master` and stick with just the branch rename for now.

mourner (approved):
👍 I also renamed the branch through GitHub (the latter makes it easy, it updated all PR targets too).

jonkoops:
FYI I have not renamed references to `master` in existing blog posts, and other historical documents (such as the changelog). It seems to me that we should not be trying to whitewash the history of the project in this regard. I have however updated the links to the main branch so they will remain functional.

.github/workflows/main.yml
mourner: Let's keep existing links working. New builds will go to updated path and that's good enough IMO.

docs/edit.html
```diff
@@ -28,9 +28,9 @@
             }
 
             if (config.version) {
-                if (config.version === 'master') {
-                    config.cssFile = 'https://leafletjs-cdn.s3.amazonaws.com/content/build/master/leaflet.css',
-                    config.jsFile = 'https://leafletjs-cdn.s3.amazonaws.com/content/build/master/leaflet-src.js'
+                if (config.version === 'main') {
```
jonkoops, line 31: I am not entirely sure where this version would be set, it seems this state is retrieved from the URL but I am not sure the reference to it has been updated.

mourner:
The new links are working: https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/main/leaflet.js

jonkoops:
@mourner Nice, be aware that the old links are also still working. Perhaps we should not remove them as there might be people relying on it to exist. However these files will slowly become outdated as we add more commits.

mourner:
@jonkoops that sounds fine to me. Those links are mostly for our team & debugging anyway, and the main purpose — providing a zip download of the bundles — is working.

Falke-Design:
@mourner have you closed #7567? Or is this automatically happened with deleting the master branch? There are many other PRs too

jonkoops:
I think this happened automatically. Strange since there are a lot of other PRs that simply changed their target branch.

jonkoops:
Ah, @mourner found the issue https://github.com/Leaflet/Leaflet/pull/6935#issuecomment-1016344317. If people deleted their fork it will close the PR, makes sense. Nothing we can do.

PR #7922, opened 2022-01-19
Narrow triggers for CI runs

jonkoops:
Narrows down the trigger for our CI so it will only run on PRs and pushes to `main`.

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -9,8 +9,8 @@
     [
       "github",
       {
-        "checkRun": false,
-        "commitStatus": true,
+        "checkRun": true,
+        "commitStatus": false,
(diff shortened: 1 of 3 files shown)


mourner:
@jonkoops 🤔 why does it still show double the bundle checks on the commit that narrows down triggers?

jonkoops:
@mourner Yeah I am confused about that as well. Perhaps an issue with the tooling itself. Let me try some things.

jonkoops:
Okays so the issue is the configuration, of only `checkRun` is enabled we get what we want. Unfortunately we are comparing a production build to a dev build which causes the file size to be misreported. I'll see if I can make a fix for this.

PR #7925, opened 2022-01-19
Add shiftkey option to scroll wheel zoom

Falke-Design:
For more infos look into: #3955

-----
# #3955:
This PR adds the possibility to set the map option "scrollWheelZoom" to enable scroll wheel zoom only when the shift key is pressed. Useful when you want to disable scroll wheel zoom but still make it possible to use the scroll wheel in some way. 

I'm not sure if this breaks any other mouse/keyboard events so it would be great if someone with better Leaflet knowledge could take a look at it.

### DIFF
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -25,6 +25,7 @@ L.Map.ScrollWheelZoom = L.Handler.extend({
 	},
 
 	_onWheelScroll: function (e) {
+		if (this._map.options.scrollWheelZoom === 'shiftkey' && !e.shiftKey) { return false; }
 		var delta = L.DomEvent.getWheelDelta(e);
 		var debounce = this._map.options.wheelDebounceTime;
 


src/map/handler/Map.ScrollWheelZoom.js
```diff
@@ -25,6 +25,7 @@ L.Map.ScrollWheelZoom = L.Handler.extend({
 	},
 
 	_onWheelScroll: function (e) {
+		if (this._map.options.scrollWheelZoom === 'shiftkey' && !e.shiftKey) { return false; }
```
IvanSanchez, line 28: I must ask why the `return false`. If my memory serves right, the return value of an event handler was used to stop event propagation... but that was many, many years ago. The current documentation says the return value of event handlers is ignored.

IvanSanchez (changes requested):
This PR is based on an incredibly outdated revision of `Map.ScrollWheelZoom.js` - more than six years old, even before the great Leafdoc and Rollup rewrites.

@Falke-Design Try to do a `git rebase main` after a `git branch shiftkey_scroll_wheel`. If rebasing gets messy, I recommend closing this PR and creating a new one. After all, it's a one-line change.

Besides that, this will need changes in the docstrings for the relevant map option.

src/map/handler/Map.ScrollWheelZoom.js
Falke-Design: Yes you are right, return values are ignored. For DomEvent and also for the Leaflet-Events.

Falke-Design:
I tried to implement the same for DoubleClick but there is already a usage of holding shift -> zooming out.

I didn't added test because they are generally missing. I will create a new PR for the test else we get merge conflicts.

IvanSanchez (approved):
LGTM.

Since this is a behaviour change, I'd like some other maintainer to give a heads-up.

I'm slightly concerned that `shiftkey` and `center` are mutually exclusive, but the alternative is adding more map options... and I'm not too comfortable about that.

Falke-Design:
> I'm slightly concerned that shiftkey and center are mutually exclusive, but the alternative is adding more map options... and I'm not too comfortable about that.

Good point ... what if we allow following `scrollWheelZoom: 'center,shiftkey'`

IvanSanchez:
> Good point ... what if we allow following `scrollWheelZoom: 'center,shiftkey'`

I'd rather `scrollWheelZoom: 'center', scrollWheelZoomKey: 'shift'`. And add support for `'control'`, `'alt'` and `'meta'` while at it.

Then again, one of the ideas in the back of my head is enabling scrollwheel zoom only when the map occupies less than a threshold percent of (e.g. "less than half of") the available screen space. So I'm unsure at how a clean API for this would look like.

Issue #2199, opened 2013-11-19
Clicking or using a keyboard to open popup doesn't set focus to popup


acbramley:
We have very strict accessibility guidelines that we must adhere to. Thankfully leaflet almost solved all our issues we had with Openlayers :) 

There are 2 issues that have been raised by our client's accessibility tester. One being that when a popup is opened (either with the keyboard or mouse) the focus is not set to the popup. This is most important for keyboard users as if a user tabs to a feature, presses enter to open the popup, then tries to get to the popup using the keyboard, they have to tab through ALL other features on the map until they hit the popup. Solution could be to set focus directly to the popup, or make it take the next tabindex?

IvanSanchez:
This has been open for almost two years. @acbramley, do you have any thoughts on the current state of things in 1.0.0-beta2?

acbramley:
Unfortunately I don't, I moved off the project that this work was for a long time ago, I can get in touch with some guys that are working on it still and find out if they still have the issue.

techieshark:
I ran across this issue while trying to figure out how to send focus to a popup for keyboard users for accessibility reasons.

Perhaps someone knows a better way to do this, but for now I just used a hack (getting at internal properties) like this:

```
var clickMarker = function (e) {
      // open popup
      var marker = e.target;
      marker.openPopup();

      // hack for accessibility - no good way to tell popup focus to shift ? - see
      // https://github.com/Leaflet/Leaflet/issues/2199
      var popupNode = e.target._popup._contentNode;
      popupNode.querySelector('a').focus();
};
```

skorasaurus:
I've come across this as well and made a test case: 
http://skorasaurus.github.io/maptemplates/leaflet-test-2199.html and http://bl.ocks.org/skorasaurus/9b2b96967335da91eac7676e815755d2
has the same code. 

how to replicate: 
visit the page 
tab through to a specific marker, press enter to open pop-up 

Expected behavior: 
focus is sent to the marker's pop-up, so use can press enter to access content (a hyperlink) in the pop-up. 

What happens: 
Focus goes to the markers that are east of the marker clicked. (If you open the marker further east, the expected behavior occurs). 

I've tried @techieshark 's workaround but that doesn't work; I haven't tried anything else yet.

techieshark:
[...]

PR #7927, opened 2022-01-19
Enhance Map accessibility

Falke-Design:
For more infos look into: #6248

-----
# #6248:
Changes:

    Replace anchor tags with buttons.
    Add Alt attribute to images.
    Add roles to elements.
    Make tooltip focusable.

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -159,7 +159,7 @@ describe("Marker", function () {
 			} else {
 				expect(marker._icon.src).to.contain('marker-icon.png');
 			}
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
 		});
 
 		it("changes an image to a DivIcon", function () {
@@ -173,7 +173,7 @@ describe("Marker", function () {
 			expect(oldIcon.parentNode).to.be(null);
 
 			expect(marker._icon.innerHTML).to.contain('Inner1Text');
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
 		});
 
 		it("reuses the icon/shadow when changing icon", function () {
@@ -187,7 +187,7 @@ describe("Marker", function () {
 			expect(oldIcon).to.be(marker._icon);
 			expect(oldShadow).to.be(marker._shadow);
 
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
 			expect(marker._shadow.parentNode).to.be(map._panes.shadowPane);
 		});
 
@@ -196,7 +196,7 @@ describe("Marker", function () {
 			map.addLayer(marker);
 			var icon = marker._icon;
 			expect(icon.hasAttribute('alt')).to.be(true);
-			expect(icon.alt).to.be('');
+			expect(icon.alt).to.be('Location');
 		});
 
 		it("doesn't set the alt attribute for DivIcons", function () {
--- a/src/control/Control.Zoom.js
+++ b/src/control/Control.Zoom.js
@@ -80,23 +80,18 @@ export var Zoom = Control.extend({
 	},
 
 	_createButton: function (html, title, className, container, fn) {
-		var link = DomUtil.create('a', className, container);
-		link.innerHTML = html;
-		link.href = '#';
-		link.title = title;
+		var button = DomUtil.create('button', className, container);
+		button.innerHTML = html;
+		button.title = title;
 
-		/*
-		 * Will force screen readers like VoiceOver to read this as "Zoom in - button"
-		 */
-		link.setAttribute('role', 'button');
-		link.setAttribute('aria-label', title);
+		button.setAttribute('aria-label', title);
 
-		DomEvent.disableClickPropagation(link);
-		DomEvent.on(link, 'click', DomEvent.stop);
-		DomEvent.on(link, 'click', fn, this);
-		DomEvent.on(link, 'click', this._refocusOnMap, this);
+		DomEvent.disableClickPropagation(button);
+		DomEvent.on(button, 'click', DomEvent.stop);
+		DomEvent.on(button, 'click', fn, this);
+		DomEvent.on(button, 'click', this._refocusOnMap, this);
 
-		return link;
+		return button;
 	},
 
 	_updateDisabled: function () {
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -179,7 +179,11 @@ export var Popup = DivOverlay.extend({
 			' leaflet-zoom-animated');
 
 		var wrapper = this._wrapper = DomUtil.create('div', prefix + '-content-wrapper', container);
-		this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+		var wrapperContent = this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+
+		wrapperContent.setAttribute('role', 'tooltip');
+		wrapperContent.setAttribute('id', 'popup-tooltip');
+		wrapperContent.parentElement.parentElement.setAttribute('tabIndex', '-1');
 
 		DomEvent.disableClickPropagation(container);
 		DomEvent.disableScrollPropagation(this._contentNode);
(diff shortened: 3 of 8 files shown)


Malvoz:
Should this be closed? I think it's doing a bit much for one PR. And it's changing `<a>` to `<button>` which is a breaking change that blocks (until v2?) some non-breaking changes that this PR also introduces?

Anyhow, there are unresolved comments (in the original PR https://github.com/Leaflet/Leaflet/pull/6248).

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -159,7 +159,7 @@ describe("Marker", function () {
 			} else {
 				expect(marker._icon.src).to.contain('marker-icon.png');
 			}
-			expect(marker._icon.parentNode).to.be(map._panes.markerPane);
+			expect(marker._icon.parentNode.parentNode).to.be(map._panes.markerPane);
```
Malvoz, line 162: > Parent of parent is an overkill.
> It's obvious that something is wrong in marker design.

_Originally posted by @johnd0e in https://github.com/Leaflet/Leaflet/pull/6248#r412835549_

src/control/Control.Zoom.js
```diff
@@ -80,23 +80,18 @@ export var Zoom = Control.extend({
...
-
-		return link;
+		var button = DomUtil.create('button', className, container);
+		button.innerHTML = html;
+		button.title = title;
+
+		button.setAttribute('aria-label', title);
+
+		DomEvent.disableClickPropagation(button);
+		DomEvent.on(button, 'click', DomEvent.stop);
```
Malvoz, line 90: > `DomEvent.stop` originally was needed to prevent default action of link.
> With button that is not needed.

_Originally posted by @johnd0e in https://github.com/Leaflet/Leaflet/pull/6248#r412837921_

src/layer/Popup.js
```diff
@@ -179,7 +179,11 @@ export var Popup = DivOverlay.extend({
 			' leaflet-zoom-animated');
 
 		var wrapper = this._wrapper = DomUtil.create('div', prefix + '-content-wrapper', container);
-		this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+		var wrapperContent = this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+
+		wrapperContent.setAttribute('role', 'tooltip');
```
Malvoz, line 184: > [`role="tooltip"`](https://www.w3.org/TR/wai-aria-1.1/#tooltip) doesn't seem appropriate for popups, rather [`role="dialog"`](https://www.w3.org/TR/wai-aria-1.1/#dialog).

_Originally posted by @Malvoz in https://github.com/Leaflet/Leaflet/pull/6248#r766351953_

src/layer/marker/Icon.js
```diff
@@ -136,9 +136,10 @@ export var Icon = Class.extend({
 		}
 	},
 
-	_createImg: function (src, el) {
+	_createImg: function (src, el, alt) {
 		el = el || document.createElement('img');
 		el.src = src;
+		el.alt = alt || '';
```
Malvoz, line 142: > In other part you use 'Location' string instead of empty string.

_Originally posted by @johnd0e in https://github.com/Leaflet/Leaflet/pull/6248#r412839207_

Malvoz: > It depends on the use case.
> 
> in this case, you create a new image tag so you need to add alt value or leave it with an empty string.

_Originally posted by @Muhnad in https://github.com/Leaflet/Leaflet/pull/6248#r414186918_

src/layer/marker/Marker.js
```diff
@@ -217,17 +221,14 @@ export var Marker = Layer.extend({
 			}
 
 			if (icon.tagName === 'IMG') {
-				icon.alt = options.alt || '';
+				icon.alt = options.alt || 'Location';
```
Malvoz, line 224: > Instead of hardcoding some arbitrary string you better change `options.alt` defaults.

_Originally posted by @johnd0e in https://github.com/Leaflet/Leaflet/pull/6248#r412840042_

Malvoz: > "location" seems opinionated, and may not make sense in a non-geographic map. How about "Marker"?

_Originally posted by @Malvoz in https://github.com/Leaflet/Leaflet/pull/6248#r563012140_

src/layer/Popup.js
```diff
@@ -189,8 +193,9 @@ export var Popup = DivOverlay.extend({
 		this._tip = DomUtil.create('div', prefix + '-tip', this._tipContainer);
 
 		if (this.options.closeButton) {
-			var closeButton = this._closeButton = DomUtil.create('a', prefix + '-close-button', container);
-			closeButton.href = '#close';
+			var closeButton = this._closeButton = DomUtil.create('button', prefix + '-close-button', container);
+
+			closeButton.setAttribute('aria-label', 'Close');
```
Malvoz, line 198: This was fixed in [#7794](https://github.com/Leaflet/Leaflet/pull/7794/files#diff-2b5b78f24babc48279b336b4e9d85623e3d1fd37c43fdfe5e7240e736d5fcf7aR194), FWIW.

src/layer/Popup.js
```diff
@@ -179,7 +179,11 @@ export var Popup = DivOverlay.extend({
 			' leaflet-zoom-animated');
 
 		var wrapper = this._wrapper = DomUtil.create('div', prefix + '-content-wrapper', container);
-		this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+		var wrapperContent = this._contentNode = DomUtil.create('div', prefix + '-content', wrapper);
+
+		wrapperContent.setAttribute('role', 'tooltip');
+		wrapperContent.setAttribute('id', 'popup-tooltip');
```
Malvoz, line 185: This is a fixed ID, but IDs in HTML must be unique.

src/layer/marker/Marker.js
```diff
@@ -202,9 +202,13 @@ export var Marker = Layer.extend({
 		var options = this.options,
 		    classToAdd = 'leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide');
 
+		var button = DomUtil.create('button', 'leaflet-marker-button');
 		var icon = options.icon.createIcon(this._icon),
 		    addIcon = false;
 
+		button.setAttribute('aria-haspopup', 'true');
```
Malvoz, line 209: Why is this needed? Note that [`aria-haspopup="true"`](https://www.w3.org/TR/wai-aria-1.1/#aria-haspopup) "_Indicates the popup is a [`menu`](https://www.w3.org/TR/wai-aria-1.1/#menu)_." (and the leaflet popups are not menus).

src/layer/marker/Marker.js
```diff
@@ -202,9 +202,13 @@ export var Marker = Layer.extend({
 		var options = this.options,
 		    classToAdd = 'leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide');
 
+		var button = DomUtil.create('button', 'leaflet-marker-button');
 		var icon = options.icon.createIcon(this._icon),
 		    addIcon = false;
 
+		button.setAttribute('aria-haspopup', 'true');
+		button.setAttribute('aria-describedby', 'popup-tooltip');
```
Malvoz, line 210: How is `aria-describedby` supposed to work here, if popups (why are popups called tooltips in this PR btw? Am I misunderstanding this PR?) are not visible to the user? In fact, popups are not even in the DOM for reference until they're already opened.

src/layer/marker/Marker.js
Malvoz: The default - as @johnd0e suggested - was change to `Marker` in [#7895](https://github.com/Leaflet/Leaflet/pull/7895/files#diff-9f4983a1c995bb3bbfe108613622e7270372359622822ba3c3cbbf3359bcc3f4R45).

src/layer/Popup.js
Malvoz: See issue https://github.com/Leaflet/Leaflet/issues/7968 for motivation for `role="dialog"`.

Falke-Design:
@Malvoz would it be ok for you, when you create new PRs with the useful suggestions from this PR and then close this one?

Malvoz:
> create new PRs with the useful suggestions from this PR and then close this one?

Most suggestions in this PR are specific to this PR. Though almost everything in here is proposed in separate issues:

> Replace anchor tags with buttons

Issue: https://github.com/Leaflet/Leaflet/issues/7821.

> Add Alt attribute to images

The new default was added in https://github.com/Leaflet/Leaflet/pull/7895.

> Add roles to elements

(This PR only adds 1 new role - `tooltip` - which I think is an incorrect role suggestion.)

Issue: https://github.com/Leaflet/Leaflet/issues/7968.

> Make tooltip focusable

(Again I think _popup_ is confused with _tooltip_ in this PR)

Issue: https://github.com/Leaflet/Leaflet/issues/2199.

---

I don't see this PR going anywhere since the original author hasn't responded in a long time, so let's close this.

PR #7928, opened 2022-01-19
Add `playsInline` attribute on the video element

### LINKED ISSUE #7357: Video overlay: `playsinline` attribute required for phones on iOS

Falke-Design:
For more infos look into: #7364

Fixes #7357


Malvoz:
re https://github.com/Leaflet/Leaflet/pull/7364#issuecomment-738780363:

> I'd rather not merge until it has been tested (and I don't have a iOS device either)

I think this should be good to merge. `playsinline` would simply allow videos to play without automatically going into fullscreen mode, in browsers where they otherwise would.

Without `playsinline`:
https://css-tricks.com/what-does-playsinline-mean-in-web-video/#:~:text=what%20it%20looks%20like%20without%20playsinline

With `playsinline`:
https://css-tricks.com/what-does-playsinline-mean-in-web-video/#:~:text=what%20it%20looks%20like%20with%20playsinline

Malvoz (approved):

Falke-Design:
We should setup a demo, so that we can easy share a link to check if it works. 
We can set it up in leafletjs.com/edit.html and then I can upload it to my server, because on phones the plunk doesn't work.

src/layer/VideoOverlay.js
```diff
@@ -78,6 +78,7 @@ export var VideoOverlay = ImageOverlay.extend({
 		vid.autoplay = !!this.options.autoplay;
 		vid.loop = !!this.options.loop;
 		vid.muted = !!this.options.muted;
+		vid.playsInline = true;
```
jonkoops, line 81: Let's make this an option like all the other ones, there might be use-cases where someone would actually want to have this play in full-screen.

jonkoops (changes requested):

Falke-Design:
Here a link to test on a mobile device: http://falke-design.bplaced.net/leaflet/video/videooverlay.html

jonkoops (approved):
LGTM

jonkoops:
@Falke-Design could you rebase this on `main`? I'd like to see if Bundlemon reports the file size increase properly.

Falke-Design:
@jonkoops does it work?

jonkoops:
Strangely enough it reports a smaller size, which seems impossible. I'll wait for some more PRs so we can see is this is a persistent problem.

PR #7933, opened 2022-01-20
added test for maxZoom isNaN

chasbani:
Added checks for isNaN conditional on maxZoom of layer.


Falke-Design:
Thank you for your contribution. 
If you want to create a Test please use the karma test suit. You need to add the test in the correct namespaced-file: https://github.com/Leaflet/Leaflet/tree/main/spec/suites

Does your PR close an issue or what is the purpose of this test?

PR #7945, opened 2022-01-26
Add referrerPolicy to TileLayer

### LINKED ISSUE #7943: Support a custom referrerPolicy option for TileLayer

We recently updated our web framework [to Django 3.2 in case others hit a similar issue] and our [MapBox] tile provider API key stopped working.

It looks like Django 3.1 [made a change](https://chipcullen.com/django-3-referrer-policy-change/) to now send a `Referrer-Policy: same-origin` header by default (on the HTTP response for the page containing our Leaflet map). So the referer header that used to be sent with the tile fetch requests is no longer sent and the tile provider (which ties API tokens to specific domains) rejects the requests.

I would like a new `referrerPolicy` option available on TileLayer. When set, the option's value would then be copied to the [`HTMLImageElement.referrerPolicy` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy) of each image node that particular TileLayer creates. E.g.

```
L.tileLayer(myUrlTemplate, {
  referrerPolicy: 'strict-origin-when-cross-origin'
}).addTo(map);
```

IIUC, this explicit referrer policy would override the document's default policy and the fetch requests would be accepted as they were before.

We could change back the Referrer-Policy sitewide, but ideally we wouldn't have to do that just for one feature of the site. I.e. whether Django was right-or-wrong to make this change, ideally we should be able to tighten up the *default* policy for our pages like Django did to us, and then override it for some of the requests which Leaflet makes.

**Additional context**

There is already a conceptually similar [`crossOrigin`](https://leafletjs.com/reference.html#tilelayer-crossorigin) option available for TileLayer configuration: https://github.com/Leaflet/Leaflet/blob/911f578/src/layer/tile/TileLayer.js#L140

natevw:
This should fix #7943 — I basically just copy-pasted the existing `crossOrigin` logic for `referrerPolicy` with appropriately modified documentation notes.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -77,7 +77,15 @@ export var TileLayer = GridLayer.extend({
 		// Whether the crossOrigin attribute will be added to the tiles.
 		// If a String is provided, all tiles will have their crossOrigin attribute set to the String provided. This is needed if you want to access tile pixel data.
 		// Refer to [CORS Settings](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) for valid String values.
-		crossOrigin: false
+		crossOrigin: false,
+
+		// @option referrerPolicy: Boolean|String = false
+		// Whether the referrerPolicy attribute will be added to the tiles.
+		// If a String is provided, all tiles will have their referrerPolicy attribute set to the String provided.
+		// This may be needed if you your map's rendering context has a strict default but your tile provider expects a valid referrer
+		// (e.g. to validate an API token).
+		// Refer to [HTMLImageElement.referrerPolicy](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy) for valid String values.
+		referrerPolicy: false
 	},
 
 	initialize: function (url, options) {
@@ -138,6 +146,10 @@ export var TileLayer = GridLayer.extend({
 
 		if (this.options.crossOrigin || this.options.crossOrigin === '') {
 			tile.crossOrigin = this.options.crossOrigin === true ? '' : this.options.crossOrigin;
+		}
+
+		if (this.options.referrerPolicy || this.options.referrerPolicy === '') {
+			tile.referrerPolicy = this.options.referrerPolicy === true ? '' : this.options.referrerPolicy;
 		}
 
 		/*
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -82,7 +82,7 @@ export var TileLayer = GridLayer.extend({
 		// @option referrerPolicy: Boolean|String = false
 		// Whether the referrerPolicy attribute will be added to the tiles.
 		// If a String is provided, all tiles will have their referrerPolicy attribute set to the String provided.
-		// This may be needed if you your map's rendering context has a strict default but your tile provider expects a valid referrer
+		// This may be needed if your map's rendering context has a strict default but your tile provider expects a valid referrer
 		// (e.g. to validate an API token).
 		// Refer to [HTMLImageElement.referrerPolicy](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy) for valid String values.
 		referrerPolicy: false
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -148,8 +148,10 @@ export var TileLayer = GridLayer.extend({
 			tile.crossOrigin = this.options.crossOrigin === true ? '' : this.options.crossOrigin;
 		}
 
-		if (this.options.referrerPolicy || this.options.referrerPolicy === '') {
-			tile.referrerPolicy = this.options.referrerPolicy === true ? '' : this.options.referrerPolicy;
+		// for this new option we follow the documented behavior
+		// more closely by only setting the property when string
+		if (typeof this.options.referrerPolicy === 'string') {
+			tile.referrerPolicy = this.options.referrerPolicy;
 		}
 
 		/*


src/layer/tile/TileLayer.js
```diff
@@ -140,6 +148,10 @@ export var TileLayer = GridLayer.extend({
 			tile.crossOrigin = this.options.crossOrigin === true ? '' : this.options.crossOrigin;
 		}
 
+		if (this.options.referrerPolicy || this.options.referrerPolicy === '') {
+			tile.referrerPolicy = this.options.referrerPolicy === true ? '' : this.options.referrerPolicy;
```
IvanSanchez, line 152: Copy-pasting went a bit too far :-)

The `crossOrigin` behaviour accepts an empty string, since that's one of the possible values for the HTML `crossorigin` attribute is an empty string.

But AFAICS, [the HTML `referrerpolicy` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-referrerpolicy) is different. There's no semantics for an empty string, and I don't think it makes much sense to accept a `true` `Boolean` here. It makes no sense to allow or provide) empty strings as the value of the attribute.

Also, note that `referrerpolicy` is lowercase in the MDN docs.

IvanSanchez (changes requested):

src/layer/tile/TileLayer.js
Falke-Design: > Also, note that referrerpolicy is lowercase in the MDN docs.

Same as for [crossOrigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-crossorigin). The html tag is lowercase but the JavaScript tag is `referrerPolicy `

src/layer/tile/TileLayer.js
```diff
@@ -77,7 +77,15 @@ export var TileLayer = GridLayer.extend({
 		// Whether the crossOrigin attribute will be added to the tiles.
 		// If a String is provided, all tiles will have their crossOrigin attribute set to the String provided. This is needed if you want to access tile pixel data.
 		// Refer to [CORS Settings](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) for valid String values.
-		crossOrigin: false
+		crossOrigin: false,
+
+		// @option referrerPolicy: Boolean|String = false
+		// Whether the referrerPolicy attribute will be added to the tiles.
+		// If a String is provided, all tiles will have their referrerPolicy attribute set to the String provided.
+		// This may be needed if you your map's rendering context has a strict default but your tile provider expects a valid referrer
```
Malvoz, line 85: typo here: "you your map's"

src/layer/tile/TileLayer.js
natevw: So as @Falke-Design noted, there's an HTML attribute vs. DOM property camel-casing distinction here.  So my JS version here is correct as far as the name itself goes.

And admittedly you caught an issue that I wondered about re. the empty string. I glossed over it because it's sort of a common idiom (*attribute*-wise) for "true" and "present but empty" to be analogous and I figured it was harmless. But since you noticed, I went ahead and looked more deeply into it and — now I still think my code is okay :-)

Or at least, just as correct as the `crossOrigin` I copy-pasted from. For [both](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-referrerpolicy) [properties](https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-crossorigin) the spec has the same "IDL attribute must reflect the content attribute" text and in both cases a "missing value default" is defined re. the empty string:

* https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attribute
* https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-empty-string

Basically afaict, an empty string is just as useful [or not] for *both* `crossOrigin` and `referrerPolicy` and basically means "keep the default" and so while I considered using `if (typeof this.options.referrerPolicy === 'string')` in the end I opted to match the current approved code as a "least surprise" thing so the two blocks for the two properties would simply match each other.

src/layer/tile/TileLayer.js
natevw: Good catch, fixed via ed83781

src/layer/tile/TileLayer.js
IvanSanchez: So the lower/uppercase confusion is on me. We should be looking at https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy and not at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-referrerpolicy . Got it.

But the empty string...

In `crossOrigin`, the lack of value means that the CORS mode is `no-cors`, and a value of an empty string means the CORS mode is `anonymous`. So, in the `L.TileLayer` implementation, `false` becomes `no-cors`, and `true` becomes `anonymous`.

But in `referrerPolicy`, a lack of value means "use the global default for the document", and an empty string means "use the global default for the document". So, in your implementation, `false` becomes "use the default", and `true` becomes.... "use the default". Setting the option from `false` to `true` does nothing, and I don't think that should be the documented behaviour.

natevw: > Setting the option from false to true does nothing, and I don't think that should be the documented behaviour.

Reasonable enough, I'll make the code match the (already) documented behavior.

natevw: I think this is resolved now, it now follows the "If a String is provided…" claim the documentation originally made and I've added a brief code comment acknowledging an intentional difference between the divergent implementation of this new one vs. the older `crossOrigin` lines above right above to avoid any confusion in future maintenance.

IvanSanchez (approved):
LGTM.

Falke-Design:
@IvanSanchez @natevw why is `Boolean` according to the docs a vaid option? If `true` or `false` do nothing, I would suggest to allow only `String` and set the default to `undefined`

PR #7955, opened 2022-01-27
create L.switchBasemap

clavijojuan:


docs/_plugins/user-interface/L.cascadeButtons
```diff
@@ -0,0 +1,12 @@
+---
```
Falke-Design, line 1: L.cascadeButtons are already added in #7942

PR #7967, opened 2022-02-04
Fix Marker position while zooming

### LINKED ISSUE #5181: Marker.setLatLng sets x and y based on pre-zoom projection during zoom animation.
- [x] I'm reporting a bug, not asking for help
- [x] I've looked at the [documentation](http://leafletjs.com/reference-1.0.0.html) to make sure the behaviour is documented and expected
  * I tried but there might be duplicates. Maybe I'm not using the right keywords.

----

When putting animated overlays or markers on map, during the zoom transition those animations will look off because the underlying tiles pane translate and make the marker or overlay look off location. 

For example [Leaflet Ant Path](http://rubenspgcavalcante.github.io/leaflet-ant-path/) looks bad during zoom transition:

lag

[Leaflet Animated Marker](http://openplans.github.io/Leaflet.AnimatedMarker/) is suffering from similar issue:

bike

I think Leaflet should provide an API that allows markers and overlays attach themselves to the element that zoom CSS transition is applied to so they are scaled with the tiles layer.

### LINKED ISSUE #5505: Marker events fired when zooming causing markers to move to preceding offset, then to the correct screen location.
## How to reproduce

- Browser (with version) I'm using: Chrome
- OS/Platform (with version) I'm using: W10

0. Zoom between 14-16 to see this effectively.
1. Create a new marker with default settings. To make life easier, create this marker on a click event.
2.  As the marker is created, zoom out 1 level. The marker should jump to the preceding screen offset, and then adjust itself to the current zoom level.

This may take a few tries. I've noticed it occuring semi-frequently when pinging new markers to my Leaflet app from a TCP server. When a new location arrives from the server, a marker is created and then added to the map. Zooming out anywhere in this process causes it to jump.

### LINKED ISSUE #5994: Marker drifts on zoom / marker jumps around on zoom

### LINKED ISSUE #7017: Marker drifts if setLatLng() is called during a zoom animation
Example: [jsfiddle](http://jsfiddle.net/cwad8p4m/1/)

**Steps to reproduce**
 - Add a marker with a custom icon (including iconSize and iconAnchor)
 - Update the markers position with `setLatLng` in an interval

**Expected behavior**
 - While zooming i'd expect the marker to maintain it's geographical position

**Current behaviour**
 - The marker drifts along it's screen position (similar to how a custom icon without setting an iconAnchor will drift).  It should be pretty apparent when you scoll wheel zoom in that fiddle.

**Additional context**
I'm writing a satellite overlay where the markers position changes rapidly and it needs a custom icon.

Falke-Design:
If a marker is set to a new latlng while zooming the marker jumps around. The problem is that the `this._pixelOrigin` (which is used for the calculation of the marker position) is only set after the zoom (`_onZoomTransitionEnd`), so while zooming the old pixelOrigin is used and the marker jumps. 

I tried to create a test but it was not possible.

- Fixes #7017
- Fixes #5181
- Fixes #5505
- Fixes #5994 (It describes exactly that issue but I'm not abel to reproduce it with the demo)

New src: [link](https://florianbischof.net/leaflet/marker-zoom/leaflet-src.js)

marker_jumping_wrong

Fixed:
marker_jumping_fix

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1211,7 +1211,7 @@ export var Map = Evented.extend({
 		return this;
 	},
 
-	_move: function (center, zoom, data) {
+	_move: function (center, zoom, data, supressEvent) {
 		if (zoom === undefined) {
 			zoom = this._zoom;
 		}
@@ -1221,17 +1221,20 @@ export var Map = Evented.extend({
 		this._lastCenter = center;
 		this._pixelOrigin = this._getNewPixelOrigin(center);
 
-		// @event zoom: Event
-		// Fired repeatedly during any change in zoom level,
-		// including zoom and fly animations.
-		if (zoomChanged || (data && data.pinch)) {	// Always fire 'zoom' if pinching because #3530
-			this.fire('zoom', data);
-		}
+		if(!supressEvent) {
+			// @event zoom: Event
+			// Fired repeatedly during any change in zoom level,
+			// including zoom and fly animations.
+			if (zoomChanged || (data && data.pinch)) {	// Always fire 'zoom' if pinching because #3530
+				this.fire('zoom', data);
+			}
 
-		// @event move: Event
-		// Fired repeatedly during any movement of the map,
-		// including pan and fly animations.
-		return this.fire('move', data);
+			// @event move: Event
+			// Fired repeatedly during any movement of the map,
+			// including pan and fly animations.
+			this.fire('move', data);
+		}
+		return this;
 	},
 
 	_moveEnd: function (zoomChanged) {
@@ -1697,6 +1700,12 @@ export var Map = Evented.extend({
 			noUpdate: noUpdate
 		});
 
+		if(!this._tempFireZoomEvent){
+			this._tempFireZoomEvent = this._zoom !== this._animateToZoom;
+		}
+
+		this._move(this._animateToCenter, this._animateToZoom, undefined, true);
+
 		// Work around webkit not firing 'transitionend', see https://github.com/Leaflet/Leaflet/issues/3689, 2693
 		setTimeout(Util.bind(this._onZoomTransitionEnd, this), 250);
 	},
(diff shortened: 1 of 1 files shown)


src/map/Map.js
```diff
@@ -1697,6 +1700,12 @@ export var Map = Evented.extend({
 			noUpdate: noUpdate
 		});
 
+		if (!this._tempFireZoomEvent) {
+			this._tempFireZoomEvent = this._zoom !== this._animateToZoom;
```
Falke-Design, line 1704: we need to store if the zoom is different, to fire later the `zoom` event.

src/map/Map.js
```diff
@@ -1697,6 +1700,12 @@ export var Map = Evented.extend({
 			noUpdate: noUpdate
 		});
 
+		if (!this._tempFireZoomEvent) {
+			this._tempFireZoomEvent = this._zoom !== this._animateToZoom;
+		}
+
+		this._move(this._animateToCenter, this._animateToZoom, undefined, true);
```
Falke-Design, line 1707: This is the necessary call to set the `_pixelOrigin` but we need to prevent firing the `zoom` event (to keep the current behaviour)

mourner (approved):
Nice!

mitow7821:
@Falke-Design In 1.8 leaflet version this problem occurs with permanent tooltip bound to that marker.

Issue #555, opened 2012-03-03
Marker drag event after dragend event


DennisSchiefer:
I encountered the following problem when dragging a Marker object over the map:
When I quickly drag the marker and release the mouse button, frequently the dragend event is fired before the last drag event. The final position of the Marker is that of the last drag event and not that of the dragstop event.

Thus, one cannot rely on the dragend event to correctly tell the end position of a drag operation.

EDIT: Tested on Windows XP, Firefox 10, Leaflet 0.3.1 stable, MS Mouse Explorer 3 with maximum mouse speed

PR #7972, opened 2022-02-06
Fix position:sticky which causes wrong document positioning

NICEXAI:
position:sticky causes the page to fail to be positioned by the '#' symbol, so the problem is solved by adding a normal element with zero height.

### DIFF
--- a/docs/reference.html
+++ b/docs/reference.html
@@ -924,8 +924,8 @@ are respected, and that the renderers do exist on the map.</p>
 
 </section><section class='collapsable'>
 
-<h4 id='map-methods-for-layers-and-controls'>Methods for Layers and Controls</h4>
-
+<h4>Methods for Layers and Controls</h4>
+<div class="ghost-pointer" id='map-methods-for-layers-and-controls'></div>
 
 <table><thead>
 	<tr>
@@ -1014,8 +1014,8 @@ are respected, and that the renderers do exist on the map.</p>
 
 </section><section class='collapsable'>
 
-<h4 id='map-methods-for-modifying-map-state'>Methods for modifying map state</h4>
-
+<h4>Methods for modifying map state</h4>
+<div class="ghost-pointer" id='map-methods-for-modifying-map-state'></div>
 
 <table><thead>
 	<tr>
@@ -1199,8 +1199,8 @@ and aborts resetting the map view if map.locate was called with
 
 </section><section class='collapsable'>
 
-<h4 id='map-other-methods'>Other Methods</h4>
-
+<h4>Other Methods</h4>
+<div class="ghost-pointer" id='map-other-methods'></div>
 
 <table><thead>
 	<tr>
@@ -1260,8 +1260,8 @@ if it's already initialized, optionally passing a function context.</p>
 
 </section><section class='collapsable'>
 
-<h4 id='map-methods-for-getting-map-state'>Methods for Getting Map State</h4>
-
+<h4>Methods for Getting Map State</h4>
+<div class="ghost-pointer" id='map-methods-for-getting-map-state'></div>
 
 <table><thead>
 	<tr>
@@ -1340,8 +1340,8 @@ If <code>zoom</code> is omitted, the map's current zoom level is used.</p>
 
 </section><section class='collapsable'>
 
-<h4 id='map-conversion-methods'>Conversion Methods</h4>
-
+<h4>Conversion Methods</h4>
+<div class="ghost-pointer" id='map-conversion-methods'></div>
 
 <table><thead>
 	<tr>
@@ -2100,7 +2100,7 @@ panning inertia).</td>
 	</div>
 </div>
 
-</section><h2 id='marker'>Marker</h2><p>L.Marker is used to display clickable/draggable icons on the map. Extends <a href="#layer"><code>Layer</code></a>.</p>
+</section><h2>Marker</h2><div class="ghost-pointer" id='marker'></div><p>L.Marker is used to display clickable/draggable icons on the map. Extends <a href="#layer"><code>Layer</code></a>.</p>
 
 <section>
 <h3 id='marker-example'>Usage example</h3>
@@ -2969,7 +2969,7 @@ properties. The event can optionally be propagated to event parents.</p>
 </section>
 
 
-</section><h2 id='popup'>Popup</h2><p>Used to open popups in certain places of the map. Use <a href="#map-openpopup">Map.openPopup</a> to
+</section><h2>Popup</h2><div class="ghost-pointer" id='popup'></div><p>Used to open popups in certain places of the map. Use <a href="#map-openpopup">Map.openPopup</a> to
 open popups while making sure that only one popup is open at one time
 (recommended for usability), or use <a href="#map-addlayer">Map.addLayer</a> to open as many as you want.</p>
 
@@ -3688,7 +3688,7 @@ properties. The event can optionally be propagated to event parents.</p>
 	</div>
 </div>
 
-</section><h2 id='tooltip'>Tooltip</h2><p>Used to display small texts on top of map layers.</p>
+</section><h2>Tooltip</h2><div class="ghost-pointer" id='tooltip'></div><p>Used to display small texts on top of map layers.</p>
 
 <section>
 <h3 id='tooltip-example'>Usage example</h3>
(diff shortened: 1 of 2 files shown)


docs/reference.html
```diff
@@ -924,8 +924,8 @@ <h3 id='map-method'>Methods</h3>
 
 </section><section class='collapsable'>
 
-<h4 id='map-methods-for-layers-and-controls'>Methods for Layers and Controls</h4>
-
+<h4>Methods for Layers and Controls</h4>
+<div class="ghost-pointer" id='map-methods-for-layers-and-controls'></div>
```
Malvoz, line 928: There are some implications with "ghost pointers":

- I think search engines use the names of elements with an ID in their search results:

  docs

  Will this change affect the search results in search engines?

- If an empty `<div>` is focused nothing will be announced to screen reader users. And yes, blind users are also developers. The creator of the [Audiom](https://audiom.xrnavigation.io/) <q>cross-sensory map component</q> which is based on Leaflet was created by a blind user.

NICEXAI: Thank you very much for your reply.

Question 1: The principle of Soso engine is to index the keywords of the page and then search, I think this change will not affect the normal search.
Question 2: I think you are right, maybe we can add the same content in the div and hide it, what do you think?

NICEXAI:
In any case, I think this issue of mislocation is necessary to fix, and it has caused me some inconvenience in my use of the documentation.

docs/reference.html
Malvoz: Actually, the docs may have accessibility problems already, I haven't looked into it much yet. But we shouldn't introduce new problems.

> maybe we can add the same content in the div and hide it

Is it necessary to duplicate the content? Shouldn't CSS changes be sufficient to fix the issue? Perhaps modify the margins, scroll-margin, scroll-padding or other related properties?

NICEXAI: I've checked the information and haven't found a better way to solve this problem yet, unless we drop position:sticky.

NICEXAI: Positioning the document accurately is probably more important to me than position:sticky.

NICEXAI: Is there a better way to solve this problem from your side?

Malvoz: So there seem to be 2 problems with the sticky headings:
- Clicking a link in the toc does not always scroll to the heading / top of that section
- Even when the heading is scrolled to, there's a too large gap due to:
  https://github.com/Leaflet/Leaflet/blob/48a38c797b19cefb28002f7d09199d947260106a/docs/docs/css/main.css#L13

/cc @avioli who introduced sticky headings in https://github.com/Leaflet/Leaflet/pull/7703.

NICEXAI: Current information indicates that position:sticky will cause the page to not find the ID correctly.

NICEXAI: So far it seems that only the first problem is encountered on my side for the time being.

Malvoz: `reference.html` is generated from `npm run docs`, so if I understand it correctly this file shouldn't be modified directly in most cases. But someone else can confirm that.

NICEXAI: Yes, I think you are right, I'll think about it and see if there is a better way to solve this problem I'm having.

avioli:
Actually the issue is because the sticky elements are sticking to the top and are not pushed off-screen, which is caused because of all headings being in one container.

Current:

https://user-images.githubusercontent.com/524259/152708451-782c52b8-6f9c-44d1-b566-44ac58e23116.mov

Possibly improved:

https://user-images.githubusercontent.com/524259/152708428-cb37d3c7-54c7-425c-ac19-f6cf98dafb50.mov

The DOM structure for the above cases is:

Current:

Screen Shot 2022-02-07 at 11 32 18 am

Possibly improved:

Screen Shot 2022-02-07 at 11 32 47 am

Following is a snippet I've came up with as a proof - to wrap all elements after the `toc` within a `div.content-item` and add the necessary style overrides:

(You can copy/paste it in your web-dev console, then close the console for the aside navigation to actually work)

```js
(() => {
  let collect = false;
  let collected = [];
  const groups = [];
  document.querySelectorAll('div.container > *').forEach(el => {
    if (
      collected.length &&
      el.nodeName === 'A' &&
      el.getAttribute('anchor')
    ) {
      el.id = el.getAttribute('anchor');
      groups.push(collected);
      collected = [];
    }
    if (el.nodeName === 'H2' && el.id) {
      el.id = null;
    }
    if (collect) {
      collected.push(el);
    } else {
      collect = el.id === 'toc';
    }
  });
  if (collected.length) {
    groups.push(collected);
  }
  groups.forEach(group => {
    const div = document.createElement('div');
    div.className = 'content-item';
    group[0].insertAdjacentElement('beforebegin', div);
    group.forEach(item => div.appendChild(item));
  });
  const style = document.createElement('style');
  style.textContent = `
  html {
    scroll-padding-top: 0;
  }
  .content-item {
    padding-top: 2em;
  }

  /* added */
  .api-page #toc ~ .content-item > h2,
  /* copied */
  .api-page #toc ~ h2 {
    font-weight: 700;
    font-size: 1.75em;
    color: #333;
    border-bottom: 3px solid #555;
    transition: border-color .25s ease;
    position: sticky;
    top: 0;
    z-index: 1;
    background-color: white;
    /* added */
    margin: 0;
  }

  /* added */
  .api-page #toc ~ .content-item > h2 ~ * {
    margin-top: 1.5em;
  }
  `;
  document.querySelector('head').appendChild(style);
})();
```

As you may have seen in the snippet - the `H2` ids are removed and "put" to the anchors before them, but I'm unsure of the implication for accessibility. I also have a feeling these anchors are generated by JavaScript, but I might be wrong.

Unfortunately the Map Methods and Map Misc are still not positioned correctly.

docs/reference.html
avioli: I added a comment to the main thread of the PR

NICEXAI:
> Actually the issue is because the sticky elements are sticking to the top and are not pushed off-screen, which is caused because of all headings being in one container.
> 
> Current:
> 
>  Screen.Recording.2022-02-07.at.11.28.34.am.mov 
> Possibly improved:
> 
>  Screen.Recording.2022-02-07.at.11.27.38.am.mov 
> The DOM structure for the above cases is:
> 
> Current:
> 
> Screen Shot 2022-02-07 at 11 32 18 am
> 
> Possibly improved:
> 
> Screen Shot 2022-02-07 at 11 32 47 am
> 
> Following is a snippet I've came up with as a proof - to wrap all elements after the `toc` within a `div.content-item` and add the necessary style overrides:
> 
> (You can copy/paste it in your web-dev console, then close the console for the aside navigation to actually work)
> 
> ```js
> (() => {
>   let collect = false;
>   let collected = [];
>   const groups = [];
>   document.querySelectorAll('div.container > *').forEach(el => {
>     if (
>       collected.length &&
>       el.nodeName === 'A' &&
>       el.getAttribute('anchor')
>     ) {
>       el.id = el.getAttribute('anchor');
>       groups.push(collected);
>       collected = [];
>     }
>     if (el.nodeName === 'H2' && el.id) {
>       el.id = null;
>     }
>     if (collect) {
>       collected.push(el);
>     } else {
>       collect = el.id === 'toc';
>     }
>   });
>   if (collected.length) {
>     groups.push(collected);
>   }
>   groups.forEach(group => {
>     const div = document.createElement('div');
>     div.className = 'content-item';
>     group[0].insertAdjacentElement('beforebegin', div);
>     group.forEach(item => div.appendChild(item));
>   });
>   const style = document.createElement('style');
>   style.textContent = `
>   html {
>     scroll-padding-top: 0;
>   }
>   .content-item {
>     padding-top: 2em;
>   }
> 
>   /* added */
>   .api-page #toc ~ .content-item > h2,
>   /* copied */
>   .api-page #toc ~ h2 {
>     font-weight: 700;
>     font-size: 1.75em;
>     color: #333;
>     border-bottom: 3px solid #555;
>     transition: border-color .25s ease;
>     position: sticky;
>     top: 0;
>     z-index: 1;
>     background-color: white;
>     /* added */
>     margin: 0;
>   }
> 
>   /* added */
>   .api-page #toc ~ .content-item > h2 ~ * {
>     margin-top: 1.5em;
>   }
>   `;
>   document.querySelector('head').appendChild(style);
> })();
> ```
> 
> As you may have seen in the snippet - the `H2` ids are removed and "put" to the anchors before them, but I'm unsure of the implication for accessibility. I also have a feeling these anchors are generated by JavaScript, but I might be wrong.
> 
> Unfortunately the Map Methods and Map Misc are still not positioned correctly.

Nice, but he only seems to solve the problem of title overlays, not positioning errors.

PR #7982, opened 2022-02-07
Fix scrolling in docs

### LINKED ISSUE #7971: Document positioning error within the API document by clicking on the sidebar button

**Steps to reproduce**

**Expected behavior**
leaflet-after

**Current behavior**
leaflet-before

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Google Chrome
- OS/Platform (with version): Windows10

Falke-Design:
Scrolling to the anchor is now be done manually instead of the browser internal function.

Fixes #7971
Closes #7972

Chromium browsers have a weird sideeffect. If we scroll to the same anchor in chromium browsers again, the scroll position has changed and the height of the sticky anchor was subtracted from the total scroll height / scroll position

@Malvoz, @NICEXAI, @avioli please check [Demo](https://florianbischof.net/leaflet/docs-scroll/reference.html#latlng)

### DIFF
--- a/docs/docs/css/main.css
+++ b/docs/docs/css/main.css
@@ -533,6 +533,7 @@ table.plugins td:first-child a {
 
 #toc-copy {
 	display: none;
+	z-index: 1;
 }
 
 #toc-copy div {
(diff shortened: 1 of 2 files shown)


NICEXAI:
Nice, this is a good way.

Malvoz:
Looks really good! Can this `z-index`:

https://github.com/Leaflet/Leaflet/blob/28d8ed5c6380c111f7f553332c12f28be5e69437/docs/docs/css/main.css#L498

that was introduces in https://github.com/Leaflet/Leaflet/pull/7703 be removed now? It seems to cause the following on mobile:

Falke-Design:
No removing `z-index: 1` is no option look:
grafik

But it works with adding `z-index: 1` to `#toc-copy` I will change this tomorrow

PR #7988, opened 2022-02-08
Add cache-version to tile url

### LINKED ISSUE #6422: With redraw (), the tiles are not requested again
## How to reproduce

- Browser (with version) I'm using: chrome 40
- OS/Platform (with version) I'm using: window 10
- step 1
add tileLayer
`    var test=L.tileLayer('https://api.tiles.mapbox.com/v4/{id}/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw', {
        maxZoom: 18,
        attribution: 'Map data &copy; <a href="https://www.openstreetmap.org/">OpenStreetMap</a> contributors, ' +
        '<a href="https://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, ' +
        'Imagery © <a href="https://www.mapbox.com/">Mapbox</a>',
        id: 'mapbox.streets'
    }).addTo(mymap);`
- step 2
use redraw()
` setTimeout(function(){
        test.redraw();
    },5000);`
 the tile are requested  once
image
if i use redraw(), the tiles should be requested twice

`    var mymap = L.map('mapid').setView([51.505, -0.09], 13);

    var test=L.tileLayer('https://api.tiles.mapbox.com/v4/{id}/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw', {
        maxZoom: 18,
        attribution: 'Map data &copy; <a href="https://www.openstreetmap.org/">OpenStreetMap</a> contributors, ' +
        '<a href="https://creativecommons.org/licenses/by-sa/2.0/">CC-BY-SA</a>, ' +
        'Imagery © <a href="https://www.mapbox.com/">Mapbox</a>',
        id: 'mapbox.streets'
    }).addTo(mymap);

 
    setTimeout(function(){
        test.redraw();
    },5000);`

Falke-Design:
Fixes #6422

Now a cache-version is added to the tile / image url (https://b.tile.openstreetmap.org/15/19160/11042.png?v=1644357627558).
This means if `redraw()` is called, it changes the cache-version and the tiles are reloaded instead of loaded from the browser cache.

### DIFF
--- a/spec/suites/SpecHelper.js
+++ b/spec/suites/SpecHelper.js
@@ -25,6 +25,13 @@ if (!Array.prototype.map) {
 	};
 }
 
+if (!String.prototype.startsWith) {
+	String.prototype.startsWith = function (searchString, position) {
+		position = position || 0;
+		return this.indexOf(searchString, position) === position;
+	};
+}
+
 expect.Assertion.prototype.near = function (expected, delta) {
 	expected = L.point(expected);
 	delta = delta || 1;
--- a/spec/suites/SpecHelper.js
+++ b/spec/suites/SpecHelper.js
@@ -43,6 +43,10 @@ expect.Assertion.prototype.nearLatLng = function (expected, delta) {
 		.be.within(expected.lng - delta, expected.lng + delta);
 };
 
+expect.Assertion.prototype.startsWith = function (expected) {
+	expect(this.obj.startsWith(expected)).to.be.ok();
+};
+
 happen.at = function (what, x, y, props) {
 	this.once(document.elementFromPoint(x, y), L.Util.extend({
 		type: what,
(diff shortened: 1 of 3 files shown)


mourner (changes requested):
This is not an acceptable solution, since you abandon tiles cached by the browser every time you reload the page. I believe we should keep the current behavior and not over-engineer this — for redraw to work properly, you either set up the tile server properly (with the right cache headers etc), or use the workaround from the comment. Doing this invasive hack by default is not the right approach.

Falke-Design:
Yes makes sense. A solution would be to keep the version in the LocalStorage of the Browser but I would say that we keep away from stroing things between sessions.

PR #7991, opened 2022-02-09
Fix responsiveness of media elements in all docs

### LINKED ISSUE #7420: Examples are not mobile-friendly

Malvoz:
Fix #7420:

> - [ ] [Responsive media elements](https://github.com/Leaflet/Leaflet/issues/7420#issuecomment-1023022670)

### DIFF
--- a/docs/docs/css/main.css
+++ b/docs/docs/css/main.css
@@ -93,6 +93,14 @@ iframe {
 	border: none;
 }
 
+img {
+	height: auto;
+}
+
+img,
+video {
+	max-width: 100%;
+}
 
 /* headings */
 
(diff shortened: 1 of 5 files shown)


mourner (commented):
Can we do a more generic fix, e.g. add `max-width: 100%` through CSS for all images/videos under tutorial/blog content, and make sure embedded maps still work properly? Adding to each image explicitly is fragile because we'll definitely forget to add it somewhere when updating tutorials in future.

mourner:
> :not(.leaflet-container)

In theory it should work without the filter — Leaflet CSS includes overrides for these kinds of rules because they are extremely common.

docs/docs/css/main.css
```diff
@@ -93,6 +93,14 @@ iframe {
 	border: none;
 }
 
+img {
+	height: auto;
```
Malvoz, line 97: Ensures images don't stretch (if `max-width: 100%` would cause an image to be smaller than its declared or intrinsic width).

mourner (approved):

PR #7996, opened 2022-02-12
Enable outlines on `leaflet-container` for keyboard users

Malvoz:
To pass [SC 2.4.7 Focus Visible](https://www.w3.org/TR/WCAG21/#focus-visible), the `outline` on `leaflet-container` must be visible on keyboard focus:

> Any keyboard operable user interface has a mode of operation where the keyboard focus indicator is visible.

Passing the SC can be accomplished (together with https://github.com/Leaflet/Leaflet/pull/7259) by removing `outline: 0` on `leaflet-container`.

## Preview

1. https://github.com/Leaflet/Leaflet/commit/f2ffb3bec57b71caaa56acc691089d52196a5be8:

   keyboard outline restored
   
    (note how the outline is clipped and seems to be thicker at the bottom, for some reason, I haven't investigated much but seems to be affected both by stacking contexts and `overflow: hidden` on the container)

2. https://github.com/Leaflet/Leaflet/commit/87ef061c2fadb1907d399d911a8ed2b5ac1b5dc6:

   keyboard outline adjusted for better visibility

   (opinionated change: default outline adjusted slightly)

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -251,6 +251,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 .leaflet-container {
 	background: #ddd;
+	outline-offset: 1px;
 	}
 .leaflet-container a {
 	color: #0078A8;
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -251,7 +251,6 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 .leaflet-container {
 	background: #ddd;
-	outline: 0;
 	}
 .leaflet-container a {
 	color: #0078A8;


Falke-Design (approved):
Valid for me.
Firefox outline:
grafik

Btw: @Malvoz did you know that Chrome and Firefox are starting focus on different containers? I used debug/map/markers.html and there I clicked somewhere on the white page. Chrome focus the leaflet-container, Firefox focus the first marker

Malvoz:
> Btw: @Malvoz did you know that Chrome and Firefox are starting focus on different containers? I used debug/map/markers.html and there I clicked somewhere on the white page. Chrome focus the leaflet-container, Firefox focus the first marker

Interesting find; when you click the area outside of the map and then tab then indeed the first focused element is different. 

Though, to clarify: it's unrelated to this PR change - it is a Firefox issue - whether they'd consider it an actual bug or not I don't know, since keyboard-only users are the users who rely on outlines, most of which probably don't use a mouse pointer at all.

PR #8019, opened 2022-02-22
Add check for blocker issues/PRs in RELEASE.md

Malvoz:
Per https://github.com/Leaflet/Leaflet/discussions/8018#discussioncomment-2229603:

> With the assumption that issues/PRs with this label must be checked manually - I suggest that we add a check in [RELEASE.md](https://github.com/Leaflet/Leaflet/blob/main/RELEASE.md) to ensure these issues/PRs are never skipped on releases.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,6 @@
 ## Releasing a new version of Leaflet
 
+- [ ] Ensure no [blocker](https://github.com/Leaflet/Leaflet/labels/blocker) issues and pull requests are unresolved.
 - [ ] Update [the changelog](https://github.com/Leaflet/Leaflet/blob/main/CHANGELOG.md) since last release and commit.
 - [ ] Run `npm version <patch | minor | major>` (this will bump the version in `package.json` and create a new tag).
 - [ ] Run `git push --follow-tags` to push the commit created by NPM to Github (together with the tag).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,6 @@
 ## Releasing a new version of Leaflet
 
-- [ ] Ensure no [blocker](https://github.com/Leaflet/Leaflet/labels/blocker) issues and pull requests are unresolved.
+- [ ] Ensure all https://github.com/Leaflet/Leaflet/labels/blocker issues and pull requests are resolved.
 - [ ] Update [the changelog](https://github.com/Leaflet/Leaflet/blob/main/CHANGELOG.md) since last release and commit.
 - [ ] Run `npm version <patch | minor | major>` (this will bump the version in `package.json` and create a new tag).
 - [ ] Run `git push --follow-tags` to push the commit created by NPM to Github (together with the tag).


RELEASE.md
```diff
@@ -1,5 +1,6 @@
 ## Releasing a new version of Leaflet
 
+- [ ] Ensure no [blocker](https://github.com/Leaflet/Leaflet/labels/blocker) issues and pull requests are unresolved.
```
Malvoz, line 3: > are unresolved

Maybe this should say "remain unresolved"?

Or maybe the sentence should rather say "Ensure all blocker issues and pull requests are resolved"?

Note that I refrained from using the words "open" or "closed", because that could technically indicate that closing them just to release is acceptable, and while I think everyone understands that it wouldn't be acceptable I think it's good to avoid any such indicators.

Falke-Design: Let us use `"Ensure all blocker issues and pull requests are resolved"`

Do we want that the label is rendered? Does it work in a .md file?
- [blocker](https://github.com/Leaflet/Leaflet/labels/blocker) `[blocker](https://github.com/Leaflet/Leaflet/labels/blocker)`
- https://github.com/Leaflet/Leaflet/labels/blocker `https://github.com/Leaflet/Leaflet/labels/blocker`

Malvoz: > Let us use `"Ensure all blocker issues and pull requests are resolved"`

Agree, done!

> Do we want that the label is rendered? Does it work in a .md file?

I can confirm it works in .md files. Changed to that, thanks! Looks much prettier 🙂

Falke-Design: > I can confirm it works in .md files. Changed to that, thanks! Looks much prettier 🙂

Are you sure? 
grafik

Malvoz: @Falke-Design Yes I tested in a private repo before committing the changes.

Preview test from editing directly:

blocker-label

Falke-Design: But when you open it in the "default" way it is not visible: https://github.com/Malvoz/Leaflet/blob/release-blocker-check/RELEASE.md

Malvoz: In my repo I do not have a `blocker` label: https://github.com/Malvoz/Leaflet/labels

Falke-Design (approved):

RELEASE.md
```diff
@@ -1,5 +1,6 @@
 ## Releasing a new version of Leaflet
 
+- [ ] Ensure all https://github.com/Leaflet/Leaflet/labels/blocker issues and pull requests are resolved.
```
mourner, line 3:
```suggestion
- [ ] Ensure all [blocker](https://github.com/Leaflet/Leaflet/labels/blocker) issues and pull requests are resolved.
```

Falke-Design: So you think we should only use a link and not the label badge self https://github.com/Leaflet/Leaflet/labels/blocker?

https://github.com/Leaflet/Leaflet/pull/8019#discussion_r812571731

Malvoz: The difference here is a "label link" vs normal link as seen in https://github.com/Leaflet/Leaflet/pull/8019#discussion_r812571731. We're going with a normal link then?

mourner: Ah, I forgot the GitHub does this with label links! Disregard this then.

PR #8033, opened 2022-03-04
OpenCage Geocoding and Geosearch plugins

freyfogle:
OpenCage now offers a Geosearch (ie location autosuggest) as a separate service from geocoding. This PR corrects the URL of the old OpenCage.Search plugin which no longer exists (is renamed to OpenCage.Geocoding) and adds OpenCage.Geosearch


IvanSanchez (approved):
All looks good!

PR #8037, opened 2022-03-08
test: add coverage for Layer\_addZoomLimit

### LINKED ISSUE #7762: Layer\_addZoomLimit not covered by tests
https://github.com/Leaflet/Leaflet/blob/d03a2d09aab86481793836cc6a4b80dff19b32ca/src/layer/Layer.js#L238

First part of condition ignored - tests are passing with or without.
Ref: #7609. Related: #7761

zishiwu123:
Hi everyone! This is my first time submitting a PR to Leaflet but don't go easy on me in the review.

This PR resolves #7762 by adding test coverage for the `addZoomLimit` function in Layer.js to check that zoom limits are updated when that function is given a Layer with non-NaN values for `minZoom` and `maxZoom`. I do have a few questions:

1) The `_addZoomLimit` function in Layer.js is actually part of the Map object, so I added the test functions into MapSpec.js instead of creating a new file called LayerSpec.js. Was this the right thing to do?

2) The `_addZoomLimit` function works differently when adding the first Layer to the Map versus adding subsequent Layers. When adding the first Layer, any non-NaN value of `minZoom` or `maxZoom` in that Layer will automatically update the `_layersMinZoom` and `_layersMaxZoom` variables in the Map to that non-NaN value. However, when adding any subsequent Layers, `_layersMinZoom` will be set to the minimum of the `minZoom` values from all the Layers in the Map, and vice versa for `_layersMaxZoom`. Hence we can only assert that `Map._layersMinZoom` is equal to `Layer.options.minZoom` when inserting the first Layer to the Map, and vice versa for `Map._layersMaxZoom`. When inserting the second Layer to the Map, we cannot guarantee the aforementioned assertion because there is the possibility that:
    - the first Layer's `minZoom` is less than the second Layer's **OR**
    - the first Layer's `maxZoom` is greater than the second Layer's

   However, I figured these assertions were enough since the goal of this PR is to provide test coverage for the function `_addZoomLimit` and not for `_updateZoomLevels`.

3) Would it be a good idea to open up a new issue to provide test coverage for the `updateZoomLevels` function?


Falke-Design (changes requested):
Hi this is a very good start thank you. Can you please add tests if two layers with different Zoom levels are added?

To your questions:

> I added the test functions into MapSpec.js

Correct

> Would it be a good idea to open up a new issue to provide test coverage for the updateZoomLevels function?

Yep, feel free to add tests for this function in a different PR. But first check if there are not already tests.

zishiwu123:
I amended the commit to include tests that check if two layers with different zoom levels will update the zoom limit. For example, if `minZoom` is 6 for the first layer and 4 for the second layer, then the min zoom limit of the map will updated to 4.

However, in the process I discovered an edge case that I think is a bug but wanted your opinion on. When only `minZoom` or only `maxZoom` is explicitly set in at least one layer of one or many layers added to the map, and when the set value is `NaN`, then the corresponding zoom limit in the map is updated to `NaN`. This is because:
- `minZoom` has a default value of 0 and `maxZoom` has a default value of 18 if not explicitly set in a layer
- therefore if either `minZoom` or `maxZoom` is set to `NaN` but the other is not explicitly set, then the other will get its default value which will cause the boolean condition in `_addZoomLimit` to evaluate to true
- then when `_updateZoomLevels` is called, both Math.min(`NaN`, <some_number>) and Math.min(`NaN`, <some_number>) will always return `NaN` causing the corresponding zoom limit to be updated to `NaN`

The boolean condition is: 
https://github.com/Leaflet/Leaflet/blob/17dcbe9b087c9fd16899bf8e0b27b72c34f7063e/src/layer/Layer.js#L230

This can be a hard-to-find edge case if the almost all the layers had valid numerical values, but one of the layers (in any order of being added into the map) met the condition I described. I have added test cases that check for `NaN` to show what I mean.

I searched for issues with the `NaN` keyword but I could not find a issue that referenced this error. I am wondering if:
1) this is a bug that needs to be addressed?
2) if yes to question 1, should a separate issue be opened up to describe this bug?

mourner:
@Falke-Design want to take another review pass?

spec/suites/map/MapSpec.js
```diff
@@ -1336,4 +1336,57 @@ describe("Map", function () {
 			}).to.not.throwException();
 		});
 	});
+
+	describe("_addZoomLimit", function () {
+		it("update zoom levels when min zoom is not NaN in a layer that is added to map", function () {
```
Falke-Design, line 1341: Please check each test describtion, the min zoom is a number. I think copy&paste issue.

spec/suites/map/MapSpec.js
```diff
@@ -1336,4 +1336,57 @@ describe("Map", function () {
...
+			expect(map._layersMinZoom).to.be(4);
+		});
+
+		it("update zoom levels when max zoom is not NaN in two layers that are added to map", function () {
+			map._addZoomLimit(L.tileLayer("", {maxZoom: 10}));
+			map._addZoomLimit(L.tileLayer("", {maxZoom: 8}));
+			expect(map._layersMaxZoom).to.be(10);
+		});
+
+		it("update zoom levels, with min zoom level to NaN, when min zoom is NaN in a layer that is added to map", function () {
```
Falke-Design, line 1363: Please add to each `NaN` test following line:
> // This test shows the NaN usage - it's not clear if NaN is a wanted "feature"

I think we should add this comment because else in the future we maybe think `NaN` is a wanted feature.

Falke-Design (changes requested):
@mourner please read the comment above and tell us what do you think over the `NaN` usage for min and max zoom. I think we should take the defaults if it is `NaN`

spec/suites/map/MapSpec.js
zishiwu123: Added comments on top of each test function that shows NaN usage for min zoom or max zoom.

jonkoops:
@zishiwu123 could you give this PR a rebase and fix the merge conficts? @Falke-Design let's do another review pass, I think we'd like to get this in asap.

Falke-Design (approved):
For the open question / bug I created a new issue: #8037

@zishiwu123 thank you for your contribution

PR #8039, opened 2022-03-08
Parse dist-tag from version for publishing on NPM

### LINKED ISSUE #8038: Support alpha/beta NPM dist tags when publishing through Github Actions
Currently our Github Actions are set up so that we do `npm publish` which assigns `latest` NPM tag by default, regardless whether the version is stable (`v1.8.0`) or not (e.g. `v1.8.0-alpha.1`, `v1.8.0-beta.1`). We should parse out the word that comes after the version (e.g. `beta`) in the publish action, and use it as a tag if present (and otherwise `latest`). `npm publish --tag <tag>`. cc @jonkoops

mourner:
Makes it so that we do `npm publish --tag latest` for stable releases and e.g. `npm publish --tag beta` for versions like `v1.8.0-beta.1`. Closes #8038.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -211,6 +211,8 @@ jobs:
 
       - name: test dist-tag parse
         run: echo $VERSION | grep -oP '^v\d+\.\d+\.\d+-?\K(\w+)?'
+        env:
+          VERSION: v1.8.0-alpha.1
 
       - name: Publish to NPM
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -194,7 +194,7 @@ jobs:
 
   publish-npm:
     needs: build
-    # if: github.repository_owner == 'Leaflet' && startsWith(github.ref, 'refs/tags/v')
+    if: github.repository_owner == 'Leaflet' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - name: Restore build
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -210,6 +210,8 @@ jobs:
           registry-url: https://registry.npmjs.org
 
       - name: Publish to NPM
-        run: npm publish
+        run: |
+          TAG=$(echo $GITHUB_REF_NAME | grep -oP '^v\d+\.\d+\.\d+-?\K(\w+)?')
```
Falke-Design, line 214: Does this work on Windows too? I mean that `grep` is not installed per default

mourner: It doesn't matter, this script is specifically for running on Ubuntu within GitHub Actions.

Falke-Design: Ah ok, so `npm publish` is not called manually? Is it triggered over a Github Release?

mourner: It's triggered when we create and push a `v*` git tag (@jonkoops although maybe we should make the check more conservative, or create another workflow so that it could be triggered by a new GitHub Release).

jonkoops: Instead of using `$GITHUB_REF_NAME` you could also get the version using `npm view version leaflet` (which will not include the `v` prefix). But this works excellently as well.

jonkoops (approved):

PR #8042, opened 2022-03-08
Map is panned to show markers when they're focused

### LINKED ISSUE #8040: Autopan out of view markers into view on focus

When keyboard-only users <kbd>tab</kbd> through markers there's a high probability that focused markers are out of view, which is disorienting and also makes it impossible for these users to nondisruptively interact with the maps (as they'd have to <kbd>shift</kbd> + <kbd>tab</kbd> back to `leaflet-container` to pan, and at that point it's a guessing game where the (previously) focused marker exists on the map).

Relates to WCAG [SC 2.4.7 Focus Visible](https://www.w3.org/TR/WCAG21/#focus-visible).

A new option to auto pan markers into view on focus.  It should default to `true`.

For example `AutoPanOnFocus` or `KeyboardAutoPan` (for consistency with other keyboard options?).

Universally meaningful pan and zoom: https://github.com/mapbox/mapbox-gl-js/issues/10114 / https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396#issue-858020120. This could be experimented with as a Leaflet plugin.

**Additional context**

Video to showcase the feature (from https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/571, although in that case it happens to also be applicable to vectors):

https://user-images.githubusercontent.com/55214462/141527065-f3e91ac0-0805-482f-abaf-05d64b1b1597.mp4

IvanSanchez:
Should fix #8040.

@Malvoz This is a quick implementation that I haven't tested thoroughly. If you can, please test with `DivIcon`s and marker images of different sizes.

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -2,7 +2,9 @@ import {Layer} from '../Layer';
 import {IconDefault} from './Icon.Default';
 import * as Util from '../../core/Util';
 import {toLatLng as latLng} from '../../geo/LatLng';
+import {toPoint as point} from '../../geometry/Point';
 import * as DomUtil from '../../dom/DomUtil';
+import * as DomEvent from '../../dom/DomEvent';
 import {MarkerDrag} from './Marker.Drag';
 
 /*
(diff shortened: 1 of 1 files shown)


Malvoz:
This is already a huge accessibility win. Awesome, thanks!

> I'm concerned that panning a marker barely into view (i.e. right now the marker touches the edge of the map container), it may be overlapped and obscured by map controls.

> Maybe bigger margin?

Yes, or rather if it's possible to have the marker panned to the center of the map? That'd increase the likelihood of popups' content to also be in view.

IvanSanchez:
> Yes [to the bigger margin], or rather if it's possible to have the marker panned to the center of the map?

The way that the popup's `autoPan` works is by giving it a margin wide enough to accommodate the popup. So maybe the logic has to check if there's a popup bound and open, grab its size, and on top of that add a margin of, like, 20% of the map's container size.

I'm not sure about these kind of specifics right now.

Malvoz (approved):

Issue #8045, opened 2022-03-09
Artifacts upload doesn't pick up tags


mourner:
Looks like artifacts upload isn't working correctly for tagged commits — e.g. https://github.com/Leaflet/Leaflet/runs/5477985170 still uploads to `/main`, even though I expected it to upload to `/v1.8.0-beta.0`. Let's investigate so that further betas / final are correctly exposed for download. cc @jonkoops

PR #8046, opened 2022-03-09
Potentially fix path for tagged artifacts upload

mourner:
A blind attempt at fixing #8045. I have no idea why `git tag --points-at HEAD` didn't work here, but maybe let's try with the built in environment var?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -175,10 +175,6 @@ jobs:
         working-directory: dist
         run: zip -r leaflet.zip .
 
-      - name: Determine directory for artifacts
-        id: artifacts-directory
-        run: echo "::set-output name=path::content/leaflet/${GITHUB_REF_NAME:-main}"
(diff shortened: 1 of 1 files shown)


mourner:
Simplified the setup with the second commit, @jonkoops this should work right?

jonkoops (approved):
@mourner reading the docs this might work, let's give it a try.

mourner:
Fucking hell, it deleted the whole S3 bucket :( I'm dumb. https://github.com/Leaflet/Leaflet/runs/5515431118

mourner:
OK, I fixed the blunder, and restored `v1.7.1` and `main` artifacts. Maybe deleting all the old stuff is actually good, it was never supposed to be hotlinked anyway.

jonkoops:
Oh shit, I guess we were playing with fire here hahah. Good to hear it's restored.

PR #8050, opened 2022-03-10
Fix bundle export error

### LINKED ISSUE #8048: leaflet export broken in 1.8.0-beta.0
**Steps to reproduce**
- import leaflet in a js file
- use a bundler like webpack or rollup to bundle the code
- load the bundle in any html file that does not have a div with id `leaflet`

**Expected behavior**
No error

**Current behavior**
Javascript Error: `Uncaught ReferenceError: leaflet is not defined`

**Environment**
- Leaflet version: 1.8.0-beta.0
- Browser (with version): Chrome 99.0.4844.51, Firefox 97.0.2
- OS/Platform (with version): Mac OS 12.2.1 (21D62)

**Additional context**
The problem seems to be that at the end of leaflet-src.js `leaflet` is used instead of `exports` now.

``` JavaScript
  var oldL = window.L;
  leaflet.noConflict = function() {
  	window.L = oldL;
  	return this;
  }
  // Always export us to window global (see #2364)
  window.L = leaflet;
```

A fun little detail is that if you have a div with the id `leaflet` in the html it is added to the global scope and everything works as intended. But I'm not sure yet why that fixes it.

https://github.com/Intermaps/leaflet-1.8.0-beta.0-export-bug

You can build with either webpack or rollup (I first thought the issue was related to my build process) and then open index.html

Falke-Design:
Fix #8048

@Eschon can you please check?

### DIFF
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -13,12 +13,12 @@ const version = release ? pkg.version : `${pkg.version}+${gitRev.branch()}.${git
 const banner = createBanner(version);
 
 const outro = `var oldL = window.L;
-leaflet.noConflict = function() {
+exports.noConflict = function() {
 	window.L = oldL;
 	return this;
 }
 // Always export us to window global (see #2364)
-window.L = leaflet;`;
+window.L = exports;`;
 
 /** @type {import('rollup').RollupOptions} */
 const config = {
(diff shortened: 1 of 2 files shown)


Eschon:
Just tried it out and the error is gone. Looks good to me.

Thanks for the quick fix!

IvanSanchez (approved):
If @Eschon approves, then I'll merge.

Issue #8051, opened 2022-03-11
Use the normal document font size in general (on `leaflet-container`)?


Malvoz:
**Is your feature request related to a problem? Please describe.**

I believe small fonts undermine the importance of text in popups, the layers control, etc. - relatively to other body text:

https://github.com/Leaflet/Leaflet/blob/0cea36c16934e616e66e1ab4287640dbe9bf3658/dist/leaflet.css#L266-L269

**Describe the solution you'd like**

Don't generally decrease the font size.

Excluding things like attribution and scale control that are intended to be subtle, a normal font size on `leaflet-container` is more readable (maybe especially so on mobile).

Issue #7952, opened 2022-01-26
Potential illegible font size for map attribution and popup content


natevw:
As was foretold by https://github.com/Leaflet/Leaflet/pull/7571#issuecomment-960906568, the changes made in https://github.com/Leaflet/Leaflet/pull/7571 / https://github.com/Leaflet/Leaflet/pull/7800 can result in very small font sizes in some contexts. I.e. looks like the IE compatibility concern was addressed, but **not** the general issue of "what happens if a site has its own `html { font-size: --arbitrary-value; }`"?

For example one site I'm maintaining was started out years ago based on a Bootstrap stylesheet that makes a `rem` work out to `10px` and this led to nearly-illegible popup and attribution content after updating to the latest Leaflet. [To be clear: when the `rem` is set to `10px` or `62.5%` at the `html` level, usually the initial `em` font size for the page is then scaled back **up** at the `body` level. I'm not complaining about the page **and** Leaflet having too-small text — only Leaflet ends up with tiny text in practice! See example in my later comment below.]

~~In my case I simply worked around by adding a subset of this~~

```
// [old] WORKAROUND: 
.leaflet-container { font-size: 12px; }
.leaflet-container .leaflet-control-attribution,
.leaflet-container .leaflet-control-scale,
.leaflet-control-scale-line { font-size: 11px; }
```

**UPDATE**: now after https://github.com/Leaflet/Leaflet/pull/8057 the workaround should need only:

```
// set up units to match Leaflet expectation
.leaflet-container { font-size: 1.2rem; }
```

to the site's own stylesheet which is included after the vendored Leaflet one and so takes precedence.

Falke-Design:
Can you please create a sample www.leafletjs.com/edit.html

Malvoz:
What do you mean by "illegal font size"? [Most countries apply laws and policies based on WCAG](https://www.w3.org/WAI/policies/). But WCAG does not have a "minimum font size required" success criterion. Though there is https://www.w3.org/TR/WCAG21/#resize-text which can't reliably be met using `px` units, as I understand it. See references:

- https://www.craigabbott.co.uk/blog/accessibility-and-font-sizes
- https://www.24a11y.com/2019/pixels-vs-relative-units-in-css-why-its-still-a-big-deal/<br>
  <img width="500" src="https://user-images.githubusercontent.com/26493779/151317439-9901e985-b670-4fe2-84f8-d9462812db25.png">

[...]

PR #7800, opened 2021-11-22
Add fallbacks to font-size

Chandu-4444:
Discussed [here](https://github.com/Leaflet/Leaflet/pull/7571#discussion_r753873292).

PR #8052, opened 2022-03-11
Refactor font sizes for text content

Malvoz:
## Comparison

<table>
  <tr>
    <th></th>
    <th>Default</th>
    <th>Custom <code>font-size</code><sup> <a href="#note">[NOTE]</a></sup></th>
    <th>Browser setting: <q>Very large</q></th>
  </tr>
  <tr>
    <th>1.7.1</th>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <th>1.8.0-beta.0</th>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <th>PR preview</th>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

<details id="note">

<summary>[NOTE]: The custom font sizes used</summary><br>

```css
html {
  font-size: 200%; /* changes root font-size */
}

body {
  font-size: 0.3em; /* changes parent element font-size */
}
```

With these particular custom styles, the computed font-size _SHOULD_ be smaller than the _default_ font-size in the screenshots.

<hr>

</details>

<table>
  <tr>
    <th></th>
    <th>1.7.1</th>
    <th>1.8.0-beta.0</th>
    <th>PR preview</th>
  </tr>
  <tr>
    <th>Scales relatively to the root (<code>&lt;html&gt;</code>) font size?</th>
    <td>❌</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
  <tr>
    <th>Scales relatively to custom font sizes on (non-root) ascendant elements?</th>
    <td>❌</td>
    <td>❌</td>
    <td>✔️</td>
  </tr>
  <tr>
    <th>Respects the user's browser settings?</th>
    <td>❌</td>
    <td>✔️</td>
    <td>✔️</td>
  </tr>
</table>

---

- Resolves https://github.com/Leaflet/Leaflet/issues/7952
  (by accepting that the new relative font sizes may cause illegible fonts depending on the application's CSS (i.e. responsibility is on the developer to not use too small font sizes), this is simply the nature of relative font sizes. In other words things like https://github.com/Leaflet/Leaflet/discussions/8044#discussioncomment-2331708 is expected and some developers may have to make CSS changes, the alternative is to close this PR, revert changes in https://github.com/Leaflet/Leaflet/pull/7800, and go back to the 1.7.1 scenario)
- Fixes https://github.com/Leaflet/Leaflet/pull/7346
  (because the attribution and scale control now has the same font size as the `<small>` element due to `font-size: smaller`, which is larger than 12px)
- Fixes https://github.com/Leaflet/Leaflet/issues/8051
  (because the default font size in Leaflet is now equal to that of body text, except for attribution and scale control which remain smaller)

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -264,8 +264,8 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 /* general typography */
 .leaflet-container {
-	font: 12px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
-	font: 0.75rem/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
+	font-family: "Helvetica Neue", Arial, Helvetica, sans-serif;
+	line-height: 1.5;
 	}
 
 
@@ -417,8 +417,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-container .leaflet-control-attribution,
 .leaflet-container .leaflet-control-scale {
-	font-size: 11px;
-	font-size: 0.69rem;
+	font-size: smaller;
 	}
 .leaflet-left .leaflet-control-scale {
 	margin-left: 5px;
@@ -431,8 +430,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-top: none;
 	line-height: 1.1;
 	padding: 2px 5px 1px;
-	font-size: 11px;
-	font-size: 0.69rem;
+	font-size: smaller;
 	white-space: nowrap;
 	overflow: hidden;
 	-moz-box-sizing: border-box;


Malvoz:
An alternative solution to the changes suggested here is to convert and use `em` (as proposed in https://github.com/Leaflet/Leaflet/issues/7952#issuecomment-1030333605) instead of `rem` (https://github.com/Leaflet/Leaflet/pull/7800), though which would not increase the default font size (and as such would not fix https://github.com/Leaflet/Leaflet/pull/7346 and https://github.com/Leaflet/Leaflet/issues/8051).

mourner:
This is great! But the default font sizes look unreasonably large to me, and completely out of character for interactive maps. The font size legibility guidelines make sense for content-heavy web pages, but it's very different here — popups are usually small and have very limited space to afford gigantic default text size, and attribution isn't really meant to be prominent and read by everyone, it's more of a small aside, unobtrusive and not taking attention away from the map itself.

Malvoz:
@mourner Thanks for the feedback. I don't share the same opinion (https://github.com/Leaflet/Leaflet/issues/8051), but you're the man. Feel free to close this PR.

@natevw Do you want to send a PR for your proposal for `em` in https://github.com/Leaflet/Leaflet/issues/7952? They should have the same affect as this PR but without changes to the default font size, I think: https://github.com/Leaflet/Leaflet/issues/7952#issuecomment-1065376804.

mourner:
Maybe we could settle on some kind of a compromise? Still fix relative font sizes and "large" setting, and increase the default size a little bit, but not that much?

Falke-Design:
I see the benefits for the accessibility but when I think out of my project view and not from Leaflet view, I wouldn't like that the text is after an update to 1.8.0 much bigger especially the attribution control. 

I add the `blocker` label to this PR because we need a descision for the problem: https://github.com/Leaflet/Leaflet/discussions/8044#discussioncomment-2331708

dist/leaflet.css
```diff
@@ -431,8 +430,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	border-top: none;
 	line-height: 1.1;
 	padding: 2px 5px 1px;
-	font-size: 11px;
-	font-size: 0.69rem;
+	font-size: smaller;
```
mourner, line 433: Also, do we know if relative sizes like `smaller` work in older IE versions? I'd maybe leave the adjusted pixel definitions before new ones just in case.

natevw:
> @natevw Do you want to send a PR for your proposal for em in https://github.com/Leaflet/Leaflet/issues/7952? They should have the same effect as this PR I think: https://github.com/Leaflet/Leaflet/issues/7952#issuecomment-1065376804, but without changes to the default font size.

Yes, I still have my VM set up for this project and could do so this coming week. Basically my approach would be to replace the new `rem` to just `em`.

dist/leaflet.css
Malvoz: Thanks for pointing this out. I thought about it but it still managed to slip my mind. I did a quick search for browser support but didn't find anything. The lack of support data is usually indicative of legacy features, but unless we find any support data we should keep the `px` variants as fallback.

Malvoz:
Closing per https://github.com/Leaflet/Leaflet/pull/8057:

> I'm wary about addressing https://github.com/Leaflet/Leaflet/issues/7952 (like in https://github.com/Leaflet/Leaflet/pull/8052) because once we fix it, we're getting a bunch of potential problems from the other side

The "potential problems" were expected and described in this PR, but maybe a big breaking change like this (if accepted) is better suited for v2.0.0.

PR #8057, opened 2022-03-14
Increase default font sizes for improved legibility

mourner:
I have spent some time tweaking the default Leaflet font sizes for improved legibility, while not compromising on the look and compactness too much — here are the changes.

I'm wary about addressing #7952 (like in #8052) because once we fix it, we're getting a bunch of potential problems from the other side (cc @natevw) — Leaflet maps are very often embedded in content contexts like blogs (with parent font sizes varying to a great extent), and this will have breaking consequences on a lot of maps (needing design adjustments), while we want to retain relatively consistent look of Leaflet maps regardless of context — so that's more of an independent component within a certain design. The old `html { font-size: 62.5% }` hack is an acceptable edge case to leave to the user to address manually IMO.

I also want to retain the current base font-size of `.leaflet-container` because many plugins depend on it for their design choices. Thus I'm increasing font sizes on children controls instead.

Closes #7346. I've come to like the 1px increase — sorry for doubting this earlier @Malvoz! Also adjusted margins and line height a little bit to compensate, and bumped the opacity of the attribution control.

Before:
image 
After:
image

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -264,8 +264,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 /* general typography */
 .leaflet-container {
-	font: 12px/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
-	font: 0.75rem/1.5 "Helvetica Neue", Arial, Helvetica, sans-serif;
+	font-family: "Helvetica Neue", Arial, Helvetica, sans-serif;
+	font-size: 12px;
+	font-size: 0.75rem;
+	line-height: 1.5;
 	}
 
 
@@ -383,6 +385,8 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-control-layers label {
 	display: block;
+	font-size: 13px;
+	font-size: 1.08333em;
 	}
 .leaflet-control-layers-separator {
 	height: 0;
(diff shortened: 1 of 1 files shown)


dist/leaflet.css
```diff
@@ -400,13 +404,14 @@ svg.leaflet-image-layer.leaflet-interactive path {
 
 .leaflet-container .leaflet-control-attribution {
 	background: #fff;
-	background: rgba(255, 255, 255, 0.7);
+	background: rgba(255, 255, 255, 0.8);
```
Malvoz, line 407: @mourner This is potentially a fix for https://github.com/Leaflet/Leaflet/issues/7538, depending on the consensus of that issue. See https://github.com/Leaflet/Leaflet/issues/7538#issuecomment-1038285629.

mourner: Yep! We can keep the issue to maybe consider bumping it further, but at least 0.8 seems like a definite improvement without compromising the design aesthetic much, so we can do it now.

Malvoz (approved):
LGTM!

natevw:
@mourner This looks good to me too, I saw as I went to create a followup PR for #7952 that you'd already taken the same approach I would have. Specifically to use `rem` only on the `.leaflet-container` and then everything else inside having `em`.

Maybe ideally it would be based on `.leaflet-container { font-size: 1rem; }` but that probably would require lots of messing with all the other `em` units so for now 👍  to your `0.75rem` as well for a minor release.

I'll update my ticket too since this approach won't work for 100% of cases but significantly simplifies the workaround when it doesn't. Thanks for revisiting these sizes!

🌻

Issue #8029, opened 2022-03-02
Marker.bindPopup() should throw exception if passed invalid argument


php4fan:
In my code I had:
```
someMarker.bindPopup({ content: '<p>Hello</p>'}); // WRONG
```
Obviously that's wrong, as I should just pass it the content as a string:
```
someMarker.bindPopup( '<p>Hello</p>'); // CORRECT
```

**Expected behavior:**

The above wrong line of code should trigger an exception, with an error message similar to:
```
 Invalid argument passed to `bindPopup()`: expect string or <whatever else is accepted>, object passed instead.
```
Well, something like that, I am very bad at phrasing error messages. Anyway that's when the error should be thrown and that's the kind of information it should contain. When I do something wrong, the error must be thrown at the time of my wrongdoing, and the error message must point to what I did wrong. 
If the second line of the stack trace doesn't point to my code, there's a bug.

**Observed behavior**
`bindPopup()` doesn't throw an error, it blindly accepts and stores the invalid content I passed, and then when the user actually clicks on the popup, things explode with this error:
```
DivOverlay.js:206 Uncaught TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'.
_updateContent @ DivOverlay.js:206
update @ DivOverlay.js:117
onAdd @ DivOverlay.js:54
onAdd @ Popup.js:112
_layerAdd @ Layer.js:114
whenReady @ Map.js:1465
addLayer @ Layer.js:176
openPopup @ Popup.js:336
openPopup @ Popup.js:424
_openPopup @ Popup.js:499
fire @ Events.js:190
_fireDOMEvent @ Map.js:1440
_handleDOMEvent @ Map.js:1397
c @ DomEvent.js:92
```
which doesn't trace to my code and costs me hell to debug, when in fact it was a very trivial error.

PR #8079, opened 2022-03-21
Marker.bindPopup() throws exception if passed invalid argument

KonstantinBiryukov:
Error throwing was added for this issue: 
[(https://github.com/Leaflet/Leaflet/issues/8029](url)
Marker.bindPopup() expects 'string', 'function', 'HTMLElement' or 'Popup'. Otherwise, the error is thrown.


mourner:
We intentionally avoid excessive validation of arguments so that the code doesn't become too verbose. The benefit of this is too low for the added complexity if we start applying this consistently all over the codebase.

jonkoops:
If you want to ensure these kind of exceptions do not happen in your application you can use TypeScript to get compile-time guarantees about the correctness of your code.

PR #8084, opened 2022-03-25
Fix undefined iconAnchor

Falke-Design:
With the implementation of #8042 a error occurs (mostly on DivIcons) if `iconAnchor` is not defined.

Fixes issues in #8044.


mourner (approved):

jonkoops:
This is a good example of a bug that TypeScript would have caught for us :smile:

Issue #8086, opened 2022-03-26
UML diagram is outdated


Falke-Design:
The UML diagram in the Tutorial [Extending Leaflet](https://leafletjs.com/SlavaUkraini/examples/extending/extending-1-classes.html) is outdated.

On a fast check VideoOverlay and SVGOverlay are missing.

@IvanSanchez @mourner how was the diagram created?

Malvoz:
> how was the diagram created?

Can we use Mermaid perhaps?  See https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/.

This:


> <pre>
> 
> ```mermaid
>   graph TD;
>       A-->B;
>       A-->C;
>       B-->D;
>       C-->D;
> ``` 
> 
> </pre>

produces:

```mermaid
  graph TD;
      A-->B;
      A-->C;
      B-->D;
      C-->D;
```

(this would also make the diagram available to devs who use screen readers, related: https://github.com/Leaflet/Leaflet/discussions/8006#discussioncomment-2190730)

IvanSanchez:
> @IvanSanchez @mourner how was the diagram created?

I created that manually, ages ago. The tool I used and the "source" for the diagram are at https://github.com/Leaflet/Leaflet/blob/main/docs/SlavaUkraini/examples/extending/class-diagram.txt

But last couple of years I've been [perfecting some Leafdoc templates](https://gitlab.com/IvanSanchez/gleo/-/tree/master/build/leafdoc-templates/graphviz-class-diagram) for generating [graphviz](https://graphviz.org/)-based UML class diagrams, and they look great (to me).

![imagen](https://user-images.githubusercontent.com/1125786/160275323-5aecefff-12a4-43a8-afb5-51a5e7e07d96.png)

This would need updating the Leafdoc version, and grabbing those graphviz templates (plus having the graphviz tools to convert the `.dot` file into an image). Also, the leafdoc version I'm using supports the `@relationship` directive, so other kinds of arrows can be drawn in the UML diagram. I'm a **big** fan of keeping the UML diagram in sync with the docstrings.

I've become a big fan of graphviz - it's tried and true tech. If you *really* fancy the mermaid thing, it'd be possible to create similat leafdoc templates for that output format.

Falke-Design:
@IvanSanchez this looks great! Keeping the UML diagram in sync is a good idea and makes totally sense.

Two things from my side, but I have no clue of the UML standard:
[...]

PR #8094, opened 2022-03-29
Leafdoc now builds UML diagrams

IvanSanchez:
Fixes part of #8086.

This changes the documentation build mechanism, so it runs three times: one for the API documentation, as now, and twice to produce two UML diagrams: one bare (only class names and relationships) and one full (showing every method, option, property and event).

These diagrams use [graphviz](https://graphviz.org/)'s `.dot` format. In order to render them to images, one needs the graphviz utilities (since I'm a Debian person, I just need to `apt install graphviz`).

Once rendered, the "bare" diagram looks pretty clean and gives an overview of relationships between classes, both inheritance and aggregation:

imagen

And the "full" diagram, while complete, is quite a monster:

imagen

There are a number of small tweaks in the leafdoc templates for the UML diagrams - to group the utility classes together and prevent a couple of ugly things. Shouldn't be too bad.

The docs are built just as before - with a `npm run docs`.


build/docs.js
```diff
@@ -1,31 +1,41 @@
...
+		showInheritancesWhenEmpty: true,
+		leadingCharacter: '@'
+	});
 
-// Note to Vladimir: Iván's never gonna uncomment the following line. He's
-// too proud of the little leaves around the code.
-// doc.setLeadingChar('@');
+	// Note to Vladimir: Iván's never gonna uncomment the following line. He's
+	// too proud of the little leaves around the code.
+	// doc.setLeadingChar('@');
```
Falke-Design, line 20: Are this lines still correct? In the options `leadingCharacter: '@'` is defined.

Falke-Design (commented):
What is if a user don't have `graphviz` installed? Is it still possible to generate the docs?

build/docs.js
IvanSanchez: Well, ***I*** didn't uncomment the line, so it's *technically* correct.

But, yes, that comment is superfluous right now.

Malvoz:
> What is if a user don't have `graphviz` installed? Is it still possible to generate the docs?

I don't have graphviz installed and the docs + `.dot` files were generated successfully.

Though I take it the intention for now is to generate the images on command (to e.g. resolve https://github.com/Leaflet/Leaflet/issues/8086, and to keep the diagram updated)?

Falke-Design (changes requested):
## Release & Generation

The image generation needs to be done in a separate step and should be up-to-date with the release. So there should be a explanation in [RELEASE.md](https://github.com/Leaflet/Leaflet/blob/main/RELEASE.md) how to generate the diagram and where it should be placed.

I tried a little bit arround and came to following solution (on Windows). But I'm far away to understand graphviz and all the possiblities. 
```
dot ./docs/class-diagram-bare.dot -Tpng -Gdpi=200 -o ./docs/SlavaUkraini/examples/extending/class-diagram.png
```
The command from above could be also a node script command.

So I suggest to add in the RELEASE.md:
1. Generate class-diagram over cmd `dot ./docs/class-diagram- ....`
2. Change the bounds in `./docs/SlavaUkraini/examples/extending/class-diagram.md` to match with the generated image.

#### Available in top level of docs

I think with the possibilities now, it would be nice to have in `./docs` the generated diagrams placed.

## Example:

I suggest to add `background: #fff;` in `class-diagram.md` because the generated png is not transparent anymore:
grafik

Also I think a change from `minZoom: -4` to `-2` would be good but that is not very important.

## Design:

- What is the meaning of this arrow? It makes not really sense for me but that can be caused of my lack of diagram knowledge.
grafik


- This now a very subjectiv thing but I like the prefix `L` (`L.Class`)  in the diagram.

## Documentation

Is there a documentation how the `.hbs` files needs to be created / used?

BadrAlzein:
Hello there, 

for a module at my university i have to make a contribution to this project and i saw this diagram that is great and i mad an editable version of it using open source uml editing website (draw.io)
here is it as HTML 
[link in gituhb](https://github.com/gernhard1337/Leaflet/blob/SE-GA/submission/phase%203/verbesserung%20von%20teil%202/final.html)
here is the editable version 
[editable version as .draw.io](https://raw.githack.com/gernhard1337/Leaflet/SE-GA/submission/phase%203/verbesserung%20von%20teil%202/final.drawio)

Please let me know if this addition make sense for future changes in the structure so i can create a pull request. thanks you so much. an accepted pull request will let me pass this module :)

best regards
Badr

Falke-Design:
@BadrAlzein thank you for your work but the purpose of this PR is that we want a UML diagram which is generated over the inline documentation and don't need maintenance work, like your draw.io version. We will keep working on this PR, only a few adjustments are open.

You can contribute for example with creating a few tests, look for this into #8145 and #8206. [Contribution docu](https://github.com/Leaflet/Leaflet/blob/main/CONTRIBUTING.md#running-the-tests)

Falke-Design:
@IvanSanchez maybe we can finish this in the near future?

IvanSanchez:
> @IvanSanchez maybe we can finish this in the near future?

I'll try to have a look during the upcoming days.

IvanSanchez:
I've tweaked the docs build system so now there's a step to (re-)generate the UML diagram image, via the `graphviz` NPM package (https://www.npmjs.com/package/graphviz). It's an incomplete javascript implementation, but it should work for the job.

> I suggest to add `background: #fff;` in `class-diagram.md`

No need, since the image now does have a white background.

> What is the meaning of this arrow?

That's [aggregation](https://en.wikipedia.org/wiki/Class_diagram#Aggregation). A `Marker` contains an icon; a `LayerGroup` contains zero or more `Layer`s, a `Map` contains one `CRS`, etc.

> Is there a documentation how the `.hbs` files needs to be created / used?

That's [handlebars tempaltes](https://handlebarsjs.com/). It's a Leafdoc thing.

Falke-Design (changes requested):
I get an error:
grafik

Windows 10

PR #8105, opened 2022-04-02
Small typo in map dragging option description

adam-lynch:


Falke-Design (changes requested):
Please revert your change and only change it in this [line](https://github.com/Leaflet/Leaflet/blob/main/src/map/handler/Map.Drag.js#L17)

adam-lynch:
OK done.

By the way, that was my first time using a "Codespace" to fix up a PR and it was amazing;

image

It launched a VSCode-like instance in a browser tab. I edited from there and committed using the "VS Code" terminal

mourner (approved):

PR #8109, opened 2022-04-03
Add a Ukrainian flag to default attribution 🇺🇦 

mourner:
Add a 🇺🇦 flag to the default attribution prefix (where SVG is supported). Just a tiny, unobtrusive way to bring more attention to Leaflet's Ukrainian background across websites that use it.

image

### DIFF
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -4,6 +4,7 @@ import {Map} from '../map/Map';
 import * as Util from '../core/Util';
 import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
+import Browser from '../core/Browser';
 
 /*
  * @class Control.Attribution
@@ -21,8 +22,8 @@ export var Attribution = Control.extend({
 
 		// @option prefix: String|false = 'Leaflet'
 		// The HTML text shown before the attributions. Pass `false` to disable.
-		prefix: '<a href="https://leafletjs.com" title="A JavaScript library for interactive maps">Leaflet</a>'
-
+		prefix: '<a href="https://leafletjs.com" title="A JavaScript library for interactive maps">' +
+				(Browser.emoji ? '&#127482;&#127462; ' : '') + 'Leaflet</a>'
 	},
 
 	initialize: function (options) {
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -6,7 +6,7 @@ import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
 import Browser from '../core/Browser';
 
-const ukrainianFlag = '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';
+var ukrainianFlag = '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';
 
 /*
  * @class Control.Attribution
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -6,6 +6,8 @@ import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
 import Browser from '../core/Browser';
 
+const ukrainianFlag = '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';
+
 /*
  * @class Control.Attribution
  * @aka L.Control.Attribution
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

MaxymVlasov (approved):

Malvoz:
The flag doesn't render for me, guess I don't have the necessary emoji font installed:

ua

But I think this is fine too.

mourner:
@Malvoz does the emoji render for you on https://emojipedia.org/flag-ukraine/? I might need to specify a different font set in CSS. I'll look into SVG as an alternative too — good point!

Falke-Design:
I would like to have a new option (maybe on the map while initialization) which is per default true and enables the flag. It should be easy to enable / disable the flag. Right now we force sites to be politcal also that ones which have in theier statutes to be not politcal. In this case it should be easy to disable it.

mourner:
> I would like to have a new option (maybe on the map while initialization) which is per default true and enables the flag. It should be easy to enable / disable the flag. Right now we force sites to be politcal also that ones which have in theier statutes to be not politcal. In this case it should be easy to disable it.

@Falke-Design If you still think this is a political issue, I don't know what to say. It's not about politics, it's about basic humanity. Those who can't have Leaflet attribution can still do that with:

```js
map.attributionControl.setPrefix('')
```

Malvoz:
> @Malvoz does the emoji render for you on https://emojipedia.org/flag-ukraine/?

It doesn't. I'm currently on Windows 10.

Falke-Design:
I thought that the attribution `Leaflet` is a must have and then clearing the prefix would be not working. But while a quick search I didn't find any must have reference to Leaflet so `setPrefix('')` is more then enough.

> If you still think this is a political issue, I don't know what to say. It's not about politics, it's about basic humanity.

Yes you are right, I didn't wanted to understate this. I appreciate this change, I only wanted to be sure that the possiblity to don't show it, is given. And it is with `setPrefix`.

MaxymVlasov:
> > @Malvoz does the emoji render for you on [emojipedia.org/flag-ukraine](https://emojipedia.org/flag-ukraine/)?
>
> It doesn't. I'm currently on Windows 10.

Interesting, I'm on Win10 too and it works in Firefox but does not work in Chrome, Edge, and IE11.

src/control/Control.Attribution.js
```diff
@@ -6,6 +6,8 @@ import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
 import Browser from '../core/Browser';
 
+const ukrainianFlag = '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';
```
Malvoz, line 9: Do we want to label this for a11y reasons?

e.g. using SVG `<title>` as a child of `<svg>`, or the global `title` attribute? Both these produce tooltips (only on hover in most browsers).

Otherwise SVG `<desc>` can be used, or global `aria-label`.

Malvoz: Or else set to `<svg>` to `aria-hidden="true"`, as some screen readers may announce "Graphic" or similar.

mourner: I think no, it's pure decoration — people following the link will get the meaning on the website anyway.

mourner:
@Malvoz @MaxymVlasov replaced the emoji with an SVG equivalent (and updated the screenshot above).

src/control/Control.Attribution.js
```diff
@@ -4,6 +4,9 @@ import {Map} from '../map/Map';
 import * as Util from '../core/Util';
 import * as DomEvent from '../dom/DomEvent';
 import * as DomUtil from '../dom/DomUtil';
+import Browser from '../core/Browser';
+
+var ukrainianFlag = '<svg xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';
```
Malvoz, line 9:
```suggestion
var ukrainianFlag = '<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" width="12" height="8"><path fill="#4C7BE1" d="M0 0h12v4H0z"/><path fill="#FFD500" d="M0 4h12v3H0z"/><path fill="#E0BC00" d="M0 7h12v1H0z"/></svg>';

```

Malvoz: I think we should hide it from ATs per https://github.com/Leaflet/Leaflet/pull/8109#discussion_r841212780.

Malvoz (approved):

junkym0nk3y:
> > I would like to have a new option (maybe on the map while initialization) which is per default true and enables the flag. It should be easy to enable / disable the flag. Right now we force sites to be politcal also that ones which have in theier statutes to be not politcal. In this case it should be easy to disable it.
> 
> @Falke-Design If you still think this is a political issue, I don't know what to say. It's not about politics, it's about basic humanity. Those who can't have Leaflet attribution can still do that with:
> 
> ```js
> map.attributionControl.setPrefix('')
> ```

No, it's kind of political in Russia. Even if every single web developer will go to raid Kremlin - in another day half of them will be in prison and will be torture the same way (for 10 years for sure - look at "Russian fake news laws" in Wiki and article "A culture of torture and fear is rife in Russia's prison" on dw.com). 

I don't want to be devil's advocate, and I hate this regime too, but I just want to say - that we are people too and we are trying to survive, don't put us at risk of losing this life - even it just live in totalitarian gulag.

MaxymVlasov:
> > > I would like to have a new option (maybe on the map while initialization) which is per default true and enables the flag. It should be easy to enable / disable the flag. Right now we force sites to be politcal also that ones which have in theier statutes to be not politcal. In this case it should be easy to disable it.
> > 
> > 
> > @Falke-Design If you still think this is a political issue, I don't know what to say. It's not about politics, it's about basic humanity. Those who can't have Leaflet attribution can still do that with:
> > ```js
> > map.attributionControl.setPrefix('')
> > ```
> 
> No, it's kind of political in Russia. Even if every single web developer will go to raid Kremlin - in another day half of them will be in prison and will be torture the same way (for 10 years for sure - look at "Russian fake news laws" in Wiki and article "A culture of torture and fear is rife in Russia's prison" on dw.com).
> 
> I don't want to be devil's advocate, and I hate this regime too, but I just want to say - that we are people too and we are trying to survive, don't put us at risk of losing this life - even it just live in totalitarian gulag.

1. That is not related to politics at all, it is about pride to be a Ukrainian product.
2. russians had more than 20 years to stop things that happened now.
3. If you have any "political problems" in a territory where you live with Leaflet or any other product - you always can choose "импортозамещение".

Malvoz:
It's not healthy to the Leaflet community to continue this discussion here. If you don't agree with the 🇺🇦 flag in the attribution, remove it, or write yourself a blog post about it.

PR #8110, opened 2022-04-03
Pin actions to a full length commit SHA

naveensrinivasan:
- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

[How do I validate these pinned actions?](https://gist.github.com/naveensrinivasan/ca008c07279176acce28969fb77d056f)

Also, dependabot supports upgrading based on SHA. https://github.com/ossf/scorecard/pull/1700

GitHub's own repository pin's their checkout actions by SHA and doesn't use the version tag
https://github.com/github/docs/blob/ea7f218c91ecbae9a700a8702b51a7d2736e0d2c/.github/workflows/docs-review-collect.yml#L23

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,10 +10,10 @@ jobs:
         os: [ubuntu-latest, windows-2019]
     steps:
       - name: Check out repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # v3
 
       - name: Set up Node
-        uses: actions/setup-node@v3.1.0
+        uses: actions/setup-node@5b52f097d36d4b0b2f94ed6de710023fbb8b2236 # v3.1.0
         with:
           node-version: ${{ env.NODE_VERSION }}
           check-latest: true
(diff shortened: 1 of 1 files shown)


mourner:
I'm not sure about this personally. It makes the actions setup more verbose, and harder to follow which action version is used where, to mitigate risk that I'd consider negligible specifically in context of how and which actions are used in Leaflet.

jonkoops:
I'm against this for several reasons:

1. Increase in the maintenance burden, as we now have to update our actions more regularly.
2. Too easy to make mistakes with versions of the same action due to the cognitive overhead of parsing commit hashes as a human.
3. Most of these actions are from Github themselves which I consider trustworthy.

Closing this PR as I just don't see the advertised benefits of this as worth the downsides.

PR #8119, opened 2022-04-05
Fix potential Ukrainian flag CSS conflicts

mourner:
Fixes an attribution issue reported in https://github.com/Leaflet/Leaflet/discussions/8044#discussioncomment-2508676.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -423,6 +423,9 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-control-attribution a:focus {
 	text-decoration: underline;
 	}
+.leaflet-control-attribution svg {
+	display: inline !important;
+	}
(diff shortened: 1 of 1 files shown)


dist/leaflet.css
```diff
@@ -423,6 +423,9 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-control-attribution a:focus {
 	text-decoration: underline;
 	}
+.leaflet-control-attribution svg {
```
Malvoz, line 426: We could give it a unique class, but how often do people use SVGs in their attribution, and change their `display`?

Malvoz (approved):

dist/leaflet.css
Malvoz: (as this would override other SVGs in the attribution that have their `display` set to other than `inline`)

PR #8123, opened 2022-04-07
Add minified stylesheet to distribution

### LINKED ISSUE #4737: CSS not minified on CDNs
Hi,

is it possible to minimize the CSS before it's pushed to the CDNs?

Thanks
Rob

jonkoops:
Closes #4737


mourner (changes requested):
As you can see by the diff in package-lock, CSSNano + PostCSS is a pretty bloated ecosystem which I'd like to avoid. The fact PostCSS is Russian-made doesn't help either.

Maybe we could try [CleanCSS](https://github.com/clean-css/clean-css) (and corresponding [CLI](https://github.com/clean-css/clean-css-cli)) instead? It's small and has no dependencies on preprocessors.

.gitignore
```diff
@@ -8,6 +8,7 @@ _site
 *.sublime-*
 _site
 dist/*.js
+dist/*.min.css
```
Malvoz, line 11: I'm not trying to complicate things, but now there are both non-`.min` (`leaflet.js`) and `.min` (`leaflet.min.css`) files that are minified/uglified. I wonder:

- If the minified CSS is `leaflet.min.css`, should all instances of `leaflet.css`  be changed to `leaflet.min.css` in [download.md](https://github.com/Leaflet/Leaflet/blob/918b65f1cc2dea81dbdfa67b676602524c57ee23/docs/SlavaUkraini/download.md?plain=1) (and maybe other places)?
- Or, should there be a `src/leaflet.css` to output the minified version as `dist/leaflet.css`? I think this has the benefit that there's no need to change references from `leaflet.css` to `leaflet.min.css`.

jonkoops:
> Maybe we could try [CleanCSS](https://github.com/clean-css/clean-css) (and corresponding [CLI](https://github.com/clean-css/clean-css-cli)) instead? It's small and has no dependencies on preprocessors.

It looks like this package is in maintenance mode, so I don't know if it makes sense to use this as an alternative.

> The fact PostCSS is Russian-made doesn't help either.

I'd possibly agree with this with if it was made by a Russian organization that publicly supports the war. However @ai has publicly renounced the war, and PostCSS seems to have a diverse group of contributors.

The argument of PostCSS being a large dependency is certainly valid, but we'd have to have a maintained alternative.

.gitignore
jonkoops: This is a valid comment, I chose to keep the PR simple for now to start this discussion. We could certainly change these references if that would make sense.

We could also first land this change and then start a discussion on if and how we further integrate it.

IvanSanchez:
I, for one, would like to bring https://github.com/Leaflet/Leaflet/pull/4586 back from the dead. If we're adding tooling, might as well get some code separation from it.

> If the minified CSS is leaflet.min.css, should all instances of leaflet.css be changed to leaflet.min.css

I'd say: follow the current convention of `dist/leaflet-src.js` vs `dist/leaflet.js`. Unminified CSS goes into `dist/leaflet-src.css`, minified goes into `dist/leaflet.css`.

IvanSanchez:
> > Maybe we could try [CleanCSS](https://github.com/clean-css/clean-css) (and corresponding [CLI](https://github.com/clean-css/clean-css-cli)) instead? It's small and has no dependencies on preprocessors.
> 
> It looks like [CleanCSS] is in maintenance mode, so I don't know if it makes sense to use this as an alternative [to CSSNano+PostCSS].

I have no problems with packages that haven't had a release in some time. I mean, look at Leaflet. From my perspective, it means the stuff is reliable and the foundations are not made out of quicksand changing every two weeks.

There seems to be https://www.npmjs.com/package/csso also - under active development, and just 5 dependencies for a total of ~3.4MiB in your `node_modules`. It might be a good compromise solution.

ai:
I can recommend Parcel CSS (which can be used without Parcel, it is a stand-alone tool).

It is much faster (written on Rust). However, it is written by US developers and US is the same bloody-based empire (but I am bad with this nationalism way of thinking since I do not support this way of thinking).

.gitignore
Malvoz: To clarify: per https://github.com/Leaflet/Leaflet/pull/8123#issuecomment-1091610570, the idea is now to rename the current `dist/leaflet.css` file to `dist/leaflet-src.css` and generate the minified `dist/leaflet.css` (and not change any references in the docs).

mourner:
> I have no problems with packages that haven't had a release in some time. I mean, look at Leaflet. From my perspective, it means the stuff is reliable and the foundations are not made out of quicksand changing every two weeks.

Agreed. To expand on this, last release of Clean-CSS is 7 days ago, and the author [explains "maintenance mode"](https://github.com/clean-css/clean-css/discussions/1209) to mean "no new features because existing ones are good enough, and no support for old versions", which matches Leaflet's philosophy as well.

I'm still opposed to complicating CSS maintenance by splitting into multiple files, etc. What we have is good enough. A small prepublishOnly minification step is simple enough. Although I'm not sure how much it will benefit end users — they usually have pipelines for app bundle minification anyway, so a minified Leaflet CSS will only be relevant when used the old way, included as a `link` style tag, and even then it will be a negligible difference when taking gzip into account.

jonkoops:
I've replaced PostCSS with Parcel CSS and the output looks good. Perhaps someone here can independently verify that the CSS changes are not negatively affecting anything in an application?

jonkoops:
What are the opinions on changing the source file's name to `leaflet-src.css` and making `leaflet.css` minified by default? @Malvoz and @IvanSanchez proposed this, and I am neutral on this subject.

mourner:
> What are the opinions on changing the source file's name to `leaflet-src.css` and making `leaflet.css` minified by default?

Yes, I think we should be consistent here between CSS & JS.

One last thing I'm a little worried about is whether Parcel minification could trip up legacy browsers that we formally still support — e.g. some of the minification tricks could only work on modern browsers, we might need to test this.

.gitignore
jonkoops: Done :+1:

jonkoops:
I've updated the CSS file to use the `-src` suffix, this also means that all references in the project are now using the minified version.

Perhaps someone can test these changes in Internet Explorer (as this is the most likely browser to trip up). I can't install it as I don't have any machine that runs Windows.

We should also think about a couple of other things:
1. Do we want to include a source-map (as we already do this for the JavaScript files)
2. Should the `style` entry-point in the package refer to `leaflet-src.css`? https://github.com/Leaflet/Leaflet/blob/25546fc04fbff9c5fec26f1cb42bda710be8c7ae/package.json#L34

I am leaning to yes on both of these, WDYT?

Falke-Design:
The differences that I found:
Last style has no `;` but in my eyes this should be generated too:
```diff
.leaflet-safari .leaflet-tile-container {
  width: 1600px;
  height: 1600px;
-  -webkit-transform-origin: 0 0;
+ -webkit-transform-origin: 0 0
}
```
```diff
/* Prevents IE11 from highlighting tiles in blue */
.leaflet-tile::selection {
-  background: transparent;
+ background: 0 0
}
```
rgb changed to hex, and color `black` to `#000`
```diff
.leaflet-container a {
-  -webkit-tap-highlight-color: rgba(51, 181, 229, 0.4);
+ -webkit-tap-highlight-color: #33b5e566
}
```

**BIG Problem:** every font-size fallback is removed (same for backgrond & margin)
```diff
.leaflet-container {
  font-family: Helvetica Neue, Arial, Helvetica, sans-serif;
-  font-size: 12px;
-  font-size: 0.75rem;
+  font-size: .75rem;
  line-height: 1.5
}
```
```diff
.leaflet-bar a,
.leaflet-control-layers-toggle {
-  background-position: 50% 50%;
+  background-position: 50%;
  background-repeat: no-repeat;
  display: block;
}
```
```diff
.leaflet-bar a:last-child {
- border-bottom: none;
+  border-bottom: 0;
  border-bottom-left-radius: 4px;
  border-bottom-right-radius: 4px
}
```
```diff
.leaflet-control-zoom-in,
.leaflet-control-zoom-out {
-  font: bold 18px 'Lucida Console', Monaco, monospace;
+  font: 700 18px Lucida Console, Monaco, monospace
  text-indent: 1px;
}
```
```diff
.leaflet-control-layers-scrollbar {
-  overflow-y: scroll;
-  overflow-x: hidden;
+  overflow: hidden scroll
  padding-right: 5px;
}
```
Background fallbacks are removed:
```diff
.leaflet-container .leaflet-control-attribution {
-  background: #fff;
-  background: rgba(255, 255, 255, 0.8);
+  background: #fffc;
  margin: 0;
}
```
```
.leaflet-oldie .leaflet-popup-tip {
  width: 24px;
  margin: 0 auto;

-  -ms-filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678)";
-  filter: progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678);
+  -ms-filter: "progid:DXImageTransform.Microsoft.Matrix(M11=0.70710678, M12=0.70710678, M21=-0.70710678, M22=0.70710678)";
+  filter: progid:DXImageTransform.Microsoft.Matrix(M11=.707107, M12=.707107, M21=-.707107, M22=.707107);
}
```

everything else is same, order of styles, attribution logic `:not()`, `:hover`, ...

mourner:
@Falke-Design yikes. Looks like we'll likely want to postpone this until v2.0 at least, since it's guaranteed to regress on IE. ☹️

Falke-Design:
Maybe it is possible to fix over the config

IvanSanchez:
> We should also think about a couple of other things:

Also, the changes should be reflected in the `download.md` webpage - right now, it tells users the difference between `leaflet-src.js` and `leaflet.js`, and lists the SHA-sums of the files for a release. The addition of `leaflet-src.css` should be reflected there.

jonkoops:
@Falke-Design Nice comparison, that is actually super useful thank you!

devongovett:
Hey, I work on Parcel CSS and stumbled across this PR. Glad to see you're trying it out.

I think you should set the --targets option so that the output is compiled for the proper browsers. When no targets are set we assume modern browsers. That should help with a few of the differences you listed. 

Let me know if I can help with anything else. 🙂

Falke-Design:
@devongovett I don't get it to run with `targets` 😕

When I run `parcel-css --minify --targets '>= 0.25%' dist/leaflet-src.css -o dist/leaflet.css`, I get following error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Nom("")', src/main.rs:85:59
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We currently support IE7, so would it be correct to use `--targets 'ie7'`?

jonkoops:
I'll try to get back to this PR soon, currently have some pressing matters.

jonkoops:
I suggest we push this back to v2. That way we can be sure this won't introduce any breaking changes to older browsers.

Issue #8131, opened 2022-04-11
Popup should have a min-height


Falke-Design:
The close button overflows the popup container if the container is empty:

![grafik](https://user-images.githubusercontent.com/19800037/162693049-187cd57f-6e05-4ab6-af11-6f83da77e72c.png)

```
var marker = new L.Marker(latlng).addTo(map);
marker.bindPopup("");
```

Falke-Design:
Hi @MassiGy,

thank you for your suggestion but it should be a "static" value in the leaflet.css file. 

The problem with your approach is that it needs to be explicit set, which will result in not adding the option because normally a popup has content and then the option `minHeight` is not needed. And if then the case happens that it has no content, the option `minHeight` is not set.

Malvoz:
@MassiGy You could utilize the [`:empty`](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) pseudo-class. Also, the `min-height` doesn't have to be that excessive. Use the smallest `min-height` that would fix the issue.

@Falke-Design Is it really necessary to fix this? Seems like an enhancement for development only. If popups are empty in production that's the wrong use of a popup.

Falke-Design:
It's not important but I think it would be good to have it, because it is clearly an UI topic which comes from Leaflet.

ansh-ag:
Looks like a straight forward change, picking this up guys.

PR #8136, opened 2022-04-12
Popup should have a min-height

ansh-ag:
Fixes: https://github.com/Leaflet/Leaflet/issues/8131

[](https://user-images.githubusercontent.com/19800037/162693049-187cd57f-6e05-4ab6-af11-6f83da77e72c.png)https://user-images.githubusercontent.com/19800037/162693049-187cd57f-6e05-4ab6-af11-6f83da77e72c.png

### Proposed Change :
Have added a min-height element is empty.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -488,6 +488,9 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	margin: 17px 0;
 	margin: 1.3em 0;
 	}
+.leaflet-popup-content:empty {
+		min-height: 1px;
+	}
 .leaflet-popup-tip-container {
 	width: 40px;
 	height: 20px;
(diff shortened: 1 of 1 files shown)


Falke-Design:
Like @Malvoz said:
> Edit: OTOH there could be really small content in popups, that this would fix. In which case :empty is not a good solution.

dist/leaflet.css
```diff
@@ -488,6 +488,9 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	margin: 17px 0;
 	margin: 1.3em 0;
 	}
+.leaflet-popup-content:empty {
```
Malvoz, line 491: The `:empty` pseudo-class is not desirable, https://github.com/Leaflet/Leaflet/issues/8131#issuecomment-1095181697:

> OTOH there could be really small content in popups, that this would fix. In which case `:empty` is not a good solution.

Falke-Design (approved):
grafik

Malvoz (approved):

PR #8144, opened 2022-04-16
Cover Rectangle with unit Tests

### LINKED ISSUE #8140: Rectangle tests are missing
There is not much difference to a polygon but the initalization and `setBounds` should be tested.

https://leafletjs.com/SlavaUkraini/reference.html#rectangle

stephenspol:
Tests cover `initialize `and `setBounds`. Most tests mirrored `PolygonSpec.js` but had to drop a few tests. Some tests that were mirrored from `Polygon` now fail such as an empty array fails when initializing and setting bounds. I am assuming this is because of `_boundsToLatLngs` being overwritten as `lat` throws an error due to being undefined. Rectangle also fails when putting more than 2 coords, not sure if this is attended.

Please let me know if anything needs to be changed. 

PR closes #8140.

### DIFF
--- a/spec/suites/layer/vector/RectangleSpec.js
+++ b/spec/suites/layer/vector/RectangleSpec.js
@@ -33,35 +33,21 @@ describe.only('Rectangle', function () {
 			expect(rectangle._latlngs).to.not.eql(sourceLatLngs);
 		});
 
-		it("cannot be called with an empty array", function () {
+		it.skip("cannot be called with an empty array", function () {
 			// Throws error due to undefined lat
-			expect(L.rectangle).withArgs([]).to.throwException();
+			expect(L.rectangle([])).to.throwException();
 		});
 
-		it("can be initialized with holes", function () {
+		it("can be initialized with extending bounds", function () {
 			var originalLatLngs = [
-				[[0, 10]], // external ring
-				[[20, 30]] // hole
+				[0, 10], [20, 30],
+				[40, 50], [60, 70] // extended bounds
 			];
 
 			var rectangle = L.rectangle(originalLatLngs);
 
 			expect(rectangle._latlngs).to.eql([
-				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
-			]);
-			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
-		});
-
-		it("cannot be initialized with more than 2 coords", function () {
-			var originalLatLngs = [
-				[0, 10], [20, 30],
-				[2, 3], [4, 5] // extra coords
-			];
-
-			var rectangle = L.rectangle(originalLatLngs);
-
-			expect(rectangle._latlngs).to.not.eql([
-				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
+				[L.latLng([0, 10]), L.latLng([60, 10]), L.latLng([60, 70]), L.latLng([0, 70])]
 			]);
 			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
 		});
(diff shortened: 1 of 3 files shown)


spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+
+			var rectangle = L.rectangle(originalLatLngs);
+			rectangle.setBounds(newLatLngs);
+
+			expect(rectangle._latlngs).to.eql([
+				[L.latLng([5, 6]), L.latLng([7, 6]), L.latLng([7, 8]), L.latLng([5, 8])]
+			]);
+
+			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
+		});
```
stephenspol, line 104: Let me know if this test is needed.

spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+
+			var rectangle = L.rectangle(sourceLatLngs);
+
+			expect(sourceLatLngs).to.eql(originalLatLngs);
+			expect(rectangle._latlngs).to.not.eql(sourceLatLngs);
+		});
+
+		it("cannot be called with an empty array", function () {
+			// Throws error due to undefined lat
+			expect(L.rectangle).withArgs([]).to.throwException();
```
Falke-Design, line 38:
```suggestion
			expect(L.rectangle([])).to.throwException();
```

spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+				[[20, 30]] // hole
+			];
+
+			var rectangle = L.rectangle(originalLatLngs);
+
+			expect(rectangle._latlngs).to.eql([
+				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
+			]);
+			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
+		});
```
Falke-Design, line 53:
```suggestion
```

I would say that we remove this test, because holes are not the wanted usage.

spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+			];
+
+			var rectangle = L.rectangle(originalLatLngs);
+			rectangle.setBounds(newLatLngs);
+
+			expect(rectangle._latlngs).to.eql([
+				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
+			]);
+			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
+		});
```
Falke-Design, line 123:
```suggestion
```

same as above

spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+				[2, 3], [4, 5] // extra coords
+			];
+
+			var rectangle = L.rectangle(originalLatLngs);
+
+			expect(rectangle._latlngs).to.not.eql([
+				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
+			]);
+			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
+		});
```
Falke-Design, line 67: Not correct. If multiple latlngs are passed to the rectangle, all latlngs will be covered by the rectangle bounds. Means each latlng is extending the bounds.

spec/suites/layer/vector/RectangleSpec.js
```diff
@@ -0,0 +1,144 @@
...
+
+			var rectangle = L.rectangle(originalLatLngs);
+			rectangle.setBounds(newLatLngs);
+
+			expect(rectangle._latlngs).to.not.eql([
+				[L.latLng([0, 10]), L.latLng([20, 10]), L.latLng([20, 30]), L.latLng([0, 30])]
+			]);
+			expect(rectangle.getLatLngs()).to.eql(rectangle._latlngs);
+		});
+	});
```
Falke-Design, line 143: Not correct. Same as above

Falke-Design (changes requested):
Thank you for your contribution!

spec/suites/layer/vector/RectangleSpec.js
Falke-Design: yes, keep it

spec/suites/layer/vector/RectangleSpec.js
stephenspol: I tried to do this previously but got the following error message:

```
Chrome Headless 99.0.4844.84 (Windows 10) Rectangle #initialize cannot be called with an empty array FAILED
        TypeError: Cannot read properties of undefined (reading 'lat')
            at LatLngBounds.getNorth (src/Leaflet.js:1260:28)
            at LatLngBounds.getNorthWest (src/Leaflet.js:1230:28)
            at NewClass._boundsToLatLngs (src/Leaflet.js:13092:19)
            at NewClass.initialize (src/Leaflet.js:13079:50)
            at new NewClass (src/Leaflet.js:297:22)
            at Object.rectangle (src/Leaflet.js:13102:11)
            at Context.<anonymous> (spec/suites/layer/vector/RectangleSpec.js:38:13)
```
Any suggestions on how to fix this?

Falke-Design:
```
expect(function () {
    L.rectangle([]);
}).to.throwException();
```

Falke-Design (approved):

Issue #7810, opened 2021-11-25
Release Leaflet v1.8.0


mourner:
We long overdue for a v1.8.0 release, and I've cleaned up the process so that it's much easier now. The remaining question is — which of the outstanding PRs we should definitely land for the release? @johnd0e @Falke-Design — I created a `v1.8` milestone, feel free to add must-have PRs to that it would be easier for me to review.

Falke-Design:
Without #7026 we don't need to release 1.8. it is the number 1 issue 😄

johnd0e:
> which of the outstanding PRs we should definitely land

Added also a couple of important but unresolved issues,

Falke-Design:
Is ok for me but can you please create a milestone where we can move the issues / pRs for 1.8.1? To keep them on track, what we should check after 1.8.0

mourner:
@Falke-Design good point, created a 1.8.1 milestone.

Falke-Design:
@mourner can we keep merging the 1.8.0 milestone?
Ready to check would be: #7782, #7855

@johnd0e can you please go through the milestone and mark the PRs as "Ready for review" (not draft) and remove not necessary PRs.

Falke-Design:
@mourner it would be nice if we can release 1.8 with march, what do you think? I remove as much as possible from the 1.8.0 milstone.

Fast and easy to check:
 -  #7763

Needs a decission:
 - #7302 (maybe it is enough to say if this topic is not important for 1.8.0)

Needs more time:
 - #7532
 - #7446
 - #7965

@IvanSanchez, @jonkoops, @johnd0e, @Malvoz help and approving PRs would be nice. Are you interesseted to make an appointment get rid of it?

Malvoz:
> help and approving PRs would be nice

I'm mostly comfortable reviewing a11y-related PRs.

mourner:
> it would be nice if we can release 1.8 with march, what do you think?

Yes! Maybe we could even release in the nearest weeks. Unless my home is invaded by Russia, but I truly hope not.

mourner:
> * #7302 (maybe it is enough to say if this topic is not important for 1.8.0)

Moved to v1.8.1, this doesn't seem critical to fix — there are easy workarounds.

> * #7532

Resolved the conflicts there, @Falke-Design can you take another look at it to see if we can merge?

Falke-Design:
If you have currently time there would be some other easy to merge PRs:
- #7814
- #7956
- #7914
[...]

PR #8150, opened 2022-04-18
Update website & docs for v1.8.0

mourner:


Malvoz:
@mourner Sorry for the slow response, consider this my approval, since I can't actually press "Approve" after the fact. 😋

I believe we should close and unpin https://github.com/Leaflet/Leaflet/issues/7810 now, and lock(?) and unpin https://github.com/Leaflet/Leaflet/discussions/8044(?).

Falke-Design:
> I believe we should close and unpin https://github.com/Leaflet/Leaflet/issues/7810 now, and lock(?) and unpin https://github.com/Leaflet/Leaflet/discussions/8044.

Done

PR #8160, opened 2022-04-19
Fix regression about popup close button modifying the URL

### LINKED ISSUE #8159: Popup close button adding #close hash to URL
Starting with version v1.8.0, clicking the popup close 'x' button adds #close to the URL

Behavior is reproducible on the leaflet.js website in examples with popups such as:
https://leafletjs.com/SlavaUkraini/examples/quick-start/example.html

Previous to v1.8.0 clicking the popup close button prevented #close hash from being added to the URL

Not ideal for Single Page Applications that rely on hash routing.

IvanSanchez:
Fixes #8159

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -203,7 +203,10 @@ export var Popup = DivOverlay.extend({
 			closeButton.href = '#close';
 			closeButton.innerHTML = '<span aria-hidden="true">&#215;</span>';
 
-			DomEvent.on(closeButton, 'click', this.close, this);
+			DomEvent.on(closeButton, 'click', function (ev) {
+				DomEvent.preventDefault(ev);
+				this.close();
+			}, this);
 		}
 	},
 


Falke-Design (approved):
I don't know if we should use `DomEvent.stop(e)` to prevent propagation.

Should we make a named functions instead of a anonymus function? The developers would have access to the function for plugins / overwrites.

IvanSanchez:
> I don't know if we should use `DomEvent.stop(e)` to prevent propagation.

I'm not 100% sure. On one hand, and after giving this a bit of thought, I'm partial to go back to how things were before #7439, and use `DomEvent.stop(ev)`. I cannot see how propagating the click event from the close button to the popup container can be useful.

On the other hand, I don't know if @johnd0e has a good use case about clicks on the popup close button being propagated. And, there's a `disableClickPropagation()` call being made on the popup container anyway, so there's no risk of propagating the event to the map container.

mourner:
Let's do away with just `preventDefault` for the fix, and look into propagation separately, since it's not causing any apparent issues at the moment.

IvanSanchez:
I think this is good "as is", then. No plugin author complained about the lack of a `_onCloseButton` method, and the `preventDefault()` should suffice.

PR #8161, opened 2022-04-19
Feature / Fix: Extend Events.listens to search for specific function

### LINKED ISSUE #7951: "Listener not found" warning when passing maxBounds option to map
**Steps to reproduce**

1. Open the "debug/map/max-bounds.html" on current `main` branch
2. Look in the browser's JS console

**Expected behavior**

No error or warning messages should be present in the JS console.

**Current behavior**

A spurious "Listener not found" warning is logged by Leaflet. The warning is "harmless" in that it does not appear to break any map behavior, but it is noise for developers using Leaflet and a potential troubleshooting distraction to anyone else who might be peeking into the JS console on a project.

Screenshot of demo map with the warning visible underneath via open developer tools pane

**Environment**
- Leaflet version: testing at commit `409c17c` (which is unrelated to this issue, just happens to be at tip)
- Browser (with version): presumedly any, appears to be a logic issue
- OS/Platform (with version): presumedly any

**Additional context**

The warning and/or the conditions leading to it seem to have been introduced after the last official (1.7.1) release but I did not bisect to determine the exact commit.  It appears that the first call to `setMaxBounds` tries to unregister a `this._panInsideMaxBounds` listener before it has ever registered one.

The "debug/map/max-bounds.html" example in this repo is sufficient.

Falke-Design:
```
@method listens(type: String, fn: Function, context?: Object, propagate?: Boolean): Boolean
```

With extending the `listens` function, we can check now if the `_panInsideMaxBounds` listener is existing before removing it. Fixes #7951
```
if (this.listens('moveend', this._panInsideMaxBounds)) {
	this.off('moveend', this._panInsideMaxBounds);
}
```


src/core/Events.js
```diff
@@ -216,19 +216,47 @@ export var Events = {
...
+				// find fn
+				for (var i = 0, len = listeners.length; i < len; i++) {
+					var l = listeners[i];
+					if (l.ctx !== ctx) {
+						continue;
+					}
+					if (l.fn === fn) {
+						return true;
+					}
+				}
```
mourner, line 250: Nit: we can write this in a more compact way for clarity:

```suggestion
				var ctx = context === this ? undefined : context;
				for (var i = 0, len = listeners.length; i < len; i++) {
					if (listeners[i].ctx === ctx && listeners[i].fn === fn) {
						return true;
					}
				}
```

Falke-Design: I copied it from:
https://github.com/Leaflet/Leaflet/blob/118ca443509476beec7f99e6d24694e4b9c1a6f2/src/core/Events.js#L164-L179

I suggest to use following from `_on` (same as your suggestion) and then use the new `listens` in `_on`
https://github.com/Leaflet/Leaflet/blob/118ca443509476beec7f99e6d24694e4b9c1a6f2/src/core/Events.js#L120-L125

mourner: Sounds good to me.

src/core/Events.js
```diff
@@ -216,24 +212,55 @@ export var Events = {
...
 			for (var id in this._eventParents) {
-				if (this._eventParents[id].listens(type, propagate)) { return true; }
+				if (this._eventParents[id].listens(type, fn, context, propagate)) { return true; }
 			}
 		}
 		return false;
 	},
 
+	_listens: function (type, fn, context) {
+		var listeners = this._events && this._events[type] || [];
```
mourner, line 250: I'm wondering... What if we handle the `(type)` case here by adding `if (!fn) return true`? Then we would be able to remove a block of code above (229–238) and replace it with just `if (this._listens(type, fn, context)) { return true; }`.

mourner (approved):

src/core/Events.js
Falke-Design: Good point, changed

src/core/Events.js
```diff
@@ -152,30 +145,24 @@ export var Events = {
...
+		if (listener) {
+			if (this._firingCount) {
+				// set the removed listener to noop so that's not called if remove happens in fire
+				listener.fn = Util.falseFn;
 
-				return;
+				/* copy array in case events are being fired */
+				this._events[type] = listeners = listeners.slice();
 			}
+			var idx = listeners.indexOf(listener);
```
mourner, line 163: I just realized that this is a perf regression for `off`, since we now essentially have to loop over the listeners twice — first to find the listener in `_listen`, and then to fined the index for removal. 

One way to fix this would be to return the index rather than the listener in `_listen`, but then we have to be careful in `if` conditions elsewhere, changing from `if (this._listens(...))` to `if (this._listens(...) !== false)`.

mourner (changes requested):

src/core/Events.js
Falke-Design: I fixed that with returning a object:
```
return {
	index: i,
	listener: listeners[i]
};
```

src/core/Events.js
```diff
@@ -216,22 +203,61 @@ export var Events = {
...
+			// Less memory footprint.
+			context = undefined;
+		}
+
+		for (var i = 0, len = listeners.length; i < len; i++) {
+			if (listeners[i].fn === fn && listeners[i].ctx === context) {
+				return {
+					index: i,
+					listener: listeners[i]
+				};
```
mourner, line 256: The problem with this approach is that while it's certainly more readable, it has an overhead of an extra object allocation for the return result. It might not be significant, but it's still overhead and if we can avoid it, I think we should.

mourner (changes requested):

src/core/Events.js
```diff
@@ -233,14 +233,15 @@ export var Events = {
...
+	// returns the index (number) or false
 	_listens: function (type, fn, context) {
 		if (!this._events) {
 			return false;
 		}
 
 		var listeners = this._events[type] || [];
 		if (!fn) {
-			return !!listeners.length;
+			return listeners.length > 0 ? 0 : false;
```
mourner, line 244: We can return `true` here like before, it still fits the `!== false` condition where necessary.

mourner (approved):

PR #8163, opened 2022-04-19
Add FeatureCollection to geometryToLayer

### LINKED ISSUE #8141: Missing FeatureCollection case in the leaflet/GeoJSON.geometryToLayer switch.
Missing FeatureCollection case in the [leaflet/GeoJSON.geometryToLayer](https://github.com/Leaflet/Leaflet/blob/main/src/layer/GeoJSON.js
) switch.

Current version throws error, when i provided "FeatureCollection" type.

`

  	case 'GeometryCollection':
  		for (i = 0, len = geometry.geometries.length; i < len; i++) {
  			var layer = geometryToLayer({
  				geometry: geometry.geometries[i],
  				type: 'Feature',
  				properties: geojson.properties
  			}, options);

  			if (layer) {
  				layers.push(layer);
  			}
  		}
  		return new FeatureGroup(layers);

  	default:
  		throw new Error('Invalid GeoJSON object.');
  	}
`

Fixed version: 

`

  	case 'GeometryCollection':
  		for (i = 0, len = geometry.geometries.length; i < len; i++) {
  			var layer = geometryToLayer({
  				geometry: geometry.geometries[i],
  				type: 'Feature',
  				properties: geojson.properties
  			}, options);

  			if (layer) {
  				layers.push(layer);
  			}
  		}
  		return new FeatureGroup(layers);

        case 'FeatureCollection':
              for (i = 0, len = geometry.features.length; i < len; i++) {
                  var flayer = this.geometryToLayer(geometry.features[i], options);
        
                  if (flayer) {
                      layers.push(layer);
                  }
              }
              return new FeatureGroup(layers);

  	default:
  		throw new Error('Invalid GeoJSON object.');
  	}
`

Falke-Design:
Fixes #8141

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -571,6 +571,19 @@ describe("L.GeoJSON functions", function () {
 			]
 		};
 
+		var featureCollection  = {
+			type: "FeatureCollection",
+			features: [
+				{
+					type: "Feature",
+					geometry: {
+						type: "Point",
+						coordinates: [0, 0]
+					}
+				},
+			]
+		};
+
 		function customPointToLayer(geojsonPoint, latLng) {
 			return L.circle(latLng, {
 				radius: geojsonPoint.properties.radius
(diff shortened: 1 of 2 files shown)


mourner (approved):
BTW, it seems weird in retrospect that the method is called `geometryToLayer` — it used to only handle geometry but then expanded into features for some reason. Maybe worth renaming to something like `geojsonToLayer`.

PR #8167, opened 2022-04-20
Fix `map.getCenter()` returning a mutable object

### LINKED ISSUE #8164: map.flyTo shifts canvas
Upon upgrading to version 1.8.0, I've noticed new intermittent behavior with the map.flyTo function, where all items on the canvas shift, and the tiles remain stationary. This is fixed after a pan of the map or re-zooming. 

**Steps to reproduce**
- Set a lat/lng and zoom
- run `map.flyTo()` with a lat/lng that is near the map center (within view) and the same zoom
- notice markers on canvas have shifted but tiles underneath have not changed

**Expected behavior**
I expect the map to pan to the new centerpoint and update the tiles as well

**Current behavior**
Items on canvas are moving, and the tiles/center point of map is staying the same

**Environment**
- Leaflet version: 1.8.0
- Browser (with version): Chrome Version 100.0.4896.88
- OS/Platform (with version): macOS Catalina 10.15.5

In development -- hoping if others encounter they can add a functioning example, it's still intermittent albeit reproducible for me.

mourner:
Closes #8164 by cloning the `LatLng` object returned by `map.getCenter()` so that mutating it afterwards doesn't lead to unexpected bugs. Also fixes a `panInside` test that didn't work properly before. cc @Sillson

### DIFF
--- a/spec/suites/map/handler/Map.DragSpec.js
+++ b/spec/suites/map/handler/Map.DragSpec.js
@@ -132,7 +132,7 @@ describe("Map.Drag", function () {
 			});
 
 			var originalCenter = L.latLng(0, 0);
-			map.setView(originalCenter, 1);
+			map.setView(originalCenter.clone(), 1);
 
 			var spy = sinon.spy();
 			map.on('drag', spy);
@@ -142,7 +142,7 @@ describe("Map.Drag", function () {
 				onStop: function () {
 					expect(map.getZoom()).to.be(1);
 					// Expect center point to be the same as before the click
-					expect(map.getCenter()).to.be(originalCenter);
+					expect(map.getCenter()).to.eql(originalCenter);
 					expect(spy.callCount).to.eql(0); // No drag event should have been fired.
 
 					done();
@@ -273,7 +273,7 @@ describe("Map.Drag", function () {
 				inertia: false
 			});
 			var originalCenter = L.latLng(0, 0);
-			map.setView(originalCenter, 1);
+			map.setView(originalCenter.clone(), 1);
 
 			map.on('mousedown', function () {
 				map.dragging.disable();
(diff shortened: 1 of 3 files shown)


spec/suites/map/handler/Map.DragSpec.js
```diff
@@ -132,7 +132,7 @@ describe("Map.Drag", function () {
 			});
 
 			var originalCenter = L.latLng(0, 0);
-			map.setView(originalCenter, 1);
+			map.setView(originalCenter.clone(), 1);
```
Falke-Design, line 135: Maybe `clone()` should be not necessary here. It should be cloned in the backend logic of `setView`.

spec/suites/map/handler/Map.DragSpec.js
```diff
@@ -142,7 +142,7 @@ describe("Map.Drag", function () {
 				onStop: function () {
 					expect(map.getZoom()).to.be(1);
 					// Expect center point to be the same as before the click
-					expect(map.getCenter()).to.be(originalCenter);
+					expect(map.getCenter()).to.eql(originalCenter);
```
Falke-Design, line 145: I would replace `map.setView(originalCenter.clone(), 1);` with `map.setView(originalCenter, 1);` and then add following test:
```
expect(map.getCenter()).to.eql(originalCenter);
expect(map.getCenter()).to.not.be(originalCenter);
```

Meaning: Same data but not the same object, because we break the reference with `clone()` in the backend


Maybe worth to add a own test instead of extending this one.

Falke-Design (commented):
Currently I can only review it not test it. But looks good for me

spec/suites/map/handler/Map.DragSpec.js
mourner: It's necessary because it makes sure the tests below compare with the original value regardless of Leaflet's internal implementation it is testing against.

spec/suites/map/handler/Map.DragSpec.js
mourner: The fact it's not the same object is tested elsewhere, in the added test (where `map.getCenter()` value is mutated).

spec/suites/map/handler/Map.DragSpec.js
Falke-Design: Ah yes I understand

Falke-Design (approved):

PR #8168, opened 2022-04-20
Fix "listener not found" warning when setting `maxBounds`

### LINKED ISSUE #7951: "Listener not found" warning when passing maxBounds option to map

mourner:
Fixes #7951. A simpler alternative to #8161.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -449,11 +449,11 @@ export var Map = Evented.extend({
 		if (this._maxBoundsSet) {
 			this._maxBoundsSet = false;
 			this.off('moveend', this._panInsideMaxBounds);
+		}
 
-			if (!bounds.isValid()) {
-				this.options.maxBounds = null;
-				return this;
-			}
+		if (!bounds.isValid()) {
+			this.options.maxBounds = null;
+			return this;
 		}
 
 		this.options.maxBounds = bounds;
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -446,11 +446,14 @@ export var Map = Evented.extend({
 	setMaxBounds: function (bounds) {
 		bounds = toLatLngBounds(bounds);
 
-		if (!bounds.isValid()) {
-			this.options.maxBounds = null;
-			return this.off('moveend', this._panInsideMaxBounds);
-		} else if (this.options.maxBounds) {
+		if (this._maxBoundsSet) {
+			this._maxBoundsSet = false;
 			this.off('moveend', this._panInsideMaxBounds);
+
+			if (!bounds.isValid()) {
+				this.options.maxBounds = null;
+				return this;
+			}
 		}
 
 		this.options.maxBounds = bounds;
@@ -459,6 +462,7 @@ export var Map = Evented.extend({
 			this._panInsideMaxBounds();
 		}
 
+		this._maxBoundsSet = true;
 		return this.on('moveend', this._panInsideMaxBounds);
 	},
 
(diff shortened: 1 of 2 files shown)


src/map/Map.js
```diff
@@ -446,11 +446,14 @@ export var Map = Evented.extend({
...
-			return this.off('moveend', this._panInsideMaxBounds);
-		} else if (this.options.maxBounds) {
+		if (this._maxBoundsSet) {
+			this._maxBoundsSet = false;
 			this.off('moveend', this._panInsideMaxBounds);
+
+			if (!bounds.isValid()) {
+				this.options.maxBounds = null;
+				return this;
+			}
```
Falke-Design, line 456: If the variable `bounds` is passed invalid, you don't return (like it was before) and apply it to `maxBounds` in line 459.

Falke-Design (changes requested):
Is a valid approche but it does **not** close #8161 it makes it only for this usecase obsolete.
The ability to check for an event handler with a certain function is missing generally and is provided with #8161.

src/map/Map.js
mourner: @Falke-Design what does the `return this` line do then?

mourner:
> Is a valid approche but it does **not** close #8161 it makes it only for this usecase obsolete.

Yeah, but I don't think such a feature is necessary given that it complicates the API, unless more use cases arise (and they haven't in 11 years as far as I can tell). Leaflet's philosophy is minimalism and doing away with less.

src/map/Map.js
Falke-Design: If the bounds are invalid but `_maxBoundsSet` is not set, you don't come to the part `if (!bounds.isValid()) {`

mourner: Oh, I see now, this is the case where max bounds wasn't set before.

Falke-Design:
I had this usecase many times ... 

> complicates the API

I don't really see that it makes it more complex, also I think that if we provide adding (`on`) / removing (`off`), we should offer a getting (`listens`).
But let us move this disscussion to #8161

Falke-Design (approved):

mourner:
@Falke-Design you're right — now that I think about it more, it makes sense to make this consistent with other methods, and you're more qualified than me with regards to possible use cases because I wasn't very active in Leaflet development for so long. Let's land that one too then.

dmytro-shchurov:
Why do I see this warning in `"version": "1.8.0",` updated today? I mean different warning about missed listeners (moveend including)

IvanSanchez:
> Why do I see this warning in `"version": "1.8.0",` updated today? I mean many different warnings about missed listeners (moveend including)

Because 1.8.0 was released on the 18th of April, and #8168 was merged into `main` on the 20th of April.

I mean, man, look at the dates.

Issue #7632, opened 2021-08-04
Validation Error: img element with empty alt attribute must not have role attribute


robinson2:
The HTML code of the tiles produced by Leaflet …

`<img crossorigin="" alt="" role="presentation" src="https://api.maptiler.com/maps/voyager/5/16/10.png?key=…" class="leaflet-tile leaflet-tile-loaded" style="width: 512px; height: 512px; transform: translate3d(886px, -296px, 0px); opacity: 1;">`

… has an error when checked via https://validator.w3.org/nu/ :

"An img element which has an alt attribute whose value is the empty string must not have a role attribute."

In the HTML source code I saw that both were implemented for accessibility reasons:

```
/*
Alt tag is set to empty string to keep screen readers from reading URL and for compliance reasons
http://www.w3.org/TR/WCAG20-TECHS/H67
*/
tile.alt = '';

/*
Set role="presentation" to force screen readers to ignore this
https://www.w3.org/TR/wai-aria/roles#textalternativecomputation
*/
tile.setAttribute('role', 'presentation');
```

On the page https://www.w3.org/WAI/tutorials/images/decorative/ both possibilities are presented and regarding the role attribute the W3C writes: "currently, this feature is not as widely supported as using a null alt attribute."

I would therefore suggest that the role attribute is removed at this point, or that you can choose between both options (alt or role) via configuration.

**Environment**
- Leaflet version: latest 1.7.1 master

Falke-Design:
@Malvoz you are an expert for this things, what do you think?

Malvoz:
Using only one of `alt` or `role` on images results in <q>Image</q> being announced by screen readers and a new error message by the validator, respectively.

If you really wanted to suppress the error message(s) right now you could use `<img ... aria-hidden="true" alt>` without `role`. Although `aria-hidden="true"` is widely supported there may be [compatibility issues](https://www.a11yproject.com/posts/2020-05-13-aria-has-perfect-support/) so `alt` would act as a fallback, in which case <q>Image</q> would be announced by screen readers due to no `role="presentation"`.

I would wait for resolution in https://github.com/validator/validator/issues/1000 before making any changes.

Malvoz:
@timveld 

> on what basis do you claim that setting alt="" on an image results in the screen reader speaking "image"?

[...]

PR #8172, opened 2022-04-21
Remove `role="presentation"` from image tiles (HTML validation error fix)

Malvoz:
Fixes https://github.com/Leaflet/Leaflet/issues/7632.

[`img` with an empty `alt` is implicitly `role="presentation"`](https://www.w3.org/TR/html-aria/#el-img-empty-alt) and it's [not clear that `role="presentation"` was ever useful](https://github.com/Leaflet/Leaflet/issues/7632#issuecomment-1088659800).

---

Side note:

I can no longer reproduce [the TalkBack issue that would potentially justify setting `aria-hidden="true"`](https://github.com/Leaflet/Leaflet/issues/7632#issuecomment-1066980024), nor can I [Get image descriptions](https://support.google.com/chrome/answer/9311597?co=GENIE.Platform%3DAndroid&oco=1) to test if TalkBack would even respect `aria-hidden="true"`. Perhaps the feature was removed from Android, or it's very well hidden and now disabled by default. I'll revisit the TalkBack issue in the future, to see if `aria-hidden="true"` is desired.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -154,17 +154,11 @@ export var TileLayer = GridLayer.extend({
 			tile.referrerPolicy = this.options.referrerPolicy;
 		}
 
-		/*
-		 Alt tag is set to empty string to keep screen readers from reading URL and for compliance reasons
-		 https://www.w3.org/TR/WCAG20-TECHS/H67
-		*/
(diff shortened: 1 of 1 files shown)


Falke-Design (approved):

Falke-Design:
@timveld what do you think?

PR #7818, opened 2021-11-26
Replace Mapbox with OpenStreetMap

Falke-Design:
Because of https://github.com/Leaflet/Leaflet/issues/6336#issuecomment-954095682 I suggest to use OpenStreetMap tiles.

> You need permission for heavy use - is this heavy use? Who knows because that term is completely undefined which is why everybody that asks OWG just gets a shrug in response. That part of the policy is just stupid but nobody knows what else to say.

_Originally posted by @tomhughes in https://github.com/Leaflet/Leaflet/issues/6336#issuecomment-954095682_

Fix: #6336

PR #8177, opened 2022-04-22
Add Ukraini dialog

Falke-Design:
Replaces the existing Urkaini war site with an dialog. 

With this change existing PRs don't get merge conflicts like #7818 and existing links, redirects are working again.
The dialog opens only once every 24 hours.

@mourner @IvanSanchez @Malvoz do you see any problems with the usage of `sessionStorage` and `iframe`?
The changed files for the dialog are:
- /SlavaUkraini/*
- index.html
- /_layouts/v2.html
- /docs/css/main.css

Here a [demo](https://florianbischof.net/leaflet/ukraini-dialog/). In the demo I changed that the dialog opens every 10 seconds again (on a site change).


docs/SlavaUkraini/dialog.js
```diff
@@ -0,0 +1,34 @@
...
+
+	if (sessionStorage) {
+		var session = sessionStorage.getItem(SESSION_KEY);
+		// open the dialog only every 24 hours
+		if (session && Date.now() - session < ONE_DAY_MILLI_SEC) {
+			return;
+		}
+	}
+
+	var html = '<div class=\'container iframe-container\'><span class=\'close-dialog\' aria-label=\'close\'>&times;</span><iframe src=\'/SlavaUkraini/\' /></div>';
```
Malvoz, line 14: Please use `<button>` or `<a>` with `role="button"` instead of `<span>`.

docs/docs/css/main.css
```diff
@@ -1118,3 +1118,50 @@ iframe[src*='youtube.com'] {
...
+	text-align: center;
+	top: 10px;
+	right: 20px;
+	position: absolute;
+	font-size: 40px;
+	line-height: 40px;
+	cursor: pointer;
+}
+
+.close-dialog:hover{
```
Malvoz, line 1163: Always use `:focus` in conjunction to `:hover` please. 🙂

docs/SlavaUkraini/dialog.js
Malvoz, line 14: For keyboard-friendly dialog:

1. [`autofocus`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus) the dialog
2. Change the DOM order of iframe and the close button

Malvoz:
[The change to logo-ua.png](https://github.com/Leaflet/Leaflet/pull/8177/files#diff-99b9f5abae0c198c703f28f6c1ff621a024826f2580433b1691ce641791dc6e8) is reverting @mourner's https://github.com/Leaflet/Leaflet/commit/3f22bb87e64ea117c829a05909402321aef30532.

mourner (changes requested):
This looks great! I was thinking about moving it back to help with the conflicts, but the dialog idea is brilliant. 👍 from me apart from addressing things @Malvoz brought up.

docs/SlavaUkraini/dialog.js
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+
+	var html = '<div class=\'container iframe-container\'><iframe tabindex="1" src=\'/SlavaUkraini/\'></iframe><a tabindex="2" role="button" class=\'close-dialog\' aria-label=\'close\'>&times;</a></div>';
+
+	var dialog = document.createElement('div');
+	dialog.id = 'ukraini-dialog';
+	dialog.innerHTML = html;
+
+	document.body.appendChild(dialog);
```
Malvoz, line 20: Also, the logical order of the dialog is to be the first child of body since it's the first content presented to the user, although it matters less if it's autofocused.

docs/SlavaUkraini/dialog.js
Falke-Design: `autofocus` does only work if the element is already existing while page loading. I used now a JS solution `focus()`.

Is there a way to make a `button` / `a` element clickable over enter / space without adding keypress listeners?

Malvoz: > autofocus does only work if the element is already existing while page loading

Ok, otherwise could not be added/removed from DOM but instead have its visibility toggled?

> Is there a way to make a `button` / `a` element clickable over enter / space without adding keypress listeners?

`<button type="submit">` inside a `<form method="dialog">` is one way. But [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) requires a [polyfill](https://github.com/GoogleChrome/dialog-polyfill) due to current [support](https://caniuse.com/dialog).

Malvoz (approved):
Seems to work well, nice!

mourner (approved):

PR #8184, opened 2022-04-24
Added aria-label when Layers collapsed = false

### LINKED ISSUE #7529: Contextual information missing for `leaflet-control-layers-list` when control is collapsed
The `<a class="leaflet-control-layers-toggle">`'s `title="Layers"` provides a summary (context) for the content it controls, however when the control's `collapsed` option is set to `false` the control is `display: none` and no summary is provided for the content.

In this case the `leaflet-control-layers-list` element could/should have `aria-label="Layers"` (this is analogous to a `<legend>` providing the accessible name in a `<fieldset>`).

Relates to <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> <span title="Success Criteria">SC</span> [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).

<hr>

IMO, this issue is of **low priority** compared to existing [accessibility issues](https://github.com/Leaflet/Leaflet/issues?q=is%3Aissue+is%3Aopen+label%3Aaccessibility) to date.

whtann:
Hello! I believe this simple change should fix #7529 .
I am relatively new to using aria-labels, so if I am wrong please let me know if I am wrong.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -208,6 +208,8 @@ export var Layers = Control.extend({
 
 		if (!collapsed) {
 			this.expand();
+
+			link.setAttribute('aria-label', Layers);
 		}
 
 		this._baseLayersList = DomUtil.create('div', className + '-base', section);


src/control/Control.Layers.js
```diff
@@ -208,6 +208,8 @@ export var Layers = Control.extend({
 
 		if (!collapsed) {
 			this.expand();
+
+			link.setAttribute('aria-label', Layers);
```
Falke-Design, line 212: Should be `Layers` not `'Layers'`?

Falke-Design (changes requested):

src/control/Control.Layers.js
Malvoz, line 212: @whtann Note that this sets the `aria-label` on the `<a>` control. That's not what is asked for in https://github.com/Leaflet/Leaflet/issues/7529:

> When the layer control's `collapsed` option is set to `false` _the [`leaflet-control-layers-list`](https://github.com/Leaflet/Leaflet/blob/3466cbba34351ee0d7477a83eb0215cc0e6b454b/src/control/Control.Layers.js#L184) element should have `aria-label="Layers"`_.

(again, this is because in that scenario the `<a>` is not available to provide the necessary context)

Malvoz (changes requested):

PR #6123, opened 2018-04-06
Add pre-commit hook

denysdovhan:
This PR introduces better contributing experience. From `CONTRIBUTING.md`:

> Before committing your changes, run `npm run lint` to catch any JS errors in the code and fix them.

This PR adds `pre-commit` hook for linting `**/*.js` files and `pre-push` hook for running tests.

PR #8189, opened 2022-04-25
Add pre-commit hook to fix linting issues

jonkoops:
Sets up [`lint-staged`](https://github.com/okonet/lint-staged) to lint JavaScript and Markdown files. This will ensure the files are linted and fixed before a commit is made.

This PR is intended as a replacement of #6123. Since the packages involved have changed so much that the original PR no longer matches the expectations.


package.json
```diff
@@ -44,13 +46,13 @@
     "test": "karma start ./spec/karma.conf.js",
     "build": "npm run rollup && npm run uglify",
     "lint": "eslint .",
-    "lintfix": "npm run lint -- --fix",
```
jonkoops, line 47: I've removed this script to fix linting issues as it is expected the pre-commit hook will take care of this for us.

Falke-Design: Sometimes lintfix can not resolve everything and then we need a log what doesn't worked

jonkoops: Should be fine as a commit cannot be made when there are unfixable linting errors. For example:

An example of a lint failing on a commit

Falke-Design: Nice 😄 
But I would like to keep it. If I don't want to commit but to fix the issues, I can't do it anymore. I think it doesn't hurt if we keep it.

Falke-Design:
Is there a way to commit into an branch without fixing lint issues for projects "work in progress"? Probably not

jonkoops:
> Is there a way to commit into an branch without fixing lint issues for projects "work in progress"? Probably not

You can always skip any pre-commit checks by passing `--no-verify`. For example: `git commit -m "WIP" --no-verify`.

package.json
Falke-Design: >So then this behavior is fine right? The commit is not actually made unless all issues are resolved, either manually or by auto-fixing.

Yes is fine for the commit part but I meant that we should keep `lintfix`.

jonkoops: You can always run `npm run lint -- --fix`. But I don't see a scenario where you need this unless you explicitly opt out of pre-commit checks.

Falke-Design: > You can always run npm run lint -- --fix.

Sure but we can also simply keep `lintfix`. What is the benefit of removing it? Maybe you are right and it's not needed anymore, but I would like to test it first in practice and check if it is obsolete.

> But I don't see a scenario where you need this

For example if I want to commit and then it failed because of lint issues. Then I fix them but want to make sure that everything is fine without commiting directly, then I need `lintfix`

mourner:
This looks great, the only reservation I have is whether we should `--fix` in the hook. Some people might find their code being automatically fixed more convenient, but maybe it's better to make them fix the errors explicitly so that they're more aware of the issues and write correct code more intentionally. Some eslint auto fixes (especially non-formatting ones) can be non-trivial.

jonkoops:
I've used this on several projects with the auto-fixing enabled and I've been quite happy with it as it fixes a lot of trivial issues especially when combined with a code formatter. I personally have auto fixing enabled in my IDE so my opinion and experiences are biased.

It can go either way, I don't have a preference.

Falke-Design:
I think we should run `--fix` in the hook. Maybe we can call first without `--fix` so that all lint errors are in the log.

PR #8190, opened 2022-04-25
Fix fire same event inside of `once()`

Falke-Design:
If the same event is called in `once()` it ends up in a endless loop.

Now `once()` has the same logic as `on()` so it would be possible to be merged them together but then we need a new function or a 4 boolean argument in `on()` if it is `once`.

```js
var obj = new L.Evented();

obj.once('test', function(){
	console.log('Called in once function');
	obj.fire('test');
}, obj);

obj.fire('test');
```

[1.8.0 Demo](https://plnkr.co/edit/JQT7ULLm8JfSPnAi)

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -192,7 +192,7 @@ export var Events = {
 					// off overwrites l.fn, so we need to copy fn to a var
 					var fn = l.fn;
 					if (l.once) {
-						this.off(type, l.fn, l.ctx);
+						this.off(type, fn, l.ctx);
 					}
 					fn.call(l.ctx || this, event);
 				}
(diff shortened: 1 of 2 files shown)


jonkoops:
@Falke-Design there are some linting issues, can you give them a look?

src/core/Events.js
```diff
@@ -182,12 +185,16 @@ export var Events = {
...
-
 			if (listeners) {
 				this._firingCount = (this._firingCount + 1) || 1;
 				for (var i = 0, len = listeners.length; i < len; i++) {
 					var l = listeners[i];
-					l.fn.call(l.ctx || this, event);
+					// off overwrites l.fn, so we need to copy fn to a var
+					var fn = l.fn;
+					if (l.once) {
+						this.off(type, l.fn, l.ctx);
```
mourner, line 195: Nit: 
```suggestion
						this.off(type, fn, l.ctx);
```

mourner (approved):

PR #8193, opened 2022-04-27
Refactor dialog

Malvoz:
Attempt on a small refactoring of the dialog:

- [x] https://github.com/Leaflet/Leaflet/commit/d30c7dfc59f6532abf9d351a44d9d2efa40ddcd3 corrects the `logo-ua.png` version in `docs/docs/images/`
- [x] https://github.com/Leaflet/Leaflet/commit/4667dd97478726b40831ac82c9575b635e975319 dialog changes:
   - refactors dialog to be a more generic dialog
   - excludes the google analytics script from the iframe
   - removes the duplicate files in [`docs/SlavaUkraini/docs`](https://github.com/Leaflet/Leaflet/tree/main/docs/SlavaUkraini/docs) to re-use files in [`docs/docs`](https://github.com/Leaflet/Leaflet/tree/main/docs/docs) instead

"Off-topic" but related to the back and forth URL change:

- [x] https://github.com/Leaflet/Leaflet/commit/5085c68cd86b9515526596daa1fdd8ab21aabcf5 changes out of date `/SlavaUkraini/` link (in changelog)
- [x] https://github.com/Leaflet/Leaflet/commit/9b9c6f53320a3a91fd97cb19f9e3bd0420ab0c03 fixes #8191

PR #8194, opened 2022-04-27
Use the topmost browsing context for links in the dialog iframe

Malvoz:
Per https://github.com/Leaflet/Leaflet/pull/8193#pullrequestreview-954886800.

(`target="_top"` escapes all levels of iframes, not only parents.)

### DIFF
--- a/docs/_layouts/tutorial_frame.html
+++ b/docs/_layouts/tutorial_frame.html
@@ -1,7 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-	<base target="_top">
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


docs/_layouts/tutorial_frame.html
```diff
@@ -1,11 +1,11 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+	<base target="_top">
+	<meta charset="utf-8">
+	<meta name="viewport" content="width=device-width, initial-scale=1">
```
Malvoz, line 6: off-topic: `<meta charset>` should be present in the first 1024 bytes of the document: https://web.dev/charset/

Malvoz:
For `tutorial_frame.html` this'll affect clicks on the popup's close button, due to regression described in https://github.com/Leaflet/Leaflet/issues/8159. Maybe we should just change this for the dialog for now?

mourner (approved):

mourner:
Yeah, let's just do the dialog frame for now.

PR #8196, opened 2022-04-27
Drop legacy VML code & official support for IE7–8

mourner:
Looks like we haven't really checked whether Leaflet is still working in IE8 for at least 4 years. The latest mention of us fixing anything related to it goes back to #5731 — 5 years ago. We don't test it manually, and we don't run any tests on CI as well. So we might as well just drop the pretence we're still supporting old IE and remove the legacy VML code at least. This probably won't signify a major breaking change (so it can go to v1.9.0). Related: #6136 

According to the latest version stats from statcounter, IE7 is not even in the stats, and IE8 has 0.03% market share.


Falke-Design:
It is a breaking change, so I suggest that we make a v2 with dropping IE support and move the other changes to v3. It would be much cleaner and correcter.

IvanSanchez (approved):
Hell yeah.

mourner:
> It is a breaking change, so I suggest that we make a v2 with dropping IE support and move the other changes to v3. It would be much cleaner and correcter.

@IvanSanchez @jonkoops @Malvoz what do you think? Should we drop IE7–8 for v2 and then target no IE for v3, or drop IE7–8 for v1.9?

IvanSanchez:
> Should we drop IE7–8 for v2 and then target no IE for v3, or drop IE7–8 for v1.9?

I believe in semantic versioning (https://semver.org/), so my vote goes to "v2 drops IE8, v3 drops IE10/11"

jonkoops:
I also believe we should save this one for the v2 release.

Malvoz:
I don't have a strong opinion on this, though I believe it'd be much clearer to those who want to support IE if support was dropped for all versions in a single release.

Also, I was under the impression that the aim was to drop support for IE in v2 specifically. People seem to be on board with that idea: https://github.com/Leaflet/Leaflet/issues/6136#issuecomment-1108628953.

Falke-Design:
Let us discuss this in #6136

docs/index.html
```diff
@@ -152,7 +152,7 @@ <h4>Desktop</h4>
             <li>Firefox</li>
             <li>Safari 5+</li>
             <li>Opera 12+</li>
-            <li>IE 7&ndash;11</li>
+            <li>IE 9&ndash;11</li>
```
Falke-Design, line 155: No support too

Falke-Design (changes requested):

docs/index.html
mourner: This specific PR didn't drop support for it, I'm just keeping things narrowly scoped. I think we'll overhaul the browser support section anyway when pushing on v2.

Issue #7191, opened 2020-07-01
Layers control should not expand on focus


Malvoz:
- [ ] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

> **Note** <ins>Issue updated per https://github.com/Leaflet/Leaflet/pull/7057/files#r406663620.</ins>

**Steps to reproduce**

1. Visit https://leafletjs.com/examples/layers-control/example.html;
2. <kbd>Tab</kbd> until the layer control is focused;
3. Observe how <s>in Chrome</s> the layers list is expanded<s>, in Firefox it is not expanded</s>.

**Expected behavior**

The layers list should only be expanded after the layer control has been pressed.

I believe this is the original intent, as described in https://github.com/Leaflet/Leaflet/issues/6946#issuecomment-567620007 (emphasis mine): "the default collapsible `L.Control.Layers` can be navigated via keyboard: _**tab-focus on the collapsed layers control, then enter**_, then tab-focus on the set of radio boxes or on the checkboxes."

**Current behavior**

Layer list is expanded <s>in Chrome</s> <ins>(in all browsers)</ins> when the layer control recieves focus.

**Environment**
- Leaflet version: <s>1.6.0, 1.7.1</s> <ins>1.9.2</ins>
- Browser (with version): Firefox 77.0.1 / Chrome 83.0.4103.116
- OS/Platform (with version): Windows 10

**Additional context**

Having the layers list only expand after the layer control has been pressed would also meet WCAG success criterion 2.4.1 Bypass Blocks, the "[Understanding 2.4.1: Bypass Blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks.html)" documentation lists several benefits of hiding content (such as a layer list) behind a menu, including:

> - People who use only the keyboard or a keyboard interface can reach content with fewer keystrokes. Otherwise, they might have to make dozens of keystrokes before reaching a link in the main content area. This can take a long time and may cause severe physical pain for some users.

IvanSanchez:
I suspect that Chromium fires some `pointerover` DOM events (or similar) when tabbing, but would like confirmation of this.

A quick&dirty solution for this would be to explicitly expand the layers control on a `focus` event.

[...]

PR #8202, opened 2022-04-30
Added `aria-expanded` state with toggle feature for `leaflet-control-layers-toggle`, regarding Issue #7526

### LINKED ISSUE #7526: `leaflet-control-layers-toggle` should convey the `aria-expanded` state of `leaflet-control-layers`
## Motivation

(Same as #7527, but for the layers control.)

The [`aria-expanded`](https://www.w3.org/TR/wai-aria-1.2/) attribute:

> Indicates whether a grouping element owned or controlled by this element is expanded or collapsed.
>
> The `aria-expanded` attribute is applied to a focusable, interactive element that toggles visibility of content in another element.
>
> For example [...] a button that controls visibility of a section of page content.

## Proposed solution

The [layers control](https://github.com/Leaflet/Leaflet/blob/76b0d406d78ebdcc083d12d2296087a79cdec9bf/src/control/Control.Layers.js#L201) (`leaflet-control-layers-toggle`) should convey the [`aria-expanded`](https://www.w3.org/TR/wai-aria-1.2/#aria-expanded) [state](https://www.w3.org/TR/wai-aria-1.2/#dfn-state):

- `false` when the layers menu is collapsed (`leaflet-control-layers`)
- `true` when the layers menu is expanded (`leaflet-control-layers-expanded`)

## Additional context

- MDN web docs: [`aria-expanded`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded)
- Relates to <span title="Web Content Accessibility Guidelines">WCAG 2.1</span> <span title="Success Criteria">SC</span> [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships).

ShehanAT:
# Description

This PR adds the `aria-expanded` state to the `leaflet-control-layers-toggle`  anchor element(`<a/`). This `aria-expanded` state is toggled to `true` on mouse enter event and to `false` on the mouse leave event. 

Here is a GIF demonstrating this change:
GIF of the `aria-expanded` attribute toggling on mouse hover

Fixes #7526

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Accessibility feature 
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Prior to committing the `npm run lint` command was run successfully with no issues

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -177,7 +177,6 @@ export var Layers = Control.extend({
 
 		// makes this work on IE touch devices by stopping it from firing a mouseout event when the touch is released
 		container.setAttribute('aria-haspopup', true);
-
 		DomEvent.disableClickPropagation(container);
 		DomEvent.disableScrollPropagation(container);
 
@@ -188,6 +187,7 @@ export var Layers = Control.extend({
 		link.href = '#';
 		link.title = 'Layers';
 		link.setAttribute('role', 'button');
+
 		DomEvent.on(link, 'click', DomEvent.preventDefault); // prevent link function
 		DomEvent.on(link, 'focus',
 			() => {
@@ -199,8 +199,14 @@ export var Layers = Control.extend({
 				keyboardFocus = false;
 			}
 		);
-		DomEvent.on(link, 'keydown', () => {
+		DomEvent.on(container, 'click', () => {
 			if (keyboardFocus) {
+				this.expand();
+				link.setAttribute('aria-expanded', 'true');
+			}
+		});
+		DomEvent.on(link, 'keydown', (e) => {
+			if (keyboardFocus && e.key === 'Enter') {
 				this.expand();
 				link.setAttribute('aria-expanded', 'true');
 			}
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -184,15 +184,34 @@ export var Layers = Control.extend({
 		var section = this._section = DomUtil.create('section', className + '-list');
 
 		var link = this._layersLink = DomUtil.create('a', className + '-toggle', container);
+		var keyboardFocus = false;
 		link.href = '#';
 		link.title = 'Layers';
 		link.setAttribute('role', 'button');
 		DomEvent.on(link, 'click', DomEvent.preventDefault); // prevent link function
-		DomEvent.on(link, 'focus', this.expand, this);
+		DomEvent.on(link, 'focus',
+			() => {
+				keyboardFocus = true;
+			}
+			, this);
+		DomEvent.on(link, 'focusout',
+			() => {
+				keyboardFocus = false;
+			}
+		);
+		DomEvent.on(link, 'keydown', () => {
+			if (keyboardFocus) {
+				this.expand();
+				link.setAttribute('aria-expanded', 'true');
+			}
+		});
 		link.setAttribute('aria-expanded', 'false');
 
 		if (collapsed) {
-			this._map.on('click', this.collapse, this);
+			this._map.on('click', () => {
+				this.collapse();
+				link.setAttribute('aria-expanded', 'false');
+			}, this);
 
 			DomEvent.on(container, {
 				mouseenter: function () {
@@ -207,14 +226,6 @@ export var Layers = Control.extend({
 					this.collapse();
 					link.setAttribute('aria-expanded', 'false');
 				},
-				focusin: function () {
-					link.setAttribute('aria-expanded', 'true');
-				},
-				focusout: function () {
-					this.collapse();
-					link.setAttribute('aria-expanded', 'false');
-				}
-
 			}, this);
 		}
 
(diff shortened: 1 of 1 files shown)


Malvoz (changes requested):
Toggling of `aria-expanded` should not be tied to certain input mechanisms, it should work with keyboard as well for example.

ShehanAT:
Ok I'll try to implement the changes to keyboard inputs as well

ShehanAT:
Added keyboard input support for the `aria-expanded` attribute for `leaflet-control-layers-toggle` element.

Here is a GIF demonstrating the changes:
Leaflet_AriaExpanded_Fix_2

Malvoz:
Sorry @ShehanAT, I forgot about the layers control's behavior. While the changes are technically correct, what should happen is:

1. https://github.com/Leaflet/Leaflet/issues/7191 - rather it should be collapsed/expanded when the layers control is activated
2. `aria-expanded` should be toggled based on the collapsed/expanded state

ShehanAT:
Alright I'll look into this soon

ShehanAT:
Added the request changes so that:
1. The layers control is not expanded upon `focus` event. Rather, only pressing `enter` or `tab` again after the `focus` event will cause it to expand 
2. Upon layers control expanding via `focus` then `enter` key, the `aria-expanded` attribute is set to true. When a mouse click event on the map is registered, the layers control collapses and the `aria-expaned` attribute is set to false. 

Here is a GIF demonstrating the new changes:
Leaflet_AriaExpanded_Fix_3

src/control/Control.Layers.js
```diff
@@ -184,15 +184,34 @@ export var Layers = Control.extend({
...
+			() => {
+				keyboardFocus = false;
+			}
+		);
+		DomEvent.on(link, 'keydown', () => {
+			if (keyboardFocus) {
+				this.expand();
+				link.setAttribute('aria-expanded', 'true');
+			}
+		});
```
Malvoz, line 207: `keydown` is intercepted by both the NVDA and Narrator [screen readers](https://leafletjs.com/examples/accessibility/#test-your-maps), and the control is not toggled. `click` works OTOH.

ShehanAT:
Added the fixes regarding `keydown` event. Here are more details:
1. Pressing <kbd>Tab</kbd> to navigate the map doesn't not open the `layers-control` unless it is focused on and then user presses <kbd>Enter</kbd>
2. I tested with both NVDA and Windows Narrator screen readers and the above point functions as expected while the screen readers are in use. 

Side note: Using Windows Narrator with input learning mode on doesn't allow the user to navigate the webpage by pressing <kbd>Tab</kbd>, therefore expanding the `layers-control` won't be possible.

Here is a GIF demonstrating the new changes:
Leaflet_AriaExpanded_Fix_4

src/control/Control.Layers.js
ShehanAT: I have addressed these issues in commit [3812689](https://github.com/Leaflet/Leaflet/pull/8202/commits/381268965b3badb0f34ae6eb446660026af2dadd). Let me know if it meets your requirements

Malvoz:
> Side note: Using Windows Narrator with input learning mode on doesn't allow the user to navigate the webpage by pressing <kbd>Tab</kbd>, therefore expanding the layers-control won't be possible.

Narrator's [input learning](https://support.microsoft.com/en-us/windows/chapter-2-narrator-basics-5ff4591e-7b6d-245e-c95d-ce83c0a1a8d4#:~:text=Explore%20your%20keyboard) is merely a mode for users to learn their keyboard layout and the associated commands, so that's not a problem. 😋

While `aria-expanded="false"` is conveyed now (and as such allows the user to understand that new content is available after activating the control) - `aria-expanded="true"` is not conveyed due to the fact that the control is `display: none` (and as such can't keep the focus to convey the new state), this can be considered a separate issue (which relates to the inability to close the layers menu from the menu/control itself, which is problematic in some cases, mostly on mobile). With that said, ideally the layers control always toggles the collapsed/expanded state (and `aria-expanded`), i.e. if you were to temporarily set `leaflet-control-layers-toggle` to `display: block!important` in dev tools, it should be possible to toggle on repeated clicks. However I feel that's stretching the intent of this PR. I've separated some of the problems with the layers control into smaller issues, but maybe there needs to be a holistic discussion about the fundamentals of the layers control first...

Malvoz:
> maybe there needs to be a holistic discussion about the fundamentals of the layers control first

@ShehanAT Sorry for this oversight, `aria-expanded` is [currently not appropriate](https://github.com/Leaflet/Leaflet/issues/7526#issuecomment-1158258359). I think the layers control can be improved in other areas as well, so will need a new discussion first. Thanks for taking the time, and again, sorry!

Issue #7846, opened 2021-12-08
Factory returns not the same as the class


Falke-Design:
`new L.Bounds()` has a different result as `L.point()`.

I didn't checked all factories , but same behavior in:
- Bounds
- Point
- LatLng
https://plnkr.co/edit/bfWXNSffczjx9Lsp

Mostly in the factory is the part:
```
if (x === undefined || x === null) {
	return x;
}
```

I suggest to remove this checks. What do you think?

johnd0e:
Related: #7128, #7767

PR #8214, opened 2022-05-04
Add Validation and Number parsing

### LINKED ISSUE #7372: map.panInside() does not pan along y axis when "padding" option is set

Falke-Design:
Includes: #7768

### Validation:
Supersede PRs: #7767, #7432, #7418, #7765
Fix: #6072, fix #7372, fix: #7128,  

Maybe: #7761 

### Facotry methods:
I moved the logic from the factory methods of Point, Bounds, LatLng, LatLngBounds to the constructers to make sure that a `L.bounds` returns the same result as `L.Bounds()`.
fix: #7846


src/core/Util.js
```diff
@@ -239,3 +239,26 @@ export function cancelAnimFrame(id) {
...
 	}
 }
+
+function isNumeric(str) { // https://stackoverflow.com/a/175787/2520247
+	return !isNaN(str) && !isNaN(parseFloat(str));
+}
+
+// keep it private intentionally, because such function implementation
+// heavily depends on it's particular application and unable to suit all the cases
+export function _checkNumber(a) {
```
IvanSanchez, line 249: I'm going to nitpick on the semantics here. This doesn't *check* for a `Number`, but rather *casts* the parameter into a `Number`.
I'll argue that this function should instead be called `castFinite`. No real reason to keep this private or undocumented IMO.

Falke-Design: We had already the same discussion [here](https://github.com/Leaflet/Leaflet/pull/7768#discussion_r747495070) with no result.

Yes the function is casting to a number but also throws an error if it is not a number `throw new Error('Number expected');`. I personally would like to keep `Number` instead of `Finite` but if we call it `cast` or `check` doesn't matter for me.

IvanSanchez:
There's an edge case that I think will be broken by this change: Setting the map's `maxBounds` to an infinite horizontal strip, in order to prevent vertical movement but allow horizontal (or viceversa in a `L.CRS.Simple`).

i.e.:
```
map.setMaxBounds([[-10, -Infinity],[10, Infinity]]);
```

I know this has been marked with a "breaking" tag. I'm undecided, since I haven't been hit by the problems this change would solve.

Falke-Design:
@IvanSanchez I changed it to a public function and also allow `Infinity`

jonkoops:
@Falke-Design considering the factory functions have been removed, I think this PR can be closed. Feel free to re-open if you feel differently.

Issue #8226, opened 2022-05-10
Event.composedPath()


Th3AnG3L:
Hello i have been using leaflet.js for a months but now i receive error in web browser:

Deprecated Feature Used
'Event.path' is deprecated and will be removed in M109, around January 2023. Please use 'Event.composedPath()' instead. See https://www.chromestatus.com/feature/5726124632965120 for more details.

or maybe i need new version leaflet.js, can someone help me out with builded new one .js file.
Thank you.

Falke-Design:
@Th3AnG3L you can use the already builded source files:
https://www.jsdelivr.com/package/npm/leaflet
```
<script src="https://cdn.jsdelivr.net/npm/leaflet@1.8.0/dist/leaflet-src.min.js"></script>
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/leaflet@1.8.0/dist/leaflet.css">
```

Th3AnG3L:
> @Th3AnG3L you can use the already builded source files: https://www.jsdelivr.com/package/npm/leaflet
> 
> ```
> <script src="https://cdn.jsdelivr.net/npm/leaflet@1.8.0/dist/leaflet-src.min.js"></script>
> <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/leaflet@1.8.0/dist/leaflet.css">
> ```

Thank you i will download it and will take a look.

Falke-Design:
@Th3AnG3L can you please set up a demo: https://leafletjs.com/edit

Th3AnG3L:
i cant because its implemented as plugin in the system, but i can show live demo
https://resido.thesky9.com/properties/6007-applegate-lane
- when click zoom or out it shows this error
- by functions it reads directly leaflet.js codes and i had download the last one 1.8.0 and when make that zoom in zoom out pop ups in browser this error

Falke-Design:
The error happens in Chromium Browsers while dragging the map. But to see the problem you need to open the issues Tab in the Developer Tools, the console is not enough (for me).

![grafik](https://user-images.githubusercontent.com/19800037/167694550-8becc504-8674-4275-bd32-1392bac38035.png)

IvanSanchez:
Then this is the chromium equivalent of #7078, triggers when copying all properties of an event. Safe to ignore.

Lookintothesoul:
Thanks Chrome, I spent an hour to figure out what was wrong and thanks, now I know where to look for chrome updates

PR #8227, opened 2022-05-11
DomEvent.DoubleTap ignores clicks on <label>s

### LINKED ISSUE #8225: Firefox 100: Clicking control fires ghost dblclick
**Steps to reproduce**
- Click the control in the top right corner with Leaflet 1.7.1 activated
- Click the control in the top right corner with Leaflet 1.8.0 activated

**Expected behavior**
The map has a button in the top right corner and a text area in the bottom left. Clicking the button should toggle the area's visibility.

**Current behavior**
Clicking the button not only toggles the area's visibility, but also zooms in on the map. This only happens in Firefox, not in Chrome, since upgrading to leaflet 1.8.0.

**Environment**
- Leaflet version: 1.8.0
- Browser (with version): Firefox 100.0

~~See this JSFiddle: https://jsfiddle.net/petergrondelaers/pvk4hxmw/~~

https://jsfiddle.net/51bLtj9x/ or https://run.plnkr.co/preview/cl30e37zy0007206kqu0it63k/

IvanSanchez:
Fixes #8225.

When clicking on a `<label>` linked to an `<input>` (or vice versa), browsers trigger an extra click on the related element, as per https://html.spec.whatwg.org/multipage/forms.html#the-label-element . This causes the `DomEvent.DoubleTap.js` logic to activate, and fire a ghost `dblclick` event.

So after sleeping on the problem, I found what should be a reliable approach: if the problem is the extra clicks that browsers introduce on `<label>`s... ignore those! Get the propagation path of the event, and check if:
- The event happened on a `<label>`, or on a child of a `<label>`
- The `<label>` has a `for` attribute (thus ignoring standalone labels and nested labels)
- The event is not happening on an `<input>` or child of an `<input>` (which means a nested label)

I think this approach is sound, but I haven't tested it extensively (or, to be honest, at all). I'm assuming that browsers do *not* trigger an extra click when the `<input>` is nested within a `<label>`, but I might be wrong about that.

Do note that `composedPath` is not implemented by all browsers (hence the `if (e.composedPath)` branch), so... (*checks notes*)... IE, Edge<74 and Safari<=9 would still display the buggy behaviour of #8225.

cc @pgrondelaers

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -1,3 +1,5 @@
+import * as DomEvent from './DomEvent';
+
 /*
  * Extends the event handling code with double tap support for mobile browsers.
  *
@@ -49,20 +51,18 @@ export function addDoubleTapListener(obj, handler) {
 		// This ignores clicks on elements which are a label with a 'for'
 		// attribute (or children of such a label), but not children of
 		// a <input>.
-		if (e.composedPath) {
-			var path = e.composedPath();
-			if (path.some(function (el) {
-				return el instanceof HTMLLabelElement && el.attributes.for;
-			}) &&
-				!path.some(function (el) {
-					return (
-						el instanceof HTMLInputElement ||
-						el instanceof HTMLSelectElement
-					);
-				})
-			) {
-				return;
-			}
+		var path = DomEvent.getPropagationPath(e);
+		if (path.some(function (el) {
+			return el instanceof HTMLLabelElement && el.attributes.for;
+		}) &&
+			!path.some(function (el) {
+				return (
+					el instanceof HTMLInputElement ||
+					el instanceof HTMLSelectElement
+				);
+			})
+		) {
+			return;
 		}
 
 		var now = Date.now();
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -44,6 +44,27 @@ export function addDoubleTapListener(obj, handler) {
 			return;
 		}
 
+		// When clicking on an <input>, the browser generates a click on its
+		// <label> (and vice versa) triggering two clicks in quick succession.
+		// This ignores clicks on elements which are a label with a 'ref'
+		// attribute (or children of such a label), but not children of
+		// a <input>.
+		if (e.composedPath) {
+			var path = e.composedPath();
+			if (path.some(function (el) {
+				return el instanceof HTMLLabelElement && el.attributes.ref;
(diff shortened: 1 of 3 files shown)


src/dom/DomEvent.DoubleTap.js
```diff
@@ -44,6 +44,27 @@ export function addDoubleTapListener(obj, handler) {
...
 
+		// When clicking on an <input>, the browser generates a click on its
+		// <label> (and vice versa) triggering two clicks in quick succession.
+		// This ignores clicks on elements which are a label with a 'ref'
+		// attribute (or children of such a label), but not children of
+		// a <input>.
+		if (e.composedPath) {
+			var path = e.composedPath();
+			if (path.some(function (el) {
+				return el instanceof HTMLLabelElement && el.attributes.ref;
```
Malvoz, line 55: Is `ref` supposed to be [`for`](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for)?

IvanSanchez: ...and that's why I should be fully awake before sending PRs.

Falke-Design: Then the following is wrong too?
> The <label> has a `ref` attribute (thus ignoring standalone labels and nested labels)

Falke-Design:
Can we use `Event.path` instead of `composedPath`? Also see: #8226

Malvoz:
> Can we use `Event.path` instead of `composedPath`? Also see: #8226

Would that result in a compatibility issue?

from https://chromestatus.com/feature/5726124632965120#comments:

> Existing pages may use the following polyfill to keep functions:
> 
> ```js
> if (!Event.prototype.path) {
>   Object.defineProperty(Event.prototype, 'path', {
>     get() { return this.composedPath(); }
>   });
> }
> ```

IvanSanchez:
> Can we use `Event.path` instead of `composedPath`? Also see: #8226

And do something like `var path = (e.composedPath && e.composedPath()) || e.path || []`? 

I think the answer is "not really". According to a comment on https://github.com/tinymce/tinymce/issues/4490 , it's chrome-specific, not IE-specific.

Might as well polyfill as

```js
L.DomEvent.getEventPath = function (ev) {
  if (ev.composedPath) {
    return ev.composedPath();
  }

  var path = [];
  var el = ev.target;

  while (el) {
    path.push(el);
    el = el.parentNode;
  }
  return path;
};
```

Falke-Design (approved):
@IvanSanchez I added also an test, please take a look on it before merging

Issue #6551, opened 2019-03-11
Closing Popup by pressing Escape doesn't work in all cases


Schleuse:
### Description

The keyboard handler adds the `keydown`-Listener to the whole document and not to the map-container element.

In order not to interfere with other keyboard-events, the `keydown`-listener gets registered when the map-container receives focus and gets removed the a blur event on the map-container occurs.

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/map/handler/Map.Keyboard.js#L56-L59

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/map/handler/Map.Keyboard.js#L71-L74

https://github.com/Leaflet/Leaflet/blob/37d2fd15ad6518c254fae3e033177e96c48b5012/src/map/handler/Map.Keyboard.js#L132-L138

When the marker gets clicked, it receives focus, therefore the map container looses focus and triggers a blur event. The blur event disables the Keyboard-Handler and the opened popup can't be closed by pressing Escape.

- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
- Add a marker to a map
- Bind a popup to that marker
- Click the marker to open the popup
- Try to close the opened popup by pressing <kdb>Esc</kdb>

**Expected behavior**
The opened popup should be closed then the <kdb>Esc</kdb>-Key is pressed

**Current behavior**
The popup get's only closed when the layer the popup is bound to is not focusable

**Environment**
- Leaflet version: 1.4
- Browser (with version): Chrome 73
- OS/Platform (with version): Windows 10

**Minimal example reproducing the issue**

https://next.plnkr.co/edit/Soh1otCSsfdBBCi0

- [x] this example is as simple as possible
- [x] this example does not rely on any third party code

Schleuse:
I'm wondering why the `keydown` listeners are added to the `document` and not the map-container.
Is it necessary for some legacy browser?

I've made a local branch where the KeyBoard-Handler add the listeners to container and note the whole `document`. By doing that it is not necessary to add and remove the listeners on `focus` and `blur` and it seems somewhat simpler.

[...]

PR #8239, opened 2022-05-15
fix inconsistency when pressing escape to close popup (#6551)

joeyskuo:
Modified the Popup class's Layer to include a keydown eventListener to close the popup upon esc keydown

fix for issue #6551

### DIFF
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -369,6 +369,7 @@ Layer.include({
 			this.on({
 				click: this._openPopup,
 				keypress: this._onKeyPress,
+				keydown: this._onKeyDown,
 				remove: this.closePopup,
 				move: this._movePopup
 			});
@@ -385,6 +386,7 @@ Layer.include({
 			this.off({
 				click: this._openPopup,
 				keypress: this._onKeyPress,
+				keydown: this._onKeyDown,
 				remove: this.closePopup,
 				move: this._movePopup
 			});
@@ -472,6 +474,12 @@ Layer.include({
 	_onKeyPress: function (e) {
 		if (e.originalEvent.keyCode === 13) {
 			this._openPopup(e);
+		}
+	},
+
+	_onKeyDown: function (e) {
+		if (e.originalEvent.keyCode === 27) {
+			this.closePopup(e);
 		}
 	}
 });


IvanSanchez:
@joeyskuo This needs to be tested in several browsers and platform. In how many different platforms have you tested this?

I'm not entirely sure this will fix all the issues from #6551 - are popups focusable elements capable of receiving `keydown` events reliably?

joeyskuo:
> @joeyskuo This needs to be tested in several browsers and platform. In how many different platforms have you tested this?
> 

Hi Ivan, thank you for the review; I have tested this change in the following configurations:

**Dev Environment**
Leaflet version: 1.8.0
Browser (with version): Chrome 101.0.4951.54 (Official Build) (x86_64)
OS/Platform (with version): macOS Mojave 10.14.5

**Test Envrionment**
Viewing (https://plnkr.co/edit/4hXgn7EMinn9Wwn6)
Viewing (https://plnkr.co/edit/pkOzSCsI5QvtAcAe)

Leaflet version: 1.8.0
Browsers (with version): Safari 12.1.1
OS/Platform (with version): macOS Mojave 10.14.5

**Using Safari UserAgent**
Viewing (https://plnkr.co/edit/4hXgn7EMinn9Wwn6)

Leaflet version: 1.8.0
Browsers: Microsoft Edge
IE 11
IE 10
Firefox - Mac
Firefox - Windows
Chrome - Windows

> I'm not entirely sure this will fix all the issues from #6551 - are popups focusable elements capable of receiving `keydown` events reliably?

The popups aren't focused/focusable from what I've worked with so far. The keydown events are triggered when the marker is focused. Referencing the original poster,

**Steps to reproduce**
Add a marker to a map
Bind a popup to that marker
Click the marker to open the popup
Try to close the opened popup by pressing Esc

**Expected behavior**
The opened popup should be closed then the Esc-Key is pressed

**Current behavior**
The popup get's only closed when the layer the popup is bound to is not focusable

> I'm not entirely sure this will fix all the issues from #6551

I've noticed that in the test case (https://plnkr.co/edit/4hXgn7EMinn9Wwn6), if a popup is opened and you tab through until you focus on a child element of the popup (either the hyperlink or the close button), the esc button does not close the popup. I think that closely relates to this issue but I can agree that this commit does not address that

IvanSanchez:
> The popups aren't focused/focusable from what I've worked with so far. The keydown events are triggered when the marker is focused. 

Ah, I misunderstood your approach. In that case, I suggest moving the functionality to the already existing `_onKeyPress` method, the one that checks against keycode 13 ("Enter" key).

I'm slightly concerned that this might fail for popups on non-interactive layers (since they're not focusable). But since this shouldn't have any negative side effects, this addition should be fine.

joeyskuo:
> In that case, I suggest moving the functionality to the already existing `_onKeyPress` method, the one that checks against keycode 13 ("Enter" key).

I initially added the functionality to the '_onKeyPress' method, but it didn't end up working due to the onKeyPress event not being fired for all keys (e.g. ALT, CTRL, SHIFT, ESC) in all browsers. After rediscovering that I added it to the '_onKeyDown' event

> 
> I'm slightly concerned that this might fail for popups on non-interactive layers (since they're not focusable). But since this shouldn't have any negative side effects, this addition should be fine.

It may not affect non-interactive layers due to focus as you said

Issue #8112, opened 2022-04-04
Tooltips should be (reliably and deliberately) announced to screen reader users


Malvoz:
**Is your feature request related to a problem? Please describe.**

When tooltips are toggled visible they're not reliably announced by screen readers.

This is because screen readers generally don't announce elements on DOM updates or changes to CSS (e.g. `display`).
(I.e. https://github.com/Leaflet/Leaflet/issues/8111 doesn't help the case for screen reader users.)

(I say "not _reliably_ announced" because a side effect of the focus loss described in https://github.com/Leaflet/Leaflet/issues/8114 causes Narrator to announce _all_ tooltips when only 1 is toggled opened. OTOH, in NVDA tooltips are not announced at all after being toggled visible.)

**Describe the solution you'd like**

For each open tooltip:

1. Set [`role="tooltip"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) (on `leaflet-tooltip`)
2. Set [`aria-describedby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) on the bound element(s) with an `id` reference to the tooltip(s)
  (space-separated `id` references if pointing to multiple tooltips)

<details>
<summary>Caveat: <i>standalone</i> tooltips</summary><br>

This proposed solution is not applicable to standalone tooltips since there's no element to be described using `aria-describedby`. No natural element anyway, could perhaps be hacked around... Anyway, this is the safest option, and as such another reason for implementers to [_try to avoid standalone tooltips_](https://github.com/Leaflet/Leaflet/discussions/8006#:~:text=Mention%20that%20standalone%20popups%20and%20tooltips%20can%20be%20problematic).

<hr>
</details>

**Describe alternatives you've considered**

<details>
<summary>1. Make the <code>leaflet-tooltip-pane</code> an <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions">ARIA Live region</a></summary><br>

[...]

Issue #8111, opened 2022-04-04
Show tooltips on focus


Malvoz:
**Is your feature request related to a problem? Please describe.**

Tooltips are toggled visible if the element that is bound to it is hovered, but not on focus.

The current behavior doesn't allow keyboard-only users to access tooltips. To confirm this behavior:

1. Go to `debug/map/tooltip.html`
2. Focus any marker that does not have its tooltip initially visible

**Describe the solution you'd like**

I suppose:

- trigger `openTooltip` when the bound element is _focused_ (in concjunction to hover)
- `closeTooltip` if the user moves away from the element

**Additional context**

- https://www.w3.org/TR/wai-aria-practices/#tooltip (emphasis mine):

  > A tooltip is a popup that displays information related to an element _**when the element receives keyboard focus** or the mouse hovers over it_.
- Similar problem to: https://github.com/Leaflet/Leaflet/issues/8078, for keyboard users.

PR #8247, opened 2022-05-16
Tooltip accessibility improvements: focus and voice over

alekzvik:
This PR addresses #8111 & #8112.

I've used the approach described in mentioned issues. I had to add logic for checking if the layer we're working with is a single `Layer` (one Marker, Polygon, etc.) or a `LayerGroup` (e.g., a group of Markers).

Changes include:
1. The layer a tooltip is bound to now has an `aria-describedby` attribute
2. Every tooltip has an id to accommodate the change above
3. Tooltips are open when acquiring focus and close on blur

Here is a before/after demo on a `debug/map/tooltip.html` with Voice Over narration visible:
<details><summary>Before</summary>

</details>

<details><summary>After</summary>

</details>

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -305,21 +305,18 @@ Layer.include({
 
 	_addFocusHooks: function () {
 		if (this.getElement) {
-			this._addListenersOnLayer(this);
+			this._addFocusListenersOnLayer(this);
 		} else if (this.eachLayer) {
-			this.eachLayer(this._addListenersOnLayer.bind(this));
+			this.eachLayer(this._addFocusListenersOnLayer.bind(this));
 		}
 	},
 
-	_addListenersOnLayer: function (layer) {
-		DomEvent.on(layer.getElement(), 'focus', () => this._openTooltipWithExplicitSource(layer));
+	_addFocusListenersOnLayer: function (layer) {
+		DomEvent.on(layer.getElement(), 'focus', () => {
+			this._tooltip._source = layer;
+			this.openTooltip();
+		});
 		DomEvent.on(layer.getElement(), 'blur', this.closeTooltip, this);
-	},
-
-	// since the focus event comes from an HTML Element and does not have .layer or .map downstream, we have to explicitly set _source to the leaflet class here.
-	_openTooltipWithExplicitSource: function (layer) {
-		this._tooltip._source = layer;
-		this.openTooltip();
 	},
 
 	// @method openTooltip(latlng?: LatLng): this
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -3,6 +3,8 @@ import {toPoint} from '../geometry/Point';
 import {Map} from '../map/Map';
 import {Layer} from './Layer';
 import * as DomUtil from '../dom/DomUtil';
+import * as DomEvent from '../dom/DomEvent';
+import * as Util from '../core/Util';
 
 /*
  * @class Tooltip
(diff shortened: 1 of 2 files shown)


Malvoz:
Will need others to review the logic, but this looks really good IMO, @alekzvik!

src/layer/Tooltip.js
```diff
@@ -115,6 +117,9 @@ export var Tooltip = DivOverlay.extend({
 		    className = prefix + ' ' + (this.options.className || '') + ' leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide');
 
 		this._contentNode = this._container = DomUtil.create('div', className);
+
+		this._container.role = 'tooltip';
+		this._container.setAttribute('id', 'tooltip-' + Util.stamp(this));
```
Malvoz, line 122: I believe we'll want a `leaflet-tooltip-` prefix on these IDs.

src/layer/Tooltip.js
```diff
@@ -115,6 +117,9 @@ export var Tooltip = DivOverlay.extend({
 		    className = prefix + ' ' + (this.options.className || '') + ' leaflet-zoom-' + (this._zoomAnimated ? 'animated' : 'hide');
 
 		this._contentNode = this._container = DomUtil.create('div', className);
+
+		this._container.role = 'tooltip';
```
Malvoz, line 121:
```suggestion
		this._container.setAttribute('role', 'tooltip');
```

src/layer/Tooltip.js
```diff
@@ -341,6 +357,27 @@ Layer.include({
...
+	_addFocusListeners: function () {
+		if (this.getElement) {
+			this._addFocusListenersOnLayer(this);
+		} else if (this.eachLayer) {
+			this.eachLayer(this._addFocusListenersOnLayer, this);
+		}
+	},
+
+	_addFocusListenersOnLayer: function (layer) {
+		DomEvent.on(layer.getElement(), 'focus', () => {
```
Malvoz, line 369: Arrow functions are not supported in IE.

Malvoz, line 369: Arrow functions are not supported in IE.

Malvoz (approved):

mourner (approved):
👍 Looks great to me!

MartijnHarmenzon:
Hi all. Is there a way to prevent this behaviour?

I have CircleMarkers of which the color changes when clicked on. But now they are getting the black and white outlines?

jonkoops:
@MartijnHarmenzon if this change is causing problems for you could you create a new issue with a reproducible case? This makes things easier to keep track of for us.

Malvoz:
> I have CircleMarkers of which the color changes when clicked on (using the setStyle function). But now they are getting the black and white outlines?

@MartijnHarmenzon The outlines (or other [obvious focus styles](https://developer.mozilla.org/en-US/docs/Web/CSS/outline#accessibility_concerns)) are essential for keyboard-only users when navigating features. You can keep the outlines only if the UA determines them "necessary" (pretty much boils down to programmatic focus + keyboard focus) by utilizing [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) like so:

```css
.leaflet-container :focus:not(:focus-visible) {
  outline: 0;
}
```

FloChehab:
Hey @Malvoz,

We have just encountered a similar issue as @MartijnHarmenzon .

We obviously get the accessibility requirements (and it's actually super nice with a keyboard), but it's causing visual boxes on Google Chrome (not Firefox) by using our mouse and no keyboards, which make things look buggy (we get the "box outline" on a click on a polygon without touching the keyboard from my point of view this is UX/UI bug).
I'll try to build a reproducible example in the coming days.

image

Malvoz:
@FloChehab Please create a new issue for discussion.

PR #8250, opened 2022-05-18
Marker accessibility: conditional `role="button"` and `aria-expanded`

### LINKED ISSUE #7527: Markers that bind to popups should convey the `aria-expanded` state of the popup
Same as https://github.com/Leaflet/Leaflet/issues/7526, but for interactive markers:

Interactive markers that bind to popups should convey the [`aria-expanded`](https://www.w3.org/TR/wai-aria-1.2/#aria-expanded) [state](https://www.w3.org/TR/wai-aria-1.2/#dfn-state) of the corresponding popup.

### LINKED ISSUE #8116: Discern interactive markers from non-interactive markers

`role="button"` informs screen reader users that it is an interactive element, and that they can expect something to happen once activated.

The role is currently set unconditionally on markers (introduced in https://github.com/Leaflet/Leaflet/pull/7895, allows screen reader users to activate markers).

Ideally, `role="button"` should only be set on markers if they actually do something upon activation (e.g. if they toggle a popup or tooltip, for example).

Similar to [Google Maps](https://cloud.google.com/blog/products/maps-platform/improved-accessibility-maps-javascript-api):

> Markers now default to the “img” accessibility role, or the “button” accessibility role if they are interactive (such as when a click event listener is registered)

**Additional context**

A conditional for setting attributes based on whether a marker is interactive or not is also necessary for https://github.com/Leaflet/Leaflet/issues/7527.

alekzvik:
- Fixes #8116
- Fixes #7527

**Changes include:**
- Marker has default `role` of `img` as a non-interactive element.
- When Popup is bound, it sets the `role` of the element to `button`
- When Popup is bound, it reflects its state (open/close) in the `aria-expanded` attribute of the bound element (as true/false)

**Demo**
<details><summary>Demo</summary>

</details>

**Discussion points**
Both issues mentioned discuss Marker/Popup interaction specifically, but changes made here are on the layer level. This part is controversial since I am not sure if this is the responsibility of a Popup or a Marker. I am open to discussion and feedback here. The main problem I see here is that the behavior will work for some Layers and not others.

**Alternative**
The alternative implementation I see is to move this code to Marker and add Popup events for bind/unbind marker can react to. This way, the behavior will work only for Marker/Popup interaction and be more explicit.

Please let me know what you think in the comments.

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -235,7 +235,7 @@ export var Marker = Layer.extend({
 
 		if (options.keyboard) {
 			icon.tabIndex = '0';
-			icon.setAttribute('role', 'button');
+			icon.setAttribute('role', 'img');
 		}
 
 		this._icon = icon;
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -69,7 +69,7 @@ describe('Popup', function () {
 		map.addLayer(marker);
 
 		marker.bindPopup('Popup1');
-		var element = marker.getElement()
+		var element = marker.getElement();
 		expect(element.getAttribute('aria-expanded')).to.be('false');
 
 		// toggle open popup
@@ -84,7 +84,7 @@ describe('Popup', function () {
 	it("changes the role of a marker to button", function () {
 		var marker = L.marker(center);
 		map.addLayer(marker);
-		var element = marker.getElement()
+		var element = marker.getElement();
 		expect(element.getAttribute('role')).to.be('img');
 
 		marker.bindPopup('Popup1');
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -64,6 +64,33 @@ describe('Popup', function () {
 		expect(map.hasLayer(marker._popup)).to.be(false);
 	});
 
+	it("reflects its visibility when marker is clicked in aria-expanded", function () {
+		var marker = L.marker(center);
+		map.addLayer(marker);
+
+		marker.bindPopup('Popup1');
+		var element = marker.getElement()
+		expect(element.getAttribute('aria-expanded')).to.be('false');
+
+		// toggle open popup
+		happen.click(marker._icon);
+		expect(element.getAttribute('aria-expanded')).to.be('true');
+
+		// toggle close popup
+		happen.click(marker._icon);
+		expect(element.getAttribute('aria-expanded')).to.be('false');
+	});
+
+	it("changes the role of a marker to button", function () {
+		var marker = L.marker(center);
+		map.addLayer(marker);
+		var element = marker.getElement()
+		expect(element.getAttribute('role')).to.be('img');
+
+		marker.bindPopup('Popup1');
+		expect(element.getAttribute('role')).to.be('button');
+	});
+
 	it("it should use a popup with a function as content with a FeatureGroup", function () {
 		var marker1 = L.marker(center);
 		var marker2 = L.marker([54.6, 38.2]);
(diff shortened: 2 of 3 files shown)


Malvoz:
> changes made here are on the layer level. This part is controversial since I am not sure if this is the responsibility of a Popup or a Marker. I am open to discussion and feedback here. The main problem I see here is that the behavior will work for some Layers and not others.

Is that the reason why these changes don't affect markers in e.g. `debug/map/map.html`?

src/layer/marker/Marker.js
```diff
@@ -235,7 +235,7 @@ export var Marker = Layer.extend({
 
 		if (options.keyboard) {
 			icon.tabIndex = '0';
-			icon.setAttribute('role', 'button');
+			icon.setAttribute('role', 'img');
```
Malvoz, line 238: For `` markers this `role` is redundant (because that element is implicitly `role="img"`).

But I'm not sure it's desired for divIcons either, because they can have custom HTML, so implementers may end up with invalid HTML (the same consideration applies to the use of `role="button"` and `aria-expanded`  for divIcons). Perhaps it should be their responsibility to not use elements that are disallowed inside `img` (all) or buttons (interactive elements) per the HTML spec, but that'd restrict the kind of HTML elements that should be used in the custom HTML in divIcons.

@mourner, @IvanSanchez, @Falke-Design, I believe we need to decide whether divIcons should be as accessible as regular markers by default but which may result in invalid HTML if the implementer isn't careful, or simply not have the same accessible defaults.

alekzvik:
> Is that the reason why these changes don't affect markers in e.g. `debug/map/map.html`?

Thanks for spotting. This example does not work because of `LayerGroup` usage. Since I am manipulating underlying DOM objects, I will need to add code to call `getElement` for each layer of the LayerGroup, similar to what I did [here](https://github.com/Leaflet/Leaflet/blob/main/src/layer/Tooltip.js#L361-L365)

Malvoz:
It's not clear to me that this should apply to `divIcon`s. I retracted an earlier review comment, in an attempt to clarify, let's break it down:

- `role="img"`:

  - `img` is not named from (child) content in [custom HTML](https://leafletjs.com/reference.html#divicon-html):<br>
    
  - and is redundant for `` element markers

- `role="button"` (and `aria-expanded`):

  Can cause invalid HTML, e.g. if custom HTML is used to insert a `<button>` inside (now) `role="button"`

- `divIcon` is named that way because it uses a `div`, but it is technically not just a `div` if we change the semantics using `role`

alekzvik:
> * `role="img"`:

Yes, seems like `role` `img` is redundant.

> * `role="button"` (and `aria-expanded`):
> Can cause invalid HTML, e.g. if custom HTML is used to insert a `<button>` or `<a href>` inside (now) `role="button"`

The mitigation strategy that comes to mind is to make an option to disable `role` and `aria-expanded` modifications.

> * `divIcon` is named that way because it uses a `div`, but it is technically not just a `div` if we change the semantics using `role`.

But if we add handlers for `click` and `keypress` to open a popup, that makes it a button, and the button role is used here to clarify this interaction possibility to the screen reader.

src/layer/marker/Marker.js
Malvoz, line 238:
```suggestion

```

Per https://github.com/Leaflet/Leaflet/pull/8250#issuecomment-1131515329.

ImprobableGenius:
The current implementation sets the `role='button'` for the markers which are rendered as `img` tags. This makes the `alt` attribute invalid, therefore we've to use `aria-label` to provide alternate marker text. I am wondering if it is possible to follow semantic html and wrap the `img` tags inside a `button`? This will help meet all interactivity requirements out of the box and will streamline usage for fellow developers. 

Let me know if there are questions, I am using a custom `L.icon` implementation on one of my project to do just that and can jump in and contribute if needed.

Issue #7822, opened 2021-11-28
Accessible vectors


Malvoz:
**Is your feature request related to a problem? Please describe.**

> There is a larger issue that I really don't have a solution for: All vectors, not just markers, need to be tabbable. The way things are currently, only markers are tabbable. This means that VoiceOver and other accessibility technologies won't work for other vectors.

_Originally posted by @nateirwin in https://github.com/Leaflet/Leaflet/issues/3210#issuecomment-73268793_

Malvoz:
Adding some substance to this issue:

For basic keyboard and screen reader accessibility, vectors would:

- [ ] Need to be focusable (i.e. set the SVG [`tabindex`](https://svgwg.org/svg2-draft/struct.html#tabindexattribute) attribute to `0`), similar to markers:

  https://github.com/Leaflet/Leaflet/blob/2960f58a02e0858f9403788f14fa9f4bd9fd2736/src/layer/marker/Marker.js#L226-L227

- [ ] Have a default [accessible name](https://www.w3.org/TR/accname/#dfn-accessible-name), similar to markers:

  https://github.com/Leaflet/Leaflet/blob/2960f58a02e0858f9403788f14fa9f4bd9fd2736/src/layer/marker/Marker.js#L43-L45

  - What should the default accessible name be?
     
     IIRC, using NVDA with Firefox: "blank" is announced for e.g. SVG `<path>`, whereas in Chrome: "graphic" is announced. Maybe vectors should default to `Graphic`.
  - How should it be exposed?
    
     Probably as `aria-label`, with an option for tooltips via the `title` attribute or the SVG `<title>` element, similar to markers:

     https://github.com/Leaflet/Leaflet/blob/2960f58a02e0858f9403788f14fa9f4bd9fd2736/src/layer/marker/Marker.js#L39-L41

- [ ] If a vector binds to a popup or otherwise is clickable it should have `role="button"`, similar to markers: https://github.com/Leaflet/Leaflet/blob/2960f58a02e0858f9403788f14fa9f4bd9fd2736/src/layer/marker/Marker.js#L228

Further enhancements:

- [ ] If a vector binds to a popup it should convey the expanded/collapsed state of the popup using `aria-expanded="true|false`, same as for markers: [#7527](https://github.com/Leaflet/Leaflet/issues/7527). `aria-expanded` relies on the presence of `role="button"`.

Malvoz:
[...]

Issue #8114, opened 2022-04-04
Focus shifts to `leaflet-container` after activating markers when using a screen reader


Malvoz:
> **Warning**
>
> This issue is blocked on <s>[#8112](https://github.com/Leaflet/Leaflet/issues/8112) and</s> [#8115](https://github.com/Leaflet/Leaflet/issues/8115)</ins>, or there'll be accessibility regressions in some screen readers! This is because the side effect of the focus shift described here actually causes some screen readers to announce tooltips and popups.

- [x] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [x] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [x] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**

1. Go to [quick-start/example.html](https://leafletjs.com/examples/quick-start/example.html)
2. Turn on a [screen reader of your choice](https://leafletjs.com/examples/accessibility/#test-your-maps) (note that you can pause speech by pressing <kbd>ctrl</kbd> at any time, if desired)
3. <kbd>Tab</kbd> to the marker and press <kbd>Enter</kbd>

**Expected behavior**

When activating the marker, focus should stay on the marker.

**Current behavior**

When activating the marker, focus is shifted to the map container.

This results in a reset of the focus order for keyboard users, and irrelevant and verbose announcements for screen reader users. It is a failure of WCAG success criterion [3.2.2 On Input](https://www.w3.org/TR/WCAG22/#on-input).

**Environment**

- Leaflet version: 1.7.1 / 1.8.0 / 1.9.4
- Browser (with version): Chrome 127.0.6533.90
- OS/Platform (with version): Windows 10
- Assistive tech: Narrator, NVDA

**Additional context**

  The current behavior has the side effect that popups actually are announced (although not _only_ the popup the user is interested in), so this issue is blocked on <s>https://github.com/Leaflet/Leaflet/issues/8112 and</s> https://github.com/Leaflet/Leaflet/issues/8115! Or there'll be accessibility regressions.

- The problem for end-users (who use screen readers) is pretty much the same as https://github.com/Leaflet/Leaflet/issues/7947.

MichaelAllenWarner:
FYI, I don't see this behavior with VoiceOver on my Mac. Focus just remains on the Marker, as expected.

PR #8251, opened 2022-05-18
Accessibility: make vector focusable

alekzvik:
This PR addresses #7822
This is my initial attempt that is working. It should also play nice with #8250, but additional testing is needed. 

**Changes include**:
- Vectors rendered in SVG have `role` set to `img` and `tabIndex` to be focusable by keyboard
- Vectors rendered in Canvas have "shadow" elements for each. Shadow elements have a role set to `img` and `tabIndex,` making them focusable by a keyboard. When the shadow element is focused, `redraw` is called, and the tied canvas element is outlined.
- `Path` class has new option `keyboard` with default `true` similar to `Marker`
- New page in the debug that renders the same simple map with both SVG and Canvas (see on the demo)

**Areas of improvement**
- Popups do not open with keyboard in Canvas but work in SVG

**Demo**
<details><summary>Demo</summary>

</details>

**Possible discussion**
I tied canvas shadow elements to `layer._path`. `layer._path` is already used in SVG to store the actual path element and is returned by Path's `getElement.` On one hand., using `_path` is confusing. Still, it allows us to utilize changes for popup interaction that rely on `getElement` in #8250. I'll be happy to hear different opinions on the matter.

### DIFF
--- a/debug/vector/simple-canvas-svg.html
+++ b/debug/vector/simple-canvas-svg.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<link rel="stylesheet" href="../css/mobile.css" />
+
+	<script src="../../dist/leaflet-src.js"></script>
+</head>
+<body>
+	<h1>SVG</h1>
+	<div id="svg-map" style="height: 40%"></div>
+	<h1>Canvas</h1>
+	<div id="canvas-map" style="height: 40%"></div>
+
+
+	<script>
+		var map1 = L.map('svg-map');
+		var map2 = L.map('canvas-map', {preferCanvas: true});
+
+		function populateMap(map) {
+			map.setView([51.505, -0.09], 13);
+
+			var marker = L.marker([51.5, -0.09])
+				.bindPopup("<b>Hello world!</b><br />I am a popup.")
+				.addTo(map);
+
+			var circle = L.circle([51.508, -0.11], 500, {color: '#f03', opacity: 0.7})
+				.bindPopup("I am a circle.")
+				.addTo(map);
+
+			var polygon = L.polygon([
+				[51.509, -0.08],
+				[51.503, -0.06],
+				[51.51,  -0.047]])
+				.bindPopup("I am a polygon.")
+				.addTo(map);
+
+			var osmUrl = 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+				osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+				osm = L.tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib}).addTo(map);
+		}
+
+		populateMap(map1);
+		populateMap(map2);
+	</script>
+</body>
+</html>
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -138,6 +138,13 @@ export var Canvas = Renderer.extend({
 	_initPath: function (layer) {
 		this._updateDashArray(layer);
 		this._layers[Util.stamp(layer)] = layer;
+		if (layer.options.keyboard) {
+			layer._path = DomUtil.create('div', 'leaflet-canvas-shadow-path', this._container);
+			layer._path.tabIndex = '0';
+			layer._path.setAttribute('role', 'img');
+			DomEvent.on(layer._path, 'focus', this._redraw, this);
+			DomEvent.on(layer._path, 'blur', this._redraw, this);
+		}
 
 		var order = layer._order = {
 			layer: layer,
@@ -269,6 +276,7 @@ export var Canvas = Renderer.extend({
 			layer = order.layer;
 			if (!bounds || (layer._pxBounds && layer._pxBounds.intersects(bounds))) {
 				layer._updatePath();
+				this._ctx.drawFocusIfNeeded(layer._path);
 			}
 		}
 
--- a/src/layer/vector/Path.js
+++ b/src/layer/vector/Path.js
@@ -71,7 +71,11 @@ export var Path = Layer.extend({
 		// @option bubblingMouseEvents: Boolean = true
 		// When `true`, a mouse event on this path will trigger the same event on the map
 		// (unless [`L.DomEvent.stopPropagation`](#domevent-stoppropagation) is used).
-		bubblingMouseEvents: true
+		bubblingMouseEvents: true,
+
+		// @option keyboard: Boolean = true
+		// Whether the marker can be tabbed to with a keyboard and clicked by pressing enter.
+		keyboard: true
 	},
 
 	beforeAdd: function (map) {
--- a/src/layer/vector/SVG.js
+++ b/src/layer/vector/SVG.js
@@ -104,6 +104,11 @@ export var SVG = Renderer.extend({
 			DomUtil.addClass(path, 'leaflet-interactive');
 		}
 
+		if (layer.options.keyboard) {
+			path.tabIndex = '0';
+			path.setAttribute('role', 'img');
+		}
+
 		this._updateStyle(layer);
 		this._layers[stamp(layer)] = layer;
 	},


src/layer/vector/SVG.js
```diff
@@ -104,6 +104,11 @@ export var SVG = Renderer.extend({
 			DomUtil.addClass(path, 'leaflet-interactive');
 		}
 
+		if (layer.options.keyboard) {
+			path.tabIndex = '0';
+			path.setAttribute('role', 'img');
```
Malvoz, line 109: SVG `path` is [implicitly](https://www.w3.org/TR/SVG2/struct.html#:~:text=%E2%80%98path%E2%80%99-,graphics%2Dsymbol%20role,-if%20the%20element) [`role="graphics-symbol"`](https://www.w3.org/TR/graphics-aria-1.0/#graphics-symbol) (albeit only [if `aria-label`, `aria-labelledby`, or `title` is used](https://www.w3.org/TR/SVG2/struct.html#TermInclusionCriteria)), I don't think there's a need to override it.

src/layer/vector/Canvas.js
```diff
@@ -138,6 +138,13 @@ export var Canvas = Renderer.extend({
 	_initPath: function (layer) {
 		this._updateDashArray(layer);
 		this._layers[Util.stamp(layer)] = layer;
+		if (layer.options.keyboard) {
+			layer._path = DomUtil.create('div', 'leaflet-canvas-shadow-path', this._container);
```
Malvoz, line 142: I think we should avoid the term "shadow", it sounds like Shadow DOM, or could perhaps be confused with marker shadows.

src/layer/vector/Canvas.js
```diff
@@ -138,6 +138,13 @@ export var Canvas = Renderer.extend({
...
+			layer._path.tabIndex = '0';
+			layer._path.setAttribute('role', 'img');
```
Malvoz, line 144:
```suggestion
			layer._path.setAttribute('role', 'graphics-symbol img');
```

Per the note in https://www.w3.org/TR/graphics-aria-1.0/#graphics-symbol:

> To support user agents and assistive technologies based on the ARIA 1.0 specification, authors may wish to include the [`img`](https://www.w3.org/TR/wai-aria-1.1/#img) role as a fallback value, in the form `role="graphics-symbol img"`, if that is not already the default semantic role for the element.

(Relates to my previous [comment](https://github.com/Leaflet/Leaflet/pull/8251#discussion_r876414033).)

src/layer/vector/Canvas.js
alekzvik: Agree, changed the class name to avoid confusion

src/layer/vector/SVG.js
alekzvik: Agree, now after I added `<title>`, it is unnecessary and gone.

src/layer/vector/SVG.js
```diff
@@ -104,6 +104,19 @@ export var SVG = Renderer.extend({
 			DomUtil.addClass(path, 'leaflet-interactive');
 		}
 
+		if (layer.options.keyboard) {
+			path.setAttribute('tabIndex', '0');
```
Malvoz, line 108:
```suggestion
			path.setAttribute('tabindex', '0');
```

alekzvik:
> @alekzvik I have updated the initial PR description of the changes to reflect some of the latest changes and also avoid the term "shadow".

Thank you, I have also updated the demo gif to reflect changes after adding the title/description.

src/layer/vector/Canvas.js
```diff
@@ -138,6 +138,16 @@ export var Canvas = Renderer.extend({
 	_initPath: function (layer) {
 		this._updateDashArray(layer);
 		this._layers[Util.stamp(layer)] = layer;
+		if (layer.options.keyboard) {
+			layer._path = DomUtil.create('div', 'leaflet-canvas-accessibility-hidden', this._container);
```
Malvoz, line 142:
```suggestion
			layer._path = DomUtil.create('div', 'leaflet-canvas-interactive', this._container);
```

Suggesting a different class name here. Similar but different to `leaflet-interactive`. I think `accessibility` and `hidden` sounds a bit contradictive, although I realize it probably intends to mean "visually" hidden. A shorter class name is also nice.

Malvoz:
> ****Areas of improvement****
> 
> - Popups do not open with keyboard in Canvas but work in SVG

The work on this PR is phenomenal and I highly appreciate it, however keyboard operability is quite important, do you think it should be implemented separately?

alekzvik:
> The work on this PR is phenomenal and I highly appreciate it, however keyboard operability is quite important, do you think it should be implemented separately?

I was going to suggest moving it into a separate PR, but I figured out how to do it, so it's no longer a problem.

src/layer/vector/Path.js
```diff
@@ -71,7 +71,21 @@ export var Path = Layer.extend({
...
+
+		// @option title: String= ''
+		// Alternative text for a Figure. Used as `<title>`` in SVG or `aria-label` in Canvas renderer.
+		// Useful for accessibility.
+		title: '',
+
+		// @option description: String= ''
+		// Text describing what is on the figure.
+		// Useful for accessibility.
+		description: '',
```
Malvoz, line 88: I suppose this option could be named only `desc` to reflect (in the SVG case) the `<desc>` element, similar to how the `alt` option is not named `alternative`~ for the `alt` attribute.

But let's hear what others think.

src/layer/vector/Canvas.js
```diff
@@ -145,6 +152,8 @@ export var Canvas = Renderer.extend({
 			layer._path.setAttribute('aria-label', layer.options.title);
 			layer._path.setAttribute('title', layer.options.description);
 
+
+			DomEvent.on(layer._path, 'keypress', Util.bind(this._onKeyPress, this, layer));
```
Malvoz, line 156: I notice that this does not work when using a screen reader (tested in Narrator and NVDA). Similar to https://github.com/Leaflet/Leaflet/pull/8202/commits/381268965b3badb0f34ae6eb446660026af2dadd.

src/layer/vector/Canvas.js
```diff
@@ -135,9 +135,28 @@ export var Canvas = Renderer.extend({
...
+
 	_initPath: function (layer) {
 		this._updateDashArray(layer);
 		this._layers[Util.stamp(layer)] = layer;
+		if (layer.options.keyboard) {
+			layer._path = DomUtil.create('div', 'leaflet-canvas-interactive', this._container);
+			layer._path.setAttribute('tabindex', 0);
+			layer._path.setAttribute('role', 'graphics-symbol img');
+			layer._path.setAttribute('aria-label', layer.options.title);
+			layer._path.setAttribute('title', layer.options.description);
```
Malvoz, line 153: I wonder if we should go with:

```diff
-layer._path.setAttribute('aria-label', layer.options.title);
+layer._path.setAttribute('title', layer.options.title);

-layer._path.setAttribute('title', layer.options.description);
+layer._path.setAttribute('aria-description', layer.options.description);
```

[`aria-description`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) is a relatively new ARIA attribute, and [not yet supported by VoiceOver](https://a11ysupport.io/tech/aria/aria-description_attribute#support-table-0). But it may be the correct thing to use, at least in the long run. 🤔

src/layer/vector/Canvas.js
Malvoz: Screen readers have a tendency to intercept keyboard events, not exactly sure under which conditions though.

Other cases where things work(ed) well without a screen reader, but doesn't if you are using one (may be totally unrelated):
- Popups only opening on <kbd>enter</kbd> if markers have `role="button"`: https://github.com/Leaflet/Leaflet/pull/7895#issue-1097336107
- https://github.com/Leaflet/Leaflet/issues/8114

alekzvik: Yes, I need to rethink this part and make it more robust. Unfortunately, I can not test things with Narrator or NVDA. But I currently can think of 2 possible reasons for this:
- Screen readers send may some other events (like a fake `click`), but I pass through only `keypress`. I will change that and pass other events too.
- Maybe it's related to the role being `graphics element` and not telling the user they can interact with it, as mentioned in the linked issue (#7895). If that's the case, #8250 should take care of that.

src/layer/vector/Canvas.js
alekzvik: I am testing the changes using VoiceOver on a mac laptop with 3 browsers: Firefox, Chrome & Safari. `aria-description` is announced in all of them.

[The link you mentioned](https://a11ysupport.io/tech/aria/aria-description_attribute#support-table-0) says it does not work for the `input` tag, but we use `div`, maybe that's the reason.

Malvoz: That's good to hear!

Yes, a11ysupport.io doesn't always test the most simple cases, which is kind of unfortunate. But that initiative was created and mostly maintained by one individual.

alekzvik: Oh, wait, I misunderstood you. `aria-description` does not work for me with VoiceOver. I was talking about the current implementation with `title` and `aria-label`.

If I use `aria-description` it is **not** announced as an additional description. But `title` is announced as an element name.
If I use `aria-label` & `title`, then `aria-label` is announced as an element name and `title` is used as a description.

src/layer/vector/Path.js
alekzvik: Although this is not as straightforward as `alt` in Marker since it has two implementations depending on a renderer, I think `desc` is descriptive enough and shorter, and I like the proposed change.

src/layer/vector/Canvas.js
Malvoz: Ah, unfortunate. With the current approach there will be confusion with `title` option setting `aria-label` and `desc(ription)` setting `title` attribute.

alekzvik: > Ah, unfortunate. With the current approach it may confuse people that the title option is setting aria-label and description option is setting the title attribute.

Agree, sounds confusing. We can introduce better names for those options. The name might be more independent of implementation. Something like `accessibleName` & accessibleDescription`. What do you think?

Malvoz: I've thought of that (well `accName` and `accDesc`), but we'd still be stuck with `alt` and `title` for markers to avoid breakage, so I think we'd want `title` here as well.

alekzvik: To sum up, we have thee options:
- use `title` and `desc` as option names. Use `title` & `aria-description` in canvas sub DOM. 
The naming is more consistent with the `Marker`, but this does not currently work with VoiceOver.
- use `title` & `desc` as option names. Use `aria-label` & `title` in canvas sub DOM.
The naming is more consistent with the `Marker`, but we introduce a source of confusion since `option.title` does not go into the `title` attribute.
- use `accName` & `accDesc` as option names. Use `aria-label` & `title` in canvas sub DOM.
The naming is not consistent with the `Marker`, and I do not believe we will change the option names of Marker. But the Voice Over works and works the same for canvas & SVG renderer.

I prefer the last option and am curious what others think.

src/layer/vector/Canvas.js
```diff
@@ -138,6 +138,18 @@ export var Canvas = Renderer.extend({
...
```
Malvoz, line 146: I notice the canvas `<title>` isn't producing a tooltip on hover (or focus, as is also the case in some browsers).

src/layer/vector/Canvas.js
Malvoz: > - use `accName` & `accDesc` as option names. [...]
>    The naming is not consistent with the `Marker`, and I do not believe we will change the option names of Marker.

With this option, I suppose there's also the possibility to introduce `accName` and `accDesc` as synonyms for `Marker`'s `alt` and `title` for consistency? I think consistency is appreciated by developers.

However I'm kind of leaning towards keeping just `alt` and `title` for Marker, and `desc` and `title` for canvas and SVG. I'm not opposed to use `aria-description` for the `desc` option, with a note in the docs stating it is not fully supported (with a link to a11ysupport.io) which could simply be removed in the future. Assuming this is an acceptable approach:
- developers will always know that the `title` option will produce the browser's native tooltips
- there's consistency
- the `title` options always give exactly that: `title` attribute, or `<title>`

src/layer/vector/CircleMarker.js
```diff
@@ -21,7 +21,12 @@ export var CircleMarker = Path.extend({
 
 		// @option radius: Number = 10
 		// Radius of the circle marker, in pixels
-		radius: 10
+		radius: 10,
+
+		// @option title: String= 'CircleMarker'
+		// Alternative text for a Figure. Used as <title> in SVG or aria-label in Canvas renderer.
```
Malvoz, line 27:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Path.js
```diff
@@ -71,7 +71,21 @@ export var Path = Layer.extend({
...
 		// (unless [`L.DomEvent.stopPropagation`](#domevent-stoppropagation) is used).
-		bubblingMouseEvents: true
+		bubblingMouseEvents: true,
+
+		// @option keyboard: Boolean = true
+		// Whether the marker can be tabbed to with a keyboard and clicked by pressing enter.
+		keyboard: true,
+
+		// @option title: String= ''
+		// Alternative text for a Figure. Used as `<title>`` in SVG or `aria-label` in Canvas renderer.
```
Malvoz, line 81:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Polygon.js
```diff
@@ -54,7 +54,12 @@ import * as PolyUtil from '../../geometry/PolyUtil';
 export var Polygon = Polyline.extend({
 
 	options: {
-		fill: true
+		fill: true,
+
+		// @option title: String= 'Polygon'
+		// Alternative text for a Figure. Used as <title> in SVG or aria-label in Canvas renderer.
```
Malvoz, line 60:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Polyline.js
```diff
@@ -57,7 +57,12 @@ export var Polyline = Path.extend({
 
 		// @option noClip: Boolean = false
 		// Disable polyline clipping.
-		noClip: false
+		noClip: false,
+
+		// @option title: String= 'Polyline'
+		// Alternative text for a Figure. Used as <title> in SVG or aria-label in Canvas renderer.
```
Malvoz, line 63:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Rectangle.js
```diff
@@ -29,6 +29,13 @@ import {toLatLngBounds} from '../../geo/LatLngBounds';
 
 
 export var Rectangle = Polygon.extend({
+	options: {
+		// @option title: String= 'Rectangle'
+		// Alternative text for a Figure. Used as <title> in SVG or aria-label in Canvas renderer.
```
Malvoz, line 34:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Circle.js
```diff
@@ -24,6 +24,13 @@ import {Earth} from '../../geo/crs/CRS.Earth';
 
 export var Circle = CircleMarker.extend({
 
+	options: {
+		// @option title: String= 'Circle'
+		// Alternative text for a Figure. Used as <title> in SVG or aria-label in Canvas renderer.
```
Malvoz, line 29:
```suggestion
		// Alternative text for a Figure. Used as `<title>` in SVG or `aria-label` in Canvas renderer.
```

src/layer/vector/Path.js
Malvoz: Yes, let's go ahead and rename it `desc`. 👍🏼

src/layer/vector/Canvas.js
Malvoz: Other options include:

- `aria-describedby` instead of `aria-label` (and use for the `desc` option) since `aria-description` isn't fully supported
- only implement `title` for now

src/layer/vector/Canvas.js
alekzvik: Canvas title support is hardly possible at the moment. The unrelated example you mentioned uses a technique similar to what Leaflet does under the hood to show `Tooltip`s on canvas - calculate what related "figures" are under the mouse current position. I do not think implementing this for a native tooltip is necessary as we already have a `Tooltip` to achieve that.

src/layer/vector/Canvas.js
Malvoz: > Do you think we should add another element to this `<div>`?

Possibly. Or sibling, but I don't think that should be necessary. The following test HTML works (tested in NVDA, please test in VO):

```html
<canvas>
  <div tabindex="0" role="graphics-symbol img" title="title" aria-describedby="leaflet-canvas-interactive-desc">
    <span id="leaflet-canvas-interactive-desc">desc</span>
  </div>
</canvas>
```

In this case `aria-describedby` is used only because the roles do not support [name from content](https://www.w3.org/TR/wai-aria-1.1/#namefromcontent).

This would result in better consistency in the names, but also allow us to change `aria-describedby` to `aria-description` (to have only 1 element) in the future.

debug/vector/simple-canvas-svg.html
```diff
@@ -0,0 +1,52 @@
...
+				.addTo(map);
+
+			var circle = L.circle([51.508, -0.11], 500, {color: '#f03', opacity: 0.7, title: 'Wow, a Circle', description: 'A regular red circle with a popup'})
+				.bindPopup("I am a circle.")
+				.addTo(map);
+
+			var polygon = L.polygon([
+				[51.509, -0.08],
+				[51.503, -0.06],
+				[51.51,  -0.047]], {title: 'a polygon', description: 'blue polygon, also has a popup'})
```
Malvoz, line 39:
```suggestion
			var circle = L.circle([51.508, -0.11], 500, {color: '#f03', opacity: 0.7, title: 'Wow, a Circle', desc: 'A regular red circle with a popup'})
				.bindPopup("I am a circle.")
				.addTo(map);

			var polygon = L.polygon([
				[51.509, -0.08],
				[51.503, -0.06],
				[51.51,  -0.047]], {title: 'a polygon', desc: 'blue polygon, also has a popup'})
```

src/layer/vector/Canvas.js
alekzvik: I like this approach and have implemented it. 
Tested it with VoiceOver on Safari, Firefox and Chrome. 
It works consistently on all browsers and narrates SVG and canvas versions in the same way.

Malvoz: Awesome, great work! 🙂 Canvas sub DOM is pretty cool.

src/layer/vector/Circle.js
```diff
@@ -24,6 +24,13 @@ import {Earth} from '../../geo/crs/CRS.Earth';
 
 export var Circle = CircleMarker.extend({
 
+	options: {
+		// @option title: String= 'Circle'
```
Malvoz, line 28: We should consider not using default titles.

Relatedly, I've been thinking of partially reverting https://github.com/Leaflet/Leaflet/pull/7895 to remove the default `alt="Marker"`, for several reasons (passing accessibility audits by default doesn't encourage unique titles, localization, inconsistent names depending on if `title` or `alt`/`desc` was used). But we could open a new issue to discuss both.

alekzvik: I added defaults primarily for consistency with the Marker. I can remove defaults here, and we can open a separate issue to discuss `alt` for the Marker. And depending on the discussion there, we can either remove the default for Marker or add for the vector types.

alekzvik: Eight minutes passed, and my opinion has changed 🙃 
Let's leave defaults as is and discuss `alt` for Marker and default `title`s for vectors in a separate issue. I can see the benefits of both providing defaults and not. I also think that consistency is important here. So discussing them together is a good way forward.

Malvoz: Yep, we can always change this later. Opened https://github.com/Leaflet/Leaflet/issues/8267 as a reminder.

Malvoz (approved):
I think the options' descriptions could be improved a bit (to help distinguish use cases for `title` vs `desc`, mention that SVG `title` produces native tooltips, same as [here](https://leafletjs.com/reference.html#marker-title)), but generally this looks really good!

<details>
<summary>Things that can be discussed/addressed separately</summary><br>

> **Note**
> While there may be some imperfections (pointed out below), _this PR does enable vector accessibility_(!).

- [Canvas's `title` option doesn't result in standard native tooltips](https://github.com/Leaflet/Leaflet/pull/8251#discussion_r878709448).
- Strange behavior with popups opened from vectors **when using a screen reader**

  <details><br>
   
   The behavior surfaces here (test in `debug/vector/simple-canvas-svg.html`) but can be reproduced in v1.7.1. For example in https://maps4html.org/experiments/linking/features/:

  1. Remove `role="application"` from the `<mapml-viewer>` element
  2. Start a screen reader (I can reproduce in both NVDA and Narrator)
  3. <kbd>Tab</kbd> 4 times until the "Saskatchewan" vector has focus
  4. Press <kbd>Enter</kbd> and observe **the popup is offset from the vector**
  <br>

  Note that while [`role="application"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/application_role) mitigates this issue in some screen readers, the role has other implications so not an ideal "solution" (`mapml-viewer` uses it for other purposes).

  </details>

- Potential for [`autoPanOnFocus`](https://leafletjs.com/reference.html#marker-autopanonfocus) for vectors
- This will increase the amount of tab stops in some maps (solution on mitigating this has been suggested in https://github.com/Leaflet/Leaflet/issues/7479#arrow-key-navigation)
- https://github.com/Leaflet/Leaflet/issues/8267
- [Incorporate advice on labelling vectors in the accessibility guidelines](https://github.com/Leaflet/Leaflet/discussions/8006#discussioncomment-2190734) (and link to it from the "Useful for accessibility" docs descriptions).

</details>

Malvoz:
For reviewers to test this out locally:

1. Go to `debug/vector/simple-canvas-svg.html`
2. [Start a screen reader](https://leafletjs.com/examples/accessibility/#test-your-maps) (if things get too noisy use <kbd>ctrl</kbd> to temporarily pause speech)
3. <kbd>tab</kbd> to vectors and press <kbd>enter</kbd>

Keep in mind https://github.com/Leaflet/Leaflet/pull/8251#pullrequestreview-985965212:

> While there may be some imperfections (pointed out below), this PR does enable vector accessibility!

nfreear:
Hi @alekzvik,

> Vectors rendered in SVG have a tabindex to be focusable by keyboard

Are these vectors always supposed to be interactive, for example, to display a popup or tooltip, or as a form field to adjust a value?

`tabindex="0"` is ONLY for interactive elements that need to be focusable using Tab and Shift + Tab, for example, buttons, links, form fields. (`tabindex` is often mis-understood and mis-used on the wrong elements.)

If the Vector needs to be available to assistive technology like screen readers, maybe it should be identified as an image?

Here is a basic example:

```
<svg aria-labelledby="myLabelId">
  <title>A descriptive name</title>
  ...
</svg>
```

Does that help?

Nick

src/layer/vector/CircleMarker.js
```diff
@@ -21,7 +21,12 @@ export var CircleMarker = Path.extend({
 
 		// @option radius: Number = 10
 		// Radius of the circle marker, in pixels
-		radius: 10
+		radius: 10,
+
+		// @option title: String= 'CircleMarker'
+		// Alternative text for a Figure. Used as `<title>` tag in SVG or `title` attribute in Canvas renderer.
+		// Useful for accessibility.
+		title: 'CircleMarker',
```
mjumbewu, line 29: **nit**: What about making this default to `'Circle marker'` instead? It seems a more appropriate thing to expose to a human interface. Since most of the default `title` and `alt` values in Leaflet are a single word, there aren't many benchmarks that I could point to, except maybe the default titles for [zoom in](https://leafletjs.com/reference.html#control-zoom-zoomintitle) and [zoom out](https://leafletjs.com/reference.html#control-zoom-zoomouttitle) buttons.
```suggestion
		// @option title: String= 'Circle marker'
		// Alternative text for a Figure. Used as `<title>` tag in SVG or `title` attribute in Canvas renderer.
		// Useful for accessibility.
		title: 'Circle marker',
```

spec/suites/layer/vector/PathSpec.js
```diff
@@ -106,4 +106,39 @@ describe('Path', function () {
...
 	});
+
+	describe('#accessibility', function () {
+		it('should have title', function () {
+			var path = L.circleMarker([1, 2]);
+			path.addTo(map);
+
+			var element = path.getElement();
+
+			expect(element.querySelector('title').innerHTML).to.eql('CircleMarker');
```
mjumbewu, line 117: This would also need to change if the default changes.
```suggestion
			expect(element.querySelector('title').innerHTML).to.eql('Circle marker');
```

I love this! Hope it gets merged.

src/layer/vector/CircleMarker.js
nfreear: :+1: - Plus one to `'Circle marker'` as the default. Spaces and plain English words are good for assistive technology!

Thank you all for your contributions to this ❤️ 

(And, take my comment above with a pinch of salt - I'm late to the party!)

Nick

Malvoz:
> > Vectors rendered in SVG have a tabindex to be focusable by keyboard
> 
> Are these vectors always supposed to be interactive, for example, to display a popup or tooltip, or as a form field to adjust a value?
> 
> `tabindex="0"` is ONLY for interactive elements that need to be focusable

(Thanks for your feedback @nfreear, across multiple issues and PRs, highly appreciated!)

No these vectors are not always supposed to be interactive. The distinction between interactive and non-interactive vectors (similar to for markers: https://github.com/Leaflet/Leaflet/issues/8116) could be addressed separately, IMO.

PR #8256, opened 2022-05-22
Replace "var map" by "let map"

KristjanESPERANTO:
That may be a harsh move without prior discussion. But I give it a try.

### DIFF
--- a/debug/hacks/jitter.html
+++ b/debug/hacks/jitter.html
@@ -31,7 +31,7 @@
 		//Disable the hack fix
 		L.Browser.chrome = true;
 
-		var map = L.map('map')
+		let map = L.map('map')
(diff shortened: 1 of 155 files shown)


jonkoops:
We are currently in the process of stabilizing Leaflet 1.8 and eventually working our way up to version 2.0 which will drop support for older browsers.

Most likely our minimum support target for version 2.0 will be ES2015, at which point we'd  want to refactor all of our code (including documentation).

I am closing this PR now as it does not fit into the current timeline and will likely end up out of sync over time. When the time comes that we move to a newer syntax in general help would be appreciated.

KristjanESPERANTO:
Sounds like a good way to proceed 🙂 Thanks!

Issue #6218, opened 2018-07-05
Horizontal lines in popup when scrolling is activated


cccruzr:
## How to reproduce

- Leaflet version I'm using: 1.3.1
- Browser (with version) I'm using: Firefox 61 / Chrome 64
- OS/Platform (with version) I'm using: Windows 10

## What behaviour I'm expecting and which behaviour I'm seeing
I'm not really sure if this is a bug of Leaflet CSS, that's why I'm asking...

Since I'm trying to code "mobile-first", I made my popup scrollable with:

`layer.bindPopup(popupLayout, {
                    maxWidth: 300,
                    maxHeight: 400,
                    closeOnClick: true
});`

However when the content is large, I'm getting the following horizontal lines:

[![enter image description here][1]][1]

Compared with a normal non-scrollable popup:

[![enter image description here][2]][2]

Any idea on how to get rid of those ugly lines?

  [1]: https://i.stack.imgur.com/P0tGw.jpg
  [2]: https://i.stack.imgur.com/hy1yg.jpg

IvanSanchez:
Can you publish a fiddle/codepen/[plunkr](https://leafletjs.com/edit) that displays this behaviour?

cccruzr:
This is the [webmap](https://cccruzr.github.io/LideresSociales/)
Here you can find the [code of the map](https://github.com/cccruzr/LideresSociales/blob/master/docs/js/d3_map.js#L73)

Thanks for your help!

Malvoz:
The lines you're seeing are borders:

https://github.com/Leaflet/Leaflet/blob/2a60e29ad6f5b11b3ce6be90c34c8f1910a88804/dist/leaflet.css#L538-L542

We should consider removing them.

Malvoz:
> Any idea on how to get rid of those ugly lines?

@cccruzr add the following to your CSS:

```css
.leaflet-popup-scrolled {
  border: 0;
}
```

PR #8260, opened 2022-05-23
Remove border styles in overflowing popups

Malvoz:
Fix https://github.com/Leaflet/Leaflet/issues/6218.


Malvoz:
IMO we should go ahead with this.

The border styles can be a bit unexpected, only used if `maxHeight`  is set. And `lealet-popup-scrolled` can still be styled with custom CSS.

Malvoz:
> And `leaflet-popup-scrolled` can still be styled with custom CSS

I guess this should be documented.

Falke-Design (approved):

PR #8286, opened 2022-06-07
Use major version ranges for Github Actions

jonkoops:
Don't want to deal with constant updates of minor versions from Dependabot for Github Actions (such as #8284). This resolves the issue by using a major version range so the latest is always used.


Falke-Design (approved):
If this works (what i think that it does) it is ok for me

Issue #8287, opened 2022-06-07
`_isClickDisabled` throws error if parent is removed from DOM


Falke-Design:
Since I upgraded leaflet to version 1.8.0, two components in my app have stopped working. They are a [react-search-box](https://www.npmjs.com/package/react-search-box) and a [react-select](https://react-select.com/home) that I placed inside my own custom control. I have applied `DomEvent.disableClickPropagation()` to the control container, like so:
`L.DomEvent.disableClickPropagation(container);`
and it used to work without any problem before the upgrade. Since the upgrade, whenever I select one of the options available from the search (react-search-box) or the select (react-select) dropdown menu, I get this error:

![image](https://user-images.githubusercontent.com/73307948/172419018-09915981-18f6-455b-bf74-75f869c4c2e1.png)

![image (1)](https://user-images.githubusercontent.com/73307948/172419199-4a5ed719-ddbd-42e7-b310-8230b481dee1.png)

@mourner and @Falke-Design do you guys have any idea what I could do to fix this? Or is there anything that you guys can do on your side? Thanks for your precious time!

_Originally posted by @AngelaCandela in https://github.com/Leaflet/Leaflet/issues/7439#issuecomment-1148829666_

Falke-Design:
@Falke-Design I set up [this demo](https://codesandbox.io/s/infallible-wood-zl59xo?file=/src/components/Map/components/CustomControl.tsx) in codesandbox, I hope it helps! Once again, thank you so much for your help!

_Originally posted by @AngelaCandela in https://github.com/Leaflet/Leaflet/issues/7439#issuecomment-1148999564_

PR #8288, opened 2022-06-07
Fix: `_isClickDisabled` throws error if parent is removed from DOM 

Falke-Design:
Fix: #8287

@AngelaCandela you can test this by adding following before you create your map (`<MapContainer` ...):
```
L.Map.include({
  _isClickDisabled: function (el: any) {
    while (el !== this._container && el) {
      if (el["_leaflet_disable_click"]) {
        return true;
      }
      el = el.parentNode;
    }
  }
});
```

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1376,7 +1376,7 @@ describe("Map", function () {
 			parent.appendChild(child);
 			container.appendChild(parent);
 			L.DomEvent.on(child, 'click', function () {
-				parent.remove();
+				L.DomUtil.remove(parent);
 			});
 			expect(function () {
 				happen.once(child, {type: 'click'});


spec/suites/map/MapSpec.js
```diff
@@ -1367,4 +1367,20 @@ describe("Map", function () {
...
+
+	describe('#disableClickPropagation', function () {
+		it('does not break if element is not in the DOM anymore', function () {
+			map.setView([0, 0], 0);
+			var parent = document.createElement('div');
+			var child = document.createElement('div');
+			parent.appendChild(child);
+			container.appendChild(parent);
+			L.DomEvent.on(child, 'click', function () {
+				parent.remove();
```
mourner, line 1379: Needs `DomUtil.remove` to fix CI tests in IE.

mourner (changes requested):
👍 apart from IE tests

spec/suites/map/MapSpec.js
Falke-Design, line 1379:
```suggestion
				L.DomUtil.remove(parent);
```

mourner (approved):

AngelaCandela:
@Falke-Design and @mourner thank you so so much! That indeed fixed it!

Issue #8228, opened 2022-05-11
Site SEO


Malvoz:
## Problems

(While the `/SlavaUkraini/` URL change may have (temporarily) affected search engine indexing, that's not my concern with this issue.)

I've found that (before, during, and after the URL change), in some cases, when explicitly searching for a title/heading of a Leaflet page:

- search engines aren't showing optimal/relevant results
- other sites get priority in the search results

Of course, the search results will differ depending on the search engine and keywords used to search. However I think there are some things that can be improved, described below.

### [Avoid repeated or boilerplate text in `<title>` elements](https://developers.google.com/search/docs/advanced/appearance/title-link#:~:text=Avoid%20repeated%20or%20boilerplate%20text%20in%20%3Ctitle%3E%20elements)

  - All pages have "Leaflet - a JavaScript library for interactive maps" in their title:

    https://github.com/Leaflet/Leaflet/blob/c46b6f044e6e6b00e8a97796700b9dd9ba5ece75/docs/_layouts/v2.html#L6

    - [ ] Remove "Leaflet - a JavaScript library for interactive maps" or at least "- a JavaScript library for interactive maps"?
           (And set it as title for [docs/index.html](https://github.com/Leaflet/Leaflet/blob/main/docs/index.html) only. The "Leaflet" part could be repeated, see [Brand your titles concisely](https://developers.google.com/search/docs/advanced/appearance/title-link#:~:text=Brand%20your%20titles%20concisely))

  - Tutorials and their examples have identical or too similar titles:

    https://github.com/Leaflet/Leaflet/blob/c46b6f044e6e6b00e8a97796700b9dd9ba5ece75/docs/examples/quick-start/index.md?plain=1#L3

    https://github.com/Leaflet/Leaflet/blob/c46b6f044e6e6b00e8a97796700b9dd9ba5ece75/docs/examples/quick-start/example-basic.md?plain=1#L3

    https://github.com/Leaflet/Leaflet/blob/c46b6f044e6e6b00e8a97796700b9dd9ba5ece75/docs/examples/quick-start/example-overlays.md?plain=1#L3

    - [ ] Rewrite titles for example pages?
[...]

PR #8294, opened 2022-06-18
Deprecate old versions' docs

Malvoz:
Partial fix for both https://github.com/Leaflet/Leaflet/issues/7712, and https://github.com/Leaflet/Leaflet/issues/8228:

> - All reference versions use the same title
>
>   - [ ] Add unique titles for the old versions (e.g. "Documentation - v0.7.7" etc.) or remove old docs per https://github.com/Leaflet/Leaflet/issues/7712?


Malvoz:
Obviously this change will make it a bit harder for people to find the docs for older Leaflet versions.

RELEASE.md
```diff
@@ -14,7 +14,6 @@
 
 - [ ] Make a new branch for the update
 - [ ] Write a blog post about the new release and put it in `/docs/_posts`
-- [ ] If necessary to preserve previous version's docs, rename `dist/reference.html` to `dist/reference-X.Y.Z.html` and add it to the list in `docs/reference-versions.html`
```
Falke-Design, line 17: There should be a new step to add a link from the web archiv

Falke-Design (dismissed):

RELEASE.md
Malvoz: I'm not so sure, it depends on how important you consider old versions to be. One of the reasons for deprecating the old docs was to simplify the release process.

Let's hear what others think.

mourner: @Malvoz the second quote sounds good to me! Let's get this in.

mourner:
@Malvoz let's rebase and get this merged!

mourner (approved):

PR #8298, opened 2022-06-18
Fix 2 step mouse wheel scrolling

Falke-Design:
We compared different Browsers (https://github.com/Leaflet/Leaflet/issues/7403#issuecomment-1101669097) and it shows us that all other Browsers than OSX(Mac), Safari, and Chrome on Linux have a double deltaX value. So we need for this Browsers `2 * devicePixelRatio`.

Fix: #7403, Close #7477.

We need to check if every Browser works fine:

- [x] **Win 10**
  - [x] Firefox (@Falke-Design, @KommX, @jinchun-wang)
  - [x] Chrome (@Falke-Design, @jinchun-wang)
  - [x] Edge (@Falke-Design, @jinchun-wang)
- [x] **Win 11**
  - [x] Firefox (@jinchun-wang)
  - [x] Chrome (@jinchun-wang)
  - [x] Edge (@jinchun-wang)
- [x] **Linux**
  - [x] Firefox (@zdila, @paulatthehug)
  - [x] Chrome (@paulatthehug)
  - [x] Edge (@IvanSanchez)
- [x] **OSX**
  - [x] Firefox (@jeffreykog)
  - [x] Chrome (@jeffreykog)
  - [x] Edge (@jeffreykog)
  - [x] Safari (@paulatthehug, @jeffreykog)

Demo: 
v1: https://plnkr.co/edit/gqQUNU9mf6RiqxD4
v2 (incl. macOS): https://plnkr.co/edit/vtEzxb7P0EFCkHoq

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -169,6 +169,13 @@ var vml = !svg && (function () {
 	}
 }());
 
+
+// @property mac: Boolean; `true` when the browser is running in a Mac platform
+var mac = navigator.platform.indexOf('Mac') === 0;
+
+// @property mac: Boolean; `true` when the browser is running in a Linux platform
+var linux = navigator.platform.indexOf('Linux') === 0;
+
 function userAgentContains(str) {
 	return navigator.userAgent.toLowerCase().indexOf(str) >= 0;
 }
@@ -207,5 +214,7 @@ export default {
 	canvas: canvas,
 	svg: svg,
 	vml: vml,
-	inlineSvg: inlineSvg
+	inlineSvg: inlineSvg,
+	mac: mac,
+	linux: linux
 };
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -263,8 +263,9 @@ export function getMousePosition(e, container) {
 	);
 }
 
-// OSX (Mac), Safari and Chrome on Linux scrolls double the pixels as in other platforms (see #7403 and #4538),
-// for all other Browsers we need double the scroll pixels by our self
+
+// For all Browsers except OSX (Mac), Safari and Chrome on Linux
+// we need double the scroll pixels (see #7403 and #4538)
 var wheelPxFactor =
 	(Browser.mac || Browser.safari || (Browser.linux && Browser.chrome)) ? window.devicePixelRatio :
 	window.devicePixelRatio > 0 ? 2 * window.devicePixelRatio : 1;
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -264,12 +264,14 @@ export function getMousePosition(e, container) {
 }
 
 
-// For all Browsers except OSX (Mac), Safari and Chrome on Linux
-// we need double the scroll pixels (see #7403 and #4538)
-var wheelPxFactor =
-	(Browser.mac || Browser.safari || (Browser.linux && Browser.chrome)) ? window.devicePixelRatio :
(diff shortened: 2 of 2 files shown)


jinchun-wang:
Hello,

I've tested 
Windows 10 version 21H2 (OS Build 19044.1766)
Windows 11 version 21H2 (OS Build 22000.739)
Windows Server 2022 version 21H2 (OS Build 22348.768)
Firefox 101.0.1 (64-bit)
Chrome 102.0.5005.115 (64 bits)

The absolut delta is 1 but the direction of zoom in / out seems inverse.
Please have a check again.

jinchun-wang:
The same result of my test on Edge on Win10, 11, server 2022, the absolute delta is 1, the direction is inverse.

jinchun-wang:
Sorry you did 'zoomBefore - map.getZoom()' so it is ok now for all. Excellent!!

Falke-Design:
It looks like there is a problem on Firefox on Mac, I need to investigate more

paulatthehug:
@Falke-Design wrote:

> It looks like there is a problem on Firefox on Mac, I need to investigate more

Yup, I just tried it on Firefox on my Mac VM and it fails. Badly. Doesn't even render a map. Let me know when you think you've for a fix and I'll try it again.

Falke-Design:
> Why did not 'Map options - zoomDelta' apply to the change of the mouse wheel? And however it's working when pressing + or - on the keyboard

@jinchun-wang [zoomDelta](https://leafletjs.com/reference.html#map-zoomdelta) is not applied to the mousewheel zooming.
> Controls how much the map's zoom level will change after a [zoomIn()](https://leafletjs.com/reference.html#map-zoomin), [zoomOut()](https://leafletjs.com/reference.html#map-zoomout), pressing + or - on the keyboard, or using the [zoom controls](https://leafletjs.com/reference.html#control-zoom). Values smaller than 1 (e.g. 0.5) allow for greater granularity.

If you want to change the zoomDelta while scrolling you need to play around with [wheelPxPerZoomLevel](https://leafletjs.com/reference.html#map-wheelpxperzoomlevel)

jinchun-wang:
But wheelPxPerZoomLevel depends on device, os and browser.

jinchun-wang:
Is it possible to add a new option 'maxZoomDelta' to apply on mouse wheel action? By default it can be no limit. In our cases we don't need fast zoom in/zoom out by mouse wheel.

jeffreykog:
Hi!

I tested this on OSX (M1 Max), on the built-in high-dpi monitor, an external high-dpi monitor, and an external 1080p monitor.

- Chrome 103.0.5060.134: Working
- Firefox 103.0.1: Working
- Safari 15.6 (17613.3.9.1.5): Working
- Edge 103.0.1264.77: Working

Zoom delta used in every browser is 1, as it should. So this is a go for OSX.

Falke-Design:
@jeffreykog very nice thank you!

Falke-Design:
@IvanSanchez do you have Edge on Linux? I know the chance is very low 😄

IvanSanchez:
> @IvanSanchez do you have Edge on Linux?

I do, but I feel dirty about it.

Zoom delta is +/-1. I do not have a high-dpi monitor for this, but I tested both at 100% and 200% in-browser scaling (the `Ctrl`+`+` thing). I'm checking off the linux+edge box.

IvanSanchez (approved):
Works In My Machine™

Issue #8300, opened 2022-06-20
GridLayer flickers when animating fractional zoom changes


sbquinlan:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

1. Setup a map that allows fractional zoom with a layer that has a maxZoom (see Plunker link for complete code)

```js
var CanvasLayer = L.GridLayer.extend({
	createTile: function(coords){
		var tile = L.DomUtil.create('canvas', 'leaflet-tile');

		var size = this.getTileSize();
		tile.width = size.x;
		tile.height = size.y;

		var ctx = tile.getContext('2d');
		ctx.fillStyle = 'green';
		ctx.fillRect(0, 0, size.x, size.y);

		return tile;
	}
});
var map = new L.Map('leaflet', {
	layers: [
		new L.TileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
			'attribution': 'Map data © <a href="https://openstreetmap.org">OpenStreetMap</a> contributors'
		}),
		new CanvasLayer({ maxZoom: 3 })
	],
	center: [0, 0],
	zoom: 3,
	zoomDelta: 0.1,
	zoomSnap: 0.1,
});
```

2. Use scroll wheel to zoom out to 3.1 (one scroll wheel 'tick')

### Expected behavior

Because maxZoom on the SomeGridLayer is 3, I expect SomeGridLayer to disappear or fade out.

### Current behavior

SomeGridLayer is removed, added back, and then removed again. 

### Minimal example reproducing the issue

https://plnkr.co/edit/TubUklLYOo7RLzDJ

### Environment

- Leaflet version: 1.8.0, master
- Browser (with version): Chrome 102.0.5005.115
- OS/Platform (with version): OSX 12.4

sbquinlan:
The start of the animation calls setView with noPrune set to true:

[...]

PR #8311, opened 2022-06-26
Fix fractional zoom flicker when out of zoom range

sbquinlan:
Fixes: #8300

Will add test and comments if maintainers support this direction. Another option is to change pruneTiles so that it uses this._tileZoom instead of pulling the zoom off the map directly. This would leave the layer rendering until Math.round(this._map.getZoom()) > maxZoom` or in the example in the attached issue 3.5.


ranxu815:
Zoom in the map on the mobile phone and put it at the maximum map level. The map layer moves with your fingers

PR #8329, opened 2022-07-06
Expose ESM entrypoint with Leaflet global

### LINKED ISSUE #7055: Add entrypoint for ES modules to package

jonkoops:
Introduces a `module` entrypoint to the package to allow some tooling to load the ECMAScript module variant of Leaflet. The goal of this PR is to introduce this as a non-breaking change, preserving the Leaflet global `L` to avoid breaking existing code.

This PR is limited in scope an purposefully and does not try to implement the following:

**[Tree-shaking](https://webpack.js.org/guides/tree-shaking/)**
This requires us to further decouple internals of Leaflet, ensuring that objects are not grouped together (e.g. `Control.Zoom`). Fully supporting this will also mean that we have to mark the [side effects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) of the package.

**[Conditional exports](https://nodejs.org/api/packages.html#conditional-exports)**
This is a new feature in Node.js that allows us to specify specific ESM and CommonJS entrypoints. It also prevents internals of the package from being exposed unintentionally since the exported API surface is defined explicitly.

This feature is not widely supported by tooling at this time, but will be preferred in the future as Node.js has never officialy supported the `module` field (unlike many bundlers). Since we currently do not offer running Leaflet under Node.js this is not a priority. 

As more tooling (such as bundlers) move to this standard we will likely end up implementing this.

Closes #7055

### DIFF
--- a/src/LeafletWithGlobals.js
+++ b/src/LeafletWithGlobals.js
@@ -1,0 +1,24 @@
+import * as L from './Leaflet';
+export * from './Leaflet';
+
+var globalL = L.extend(L, {noConflict: noConflict});
+export default globalL;
+
+var globalObject = getGlobalObject();
+var oldL = globalObject.L;
+
+globalObject.L = globalL;
+
+export function noConflict() {
+	globalObject.L = oldL;
+	return globalL;
+}
+
+function getGlobalObject() {
+	if (typeof globalThis !== 'undefined') { return globalThis; }
+	if (typeof self !== 'undefined') { return self; }
+	if (typeof window !== 'undefined') { return window; }
+	if (typeof global !== 'undefined') { return global; }
+
+	throw new Error('Unable to locate global object.');
+}
--- a/build/rollup-config.js
+++ b/build/rollup-config.js
@@ -12,24 +12,15 @@ const watch = process.argv.indexOf('-w') > -1 || process.argv.indexOf('--watch')
 const version = release ? pkg.version : `${pkg.version}+${gitRev.branch()}.${gitRev.short()}`;
 const banner = createBanner(version);
 
-const outro = `var oldL = window.L;
-exports.noConflict = function() {
-	window.L = oldL;
-	return this;
-}
-// Always export us to window global (see #2364)
-window.L = exports;`;
-
 /** @type {import('rollup').RollupOptions} */
 const config = {
-	input: 'src/Leaflet.js',
+	input: 'src/LeafletWithGlobals.js',
 	output: [
 		{
 			file: pkg.main,
 			format: 'umd',
 			name: 'leaflet',
 			banner: banner,
-			outro: outro,
 			sourcemap: true,
 			freeze: false,
 			esModule: false
@@ -43,7 +34,7 @@ const config = {
 if (!watch) {
 	config.output.push(
 		{
-			file: 'dist/leaflet-src.esm.js',
+			file: pkg.module,
 			format: 'es',
 			banner: banner,
 			sourcemap: true,
--- a/debug/tests/esm.html
+++ b/debug/tests/esm.html
@@ -1,0 +1,19 @@
+<html>
+  <head>
+    <link rel="stylesheet" href="../../dist/leaflet.css" />
+    <link rel="stylesheet" href="../css/screen.css" />
+  </head>
+  <body>
+    <div id="map"></div>
+    <script type="module">
+      import L from '../../dist/leaflet-src.esm.js';
+
+      var map = L.map('map', {
+        center: [39.84, -96.591],
+        zoom: 4,
+      });
+
+      L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(map);
+    </script>
+  </body>
+</html>
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "uglify-js": "^3.15.4"
   },
   "main": "dist/leaflet-src.js",
+  "module": "dist/leaflet-src.esm.js",
   "style": "dist/leaflet.css",
   "files": [
     "dist",
@@ -69,7 +70,8 @@
     ],
     "root": true,
     "globals": {
-      "L": true
+      "L": true,
+      "globalThis": true
     },
     "env": {
       "commonjs": true,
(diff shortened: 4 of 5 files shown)


jonkoops:
I've added a debug HTML that loads the ESM version directly into the browser without the need to add a bundler. And it seems that this is working as expected!

Screenshot showing Leaflet working in ESM mode

src/LeafletWithGlobals.js
```diff
@@ -0,0 +1,22 @@
...
+
+globalObject.L = globalL;
+
+export function noConflict() {
+	globalObject.L = oldL;
+	return globalL;
+}
+
+function getGlobalObject() {
+	if (typeof globalThis !== 'undefined') { return globalThis; }
```
IvanSanchez, line 16: What's the purpose of `globalThis`? Framework compatibility?

jonkoops: The [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) provides a standard way of accessing the global `this` value (and hence the global object itself) across environments.

Falke-Design:
Is it possible to add a demo with the esm usage? And maybe a test?

src/LeafletWithGlobals.js
IvanSanchez: I see, it was added to browsers fairly recently and wasn't aware of it.

Falke-Design:
It don't work with the new leaflet source for:
```
const L = require('../libs/leaflet');
//const L = require('leaflet'); works
require('@geoman-io/leaflet-geoman-free');
console.log(L.PM)
```` 

Maybe it is a error from me but I don't have time to check this in the next 24h hours, so maybe it helps to find a problem

File: src/LeafletWithGlobals.js (hunk 1 above)
Comment (jonkoops) [reply src/LeafletWithGlobals.js:16]: Yes it does, as `globalThis` refers to the global in Node.js as well. But it wont prevent errors that occur if certain APIs don't exist (such as the DOM).

**Comment by jonkoops** [comment]
@Falke-Design it's hard to say without knowing where `../libs/leaflet` resolves to.

**Comment by jonkoops** [comment]
@Falke-Design do you have a reproducible example somewhere I can use?

**Comment by Falke-Design** [comment]
I will create one tomorrow

**Comment by jonkoops** [comment]
So I have been looking around to see if we can test this functionality but I think that is tad hard to accomplish. The reason for this is that we are [using a Karma plugin](https://github.com/Leaflet/Leaflet/blob/be871e6f706bc0567a211f873225077f4cf05b99/spec/karma.conf.js#L38) to compile the source with Rollup. The configuration it uses is separate from all the existing Rollup configuration for the build.

This also means that we are not really end-to-end testing our package from the perceptive of the consumer. This is also true for the UMD module we are distributing now.

**Comment by Falke-Design** [comment]
@jonkoops here is the example, it is builded with webpack: https://github.com/Falke-Design/leaflet-esm-test

**Comment by jonkoops** [comment]
I've added some changes so that a default export is present which is the exact same as the global `L`. This will make it less likely that folks will run into any issues when upgrading. This allows the following:

```js
import * as L from 'leaflet';
```
To be turned into:
```js
import L from 'leafet';
```
And still allows named imports to co-exist:
```js
import L, { Map } from 'leafet';
```
It also ensures that the global `L` and the default imported `L` are equivalent, making it less error prone for plugins:
```js
import L from 'leaflet';
import '@geoman-io/leaflet-geoman-free';

L === window.L; // true
L.PM === window.L.PM; // true
```

jonkoops:
@Falke-Design I've done some testing on your example code and managed to get things working properly with the latest changes on this PR. I've opened up a PR on your repo where you can see the code I used to test (https://github.com/Falke-Design/leaflet-esm-test/pull/1).

Make sure to build Leaflet locally on this PR and then [`npm link`](https://docs.npmjs.com/cli/v8/commands/npm-link) it into your example code and it should function as expected.

jonkoops:
@Falke-Design @IvanSanchez @mourner can I ask y'all to take another look at this PR? As far as I am concerned we can go ahead with this and start rolling out betas for 1.9 (are we tracking this anywhere?).

Falke-Design:
@jonkoops thank you. I will take a look at it in the evening.

Falke-Design (approved):
Everything looks fine for me. This will be a nice improvement! Thank you @jonkoops

mourner (approved):

Issue #8145, opened 2022-04-16
Missing Map tests


Falke-Design:
[Map](https://leafletjs.com/SlavaUkraini/reference.html#map-setview):
- [x] setZoom
- [x] setZoomAround
- [x] fitWorld
- [x] #8365
- [ ] #8366
- [ ] #8367
- [x] setMinZoom
- [x] setMaxZoom
- [ ] #8368
- [ ] #8369
- [ ] #8370
- [ ] #8371
- [ ] #8372
- [x] createPane
- [x] getPane
- [x] getPanes
- [x] getContainer
- [x] getSize
- [x] getPixelBounds
- [x] getPixelOrigin
- [x] getPixelWorldBounds
- [ ] #8373
- [ ] #8374
- [ ] #8375
- [x] #8376
- [x] distance
- [x] containerPointToLayerPoint
- [x] layerPointToContainerPoint
- [x] #8377
- [x] #8378
- [ ] #8379
- [ ] #8380
- [ ] #8381

stephenspol:
Can you assign this to me @Falke-Design?

Malvoz:
@stephenspol Done! 👍🏼

NeverGiveUp23:
Hello, can you assign me a task. doesn't matter which one.

Falke-Design:
@NeverGiveUp23 maybe the easiest ones are:
 - containerPointToLatLng
 - latLngToContainerPoint
 - mouseEventToContainerPoint
 - mouseEventToLayerPoint
 - mouseEventToLatLng

abhi3315:
It seems no one is assigned to this issue. Can I pick up this issue?

Malvoz:
@abhi3315 Absolutely!

jonkoops:
I've created some more granular issues so that we can track the changes better, feel free to sign up for any of them.

Malvoz:
Closing this per the above comment / in favor of the more granular issues.

jonkoops:
Let's keep this open as not all tasks inside of this tracking issue have been resolved.

Malvoz:
Ok.

adeeb18:
Hello! I am a first time contributor, are there any tasks I can be assigned and do you have any insight on how to best tackle them? Thank you!

jonkoops:
@adeeb18 I'd recommend to take a look at the PRs for the issues that have been closed. They will give you a good impression on how to get started.

jonkoops:
[...]

PR #8343, opened 2022-07-18
add report asa

albenahm:
Automatic Static Analysis report -> docs/asa/Leaflet ASA.pdf
( #8145 ) test for all project


mourner:
Nah, thanks

Issue #8349, opened 2022-07-20
Be able to use main branch with NPM


davetapley:
### Checklist

- [X] I've searched through the [plugins](https://leafletjs.com/plugins.html) to make sure this feature isn't already available, or think it shouldn't require a plugin.
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this feature hasn't been requested already.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Motivation

There is no easy way to install the current `main` with `npm`,
which makes it hard to develop against unreleased bug fixes (in my case: https://github.com/Leaflet/Leaflet/pull/8288),
ahead of them being release in the next version of Leaflet.

This is especially awkward since the [downloads page](https://leafletjs.com/download.html) gives a `.zip` for:
> In-progress version, developed on the main branch.

---

See also:
- https://github.com/Leaflet/Leaflet/issues/8169
- https://github.com/Leaflet/Leaflet/issues/8165
- https://github.com/Leaflet/Leaflet/issues/8053
- https://github.com/Leaflet/Leaflet/issues/3617

### Suggested solution

Two options:

1. Require build and push `dist` folder, which would allow using [a GitHub URL](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#github-urls)
2. Build and push `-dev` package to [NPM](https://www.npmjs.com/package/leaflet), which would match how downloads page looks

### Alternatives considered

Work around as outlined [here](https://stackoverflow.com/a/73044271/21115).

jonkoops:
We could attempt to publish a `-dev` version to NPM on every commit although this would require some automation to create new version commits as NPM does not allow you to overwrite versions. So that would end up becoming something like `x.x.x-dev.1`, `-dev.2, etc.

Alternatively if we end up turning the library into a pure ESM package I think it would be possible install the package from GIT (as long as the entrypoint is the source code).

davetapley:
@jonkoops hmm, good point, I think that makes pushing to NPM a no go.

But! I also just learned that NPM can take a [tarball URL](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#urls-as-dependencies), so it might be as simple as making ⬇️ also push a `.tar` (I already checked and it doesn't work with the `.zip`) 😞 
[...]

PR #8350, opened 2022-07-20
Build tarball for NPM install

davetapley:
Hard to test without running the action, but should do https://github.com/Leaflet/Leaflet/issues/8349#issuecomment-1190756122

But I tested running commands myself and uploaded to https://davetapley.com/assets/leaflet.tar,
which can be used in a `package.json` as:
```json
 "dependencies": {
  "leaflet": "https://davetapley.com/assets/leaflet.tar",
```

Assuming if this works it'll push to https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/main/leaflet.tar
as existing https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/main/leaflet.zip does.


.github/workflows/main.yml
```diff
@@ -171,10 +171,17 @@ jobs:
           path: ./*
           key: ${{ runner.os }}-${{ github.sha }}-build
 
-      - name: Compress artifacts
+      - name: Compress artifacts to zip file
         working-directory: dist
         run: zip -r leaflet.zip .
 
+      - name: Create tarball for NPM install
+        run: |
```
jonkoops, line 179: To create a tarball I think it would be better to use [`npm pack`](https://docs.npmjs.com/cli/v8/commands/npm-pack) which creates a package exactly as NPM expects it.

davetapley: Oh nice! I didn't know about that, I'll amend...

davetapley: @jonkoops so long story short I don't think this is going to be so easy 😬 

Just doing `npm pack` on a fresh clone appears to work, but using the `.tgz` causes:
```
Attempted import error: 'leaflet' does not contain a default export (imported as 'L').
```

... which I _assume_ is because the actual build uses `rollup`.

If I `npm build` first, then `npm pack` I get https://github.com/Esri/esri-leaflet/issues/1200:

```
./node_modules/esri-leaflet/src/Layers/FeatureLayer/FeatureGrid.js
Attempted import error: 'toLatLngBounds' is not exported from 'leaflet' (imported as 'latLngBounds').
```

... any ideas? 😬 

---

I know my PR as given (using `tar` directly) isn't _great_, but it does work 😇

jonkoops:
I'm holding off on merging this in for now as I think we can actually start thinking about removing the compilation step from Leaflet entirely.

PR #8387, opened 2022-08-03
Fix touchzoom mobile

ranxu815:
修改移动端放大到最大级别后，图层随手指移动的问题

### DIFF
--- a/src/map/handler/Map.TouchZoom.js
+++ b/src/map/handler/Map.TouchZoom.js
@@ -74,7 +74,7 @@ export var TouchZoom = Handler.extend({
 
 		this._zoom = map.getScaleZoom(scale, this._startZoom);
 
-		//修改移动端放大到最大级别后，图层随手指移动的问题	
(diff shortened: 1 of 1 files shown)


IvanSanchez:
请用英文写.

I don't understand what this is supposed to fix. Zooming past the map's maxzoom is already handled by running `map._limitZoom(this._zoom)` a few lines later.

ranxu815:
hello,
        If the current map level exceeds the maximum level, you should return, otherwise the map will follow the mobile phone



 
From: Iván Sánchez Ortega
Date: 2022-08-03 17:41
To: Leaflet/Leaflet
CC: ranxu815; Author
Subject: Re: [Leaflet/Leaflet] Fix touchzoom mobile (PR #8387)
请用英文写.
I don't understand what this is supposed to fix. Zooming past the map's maxzoom is already handled by running map._limitZoom(this._zoom) a few lines later.

PR #8391, opened 2022-08-05
getZoom() after passing zoom options in panTo not working fix

anurag-dhamala:
```
map.panTo([50, 30], {zoom: 15});
map.getZoom();

```
The output was NaN. 
It will result in 15 as of now.

Hope this PR helps.


IvanSanchez:
The `panTo` options do not include `zoom`. This is [documented](https://leafletjs.com/reference#pan-options). Users setting the map center and zoom at the same time should be using the `setView` method already.

PR #8404, opened 2022-08-18
Fix added for tests failing: change the option wheelPxPerZoomLevel

kreloaded:
closes: #8479


Falke-Design:
Thank you for your contribution but your change will break the intention of the test. The reason for failing is in the Leaflet code and not in the test

PR #8418, opened 2022-08-27
Use preferred tile.openstreetmap.org URL

Firefishy:
Now that tile.openstreetmap.org supports HTTP/2 + HTTP/3 the old (a|b|c).tile.openstreetmap.org is no longer recommended.

Use tile.openstreetmap.org directly is faster and will likely result in a better cache hit ratio.

Signed-off-by: Grant Slater <git@firefishy.com>

### DIFF
--- a/debug/hacks/jitter.html
+++ b/debug/hacks/jitter.html
@@ -24,7 +24,7 @@
 </div>
 	<script>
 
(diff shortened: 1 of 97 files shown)


mourner (approved):
Thank you!

Falke-Design (approved):
Checked local, works well, thank you!

Firefishy:
Link for reference ;-) See: https://github.com/openstreetmap/operations/issues/737

PR #8431, opened 2022-09-07
Update download.md

andresrivas20:
I included an overview of the different types of versions that exist for Leaflet. This will allow the user to view the options quickly and easily. Also, I included an html overview section which outlines the html structure on how to incorporate Leaflet's hosted version within the header of the html file. I also included other aspects like the user's index.js and css. The goal is to show all together how the main setup should look like to allow the user to get a good base when first starting. Lastly, I referenced the quick start guide to direct the user just in case they have more intricate details about starting.

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -30,6 +30,13 @@ bodyclass: download-page
 Note that the main version can contain incompatible changes,
 so please read the changelog carefully when upgrading to it.
 
+### Leaflet Versions Overview
+There are many ways to download and use Leaflet, the following encompass the different ways. <br />
+	-Using a Hosted Version of Leaflet <br />
+	-Using a Downloaded Version of Leaflet <br />
+	-Using a JavaScript package manager <br />
+	-Leaflet source code <br />
+
 ### Using a Hosted Version of Leaflet
 
 The latest stable Leaflet release is available on several CDN's &mdash; to start using
@@ -87,3 +94,39 @@ Here are the steps to set it up:
 Now that you have everything installed, run `npm run build` inside the Leaflet directory.
 This will combine and compress the Leaflet source files, saving the build to the `dist` folder.
 
+### HTML Overview with Leaflet
+
+If choosing to use the hosted version of Leaflet, the example below can demonstrate how to set up the HTML page.
+```
+index.html
+<html>
+  <head>
+    <title>Leaflet Map</title>
+    
+    <!-- Leaflet CSS -->
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.8.0/dist/leaflet.css"
+    integrity="sha512-hoalWLoI8r4UszCkZ5kL8vayOGVae1oxXe/2A4AO6J9+580uKHDO3JdHb7NzwwzK5xr/Fs0W40kiNHxM9vyTtQ=="
+    crossorigin=""/>
+   
+    <!-- Leaflet JS -->
+    <script src="https://unpkg.com/leaflet@1.8.0/dist/leaflet.js"
+    integrity="sha512-BB3hKbKWOc9Ez/TAwyWxNXeoV9c1v6FIeYiBieIWkpLjauysF18NzgR1MBNBXf8/KABdlkX68nAhlwcDFLGPCQ=="
+    crossorigin=""></script>
+   
+    <!-- Main JS -->
+    <script type="module" src="./index.js"></script>
+  </head>
+  
+  <body>
+    <div id="map"></div>
+  </body>
+  
+</html>
+```
+```
+style.css
+#map {
+  height: 100%;
+}
+```
+For more information please take a look at the quick-start guide.


docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.8.0/dist/leaflet.css"
+    integrity="sha512-hoalWLoI8r4UszCkZ5kL8vayOGVae1oxXe/2A4AO6J9+580uKHDO3JdHb7NzwwzK5xr/Fs0W40kiNHxM9vyTtQ=="
+    crossorigin=""/>
+   
+    <!-- Leaflet JS -->
+    <script src="https://unpkg.com/leaflet@1.8.0/dist/leaflet.js"
+    integrity="sha512-BB3hKbKWOc9Ez/TAwyWxNXeoV9c1v6FIeYiBieIWkpLjauysF18NzgR1MBNBXf8/KABdlkX68nAhlwcDFLGPCQ=="
+    crossorigin=""></script>
+   
+    <!-- Main JS -->
```
Falke-Design, line 116:
```suggestion
    <!-- Your Main JS -->
```

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+
+If choosing to use the hosted version of Leaflet, the example below can demonstrate how to set up the HTML page.
+```
+index.html
+<html>
+  <head>
+    <title>Leaflet Map</title>
+    
+    <!-- Leaflet CSS -->
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.8.0/dist/leaflet.css"
```
Falke-Design, line 107: Please use a `latest` link `https://unpkg.com/leaflet@latest/dist/leaflet.css`

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+    integrity="sha512-hoalWLoI8r4UszCkZ5kL8vayOGVae1oxXe/2A4AO6J9+580uKHDO3JdHb7NzwwzK5xr/Fs0W40kiNHxM9vyTtQ=="
+    crossorigin=""/>
+   
+    <!-- Leaflet JS -->
+    <script src="https://unpkg.com/leaflet@1.8.0/dist/leaflet.js"
```
Falke-Design, line 112: Use `latest` link `https://unpkg.com/leaflet@latest/dist/leaflet.js`

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+  
+</html>
+```
+```
+style.css
+#map {
+  height: 100%;
+}
+```
+For more information please take a look at the quick-start guide.
```
Falke-Design, line 132: Please add a link to the quick-start guide

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+    <script src="https://unpkg.com/leaflet@1.8.0/dist/leaflet.js"
+    integrity="sha512-BB3hKbKWOc9Ez/TAwyWxNXeoV9c1v6FIeYiBieIWkpLjauysF18NzgR1MBNBXf8/KABdlkX68nAhlwcDFLGPCQ=="
+    crossorigin=""></script>
+   
+    <!-- Main JS -->
+    <script type="module" src="./index.js"></script>
+  </head>
+  
+  <body>
+    <div id="map"></div>
```
Falke-Design, line 121: Let us use inline-css
```suggestion
    <div id="map" style="height: 100%;"></div>
```

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+  </body>
+  
+</html>
+```
+```
+style.css
+#map {
+  height: 100%;
+}
+```
```
Falke-Design, line 131: Let us use inline-css
```suggestion
```

docs/download.md
```diff
@@ -30,6 +30,13 @@ bodyclass: download-page
 Note that the main version can contain incompatible changes,
 so please read the changelog carefully when upgrading to it.
 
+### Leaflet Versions Overview
```
Falke-Design, line 33: Please make all headlines like `Using a Hosted Version of Leaflet` a sub-headline of `Leaflet Versions Overview`. So add one `#`

Falke-Design (changes requested):
Thank you for contribution, please check out the requested changes.

docs/download.md
mourner: I'm generally not convinced this overview is needed. The page is already short and easy to skim headings on, and this only adds unnecessary bloat IMO.

docs/download.md
mourner: I'd prefer a specific version — we can use templated one here like above (`{{ site.latest_leaflet_version}}`).

docs/download.md
```diff
@@ -87,3 +94,39 @@ Here are the steps to set it up:
...
+If choosing to use the hosted version of Leaflet, the example below can demonstrate how to set up the HTML page.
+```
+index.html
+<html>
+  <head>
+    <title>Leaflet Map</title>
+    
+    <!-- Leaflet CSS -->
+    <link rel="stylesheet" href="https://unpkg.com/leaflet@1.8.0/dist/leaflet.css"
+    integrity="sha512-hoalWLoI8r4UszCkZ5kL8vayOGVae1oxXe/2A4AO6J9+580uKHDO3JdHb7NzwwzK5xr/Fs0W40kiNHxM9vyTtQ=="
```
mourner, line 108: Same for integrity hashes — needs `{{site.integrity_hash_css}}` etc.

mourner (changes requested):

docs/download.md
Falke-Design: In general I would also say a specific version is better but this means maintainance work for us and then I see more benefit if we have less work then have a specific version in the docs.

docs/download.md
Falke-Design: This makes it more complicated because we need to update the docs **after** the release. The hash is only accessible after the push to `unpkg` as far as I know.

mourner: @Falke-Design it's already used above on this page, it would remain like it's been before.

jonkoops:
I'm not a big fan of this change, it adds a lot of details to the page that don't necessarily make it easier to understand how to integrate Leaflet. Personally I am a fan of how this page looks now, which is nice and simple with links to more details when you actually need them.

Falke-Design:
@andresrivas20 thank you for your work but we decided that we will keep the page like it is now.

PR #8433, opened 2022-09-08
Update Center calculation to work for very small layers too

Falke-Design:
The new center calculation (#7603) caused on small layers that the center was outside of the layer. This is fixed now.

Old:
grafik

New:
grafik


mourner (approved):
Nice catch!

PR #8455, opened 2022-09-22
Update browser support list

jonkoops:
Removes browsers we no longer want to support for v2 from the site.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -145,26 +145,11 @@ If you have any questions, take a look at the <a href="https://github.com/Leafle
     <div class="no-break">
         <h3>Browser Support</h3>
 
-        <h4>Desktop</h4>
-
         <ul>
             <li>Chrome</li>
             <li>Firefox</li>
-            <li>Safari 5+</li>
-            <li>Opera 12+</li>
-            <li>IE 9&ndash;11</li>
+            <li>Safari</li>
             <li>Edge</li>
-        </ul>
-    </div>
-
-    <div class="no-break">
-        <h4>Mobile</h4>
-
-        <ul>
-            <li>Safari for iOS 7+</li>
-            <li>Chrome for mobile</li>
-            <li>Firefox for mobile</li>
-            <li>IE10+ for Win8 devices</li>
         </ul>
     </div>
 


docs/index.html
```diff
@@ -145,29 +145,14 @@ <h3>Map Controls</h3>
...
-            <li>Safari 5+</li>
-            <li>Opera 12+</li>
-            <li>IE 9&ndash;11</li>
+            <li>Safari</li>
             <li>Edge</li>
         </ul>
     </div>
 
-    <div class="no-break">
-        <h4>Mobile</h4>
```
jonkoops, line 161: Removed the sections for mobile and desktop as we want to support the same browsers on both.

Falke-Design:
Good point, yes don't merge this.

@mourner the doc changes of #8196 already published too

IvanSanchez (approved):
LGTM.

Do we want to mention Brave? Do we want to group Chrome & Edge under "blink engine browsers" or something like that?

jonkoops:
> Do we want to mention Brave? Do we want to group Chrome & Edge under "blink engine browsers" or something like that?

Yea, I was also not quite sure how to formulate this, naturally supporting Chromium-based browsers such as Brave and Samsung Internet would be fine considering they're all based on the same platform.

Malvoz (approved):

docs/index.html
```diff
@@ -145,29 +145,14 @@ <h3>Map Controls</h3>
...
-        <h4>Desktop</h4>
-
         <ul>
             <li>Chrome</li>
             <li>Firefox</li>
-            <li>Safari 5+</li>
-            <li>Opera 12+</li>
-            <li>IE 9&ndash;11</li>
+            <li>Safari</li>
             <li>Edge</li>
```
simon04, line 152: What about legacy Edge? It is [end of life](https://learn.microsoft.com/en-us/lifecycle/announcements/m365-ie11-microsoft-edge-legacy) since March 9, 2021.

jonkoops: Well, since Microsoft makes no distinction I think we can do the same. Naturally we don't support 'legacy' edge.

Falke-Design: Edge is now removed from the code

jonkoops: The version of Edge that was using the old engine yes, but we still support the Chromium version.

Falke-Design:
Merged because docs are built from branch `v1`

PR #8456, opened 2022-09-22
Fix the resource integrity hashes

jonkoops:
Also adds additional steps to the release process to prevent this from happening in the future.


Falke-Design:
> adds additional steps to the release process

are missing 😄

jonkoops:
> are missing :smile: 

No longer :+1:

Falke-Design (approved):

jonkoops:
@mourner still seeing errors on https://leafletjs.com/, not sure why these resources do not match the hash.

mourner:
@jonkoops I think I know — https://unpkg.com/leaflet@1.9.0/dist/leaflet.js has non-clean version in the header (should be just v1.9.0 but it also has a hash). Not sure why that happened, needs investigating. Meanwhile we can calculate the hashes for the files manually and update

mourner:
I updated them manually, https://leafletjs.com/ is OK now.

PR #8457, opened 2022-09-22
Fix Events.listens for nested propagations

### LINKED ISSUE #8450: Leaflet 1.9.0 breaks Leaflet Markercluster
### Checklist

### Steps to reproduce

Switching from Leaflet 1.8 to 1.9 breaks the Markercluster click event.

Nothing in the console, no errors, but no cluster-related event on the cluster marker. Instead it seems to have a "focus" event. But expanding a cluster doesn't work anymore.

I'm aware that this might eventually be a Markercluster bug, but at least it deserves some backwards compatibility considerations in Leaflet core.

### Expected behavior

Clicking on a Markercluster should expand it and zoom in.

### Current behavior

Nothing at all.

Note that with tabbing I can get a focus on that icon.

http://olga.indigofloat.at/bugdemo/demo.html

### Environment

- Leaflet version: 1.9.0
- Browser (with version): Firefox ESR (91.12.0esr), Chromium (104.0.51)
- OS/Platform (with version): Linux

Falke-Design:
The problem was that the value of `fn` was overwritten and then passed again to `listens` in the propagation.

closes #8450

### DIFF
--- a/spec/suites/core/EventsSpec.js
+++ b/spec/suites/core/EventsSpec.js
@@ -706,6 +706,17 @@ describe('Events', function () {
 			expect(marker.listens('test', true)).to.be(true);
 		});
 
+		it('is true if there is an event handler on parent parent', function () {
+			var fgP = L.featureGroup(),
+			fg = L.featureGroup().addTo(fgP),
+			marker = L.marker([0, 0]).addTo(fg),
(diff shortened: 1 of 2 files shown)


mourner (approved):
Nice fix 👍

PR #8459, opened 2022-09-22
Improve integrity generation for releases

### LINKED ISSUE #8458: Update integrity script to generate hashes from published files
https://github.com/Leaflet/Leaflet/pull/8456 shows it's very fragile to depend on local generation of integrity hashes. We could simplify the process and eliminate any potential errors by generating them from already published files from `unpkg` instead of relying on local files.

mourner:
Closes #8458. Instead of relying on the local files being correct when generating integrity hashes, just fetch the published ones from `unpkg`. This makes this step way less fragile, and simplifies the process.

I also updated the algorithm to `sha256` (one of the options per the spec) because it's shorter and looks less intimidating in examples, and we don't care about super duper security here.


jonkoops (approved):
Looks good to me 👍

PR #8463, opened 2022-09-23
Upgrade dependencies, including Sinon & Mocha

mourner:
We've put off upgrading ancient Sinon version because of IE, but now we can safely upgrade tooling to the latest versions.


jonkoops (changes requested):
Let's also remove `sinon` and `mocha` from the Dependabot [ignore list](https://github.com/Leaflet/Leaflet/blob/fd202ce5b618fd23f998fff79c865b1baedbd41d/.github/dependabot.yml#L14-L17) so it will offer upgrades for these dependencies.

jonkoops (approved):

PR #8474, opened 2022-09-26
Fix web archive link to v1.8.0 reference

Malvoz:
I suppose `reference-1.8.0.html` wasn't live for long enough between its creation and deletion for the web archive to keep it in the archive? Some sort of abuse protection perhaps?

Let's use https://web.archive.org/web/20220908180925/https://leafletjs.com/reference.html instead.


mourner:
Hmm, the old link works for me, what does it say when you try to visit it?

Malvoz:
@mourner Indeed! Yesterday night I was seeing a message stating something along the lines of "This page doesn't exist on the web", glad that was only temporarily. Thanks for checking!

PR #8480, opened 2022-09-26
Make test runner output cleaner

mourner:
When running `npm test`:

- Silence `console.warn` output in the terminal which was annoying and unnecessary.
- Silence Rollup warnings (circular deps / export) because we don't care about those during tests.

### DIFF
--- a/spec/karma.conf.js
+++ b/spec/karma.conf.js
@@ -48,6 +48,7 @@ module.exports = function (config) {
 		// Rollup the ES6 Leaflet sources into just one file, before tests
 		preprocessors: preprocessors,
 		rollupPreprocessor: {
+			onwarn: () => {}, // silence Rollup warnings
 			plugins: [
 				json()
 			],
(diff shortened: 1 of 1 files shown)


Falke-Design:
I don't know if this is so good. I think this one of the only ways for us to see if a (unwanted) warning is printed.

mourner:
There's no way for us to determine if a warning is unwanted unless we test specifically for warning output in the tests. As for Rollup, it's reusing the config used for the builds and we can see the warnings when running the build script.

Falke-Design:
I also use the test to developing, so I can clearly say if a warning is expected or not. 
I personally have no problems if the warnings are visible but it's also ok for me if we hide them.

jonkoops:
TBH I don't find the test output that verbose, and I'd prefer to see issues rather than to suppress that.

mourner:
IDK maybe that's a sign of OCD in me, but I just find the output ... wrong. Like, not crafted in the way it was intended. Maybe I should try supressing each of the current logs manually in each test then...

jonkoops (approved):
Actually, looking at the test result after running this PR it is a lot cleaner. LGTM.

PR #8481, opened 2022-09-26
Fix test on mac `changes the option 'wheelPxPerZoomLevel'`

### LINKED ISSUE #8479: `wheelPxPerZoomLevel` unit test always fails on Mac
The last unit test in `Map.ScrollWheelZoomSpec.js` added by @Falke-Design always fails for me locally on Mac:

```
Firefox 105.0 (Mac OS 10.15) Map.ScrollWheelZoom changes the option 'wheelPxPerZoomLevel' FAILED
	Error: expected 1 to be above 1 (node_modules/expect.js/index.js:96)
```

We should figure out if it's expected to fail and fix either the behavior of the test, making it more platform-agnostic.

Falke-Design:
Fix #8479

Changed `map.options.wheelPxPerZoomLevel = 30;` to `20` 

According to the fixes of #8298, the calculation for mac has changed.

Was easy to fix with BrowserStack ❤️


mourner (approved):

mourner:
Thanks!

PR #8484, opened 2022-09-27
Replace `Util.bind()` with `Function.bind()`

jonkoops:
Removes the `Util.bind()` function and replaces it with the standard [`Function.bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) method. `Util.bind()` included backwards compatibility with older browsers, which is no longer required now that we have raised our target browsers.

This change is also needed to update the linter configuration as it introspects `.bind()` calls for certain rules (such as [`no-invalid-this`](https://eslint.org/docs/latest/rules/no-invalid-this)).

### DIFF
--- a/src/layer/ImageOverlay.js
+++ b/src/layer/ImageOverlay.js
@@ -193,8 +193,8 @@ export var ImageOverlay = Layer.extend({
 
 		// @event load: Event
 		// Fired when the ImageOverlay layer has loaded its image
-		img.onload = Util.bind(this.fire, this, 'load');
-		img.onerror = Util.bind(this._overlayOnError, this, 'error');
+		img.onload = this.fire.bind(this, 'load');
+		img.onerror = this._overlayOnError.bind(this);
 
 		if (this.options.crossOrigin || this.options.crossOrigin === '') {
 			img.crossOrigin = this.options.crossOrigin === true ? '' : this.options.crossOrigin;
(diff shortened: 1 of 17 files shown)


src/layer/ImageOverlay.js
```diff
@@ -193,8 +193,8 @@ export var ImageOverlay = Layer.extend({
 
 		// @event load: Event
 		// Fired when the ImageOverlay layer has loaded its image
-		img.onload = Util.bind(this.fire, this, 'load');
-		img.onerror = Util.bind(this._overlayOnError, this, 'error');
```
jonkoops, line 197: Removed the `'error'` parameter here as the `_overlayOnError` method accepts no arguments.

src/layer/tile/TileLayer.js
```diff
@@ -197,12 +197,7 @@ export var TileLayer = GridLayer.extend({
 	},
 
 	_tileOnLoad: function (done, tile) {
-		// For https://github.com/Leaflet/Leaflet/issues/3332
-		if (Browser.ielt9) {
```
jonkoops, line 201: Removed this condition as we no longer support any version of Internet Explorer.

Falke-Design (approved):

mourner:
Should we include `L.Util.bind` as an alias to `fn.bind` just to retain compatibility with some older code or plugins that may still use it?

jonkoops:
@mourner Yeah, I wanted your input on this. We could leave a minimal implementation for it to retain backwards compatibility. Personally, I would like to get rid of large swaths of `Util` (and other components) and replace it with standard browser APIs, leaving a very minimal API exposed from Leaflet.

jonkoops:
@mourner I'd like to go ahead and merge this in, do you have any objections?

mourner (approved):

Issue #8483, opened 2022-09-27
Some of the unit tests are too slow


mourner:
Currently it takes ~40s to run the test suite. It used to be much faster, but we added a lot of unit tests since then and some of them are bottlenecks. We can use `karma-time-stats-reporter` plugin to discover which of the tests are slowing down the suite. Here's the list of the slowest ones:

```
4707ms Map #flyTo move to requested center and zoom, and call zoomend once                                                           
2877ms Map #flyTo flyTo start latlng == end latlng                                                                                   
2862ms TileLayer number of kittens loaded Loads 290, unloads 275 kittens on MAD-TRD flyTo()                                          
2699ms GridLayer number of 256px tiles loaded in synchronous animated grid @800x600px Loads 290, unloads 275 tiles on MAD-TRD flyTo()
1329ms ImageOverlay #getCenter should open popup at the center                                                                       
808ms  Popup L.Map#openPopup moves the map over a long distance to the popup if it is not in the view (keepInView)                   
683ms  Map.ScrollWheelZoom changes the option 'wheelPxPerZoomLevel'                                                                  
612ms  Path #events can add a layer while being inside a moveend handler                                                             
358ms  Map.ScrollWheelZoom changes the option 'wheelDebounceTime'                                                                    
314ms  LineUtil #polylineCenter computes center of a small line and test it on every zoom                                            
309ms  Map.ScrollWheelZoom zooms out while firing 'wheel' event                                                                      
307ms  Map.Keyboard plus/minus keys zoom in                                                                                          
306ms  Map.ScrollWheelZoom scrollWheelZoom: 'center'                                                                                 
306ms  Map.Keyboard does not move the map if disabled no move north                                                                  
306ms  Map.Keyboard does not move the map if disabled no zoom in                                                                     
306ms  Map.Keyboard arrow keys move the map west                                                                                     
305ms  Map.Keyboard plus/minus keys zoom out                                                                                         
305ms  Map.Keyboard arrow keys move the map east                                                                                     
305ms  Map.Keyboard arrow keys move the map north                                                                                    
303ms  Map.Keyboard arrow keys move the map south                                                                                    
297ms  Map.ScrollWheelZoom zooms in while firing 'wheel' event                                                                       
266ms  Map #containerPointToLayerPoint return point relative to LayerPoint                                                           
263ms  Map.DoubleClickZoom doubleClickZoom: 'center'                                                                                 
[...]

PR #8486, opened 2022-09-27
Add slow test stats and make some tests faster

mourner:
Ref #8483. Adds a Karma plugin that displays 10 slowest tests after running the test suite:

```
2833ms GridLayer number of 256px tiles loaded in synchronous animated grid @800x600px Loads 290, unloads 275 tiles on MAD-TRD flyTo()
2813ms TileLayer number of kittens loaded Loads 290, unloads 275 kittens on MAD-TRD flyTo()                                          
809ms  Popup L.Map#openPopup moves the map over a long distance to the popup if it is not in the view (keepInView)                   
665ms  Map.ScrollWheelZoom changes the option 'wheelPxPerZoomLevel'                                                                  
606ms  Path #events can add a layer while being inside a moveend handler                                                             
365ms  Map.ScrollWheelZoom changes the option 'wheelDebounceTime'                                                                    
321ms  LineUtil #polylineCenter computes center of a small line and test it on every zoom                                            
310ms  Map.ScrollWheelZoom scrollWheelZoom: 'center'                                                                                 
306ms  Map.ScrollWheelZoom zooms out while firing 'wheel' event                                                                      
300ms  Map.ScrollWheelZoom zooms in while firing 'wheel' event     
```

Also makes a few of the unit tests faster, cutting down overall time by ~10 seconds.

### DIFF
--- a/spec/suites/map/handler/Map.KeyboardSpec.js
+++ b/spec/suites/map/handler/Map.KeyboardSpec.js
@@ -16,6 +16,9 @@ describe("Map.Keyboard", function () {
 			zoomAnimation: false	// If true, the test has to wait extra 250msec
 		});
 
+		// make keyboard-caused panning instant to cut down on test running time
+		map.panBy = function (offset) { return L.Map.prototype.panBy.call(this, offset, {animate: false}); };
+
 		map.setView([0, 0], 5);
 
 		// Keyboard functionality expects the map to be focused (clicked or cycle-tabbed)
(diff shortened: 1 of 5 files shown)


jonkoops (approved):

spec/suites/map/handler/Map.KeyboardSpec.js
```diff
@@ -16,6 +16,9 @@ describe("Map.Keyboard", function () {
 			zoomAnimation: false	// If true, the test has to wait extra 250msec
 		});
 
+		// make keyboard-caused panning instant to cut down on test running time
+		map.panBy = function (offset) { return L.Map.prototype.panBy.call(this, offset, {animate: false}); };
+
```
Falke-Design, line 21: Does it make sense to remove everywhere `animate` only to have faster tests? The most users are using everything with animation, so tests should handle that too

mourner: Not everywhere, but in places where the tests don't care whether it's animated or not. It's keyboard interactions that are being tested in this place, not `panBy` animation (which is tested elsewhere).

PR #8487, opened 2022-09-27
Use different cache keys

### LINKED ISSUE #8460: Release build for NPM publishing might randomly run in non-release mode
OK, I figured out why https://unpkg.com/leaflet@1.9.0/dist/leaflet.js has the hash in the version (meaning it was built without `NODE_ENV=release`. It's a tricky one!

According to the [raw job log of v1.9.0 tag build job](https://pipelines.actions.githubusercontent.com/serviceHosts/63923d90-4a1f-44c3-8ec1-bd59c3ff7d7c/_apis/pipelines/1/runs/1868/signedlogcontent/29?urlExpires=2022-09-22T19%3A36%3A45.3572655Z&urlSigningMethod=HMACV1&urlSignature=dhSdPG3oXNKd0kWIbXr4cSWhIvbaXD1%2FD2Td27fejV8%3D), it built with `NODE_ENV=release` successfully. But later scrapped that build because it already found a build saved under the same cache key!

```
Cache hit occurred on the primary key Linux-a7e1bbcb779f181c599086f12555cab851ded25e-build, not saving cache.
```

So, basically it fetched the build that was created on a non-tagged job of the same commit, which was run in non-release mode but saved under the same cache key. To fix this, we'll probably have to update the related cache keys to include the tag name, e.g.:

```yaml
key: ${{ runner.os }}-${{ github.sha }}-${{ github.ref_name }}-build
```

jonkoops:
Use cache keys based on the `ref` to ensure no caches are shared between regular builds and release builds.

Closes #8460.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,22 +18,14 @@ jobs:
           check-latest: true
           cache: npm
 
-      - name: Cache dependencies
-        id: cache-dependencies
-        uses: actions/cache@v3
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
-
       - name: Install dependencies
-        if: steps.cache-dependencies.outputs.cache-hit != 'true'
         run: npm ci
 
       - name: Cache setup
         uses: actions/cache@v3
         with:
           path: ./*
-          key: ${{ runner.os }}-${{ github.sha }}-setup
+          key: ${{ github.ref }}-${{ github.sha }}-setup
 
   build:
     needs: setup
@@ -43,7 +35,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ./*
-          key: ${{ runner.os }}-${{ github.sha }}-setup
+          key: ${{ github.ref }}-${{ github.sha }}-setup
 
       - name: Set up Node
         uses: actions/setup-node@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -18,22 +18,14 @@ jobs:
...
 
-      - name: Cache dependencies
-        id: cache-dependencies
-        uses: actions/cache@v3
-        with:
-          path: node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
-
       - name: Install dependencies
-        if: steps.cache-dependencies.outputs.cache-hit != 'true'
```
mourner, line 29: Can you comment on this change too? Are node_modules being cached automatically now or are disabling caching because `npm ci` is fast enough to do from scratch every time?

jonkoops: The latter. Looking at tasks that are restoring and populating this cache they take about 22 seconds, this one takes 15. So I removed it as it's not worth the added complexity in the workflow.

mourner: Should we try caching `~/.npm` instead of `node_modules` [as the docs suggest](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), to see how that works just in case? In theory the cache store should be smaller and easier to restore.

jonkoops: This is already done by providing `cache: npm` to the action that sets up Node.js (see [their docs](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data)).

mourner (approved):

PR #8415, opened 2022-08-24
Quick-start: fix link in code block.

Sjlver:
The quickstart guide contains a code block with a link;
this should be rendered as an `<a>` HTML tag, but is
rendered as an actual link. I've fixed the Markdown by
changing it to a fenced code block. For consistency,
I've also adjusted the other code blocks in the page.

PR #8488, opened 2022-09-27
Disable <a> tag in code block

KristjanESPERANTO:


Falke-Design:
Thank you for your contribution but this already fixed by #8415

Issue #8201, opened 2022-04-29
Trying to move (drag) map while zoom animation is still in progress results in "click" behavior.


RomanistHere:
Here's the video I recorded here: https://leafletjs.com/examples/quick-start/ and also is working locally with 1.8.0 for me.

https://user-images.githubusercontent.com/49812531/166066386-8cfb5889-bca6-4972-b6bb-16bfe6e1d645.mp4

So the problem is when you're trying to move (by dragging) while just zoomed - you don't move the map and it reads my actions as click. So if you show something to user on click it's kinda annoying.

**Steps to reproduce**
- zoom out
- immediately try to drag the map

**Expected behavior**
Map is moving or at least nothing happens,

**Current behavior**
Map reads my action as `.on('click'`

**Environment**
- Leaflet version: 1.8.0 (on 1.7.4 had the same issue)
- Browser (with version):
![image](https://user-images.githubusercontent.com/49812531/166067166-edb39cbc-3f48-4cc3-8c74-a5f3ae523712.png)
- OS/Platform (with version): windows 10

**Additional context**
Proposed fast fix:  add debounce on scroll to off all the handlers for some time, I think 150-200ms should do it?

**Minimal example reproducing the issue**

https://leafletjs.com/examples/quick-start/

RomanistHere:
I mean I can do it myself (disable handlers on zoom) but I wondered if it's intended or not.

RomanistHere:
It procs on the 'dragend' so if you're doing "long" drag 200ms won't be enough. I stopped at 300 but it feels unnatural, so I'd still like some built-in way to prevent the issue.

![image](https://user-images.githubusercontent.com/49812531/166098243-87575bf7-ac10-4820-a7ba-01c1b217ee6e.png)

lgimberis:
Hi, as above I found the root cause of this and have submitted a PR.

PR #8489, opened 2022-09-28
Avoid ignoring dragstart events during zoom in/out (#8201)

lgimberis:
This line of code will cancel a drag that starts during the zoom in/out animation. Said line looks extremely specific so there may well be a good reason it's there, but neither comments nor commit messages indicate a particular reason for its existence, and after removing it the tests continue to pass.

Please let me know if I've done something silly in this process. It's my first time doing a pull request ;)

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -78,8 +78,6 @@ export var Draggable = Evented.extend({
 
 		this._moved = false;
 
-		if (DomUtil.hasClass(this._element, 'leaflet-zoom-anim')) { return; }
-
 		if (e.touches && e.touches.length !== 1) {
 			// Finish dragging to avoid conflict with touchZoom
 			if (Draggable._dragging === this) {


Falke-Design:
Maybe because of pinch zoom + drag on mobile

mourner:
Looks like this goes way back to 10 years ago and fixes zooming during pan animation and panning during zoom animation: https://github.com/Leaflet/Leaflet/commit/9625cdbbb96ab2e448d7444abde908b1fc10abdb

The current way it returns (based on `leaflet-zoom-anim` class) was introduced in https://github.com/Leaflet/Leaflet/pull/3050

lgimberis:
> **fixes** zooming during pan animation and panning during zoom animation

Currently in main, you can still zoom during panning. The problem in the issue this PR references is that if you **start** panning **during** a zoom, your "mousedown" input will be ignored and when you release, you'll instead click on the map. 

> Maybe because of pinch zoom + drag on mobile

Good point. I tested it and pinch zoom + drag works (to my eyes) exactly the same as before, probably because my fingers aren't robotic enough to pinch zoom without panning.

Feel free to compare the current version with this PR now hosted [here](https://lukasgimberis.com/leaflet).

PR #8493, opened 2022-09-29
Drop ESM entrypoint from package

jonkoops:
Removes the ESM entrypoint from the package file, as this is causing too many compatibility issues with certain bundlers/tooling.


Falke-Design:
@jonkoops Tests are failing

jonkoops:
Looks like I forgot to update the build config, should be good now.

Falke-Design (approved):
Looks good, but I can't test it local.

Maybe you can try it out with one of the reported issues? If not already done

mourner (approved):

jonkoops:
I'm going to revert the entire commit, that way we can be sure that things will continue working as expected.

PR #8498, opened 2022-10-01
Check if `getElement()` is undefined before adding focus listener (Canvas)

### LINKED ISSUE #8442: Tooltips throw exceptions when using vector shapes with the canvas renderer after #8247
### Checklist

### Steps to reproduce

1. Use latest `main` (9d1a841 at the time of creating this issue)
2. Set up a map with canvas renderer and a vector shape, like a rectangle.
3. Bind a tooltip to the vector. _Check JS console._
4. Move mouse over the vector. _Check JS console._
5. Move mouse out of the vector. _Tooltip does not disappear._

### Expected behavior

Tooltip does not fail to initialise or when reacting to events. They do not require DOM elements to function.

### Current behavior

Exceptions thrown on: tooltip initialisation, mouse events due to `getElement` calls introduced in #8247 671ebdf on a canvas-rendered shape (`getElement` is defined on the class, but has no element to return).

https://codepen.io/alex4401/pen/XWqpKMW

### Environment

- Leaflet version: main `9d1a841`
- Browser (with version): Firefox 104.0.2, but got user reports from all other browsers
- OS/Platform (with version): Arch Linux

Falke-Design:
Fix #8442

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -387,15 +387,19 @@ Layer.include({
 	},
 
 	_addFocusListenersOnLayer: function (layer) {
-		DomEvent.on(layer.getElement(), 'focus', function () {
-			this._tooltip._source = layer;
-			this.openTooltip();
-		}, this);
-		DomEvent.on(layer.getElement(), 'blur', this.closeTooltip, this);
+		if (layer.getElement()) {
+			DomEvent.on(layer.getElement(), 'focus', function () {
+				this._tooltip._source = layer;
+				this.openTooltip();
+			}, this);
+			DomEvent.on(layer.getElement(), 'blur', this.closeTooltip, this);
+		}
 	},
 
 	_setAriaDescribedByOnLayer: function (layer) {
-		layer.getElement().setAttribute('aria-describedby', this._tooltip._container.id);
+		if (layer.getElement()) {
+			layer.getElement().setAttribute('aria-describedby', this._tooltip._container.id);
+		}
 	},
 
 
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -387,18 +387,20 @@ Layer.include({
 	},
 
 	_addFocusListenersOnLayer: function (layer) {
-		if (layer.getElement()) {
-			DomEvent.on(layer.getElement(), 'focus', function () {
+		var el = layer.getElement();
+		if (el) {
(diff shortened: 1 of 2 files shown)


src/layer/Tooltip.js
```diff
@@ -387,15 +387,19 @@ Layer.include({
 	},
 
 	_addFocusListenersOnLayer: function (layer) {
-		DomEvent.on(layer.getElement(), 'focus', function () {
-			this._tooltip._source = layer;
-			this.openTooltip();
-		}, this);
-		DomEvent.on(layer.getElement(), 'blur', this.closeTooltip, this);
+		if (layer.getElement()) {
+			DomEvent.on(layer.getElement(), 'focus', function () {
```
mourner, line 391: Nit: while we're at it, let's avoid repeated calls to the same thing. E.g. do `var el = layer.getElement();` first.

mourner (approved):

src/layer/Tooltip.js
```diff
@@ -387,15 +387,19 @@ Layer.include({
...
+				this._tooltip._source = layer;
+				this.openTooltip();
+			}, this);
+			DomEvent.on(layer.getElement(), 'blur', this.closeTooltip, this);
+		}
```
Falke-Design, line 396:
```suggestion
		var el = layer.getElement();
		if (el) {
			DomEvent.on(el, 'focus', function () {
				this._tooltip._source = layer;
				this.openTooltip();
			}, this);
			DomEvent.on(el, 'blur', this.closeTooltip, this);
		}
```

src/layer/Tooltip.js
```diff
@@ -387,15 +387,19 @@ Layer.include({
...
 	},
 
 	_setAriaDescribedByOnLayer: function (layer) {
-		layer.getElement().setAttribute('aria-describedby', this._tooltip._container.id);
+		if (layer.getElement()) {
+			layer.getElement().setAttribute('aria-describedby', this._tooltip._container.id);
+		}
```
Falke-Design, line 402:
```suggestion
		var el = layer.getElement();
		if (el) {
			el.setAttribute('aria-describedby', this._tooltip._container.id);
		}
```

src/layer/Tooltip.js
Falke-Design: Changed

PR #8501, opened 2022-10-01
Remove char which causes error in YAML while serving

Falke-Design:
grafik


mourner (approved):

mourner:
It would be nice to add some kind of check to our CI to make sure those things don't slip past.

PR #8503, opened 2022-10-01
Added test cases for Map:addHandler method

### LINKED ISSUE #8372: Implement test for 'addHandler' method of Map

shtelzerartem:
Resolve #8372
Added 3 test cases for Map:addHandler method

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -611,6 +611,55 @@ describe("Map", function () {
 		});
 	});
 
+	describe("#addHandler", function () {
+		let clickCounter = 0;
+
+		L.ClickHandler = L.Handler.extend({
+			addHooks: function () {
+				L.DomEvent.on(window, 'click', this.handleClick, this);
+			},
+
+			removeHooks: function () {
+				L.DomEvent.off(window, 'click', this.handleClick, this);
+			},
+
+			handleClick: function () {
+				clickCounter++;
+			}
+		});
+
+
+		it("not enabled addHandler", function () {
+			map.addHandler('clickHandler', L.ClickHandler);
+
+			happen.once(window, {type: 'click'});
+			expect(clickCounter).to.eql(0);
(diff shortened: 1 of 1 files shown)


spec/suites/map/MapSpec.js
```diff
@@ -611,6 +611,62 @@ describe("Map", function () {
...
+
+				handleClick: callback
+			});
+		}
+
+		it("checking enabled method", function () {
+			L.ClickHandler = getHandler();
+			map.addHandler('clickHandler', L.ClickHandler);
+
+			map.clickHandler.enable();
```
Falke-Design, line 633: Let us add a check, if it's disabled by default
```suggestion
			map.addHandler('clickHandler', L.ClickHandler);

			expect(map.clickHandler.enabled()).to.eql(false);

			map.clickHandler.enable();
```

spec/suites/map/MapSpec.js
```diff
@@ -611,6 +611,62 @@ describe("Map", function () {
...
+			expect(map.clickHandler.enabled()).to.eql(true);
+
+			map.clickHandler.disable();
+			expect(map.clickHandler.enabled()).to.eql(false);
+		});
+
+		it("automatically enabled, when has a property named the same as the handler", function () {
+			container = createContainer();
+			map = L.map(container, {clickHandler: true});
```
Falke-Design, line 642: Don't create a new container. Remove the old map and use the same container
```suggestion
			map.remove();
			map = L.map(container, {clickHandler: true});
```

Falke-Design (changes requested):
Very nice, thank you. Only a few small things

spec/suites/map/MapSpec.js
shtelzerartem: Ahh, yeah, makes sense. Fixing.

Falke-Design (approved):
Very nice, thank you

PR #8504, opened 2022-10-01
Add build-docs job to main github actions file

### LINKED ISSUE #8502: Add a check to our CI to make errors in jekyll visible
[Local usage of jekyll](https://github.com/Leaflet/Leaflet/blob/main/CONTRIBUTING.md#improving-documentation)

exequiel09:
Closes #8502


.github/workflows/main.yml
```diff
@@ -79,6 +80,34 @@ jobs:
...
+        with:
+          path: ./*
+          key: ${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true
+          cache-version: 1
```
jonkoops, line 98: If I get this right then this value will have to be bumped manually according to the docs:

> In some rare scenarios (like using gems with C extensions whose functionality depends on libraries found on the system at the time of the gem's build) it may be necessary to ignore contents of the cache and get and build all the gems anew. In order to achieve this, set the `cache-version` option to any value other than `0` (or change it to a new unique value if you have already used it before.)

From what I am reading here I believe we can leave this empty for now.
```suggestion
```

jonkoops (changes requested):
This is a welcome change. Great work! I just have one small comment.

.github/workflows/main.yml
exequiel09: Ah yes, I thought it might be needed so I added it. I'll push a change to remove that line.

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
           path: ./*
           key: ${{ github.ref }}-${{ github.sha }}-build-release
 
+  build-docs:
+    needs: setup
```
jonkoops, line 84: Since this job doesn't need any of the JavaScript stuff (as far as I know), I think we can remove the dependency on `setup` and check out the repo instead. This should allow the job to start earlier, and reduce the time it takes for the CI to complete.

```suggestion
```

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
           path: ./*
           key: ${{ github.ref }}-${{ github.sha }}-build-release
 
+  build-docs:
```
jonkoops, line 83: This job is by far the longest running one on the CI, let's see if we can speed it up.

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
...
+    needs: setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: Restore setup
+        uses: actions/cache@v3
+        with:
+          path: ./*
+          key: ${{ github.ref }}-${{ github.sha }}-setup
```
jonkoops, line 91:
```suggestion
      - name: Check out repository
        uses: actions/checkout@v3
```

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
...
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true
+
+      - name: Install jekyll
+        run: gem install jekyll
```
jonkoops, line 100: Would it be possible to add `jekyll` to the gem-file instead and leverage the `bundler-cache` feature?

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
...
+
+      - name: Run bundle install
+        working-directory: ./docs
+        run: bundle install
+
```
jonkoops, line 105: Looks like this can be removed as `bundler-cache: true` already runs this command.
```suggestion
```

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
...
+        uses: actions/cache@v3
+        with:
+          path: ./*
+          key: ${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ env.RUBY_VERSION }}
+          bundler-cache: true
```
jonkoops, line 97: We should add the `working-directory` here as well.
```suggestion
          bundler-cache: true
          working-directory: ./docs
```

jonkoops (changes requested):
Sorry for the additional review comments, but looks like we can optimize things a bit.

.github/workflows/main.yml
exequiel09: Done 🙏

.github/workflows/main.yml
exequiel09: Done 🙏

.github/workflows/main.yml
exequiel09: Done 🙏

.github/workflows/main.yml
exequiel09: Done 🙏

.github/workflows/main.yml
exequiel09: Done 🙏

.github/workflows/main.yml
exequiel09: Requested changes done 🙏

mourner:
@exequiel09 it's now failing with a suprising error about not being able to build some internal example for some reason...

docs/Gemfile
```diff
@@ -3,3 +3,5 @@ source 'https://rubygems.org'
 gem 'github-pages'
 
 gem "webrick", "~> 1.7"
+
+gem "jekyll", "~> 3.9"
```
mourner, line 7: Do we have to explicitly specify `jekyll` here? The docs for `github-pages` state the following:

> Note: You are not required to install Jekyll separately. Once the github-pages gem is installed, you can build your site using jekyll build, or preview your site using jekyll serve. For more information about installing Jekyll locally, please see [the GitHub Help docs on the matter](https://help.github.com/articles/using-jekyll-with-pages#installing-jekyll).

exequiel09: @jonkoops requested me to add it https://github.com/Leaflet/Leaflet/pull/8504#discussion_r985103621

exequiel09: I'll try adding a commit without it . Let's see if it passes

mourner:
@exequiel09 looks like `jekyll` is available, but it still shows that error...

docs/Gemfile
jonkoops: Oh nice, I was not aware it was a transitive dependency.

exequiel09:
Maybe related to this. still reading docs: 0000-00-00-welcome-to-jekyll.markdown.erb

exequiel09:
It's really hard debugging this pipeline issue 😅

exequiel09:
Finally! It worked https://github.com/Leaflet/Leaflet/actions/runs/3165492356/jobs/5154569243 🎊

Falke-Design (changes requested):
The building works but it doesn't break when the error happens:

https://github.com/Falke-Design/Leaflet/actions/runs/3165535999/jobs/5154637242

docs/_config.yml
```diff
@@ -1,4 +1,4 @@
-exclude: [build, debug, node_modules, spec, src, CNAME, reference-tpl.html, CHANGELOG.md, README.md, LICENSE]
+exclude: [build, debug, node_modules, spec, src, CNAME, reference-tpl.html, CHANGELOG.md, README.md, LICENSE, vendor/bundle, vendor/cache, vendor/gems, vendor/ruby]
```
jonkoops, line 1: Just thinking out loud here, perhaps it would be better to use an `include` rather than to list all the things that should be excluded?

jonkoops (approved):
Code changes look good to me, but I am interested why it doesn't fail the job as expected when an error is introduced as @Falke-Design mentioned.

jonkoops:
> It's really hard debugging this pipeline issue 😅

Yeah, debugging stuff on Github Actions can be a bit of a hassle. I usually test on my own fork as well, less waiting on upstream capacity.

exequiel09:
oh. I didn't know that I can run GH actions on a fork 😲

exequiel09:
> I usually test on my own fork as well, less waiting on upstream capacity.
Not sure on how to run on my own fork. Quick search landed me on the event `pull_request_target`: https://dev.to/github/running-github-actions-on-forks-4e75

Not sure if I'm on the right direction for running in forks.

exequiel09:
Screen Shot 2022-10-01 at 9 48 21 PM

Okay so running locally, somehow error is swallowed. Running `echo $?` after the failed command yield exit status of 0 which it shouldn't be.

exequiel09:
> bundle exec executes a command in the context of the bundle (from bundle exec --help ).
> If you pass a executable name that is not found, e.g. bundle exec thisisnotwhatyouarelookingfor, it will exit with a status code != 0.

https://stackoverflow.com/a/64785111

I think I need to do some digging further to bubble up the error

Falke-Design:
Can we store the output on a file and check if it contains `Error`?

Falke-Design:
Found a repo: https://github.com/epstechtheatre/Jekyll-Tester

exequiel09:
> Found a repo: https://github.com/epstechtheatre/Jekyll-Tester

Push commit to use the suggestion action https://github.com/Leaflet/Leaflet/pull/8504/commits/6edc829630b4ca76757e605d088e51a6fb927d2a

exequiel09:
> Found a repo: https://github.com/epstechtheatre/Jekyll-Tester

I had to add back the setup node action to make the suggestion action for jekyll tester to work

.github/workflows/main.yml
```diff
@@ -79,6 +80,33 @@ jobs:
...
+          working-directory: ./docs
+
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
```
exequiel09, line 99: > Found a repo: https://github.com/epstechtheatre/Jekyll-Tester

I had to add back the setup node action to make the suggestion action for jekyll tester to work

exequiel09:
Seems like the suggested action for jekyll doesn't work. 

This line was never emitted https://github.com/epstechtheatre/Jekyll-Tester/blob/5331113e1372cd3a625eaa7398e334fe21397507/src/index.ts#L91

in this fork run: https://github.com/exequiel09/Leaflet/actions/runs/3165731159/jobs/5154974801#step:5:141

exequiel09:
Okay so it's not related to `bundler` but with `jekyll`: https://github.com/jekyll/jekyll/issues/8367

exequiel09:
Okay I know the fix. Pushing another one. removing the suggested action as well

.github/workflows/main.yml
exequiel09: Removed. Suggested action didn't work

exequiel09:
Added `--strict_front_matter`.

– https://stackoverflow.com/a/68600279

exequiel09:
Screen Shot 2022-10-02 at 3 16 28 AM

Adding `--strict_front_matter` will make exit status be `1`

Falke-Design (approved):
I have no idea how the workflows are working but the output looks good now 😄

docs/_config.yml
jonkoops, line 1: Do we really need anything from the `vendor` directory? Perhaps we can exclude all of it?
```suggestion
exclude: [build, debug, node_modules, spec, src, CNAME, reference-tpl.html, CHANGELOG.md, README.md, LICENSE, vendor]
```

docs/Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
 gem 'github-pages'
-
-gem "webrick", "~> 1.7"
+
```
jonkoops, line 4: Can we revert this file? Looks like nothing has changed.

jonkoops (changes requested):
Workflow is looking good to me, got a couple small questions but after that I think we can merge.

docs/_config.yml
exequiel09: I'm sure on the proper solution but the jekyll website suggested that fix on CI issue https://github.com/Leaflet/Leaflet/pull/8504#issuecomment-1264430063

docs/Gemfile
exequiel09: Done

exequiel09:
> I have no idea how the workflows are working but the output looks good now 😄

@Falke-Design we need the build to command to return non-zero exit status to make GitHub actions check fail. In that case jekyll build returns 1 when there are errors and 0 if there's none.

docs/_config.yml
exequiel09: Mentioned in jekyll website:

> The proper solution is to incorporate the default setting for exclude: rather than override it completely:
For versions up to v3.4.3, the exclude: setting must look like following:
```yaml
exclude:
  - Gemfile
  - Gemfile.lock
  - node_modules
  - vendor/bundle/
  - vendor/cache/
  - vendor/gems/
  - vendor/ruby/
  - any_additional_item # any user-specific listing goes at the end
```

mourner (approved):
Nice work!

jonkoops (approved):
Looking good to me!

jonkoops:
Thank you so much for this contribution @exequiel09. This change helps us out a lot!

PR #8512, opened 2022-10-02
Add `getWheelPxFactor` and fix test `changes the option 'wheelPxPerZoomLevel'` for mac

### LINKED ISSUE #8479: `wheelPxPerZoomLevel` unit test always fails on Mac

Falke-Design:
Fixes #8479

The problem was the `devicePixelRatio`. 
I made now `getWheelPxFactor` public.

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -268,9 +268,10 @@ export function getMousePosition(e, container) {
 export function getWheelPxFactor() {
 	// We need double the scroll pixels (see #7403 and #4538) for all Browsers
 	// except OSX (Mac) -> 3x, Chrome running on Linux 1x
-	return (Browser.linux && Browser.chrome) ? window.devicePixelRatio :
-		Browser.mac ? window.devicePixelRatio * 3 :
-		window.devicePixelRatio > 0 ? 2 * window.devicePixelRatio : 1;
+	var ratio = window.devicePixelRatio;
+	return Browser.linux && Browser.chrome ? ratio :
+		Browser.mac ? ratio * 3 :
+		ratio > 0 ? 2 * ratio : 1;
 }
 
 // @function getWheelDelta(ev: DOMEvent): Number
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -263,15 +263,16 @@ export function getMousePosition(e, container) {
 	);
 }
 
+// @function getWheelPxFactor(): Number
+// Gets the wheel pixel factor based on the devicePixelRatio
+export function getWheelPxFactor() {
+	// We need double the scroll pixels (see #7403 and #4538) for all Browsers
+	// except OSX (Mac) -> 3x, Chrome running on Linux 1x
+	return (Browser.linux && Browser.chrome) ? window.devicePixelRatio :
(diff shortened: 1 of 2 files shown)


src/dom/DomEvent.js
```diff
@@ -263,23 +263,24 @@ export function getMousePosition(e, container) {
...
 
+// @function getWheelPxFactor(): Number
+// Gets the wheel pixel factor based on the devicePixelRatio
+export function getWheelPxFactor() {
+	// We need double the scroll pixels (see #7403 and #4538) for all Browsers
+	// except OSX (Mac) -> 3x, Chrome running on Linux 1x
+	return (Browser.linux && Browser.chrome) ? window.devicePixelRatio :
+		Browser.mac ? window.devicePixelRatio * 3 :
+		window.devicePixelRatio > 0 ? 2 * window.devicePixelRatio : 1;
+}
```
mourner, line 274: Does this have to be a function, or could we export a constant instead?

Falke-Design: I think a constant would be ok too. I didn't know if we want to have constants in `DomEvent` so I created a function.

Thinking loud: What if the user zooms on the page (not map) and changes the `devicePixelRatio`? With a function we would always have the correct value. How much performance costs do we have with a function instead of a constant?

mourner: Yeah, devicePixelRatio possibly changing is a good point. Let's keep it a function then.

src/dom/DomEvent.js
```diff
@@ -263,23 +263,24 @@ export function getMousePosition(e, container) {
...
 }
 
+// @function getWheelPxFactor(): Number
+// Gets the wheel pixel factor based on the devicePixelRatio
+export function getWheelPxFactor() {
+	// We need double the scroll pixels (see #7403 and #4538) for all Browsers
+	// except OSX (Mac) -> 3x, Chrome running on Linux 1x
+	return (Browser.linux && Browser.chrome) ? window.devicePixelRatio :
+		Browser.mac ? window.devicePixelRatio * 3 :
+		window.devicePixelRatio > 0 ? 2 * window.devicePixelRatio : 1;
```
mourner, line 273: Nit: I don't like the repetition.

```suggestion
	var ratio = window.devicePixelRatio;
	return Browser.linux && Browser.chrome ? ratio :
		Browser.mac ? ratio * 3 :
		ratio > 0 ? 2 * ratio : 1;
```

mourner (approved):

jonkoops:
@Falke-Design is this something that needs a backport to `v1`?

Falke-Design:
No need but doesn't hurt if we do

jonkoops:
Good to know. Just wanted to know if we were missing any essential fixes.

PR #8513, opened 2022-10-02
Fix typo in file

Guilospanck:


jonkoops:
IMHO the wording here is perfectly fine and I see no reason to change this. Also this is not a typo.

PR #8420, opened 2022-08-29
closes #8366: panBy() test

adrianaris:
closes #8366

PR #8514, opened 2022-10-03
added map panBy method unit tests

adeeb18:
I included a few unit tests for panBy that I thought could be helpful. Please let me know if I am missing anything or if anything is incorrect, as this is my first contribution! Thanks.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+npm test
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "eslint-config-mourner": "^2.0.3",
         "git-rev-sync": "^3.0.2",
         "happen": "~0.3.2",
-        "husky": "^8.0.1",
+        "husky": "^8.0.0",
         "karma": "^6.4.1",
         "karma-chrome-launcher": "^3.1.1",
         "karma-edge-launcher": "^0.4.2",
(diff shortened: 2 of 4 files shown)


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+npm test
```
Falke-Design, line 4: Don't change files that have nothing to do with this PR. If this is a necessary change, create a new PR only with this changes and we will decide if we want to merge it.

package.json
```diff
@@ -11,7 +11,7 @@
     "eslint-config-mourner": "^2.0.3",
     "git-rev-sync": "^3.0.2",
     "happen": "~0.3.2",
-    "husky": "^8.0.1",
+    "husky": "^8.0.0",
```
Falke-Design, line 14: Revert

spec/suites/map/MapSpec.js
```diff
@@ -2073,6 +2073,21 @@ describe("Map", function () {
...
+			expect(function () {
+				map.panBy();
+			}).to.throwError();
+		});
+
+		it("pans the map by the given number of pixels", function () {
+			var center = L.latLng(10, 10);
+			map.setView(center);
+			var change = L.point(200, 300);
+			expect(map.panBy(change)).to.be(map);
```
Falke-Design, line 2088: The test should be check if the map has successfull and correct panned, but in the `expect` you only test if the return value of `map.panBy()` == `map`

package-lock.json
```diff
@@ -16,7 +16,7 @@
         "eslint-config-mourner": "^2.0.3",
         "git-rev-sync": "^3.0.2",
         "happen": "~0.3.2",
-        "husky": "^8.0.1",
+        "husky": "^8.0.0",
```
Falke-Design, line 19: Revert

Falke-Design (changes requested):
Thank you for your contribution, saddly we already have a open PR for `panBy` #8420.
I will close this PR but I let you some feedback here.

PR #8516, opened 2022-10-03
Add Dependabot config for Bundler dependencies

jonkoops:
Adds a Dependabot config to keep our [Bundler](https://bundler.io/) dependencies up to date.


.github/workflows/main.yml
```diff
@@ -89,7 +88,6 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ env.RUBY_VERSION }}
```
jonkoops, line 92: We no longer need to pass this version as the action reads it from the `.ruby-version` file.

.gitignore
```diff
@@ -15,8 +15,6 @@ coverage/
 .mailmap
 component.json
 debug/local/
-Gemfile.lock
```
jonkoops, line 18: I've checked in the lockfile as is [recommended](https://bundler.io/guides/rationale.html#checking-your-code-into-version-control) by the Bundler docs. This prevents accidental upgrades of dependencies that might break the process.

.gitignore
```diff
@@ -15,8 +15,6 @@ coverage/
...
 yarn.lock
 *.log
 .eslintcache
-.jekyll-cache
```
jonkoops, line 22: I've moved all the Jekyll related files that should be ignored to their own `.gitignore` in the `docs` directory.

docs/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.2
```
jonkoops, line 1: The `github-pages` gem does not support Ruby 3, so I set this to the latest version of v2.

docs/Gemfile
```diff
@@ -1,5 +1,3 @@
 source 'https://rubygems.org'
 
 gem 'github-pages'
-
-gem "webrick", "~> 1.7"
```
jonkoops, line 5: This gem had to be added because of compatibility issues with Ruby 3. Downgrading the Ruby version allows us to remove it again.

package.json
```diff
@@ -52,7 +52,6 @@
     "watch": "rollup -w -c build/rollup-config.js",
     "uglify": "uglifyjs dist/leaflet-src.js -c -m -o dist/leaflet.js --source-map filename=dist/leaflet.js.map --source-map content=dist/leaflet-src.js.map --source-map url=leaflet.js.map --comments",
     "bundlemon": "bundlemon --subProject js --defaultCompression none && bundlemon --subProject js-gzip --defaultCompression gzip",
-    "serve": "cd docs && bundle exec jekyll serve",
```
mourner, line 55: Any reason to remove this, or does it stop working after the PR updates?

docs/.ruby-version
mourner: Do we have to be specific here, or can we just specify e.g. `2` just to avoid the hassle of bumping this regularly?

mourner (approved):

docs/Gemfile
Falke-Design: If we keep this line, it would be possible to use with Ruby 3. Why not support 2 & 3 if it is easy possible?

package.json
Falke-Design: I would like to keep it

CONTRIBUTING.md
```diff
@@ -145,14 +145,17 @@ browse the [`docs/ directory`](https://github.com/Leaflet/Leaflet/tree/main/docs
 choose a certain file for editing, click the Edit button, make changes and follow instructions from there.
 Once it gets merged, the changes will immediately appear on the website.
 
+To work on the documentation locally ensure you have Ruby installed. You can download it from the [Ruby website](https://www.ruby-lang.org/en/downloads/) or use [`rbenv`](https://github.com/rbenv/rbenv) (recommended).
+
+You'll need to install the same Ruby version as specified in [`.ruby-version`](./docs/.ruby-version). If you are using `rbenv` you can install this by running `rbenv install` from the `docs/` directory.
+
```
Falke-Design, line 151: Ruby 2 is only necessary for `github-pages` but not for running `jekyll`. I know the user should use the same as Github but I think the most users will install the newest version or already have version 3 because Ruby 2 is not maintained anymore. Can we support Ruby 2 and 3 on local for `jekyll`?

Falke-Design (changes requested):

docs/Gemfile
jonkoops: I would not run it if it's not officially supported. There might be other issues that we're not seeing at the surface.

package.json
jonkoops: Removed it since the docs pretty much live on their own and don't need NPM. This way the docs stay very pure on the Ruby side of things.

jonkoops:
https://github.com/Leaflet/Leaflet/pull/8516#discussion_r985631130 ◀️ Unfortunately this is not possible.

CONTRIBUTING.md
```diff
@@ -145,14 +145,17 @@ browse the [`docs/ directory`](https://github.com/Leaflet/Leaflet/tree/main/docs
...
- 2. Run `gem install jekyll`.
- 3. Enter the directory where you cloned the Leaflet repository
- 4. Make sure you are in the `main` branch by running `git checkout main`
- 5. Run `npm run serve`
- 6. Open `localhost:4000` in your web browser.
+1. Open a terminal in the `docs/` directory.
+2. Make sure you are on the `main` branch by running `git checkout main`.
+3. Run `bundle install` to install the dependencies.
+4. Run `bundle exec jekyll serve` to serve the documentation.
+5. Open [http://localhost:4000](http://localhost:4000) in your web browser.
```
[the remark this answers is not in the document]
jonkoops: Not quite sure what you want me to do here. The reason I put the installation of Ruby in a separate section is that it's a prerequisite, the user will likely do this once. Whereas this list of steps is something the contributor might want to repeat.

As for `npm run serve`, I removed it as I personally do not like mixing tooling. Like this everything related to the docs lives in the `docs/` directory, and JS tooling lives in the root.

jonkoops:
> Do we have to be specific here, or can we just specify e.g. `2` just to avoid the hassle of bumping this regularly?

Unfortunately, this does not support version ranges. So an exact version is always needed.

docs/.ruby-version
jonkoops: Unfortunately, this does not support version ranges. So an exact version is always needed.

mourner: If you reverted to supporting both v2 and v3, should we drop that file as well then so that there's no Ruby version hassle when running docs locally?

jonkoops:
Nice, the CI is all green now :heavy_check_mark:

CONTRIBUTING.md
```diff
@@ -145,14 +145,17 @@ browse the [`docs/ directory`](https://github.com/Leaflet/Leaflet/tree/main/docs
...
- 1. [Install Ruby](https://www.ruby-lang.org/en/downloads/) if you don't have it yet.
- 2. Run `gem install jekyll`.
- 3. Enter the directory where you cloned the Leaflet repository
- 4. Make sure you are in the `main` branch by running `git checkout main`
- 5. Run `npm run serve`
- 6. Open `localhost:4000` in your web browser.
+1. Open a terminal in the `docs/` directory.
+2. Make sure you are on the `main` branch by running `git checkout main`.
+3. Run `bundle install` to install the dependencies.
+4. Run `bundle exec jekyll serve` to serve the documentation.
```
Falke-Design, line 157: Please revert to `npm run serve` then it is ok for me

Falke-Design (changes requested):

CONTRIBUTING.md
jonkoops: Done :+1:

mourner (commented):
Looks good to me, but would still want a comment on this:

> If you reverted to supporting both v2 and v3, should we drop the `.ruby-version` file as well then so that there's no specific Ruby version installation hassle when running docs locally?

jonkoops:
I think we should keep the `.ruby-version` file around as it seems to be a convention in the world of Ruby tooling.

For example the Github Action we use to set up Ruby works out of the box with it, and `rbenv` automatically switches tot he correct version of Ruby when `cd`ing into a directory that has this file.

PR #8517, opened 2022-10-03
chore: replaced `substr` with `substring`

k-rajat19:
`substr` method is deprecated, it should be avoided using.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


mourner (approved):

jonkoops (approved):
Yeah, this seems like a valid piece of refactoring. Thanks!

PR #8518, opened 2022-10-03
Speed up PathSpec's "Add layer inside move handler" test: 611ms to 5ms (122x faster)

rjackson:
Related to #8483

The `panBy` calls within this test would invoke pan animations, adding 250ms per pan to the test's execution time. I assume these setTimeout calls were set at 300ms to wait for that animation to finish.

I don't believe this test needs to invoke animations. Disabling them and dropping the setTimeouts saves us half a second on the tests.


jonkoops (approved):

mourner:
Nice!

PR #8525, opened 2022-10-04
Typo in line 63

harshit-dugar:
There was a grammatical error in line 63


harshit-dugar:
@michael Please check this out

jonkoops:
Not sure if this is a change we want. But if you are fixing it here please also fix the `og:title` meta tag. @mourner what would you like us to do with PRs like this?

mourner:
Nah, I don't want to encourage the Hacktober crowd to do trivial and unnecessary PRs like this.

PR #8534, opened 2022-10-06
Fix intermittent wobble when setMaxBounds(map.getBounds())

### LINKED ISSUE #8532: Map shaking after setMaxBounds and setMinZoom
### Checklist

### Steps to reproduce

1. Create a map.
2. Set its maximum borders and minimum zoom to those currently visible on the screen.

`const map = L.map('map').setView([50.450036, 30.5241361], 13)`

`L.tileLayer('http://{s}.tile.osm.org/{z}/{x}/{y}.png', {`
`    attribution:`
`        '&copy; <a href="http://osm.org/copyright">OpenStreetMap</a> contributors',`
`}).addTo(map)`

`map.setMaxBounds(map.getBounds()).setMinZoom(map.getZoom())`

### Expected behavior

The map is limited by the maximum borders and the minimum zoom without any unexpected behavior.

### Current behavior

Map is limited by the maximum borders and the minimum zoom, but the map is moving (shaking).
I have looked at these issues: #[1946](https://github.com/Leaflet/Leaflet/pull/1946), #[2107](https://github.com/Leaflet/Leaflet/issues/2107), #[2011](https://github.com/Leaflet/Leaflet/issues/2011), #[2081](https://github.com/Leaflet/Leaflet/issues/2081), #[1866](https://github.com/Leaflet/Leaflet/issues/1866), #[2187](https://github.com/Leaflet/Leaflet/pull/2187). But it seems that there is no solution to my problem there, so I don't think it's a duplicate to any of previous issues. If it is, could you please help me to resolve this issue?
Thank you!

https://jsfiddle.net/fd3mertj/1/

### Environment

- Leaflet version:  1.8.0, 1.9.1
- Browser (with version): Mozilla 105.0.1 (64 bit), Opera 90.0.4480.84
- OS/Platform (with version): Windows 10 laptop

rjackson:
Fixes #8532

At certain coordinate and/or screen size combinations, `Map.setMaxBounds` ended up triggering an infinite `_panInsideMaxBounds` loop due to slight discrepancies in reprojected center points across loops.

We are supposed to break out of this loop if our recalculated centers are identical:

https://github.com/Leaflet/Leaflet/blob/4372a9fcadce23d23a21370439179c09d1118ab3/src/map/Map.js#L503-L514

And our `_limitCenter` code was already set up to catch subpixel offsets, ignore them, and returning the identical center to break the loop.

Unfortunately it seems the projection instability can cause up to 1px offsets, which this code was not catching.

This PR adds a failing test for the problem as reported, using the `setView` values from the bug report. This test does appear to consistently fail.

The PR also updates `_limitCenter` to ignore up to 1 pixel offsets, which fixes the issue.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -458,6 +458,20 @@ describe("Map", function () {
 			map.setMaxBounds(bounds, {animate: false});
 			expect(map.off.called).not.to.be.ok();
 		});
+
+		it("avoid subpixel / floating point related wobble (#8532)", function (done) {
+			map.setView([50.450036, 30.5241361], 13);
+
+			var spy = sinon.spy();
+			map.on('moveend', spy);
+			map.setMaxBounds(map.getBounds());
+
+			// Unfortunately this is one of those tests where we need to allow at least one animation tick
+			setTimeout(function () {
+				expect(spy.called).to.be(false);
+				done();
+			}, 300);
+		});
 	});
 
 	describe("#setMinZoom and #setMaxZoom", function () {
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1543,7 +1543,7 @@ export var Map = Evented.extend({
 		// If offset is less than a pixel, ignore.
 		// This prevents unstable projections from getting into
 		// an infinite loop of tiny offsets.
-		if (offset.round().equals([0, 0])) {
+		if (Math.abs(offset.x) <= 1 && Math.abs(offset.y) <= 1) {
 			return center;
 		}
 


spec/suites/map/MapSpec.js
```diff
@@ -458,6 +458,20 @@ describe("Map", function () {
 			map.setMaxBounds(bounds, {animate: false});
 			expect(map.off.called).not.to.be.ok();
 		});
+
+		it("avoid subpixel / floating point related wobble (#8532)", function (done) {
+			map.setView([50.450036, 30.5241361], 13);
+
+			var spy = sinon.spy();
+			map.on('moveend', spy);
```
mourner, line 466: Should be `once` so that it doesn't pollute other tests, right?
```suggestion
			map.once('moveend', spy);
```

Falke-Design (approved):
Very nice!

mourner (approved):
Looks good to me!

spec/suites/map/MapSpec.js
mourner: Ah right.

PR #8537, opened 2022-10-06
Remove the long-deprecated `L.Mixin.Events` and `_flat`

mourner:


Falke-Design:
Maybe we want to remove `'Deprecated use of _flat, please use L.LineUtil.isFlat instead.` too

Falke-Design:
Solves: https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-1639414

> We should remove the deprecated functions. There are a few warnings while running tests:
> 
>     * `_flat`: Deprecated use of _flat, please use L.LineUtil.isFlat instead.
> 
>     * `L.Mixin.Events`: Deprecated include of L.Mixin.Events: [...] inherit from L.Evented instead.
> 
>     * Remove the obsolete factory of Circle: https://leafletjs.com/reference.html#circle-l-circle

Falke-Design (approved):

PR #8543, opened 2022-10-08
Change default font size of scale bar

secondl1ght:
Hello 👋🏼 

I had a user notice that when the map initally loads at a zoom level of `3`, the metric scale bar slightly cuts off the unit value of `km`. Please see screenshot below. 

image

Since this is a scale bar and the width of it cannot be changed, I think the only solution must be to reduce the font size of the text slightly so that this doesn't occur. 

It is an edge-case but noticeable enough that a user reported it. I have tested font size `10px` on many different zoom levels and it seems to fix this issue. 

I also searched for related issues to this and did not find any discussion on this previously. (Maybe it was decided not to be a big enough deal to fix?). Please let me know if I missed any background discussion on this issue.

Here is how it looks with the adjusted font size at the same zoom level:

image

Thanks

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -433,6 +433,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-control-scale-line {
 	border: 2px solid #777;
 	border-top: none;
+	font-size: 10px;
 	line-height: 1.1;
 	padding: 2px 5px 1px;
 	white-space: nowrap;


IvanSanchez:
I don't think reducing the font size is the adequate answer - there's been a whole set of accessibility changes from 1.8 to 1.9, and this would go against.

If anything, there should be a `minWidth` option/value in `L.Control.Scale`, and just multiply the length and the measurement by a factor of 2 whenever the actual length is smaller.

Malvoz:
Another option is to instead allow the text to overflow the container. I.e. remove:
https://github.com/Leaflet/Leaflet/blob/3cb07e4577b9527a3453db1ee619fc7d5e730907/dist/leaflet.css#L439


Some "other" maps do it too:

Screenshot_20221008-200838_Maps

The text could however overlap the scale itself, potentially utilizing a small text-shadow for contrast could help, e.g. `text-shadow: 0 1px #fff`.

Malvoz:
Closing per https://github.com/Leaflet/Leaflet/pull/8543#issuecomment-1272377313, feel free to continue the discussion.

Falke-Design:
Just chaning the size of the scale box will result in a not correct scale

PR #8548, opened 2022-10-09
Allow the scale control's text to overflow the container

Malvoz:
Fixes the issue described by @secondl1ght in https://github.com/Leaflet/Leaflet/pull/8543:

scale-overflow-hidden

By allowing the scale control's text to overflow the container and giving it a slight `text-shadow`, as suggested in https://github.com/Leaflet/Leaflet/pull/8543#issuecomment-1272492170:

scale-text-overflow-visible

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -439,7 +439,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	-moz-box-sizing: border-box;
 	     box-sizing: border-box;
 	background: rgba(255, 255, 255, 0.8);
-  text-shadow: 1px 1px #fff;
+	text-shadow: 1px 1px #fff;
 	}
 .leaflet-control-scale-line:not(:first-child) {
 	border-top: 2px solid #777;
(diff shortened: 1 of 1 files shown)


Malvoz:
The addition of `text-shadow` here results in higher contrast, in which case the need for https://github.com/Leaflet/Leaflet/pull/8547 might not be as immediate. Anyhow, if https://github.com/Leaflet/Leaflet/pull/8547 is merged together with this PR we'd have:

scale-text-overflow-visible2

dist/leaflet.css
```diff
@@ -436,12 +436,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	line-height: 1.1;
 	padding: 2px 5px 1px;
 	white-space: nowrap;
-	overflow: hidden;
 	-moz-box-sizing: border-box;
 	     box-sizing: border-box;
-
-	background: #fff;
 	background: rgba(255, 255, 255, 0.8);
+  text-shadow: 1px 1px #fff;
```
mourner, line 442:
```suggestion
	text-shadow: 1px 1px #fff;
```

Malvoz: Oops.

mourner (approved):

secondl1ght:
Thanks for fixing this so quickly, will this be included in the next release?

Falke-Design:
@secondl1ght yes in the v2

PR #8549, opened 2022-10-09
transfer mixins into Map.js for testing purposes (#6397)

### LINKED ISSUE #6397: [ES Modules] Layer depends on Map
## How to reproduce

- Leaflet Version: 1.3.2

Importing TileLayer should not pull Map, et al. into the bundle

```js
import { TileLayer } from 'leaflet'
TileLayer.extend({});
```

> Note: I'm using source-map-explorer to view what is getting pulled in. For ease of diagnosing what is in the bundle I also alias {'leaflet': 'leaflet/src/Leaflet.js'}

But unfortunately a large portion of the world is pulled in because TileLayer -> GridLayer -> Layer -> Map -> "World"

It appears the Layer.js file is extending the Map object, but that extension doesn't appear to actually depend on the Layer object itself. The added behaviour should likely be moved to Map.js.

adrianaris:
closes #6397, perhaps

This is a intermediary PR that has several purposes:
  1. To gouge your interest for having those modifications (maybe going into 2.0 you already have a drown-out plan for how you'll go about tree-shaking and whatever else);
  2. To see if I'm going the right way with it;
  3. To spark a discussion on how you'd like to go about decoupling or if you have certain design pattern you'd like to go towards.
  4. To see if I'm actually capable of engaging this problem.

Am noob, pls no bully.


mourner:
Generally I think this is the right thing to do — thanks for taking on this! Years ago I introduced this pattern because we wanted to make things modular, and also avoid `Map` becoming too huge, but I think this is doing more harm than good — increasing complexity, introducing side effects to modules, etc. But I'll welcome more opinions.

Can you please rebase on `main`? Looks like some already landed PR changes are showing up in your diff. Also let's make sure this doesn't break the docs flow by generating them and reviewing the difference locally.

adrianaris:
Damn, failed rebase again. I'll try again later :)

Falke-Design:
I always rebase by running `git pull http://github leaflet repo clone url`

src/layer/Layer.js
```diff
@@ -117,159 +116,3 @@ export var Layer = Evented.extend({
 		map.fire('layeradd', {layer: this});
 	}
 });
-
-/* @section Extension methods
```
adrianaris, line 121: Removing this section was a mistake I believe.

src/control/Control.js
```diff
@@ -114,61 +112,3 @@ export var Control = Class.extend({
 export var control = function (options) {
 	return new Control(options);
 };
-
-/* @section Extension methods
```
adrianaris, line 118: mistake

src/layer/Layer.js
```diff
@@ -117,159 +116,3 @@ export var Layer = Evented.extend({
...
- * @method getAttribution(): String
- * This optional method should return a string containing HTML to be shown on the `Attribution control` whenever the layer is visible.
- *
- * @method beforeAdd(map: Map): this
- * Optional method. Called on [`map.addLayer(layer)`](#map-addlayer), before the layer is added to the map, before events are initialized, without waiting until the map is in a usable state. Use for early initialization only.
- */
-
-
-/* @namespace Map
- * @section Layer events
```
adrianaris, line 144: Should I bring those comments in their original location or it doesn't matter where they are? From looking at the served documentation site it seams that everything is in its place, except for completely deleted section-comments

mourner: Just generate the docs with `npm run docs` and see which lines changed in the generated file, this will be the best way to make sure nothing is missing.

mourner:
@adrianaris I meant not commit the regenerated docs, just check the difference locally to see what's generated differently and investigate/fix appropriately. Currently committed docs show way too much difference so apparently something broke.

adrianaris:
Yes, I understood what you meant, but I noticed the big difference and I didn't understand if it's normal.

I pushed it for you to see.

I'll try to figure it out but if you have some pointers for me now that would be nice.

mourner:
Summoning @IvanSanchez for ideas 🙏

IvanSanchez:
> Summoning @IvanSanchez for ideas

💨 *IvanSanchez appears in a puff of smoke* 💨

Personally I'm against this PR - the point of using ES modules (& treeshaking) is to reduce the amount of code being loaded, and `import`ing controls, svg and canvas renderers in the `Map` class immediately cancels that benefit.

In other words: there are use cases where a developer wants a Leaflet map displayed without any vector symbols (lines/circles/polygons), or without a `Canvas` renderer - and the approach proposed by this PR will *always* load the code for all controls, all handlers, and all renderers.

I don't feel like coupling/cohesion is a problem in the current architecture. Yes, the behaviour of controls/handlers is defined by options in the `Map`, but the alternative is to force the user to explicitly instantiate controls/handlers with the desired behavioural options and override the default.... and all of a sudden this becomes OpenLayers, and instantiating a tile layer requires the explicit instantiation of three different classes. (Which, granted, is not what this PR does, but it's what my brain is extrapolating)

Yes, ideally, this removes coupling between Leaflet's internal classes.... at the cost of increasing Leaflet's API surface (i.e. increasing coupling between leaflet's code and user's code).

(I'd even venture to say that the current form of this PR **increases** coupling - by having the controls/handlers/renderers use stuff from `Map` which is defined elsewhere)

I'm wondering if the ""right"" way to go is something akin to what I'm doing in my [not-so-secret-anymore WebGL map library](https://gitlab.com/IvanSanchez/gleo) - have a base `Map` class with no extraneous functionality, and then a façade `Map` class which loads a set of expected defaults - see https://gitlab.com/IvanSanchez/gleo/-/blob/master/src/MercatorMap.mjs

adrianaris:
@IvanSanchez you mean something like this: https://github.com/adrianaris/Leaflet/commit/2215804c8a78dd9257b9afc4da17c0d10682326d ???

@mourner Maybe is best to put this on hold for now? I think I understand what Ivan suggests but I don't know if I can come up with a good solution that fits the current design. Things are out of my depth as is but I'm willing to bang my head at it if I have the blessing and some help. What's your opinion on Ivan's comment?

Thanks

IvanSanchez:
@adrianaris Yeah, kinda, but it's not exactly what I meant.

On one hand, "blank" classes with no functionality are my (currently) preferred way of dealing with circular dependencies. So yes on that.

On the other: There has to be a way to both:
- (a) Allow users having a `L.Map` with the default expected controls and handlers (zoom buttons, mouse drag, etc)
- (b) Allow users having a `L.Map` with no controls nor handlers (so they set up their own), loading code strictly on demand.

My point being that this PR doesn't allow for (b). Specifically, https://github.com/Leaflet/Leaflet/pull/8549/files#diff-f7c61cee5bee8604986cfce68530ef8df7316a4e93ea6a2ddfb6e97fa4234421R12 makes things messy (and tree-shaking basically impossible)

(Sorry if I cannot provide a magical 100% correct answer - this is by no means an easy architectural problem to tackle)

adrianaris:
Closing this as I feel like it actually has the same scope with what leaflet members are working towards with their major refactoring for the 2.0 (https://github.com/Leaflet/Leaflet/discussions/8576#discussioncomment-3888966) + It's out of my depth and I believe that chances are slim for me to come up with a solution that you would accept. Cheers

jonkoops:
Thanks for the efforts, starting a discussion like this is always appreciated.

PR #8556, opened 2022-10-10
Expand layers control on Enter keydown

larsgw:
When navigating a Leaflet map by keyboard, instead of expanding the Layers control immediately when the toggle is focused, instead only expand when the Enter key is also pressed.

Fix https://github.com/Leaflet/Leaflet/issues/7191

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -204,7 +204,15 @@ export var Layers = Control.extend({
 		link.setAttribute('role', 'button');
 
 		DomEvent.on(link, 'click', DomEvent.preventDefault); // prevent link function
-		DomEvent.on(link, 'focus', this.expand, this);
+		DomEvent.on(link, 'keydown', function (e) {
+			if (e.keyCode === 13) {
+				DomEvent.on(section, 'click', DomEvent.preventDefault);
+				this.expand();
+				setTimeout(function () {
+					DomEvent.off(section, 'click', DomEvent.preventDefault);
+				});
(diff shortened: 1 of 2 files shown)


src/control/Control.Layers.js
```diff
@@ -204,7 +204,15 @@ export var Layers = Control.extend({
...
-		DomEvent.on(link, 'focus', this.expand, this);
+		DomEvent.on(link, 'keydown', function (e) {
+			if (e.keyCode === 13) {
+				DomEvent.on(section, 'click', DomEvent.preventDefault);
+				this.expand();
+				setTimeout(function () {
+					DomEvent.off(section, 'click', DomEvent.preventDefault);
+				});
+			}
+		}, this);
```
Malvoz, line 215: Thanks @larsgw, however this doesn't work using a screen reader, see related  https://github.com/Leaflet/Leaflet/pull/8202#pullrequestreview-963070739.

larsgw: I see. I don't have Windows so I don't think I can test NVDA or Narrator, but I'll try to reproduce with Orca and take a look.

Malvoz (approved):
Nice, works well, tested in both NVDA and Narrator.

Malvoz:
I'll open a separate issue later for https://github.com/Leaflet/Leaflet/issues/7191#issuecomment-1273839451.

Falke-Design (approved):
@larsgw thank you!

PR #8559, opened 2022-10-12
Remove Internet Explorer from browser detection

jonkoops:
Removes Internet Explorer from the browser detection code and removes all related code based on these checks. This removes the following APIs:

- `L.Browser.ie`
- `L.Browser.ielt9`
- `L.Browser.ie3d`

### DIFF
--- a/docs/examples/choropleth/example.md
+++ b/docs/examples/choropleth/example.md
@@ -92,7 +92,7 @@ css: "#map {
 			fillOpacity: 0.7
 		});
 
-		if (!L.Browser.ie && !L.Browser.opera && !L.Browser.edge) {
+		if (!L.Browser.opera && !L.Browser.edge) {
 			layer.bringToFront();
 		}
 
--- a/debug/vector/geojson.html
+++ b/debug/vector/geojson.html
@@ -107,9 +107,7 @@
 				fillOpacity: 0.7
 			});
 
-			if (!L.Browser.ie) {
-				layer.bringToFront();
-			}
+			layer.bringToFront();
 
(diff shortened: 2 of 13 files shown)


IvanSanchez (approved):
This feels like an underwhelming historic moment.

I can't see anything wrong with the code changes.

Malvoz (approved):

Falke-Design (approved):
Nice!
It's important that we make in the changelog clearly visible all API changes, but I think this is clear

jonkoops:
@Falke-Design Yeah, I was thinking we could perhaps already create a draft release for 2.0 with all the breaking changes and migration guide. That way we don't have to retroactively compile it all.

Falke-Design:
Sounds good. But let us only add for now the main things like breaking / API changes so that we don't forget them. And then before the release, I go over all the commits again and add everything else

jonkoops:
Exactly what I was thinking, I'll get on that after merging this in.

docs/examples/choropleth/example.md
```diff
@@ -92,7 +92,7 @@ css: "#map {
 			fillOpacity: 0.7
 		});
 
-		if (!L.Browser.ie && !L.Browser.opera && !L.Browser.edge) {
+		if (!L.Browser.opera && !L.Browser.edge) {
```
mourner, line 95: BTW, the other two switches here are suspicious. For `edge`, is this detecting the old IE-based Edge, not the Chromium one? And why would Opera not work here if it's also pretty much a flavour of Chrome now? Maybe we could get rid of both of these too.

PR #8562, opened 2022-10-13
Fix worldCopyJump with Keyboard

### LINKED ISSUE #8560: Pan with keyboard and dragging with mouse gives different result when passing dateline
### Checklist

### Steps to reproduce

Hi
_If I add map1.on('moveend', function(event){  console.log('moveend lng=',this.getCenter().lng); });_
and drag the map passed the dateline (lng=180) I get (almost) the correct mapCenter, but if I pan the map using the keyboard I get lng > 180 if I pan west to east. See attached video

### Expected behavior

I would expect the resulting lng to be the same, and also < 180

### Current behavior

https://user-images.githubusercontent.com/11357972/195583584-02b474d7-eaf0-4572-bd69-299740b1fc87.mov

_No response_

### Environment

- Leaflet version: 1.9.2
- OS/Platform (with version): Windows

Falke-Design:
Fixes #8560

### DIFF
--- a/src/map/handler/Map.Keyboard.js
+++ b/src/map/handler/Map.Keyboard.js
@@ -151,10 +151,15 @@ export var Keyboard = Handler.extend({
 					offset = toPoint(offset).multiplyBy(3);
 				}
 
-				map.panBy(offset);
-
 				if (map.options.maxBounds) {
-					map.panInsideBounds(map.options.maxBounds);
+					offset = map._limitOffset(toPoint(offset), map.options.maxBounds);
+				}
+
+				if (map.options.worldCopyJump) {
+					var newLatLng = map.wrapLatLng(map.unproject(map.project(map.getCenter()).add(offset)));
+					map.panTo(newLatLng);
+				} else {
+					map.panBy(offset);
 				}
 			}
 		} else if (key in this._zoomKeys) {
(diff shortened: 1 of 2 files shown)


jonkoops (approved):

jonkoops:
@Falke-Design is this something we would like to back-port as well?

Falke-Design:
> @Falke-Design is this something we would like to back-port as well?

Yes I think we can back-port this too

src/map/handler/Map.Keyboard.js
```diff
@@ -151,10 +151,15 @@ export var Keyboard = Handler.extend({
...
 
-				map.panBy(offset);
-
 				if (map.options.maxBounds) {
-					map.panInsideBounds(map.options.maxBounds);
+					offset = map._limitOffset(toPoint(offset), map.options.maxBounds);
+				}
+
+				if (map.options.worldCopyJump) {
+					var newLatLng = map.wrapLatLng(map.unproject(map.project(map.getCenter()).add(offset)));
```
IvanSanchez, line 159: This looks like something that could be moved to the `Map`'s `panBy()` method.

Falke-Design: I also thought about that but I do not completly understand how the original wrapping in Map.Drag works.

Falke-Design: I digged into it, it is not so easy. I created a new issue / task for this. #8625

PR #8571, opened 2022-10-14
Fixes event target of popupopen event and adds test

### LINKED ISSUE #8558: Incorrect marker popupopen event e.target reference
### Checklist

### Steps to reproduce

[https://codepen.io/philiplarner/pen/zYjyGdY](https://codepen.io/philiplarner/pen/zYjyGdY)

1. Hover the red marker (A)

### Expected behavior

It should raise an alert referencing marker A as it does in older versions of leaflet? 

To see please adjust the leaflet version 1.7.1 ("Settings" in the header > "Pen Settings" > "JS" and adjust version number)

### Current behavior

The alert from marker A references marker B

https://codepen.io/philiplarner/pen/zYjyGdY

### Environment

- Leaflet version: 1.9.2

I think there is an issue with leaflet "popupopen" event starting at version 1.8.0. When you programmatically close a popup and open it, triggering the "popupopen" event on a specific marker when it opens: the event target doesnt match what the event was bound on? N.B this is when all markers share the same popup instance

Belair34:
The main issue was that _prepareOpen in DivOverlay was changed so that it didn't have any logic to select the correct source. Let me know if anything should be changed/refactored.
Fixes #8558


src/layer/DivOverlay.js
```diff
@@ -230,23 +230,20 @@ export var DivOverlay = Layer.extend({
...
+		var source = parent;
 		if (!source._map) { return false; }
 
 		if (source instanceof FeatureGroup) {
 			source = null;
-			var layers = this._source._layers;
+			var layers = parent._layers;
 			for (var id in layers) {
-				if (layers[id]._map) {
+				if (layers[id]._map && (Number(id) === this._source._leaflet_id || this._source instanceof FeatureGroup)) {
```
IvanSanchez, line 241: I'd rather use the `stamp` function from `Util`, instead of querying `_leaflet_id`. It's what the codebase currently does. 

The only reason not to do so would be a use case where `this._source` is not a `L.Layer` - I don't think that's possible here.

src/layer/Popup.js
```diff
@@ -420,7 +420,7 @@ Layer.include({
 	// @method openPopup(latlng?: LatLng): this
 	// Opens the bound popup at the specified `latlng` or at the default popup anchor if no `latlng` is passed.
 	openPopup: function (latlng) {
-		if (this._popup && this._popup._prepareOpen(latlng || this._latlng)) {
+		if (this._popup && this._popup._prepareOpen(latlng || this._latlng, this)) {
```
Falke-Design, line 423: It would be enough to do `this._popup._source = this;`

Falke-Design (changes requested):
The test is great, but I think the other changes are not necessary. It would be enough to do  `this._popup._source = this;` in `openPopup` & `openTooltip`

src/layer/DivOverlay.js
Belair34: @Falke-Design That was both to select the correct layer as well as to preserve the behavior of FeatureGroups. However, your idea was a better way to select the correct layer; I just had to add a little bit to it to preserve FeatureGroup behavior.

src/layer/Popup.js
```diff
@@ -420,10 +421,14 @@ Layer.include({
...
 	// Opens the bound popup at the specified `latlng` or at the default popup anchor if no `latlng` is passed.
 	openPopup: function (latlng) {
-		if (this._popup && this._popup._prepareOpen(latlng || this._latlng)) {
-
-			// open the popup on the map
-			this._popup.openOn(this._map);
+		if (this._popup) {
+			if (!(this instanceof FeatureGroup)) {
+				this._popup._source = this;
+			}
```
Falke-Design, line 427: I still don't understand it, but I think it has it reason. Can you please add a test for this?

Falke-Design (changes requested):

src/layer/DivOverlay.js
```diff
@@ -260,6 +260,7 @@ export var DivOverlay = Layer.extend({
 				throw new Error('Unable to get source layer LatLng.');
 			}
 		}
+
```
Falke-Design, line 263: Please revert this

src/layer/Popup.js
Belair34: I'm not sure what kind of test I could add, but without this if statement there are 4 existing tests that fail. They all seem to relate to feature groups of popups/tooltips with functions as content (in bindTooltip/bindPopup).

Falke-Design (approved):

IvanSanchez (approved):

plarner30:
hey many thanks for the fix do you guys know if there will be a new minor release any time soon?

jonkoops:
@plarner30 see #8634

PR #8574, opened 2022-10-15
Fix issue with _latLngToNewLayerPoint

Z3rio:
This Pull Request fixes an issue with the _animateZoom function.
The error seems to be happening because _map is undefined, it is recreational simply by creating a map + a marker on the map, then just zoom in & out a bunch whilst the marker is visible (atleast how I did it).

This is solved by simply checking if the map exists in these functions.

Images & videos of the error:

image
**Video:**

https://user-images.githubusercontent.com/54480523/196011117-803344df-edbe-404e-930a-ea58d18294f6.mp4


Z3rio:
> @Z3rio could you [rebase](https://docs.gitlab.com/ee/topics/git/git_rebase.html) this PR on `main` and resolve the conflicts?

There, this should now be Done.

Falke-Design:
Closing this because of no repsone and I'm not able to reproduce the error. If we get new information I will reopen this PR.

StefanoMantero:
Hi all, i understand that this issue has been closed but the problem still persists using leaflet with Vue3 (Or a famework that proxy the map object). All the workarounds i can find are "just unwrap it", at it's not practical when you are developing something that is more complex than a map with 2 merkes on it. 

I can easily replicate this just by allowing leaflet to autoclose a popup generating the `Uncaught TypeError: Cannot read properties of null (reading '_latLngToNewLayerPoint')` error.

Could this issue be re-opened ? Thanks!

Falke-Design:
@StefanoMantero please create a demo which throws the error, then I will reopen the issue and look into it.
The best would be to use www.leafletjs.com/edit.html but this is a vanilla js sandbox.

Issue #8395, opened 2022-08-10
Extend support for {-y} tiled maps into infinite 


BrianOfTsushima:
### Checklist

- [X] I've searched through the [plugins](https://leafletjs.com/plugins.html) to make sure this feature isn't already available, or think it shouldn't require a plugin.
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this feature hasn't been requested already.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Motivation

We use leaflet to plot game data, and game worlds/coordinate systems typically have the origin centered on the map, and have +x to the right/east, and +y upwards/north.   

For all items *except* tile numbering, simply using CRS.Simple works beautifully!  Alas, the tiles themselves still load with +Y going down/south, and place the origin at the upper/left corner of tile 0,0.

My request is to modify TileLayer::getTileURL to support {-y} syntax from within infinite coordinate systems, and to do so with the origin of the bottom left corner of tile 0,0.

### Suggested solution

An example modification we've made in TileLayer::getTileURL that you see below (discussion follows)

```
if (this._map) {
			// treat all !crs.infinite as before -- inverting by subtract from globalTileRange
			if (!this._map.options.crs.infinite) {
				var invertedY = this._globalTileRange.max.y - coords.y;
				if (this.options.tms) {
					data['y'] = invertedY;
				}
				data['-y'] = invertedY;
			}
			// Infinite? Yes.  But if we have also inverted (typically by CRS.Simple) the Y axis
			// then we can supply -y by just inverting (and offset by 1, to get origin bottom left)
			else if (this._map.options.crs.transformation._c == -1) {
				data['-y'] = (-1 * coords.y) - 1;  // invert y, offset origin to bottom/left corner
			}
  		}
```
First, this change seems very safe -- because the only new behavior is supporting {-y} in some cases which previously were unsupported.  I'm presuming nobody is counting on the old/error behavior of -y being undefined!

Second the "offset origin to bottom/left corner" comment is there to clarify with the final subraction of 1 is present.  That's the most opaque bit IMO.

The only further discussion is if the check of _c == -1 is useful.  It's extra defensive but again not obviously necessary.

### Alternatives considered

Considered doing this in {y} but that seems much more dangerous to existing sites depending on the old behavior.

Issue #4338, opened 2016-03-21
TMS doesn't work with L.CRS.Simple in 1.0


hyperknot:
As provided by a user in #4333 using L.CRS.Simple with tms: true results in an exception where `this._globalTileRange` is uninitialised in tileLayer / getTileUrl.
https://github.com/Leaflet/Leaflet/blob/b60e18d5c6c1ceacdcccfc03f184c0afa516607b/src/layer/tile/TileLayer.js#L77

http://playground-leaflet.rhcloud.com/jac/1/edit?html,output

It works untouched in 0.7, but is broken in 1.0

Related issue: https://github.com/Leaflet/Leaflet/issues/2298

cc: @perliedman

perliedman:
The problem is that `L.CRS.Simple` is an so called "infinite" CRS (https://github.com/Leaflet/Leaflet/blob/master/src/geo/crs/CRS.Simple.js#L24), which means it does not have any bounds.

TMS, on the other hand, needs an upper bound for the CRS's Y coordinate, to be able to flip the tile row.

This means that TMS will not work with `L.CRS.Simple` out of the box in 1.0. On the other hand, the only reason it worked in 0.7 was that `L.CRS.Simple` had implicit bounds, and from what I remember a lot of people ran into weird problems because those implicit bounds were rarely the same as the actual limits.

If you want to use TMS and a CRS that resembles `L.CRS.Simple`, you should extend it, set `infinite: false` and give it the bounds you want. This could, of course, be made clearer with a tutorial or some kind of warning in the docs.

jieter:
Regarding #4337, should we not try to replace `{-y}` if `map.crs.infinite === true` to prevent this error ?

perliedman:
@jieter yes, sounds like a good plan. In general it would of course be very nice to provide a better error message than that some internal variable is `undefined`!

balciseri:
Yes at the moment if i reference the master version with the {-y} fix i can't use crs: L.CRS.Simple,
because it will not work.

jieter:
Hang on, working on it...

jieter:
@perliedman Leaflet doesn't seem to do much explaining in error messages. I've added an example of the error produced by a small `L.CRS.Simple`-example to the description of #4344 .

PR #8575, opened 2022-10-16
Add support `{-y}` for TileLayer

ShivangMishra:
modified TileLayer::getTileURL to support {-y} syntax from within infinite coordinate systems. 
The change is safe - the only new behavior is supporting {-y} in some cases which previously were unsupported. 

Fixes: #8395


Falke-Design (changes requested):
I didn't checked the code but a test is needed anyway

ShivangMishra:
@Falke-Design  I need to add a test ?

Falke-Design:
For some reason there was a expliciet test for `{-y}` to be not set on infinite crs (this is also then one who is failing):

https://github.com/Leaflet/Leaflet/blob/c33a57ca3a59c25f8874d09aec5fa2f1c6a98e5f/spec/suites/layer/tile/TileLayerSpec.js#L327-L345

ShivangMishra:
@Falke-Design This is what I understood - 
In the issue #4338 , the problem was that TMS "**needs an upper bound for the CRS's Y coordinate, to be able to flip the tile row**". In the case of infinite crs, we don't have an upper bound, that's probably why they added the check on {-y}.
 In our case, **we don't use the upper bound**, we simply set **{-y} = (-1 * coords.y) - 1**.

Falke-Design:
@mourner @IvanSanchez maybe you can take a look into this. The tile and rendering logic is still pretty hard for me

ShivangMishra:
@Falke-Design I reforked this repo. Should I open a new PR for this issue #8395 ?

Falke-Design:
@ShivangMishra go for it

PR #8580, opened 2022-10-16
Fractional zoom level for GridLayer

### LINKED ISSUE #8276: GridLayer fails to load map tiles when `redraw` called at fractional zoom level
### Checklist

### Steps to reproduce

Create a GridLayer and set the map zoom level to a fractional value, e.g. `2.5`. Call the `redraw()` method on the grid layer. The layer makes requests with urls like:

`<baseurl>/2.5/2/4.png`

These requests fail, since the `{z}` value is typically only available in integer steps. So the map fails to load.

### Expected behavior

The `_tileZoom` property should always be an integer value, so that the URL requested only resolves to integer values.

### Current behavior

I believe the bug was introduced in this commit: 

https://github.com/Leaflet/Leaflet/commit/2e8638761ca91506e3dc2b132da4c8185aab27f0

For example in `_setView()` the `_tileZoom` is `round(zoom)`

https://github.com/Leaflet/Leaflet/blob/c7d3e97d9ae3fada1f7ab047d4d56c06cec35d12/src/layer/tile/GridLayer.js#L552

I think [this line](https://github.com/Leaflet/Leaflet/blob/c7d3e97d9ae3fada1f7ab047d4d56c06cec35d12/src/layer/tile/GridLayer.js#L231) ought to be:

`var tileZoom = Math.round(this._clampZoom(this._map.getZoom()));`

_No response_

### Environment

- Leaflet version: git main branch (also present in leaflet@1.8.0 from npm)
- Browser (with version): firefox-100.0-4.fc35.x86_64 (also seen in chrome / safari)
- OS/Platform (with version): Fedora 35 Linux (also seen on mac and windows)

raychanks:
Round off the fractional zoom level in order to use an integer value for requesting tiles.

Closes #8276


Falke-Design:
Maybe related to #3575

raychanks:
@Falke-Design I think there are two issues here.
1. `GridLayer` will load map tiles with decimal `z` value in url, like `<baseurl>/2.5/2/4.png`, leading to unsuccessful requests. (This PR tries to address this)
2. The map may have gaps showing up between tiles. (This is the issue discussed in #3575)

I think these two are separate issues and this PR tries to address the first one, or am I missing something?

raychanks:
Oh, it seems that I messed up the auto formatting during merge and caused quite a lot of line changes. If this is what I think as a separate issue, I can close this PR and open another PR for easier review.

raychanks:
Close PR in favour of #8613 for cleaner changes.

PR #8582, opened 2022-10-16
Upgrade Rollup to version 3

jonkoops:
Upgrades the build system to the latest version of Rollup. Some migration was required as Rollup now uses Node.js to directly load the config, instead of bundling it first.


build/banner.js
```diff
@@ -1,7 +0,0 @@
-export function createBanner(version) {
```
jonkoops, line 1: I've moved the `createBanner` function into the Rollup config as that's the only place it's being used.

package.json
```diff
@@ -9,7 +9,6 @@
     "bundlemon": "^1.4.0",
     "eslint": "^8.23.1",
     "eslint-config-mourner": "^2.0.3",
-    "git-rev-sync": "^3.0.2",
```
jonkoops, line 12: The `git-rev-sync` package relies on some CommonJS features that are not available in ESM mode, so I've replaced it with a more modern package.

package.json
```diff
@@ -48,8 +48,8 @@
     "build": "npm run rollup && npm run uglify",
     "lint": "eslint .",
     "lintfix": "npm run lint -- --fix",
-    "rollup": "rollup -c build/rollup-config.js",
-    "watch": "rollup -w -c build/rollup-config.js",
+    "rollup": "rollup -c build/rollup-config.mjs",
```
jonkoops, line 51: Since we now load the config as ESM, we need to use the `.mjs` extension. Once we have moved to a pure ESM package this can be restored.

mourner (approved):

PR #8587, opened 2022-10-16
Enable `prefer-template` linting rule and fix issues

jonkoops:


Falke-Design (approved):

mourner:
👍 on the PR but I'm just wondering — why is Bundlemon saying there was no change in size? There are some code changes so it would be unlikely to have zero difference in both pure and gzipped...

jonkoops:
Yeah, that is very weird. Unless we are really really close to the same byte size.

PR #8590, opened 2022-10-17
Added option for setting value of `aria-label` for popup close button

ShivangMishra:
Added an option for setting value of aria-label for popup close button in Popup.js. The default aria-label is 'Close Popup'. 
Please accept cc @jonkoops

### DIFF
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -49,6 +49,23 @@ describe('Popup', () => {
 		expect(map.hasLayer(popup)).to.be(true);
 	});
 
+	it('sets the default \'closeButtonLabel\' on the close button', () => {
+		const popup = L.popup()
+			.setLatLng(center)
+			.openOn(map);
+
+		expect(popup.getElement().querySelector('[aria-label="Close popup"]')).not.to.be(null);
+	});
+
+	it('sets a custom \'closeButtonLabel\' on the close button', () => {
+		const closeButtonLabel = 'TestLabel';
+		const popup = L.popup({closeButtonLabel})
+			.setLatLng(center)
+			.openOn(map);
+
+		expect(popup.getElement().querySelector(`[aria-label="${closeButtonLabel}"]`)).not.to.be(null);
+	});
+
 	it('toggles its visibility when marker is clicked', () => {
 		const marker = L.marker(center);
 		map.addLayer(marker);
--- a/src/layer/Popup.js
+++ b/src/layer/Popup.js
@@ -97,6 +97,10 @@ export const Popup = DivOverlay.extend({
 		// Controls the presence of a close button in the popup.
 		closeButton: true,
 
+		// @option closeButtonLabel: String = 'Close popup'
+		// Specifies the 'aria-label' attribute of the close button.
+		closeButtonLabel: 'Close popup',
+
 		// @option autoClose: Boolean = true
 		// Set it to `false` if you want to override the default behavior of
 		// the popup closing when another popup is opened.
@@ -206,7 +210,8 @@ export const Popup = DivOverlay.extend({
 		if (this.options.closeButton) {
 			const closeButton = this._closeButton = DomUtil.create('a', `${prefix}-close-button`, container);
 			closeButton.setAttribute('role', 'button'); // overrides the implicit role=link of <a> elements #7399
-			closeButton.setAttribute('aria-label', 'Close popup');
+			closeButton.setAttribute('aria-label', this.options.closeButtonLabel);
+
 			closeButton.href = '#close';
 			closeButton.innerHTML = '<span aria-hidden="true">&#215;</span>';
 


src/layer/Popup.js
```diff
@@ -97,6 +97,10 @@ export var Popup = DivOverlay.extend({
 		// Controls the presence of a close button in the popup.
 		closeButton: true,
 
+		//@option closeButtonAriaLabel: String = 'Close popup'
```
IvanSanchez, line 100: For consistency: This docstring is lacking a space between `//` and `@`.

IvanSanchez (changes requested):

src/layer/Popup.js
```diff
@@ -97,6 +97,10 @@ export var Popup = DivOverlay.extend({
 		// Controls the presence of a close button in the popup.
 		closeButton: true,
 
+		// @option closeButtonAriaLabel: String = 'Close popup'
+		// specifies aria-label attribute of the close button
```
jonkoops, line 101:
```suggestion
		// Specifies the 'aria-label' attribute of the close button.

```

jonkoops (changes requested):
Can you add some tests to verify this is working as expected?

src/layer/Popup.js
```diff
@@ -98,7 +98,8 @@ export var Popup = DivOverlay.extend({
 		closeButton: true,
 
 		// @option closeButtonAriaLabel: String = 'Close popup'
-		// specifies aria-label attribute of the close button
+		// Specifies the 'aria-label' attribute of the close button.
+
```
jonkoops, line 102:
```suggestion

```

ShivangMishra:
> Can you add some tests to verify this is working as expected?

I've added a test. Please review.

ShivangMishra:
> > Can you add some tests to verify this is working as expected?
> 
> I've added a test. Please review. 

@jonkoops can you accept it?

src/layer/Popup.js
```diff
@@ -97,6 +97,10 @@ export var Popup = DivOverlay.extend({
 		// Controls the presence of a close button in the popup.
 		closeButton: true,
 
+		// @option closeButtonAriaLabel: String = 'Close popup'
+		// Specifies the 'aria-label' attribute of the close button.
+		closeButtonAriaLabel: 'Close popup',
```
jonkoops, line 102: Can we rename this to `closeButtonLabel`? ARIA is a technical detail, the important thing is that there is an accessible label.

jonkoops (changes requested):
Sorry for the delayed review, I've been swamped with other work recently.

src/layer/Popup.js
Malvoz: Also not sure if "Button" should be in the name, the zoom controls don't use it:

<html>
<body>


zoomInText | String | &lt;span aria-hidden="true">+&lt;/span> | The text set on the 'zoom in' button.
-- | -- | -- | --
zoomInTitle | String | 'Zoom in' | The title set on the 'zoom in' button.
zoomOutText | String | &lt;span aria-hidden="true">&#x2212;&lt;/span> | The text set on the 'zoom out' button.
zoomOutTitle | String | 'Zoom out' | The title set on the 'zoom out' button.


</body>
</html>

And maybe the name should be more descriptive about which close button it is, for consistency with the zoom controls? `PopupCloseLabel` perhaps?

jonkoops: I think I'd still prefer `closeButtonLabel` as there is already an option that is called `closeButton`, which controls the visibility of the button itself. Given the context is initializing the `Popup` I would argue that is clear enough.

jonkoops: That should be all as far as I know. Thanks @ShivangMishra

ShivangMishra:
@jonkoops  please review and approve

spec/suites/layer/PopupSpec.js
```diff
@@ -293,7 +306,7 @@ describe('Popup', function () {
 			expect(offsetTop - offset.y).to.eql(defaultTop);
 		});
 	});
-
+
```
jonkoops, line 309: Can we get rid of these unrelated white-space changes?
```suggestion
```

src/layer/Popup.js
```diff
@@ -97,6 +97,10 @@ export var Popup = DivOverlay.extend({
 		// Controls the presence of a close button in the popup.
 		closeButton: true,
 
+		// @option closeButtonLabel: String = 'Close popup'
+		// Specifies the 'aria-label' attribute of the close button.
```
jonkoops, line 101: Let's use backticks here for the attribute name, it will get special formatting in the docs.
```suggestion
		// Specifies the `aria-label` attribute of the close button.

```

jonkoops (changes requested):

jonkoops (approved):

jonkoops:
@ShivangMishra could you rebase your PR on `main`? There seem to be some merge conflicts preventing us from landing this.

jonkoops (approved):

ShivangMishra:
@IvanSanchez  @Malvoz do I need to make any other changes?

jonkoops:
Fixed up the tests and everything looks good, I am merging this one in.

jonkoops:
Thanks for you contribution @ShivangMishra!

ShivangMishra:
> Thanks for you contribution @ShivangMishra!

You're welcome @jonkoops

PR #8592, opened 2022-10-18
Enable `object-shorthand` linting rule and fix issues

jonkoops:


IvanSanchez (approved):
This is https://caniuse.com/mdn-javascript_grammar_shorthand_object_literals , right? Anything post-2016 seems to support it natively. I'm OK with the change.

Falke-Design (approved):
I appreciate that change! Conflict with LayerGroup

jonkoops:
> This is https://caniuse.com/mdn-javascript_grammar_shorthand_object_literals , right? Anything post-2016 seems to support it natively. I'm OK with the change.

Yeah, it is. I think we can safely target ES2020, perhaps even ES2021.

PR #8593, opened 2022-10-18
Enable `prefer-rest-params` linting rule and fix issues

jonkoops:


IvanSanchez (approved):
https://caniuse.com/rest-parameters

Seems supported across the board since 2016. I'm OK with this.

Issue #8598, opened 2022-10-19
Code smell - Lint configuration for unused function params


JohnnyMoonlight:
### Checklist

- [ ] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

Not a bug, more a linting/readability issue: function calls with unnecessary parameters are allowed.

See this example:
GridLayer.js calls in line [399](https://github.com/Leaflet/Leaflet/blob/2daebbb4390b8e52c0b8c7bfc2074fe577f8bf73/src/layer/tile/GridLayer.js#L396) falseFn() with a parameter:
`Util.falseFn(level.el.offsetWidth);`

falseFn() das not expect a parameter Util.js, line [92](https://github.com/Leaflet/Leaflet/blob/2daebbb4390b8e52c0b8c7bfc2074fe577f8bf73/src/layer/tile/GridLayer.js#L396).
`export function falseFn() { return false; }`

The code works, but for clarity the usage of unnecessary parameters should be restricted through linting rules.

### Expected behavior

eslint should be configured to warn about function calls with unnecessary parameters.

### Current behavior

eslint ignores unnecessary parameteres in function calls.

### Environment

- Leaflet version 1.9.2

mourner:
This code is written like this intentionally. We need to read `offsetWidth` to force relayout, without using the actual value and avoid any related linting rules or treeshaking removal.

PR #8601, opened 2022-10-20
Linting: clarify reason for browser hack

IvanSanchez:
Clarifies why there's seemingly useless unlinted code. Should appease #8598.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -392,8 +392,8 @@ export var GridLayer = Layer.extend({
 
 			this._setZoomTransform(level, map.getCenter(), map.getZoom());
 
-			// force the browser to consider the newly added element for transition
-			Util.falseFn(level.el.offsetWidth);
+			// force reading offsetWidth so the browser considers the newly added element for transition
+			Util.falseFn(level.el.offsetWidth);	// eslint-disable-line
 
 			this._onCreateLevel(level);
 		}
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -393,7 +393,7 @@ export var GridLayer = Layer.extend({
 			this._setZoomTransform(level, map.getCenter(), map.getZoom());
 
 			// force reading offsetWidth so the browser considers the newly added element for transition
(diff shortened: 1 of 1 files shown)


mourner (approved):
Should there be a `eslint-disable-line` if ESLint passes without it? I quickly browsed the list of all ESLint rules and there seems to be no rule about passing superfluous parameters.

IvanSanchez:
Perhaps we want to enable https://eslint.org/docs/latest/rules/no-unused-vars as well?

IvanSanchez:
cc @jonkoops

Falke-Design (approved):

Falke-Design:
> Perhaps we want to enable https://eslint.org/docs/latest/rules/no-unused-vars as well?

+1

jonkoops:
> Perhaps we want to enable https://eslint.org/docs/latest/rules/no-unused-vars as well?

Is this not enabled already? We're extending the [`eslint:reccomended`](https://github.com/mourner/eslint-config-mourner/blob/master/index.js#L7) config, and it should be enabled there.

src/layer/tile/GridLayer.js
```diff
@@ -392,8 +392,8 @@ export var GridLayer = Layer.extend({
 
 			this._setZoomTransform(level, map.getCenter(), map.getZoom());
 
-			// force the browser to consider the newly added element for transition
-			Util.falseFn(level.el.offsetWidth);
+			// force reading offsetWidth so the browser considers the newly added element for transition
+			Util.falseFn(level.el.offsetWidth);	// eslint-disable-line
```
jonkoops, line 396: Does the linting pass without this? Because if it does I'd rather not disable the linter here.
```suggestion
			Util.falseFn(level.el.offsetWidth);
```

jonkoops (changes requested):

src/layer/tile/GridLayer.js
JohnnyMoonlight: It passes without the comment.

jonkoops (approved):
Looks like this also works fine without disabling ESLint, so LGTM.

PR #8602, opened 2022-10-20
Enable `no-var` linting rule and fix issues

jonkoops:


Falke-Design:
Looks good, but I had no good experiences with changing var to const / let automatically. I always had problems afterward. We have already a lot of tests, but I think it would be good to do some manual tests too. I can do some in the next days.

A last thing to consider: when we merge this, I think the most of PRs getting conflicts. Maybe we don't need to enable it now.

IvanSanchez:
I agree with @Falke-Design , with the caveat that we're gonna break up all PRs sooner or later.

jonkoops:
I think breaking existing PRs is fine. It's a small change for PR authors to resolve any conflicts.

As for testing this manually, if you want to take this on @Falke-Design feel free to do so. IMHO I think that any errors would likely already have been caught by CI.

Falke-Design:
I'm still not sure if we should do this now, but I will not be the one against it.

jonkoops:
Anyone got objections against merging this in? If not I'd like to merge this in the next couple of hours.

Issue #7200, opened 2020-07-07
Prefer using PointerEvents


IvanSanchez:
TODO: Finish the [work on the `pointerify` branch](https://github.com/Leaflet/Leaflet/commit/3151aadb17a64131ecaa533a1f328c6a41339ff1), now that [the majority (~90%) of browsers support `PointerEvent`s](https://caniuse.com/#search=pointerevent):

- [ ] Make `DomEvent.on`/`off` display a deprecation warning when listening to a `MouseEvent` or `TouchEvent`.
- [ ] Rebase the [code](https://github.com/Leaflet/Leaflet/commit/3151aadb17a64131ecaa533a1f328c6a41339ff1) that polyfills the `DomEvent` stuff.
- [ ] Change all internal references to `MouseEvent`/`TouchEvent`s into `PointerEvent`s (e.g. turn `this._container.on('mousedown', ...` into `this._container.on('pointerdown', ...`)
- [ ] Ensure documentation reflects the changes (e.g. remove https://leafletjs.com/reference-1.6.0.html#mouseevent *et al*)

*Hopefully* this should remove/simplify the legacy logic for touch events. This is not tied to a particular issue, but things like #7027, #7199, #7097 should benefit from it

johnd0e:
The idea is tempting, but I have some concerns.

1. Besides Leaflet itself we have **a lot of** 'legacy' plugins code.
Are you proposing to continue support Touch- and Mouse- by implementing internal polyfills?
If so, I'm not sure that this is right way, as it would be equally error-prone as it is now.
Instead, Leaflet could left this 'untouched': use Pointer- internally, but let all 3rd-party code to prefer any they want. 

2. There are some cases where Pointer-api does not correspond to Touch-api directly.
E.g. we can easily prevent 'click' or 'contextmenu' in 'touchend' listener, but it's not possible in 'pointerup' listener.

It's just thoughts, as I do not have ultimate solution for all current issues.
But you may look to my own approach in this [project](https://github.com/Leaflet/Leaflet/projects/1), where I try to simplify the things step by step.

IvanSanchez:
> Are you proposing to continue support Touch- and Mouse- by implementing internal polyfills?

[...]

PR #8603, opened 2022-10-20
Remove Microsoft-specific Pointer Events from browser detection

jonkoops:
Removes Microsoft-specific Pointer Events from the browser detection code, and removes all related code based on this. This removes the `L.Browser.msPointer` API.

### DIFF
--- a/src/dom/DomEvent.Pointer.js
+++ b/src/dom/DomEvent.Pointer.js
@@ -1,15 +1,9 @@
-import * as DomEvent from './DomEvent';
-import Browser from '../core/Browser';
 import {falseFn} from '../core/Util';
 
-/*
- * Extends L.DomEvent to provide touch support for Internet Explorer and Windows-based devices.
- */
-
-var POINTER_DOWN =   Browser.msPointer ? 'MSPointerDown'   : 'pointerdown';
-var POINTER_MOVE =   Browser.msPointer ? 'MSPointerMove'   : 'pointermove';
-var POINTER_UP =     Browser.msPointer ? 'MSPointerUp'     : 'pointerup';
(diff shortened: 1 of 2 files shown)


src/dom/DomEvent.Pointer.js
```diff
@@ -1,15 +1,9 @@
...
 
-/*
- * Extends L.DomEvent to provide touch support for Internet Explorer and Windows-based devices.
- */
-
-var POINTER_DOWN =   Browser.msPointer ? 'MSPointerDown'   : 'pointerdown';
-var POINTER_MOVE =   Browser.msPointer ? 'MSPointerMove'   : 'pointermove';
-var POINTER_UP =     Browser.msPointer ? 'MSPointerUp'     : 'pointerup';
-var POINTER_CANCEL = Browser.msPointer ? 'MSPointerCancel' : 'pointercancel';
+var POINTER_DOWN =   'pointerdown';
```
IvanSanchez, line 3: There's no point on leaving the uppercase "constants" defined like this. The values should be refactored into the code IMO: replace all occurences of `POINTER_DOWN` with `'pointerdown`, etc.

IvanSanchez (changes requested):

IvanSanchez:
This is good as a cleanup, but in the med-long term I'd love to go back to https://github.com/Leaflet/Leaflet/issues/7200 , and only provide support for `'pointer*'` events, dropping both `'touch*'` and `'mouse*'` events altogether.

jonkoops:
@IvanSanchez Yes, agreed! I've pulled this into the milestone for 2.0. Perhaps we can get something done in this regard.

src/dom/DomEvent.Pointer.js
jonkoops: Done :+1:

IvanSanchez (approved):

PR #8604, opened 2022-10-20
Remove Karma launcher for Microsoft Edge

jonkoops:
We no longer support 'old' Edge, so let's remove it from the test suite.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,6 @@
         "husky": "^8.0.1",
         "karma": "^6.4.1",
         "karma-chrome-launcher": "^3.1.1",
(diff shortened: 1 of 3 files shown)


mourner (approved):

Falke-Design:
Then we need to remove Edge from checks in the code too.

https://github.com/Leaflet/Leaflet/blob/d777c79f7ad582c5928fe3e5e8c86d1782a15cf6/src/dom/DomEvent.js#L281

jonkoops:
@Falke-Design Yeah, working on that in a separate PRs. I'm splitting out the work to keep things easily reviewable.

PR #8607, opened 2022-10-20
Remove Android from browser detection

jonkoops:
Removes Android from the browser detection code and removes all related code based on these checks. This removes the following APIs:

- `L.Browser.android`
- `L.Browser.android23`
- `L.Browser.androidStock`

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -10,8 +10,8 @@ import {svgCreate} from '../layer/vector/SVG.Util';
  * @example
  *
  * ```js
- * if (L.Browser.android23) {
- *   alert('Upgrade your browser, dude!');
+ * if (L.Browser.chrome) {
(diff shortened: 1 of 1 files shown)


Falke-Design (approved):
Looks good.

Maybe we want to keep the browser detection because it is maybe used in some plugins. Would like to have the opinion from the others. @mourner @IvanSanchez @Malvoz

jonkoops:
I think if we are going to keep any browser detection then it should be around our supported browsers. Or even better, the browser engine. For example, `L.Browser.chrome` could become `L.Browser.blink` and `L.Browser.firefox` could be `L.Browser.gecko`.

This is mostly because there is no real difference between browsers such as Chrome, Chromium, Opera, Samsung Internet, Edge, etc. They are all just shipping different versions of the same engine.

Issue #8608, opened 2022-10-20
Superclass for L.Renderer


IvanSanchez:
This is something I've commented on already, but I cannot find an explicit issue for.

I've created a fair share of plugins - and a few of them, involving WebGL, spawn a `<canvas>` that covers the entire screen surface of the map container. And in order to sync changes to the map state, I have [code like this](https://gitlab.com/IvanSanchez/Leaflet.GLMarkers/-/blob/master/src/GLMarkers.js#L90-L125) repeated across the plugins:

```
	// In order to make the zoom animation look decent, copy most of the
	// L.Renderer functionality for the update logic:
	getEvents: L.Renderer.prototype.getEvents,
	_onAnimZoom(ev) { this._updateTransform(ev.center, ev.zoom); this._redraw(); },
	_onZoom(ev) { this._update(); this._redraw(); },
	_onZoomEnd: L.Util.falseFn,
	_updateTransform: L.Renderer.prototype._updateTransform,
	_update: function _update() {
		/* snip */
	},
	_redraw: function _redraw(ev) { return this.render(); },
	_reset: function () { this._update(); this._updateTransform(this._center, this._zoom); },
```

Other plugins, like the D3 one or the PixiJS one, are forced to implement similar logic.

In order to simplify code for plugin authors, I'd like to create an abstract superclass of `L.Renderer` - its only purpose is to handle a single `HTMLElement`, ensure that said element is as big as the map canvas, and trigger some kind of `redraw()` (implemented in the concrete child classes) on `zoomend` but not during a zoom animation (just like `L.Renderer`).

jonkoops:
Sounds like a decent abstraction to me, let's do it!

PR #8611, opened 2022-10-20
Implement BlanketOverlay as Renderer superclass

IvanSanchez:
Splits off the update-triggering logic from L.Renderer into a new L.BlanketOverlay class. L.Renderer now only holds logic for keeping a list of L.Paths.

Intended for #8608.

This is a quick&dirty implementation without major testing. At first glance it doesn't seem to break anything. But there's pending stuff:

- [x] Should there be a hook point for the `animzoom` and `zoom` events? `L.Canvas`/`L.SVG` doesn't use it (since redraws only happen at `zoomend`/`viewreset`), but WebGL-based stuff will want to trigger a redraw.
- [x] Should there be a further abstraction of `reset`/`update` concepts? The current implementation has some "unconventional" event handlers and code paths to cover browser inconsistencies when it comes to `<canvas>` synchronization, and I don't want to break any of that.
- [x] Cleanup of event handlers? Subclasses (including `L.Renderer`) should be responsible for attaching event handlers to `zoomend`.

### DIFF
--- a/debug/vector/blanket.html
+++ b/debug/vector/blanket.html
@@ -13,7 +13,8 @@
 	<div id="map"></div>
 
 	<script type="module">
-		import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../src/Leaflet.js';
+		// import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../src/Leaflet.js';
+		import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../dist/leaflet-src.esm.js';
 
 		const osmUrl = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
 			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
@@ -31,8 +32,8 @@
 				container.style.border = '2px solid red';
 
 				container.style.display= "flex";
-  container.style.justifyContent = "center";
-  container.style.alignItems = "center";
+				container.style.justifyContent = "center";
+				container.style.alignItems = "center";
 			},
 
 			_onSettled(ev){
--- a/debug/vector/blanket.html
+++ b/debug/vector/blanket.html
@@ -13,14 +13,13 @@
 	<div id="map"></div>
 
 	<script type="module">
-		// import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../src/Leaflet.js';
 		import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../dist/leaflet-src.esm.js';
 
 		const osmUrl = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
 			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
 			osm = tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
 
-		const map = window.map = new Map('map', {
+		const map = new Map('map', {
 			center: new LatLng(0,0),
 			zoom: 1,
 			layers: [osm]
--- a/debug/vector/blanket.html
+++ b/debug/vector/blanket.html
@@ -1,0 +1,61 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+
+	<meta name="viewport" content="initial-scale=1.0, user-scalable=no" />
+
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+
+	<link rel="stylesheet" href="../css/mobile.css" />
+</head>
+<body>
+	<div id="map"></div>
+
+	<script type="module">
+		import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../src/Leaflet.js';
+
+		const osmUrl = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
+
+		const map = window.map = new Map('map', {
+			center: new LatLng(0,0),
+			zoom: 1,
+			layers: [osm]
+		});
+
+		const DebugBlanket = BlanketOverlay.extend({
+			_initContainer(){
+				const container = this._container = document.createElement('div');
+				container.style.border = '2px solid red';
+
+				container.style.display= "flex";
+  container.style.justifyContent = "center";
+  container.style.alignItems = "center";
+			},
+
+			_onSettled(ev){
+				this._container.innerHTML = `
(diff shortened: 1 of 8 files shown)


src/layer/vector/Renderer.js
```diff
@@ -113,16 +54,15 @@ export const Renderer = Layer.extend({
...
+	_onViewReset() {
+		for (const id in this._layers) {
+			this._layers[id]._reset();
+		}
+	},
 
-		this._bounds = new Bounds(min, min.add(size.multiplyBy(1 + p * 2)).round());
+	_onSettled() {
+		// Subclasses are responsible of firing the 'update' event.
+		this._update();
```
Falke-Design, line 65: The function `_update()` doesn't exist in Renderer, maybe you want to add a false-function?

IvanSanchez: The comment from https://github.com/Leaflet/Leaflet/pull/8611/commits/931d98c974c0d0a7a39d772c50ee04552655f0c0 should clarify this. Basically I want to maintain the code from L.Canvas and L.SVG unchanged, and at the same time `_update` feels ambiguous.

IvanSanchez:
I'm quite happy with the PR as it stands now. I've added a new example at `debug/vector/blanket.html` illustrating how a `<div>` `BlanketOverlay` updates.

src/layer/vector/Renderer.js
Falke-Design: Some plugins are extending from `Renderer` so we should have no unimplemented methods, which are called. Currently it is more hacky then a clean solution with `_update()`.
We should just add `_update: Util.falseFn,`

src/layer/vector/SVG.js
```diff
@@ -48,29 +47,31 @@ export const SVG = Renderer.extend({
...
+			this._container.setAttribute('width', size.x);
+			this._container.setAttribute('height', size.y);
+		}
+	},
+
 	_update() {
 		if (this._map._animatingZoom && this._bounds) { return; }
 
-		Renderer.prototype._update.call(this);
+		// Renderer.prototype._update.call(this);
```
Falke-Design, line 69: Keeping the comment?

src/layer/BlanketOverlay.js
```diff
@@ -0,0 +1,165 @@
...
+export const BlanketOverlay = Layer.extend({
+	// @section
+	// @aka BlanketOverlay options
+	options: {
+		// @option padding: Number = 0.1
+		// How much to extend the clip area around the map view (relative to its size)
+		// e.g. 0.1 would be 10% of map view in each direction
+		padding: 0.1,
+
+		// @option continuous: Boolean = false
```
IvanSanchez, line 28: This covers the use case of plugins that want to update during pan animations or pinch interactions.

It won't work on `L.Canvas` nor `L.SVG` because `L.Path`s depend a bit too much on the "origin pixel" - I've tried having a look at when those origin-pixel-relative coordinates get updated, but it's a bit too convoluted and there's no quick fix.

src/layer/vector/Renderer.js
IvanSanchez: You have a good point. I've added that.

src/layer/vector/SVG.js
```diff
@@ -48,31 +47,30 @@ export const SVG = Renderer.extend({
...
 
-		// set size of svg-container if changed
-		if (!this._svgSize || !this._svgSize.equals(size)) {
-			this._svgSize = size;
-			container.setAttribute('width', size.x);
-			container.setAttribute('height', size.y);
-		}
-
 		// movement: update container viewBox so that we don't have to change coordinates of individual layers
-		DomUtil.setPosition(container, b.min);
```
IvanSanchez, line 75: This call (and the `L.Canvas` equivalent) is redundant - the `_updateTransform` method from `L.Renderer` (now moved to `BlanketOverlay`) takes care of that.

debug/vector/blanket.html
```diff
@@ -0,0 +1,71 @@
...
+<body>
+	<div id="map"></div>
+
+	<script type="module">
+		// import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../src/Leaflet.js';
+		import {tileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../dist/leaflet-src.esm.js';
+
+		const osmUrl = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
+			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
+			osm = tileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});
```
Falke-Design, line 21:
```suggestion
		import {TileLayer, Map, LatLng, BlanketOverlay, Browser, Canvas, SVG, CircleMarker} from '../../dist/leaflet-src.esm.js';

		const osmUrl = 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
			osmAttrib = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
			osm = TileLayer(osmUrl, {maxZoom: 18, attribution: osmAttrib});

```

debug/vector/blanket.html
```diff
@@ -0,0 +1,71 @@
...
+
+		const map = window.map = new Map('map', {
```
Falke-Design, line 23: do we want to keep `window.map`?

src/layer/BlanketOverlay.js
```diff
@@ -0,0 +1,165 @@
...
+	 * any map state change (including *during* pan/zoom animations).
+	 */
+	_initContainer() {
+		this._container = DomUtil.create('div');
+	},
+	_destroyContainer() {
+		DomEvent.off(this._container);
+		this._container.remove();
+		delete this._container;
+	},
```
Falke-Design, line 154: maybe adding a check if `this._container` exists. 
What if the user does `BlanketOverlay.removeFrom(map)` without adding it to the map? I'm not sure if we then call `onRemove`

Falke-Design (changes requested):
Besides that few things, this PR is ready to merge for me.

@IvanSanchez only to make sure, do this now cover all your usecases which are in #8608 described?

debug/vector/blanket.html
IvanSanchez: Oh, that was meant for debugging (unlike non-module scripts, variables defined don't become globals and cannot be referenced in the console). Will delete.

debug/vector/blanket.html
IvanSanchez: Ugh, the `import {version} from 'package.json'` was preventing me from making that work right :-/

src/layer/BlanketOverlay.js
IvanSanchez: The existing implementation didn't do a sanity check on `this._container`, so for now my stance will be "let's keep the existing behaviour".

IvanSanchez:
> @IvanSanchez only to make sure, do this now cover all your usecases which are in #8608 described?

Yes, this covers the use cases nicely. Implementing a whole-map plugin should now be simpler, since the plugin only has one entry point (`_onSettled`). The position/sizeof the container is managed by `BlanketOverlay` in a consistent manner.

There are two *future* improvements:
- Make `L.Canvas` and `L.SVG` able to work with `continuous=true`. The code for the local origin of coordinates of `L.Path`s is convoluted, and is meant to work only at integer zoom levels AFAICS.
- Handle CSS-driven zoom animations. Trying to track them (by means of timers and `requestAnimFrame`) is a headache.

Falke-Design (approved):

PR #8613, opened 2022-10-21
GridLayer construct url with integer {z} for fractional zoom

raychanks:
Round off the fractional zoom level in order to use an integer value for requesting tiles. Such as `<baseurl>/3/2/4.png` instead of `<baseurl>/3.3/2/4.png` (notice the value `3` vs `3.3`).

This PR contains the same logical changes as #8580, but without the modifications from auto-formatting for easier review.

Closes https://github.com/Leaflet/Leaflet/issues/8276


Falke-Design:
@mourner do you think this change can cause some problems? We would like to add it to the `v1` release

spec/suites/layer/tile/TileLayerSpec.js
```diff
@@ -292,6 +292,26 @@ describe('TileLayer', () => {
 			map.setView([0, 0], 2);
 		});
 
+		it("requests tiles with an integeral {z} when the map's zoom level is fractional", () => {
```
IvanSanchez, line 295: Typo: should be `integer`.

src/layer/tile/GridLayer.js
```diff
@@ -228,7 +228,7 @@ export const GridLayer = Layer.extend({
 	redraw() {
 		if (this._map) {
 			this._removeAllTiles();
-			const tileZoom = this._clampZoom(this._map.getZoom());
+			const tileZoom = Math.round(this._clampZoom(this._map.getZoom()));
```
IvanSanchez, line 231: There's gonna be an edge case when the `GridLayer`'s `minZoom` or `maxZoom` are fractional as well - the `tileZoom` might fall outside of the clamp.

I don't think this will be a problem since I don't know any case where people use fractional `min`/`maxZoom`. Perhaps a note in the docstrings for those options would be enough?

IvanSanchez (changes requested):

src/layer/tile/GridLayer.js
raychanks: Hmm, not familiar with the use cases, but that's a good catch. If there is a case where `minNativeZoom` is `2.2` and `maxNativeZoom` is `2.9`, then the `tileZoom` will always be out of bound. I will wait for the team to decide on this one.

anton-seaice: As the examples are from tileLayer, and there could be plugins where people redraw differently at fractional zoom levels, can we just change L.TileLayer and leave gridLayer as is? (Then there is no issue if plugins want to use fractional zoom levels)

Falke-Design:
I removed `Math.round` in `GridLayer` and added it to `_clampZoom` in `TileLayer`. 

> There's gonna be an edge case when the GridLayer's minZoom or maxZoom are fractional as well - the tileZoom might fall outside of the clamp.

If this would be the case the user can overwrite the _clampZoom function in TileLayer

> Hmm, not familiar with the use cases, but that's a good catch. If there is a case where minNativeZoom is 2.2 and maxNativeZoom is 2.9, then the tileZoom will always be out of bound. I will wait for the team to decide on this one.

> As the examples are from tileLayer, and there could be plugins where people redraw differently at fractional zoom levels

For this reasons, I moved it to TileLayer

IvanSanchez (approved):
Looks good to me. I don't see any major reasons to worry about edge cases.

PR #8621, opened 2022-10-24
Remove Opera from browser detection

jonkoops:
Removes Opera from the browser detection code and removes all related code based on these checks. This removes the following APIs:

- `L.Browser.opera`
- `L.Browser.opera12`
- `L.Browser.mobileOpera`

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -22,23 +22,16 @@ const style = document.documentElement.style;
 // `true` for webkit-based browsers like Chrome and Safari (including mobile versions).
 const webkit = userAgentContains('webkit');
 
-// @property opera: Boolean; `true` for the Opera browser
-const opera = !!window.opera;
-
 // @property chrome: Boolean; `true` for the Chrome browser.
 const chrome = userAgentContains('chrome');
 
 // @property gecko: Boolean; `true` for gecko-based browsers like Firefox.
-const gecko = userAgentContains('gecko') && !webkit && !opera;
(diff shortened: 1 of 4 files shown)


src/core/Browser.js
```diff
@@ -51,7 +44,7 @@ const gecko3d = 'MozPerspective' in style;
 
 // @property any3d: Boolean
 // `true` for all browsers supporting CSS transforms.
-const any3d = !window.L_DISABLE_3D && (webkit3d || gecko3d) && !opera12 && !phantom;
+const any3d = !window.L_DISABLE_3D && (webkit3d || gecko3d) && !phantom;
```
Falke-Design, line 47: Out of topic but is phantom still a valid browser?

Falke-Design (approved):
Have you tested if bringToFront() works now with opera? Or is opera now chromium too?

src/core/Browser.js
Falke-Design: #8622

jonkoops:
> Have you tested if bringToFront() works now with opera? Or is opera now chromium too?

Yeah, this works as it's all Chromium based now. These checks are meant for versions of Opera that were still using the [Presto engine](https://en.wikipedia.org/wiki/Presto_(browser_engine)).

src/core/Browser.js
jonkoops: Already have a PR for removing that as well! See #8622.

PR #8629, opened 2022-10-30
W3 validator updates

hareesh-pillai:
HTML validation updates suggested by the W3 validator


IvanSanchez:
Looks like a low-effort hacktoberfest PR, TBH. Why `jitter.html`?

Malvoz:
Thanks @hareesh-pillai, though I agree with @IvanSanchez. And while I generally welcome these changes only modifying this particular file is not very helpful.

PR #8632, opened 2022-10-31
Revamp synthetic dblclick event instantiation

### LINKED ISSUE #8631: Double click on an input element that is rendered on the leaflet map
### Checklist

### Steps to reproduce

Perform a double click on the input element and move the mouse between the clicks within the input.

It might require a few tries to do it, but it should be fairly easy to reproduce

### Expected behavior

That the error does not happen

### Current behavior

An error is thrown in the console:

```
Permission denied to access property "bind"
```

Image of the error:

image

https://codesandbox.io/s/amazing-oskar-c0cxuc

### Environment

- Leaflet version: 1.8.0, 1.9.0, 1.9.1, 1.9.2
- Browser (with version): Firefox >= 105
- OS/Platform (with version): Windows 11, MacOS 12.6.1

### Extra info

- This issue does not happen on Chrome/Edge

Edit: added the `Extra info` section

IvanSanchez:
This is a ~~blind (i.e. totally untested)~~ attempt to fix #8631.

I'm just using my code pattern for initializing synthetic events (from https://gitlab.com/IvanSanchez/gleo/-/blob/master/src/Platina.mjs#L1109 ) - instead of iterating through the properties of the existing event, this creates a init dictionary with the properties defined in the DOM specs.

This might not be the best solution, though. Perhaps the "DOM event cloning" functionality should be moved to `DomEvent.js`, and be called from the dblclick code.

### DIFF
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -54,7 +54,6 @@ function makeDblclick(ev) {
 	} else {
 		newEvent = new MouseEvent('dblclick', init);
 	}
-	newEvent._simulated = true; // for debug purposes
 	return newEvent;
 }
 
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -33,6 +33,7 @@ function makeDblclick(ev) {
 		region: ev.region,
 	};
 
+	let newEvent;
 	// The `click` event received should be a PointerEvent - but some
 	// Firefox versions still use MouseEvent.
 	if (ev instanceof PointerEvent) {
@@ -49,10 +50,12 @@ function makeDblclick(ev) {
 			pointerType: ev.pointerType,
 			isPrimary: ev.isPrimary,
 		};
-		return new PointerEvent('dblclick', init);
+		newEvent = new PointerEvent('dblclick', init);
 	} else {
-		return new MouseEvent('dblclick', init);
+		newEvent = new MouseEvent('dblclick', init);
 	}
+	newEvent._simulated = true; // for debug purposes
+	return newEvent;
 }
 
 const delay = 200;
--- a/src/dom/DomEvent.DoubleTap.js
+++ b/src/dom/DomEvent.DoubleTap.js
@@ -68,14 +68,14 @@ export function addDoubleTapListener(obj, handler) {
 	// So here we rely on that fact to avoid excessive 'dblclick' simulation when not needed.
 	let last = 0,
 	    detail;
-	function simDblclick(e) {
-		if (e.detail !== 1) {
-			detail = e.detail; // keep in sync to avoid false dblclick in some cases
+	function simDblclick(ev) {
+		if (ev.detail !== 1) {
+			detail = ev.detail; // keep in sync to avoid false dblclick in some cases
 			return;
 		}
 
-		if (e.pointerType === 'mouse' ||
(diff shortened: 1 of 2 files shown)


src/dom/DomEvent.DoubleTap.js
```diff
@@ -49,10 +50,12 @@ function makeDblclick(ev) {
 			pointerType: ev.pointerType,
 			isPrimary: ev.isPrimary,
 		};
-		return new PointerEvent('dblclick', init);
+		newEvent = new PointerEvent('dblclick', init);
 	} else {
-		return new MouseEvent('dblclick', init);
+		newEvent = new MouseEvent('dblclick', init);
 	}
+	newEvent._simulated = true; // for debug purposes
```
IvanSanchez, line 57: I'm not sure that we need this anymore.

Falke-Design: My understanding is that we have this property to see in the tests from where the event is fired. Have we another way to check if it is a Leaflet created event?

src/dom/DomEvent.DoubleTap.js
```diff
@@ -65,7 +99,7 @@ export function addDoubleTapListener(obj, handler) {
 		if (now - last <= delay) {
 			detail++;
 			if (detail === 2) {
-				handler(makeDblclick(e));
+				ev.target.dispatchEvent(makeDblclick(ev));
```
Falke-Design, line 102: Are you sure that this results in the same?

IvanSanchez: It's not "the same". It's... better. Now the event handlers do receive an instance of `MouseEvent` or `PointerEvent` instead of an empty object with copied properties. So the propagation & cancellation logic is the browser's native one. In other words: synthetic `dblclick` events will propagate through the DOM.

AFAIK this couldn't be done before, as browsers didn't have consistent behaviour for `PointerEvent`s, and/or different ways of instantiating synthetic events.

src/dom/DomEvent.DoubleTap.js
IvanSanchez: No, there's not. A synthetic event dispatched from Leaflet looks and behaves exactly the same as a synthetic event dispatched from Happen, or from Prosthetic-Hand (the libraries used in the automated tests).

src/dom/DomEvent.DoubleTap.js
Falke-Design: Okay yeah I understand now. Is much more cleaner now!

src/dom/DomEvent.DoubleTap.js
Falke-Design: We don't need this anymore. The event is now fired over the browser to the target, so the event self is not simulated.

src/dom/DomEvent.DoubleTap.js
```diff
@@ -7,19 +7,53 @@ import * as DomEvent from './DomEvent';
...
+		};
+		newEvent = new PointerEvent('dblclick', init);
+	} else {
+		newEvent = new MouseEvent('dblclick', init);
 	}
-	event = newEvent;
-	newEvent.type = 'dblclick';
-	newEvent.detail = 2;
-	newEvent.isTrusted = false;
 	newEvent._simulated = true; // for debug purposes
```
Falke-Design, line 57: Let us remove this line again

Falke-Design (changes requested):
Looks good for me. 
Should we add a test for the case #8631?

Falke-Design (approved):

Issue #5165, opened 2016-12-05
Tests for _containsPoint?


nathancahill:
Is there any test coverage for the click detection in the canvas implementation of vector layers [src/layer/vector/Canvas.js#L439-L486](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Canvas.js#L439-L486)? As far as I could tell, there isn't. If that is the case, is there interest in a PR that adds testing the math/logic of those functions?

nathancahill:
I believe from looking at the code, that it does not cover cases for polygons with holes.

IvanSanchez:
More tests are always good IMHO.

tsuday:
Hi, I am new to Leaflet. Can I work on this?

Now `_containsPoint` has moved to [src/layer/vector/CircleMarker.js#L95-L97](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/CircleMarker.js#L95-L97), [src/layer/vector/Polyline.js#L296-L315](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Polyline.js#L296-L315) and [src/layer/vector/Polygon.js#L154-L176](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Polygon.js#L154-L176), but `Polyline#_containsPoint` is still not covered according to the coverage. (`CircleMarker#_containsPoint` and `Polygon#_containsPoint` seem to be covered indirectly by tests for Canvas.)

If it is OK for me to write tests, I will add direct tests for `_containsPoint `to CircleMarkerSpec, PolylineSpec and PolygonSpec.

tsuday:
@IvanSanchez Thank you for your kind comments! I will try adding tests.

I checked the specs mentioned above, and found that these specs did not use both of the renderers. These specs directly called functions of test target classes, so I will also call `_containsPoint `directly in the same way. (Also, other private functions whose name start with "\_"(underscore) are called directly in specs, so I interpret that such calling is allowed in Leaflet testing rules.)

I will be careful about the differences between two renderers when I work on other issues and when I use Leaflet.

antoniotp2:
Hi! I would like to work on this issue.

jonkoops:
Feel free to submit a PR.

PR #8638, opened 2022-11-05
added Polygon test for _containsPoint

jotang1:
Closes part of the issue #5165 
Still new to Open Sources, let me know if anything needs to be changed.


Falke-Design:
Test is not implemented for Canvas renderer

PR #8644, opened 2022-11-07
Vector Drift Test Pinch Zoom Fix

### LINKED ISSUE #8117: Cover vector drift issue with tests
https://github.com/Leaflet/Leaflet/pull/8103 fixes a long-standing issue with vector drift on pinch zoom, but a test designed for it (from https://github.com/Leaflet/Leaflet/pull/7965) fails to reflect what happens in practice for some reason (quoting @Falke-Design):

> The best would be, if you checkout my PR #7965 and look what visually happens in my test.
> 
> 1. Go into Map.TouchZoomSpec
> 2. Comment out Line 15 `removeMapContainer(map, container);`
> 3. Go to [Line 65](https://github.com/Falke-Design/Leaflet/blob/205395d41ce70f94a1d5e57e6065b786285a2ca4/spec/suites/map/handler/Map.TouchZoomSpec.js#L65) and add `.only`
> 4. Replace `fastframe` with `frame` in [line 78](https://github.com/Falke-Design/Leaflet/blob/205395d41ce70f94a1d5e57e6065b786285a2ca4/spec/suites/map/handler/Map.TouchZoomSpec.js#L78)
> 5. If you then run the tests in the browser over the `spec/index.html` you see that the rectangle is still in the center while / after zooming. With your (this) PR the rectangle is moving away while zooming.

rectangle_pinch

We need to figure out why it fails and cover this properly. Creating a separate issue because I want to leave this as a follow up and merge the fix.

stephenspol:
Corrected test to be able to get SVG size and coords correctly. Also discovered that a bug is still present with the help of https://github.com/Leaflet/Leaflet/pull/8103#issuecomment-1235646390. 

When going through the debugger, `Draggable.disabled()` was being called if `zoomAnimation` is `false`. If `zoomAnimation` is `true`, `Draggable.enabled()` is called instead and is properly fixed.

This test will be slow due to the fact of `zoomAnimation` being enabled as the bug has not be fixed when `zoomAnimation` is `false`.

Fixes #8117.

Issue #7466 most likely will need to be reopened.


misbahkhalilaz (approved):

Falke-Design (approved):
Works perfectly, thank you!

#8794 fixes the `zoomAnimation: false` bug. An additional test will be added there.

PR #8646, opened 2022-11-08
irregular channel plugin

chenjacky131:
irregular channel for ships


docs/_plugins/irregular-channel/irregular-channel.md
```diff
@@ -0,0 +1,12 @@
+---
+name: irregular-channel
+category: irregular-channel
```
Malvoz, line 3: Does this plugin warrant a new plugin category? If that's the case the category would need its own section in plugins.md, see [example](https://github.com/Leaflet/Leaflet/blob/641e91ae50a46bc04a1c6a7bc9cc20b57e449b34/docs/plugins.md?plain=1#L104-L108).

Otherwise please move the `irregular-channel.md` file into an appropriate folder and use that category instead.

Malvoz (changes requested):

Falke-Design:
Closing because of no response

Issue #8649, opened 2022-11-10
Consider exposing `HTMLImageElement.decoding` for `ImageOverlay`


denschub:
### Motivation

Hi there! We at Mozilla see a quite a few Web Compatibility reports about various applications using an animated map layer on top of Leaflet, where the animation is flickering. For example: [there's this issue on firesmoke.ca](https://github.com/webcompat/web-bugs/issues/80232).

The primary issue here is a bit of a race between showing and decoding the image. When image tags are hidden, browsers may not decode the image files to save on memory and performance, and they'll defer decoding to when the image becomes visible. There are a couple of Leaflet plugins that use hiding and showing layers to achieve an animation effect, and that might result in flickering. Not just in Firefox, this *could* affect all browsers, and is highly dependent on the systems performance and the image decoders performance...

### Suggested solution

A reliable way to solve this is to add the `decoding="sync"` attribute to the `img` tag, or to set [`HTMLImageElement.decoding`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding) to `sync` via the DOM API. This will force the browser to decode right away, which eliminates the possibility for flickering.

While I don't think setting the decoding mode to `sync` by default is a good idea, I'd love this to be exposed as an option to the `ImageOverlay`, so that libraries could set `sync` on layers they know will be animated. 

### Alternatives considered

Sites could probably workaround that by modifying the DOM themselves, but that's not a maintainable idea. Some sites we know of have monkey-patched Leaflet to always add the `sync` decoding flag, but that'll lock them out of easy Leaflet updates, and is also not something we'd consider. Browsers could improve their logic for deciding if something should be `sync` or `async`, but that's not easy, and I think that making the intentions clear on the image tag is a cleaner solution.

IvanSanchez:
Sounds simple enough - #8650 should do the trick.

@denschub Do you know if it's only `ImageOverlay`s, or are `TileLayer`s also affected by the flickering?

I'm slightly concerned about performance issues here, since there's also users that want to delay loading any images in a Leaflet map until the user scrolls down the webpage and the map container becomes visible. I'm wondering about setting up some events to check when a map is scrolled into view, but that might be a bit overkill.

jonkoops:
Perhaps we need to start thinking about an API for `ImageOverlay` that allows users to 'swap' the image. The process could involve loading in a new image element and waiting for the decode to happen before swapping the underlying image to avoid flickering.

[...]

PR #8650, opened 2022-11-10
Added 'decoding' option to ImageOverlay

IvanSanchez:
Implements #8649

### DIFF
--- a/src/layer/ImageOverlay.js
+++ b/src/layer/ImageOverlay.js
@@ -53,7 +53,14 @@ export const ImageOverlay = Layer.extend({
 
 		// @option className: String = ''
 		// A custom class name to assign to the image. Empty by default.
-		className: ''
+		className: '',
+
+		// @option decoding: String = 'auto'
+		// Tells the browser whether to decode the image in a synchronous fashion,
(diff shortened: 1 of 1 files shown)


src/layer/ImageOverlay.js
```diff
@@ -53,7 +53,14 @@ export const ImageOverlay = Layer.extend({
...
 		// A custom class name to assign to the image. Empty by default.
-		className: ''
+		className: '',
+
+		// @option syncDecoding: Boolean = true
+		// Tells the browser to [decode the image in a synchronous fashion](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding),
+		// in order to reduce image flickering under some workloads. It should be
+		// set to `false` whenever the image overlay is not expected to be
+		// removed/readded to the map.
+		syncDecoding: true
```
jonkoops, line 63: Perhaps we can use the [same values](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding#value) here as the attribute supports? That way the user could also decide to go for `auto`?

IvanSanchez:
Changed default value as per https://github.com/Leaflet/Leaflet/issues/8649#issuecomment-1310519108

Falke-Design (approved):

Malvoz (approved):

jonkoops (approved):
LGTM

Issue #8630, opened 2022-10-30
Flickering while zooming on some Android WebView with custom CSS-filter


andreandersson:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

1. Set a custom `CSS-filter` on `img`. For example: `img { filter: invert(1); }`
2. Create an Android-app which opens the map in a WebView
3. Pinch zoom

### Expected behavior

While pinch-zooming some strange flickering can be seen; and it seems like a "none-styled map" is overlayed above the map. The issue is easier to see if you have `zoomSnap` 0.0001 or something similar.

### Current behavior

The map should be pinch zoomed correctly

### Minimal example reproducing the issue

https://github.com/andreandersson/leaflet-bug/tree/main

### Environment

NOTE: This does not seem to be the case with most WebViews, but only some phones... We have been able to see it on Pixel 6 and 7 Pro. See video in the repo linked in the example reproducing the issue.

- Leaflet version: 1.9.2
- Browser (with version): Android System WebView 106.0.5249.126
- OS/Platform (with version): Android 13 (seen in 12 as well)

jonkoops:
Android WebView is not a supported browser, as we cannot reasonably reproduce any issues in it. The fact that this only occurs on a sub-set of phones as well makes if even more problematic to reproduce.

I have not been able to reproduce this in either Chrome (107.0.5304.87) or Firefox (105.0.2) by running the provided code in a regular browser.

Since the Android WebView is based on Chrome, this issue might be fixed by a future update. Either way, this is a quirk in the browser, not Leaflet.

Malvoz:
@andreandersson You can file WebView bugs at https://bugs.chromium.org/p/chromium/issues/entry?template=Webview+Bugs.

andreandersson:
Created a bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1380149&start=0
[...]

Issue #451, opened 2012-01-31
Opacity


Mides:
Layer opacity seem broken with IE9.

Issue #90, opened 2011-05-31
TileLayer opacity option in IE 8 fails to apply to elements in its container


cgalli:
With a simple example, I can't manage to get the opacity to work with TileLayer in IE8. I also suspect this is IE 7+. The filter style property does in fact get set to the tile._container element, but there might be an inheritance issue with individual IMG elements within that container not inheriting the alpha(opacity=xxx) value. The code that handles setOpacity does exactly as intended (dom/DomUtil.js), but opacity is somehow not applying to elements in that TileLayer container.

The opacity option works fine in every other browser I've tested that uses the style.opacity CSS setting.

Anyone else seeing this problem?

PR #8651, opened 2022-11-10
Stop inheriting `filter` CSS property on tiles

jonkoops:
Removes the inheritance of the `filter` property on map tiles. This property was initially added in #451 as a fix for IE6-9, however since then this property has been standardized as an entirely different API. 

This has been causing issues (#8630), and the best course of action seems to be to simply remove this code entirely.

Since users might be relying on this implicitly, I have marked this a breaking. But perhaps I am being over presumptuous (do let me know).

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -72,7 +72,6 @@
 	-webkit-tap-highlight-color: rgba(51, 181, 229, 0.4);
 }
 .leaflet-tile {
-	filter: inherit;
 	visibility: hidden;
 	}
 .leaflet-tile-loaded {


mourner (approved):

Malvoz (approved):

Falke-Design (approved):

IvanSanchez:
> This property was initially added in https://github.com/Leaflet/Leaflet/issues/451

It seems this dates back to #90. Wow.

The default value for `filter` is `none`. Having `inherit` might apply filters from containing elements into the tiles, which seems undesired, so I'm OK with the change. I'm not aware Leaflet is applying filters to tile containers at all.

IvanSanchez (approved):

PR #8665, opened 2022-11-18
Update docs for release 1.9.3

Falke-Design:
This merged into `v1` branch but a cherry pick had to be done on the `main`. Maybe someone else can do this.

### DIFF
--- a/docs/reference.html
+++ b/docs/reference.html
@@ -6,7 +6,7 @@ bodyclass: api-page
 
 <h2>Leaflet API reference</h2>
 
-<p>This reference reflects <strong>Leaflet v1.9.2</strong>. Check <a href='reference-versions.html'>this list</a> if you are using a different version of Leaflet.</p>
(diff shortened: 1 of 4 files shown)


docs/reference.html
```diff
@@ -6,7 +6,7 @@
 
 <h2>Leaflet API reference</h2>
 
-<p>This reference reflects <strong>Leaflet v1.9.2</strong>. Check <a href='reference-versions.html'>this list</a> if you are using a different version of Leaflet.</p>
+<p>This reference reflects <strong>Leaflet </strong>. Check <a href='reference-versions.html'>this list</a> if you are using a different version of Leaflet.</p>
```
Malvoz, line 9: @Falke-Design A bit late with the review, sorry. This should've been `<strong>Leaflet v1.9.3</strong>`?

Falke-Design: Good catch!
Why is this not generated automatically? ...

jonkoops: Good question, it should have been. I do recall doing all of this manually. We should really find a way to automate the entire release process.

PR #8671, opened 2022-11-18
Simplify workflows on CI

mourner:
Collapse several jobs to make the flow simpler and hopefully faster because of less job spinning-up / cache restoration & saving: 

- `setup` + `build` => `setup`
- `build-release` + `bundlemon` => `bundlemon`

With the previous approach, some jobs like `lint` could start earlier, but `npm run build` runs in just 6 seconds and `lint` isn't the bottleneck anyway, so it doesn't matter in practice.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,66 +24,16 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Cache setup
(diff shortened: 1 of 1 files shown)


mourner:
Workflows look like this now:

image

jonkoops (approved):
Trims things down a bit, LGTM.

PR #8672, opened 2022-11-19
Fix issue whereby tooltips loaded dynamically while moving the map were never shown.

### LINKED ISSUE #8591: Tooltips are not added to dom, when moving the map (or later)
Hi there,

I want to report a bug.

We work with a realtime system, that adds markers (with sometimes tooltips) on the map based on incomming data.
A user reported a bug that sometimes items didn't have labels (tooltips), so we investigated.

Turns out, that markers that have a tooltip, don't add the tooltip when you are moving the map. The marker itself gets added, but the tooltip doesn't.
I have a simple testcase here (I abuse the moveend event to add a marker):
https://jsfiddle.net/76otz9bw/5/

The following code is the problem:
```
    _openTooltip: function (e) {
        if (!this._tooltip || !this._map || (this._map.dragging && this._map.dragging.moving())) {
            console.log('  leaving cause you are dragging... solong suckers');
            return;
        }
        this._tooltip._source = e.layer || e.target;

        this.openTooltip(this._tooltip.options.sticky ? e.latlng : undefined);
    }
```

The return statements gets hit, but there is no fallback scenario. It doesn't get added later as it should.

I can imagine a fix something along the lines of:

```
    _openTooltip: function (e) {
        if (!this._tooltip || !this._map) {
            return;
        }

        if (this._map.dragging && this._map.dragging.moving())) {
            this._map.on('zoomend', ()=> {
                this._openTooltip(e);
            });
        }
        this._tooltip._source = e.layer || e.target;

        this.openTooltip(this._tooltip.options.sticky ? e.latlng : undefined);
    }
```
(i know horrible code, but you get the idea)

theGOTOguy:
Slava Ukraini, `leaflet` team!

This pull request resolves #8591, whereby `Tooltip`s were not added to the DOM while the map was moving.  

[@BasieP's analysis of the offending code](https://github.com/Leaflet/Leaflet/issues/8591#issue-1411797195) was correct, as is the [workaround suggested](https://github.com/Leaflet/Leaflet/issues/8591#issuecomment-1282942278);  However, [ I am using `react-leaflet`](https://github.com/PaulLeCam/react-leaflet/issues/1039), where this workaround is not an option.

An automatic test is included.


src/layer/Tooltip.js
```diff
@@ -411,9 +411,18 @@ Layer.include({
 
 
 	_openTooltip(e) {
-		if (!this._tooltip || !this._map || (this._map.dragging && this._map.dragging.moving())) {
+		if (!this._tooltip || !this._map) {
 			return;
 		}
+
+		// If the map is moving, we will show the tooltip after it's done.
+		if (this._map.dragging && this._map.dragging.moving()) {
```
jonkoops, line 419: I think we can make this a bit shorter using optional chaining.
```suggestion
		if (this._map.dragging?.moving()) {
```

spec/suites/layer/TooltipSpec.js
```diff
@@ -461,6 +461,31 @@ describe('Tooltip', () => {
...
+
+		// If tooltips are dynamically loaded while the map is dragging, they need
+		// to be loaded when the dragging stops.
+		const layer = L.marker(center).bindTooltip('Tooltip', {permanent: true});
+		map.addLayer(layer);
+
+		// simulate map not dragging anymore
+		map.dragging.moving = function () {
+			return false;
+		};
```
jonkoops, line 478:
```suggestion
		map.dragging.moving = () => false;
```

IvanSanchez: What happened to `L.Util.falseFn`?

jonkoops: Yeah, I was thinking we can use that as well. Why don't we have a `L.Util.trueFn`?

src/layer/Tooltip.js
theGOTOguy: Good idea.  Done.

spec/suites/layer/TooltipSpec.js
theGOTOguy: I made the suggested simplification to functions for this test, and in other `Tooltip` tests where this pattern occurs.

If you prefer that I use `L.Util.falseFn` and add an `L.Util.trueFn` and use that in these tests, just let me know and I will do that instead.

jonkoops:
Also please rebase your PR on `main` so we can merge if needed (no merge commits).

src/layer/Tooltip.js
```diff
@@ -413,9 +413,18 @@ Layer.include({
...
+			this._map.once('moveend zoomend', () => {
+				this._openTooltip(e);
+			});
```
Falke-Design, line 424: This is the right approche but everytime the tooltip wil be opend it adds a new `once` listener

src/layer/Tooltip.js
```diff
@@ -413,9 +413,18 @@ Layer.include({
...
 
 	_openTooltip(e) {
-		if (!this._tooltip || !this._map || (this._map.dragging && this._map.dragging.moving())) {
+		if (!this._tooltip || !this._map) {
 			return;
 		}
+
+		// If the map is moving, we will show the tooltip after it's done.
+		if (this._map.dragging && this._map.dragging.moving()) {
+			this._map.once('moveend zoomend', () => {
```
Falke-Design, line 422: `zoomend` is not needed. `moveend` while always fired if `zoomend` is fired too

src/layer/Tooltip.js
```diff
@@ -413,9 +413,18 @@ Layer.include({
...
+				this._openTooltip(e);
```
Falke-Design, line 423: Calling `_openTooltip` here is to early. the `dragging.moving()` function will be still set to `true`. This needs a change in `Draggable.js`, the flag needs to be set to `false` before fireing the event.

```

		if (this._moved && this._moving) {

			// @event dragend: DragEndEvent
			// Fired when the drag ends.
			this.fire('dragend', {
				noInertia,
				distance: this._newPos.distanceTo(this._startPos)
			});
		}

		this._moving = false;
		Draggable._dragging = false;
```

to:
```
		const fireDragend = this._moved && this._moving;

		this._moving = false;
		Draggable._dragging = false;

		if (fireDragend) {
			// @event dragend: DragEndEvent
			// Fired when the drag ends.
			this.fire('dragend', {
				noInertia,
				distance: this._newPos.distanceTo(this._startPos)
			});
		}
```

Falke-Design (changes requested):
The check for map dragging was added in #7862 to prevent adding / removing of the tooltip when occasionally hovering over the marker with the mouse while dragging.

src/dom/Draggable.js
```diff
@@ -202,18 +202,19 @@ export const Draggable = Evented.extend({
 		DomUtil.enableImageDrag();
 		DomUtil.enableTextSelection();
 
-		if (this._moved && this._moving) {
+		const fireDragend = this._moved && this._moving;
 
+		this._moving = false;
+		Draggable._dragging = false;
+
+		if (fireDragend) {
```
Falke-Design, line 210: @mourner @jonkoops @IvanSanchez do you see any problems with setting the flags to `false` before fireing the event?

src/layer/Tooltip.js
```diff
@@ -413,9 +413,22 @@ Layer.include({
...
+
+		// If the map is moving, we will show the tooltip after it's done.
+		if (this._map.dragging && this._map.dragging.moving()) {
+			if (!this._openOnceFlag) {
+				this._openOnceFlag = true;
+				this._map.once('moveend', () => {
+					this._openOnceFlag = false;
+					this._openTooltip(e);
+				});
+			}
```
Falke-Design, line 428: @mourner @jonkoops @IvanSanchez have someone a good Idea to make this cleaner? The `once` listener should not be added multiple times and it is important to pass `e` as argument from the original `_openTooltip` call.

theGOTOguy: Here's a couple of ideas:

Superficially, you could do something like this:

```
                // If the map is moving, we will show the tooltip after it's done.
                if (this._map.dragging && this._map.dragging.moving() && !this._openOnceFlag) {
                        this._openOnceFlag = true;
                        this._map.once('moveend', () => {
                            this._openOnceFlag = false;
                            this._openTooltip(e);
                        });
                        return;
                }
```

It's not satisfying because we still have to store `_openOnceFlag`.

However, maybe there's a better solution.  There are only four places where `_openTooltip` is called at all, and they're all here in `Tooltip.js`.

We could replace all instances of `_openTooltip` with `_openTooltipWhenNotMoving`, with:

```
_openTooltipWhenNotMoving(e) {
    if (this._map.dragging && this._map.dragging.moving()) {
        this._map.once('moveend', () => {
            this._openTooltip(e);
        });
        return;
    } else {
        this._openTooltip(e);
    }
}
```

Then remove the moving checks from `_openTooltip` proper.

In this way, we avoid storing an extra variable, and call `_openTooltip` at most once at an appropriate time when it would be needed.

theGOTOguy: I've implemented the former solution, which at least resolves a little bit of the nesting.

I don't see a better solution than the `_openOnceFlag` that you implemented, and I've convinced myself that it is correct even in case where some external code toggles `permanent`.

src/dom/Draggable.js
theGOTOguy: It seems like the only way this would matter would be if someone was looking at `_moving` or `_dragging` during the handling of a `dragend` event.  We're not doing it internally to Leaflet, but it's possible that someone who depends on us might.  Maybe mark this as a breaking API change in case someone depends on the state of `_moving` or `_dragging` in an event triggered by `dragend`?

theGOTOguy:
This is still an issue for me in production, and I'm wondering if you could take a look at the changes I've made and let me know if the solution as it sits is appropriate or if more changes are desired.

Falke-Design (approved):
Fine for me but we need an approve from another maintainer @IvanSanchez @mourner

jonkoops (approved):

PR #8678, opened 2022-11-21
Rename 'closed' parameter of 'latLngsToCoords' to 'close' to avoid confusion

ronaldhoek:
I noticed the use of the parameter name `closed` in `latLngsToCoords` when looking at the changes or PR #7344.

But I got confused on what the usage is regarding this parameter as it suggests (because on the fact it's a noun) it states something about the current 'status' of the coords (`latlngs` parameter) being passed to the function.

Therefore I think it's better to change the name to `close`...

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -22,143 +22,143 @@ import * as LineUtil from '../geometry/LineUtil';
  *
  * ```js
  * L.geoJSON(data, {
- *  style: function (feature) {
- *    return {color: feature.properties.color};
- *  }
+ * 	style: function (feature) {
+ * 		return {color: feature.properties.color};
+ * 	}
  * }).bindPopup(function (layer) {
- *  return layer.feature.properties.description;
+ * 	return layer.feature.properties.description;
  * }).addTo(map);
  * ```
  */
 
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Please revert the changes of replacing tabs with spaces.

It is named `closed` because the argument needs to be true if the shape is `closed`.

ronaldhoek:
> It is named closed because the argument needs to be true if the shape is closed.

But when I look at the function, with _shape_ you mean the `latlngs` parameter?
If so, when I look at the `closed` is interpreted, it actually closes (an action) the newly (based on the `latlngs`) created shape, so that the created shape is actually 'closed'.

So basically inside the function, a `close` action is performed on the shapedata to close it, creating a closed shape...

Therefore I suggest renaming the parameter from a 'noun' to a 'verb' as it gives a more clear understanding...

ronaldhoek:
> Please revert the changes of replacing tabs with spaces.

Sorry about the spaces, used the wrong editor.

jonkoops (approved):
This seems fine to me, let's land this.

ronaldhoek:
Thanks!

PR #8681, opened 2022-11-25
Replace `Util.create()` with `Object.create()`

jonkoops:
Removes the `Util.create()` function and replaces it with the standard [`Object.create()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) method. `Util.create()` was a polyfill, which is no longer required now that we have raised our target browsers.

This also remove `Util.create()` as an API, thus this is a breaking change.

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -18,16 +18,6 @@ export function extend(dest, ...args) {
 	return dest;
 }
 
-// @function create(proto: Object, properties?: Object): Object
-// Compatibility polyfill for [Object.create](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object/create)
-export const create = Object.create || (function () {
-	function F() {}
-	return function (proto) {
-		F.prototype = proto;
-		return new F();
-	};
-})();
-
 // @property lastId: Number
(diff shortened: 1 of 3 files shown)


src/core/Util.js
```diff
@@ -18,16 +18,6 @@ export function extend(dest, ...args) {
 	return dest;
 }
 
-// @function create(proto: Object, properties?: Object): Object
-// Compatibility polyfill for [Object.create](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object/create)
-export const create = Object.create || (function () {
```
IvanSanchez, line 23: So apparently this was one of those IE8 compatibility polyfills.

It's safe to depend on `Object.create`, but I anticipate some plugins to still depend on `L.Util.create`, and fail. Still, this had to be done.

IvanSanchez (approved):

src/core/Util.js
jonkoops: Yeah, now is the time where we can start shedding some of that weight. It's important to advocate using the standard Web APIs.

PR #8682, opened 2022-11-25
Replace `Util.trim()` with `String.prototype.trim()`

jonkoops:
Removes the `Util.trim()` function and replaces it with the standard [`String.prototype.trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) method. `Util.trim()` was a polyfill, which is no longer required now that we have raised our target browsers.

This also remove `Util.trim()` as an API, thus this is a breaking change.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -131,7 +131,7 @@ export function removeClass(el, name) {
 	if (el.classList !== undefined) {
 		el.classList.remove(name);
 	} else {
-		setClass(el, Util.trim((` ${getClass(el)} `).replace(` ${name} `, ' ')));
+		setClass(el, ` ${getClass(el)} `.replace(` ${name} `, ' ').trim());
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/dom/DomUtil.js
```diff
@@ -131,7 +131,7 @@ export function removeClass(el, name) {
 	if (el.classList !== undefined) {
 		el.classList.remove(name);
 	} else {
-		setClass(el, Util.trim((` ${getClass(el)} `).replace(` ${name} `, ' ')));
+		setClass(el, (` ${getClass(el)} `).replace(` ${name} `, ' ').trim());
```
IvanSanchez, line 134: I think the parentheses around the backticks can be removed. Does the linter not complain about this?

jonkoops: You are right, I've removed them. And no, it doesn't look like we have a linting rule configured for this (perhaps we should).

IvanSanchez (approved):

PR #8685, opened 2022-11-26
Use the `classList` API for class manipulation methods in `DomUtil`

jonkoops:
Changes the methods for class manipulation (`hasClass`, `addClass`, `removeClass`, `setClass` and `getClass`) in `DomUtil` to use the [`classList` API](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) exclusively. Includes the following changes:

- Changes the type annotation to use the `Element` class, from which both `HTMLElement` and `SVGElement` extend.
- Updates the documentation for these methods to be more unified and explicit.
- Drops Internet Explorer specific hack for getting a class name.
- Drops deprecated usage of the [`SVGElement.className` property](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement#svgelement.classname).
- Improves testing coverage by testing both `HTMLElement` and `SVGElement` and splitting out the test cases.

Although I think this is a nice cleanup, most of these methods are now light-weight wrappers around the `classList` API. So lightweight in fact that we might as well use this API directly.

Therefore, before merging this PR I would like to know from the core members if we are perhaps better off without these methods for 2.0.

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -23,35 +23,6 @@ describe('DomUtil', () => {
 		});
 	});
 
-	describe('#addClass, #removeClass, #hasClass', () => {
-		it('has defined class for test element', () => {
-			el.className = 'bar foo baz ';
-			expect(L.DomUtil.hasClass(el, 'foo')).to.be.ok();
-			expect(L.DomUtil.hasClass(el, 'bar')).to.be.ok();
-			expect(L.DomUtil.hasClass(el, 'baz')).to.be.ok();
-			expect(L.DomUtil.hasClass(el, 'boo')).to.not.be.ok();
-		});
-
-		it('adds or removes the class', () => {
-			el.className = '';
-			L.DomUtil.addClass(el, 'foo');
-
-			expect(el.className).to.eql('foo');
-			expect(L.DomUtil.hasClass(el, 'foo')).to.be.ok();
-
-			L.DomUtil.addClass(el, 'bar');
-			expect(el.className).to.eql('foo bar');
-			expect(L.DomUtil.hasClass(el, 'foo')).to.be.ok();
-
-			L.DomUtil.removeClass(el, 'foo');
-			expect(el.className).to.eql('bar');
-			expect(L.DomUtil.hasClass(el, 'foo')).to.not.be.ok();
-
-			el.className = 'foo bar barz';
-			L.DomUtil.removeClass(el, 'bar');
-			expect(el.className).to.eql('foo barz');
-		});
-	});
 	describe('#getStyle', () => {
 		it('gets the value for a certain style attribute on an element,', () => {
 			el.style.color = 'black';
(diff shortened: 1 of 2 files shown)


jonkoops:
Note that I am in favor of removing these methods entirely for version 2.0. I think we should aim to align Leaflet as much as possible with the Web API, instead of putting our own shims around it.

I was thinking that perhaps we can add a `leaflet-compat` package that holds all of these APIs we are removing for v2 so that folks can migrate over at their own pace. WDYT?

Falke-Design:
Out of my eyes as a plugin developer removing APIs are bad. They force me to change my code. On the other hand Leaflet, should not take care about Util functionality which are built-in in browsers.
Another problem with removing the Util functions are that we lose compatibility for many plugins which are not maintained anymore. 
**So I see no way around creating an additional script for migration.** For me personally I would not name it `leaflet-compat`. I think better would be the full name `leaflet-compatibility` or another name. Let us also thinking about including it as static file in the dist folder.

jonkoops:
@Falke-Design Yeah, I was thinking of it as a drop in replacement for the `leafet` import. It would import all of the existing Leaflet APIs and re-export them, plus these APIs we consider 'bad practice'. For example, if a plugin (or user) has the following code:

```ts
import { Map, DomUtil } from 'leaflet';

const element = DomUtil.create('div');
DomUtil.addClass(element, 'map-container');

const map = new Map('map', {
    center: [51.505, -0.09],
    zoom: 13
});
```

This would break as `DomUtil.create()` and `DomUtil.addClass()` would be removed. However, it could be restored to working order by importing from `leaflet/compat` instead:

```ts
import { Map, DomUtil } from 'leaflet/compat';
```

jonkoops:
Unrelated to this, I am also considering we should move all of the official Leaflet plugins into a mono-repo with our code . That way we can test them together with the Leaflet code to catch regressions and compatibility issues early on.

Falke-Design:
> I was thinking of it as a drop in replacement for the leafet import.

I would prefer to create an additional script which needs to be imported too. With this solution the user don't just uses `leaflet/compat` everywhere instead of `leaflet`. 

```
import { DomUtil } from 'leaflet/compat';

const element = DomUtil.create('div');
DomUtil.addClass(element, 'map-container');

const map = new Map('map', {
    center: [51.505, -0.09],
    zoom: 13
});
```

> Unrelated to this, I am also considering we should move all of the official Leaflet plugins into a mono-repo with our code . That way we can test them together with the Leaflet code to catch regressions and compatibility issues early on.

What are the offical Leaflet plugins for you? I don't see this compatible with the light-weight mindset of Leaflet. Also we have already enough work with Leaflet core without the plugins.

jonkoops:
> What are the offical Leaflet plugins for you? I don't see this compatible with the light-weight mindset of Leaflet. Also we have already enough work with Leaflet core without the plugins.

It's the ones under the Leaflet organization on GitHub, but this is a discussion for another time.

> I would prefer to create an additional script which needs to be imported too. 

I think a lot of folks just want their plugins to 'just work', as if it's a drop in replacement then it can be used as an alias (such as [in WebPack](https://webpack.js.org/configuration/resolve/#resolvealias)).

Falke-Design:
> It's the ones under the Leaflet organization on GitHub, but this is a discussion for another time.

But I hope you don't want to inlcude the big libraries like Leaflet-Draw or Markercluster, neither the other not maintained ones (Leaflet.Editable, Leaflet.heat, Path.Drag.js, Leaflet.toolbar). 

> I think a lot of folks just want their plugins to 'just work'

Yes that is very important. I see the benifit in own script, that we don't have to care in later release about it. We can make now a clear cut and everyone who needs the old utilities, can include the seperate script too. But it is also ok for me to make `leaflet/compat`

Falke-Design (approved):

jonkoops:
Let's merge this in for now, I think we can always have a separate discussion about how we're going to handle compatibility in the future.

mourner:
> Therefore, before merging this PR I would like to know from the core members if we are perhaps better off without these methods for 2.0.

The way I see it, when we're talking about deprecating APIs like this, there's no better chance for dropping them than a new major version (v2), so we better cram all the ones we want before the release. It's not like we can keep a 100% compatibility anyway — we dropped some methods already (e.g. those in `Util`) and didn't provide a compat wrapper for those, which breaks consistency.

Besides, plugins would likely have to be updated for v2 compatibility anyway (because of ES stuff, etc.). Ideally we would urge plugin developers to upgrade the plugins, maybe make a "v2 compatible" badge bubbling plugins to the top in the docs, and avoid having to maintain a "compatibility" layer forever, staying lean.

jonkoops:
Yeah, I agree with that. I'll see what it would take to remove these functions.

PR #8687, opened 2022-11-26
add new method setPrecision in LatLng

stefanocudini:
I sent a pull request for a small change which I think is very useful and many people
The main reason why I decided to implement it directly in the codebase and send it as PR is because every time you need this function the code needed to do it out of the library is always dirty and inconvenient and then I think it is a consistent method with what is expected from the L.Latlng object.

Opinions are welcome

thank you all
Stef

### DIFF
--- a/src/geo/LatLng.js
+++ b/src/geo/LatLng.js
@@ -91,6 +91,14 @@ LatLng.prototype = {
 		        [this.lat + latAccuracy, this.lng + lngAccuracy]);
 	},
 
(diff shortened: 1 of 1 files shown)


Falke-Design:
For what do you need to modify the precision on the `LatLng` object? I think working with exact values and setting the precision while exporting is more efficient.
But I'm open if you let me know some usecases.

jonkoops:
Closing this due to lack of reaction from the PR author. Feel free to let me know if this needs to be re-opened.

PR #8717, opened 2022-12-15
Add loading=lazy to badges in plugins list

### LINKED ISSUE #8716: Optimize star / recency labels on the plugins page
Follow up to #8714 — while it's a huge improvement, visiting the plugins page now invokes 1000+ network requests and makes the page very slow, in addition to a bunch of errors (around 60 for me). 

As a temporary workaround, we should add `loading="lazy"` to badge images, but long-term, ideally we would do away with badges and instead fetch / calculate stars and other metadata in a GitHub Action and then use it for rendering the static page. One of the advantages is instant page load, but another one would be that we can then extend the functionality of the page further, for example, adding dynamic sorting by stars or recency per category, highlight the most popular plugins, and style the rows differently depending on recency.

cc @Falke-Design @jonkoops

IvanSanchez:
Closes #8716

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -442,9 +442,11 @@ function loadRepoData() {
 					];
 				}
 
-				badges.forEach((badge) => {
-					repoData.innerHTML += `<img src="${badge}" alt=""/>`;
-				});
+				if (badges.length) {
+					repoData.innerHTML = badges
+						.map(b => `<img src="${b}" alt="" loading="lazy"/>`)
+						.join('');
+				}
 			}
 		} catch (e) {
 			console.error(e);


jonkoops:
This doesn't actually seem to work for me, it seems related to the timing at which the DOM mutations take place. When wrapping it on the `DOMContentLoaded` event things work fine (just not in Firefox).

mourner:
Is there any way we could move the regexp/badge logic from JS to the Jekyll templates so that images would be embedded in HTML from the beginning?

jonkoops:
Yeah, I would prefer that as well. I think we'd have to create a [custom filter](https://jekyllrb.com/docs/plugins/filters/) in Jekyll which matches the regex and returns the repo name (see also the [Liquid docs](https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers#create-your-own-filters)).

Falke-Design:
@jonkoops @mourner take a look into #8718

IvanSanchez:
Superseded by #8718

PR #8718, opened 2022-12-17
Create img badges directly in jekyll

### LINKED ISSUE #8716: Optimize star / recency labels on the plugins page

Falke-Design:
Attempt to create the badges directly in Jekyll.

Closes #8716, #8717

grafik


IvanSanchez (approved):
I cannot understand much Ruby (but the "piping data to a jekyll helper" bit looks good) , and my local installation of Jekyll is a bit of a mess, so I'm just gonna assume you've tested it.

mourner (approved):

mourner:
@Falke-Design should probably be reverted — it's not working. The plugins page looks like this now:

image

Falke-Design:
@mourner Done

Falke-Design:
@mourner @jonkoops it is not possible to use custom filters with `github-pages` ... Github overwrites the directory path of the custom plugin with a random string, so it is never possible to include custom plugins / filters.

grafik
https://github.com/jekyll/jekyll/issues/5265#issuecomment-241267253

jonkoops:
@Falke-Design any chance the fix mentioned in https://github.com/jekyll/jekyll/issues/5265#issuecomment-435327578 could also work for us here?

Falke-Design:
@jonkoops no doesn't work 😕 

https://falke-design.github.io/Leaflet/plugins.html
https://github.com/Falke-Design/Leaflet/tree/docs/jekyll-badges
https://github.com/Falke-Design/Leaflet/blob/docs/jekyll-badges/docs/Gemfile

jonkoops:
Ah that's a bummer indeed. It would be really nice if we could get it to work server-side.

mourner:
@jonkoops in the meantime, maybe we could simulate the lazy loading by adding the badges when scrolling in view with `IntersectionObserver`?

jonkoops:
I actually have a working version of the `loading="lazy"` functionality on my other machine. I'll see if I can pull it up once I have it up and running (moving countries atm).

PR #8724, opened 2022-12-20
Supported file types for image overlay

anton-seaice:
Added notes about supported file types for image overlay, and link to  tags documentation

### DIFF
--- a/docs/examples/overlays/index.md
+++ b/docs/examples/overlays/index.md
@@ -14,7 +14,7 @@ In this tutorial, you’ll learn how to use these overlays.
 
 ### `ImageOverlay`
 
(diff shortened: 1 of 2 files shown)


jonkoops:
Sorry, but all of the things added here seem pretty explanatory to me. I don't see the value of adding this to the documentation.

anton-seaice:
Explanatory sounds good doesn't it? Especially in the Tutorials (maybe less so in the Docs)

There are similar details provided in the doco for Video Overlay. Are their guidelines about what to include in the doco?

jonkoops:
Sorry, meant 'self-explanatory' looks like it got auto-corrected. Still don't think we need to add this or spend time on it, our backlog is big enough at the moment.

PR #8732, opened 2022-12-22
Replace `DomUtil.removeClass()` with `classList.remove()`

jonkoops:
Removes the `DomUtil.removeClass()` function and replaces it with the standard [`classList.remove()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) method. `DomUtil.removeClass()` was a polyfill, which is no longer required now that we have raised our target browsers.

This also removes `DomUtil.removeClass()` as an API, thus this is a breaking change. See #8685 for more background information.

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -146,22 +146,6 @@ describe('DomUtil', () => {
 		});
 	});
 
-	describe('#removeClass', () => {
-		it('removes the class from an HTML element', () => {
-			const element = document.createElement('div');
-			element.classList.add('newClass');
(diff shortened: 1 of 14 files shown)


jonkoops:
Looks like Bundlemon is having some issues, but it should not be affected by the changes in this PR.

Falke-Design (approved):

jonkoops:
> Looks like Bundlemon is having some issues, but it should not be affected by the changes in this PR.

@LironEr anything going on on your side? I don't quite understand why this is failing now.

LironEr:
@jonkoops looks like a rate limit issue.
I'm using Vercel to host my backend, it seems like I reached some limits there.

I will try to fix that very soon

PR #8737, opened 2022-12-22
`toGeoJSON()` should still work if no values in coords array

Falke-Design:
After adding `coords[0].slice()` in #7344 we have the problem that if `coords[0]` is `undefined` an error is thrown.

Must released in `1.9.4`

### DIFF
--- a/spec/suites/layer/GeoJSONSpec.js
+++ b/spec/suites/layer/GeoJSONSpec.js
@@ -887,6 +887,11 @@ describe('L.GeoJSON functions', () => {
 			expect(coords).to.eql([[1, 2, 3], [4, 5, 6], [1, 2, 3]]);
 			expect(coords[0] === coords[2]).to.not.ok();
 		});
+		it('still works if no values in coords array', () => {
+			expect(() => {
+				L.GeoJSON.latLngsToCoords([[]], 1, true);
+			}).to.not.throwException();
+		});
 	});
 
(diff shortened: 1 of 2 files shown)


mourner (approved):

jonkoops (approved):

Falke-Design:
Cherry pick into `v1` done: https://github.com/Leaflet/Leaflet/commit/f42cab93c30a33828ecbc753febb3247a069c8c3

jonkoops:
Does this warrant a release or should we hold off until we have gathered more bugfixes?

Falke-Design:
I would wait a month or two and then release it.

manzanotti:
I've encountered an issue with exporting layers to GeoJSON, and I'm wondering if this will fix it?

I have layers that consist of polylines. Each line is added to the layer as a separate feature. At some point in the last few months, the coordinates property of the feature has gone from being an array of coordinates, to an array of an array of coordinates.

Looking at my project, I suspect that it was when I upgraded leaflet from 1.8 to 1.9.2.

It appears that the lines created on the map before that change still export as a flat array, whereas new lines are exported with the nested array.

I have had a look back through the history of this file and polylines, and can't see what else might have caused this, and I have tripled-checked that all I'm doing in my code is calling toGeoJSON and saving that to a file, so I don't think it's something I've done.

PR #8738, opened 2022-12-22
Remove vendor prefixes when setting `userSelect` style

jonkoops:
Removes vendor prefixes for `userSelect` style when disabling or re-enabling text selection. Since this property is now supported by all major browsers it's no longer required to add a vendor prefix.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -134,21 +134,14 @@ if ('onselectstart' in document) {
 		DomEvent.off(window, 'selectstart', DomEvent.preventDefault);
 	};
 } else {
-	const userSelectProperty = testProp(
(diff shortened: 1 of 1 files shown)


Falke-Design (approved):

jonkoops:
I feel like the `disableTextSelection` and `enableTextSelection` functions could be optimized further. The only browser that does not seem to implement the `selectstart` event is Safari on iOS (see [MDN docs](url)).

Not sure if this info is up-to-date, but I have a gut feeling this could still be optimized. Perhaps by only using `user-select`, since the property is supported everywhere.

Either way, it would require further investigation I can't get into right now.

mourner:
@jonkoops I was wary about this change before because [MDN states](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) `user-select` is still prefixed on Safari (both desktop and mobile). Can you please confirm we didn't break it there by removing the prefix?

PR #8741, opened 2022-12-23
Fix implementation for disabling & enabling text selection

### LINKED ISSUE #8740: Fix `userSelect` on Safari
#8738 removed prefixing for `userSelect`, but Webkit (Safari on iOS and desktop) still has it prefixed (`WebkitUserSelect`) and doesn't support the non-prefixed version.

Let's bring it back and make sure we're more careful about this, checking support on MDN / CanIUse before making similar changes. No need to bring back #8739 — can be a simple ternary.

jonkoops:
Changes the implementation of `DomUtil.disableTextSelection()` and `DomUtil.enableTextSelection()` to exclusively use the [`user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) CSS property.

Previously this was detecting the `selectstart` event and then using `event.preventDefault()` to ensure the user was not able to select text. This approach however was flawed as Safari on iOS now supports this event, but does not prevent the user from selecting text when the event is prevented.

Also fixes a bug where it was possible to call `DomUtil.disableTextSelection()` twice in a row allowing the internal value for restoring text selection to be incorrectly set to `none`. This then caused `DomUtil.enableTextSelection()` to 'restore' this incorrect value.

Documentation for `DomUtil.disableImageDrag()` and `DomUtil.enableImageDrag()` has also been adjusted as it contained a incorrect references.

Closes #8740

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -100,45 +100,45 @@ export function getPosition(el) {
 	return el._leaflet_pos || new Point(0, 0);
 }
 
+const documentStyle = document.documentElement.style;
+// Safari still needs a vendor prefix, we need to detect with property name is supported.
+const userSelectProp = ['userSelect', 'WebkitUserSelect'].find(prop => prop in documentStyle);
+let prevUserSelect;
+
 // @function disableTextSelection()
-// Prevents the user from generating `selectstart` DOM events, usually generated
-// when the user drags the mouse through a page with text. Used internally
+// Prevents the user from selecting text in the document. Used internally
 // by Leaflet to override the behaviour of any click-and-drag interaction on
 // the map. Affects drag interactions on the whole document.
+export function disableTextSelection() {
+	const value = documentStyle[userSelectProp];
+
+	if (value === 'none') {
+		return;
+	}
+
+	prevUserSelect = value;
+	documentStyle[userSelectProp] = 'none';
+}
 
 // @function enableTextSelection()
 // Cancels the effects of a previous [`L.DomUtil.disableTextSelection`](#domutil-disabletextselection).
-export let disableTextSelection;
-export let enableTextSelection;
-let _userSelect;
-if ('onselectstart' in document) {
-	disableTextSelection = function () {
-		DomEvent.on(window, 'selectstart', DomEvent.preventDefault);
-	};
-	enableTextSelection = function () {
-		DomEvent.off(window, 'selectstart', DomEvent.preventDefault);
-	};
-} else {
-	disableTextSelection = function () {
-		const style = document.documentElement.style;
-		_userSelect = style.userSelect;
-		style.userSelect = 'none';
-	};
-	enableTextSelection = function () {
-		document.documentElement.style.userSelect = _userSelect;
-		_userSelect = undefined;
-	};
+export function enableTextSelection() {
+	if (typeof prevUserSelect === 'undefined') {
+		return;
+	}
+
+	documentStyle[userSelectProp] = prevUserSelect;
+	prevUserSelect = undefined;
 }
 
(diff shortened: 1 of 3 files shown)


jonkoops:
Some cross browser testing would be appreciated before merging this one. I've tested this on iOS and macOS Safari (16.2), Firefox (macOS 108.0.1) and Chrome (macOS 108.0.5359.124).

I've added a test HTML file for debugging purposes (see `debug/tests/text-selection.html`).

src/dom/DomUtil.js
```diff
@@ -100,45 +100,38 @@ export function getPosition(el) {
...
-		const style = document.documentElement.style;
-		_userSelect = style.userSelect;
-		style.userSelect = 'none';
-	};
-	enableTextSelection = function () {
-		document.documentElement.style.userSelect = _userSelect;
-		_userSelect = undefined;
-	};
+export function enableTextSelection() {
+	documentStyle[userSelectProp] = prevUserSelect;
```
Falke-Design, line 123: If `enableTextSelection` is called before `disableTextSelection` it will overwrite the default value `auto`. We should set `prevUserSelect` to `auto` or add a check before setting the value

Falke-Design (changes requested):

src/dom/DomUtil.js
Falke-Design: Ok it doesn't allow to set invalid values but I still think it would be cleaner

Falke-Design:
> Some cross browser testing would be appreciated before merging this one. I've tested this on iOS and macOS Safari (16.2), Firefox (macOS 108.0.1) and Chrome (macOS 108.0.5359.124).
> 
> I've added a test HTML file for debugging purposes (see `debug/tests/text-selection.html`).

Tested via BrowserStack:
Win 10 & Win 11, Chrome 108, Edge 108, Firefox 108: working
Mac Monterey: Safari 15.6, Chrome 108, Edge 108, Firefox 108: working

src/dom/DomUtil.js
jonkoops: Agreed, I'll add a check :+1:

mourner:
Do you know why the Windows Chrome CI fails? Haven't seen this error before:

```
'karma' is not recognized as an internal or external command,
operable program or batch file.
Error: Process completed with exit code 1.
```

jonkoops:
> Do you know why the Windows Chrome CI fails? Haven't seen this error before:

It seems to happen on the latest commit on `main` as well, perhaps something has changed in GitHub actions?

jonkoops:
Got a fix for the failures on CI under #8742, they are unrelated to the changes in this PR.

mourner (approved):

jonkoops:
@Falke-Design can we merge this one as far as you are concerned?

Falke-Design (approved):
Looks good 👍

src/dom/DomUtil.js
```diff
@@ -100,45 +100,45 @@ export function getPosition(el) {
...
-		const style = document.documentElement.style;
-		_userSelect = style.userSelect;
-		style.userSelect = 'none';
-	};
-	enableTextSelection = function () {
-		document.documentElement.style.userSelect = _userSelect;
-		_userSelect = undefined;
-	};
+export function enableTextSelection() {
+	if (typeof prevUserSelect === 'undefined') {
```
Falke-Design, line 126: Is there a difference or a benefit to use `typeof prevUserSelect === 'undefined'` instead of `prevUserSelect === undefined`?

jonkoops: Don't think it makes any difference. I've had the latter fail in SSR contexts when detecting Web APIs, but that's about it.

PR #8743, opened 2022-12-24
Run CI on latest version of Windows

jonkoops:
Since we no longer need to run our tests on Internet Explorer we can use the latest version of Windows for CI.


jonkoops:
Gonna merge this in since it's just a small change to the CI. Let me know if there is any feedback.

PR #8747, opened 2022-12-25
Remove `DomUtil.getStyle()`

jonkoops:
Removes the `DomUtil.getStyle()` function and replaces it with [`getComputedStyle()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) where needed.

This function included code that was needed to handle older versions of Firefox and Internet Explorer, but now that we raised our browser target this is no longer required.

### DIFF
--- a/src/layer/marker/Icon.Default.js
+++ b/src/layer/marker/Icon.Default.js
@@ -53,11 +53,9 @@ export const IconDefault = Icon.extend({
 
 	_detectIconPath() {
 		const el = DomUtil.create('div',  'leaflet-default-icon-path', document.body);
-		let path = DomUtil.getStyle(el, 'background-image') ||
-		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
+		const path = this._stripUrl(getComputedStyle(el).backgroundImage);
 
 		document.body.removeChild(el);
-		path = this._stripUrl(path);
 		if (path) { return path; }
 		const link = document.querySelector('link[href$="leaflet.css"]');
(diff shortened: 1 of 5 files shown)


src/layer/marker/Icon.Default.js
```diff
@@ -53,11 +53,9 @@ export const IconDefault = Icon.extend({
 
 	_detectIconPath() {
 		const el = DomUtil.create('div',  'leaflet-default-icon-path', document.body);
-		let path = DomUtil.getStyle(el, 'background-image') ||
-		           DomUtil.getStyle(el, 'backgroundImage');	// IE8
```
jonkoops, line 57: Also removed this old Internet Explorer 8 fallback.

Falke-Design (approved):

src/map/Map.js
```diff
@@ -1108,7 +1108,7 @@ export const Map = Evented.extend({
 
 		container.classList.add(...classes);
 
-		const position = DomUtil.getStyle(container, 'position');
+		const {position} = container.style;
```
mourner, line 1111: Why not use computed style here? Container position could be defined in CSS, and then it would be overriden with `relative`, right?

mourner (changes requested):
Looks good apart from one moment in `Map.js`

src/map/Map.js
jonkoops: Yeah, you are totally right. We should be using computed styles for all of these.

PR #8749, opened 2022-12-26
Hold element positions in WeakMap

jonkoops:
Stores positions set with `DomUtil.setPosition()` in a [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) instead of setting a custom 'private' property (`_leaflet_pos`) on the DOM element itself.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -59,16 +59,14 @@ export function setTransform(el, offset, scale) {
 	el.style.transform = `translate3d(${pos.x}px,${pos.y}px,0)${scale ? ` scale(${scale})` : ''}`;
 }
 
+const positions = new WeakMap();
+
 // @function setPosition(el: HTMLElement, position: Point)
 // Sets the position of `el` to coordinates specified by `position`,
 // using CSS translate or top/left positioning depending on the browser
 // (used by Leaflet internally to position its layers).
 export function setPosition(el, point) {
-
-	/*eslint-disable */
-	el._leaflet_pos = point;
-	/* eslint-enable */
-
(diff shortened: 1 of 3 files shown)


src/dom/DomUtil.js
```diff
@@ -59,26 +59,24 @@ export function setTransform(el, offset, scale) {
...
 // @function getPosition(el: HTMLElement): Point
-// Returns the coordinates of an element previously positioned with setPosition.
+// Returns the coordinates of an element previously positioned with `setPosition`.
+// If the element was not previously positioned 'null' will be returned instead.
 export function getPosition(el) {
 	// this method is only used for elements previously positioned using setPosition,
 	// so it's safe to cache the position for performance
-
-	return el._leaflet_pos || new Point(0, 0);
+	return positions.get(el) ?? null;
```
mourner, line 79: This may be breaking in subtle ways, especially in downstream plugins, so I'd prefer to keep the zero position for a fallback.

mourner (changes requested):

src/dom/DomUtil.js
jonkoops: Sure, we can keep it there for now. Although I do find it strange that we would return a point that is not actually accurate, as it can lead to bad code.

mourner: @jonkoops I think it makes sense if we look at it as a "default" offset value (which is effectively zero if transform is not set).

mourner (approved):

Falke-Design (approved):

PR #8754, opened 2022-12-26
Remove `Browser.canvas` feature detection

jonkoops:
Removes the `Browser.canvas` API  to detect HTML canvas support. Our new browser support target will always support this feature, so it is no longer required to detect it.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1873,9 +1873,7 @@ describe('Map', () => {
 			expect(called).to.eql(4);
 		});
 
-		it('prevents default action of contextmenu if there is any listener', function () {
-			if (!L.Browser.canvas) { this.skip(); }
-
+		it('prevents default action of contextmenu if there is any listener', () => {
 			removeMapContainer(map, container);
 			container = createContainer();
 			map = L.map(container, {
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -92,12 +92,6 @@ const passiveEvents = (function () {
 	return supportsPassiveOption;
 }());
 
-// @property canvas: Boolean
-// `true` when the browser supports [`<canvas>`](https://developer.mozilla.org/docs/Web/API/Canvas_API).
-const canvas = (function () {
-	return !!document.createElement('canvas').getContext;
-}());
-
 // @property svg: Boolean
 // `true` when the browser supports [SVG](https://developer.mozilla.org/docs/Web/SVG).
 const svg = !!(document.createElementNS && svgCreate('svg').createSVGRect);
@@ -136,7 +130,6 @@ export default {
 	mobileGecko,
 	retina,
 	passiveEvents,
-	canvas,
 	svg,
 	inlineSvg,
 	mac,
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -490,5 +490,5 @@ export const Canvas = Renderer.extend({
 // @factory L.canvas(options?: Renderer options)
 // Creates a Canvas renderer with the given options.
 export function canvas(options) {
-	return Browser.canvas ? new Canvas(options) : null;
+	return new Canvas(options);
 }


Falke-Design (approved):
This makes the `prefer` in [preferCanvas](https://leafletjs.com/reference.html#map-prefercanvas) obsolete.

mourner (approved):

jonkoops:
> This makes the `prefer` in [preferCanvas](https://leafletjs.com/reference.html#map-prefercanvas) obsolete.

I don't think it does, as the code that determines the renderer [still exists](https://github.com/Leaflet/Leaflet/blob/9f35179419aad62d5d735f49a70e7a79fd67cfd5/src/layer/vector/Renderer.getRenderer.js#L39-L44). That said, do we want to change the default here or perhaps even get rid of one of the renderers?

IvanSanchez:
> That said, do we want to change the default here or perhaps even get rid of one of the renderers?

I think the default is just fine - most use cases get covered by `SVG`, and some minor use caseswork better with `Canvas`. Removing `Canvas` might be too much for plugin authors, even with the whole 2.0 breaking changes.

I'd propose renaming `preferCanvas = true` to `preferredRenderer = L.Canvas` (`L.` just for the sake of clarity, I know we're getting away from the namespace and into `import`s). Ideally I'd like to put some finishing touches on [Gleo](https://gitlab.com/IvanSanchez/gleo) and make it a Leaflet WebGL vector renderer. Having a `preferredRenderer` map option would allow for hot-plugging the fabled WebGL rendering backend.

Falke-Design:
> > This makes the `prefer` in [preferCanvas](https://leafletjs.com/reference.html#map-prefercanvas) obsolete.
> 
> I don't think it does, as the code that determines the renderer [still exists](https://github.com/Leaflet/Leaflet/blob/9f35179419aad62d5d735f49a70e7a79fd67cfd5/src/layer/vector/Renderer.getRenderer.js#L39-L44). That said, do we want to change the default here or perhaps even get rid of one of the renderers?

It makes sense to have a flag that the user can decide if he wants to render everything as canvas or svg. But before `preferCanvas` checked if canvas is possible for the browser else use svg. Now it will be always rendered as cavas, so the option `useCanvas` would make more sense instead of `preferCanvas`.

No need to change the option name, I only wanted to show that `prefer` is maybe not accurate anymore

jonkoops:
> Having a `preferredRenderer` map option would allow for hot-plugging the fabled WebGL rendering backend.

Doesn't `Map` already have a [`renderer`](https://leafletjs.com/reference.html#map-renderer) option? That would allow us to do this?

I think we should perhaps get rid of `preferCanvas` as an option, it's confusing to me what renderer would be used. I'd much rather only see an explicit `renderer` option.

jonkoops:
> Now it will be always rendered as canvas, so the option `useCanvas` would make more sense instead of `preferCanvas`.

Are you sure? The [change in this PR](https://github.com/Leaflet/Leaflet/pull/8754/files#diff-f4127fe2c74bbf5c584e01246849fac1954ead8521449585b61d268bfb7b059eL493) should only affect the fallback to SVG when using `L.canvas()`. The conditional logic that picks the renderer for the map [still depends](https://github.com/Leaflet/Leaflet/blob/9f35179419aad62d5d735f49a70e7a79fd67cfd5/src/layer/vector/Renderer.getRenderer.js#L43) on `preferCanvas`, which by default is falsy.

So unless I am missing something I do believe the default renderer is SVG based, no?

Falke-Design:
> > Having a `preferredRenderer` map option would allow for hot-plugging the fabled WebGL rendering backend.
> 
> Doesn't `Map` already have a [`renderer`](https://leafletjs.com/reference.html#map-renderer) option? That would allow us to do this?
> 
> I think we should perhaps get rid of `preferCanvas` as an option, it's confusing to me what renderer would be used. I'd much rather only see an explicit `renderer` option.

Yeah I think that makes sense

Falke-Design:
> > Now it will be always rendered as canvas, so the option `useCanvas` would make more sense instead of `preferCanvas`.
> 
> Are you sure? The [change in this PR](https://github.com/Leaflet/Leaflet/pull/8754/files#diff-f4127fe2c74bbf5c584e01246849fac1954ead8521449585b61d268bfb7b059eL493) should only affect the fallback to SVG when using `L.canvas()`. The conditional logic that picks the renderer for the map [still depends](https://github.com/Leaflet/Leaflet/blob/9f35179419aad62d5d735f49a70e7a79fd67cfd5/src/layer/vector/Renderer.getRenderer.js#L43) on `preferCanvas`, which by default is falsy.
> 
> So unless I am missing something I do believe the default renderer is SVG based, no?

Sorry, my explanation is getting worse and worse 😅 I try it again.
With `preferCanvas` (old behaviour) we have a flag when is set to `true`, checks if the browser supports canvas. If the browser support canvas then use canvas, else use svg.
Now (new behaviour) the check `Browser.canvas` was removed so there is no **_prefer_** anymore, it is **_use_** or **_not use_** canvas.

jonkoops:
Gotcha, yeah that is correct. What would y'all think about removing `preferCanvas` entirely and using the provided `renderer` from options instead?

PR #8759, opened 2022-12-28
Use Pointer Events for `Control.Layers`

Falke-Design:
Part of https://github.com/Leaflet/Leaflet/issues/7200

The replacement of `mouseleave` to `pointerleave` had a different result on touchdevices. `pointerleave` was triggered when a click on a checkbox happend and then the control was collapsed instead of adding the layer to the map. So a check if it is a mouse event was needed.


src/control/Control.Layers.js
```diff
@@ -165,7 +165,10 @@ export const Layers = Control.extend({
 
 	// @method collapse(): this
 	// Collapse the control container if expanded.
-	collapse() {
+	collapse(e) {
```
IvanSanchez, line 168: Nitpicking: Please use `ev` instead of `e`.

Whenever I see `e` as a parameter, I don't know if that's an `ev`ent, an `err`or or an `ex`ception.

src/control/Control.Layers.js
```diff
@@ -165,8 +165,13 @@ export const Layers = Control.extend({
 
 	// @method collapse(): this
 	// Collapse the control container if expanded.
-	collapse() {
-		this._container.classList.remove('leaflet-control-layers-expanded');
+	collapse(ev) {
+		// On touch devices `pointerleave` is fired while clicking on a checkbox.
```
IvanSanchez, line 169: Out of curiosity: have you tried with `pointerout`? It has a similar but *slightly* different behaviour than `pointerleave`, for this kind of situations.

jonkoops:
@Falke-Design did a dirty rebase and squash on this PR to fix the merge conflicts, can you take a look and see if everything is still looking good?

src/control/Control.Layers.js
Falke-Design: Just tested, is `pointerout` is fired same as `pointerleave`

jonkoops:
@Falke-Design let's try to avoid getting stuff from `main` using the merge strategy, as it can be an annoying web to untangle. If possible, a rebase (and optionally a squash) is preferred.

spec/.eslintrc.js
```diff
@@ -12,5 +12,9 @@ module.exports = {
 		chai: false,
 		sinon: false,
 		Hand: false,
+		touchEventType: false, /* defined in SpecHelper.js */
+		createContainer: false, /* defined in SpecHelper.js */
+		removeMapContainer: false, /* defined in SpecHelper.js */
+		pointerType: false /* defined in SpecHelper.js */
```
jonkoops, line 18: These should no longer be needed, as `SpecHelper.js` is now a module, and doesn't expose any globals.
```suggestion
```

jonkoops (approved):

Falke-Design:
> @Falke-Design let's try to avoid getting stuff from main using the merge strategy, as it can be an annoying web to untangle. If possible, a rebase (and optionally a squash) is preferred.

Thanks for the reminder, I already use rebase always but this time it was not possible.

PR #8760, opened 2022-12-28
Refactor Event handling and happen.js

Falke-Design:
Refactoring happen.js to make `pointer` events available


mourner (changes requested):
1. I'd really like to avoid any libraries vendored in the Leaflet repo. A more standard course of action would be to fork the library, make changes there through PRs so that it's clear what changed, publish it on NPM namespaced and then use it as a dependency.
2. You'd need to update your NPM version so that the lockfile doesn't downgrade from v2 to v1.

IvanSanchez:
@Falke-Design Gentle reminder that https://github.com/Leaflet/prosthetic-hand exists and does pointer events.

Falke-Design:
> @Falke-Design Gentle reminder that https://github.com/Leaflet/prosthetic-hand exists and does pointer events.

Is there an easy way without creating a Finger to simulate the event? Like `once(elem, 'keypress')`

Falke-Design:
> 1. I'd really like to avoid any libraries vendored in the Leaflet repo. A more standard course of action would be to fork the library, make changes there through PRs so that it's clear what changed, publish it on NPM namespaced and then use it as a dependency.
> 2. You'd need to update your NPM version so that the lockfile doesn't downgrade from v2 to v1.

1. I thought about that but I don't want to maintaine this repo, so I thought I add only the necessary code for Leaflet. Not that in a few years the repo is unmaintained again.
2. ups, will be changed :+1:

jonkoops:
Perhaps we can refactor the tests so that `happen` can be removed and only `prosthetic-hand` is used? I haven't yet intimately familiarized myself  with these libraries, but it looks like `happen` is pretty much dead.

So either that, or we create a very light-weight wrapper to quickly simulate pointer events only. No need to pull in a whole library.

WDYT?

Falke-Design:
I outsourced the code into https://github.com/Falke-Design/ui-event-simulator.

Maybe I oversee something but I don't think that `prosthetic-hand` is the right tool to simulate simple and all variants of events. If there is a way please provide a PR until then I suggest to merge this PR. 

This is needed to proceed with #7200.

IvanSanchez:
> Maybe I oversee something but I don't think that `prosthetic-hand` is the right tool to simulate simple and all variants of events. If there is a way please provide a PR until then I suggest to merge this PR.

You're right - `prosthetic-hand` is good for a succession of `PointerEvent`s, but cannot do individual `KeyboardEvent`s.

jonkoops (approved):
If this can help us land the other PRs I'd like to get this merged. We can always improve upon it in the future.

jonkoops:
@mourner @IvanSanchez shall we move this one along, or are there any objections?

mourner (approved):

jonkoops:
Merging this in so we can continue our work.

PR #8764, opened 2022-12-30
new render shape of CircleMarker as N-Corner-Star

haelue:
A new render-type of CircleMarker as an N-Corner-Star, by set CircleMarker options of: star: n (n > 2).

since it looks too monotonous of vector-symbol only with lines, polygons & circles.


Falke-Design:
Thank you for your contribution but it would be better to have this as a plugin. You can take a look into this similar plugin: https://github.com/Falke-Design/L.Donut

PR #8766, opened 2022-12-31
Use SVG for icons instead of bitmaps

### LINKED ISSUE #8756: Use SVG icons instead of bitmaps
Now that all browsers we're targeting support SVG we should make an effort to replace the bitmap icons we have. I am marking this as 'help wanted' since someone will have to recreate the icons.

jonkoops:
Closes #8756


jonkoops:
There are a couple of unresolved issues which is why I have marked this PR as a draft.

Firstly, there is no SVG version of the marker shadow. I don't have much experience designing these sort of things so I would appreciate it if someone could create an SVG version of this asset.

I am also considering the idea of removing the shadow image, and instead using the [`drop-shadow`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/drop-shadow) CSS filter. Which should automatically create shadows in the correct shape for 3rd party icons as well, without needing additional assets.

Secondly, it looks like the SVG version of the marker does not have the hole on the top filled in white (see below). Something that is present in the bitmap version.

The SVG verrsion of the marker icon, without white fill.

WDYT?

src/images/layers.svg
```diff
@@ -1 +0,0 @@
-<svg xmlns="http://www.w3.org/2000/svg" height="26" width="26"><path d="M.032 17.056l13-8 13 8-13 8-13-8" fill="#b9b9b9"/><path d="M.032 17.056l-.032.93 13 8 13-8 .032-.93-13 8z" fill="#737373"/><path d="M0 13.076l13-8 13 8-13 8-13-8" fill="#cdcdcd"/><path d="M0 13.076v.91l13 8 13-8v-.91l-13 8z" fill="#737373"/><path d="M0 8.986l13-8 13 8-13 8-13-8" fill-opacity=".585" stroke="#797979" stroke-width=".1" fill="#e9e9e9"/><path d="M0 8.986v1l13 8 13-8v-1l-13 8z" fill="#737373"/></svg>
```
Falke-Design, line 1: Why removing the source files?

Falke-Design (changes requested):

mourner:
Before landing this change, we should make sure it has no implications on performance when using a lot of markers on screen at once (e.g. thousands), and check in Blink, Webkit and Firefox since they are optimized differently with regards to rendering SVG.

Also, it would be nice to explore how the new markers could styled with CSS — e.g. add classes to some of the marker elements so that we could pass a className to a marker that we can target to make it red instead of blue.

jonkoops:
> Also, it would be nice to explore how the new markers could styled with CSS — e.g. add classes to some of the marker elements so that we could pass a className to a marker that we can target to make it red instead of blue.

We could probably use CSS variables for this even. However I do believe this would require us to in-line the SVG either way (vs. loading it as an image). Not sure what the performance implications of that would be.

src/images/layers.svg
jonkoops: Ah, I think I messed up rebasing this. Let me put them back.

Falke-Design (approved):
The changes are looking good for me

I did a small performance test on Firefox and Chrome and the results are pretty same. SVG is on Firefox a little bit faster.

My used code: (how long until everything is created & how long `flyTo` takes after calling `startTest()`)
```
function createMarkers(number, type){
	var row = 0;
	var column = 0;
	for(var i = 0; i < number; i++) {
		if(i % 200 === 0){
			row++;
		}
		column = i % 200;
		if(type === 'svg') {
			var icon = new L.Icon.Default({
				iconUrl: './marker.svg',
				shadowUrl: './marker-shadow.svg',
				iconSize: [25, 41],
				iconAnchor: [12, 41],
				popupAnchor: [1, -34],
				tooltipAnchor: [16, -28],
				shadowSize: [41, 41]
			});
			var marker = new L.Marker([latlng.lat+(0.0001*row), latlng.lng+(0.0001*column)], {icon}).addTo(map);
		} else {
			var icon = new L.Icon.Default({
				iconUrl: './marker-icon.png',
				shadowUrl: './marker-shadow.png',
				iconSize: [25, 41],
				iconAnchor: [12, 41],
				popupAnchor: [1, -34],
				tooltipAnchor: [16, -28],
				shadowSize: [41, 41]
			});
			var marker = new L.Marker([latlng.lat+(0.0001*row), latlng.lng+(0.0001*column)], {icon, draggable: true}).addTo(map);
		}
	}
}

console.time('create');
createMarkers(4000, 'svg')
console.timeEnd('create')

map.on('click',console.log	)

function startTest(){
	console.time('perf');
	map.flyTo([50.503300941513935, 30.513067245483402], 18, {animate: true});
	map.once('moveend',()=>{
		console.timeEnd('perf')
	})
}
```

mourner:
@Falke-Design `flyTo` duration should be approximately the same regardless of performance. I'd measure number of frames instead (relative to theoretical maximum, which is `duration * 60`).

timbtimbtimb:
> > Also, it would be nice to explore how the new markers could styled with CSS — e.g. add classes to some of the marker elements so that we could pass a className to a marker that we can target to make it red instead of blue.
> 
> We could probably use CSS variables for this even. However I do believe this would require us to in-line the SVG either way (vs. loading it as an image). Not sure what the performance implications of that would be.

From experience inline SVG are terribly slow. I've had to convert them to base64 in one of my projects

Falke-Design:
@jonkoops why did you closed this PR?

jonkoops:
I was cleaning my fork up, so that is why this closed as well. Re-opening.

coryasilva:
@jonkoops First thanks for all the work from you and the other maintainers and creator. I am a long time user.

### Proposal:
1. Move the svg shadow to a data uri `` for caching and faster rendering over `` (not even counting the http req)
2. Simplify icon gradients _(I have an un-researched worry about SVG gradient performance -- I remember that not all browsers have the same optimizations here as they do for css)_
3. Create the icon with `document.createElementNS` so we can leverage fill="currentColor" to change its color. I suppose the border could something like rgba(0,0,0,0.5) with a mix-blend-mode of multiply to generally get a similar effect.

#### Pros
- Get rid of all the bundler drama
  - #9637
  - #4968
- ...and the retina this retina that...
- Marginally faster than ``
- We can change the color dynamically***

#### ***Color digression:
- CSS filters work regardless of img or svg -- but that is sort of nuanced, frustrating devx, and actually slower rendering.

- Dynamically change uri string will work -- may open us up to more XSS sinks and probably requires an addition to the API.

- We could use `document.createElementNS("svg"` but then we would need to unique the gradient id's else they will collide.

I totally understand if you all say "nah" -- I do maintain [(and am currently upgrading/rewriting) a plugin](https://github.com/coryasilva/Leaflet.ExtraMarkers/tree/v2) for all this but figure it is easier for this to work OOTB.

jonkoops:
@coryasilva I want to keep this simple and focus on using as much of the existing code as possible. We can always handle enhancements as a separate proposal (issue, PR).

jonkoops:
@Leaflet/core @Leaflet/leaflet-maintainers can I get a pass at this? I've rebased the work, and I would like to get it landed as it is right now. Performance seems fine as far as I can tell, it seems the browser is able to cache the resulting bitmap, so it is not any less performant than before.

mourner:
> Performance seems fine as far as I can tell, it seems the browser is able to cache the resulting bitmap, so it is not any less performant than before.

Can we double-check this on different browsers just to be on the safe side? E.g. add a few thousand random markers on a page and see how zooming in and out performs.

coryasilva:
FWIW here is a performance script I have been using

```html
<!doctype html>
<html>
<head>
	<meta charset="utf-8">
	<meta name="viewport" content="width=device-width">
	<title>Perf</title>
	<link rel="stylesheet" href="https://unpkg.com/leaflet@2.0.0-alpha.1/dist/leaflet.css" />
</head>
<body>
	<div id="map" style="width:100%;height:400px;"></div>
	<script type="importmap">
		{
			"imports": {
				"leaflet": "https://unpkg.com/leaflet@2.0.0-alpha.1/dist/leaflet.js"
			}
		}
	</script>
	<script type="module">
		import { FeatureGroup, Map, Marker } from "leaflet";

		function perfTest() {
			const markers = new FeatureGroup();
			const count = 1000;
			const map = new Map("map", { center: [0, 0], zoom: 4 });
			const bounds = map.getBounds();
			const sw = bounds.getSouthWest();
			const ne = bounds.getNorthEast();
			const lngSpan = ne.lng - sw.lng;
			const latSpan = ne.lat - sw.lat;

			for (let i = 0; i < count; i++) {
				new Marker([
					sw.lat + latSpan * Math.random(),
					sw.lng + lngSpan * Math.random(),
				]).addTo(markers);
			}

			console.time();
			markers.addTo(map);

			window.requestAnimationFrame(() => {
				console.timeEnd();
			});
		}

		perfTest();
</script>
</body>
</html>
```

src/layer/marker/Icon.Default.js
```diff
@@ -20,9 +20,9 @@ export class IconDefault extends Icon {
 
 	static {
 		this.setDefaultOptions({
-			iconUrl:       'marker-icon.png',
-			iconRetinaUrl: 'marker-icon-2x.png',
-			shadowUrl:     'marker-shadow.png',
+			iconUrl:       'marker.svg',
+			iconRetinaUrl: 'marker.svg',
```
Falke-Design:
```suggestion
			iconUrl:       'marker-icon.svg',
			iconRetinaUrl: 'marker-icon.svg',
```

Falke-Design:
Tested with the code from @coryasilva but with 5000 markers

Windows 11:
* Firefox 143.0.1: svg: ~110ms, png: ~105ms
* Chrome 40.0.7339.128 : svg: ~200ms, png: ~200ms

spec/suites/layer/marker/MarkerSpec.js
```diff
@@ -157,11 +157,7 @@ describe('Marker', () => {
 			expect(oldIcon).to.not.equal(marker._icon); // Check that the _icon is NOT re-used
 			expect(oldIcon.parentNode).to.equal(null);
 
-			if (Browser.retina) {
-				expect(marker._icon.src).to.contain('marker-icon-2x.png');
-			} else {
-				expect(marker._icon.src).to.contain('marker-icon.png');
-			}
+			expect(marker._icon.src).to.contain('marker.svg');
```
Falke-Design:
```suggestion
			expect(marker._icon.src).to.contain('marker-icon.svg');
```

PR #8710, opened 2022-12-12
docs: minor typo in code example

Lalaluka:
Hi,
I found a very very minor typo in a code example in the Docs.
This PR fixes it :)

PR #8767, opened 2022-12-31
Added missing curly bracket in docs

i-sukhanov:
Screenshot 2022-12-31 at 17 06 20

Added missing curly bracket in popup docs example


Falke-Design (approved):
I'm merging this to update the website. The docs already fixe with #8710

jonkoops:
Are the docs not generated from the source code?

Falke-Design:
yes but then we need to run `npm run docs`, commit and push. It would have the same result

PR #8768, opened 2022-12-31
Add shadow-icon as svg icon and add white circle to marker-icon

Falke-Design:
Follow up #8766

This is my best result of creating the shadow as svg.
Maybe the `viewBox` needs to be changed but here ends my knowledge.

**Shadow icon diff:**
Left `.png`, right `.svg`:

grafik

grafik grafik

### DIFF
--- a/src/images/marker-shadow.svg
+++ b/src/images/marker-shadow.svg
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 23.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Ebene_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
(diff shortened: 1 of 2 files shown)


jonkoops:
The SVG shadow seems a bit lighter, but it still looks nice. Perhaps we should we should also optimize these file with [`svgo`](https://github.com/svg/svgo) like we did in the past?

Falke-Design:
> The SVG shadow seems a bit lighter, but it still looks nice. Perhaps we should we should also optimize these files with [`svgo`](https://github.com/svg/svgo) like we did in the past?

Yes it is a little bit lighter but was the best what I can achieve. 
SVGO sounds good but I would optimize it while generating the dist files and keep the source file like know.

jonkoops:
> Yes, it is a little bit lighter but was the best what I can achieve. SVGO sounds good but I would optimize it while generating the dist files and keep the source file like know.

Fair enough, let's get this merged 👍

jonkoops (approved):

jonkoops:
Thanks for the assist @Falke-Design 🙌

PR #8770, opened 2022-12-31
Remove Gecko-based browser detection

jonkoops:
Removes Gecko-based browser detection (`Browser.gecko` and `Browser.mobileGecko`). This code is not used internally in Leaflet so it can be safely removed. Users should do feature detection instead where possible.


Falke-Design (approved):

jonkoops:
Going to merge this as I have some other PRs in the works that are dependent on this. If anyone has any objections/remarks let me know.

PR #8779, opened 2023-01-02
Adding colors to marker

Falke-Design:
Based on #8766

Adding colors to the marker over an `css filter`:

grafik

Colors: pink, dark-blue, lila, magenta, red, orange, brown, green, mint


jonkoops:
Interesting approach!

src/layer/marker/Marker.js
```diff
@@ -99,7 +99,9 @@ export const Marker = Layer.extend({
 
 		// @option autoPanSpeed: Number = 10
 		// Number of pixels the map should pan by.
-		autoPanSpeed: 10
+		autoPanSpeed: 10,
+
+		color: null
```
IvanSanchez, line 104: The approach is interesting, but I don't like the semantics of this new `L.Marker` option. It works for the default case, but if a user puts an image of a red circle as the icon, then it breaks.

We might want to consider breaking off `L.Icon` from `L.Marker` and do some kind of `L.TintMarker`. Or have the source image to be greyscale and apply a colour to it (not possible with CSS as it seems).

Falke-Design:
Checkout my other attempt over `Icon.Default`: #8855

Vectorial1024:
Late join, but I believe it's better to split this feature into another library (e.g. https://github.com/Vectorial1024/leaflet-color-markers/issues/3), just like what this said https://github.com/Leaflet/Leaflet/pull/8779#discussion_r1111960534 : what if the user tries to use an entirely different image for the marker, and then the `color` option becomes redundant and non-sensical.

PR #8784, opened 2023-01-06
Fix rounding errors in center calculation for small layers

### LINKED ISSUE #8783: Tooltip anchor locations are incorrect for small GeoJSON polygons in 1.9.x
### Checklist

### Steps to reproduce

This JSFiddle Illustrates the problem well: https://jsfiddle.net/michaelthoreau/ekafvph8/6/

This issue seems to only affect geojson polygons that are geographically small, regardless of pixel size when drawn.

### Expected behavior

Both tooltips should appear centered on the polygons. 

### Current behavior

The smaller polygon's tooltip appears in the wrong position. This behaviour does not appear in leaflet@1.8.0.

https://jsfiddle.net/michaelthoreau/ekafvph8/6/

### Environment

- Leaflet version: 1.9.3
- Browser (with version): 108.0.5359.124 (Official Build) (arm64)
- OS/Platform (with version): Mac OS 12.6

Falke-Design:
For small areas, the center was to far away of the layer. It looks like the reason are rounding errors (Thx to turf.js for the idea: https://github.com/Turfjs/turf/blob/master/packages/turf-center-of-mass/index.ts#L39). But this solution doesn't work good for normal / larger areas, so I added a check if the area of the bounds are small then `1700` it using the rounding preventing solution.

To solve this I needed to add the new function `PolyUtil.centroid`.

This needs merged in the `main` banche too.

error:
grafik

fixed:
grafik

```
L.polygon([[42.87097909758862, -81.12594320566181], [42.87108302016597, -81.12594320566181], [42.87108302016597, -81.12576504805303], [42.87097909758862, -81.12576504805303]]).addTo(map);
```

Fixes #8783

PS: I also cleaned up some unnecessary tests.

### DIFF
--- a/src/geometry/LineUtil.js
+++ b/src/geometry/LineUtil.js
@@ -1,6 +1,8 @@
 import {Point, toPoint} from './Point';
 import * as Util from '../core/Util';
 import {toLatLng} from '../geo/LatLng';
+import {centroid} from './PolyUtil';
+import {toLatLngBounds} from '../geo/LatLngBounds';
 
 
 /*
@@ -257,9 +259,20 @@ export function polylineCenter(latlngs, crs) {
 		latlngs = latlngs[0];
 	}
 
+	var centroidLatLng = toLatLng([0, 0]);
+
+	var bounds = toLatLngBounds(latlngs);
+	var areaBounds = bounds.getNorthWest().distanceTo(bounds.getSouthWest()) * bounds.getNorthEast().distanceTo(bounds.getNorthWest());
+	// tests showed that below 1700 rounding errors are happening
+	if (areaBounds < 1700) {
+		// getting a inexact center, to move the latlngs near to [0, 0] to prevent rounding errors
+		centroidLatLng = centroid(latlngs);
+	}
+
 	var points = [];
-	for (var j in latlngs) {
-		points.push(crs.project(toLatLng(latlngs[j])));
+	for (var k in latlngs) {
+		var latlng = toLatLng(latlngs[k]);
+		points.push(crs.project(toLatLng([latlng.lat - centroidLatLng.lat, latlng.lng - centroidLatLng.lng])));
 	}
 
 	var len = points.length;
@@ -288,5 +301,7 @@ export function polylineCenter(latlngs, crs) {
 			}
 		}
 	}
-	return crs.unproject(toPoint(center));
+
+	var latlngCenter = crs.unproject(toPoint(center));
+	return toLatLng([latlngCenter.lat + centroidLatLng.lat, latlngCenter.lng + centroidLatLng.lng]);
 }
--- a/src/geometry/PolyUtil.js
+++ b/src/geometry/PolyUtil.js
@@ -119,7 +119,8 @@ export function centroid(coords) {
 	var latSum = 0;
 	var lngSum = 0;
 	var len = 0;
-	for (const coord of coords) {
+	for (var coord of coords) {
+
 		var latlng = toLatLng(coord);
 		latSum += latlng.lat;
 		lngSum += latlng.lng;
--- a/src/geometry/PolyUtil.js
+++ b/src/geometry/PolyUtil.js
@@ -119,9 +119,8 @@ export function centroid(coords) {
 	var latSum = 0;
 	var lngSum = 0;
 	var len = 0;
-	for (var coord of coords) {
-
-		var latlng = toLatLng(coord);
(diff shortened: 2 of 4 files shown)


Falke-Design:
We should release this fix. @mourner @jonkoops @IvanSanchez can you please review.

src/geometry/PolyUtil.js
```diff
@@ -95,5 +107,23 @@ export function polygonCenter(latlngs, crs) {
...
+}
+
+/* @function centroid(latlngs: LatLng[]): LatLng
+ * Returns the 'center of mass' of the passed LatLngs.
+ */
+export function centroid(coords) {
+	var latSum = 0;
+	var lngSum = 0;
+	var len = 0;
+	coords.forEach(function (coord) {
```
jonkoops, line 122: I believe we're avoiding using `.forEach()` in v1 as it's not supported by all the browsers. So this should probably be a classic loop.

jonkoops (changes requested):

src/geometry/PolyUtil.js
Falke-Design: makes sense, I will replace that

src/geometry/PolyUtil.js
```diff
@@ -95,5 +107,23 @@ export function polygonCenter(latlngs, crs) {
...
+}
+
+/* @function centroid(latlngs: LatLng[]): LatLng
+ * Returns the 'center of mass' of the passed LatLngs.
+ */
+export function centroid(coords) {
+	var latSum = 0;
+	var lngSum = 0;
+	var len = 0;
+	for (const coord of coords) {
```
Falke-Design, line 122:
```suggestion
	for (var coord of coords) {

```

jonkoops (approved):
LGTM

src/geometry/LineUtil.js
```diff
@@ -257,10 +259,21 @@ export function polylineCenter(latlngs, crs) {
 		latlngs = latlngs[0];
 	}
 
+	var centroidLatLng = toLatLng([0, 0]);
+
+	var bounds = toLatLngBounds(latlngs);
+	var areaBounds = bounds.getNorthWest().distanceTo(bounds.getSouthWest()) * bounds.getNorthEast().distanceTo(bounds.getNorthWest());
+	// tests showed that below 1700 rounding errors are happening
+	if (areaBounds < 1700) {
```
IvanSanchez, line 267: I have the suspicion that the threshold is not based on the area, but rather on the (maximum) magnitude of *either* the horizontal or vertical component.

Did these tests include very slim (e.g. small `x` but large `y`) polygons near e.g. latitude 0, longitude 179? Longitude 0 latitude 82?

IvanSanchez (changes requested):

src/geometry/LineUtil.js
Falke-Design: > I have the suspicion that the threshold is not based on the area, but rather on the (maximum) magnitude of either the horizontal or vertical component.

@IvanSanchez I do not really understand what this means, but every help is appriciated. This is error is much over my knowledge and this workaround was the best what I was eable to achieve. 

- Demo (v1.9.3) where you can edit a small polygon and see the center "jumping" https://jsfiddle.net/3m0d4Lqc/
- small `x` but large `y`: https://jsfiddle.net/p8wt2q3o/2/:
  - 1.9.3: 
  https://jsfiddle.net/p8wt2q3o/2/
grafik
  - new: 
grafik
- large`x` but small  `y`: https://jsfiddle.net/p8wt2q3o/2/:
  - 1.9.3: 
  https://jsfiddle.net/c4Lj2wup/
grafik
  - new: 
grafik

IvanSanchez: This is due to how floating point arithmetic works. One of the diagrams in the [wikipedia page about floating point](https://en.wikipedia.org/wiki/Floating-point_arithmetic) sums it nicely:

image

Precision is good near zero, but becomes worse the further from zero you are.

This can be seen in your jsfiddle. This is how it looks with a rectangle like `[[0.0445, 179.925448],[0.0447, 179.925449]]`:
image

...and see how it's different from a rectangle like `[[0.0445, 0.925448],[0.0447, 0.925449]]`:

image

Which means that your code covers the worst case for bound latitude-longitude (-90,90; -180,180), which is good. But I can *guess* it will fail for maps with `L.CRS.Simple`, which can use larger coordinates.

I'm OK with that, but want you to be aware.

PR #8790, opened 2023-01-10
Create leaflet-markerplayer.md

0n3byt3:


0n3byt3:
ok.. but is there a specific important reason for that?

Falke-Design:
It will always reference back to `MovingMarker`, but you built a new plugin, so it should be a repo without a _fork_-reference.
Also it is a quality thing that a plugin in our list should not be a forked repo of a already existing plugin. 

grafik

0n3byt3:
thnx for your info and fast response... its done.. you can check it now

0n3byt3:
🤔👀

PR #8791, opened 2023-01-12
remove duplicate assert

yuhongliang90:
duplicate assert

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -235,7 +235,7 @@ export function geometryToLayer(geojson, options) {
 function _pointToLayer(pointToLayerFn, geojson, latlng, options) {
 	return pointToLayerFn ?
 		pointToLayerFn(geojson, latlng) :
-		new Marker(latlng, options && options.markersInheritOptions && options);
+		new Marker(latlng, options && options.markersInheritOptions);
 }
 
 // @function coordsToLatLng(coords: Array): LatLng


src/layer/GeoJSON.js
```diff
@@ -235,7 +235,7 @@ export function geometryToLayer(geojson, options) {
 function _pointToLayer(pointToLayerFn, geojson, latlng, options) {
 	return pointToLayerFn ?
 		pointToLayerFn(geojson, latlng) :
-		new Marker(latlng, options && options.markersInheritOptions && options);
+		new Marker(latlng, options && options.markersInheritOptions);
```
Falke-Design, line 238: the return value is now not `options` anymore, this needs more investigations

Falke-Design (changes requested):

src/layer/GeoJSON.js
IvanSanchez: The first `options &&` ensures that `options` is an non-undefined object, whereas the second `&& options` handles the semantics of `markersInheritOptions` (returning `options` if `markersInheritOptions` is truthy)

Should be `options?.markersInheritOptions && options` instead, or maybe `options?.markersInheritsOptions ? options : {}`. The behaviour is the same, but the fancy syntax makes the purpose a bit clearer IMO.

IvanSanchez (changes requested):

src/layer/GeoJSON.js
Falke-Design: I prefer `options?.markersInheritOptions && options`

Falke-Design:
I'm closing this because everything works as designed

PR #8794, opened 2023-01-13
fix vector drifts when `zoomAnimation` is `false` and zooming via `flyTo` or pinch

### LINKED ISSUE #8117: Cover vector drift issue with tests

plainheart:
### Background

First I'd like to thank @mourner for the fix in #8103, which makes it work when `zoomAnimation` is `true`.

However, as mentioned in #7466, #8644, and my https://github.com/Leaflet/Leaflet/pull/8103#issuecomment-1235646390, it doesn't work when `zoomAnimation` is set as `false`.
You can refer to the test case `debug/vector/vector-drift.html` added in this PR to reproduce or the [online demo](https://plnkr.co/edit/qHqGIded24Z4MYtO) provided by #7466.

In this PR, I tried to set the `transform-origin` as `0 0` for the container of the SVG/Canvas renderer. By default, it might be `center`. I'm not sure if it is right way to fix this bug but it works indeed and looks simple.

closes #8117

### Comparison

| Platform | Before | After |
| :----: | :----: | :----: |
| Desktop | bug | fixed 
| Mobile | bug-mobile | fixed-mobile |

### DIFF
--- a/debug/vector/vector-drift.html
+++ b/debug/vector/vector-drift.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<title>Leaflet debug page</title>
+	<link rel="stylesheet" href="../../dist/leaflet.css" />
+	<link rel="stylesheet" href="../css/screen.css" />
+	<script src="../../dist/leaflet-src.js"></script>
+</head>
+<body>
+	<button id="btn-1">A</button>
+	<button id="btn-2">B</button>
+	<div id="map"></div>
+
+	<script>
+		var map = L.map('map', {
+			zoomAnimation: false,
+			// preferCanvas: true
+		}).setView([0, 0], 7);
+
+		var markerA = L.marker([1, 0]).addTo(map);
+		var markerB = L.marker([1, 2]).addTo(map);
+		L.polygon([
+			[0, 0],
+			[2, 0],
+			[2, 2],
+			[0, 2],
+			[0, 0],
+		])
+		.bindPopup('Hello world')
+		.addTo(map);
+
+		// or pinch zoom in mobile
+		document.getElementById('btn-1').addEventListener('click', function () {
+			map.flyTo(markerA.getLatLng(), 6);
+		});
+		document.getElementById('btn-2').addEventListener('click', function () {
+			map.flyTo(markerB.getLatLng(), 7);
+		});
+	</script>
+</body>
+</html>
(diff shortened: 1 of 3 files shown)


mourner:
@plainheart the fix looks great, thank you for looking into this! Since it's not covered by tests currently, I wonder if it would be covered by #8644 — can you take a look? Let's see if it passes if we turn zoomAnimation off there with this fix applied, and fails without.

plainheart:
I'm not familiar with how to add test cases, so I'd like to ask you to help me to add some appropriate test cases. Thanks!

mourner:
@plainheart oh, I mean there's already a test case written in that PR — we just need to check if it still passes with zoomAnimation: false and this fix.

Falke-Design (approved):
@mourner I added the test with `zoomAnimation: false`.

The current lint error in the main will be fixed with this to.

jonkoops (approved):

Falke-Design:
The reason why this is working is because dragging is prevented:
https://github.com/Leaflet/Leaflet/blob/9793d0ab1ca545e5a578c9451f84a908d3e7ce08/src/dom/Draggable.js#L81

veitbjarsch:
Would love to see this in the next release. +1

PR #8796, opened 2023-01-13
Remove deprecated `which` and change `button` to main-button

Falke-Design:
To determine the left mouse button we used [which](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which) but it is deprecated. So we need to use [button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button). We already have it implemented but for some reason we had `exit if button !== 1` which means it is only true if the middle mouse button is used.

`which`:

    0: No button
    1: Left button
    2: Middle button (if present)
    3: Right button

`button`:

    0: Main button pressed, usually the left button or the un-initialized state
    1: Auxiliary button pressed, usually the wheel button or the middle button (if present)
    2: Secondary button pressed, usually the right button
    3: Fourth button, typically the Browser Back button
    4: Fifth button, typically the Browser Forward button

@mourner this was already implemented a long time ago but you reverted it, do you maybe have an idea why?
grafik

https://github.com/Leaflet/Leaflet/commit/0d1716d7964491ae894b58ef006cda9dd07d078f
https://github.com/Leaflet/Leaflet/commit/21a748123f7411ed8a0be5824e2366b71c7b6b8b

### DIFF
--- a/spec/suites/map/handler/Map.TouchZoomSpec.js
+++ b/spec/suites/map/handler/Map.TouchZoomSpec.js
@@ -157,7 +157,7 @@ describe('Map.TouchZoom', () => {
 			.down().moveBy(-200, 0, 500).up(100);
 	});
 
-	it.skipIfNotTouch("Layer is rendered correctly while pinch zoom when zoomAnim is true", (done) => {
+	it.skipIfNotTouch('Layer is rendered correctly while pinch zoom when zoomAnim is true', (done) => {
 		map.remove();
 
 		map = new L.Map(container, {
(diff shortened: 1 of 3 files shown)


mourner:
@Falke-Design yep, traced it — reverted because it broke IE8, but we don't care anymore. https://github.com/Leaflet/Leaflet/pull/3287#issuecomment-82901827

mourner (approved):

jonkoops:
Loving all this old cruft removal :muscle:

PR #8800, opened 2023-01-15
Drop  `__super__` property from Class

jonkoops:
Removes the `__super__` property from classes. This [was used in CoffeeScript](https://github.com/Leaflet/Leaflet/commit/c3ee79b01d44484745fe59e91be88b1b502bd80e) to determine the super class in their own class system, but has since been superseded by [standardized ECMAScript classes](https://coffeescript.org/#classes).

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -28,7 +28,7 @@ Class.extend = function (props) {
 		this.callInitHooks();
 	};
 
-	const parentProto = NewClass.__super__ = this.prototype;
+	const parentProto = this.prototype;
 
 	const proto = Object.create(parentProto);
 	proto.constructor = NewClass;
@@ -37,7 +37,7 @@ Class.extend = function (props) {
 
 	// inherit parent's statics
 	for (const i in this) {
-		if (Object.hasOwn(this, i) && i !== 'prototype' && i !== '__super__') {
+		if (Object.hasOwn(this, i) && i !== 'prototype') {
 			NewClass[i] = this[i];
 		}
 	}


src/core/Class.js
```diff
@@ -28,7 +28,7 @@ Class.extend = function (props) {
 		this.callInitHooks();
 	};
 
-	const parentProto = NewClass.__super__ = this.prototype;
```
IvanSanchez, line 31: That's `__super__`, not `__proto__`!

IvanSanchez:
This dates back to https://github.com/Leaflet/Leaflet/commit/c3ee79b01d44484745fe59e91be88b1b502bd80e - it's not a deprecated JS feature, but a coffeescript convention instead. I guess it's meant to keep track of whether a class is a `L.Class` or a vanilla JS `function`/`class`/`Object`.

src/core/Class.js
jonkoops: Looks like I didn't have my morning coffee hahah

jonkoops:
Ahhh, looks like I wasn't paying attention when going through this code and I mistook the two. I'm working to refactor the class system so we can make use of standardized classes in some manner.

I'd still like to get rid of this, as it looks like CoffeeScript has moved to [using standardized classes](https://coffeescript.org/#classes) since version 2.

IvanSanchez (approved):
It *should* be safe to remove `__super__`, since it's not being used anywhere else.

I'm not aware of any leaflet plugin/integration that depends on this, but I guess there's still a risk of something breaking somewhere.

Falke-Design (approved):

Falke-Design:
> I'm working to refactor the class system so we can make use of standardized classes in some manner.

@jonkoops is it still possible to `include` and `extend` a class?

IvanSanchez:
> @jonkoops is it still possible to `include` and `extend` a class?

Standard ES6 classes have `extend` built in. For Leaflet's `include`, it should be possible to port the current technique in the form of a static method that modifies the class prototype.

jonkoops:
> Standard ES6 classes have `extend` built in. For Leaflet's `include`, it should be possible to port the current technique in the form of a static method that modifies the class prototype.

Note that that the prototype of ECMAScript classes is non-configurable, so changing it's prototype will require some [tomfoolery](https://stackoverflow.com/questions/37680766/how-to-change-classs-prototype-in-es6#answer-37680991) (honestly a better API anyways).

> Is it still possible to include and extend a class?

Ideally I'd like to keep all of the existing functionality of our `Class` system and deprecate them so that we're not breaking things too much in the next major. We can consider them for removal in the next version.

However if I run into any impediments I will make sure to raise them so they can be discussed.

PR #8806, opened 2023-01-17
Use standardized ECMAScript classes for `Class`

jonkoops:
Refactors the `Class` API so that a [standardized ECMAScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes) class is used. This also allows classes to be sub-classed by using the new ECMAScript syntax:

```js
class Parent extends L.Class {
  initialize() {
    console.log("Parent initialize called.");
  }
}

class Child extends Parent {
  initialize() {
    super.initialize();
    console.log("Child initialize called.");
  }
}

Parent.addInitHook(function () {
  console.log("Parent init hook called.");
});

Child.addInitHook(function () {
  console.log("Child init hook called.");
});

const child = new Child();
```

As shown above `addInitHook()` continues to work as expected. There are however some things that do not work as one might be used to with the `Class.extend()` method.

## Statics

*Before*:
```js
const MyClass = L.Class.extend({
  statics: {
    FOO: "bar",
  },
});

MyClass.FOO; // 'bar'
```

*After:*
```js
class MyClass extends L.Class {
  static FOO = "";
}

MyClass.FOO; // 'bar'
```

## Includes (or mixins)

*Before:*
```js
const MyMixin = {
    foo: function () {},
    bar: 5
};

const MyClass = L.Class.extend({
    includes: MyMixin
});

const a = new MyClass();
a.foo();
```

*After:*
Passing in `includes` is no longer possible as this cannot be done when the class is being constructed (proposals welcome). However calling `includes()` statically is still possible:

```js
MyClass.include(MyMixin);
```

## Options

// TODO

Recommended reading:
- [`Object.setPrototypeOf()` (MDN)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf)
- [Inheritance and the prototype chain - Building longer inheritance chains (MDN)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#building_longer_inheritance_chains)

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -4,114 +4,112 @@ import * as Util from './Util';
 // @aka L.Class
 
 // @section
-// @uninheritable
 
 // Thanks to John Resig and Dean Edwards for inspiration!
 
-export function Class() {}
+export class Class {
+	_initHooksCalled = false;
 
-Class.extend = function (props) {
-
-	// @function extend(props: Object): Function
-	// [Extends the current class](#class-inheritance) given the properties to be included.
-	// Returns a Javascript function that is a class constructor (to be called with `new`).
-	const NewClass = function (...args) {
-
+	constructor(...args) {
 		Util.setOptions(this);
 
 		// call the constructor
 		if (this.initialize) {
-			this.initialize.apply(this, args);
+			this.initialize(...args);
 		}
 
 		// call all constructor hooks
 		this.callInitHooks();
-	};
-
-	const parentProto = this.prototype;
-
-	const proto = Object.create(parentProto);
-	proto.constructor = NewClass;
-
-	NewClass.prototype = proto;
-
-	// inherit parent's statics
-	for (const i in this) {
-		if (Object.hasOwn(this, i) && i !== 'prototype') {
-			NewClass[i] = this[i];
-		}
 	}
 
-	// mix static properties into the class
-	if (props.statics) {
-		Util.extend(NewClass, props.statics);
-	}
-
-	// mix includes into the prototype
-	if (props.includes) {
-		Util.extend.apply(null, [proto].concat(props.includes));
-	}
-
-	// mix given properties into the prototype
-	Util.extend(proto, props);
-	delete proto.statics;
-	delete proto.includes;
-
-	// merge options
-	if (proto.options) {
-		proto.options = parentProto.options ? Object.create(parentProto.options) : {};
-		Util.extend(proto.options, props.options);
-	}
-
-	proto._initHooks = [];
-
-	// add method for calling all hooks
-	proto.callInitHooks = function () {
-
+	callInitHooks() {
 		if (this._initHooksCalled) { return; }
 
-		if (parentProto.callInitHooks) {
-			parentProto.callInitHooks.call(this);
+		// collect all prototypes in chain
+		const protos = [];
+		let proto = Object.getPrototypeOf(this);
+
+		while (proto !== null) {
+			protos.push(proto);
+			proto = Object.getPrototypeOf(proto);
+		}
+
+		// reverse so the parent prototype is first
+		protos.reverse();
+
+		// call init hooks on each prototype
+		for (const proto of protos) {
+			const initHooks = proto._initHooks ?? [];
+			for (const hook of initHooks) { hook.call(this); }
 		}
 
 		this._initHooksCalled = true;
-
-		for (let i = 0, len = proto._initHooks.length; i < len; i++) {
-			proto._initHooks[i].call(this);
-		}
-	};
-
-	return NewClass;
-};
-
-
-// @function include(properties: Object): this
-// [Includes a mixin](#class-includes) into the current class.
-Class.include = function (props) {
-	const parentOptions = this.prototype.options;
-	Util.extend(this.prototype, props);
-	if (props.options) {
-		this.prototype.options = parentOptions;
-		this.mergeOptions(props.options);
 	}
-	return this;
-};
 
-// @function mergeOptions(options: Object): this
-// [Merges `options`](#class-options) into the defaults of the class.
-Class.mergeOptions = function (options) {
-	Util.extend(this.prototype.options, options);
-	return this;
-};
+	// @function extend(props: Object): Function
+	// [Extends the current class](#class-inheritance) given the properties to be included.
+	// Returns a Javascript function that is a class constructor (to be called with `new`).
+	static extend({statics, includes, ...props}) {
+		const NewClass = class extends this {};
 
-// @function addInitHook(fn: Function): this
-// Adds a [constructor hook](#class-constructor-hooks) to the class.
-Class.addInitHook = function (fn, ...args) { // (Function) || (String, args...)
-	const init = typeof fn === 'function' ? fn : function () {
-		this[fn].apply(this, args);
-	};
+		// hook up the static properties
+		Object.setPrototypeOf(NewClass, this);
 
-	this.prototype._initHooks = this.prototype._initHooks || [];
-	this.prototype._initHooks.push(init);
-	return this;
-};
+		// mix static properties into the class
+		if (statics) {
+			Util.extend(NewClass, statics);
+		}
+
+		const proto = NewClass.prototype;
+
+		// mix includes into the prototype
+		if (includes) {
+			Util.extend.apply(null, [proto].concat(includes));
+		}
+
+		// mix given properties into the prototype
+		Util.extend(proto, props);
+
+		// merge options
+		if (proto.options) {
+			proto.options = Object.create(this.prototype.options ?? {});
+			Util.extend(proto.options, props.options);
+		}
+
+		proto._initHooks = [];
+
+		return NewClass;
+	}
+
+	// @function include(properties: Object): this
+	// [Includes a mixin](#class-includes) into the current class.
+	static include(props) {
+		const parentOptions = this.prototype.options;
+		Util.extend(this.prototype, props);
+		if (props.options) {
+			this.prototype.options = parentOptions;
+			this.mergeOptions(props.options);
+		}
+		return this;
+	}
+
+	// @function mergeOptions(options: Object): this
+	// [Merges `options`](#class-options) into the defaults of the class.
+	static mergeOptions(options) {
+		Util.extend(this.prototype.options, options);
+		return this;
+	}
+
+
+	// @function addInitHook(fn: Function): this
+	// Adds a [constructor hook](#class-constructor-hooks) to the class.
+	static addInitHook(fn, ...args) { // (Function) || (String, args...)
+		const init = typeof fn === 'function' ? fn : function () {
+			this[fn](...args);
+		};
+
+		this.prototype._initHooks ??= [];
+		this.prototype._initHooks.push(init);
+		return this;
+	}
+}


src/core/Class.js
```diff
@@ -8,39 +8,68 @@ import * as Util from './Util';
 
 // Thanks to John Resig and Dean Edwards for inspiration!
 
-export function Class() {}
+export class Class {
+	// @function include(properties: Object): this
+	// [Includes a mixin](#class-includes) into the current class.
+	static include = function (props) {
```
IvanSanchez, line 14: Let's be consistent: `static include(props) { ... }`

src/core/Class.js
```diff
@@ -8,39 +8,68 @@ import * as Util from './Util';
...
+	}
+}
 
 Class.extend = function (props) {
 
 	// @function extend(props: Object): Function
 	// [Extends the current class](#class-inheritance) given the properties to be included.
 	// Returns a Javascript function that is a class constructor (to be called with `new`).
-	const NewClass = function (...args) {
+	class NewClass {
```
IvanSanchez, line 50: Try `class NewClass extends this.prototype { ... }` instead.

IvanSanchez (changes requested):

src/core/Class.js
IvanSanchez: Try also `const NewClass = class extends this.prototype` (a `class` *expression* instead of a `class` *statement*) - should make it an anonymous class and get rid of us seeing `NewClass` when debugging.

Falke-Design:
Looks pretty good for me. If we have no hurry I would like to test it with some "plugins" in the next days.

For me as a reminder to test: `instanceof`, `typeof`

jonkoops:
> Looks pretty good for me. If we have no hurry I would like to test it with some "plugins" in the next days.

We're in no rush, I wanted to get this out there to get some early feedback and explore our options. Feel free to test it against some plugins if you like, but things might still be prone to change.

src/core/Class.js
jonkoops: I used the following code to extend from the prototype:
```js
const NewClass = class extends this.prototype {};
```
Unfortunately this produces the following exception:
```
Uncaught TypeError: Class extends value #<Class> is not a constructor or null
```

I believe that since this method is `static` we should actually be using `this` (without `prototype`):
```js
const NewClass = class extends this {};
```
And after adding the constructor it looks as follows:
```js
const NewClass = class extends this {
	constructor(...args) {
		super(...args);
		Util.setOptions(this);

		// call the constructor
		if (this.initialize) {
			this.initialize.apply(this, args);
		}

		// call all constructor hooks
		this.callInitHooks();
	}
};
```
After this change the class extension seems to work, however this starts to create errors when running the tests:
```
Error: Map container is already initialized.
	at NewClass._initContainer (src/LeafletWithGlobals.js:3733:12)
	at NewClass.initialize (src/LeafletWithGlobals.js:2779:10)
	at new NewClass (src/LeafletWithGlobals.js:243:24)
	at Object.createMap [as map] (src/LeafletWithGlobals.js:4397:11)
	at Context.<anonymous> (spec/suites/control/Control.AttributionSpec.js:6:11)
```

I think I'll have to move some of this logic such as the `constructor` and `callInitHooks()` to `Class` directly, I'll explore that a little.

IvanSanchez:
This is going to be somewhat tricky to get right. Ultimately I'd like to get rid of all the `.extend()` calls, or at least move **all** the functionality to the `constructor` of the base `class Class`.

jonkoops:
@IvanSanchez this is what I have so far to hopefully facilitate this:

```js
import * as Util from './Util';

// @class Class
// @aka L.Class

// @section
// @uninheritable

// Thanks to John Resig and Dean Edwards for inspiration!

export class Class {
	// @function extend(props: Object): Function
	// [Extends the current class](#class-inheritance) given the properties to be included.
	// Returns a Javascript function that is a class constructor (to be called with `new`).
	static extend(props) {
		const NewClass = class extends this {};

		const parentProto = this.prototype;
		const proto = NewClass.prototype;

		// mix static properties into the class
		if (props.statics) {
			Util.extend(NewClass, props.statics);
		}

		// mix includes into the prototype
		if (props.includes) {
			Util.extend.apply(null, [proto].concat(props.includes));
		}

		// mix given properties into the prototype
		Util.extend(proto, props);
		delete proto.statics;
		delete proto.includes;

		// merge options
		if (proto.options) {
			proto.options = parentProto.options ? Object.create(parentProto.options) : {};
			Util.extend(proto.options, props.options);
		}

		proto._initHooks = [];

		return NewClass;
	}

	// @function include(properties: Object): this
	// [Includes a mixin](#class-includes) into the current class.
	static include(props) {
		const parentOptions = this.prototype.options;
		Util.extend(this.prototype, props);
		if (props.options) {
			this.prototype.options = parentOptions;
			this.mergeOptions(props.options);
		}
		return this;
	}

	// @function mergeOptions(options: Object): this
	// [Merges `options`](#class-options) into the defaults of the class.
	static mergeOptions(options) {
		Util.extend(this.prototype.options, options);
		return this;
	}

	// @function addInitHook(fn: Function): this
	// Adds a [constructor hook](#class-constructor-hooks) to the class.
	static addInitHook(fn, ...args) { // (Function) || (String, args...)
		const init = typeof fn === 'function' ? fn : function () {
			this[fn].apply(this, args);
		};

		this.prototype._initHooks = this.prototype._initHooks || [];
		this.prototype._initHooks.push(init);
		return this;
	}

	_initHooksCalled = false;

	constructor(...args) {
		Util.setOptions(this);

		// call the constructor
		if (this.initialize) {
			this.initialize.apply(this, args);
		}

		// call all constructor hooks
		this.callInitHooks();
	}

	callInitHooks() {
		if (this._initHooksCalled) { return; }

		if (super.callInitHooks) {
			super.callInitHooks.call(this);
		}

		this._initHooksCalled = true;

		for (let i = 0, len = this.prototype._initHooks.length; i < len; i++) {
			this.prototype._initHooks[i].call(this);
		}
	}
}

Class.prototype._initHooks = [];
```

This is still failing the tests, not sure why (all very much WIP). But it might allow us to at least do `X extends Class` internally. WDYT?

src/core/Class.js
```diff
@@ -8,110 +8,108 @@ import * as Util from './Util';
...
+				if (super.callInitHooks) {
+					super.callInitHooks.call(this);
+				}
 
-	const proto = Object.create(parentProto);
-	proto.constructor = NewClass;
+				this._initHooksCalled = true;
 
-	NewClass.prototype = proto;
+				const proto = NewClass.prototype;
```
jonkoops, line 40: Managed to move the `callInitHooks()` method into the class definition, but I need to find some kind of way to refer to the prototype here without statically referencing `NewClass`.

Falke-Design:
Still some failing tests but already "better":

`prototype` is not available in this context and `super` can't be called.

```
	callInitHooks(ctx) {
		if (this._initHooksCalled) { return; }

		const prototype = Object.getPrototypeOf(this);
		if(prototype && prototype.callInitHooks){
			// for some reason the context of the initial callInitHooks call has the same _initHooks but is a different object.
			// I think this causes a duplicated execution of these hooks
			prototype.callInitHooks(ctx || this);
		}

		this._initHooksCalled = true;

		for (let i = 0, len = this._initHooks.length; i < len; i++) {
			this._initHooks[i].call(this);
		}
	}
```

With adding `proto._initHooksCalled = false;` to `extend` the  `Class` tests are successful but `Canvas` not anymore.

jonkoops:
The 'problem' is that `this` will always refer to the child class, and never the parent class it makes which impossible to get the prototype of the 'current' class.

As [per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain#inheriting_methods):
> When an inherited function is executed, the value of [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) points to the inheriting object, not to the prototype object where the function is an own property.


So I think any approach that goes against the stream is here is kinda dead on arrival. I was thinking what we could do instead is place the `callInitHooks()` in parent class `Class`, and directly traverse the prototype itself in reverse. This would replicate the existing functionality, but remove the need for super calls.

jonkoops:
This seems to work, WDYT @Falke-Design @IvanSanchez?
```js
callInitHooks() {
	if (this._initHooksCalled) { return; }

	const protos = [];
	let proto = Object.getPrototypeOf(this);

	while (proto !== null) {
		protos.push(proto);
		proto = Object.getPrototypeOf(proto);
	}

	protos.reverse();

	protos.forEach((proto) => {
		const initHooks = proto._initHooks ?? [];
		initHooks.forEach((hook) => { hook.call(this); });
	});

	this._initHooksCalled = true;
}
```

IvanSanchez:
Remember https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super#super-calling_static_methods , folks.

jonkoops:
@IvanSanchez I plan to move this from `NewClass` into the `Class` prototype, so there would be no need for a `super()` call anymore.

jonkoops:
Added the refactored method, however moving stuff to `Class` still breaks the tests :thinking:

src/core/Class.js
```diff
@@ -8,110 +8,116 @@ import * as Util from './Util';
...
 
-	// inherit parent's statics
-	for (const i in this) {
-		if (Object.hasOwn(this, i) && i !== 'prototype') {
-			NewClass[i] = this[i];
+				// call init hooks on each prototype
+				protos.forEach((proto) => {
+					const initHooks = proto._initHooks ?? [];
+					initHooks.forEach(hook => hook.call(this));
+				});
```
mourner, line 50: Nit: avoid closures, prefer the more idiomatic `of`.

```suggestion
				for (const proto of protos) {
					const initHooks = proto._initHooks ?? [];
					for (const hook of initHooks) hook.call(this);
				}
```

src/core/Class.js
```diff
@@ -8,110 +8,116 @@ import * as Util from './Util';
...
+		// mix static properties into the class
+		if (props.statics) {
+			Util.extend(NewClass, props.statics);
+		}
 
-	// add method for calling all hooks
-	proto.callInitHooks = function () {
+		// mix includes into the prototype
+		if (props.includes) {
+			Util.extend.apply(null, [proto].concat(props.includes));
```
mourner, line 73: Nit: rest args where possible?
```suggestion
			Util.extend(proto, ...props.includes);

```

src/core/Class.js
jonkoops: Done 👍

Falke-Design:
> however moving stuff to `Class` still breaks the tests 🤔

How does the code look like then?

src/core/Class.js
jonkoops: 933fa76fb87782a618b866cb920079df7463c2e3 takes care of this.

jonkoops:
Ok, I managed to move the hook initialization to the base class `Class`. Also `NewClass` now properly uses the `extends` keywork to extend `Class`.

src/core/Class.js
jonkoops: Managed to accomplish this in 3c65e75d90ae098aef7dff4b09609b2314b50ac9.

jonkoops:
@Falke-Design @IvanSanchez @mourner I think we're in a pretty good place right now. Can I ask you guys to take another look? It's now possible to extend classes with initialization hooks and all using standard ECMAScript syntax:

```js
class Parent extends L.Class {
  initialize() {
    console.log("Parent initialize called.");
  }
}

class Child extends Parent {
  initialize() {
    super.initialize();
    console.log("Child initialize called.");
  }
}

Parent.addInitHook(function () {
  console.log("Parent init hook called.");
});

Child.addInitHook(function () {
  console.log("Child init hook called.");
});

const child = new Child();
```

I am trying to document various differences and caveats in the description, feel free to supplement it with your findings. I am getting a good night sleep 😴

src/core/Class.js
```diff
@@ -4,114 +4,112 @@ import * as Util from './Util';
...
+	_initHooksCalled = false;
 
+	constructor(...args) {
 		Util.setOptions(this);
 
 		// call the constructor
 		if (this.initialize) {
-			this.initialize.apply(this, args);
+			this.initialize(...args);
 		}
```
Falke-Design, line 19: Is it correct to "convert" args twice?

src/core/Class.js
```diff
@@ -4,114 +4,112 @@ import * as Util from './Util';
...
 	}
 
-	// mix includes into the prototype
-	if (props.includes) {
-		Util.extend.apply(null, [proto].concat(props.includes));
-	}
+	// @function extend(props: Object): Function
+	// [Extends the current class](#class-inheritance) given the properties to be included.
+	// Returns a Javascript function that is a class constructor (to be called with `new`).
+	static extend({statics, includes, ...props}) {
```
Falke-Design, line 52: Should be the new default way `class Parent extends L.Class` or `L.Class.extend()`? 
If they are resulting in the exact same, then I would say we mark this function als deprecated. 
But for me it looks like they are different. With `.extend()` we apply `statics` and `includes` over `extends L.Class` we didn't. Or do I oversee something?

IvanSanchez: I would like the default to be `class Parent extends L.Class`, with `.extend()` left just for compatibility.

> With `.extend()` we apply `statics` and `includes` over `extends L.Class` we didn't. Or do I oversee something?

That's the current behaviour if I understand the code correctly. I think it'd be possible to apply `statics` and `includes` and `options` in the `constructor()` as well.

src/core/Class.js
jonkoops: Haven't found a functional way to modernize this, perhaps we leave it as is. WDYT @mourner?

src/core/Class.js
jonkoops: Yeahs should be no issue. The `args` are coming in as parameters hence the spread in the constructor to ensure we get an array. Then when calling initialize we want to pass them in the same manner, so they are spread again.

src/core/Class.js
jonkoops: The new default should become the ECMAScript syntax. However I have yet to find a way to apply the `statics`, `includes` and `options` as they all modify the prototype. This kind of modification should happen when the class is defined (e.g. `Class.extends()`), and not when it is instantiated.

Not sure if we can land this in a backwards compatible manner, but perhaps we could provide static methods on `Class` to emulate these APIs. We already have `Class.include()` so we could create a `Class.addOptions()` and `Class.addStatics()` method.

Any thoughts?

src/core/Class.js
mourner: Yeah, let's leave as is if it's not an easy fix.

jonkoops: Ok, I will consider this resolved then.

src/core/Class.js
jonkoops: Marking this as resolved so we can focus on the open issues.

jonkoops:
Just thinking out loud, but we could also provide [class decorators](https://2ality.com/2022/10/javascript-decorators.html) for example:

```js
import { Class, includes } from 'leaflet';
import MyMixin from './mixins';

@includes(MyMixin)
class MyClass extends Class { }
```

Or with options:
```js
@options({
  myOption1: 'foo',
  myOption2: 'bar'
})
class MyClass extends Class { }
```

jonkoops:
I'm converting this PR to a draft. I think it's a good place to work on and discuss this feature, but I see more benefit in chunking the work for this PR out into several smaller pieces of work that are easier to review.

IvanSanchez:
Closing this, since all the changes seem to have been merged via five individual PRs.

jonkoops:
Yeah, there are still a couple things left. But I feel this PR does not serve a useful purpose at this point.

PR #8810, opened 2023-01-22
Replace animation frame polyfill with native API

jonkoops:
Replaces the custom animation frame API from `Util` with the standardized methods. These are now available in all browsers within our support target, so there is no longer required to use a polyfill.

This removes the following APIs:
- `Util.requestAnimFrame()`
- `Util.cancelAnimFrame()`
- `Util.requestFn`
- `Util.cancelFn`

### DIFF
--- a/debug/vector/moving-canvas.html
+++ b/debug/vector/moving-canvas.html
@@ -45,8 +45,8 @@
 					]);
 				});
 
-			requestAnimationFrame(update);
-		}
+				requestAnimationFrame(update);
+			}
 
 			update();
 		</script>
--- a/debug/vector/moving-canvas.html
+++ b/debug/vector/moving-canvas.html
@@ -45,8 +45,8 @@
 					]);
 				});
 
-				requestAnimationFrame(update);
-			}
+			requestAnimationFrame(update);
+		}
 
 			update();
 		</script>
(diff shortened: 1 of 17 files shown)


IvanSanchez (approved):
Ah, one more "goodbye internet explorer" change. This is gonna break plugins for sure, but needed to be done.

jonkoops:
I am a bit at a loss as to why the tests are failing, it appears to be a rather strange bug. I am able to strip the code from `main` down to the following and the tests will still pass:

```js
const requestFn = requestAnimationFrame;
const cancelFn = cancelAnimationFrame;

export function requestAnimFrame(fn, context) {
  return requestFn(fn.bind(context));
}

export function cancelAnimFrame(id) {
  cancelFn(id);
}
```

However, when I refactor away `requestFn` and `cancelFn` so calls are made directly the tests seem to fail:
```js
export function requestAnimFrame(fn, context) {
  return requestAnimationFrame(fn.bind(context));
}

export function cancelAnimFrame(id) {
  cancelAnimationFrame(id);
}
```

jonkoops:
I believe it might have something to do with how Sinon handles [fake timers](https://github.com/sinonjs/fake-timers/blob/main/README.md).

jonkoops:
@mourner related to your recent changes under #9241 and #9236.

Falke-Design:
I don't understand why the test `auto invalidateSize after container resize` is failling on the server but not locally ...

Falke-Design (approved):

jonkoops:
Nice work! Thanks for getting this landed @Falke-Design!

PR #8823, opened 2023-02-04
Destructure `statics` and `includes` from class properties

jonkoops:
Destructures `statics` and `includes` from class properties. This avoids the need to delete these fields from the prototype after extending.
 

Falke-Design (approved):
Fine for me. Let us merge this, no need to wait for other approvals, because it is no breaking change.

PR #8824, opened 2023-02-04
Use `Object.setPrototypeOf()` to extend `Class`

jonkoops:
Changes the logic for inheritance of `Class` to use the more modern [`Object.setPrototypeOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf). The same method is now used to 'inherit' the static properties of a parent class, rather than copying it over manually.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -28,19 +28,14 @@ Class.extend = function (props) {
 		this.callInitHooks();
 	};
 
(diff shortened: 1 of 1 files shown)


Falke-Design (approved):
Fine for me

PR #8825, opened 2023-02-04
Move `callInitHooks()` to prototype of `Class`

jonkoops:
Moves the `callInitHooks()` method to the prototype of `Class` instead of defining it every time a class is `extend()`ed. The method is also re-implemented to remove the reliance on variables defined in the scope of `extend()`.


Falke-Design (approved):
Fine for me

PR #8826, opened 2023-02-05
Drop UMD and make ESM the default entrypoint

jonkoops:
Drops the UMD (Universal Module Definition) distribution and makes ESM (ECMAScript modules) distribution the default. Also changes the default [type of the package](https://nodejs.org/api/packages.html#type) to `module` so that all code in the repository is evaluated as ESM by Node.js.

Since ESM is now the default all files with the `.mjs` extension have been renamed to `.js`. Some configuration files such as those for Karma and ESLint do not yet support ESM fully, so these have been renamed to `.cjs` to remain functional.


debug/tests/esm.html
```diff
@@ -1,19 +0,0 @@
...
-    <div id="map"></div>
-    <script type="module">
-      import L from '../../dist/leaflet-src.esm.js';
-
-      var map = L.map('map', {
-        center: [39.84, -96.591],
-        zoom: 4,
-      });
-
-      L.tileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(map);
```
Falke-Design, line 16: why deleting this file? Only the import path should be changed

Falke-Design (changes requested):

debug/tests/esm.html
jonkoops: Yeah, I guess there is no harm in keeping this around. Restored the file and changed the path.

Falke-Design (changes requested):
Every file in `/debug/` doesn't work anymore, they need all to be changed to a module.
`spec/index.html` doesn't work anymore too.

jonkoops:
Hmmm, yeah. Looks like I'll have to port those over first. I wanted to clean up that code anyways so I will open up some PRs to do so.

jonkoops:
Converting this to a draft until I have all the aforementioned issues resolved.

Falke-Design:
@jonkoops @IvanSanchez @mourner please review

jonkoops (commented):
LGTM, but I cannot approve my own PRs. @Falke-Design if this looks good enough to you, feel free to approve and merge it.

Falke-Design (approved):

Falke-Design:
@mourner @IvanSanchez  the publish workflow is failing, can you fix it as admins. We deleted the file `leaflet-src.esm.js`

grafik
https://github.com/Leaflet/Leaflet/actions/runs/13902673018/job/38899588346

IvanSanchez:
@jonkoops @Falke-Design  I'm looking at https://github.com/Leaflet/Leaflet/blob/7ac98758d4eb5c2389afa01566aeb4c352abe026/.github/workflows/main.yml#L181 , but I know nothing about how S3 works (IIRC it was @mourner who set the S3 accounts long time ago).

mourner:
Updated the S3 bucket so that there's no error, but now it complains that there's no `dist` folder to upload. 😅 I think we can just remove the job, and make sure to update the download page (and rename it to "Get started", "Install" or something) after v2 is out.

PR #8830, opened 2023-02-06
Convert canvas debug page to ESM

jonkoops:
Converts the canvas debug page to ESM and cleans up the code so it's bit more modern.

### DIFF
--- a/debug/map/canvas.html
+++ b/debug/map/canvas.html
@@ -1,48 +1,44 @@
-<!DOCTYPE html>
-<html>
-<head>
-	<title>Leaflet debug page</title>
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - Canvas</title>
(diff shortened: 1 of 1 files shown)


debug/map/canvas.html
```diff
@@ -1,48 +1,44 @@
...
+		<link rel="stylesheet" href="../css/screen.css" />
+	</head>
+	<body>
+		<div id="map"></div>
+		<script type="module">
+			import {GridLayer, Map, LatLng} from '../../dist/leaflet-src.esm.js';
+
+			const tiles = new GridLayer();
+
+			tiles.createTile = function (coords) {
```
Falke-Design, line 17: Only thinking out loud: Is it better to refactor this and make it extending the GridLayer class?

Falke-Design (approved):

debug/map/canvas.html
jonkoops: Could be an option, we're doing this in a lot of places so I don't know if it's best practice?

PR #8831, opened 2023-02-06
Provide file extensions when running ESLint

jonkoops:
Adds the extensions for ESLint explicitly when running the `lint` task. The default here is `.js` only, this caused some linting issues to slip through (which have been fixed here).


jonkoops:
Gonna merge this in, as I need it for some other PRs. Let me know if there is any feedback.

PR #8834, opened 2023-02-06
Convert controls debug page to ESM

jonkoops:
Converts the controls debug page to ESM and cleans up the code so it's bit more modern.

### DIFF
--- a/debug/map/controls.html
+++ b/debug/map/controls.html
@@ -1,45 +1,41 @@
-<!DOCTYPE html>
-<html>
-<head>
-	<title>Leaflet debug page</title>
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - Control</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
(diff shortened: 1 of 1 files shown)


debug/map/controls.html
```diff
@@ -1,45 +1,41 @@
...
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - Control</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
+		<link rel="stylesheet" href="../../dist/leaflet.css" />
+		<link rel="stylesheet" href="../css/screen.css" />
+	</head>
+	<body>
+		<div id="map"></div>
+		<script type="module">
+			import {TileLayer, Map, Marker, LatLng, Control} from '../../dist/leaflet-src.esm.js';
```
Falke-Design, line 13: We suggested most of the time that the users should use the factory methods like `L.tileLayer`, is this still possible?

Falke-Design (approved):

debug/map/controls.html
jonkoops: Certainly still possible, but we might want to revisit that convention as it's really just a Leaflet practice.

Issue #8352, opened 2022-07-24
Update documentation to reflect 'ESM first' approach


jonkoops:
Now that we have landed support for ES modules in #8329 the documentation should be updated to reflect this as the preferred approach to use Leaflet. This includes:

- Replacing usages of the global `L` with imports (preferably named)
- Updating installation instructions to prefer loading Leaflet as an import from CDNs
- Possibly recommending [Import Maps](https://caniuse.com/import-maps) (needs more browser adoption).

patrickarlt:
I would be happy to take this on if you wanted some help. I've documented other projects with [the SkyPack CDN](https://www.skypack.dev/) which provides a CDN optimized for delivering ES modules. Once a beta/rc is on NPM it should be pretty strightforward to get the doc updated.

jonkoops:
@patrickarlt feel free to give it a first draft if you like, I think it's good to start putting some ideas out there. Most likely this is something we'd try to land in multiple PRs, starting with the [download](https://leafletjs.com/download.html) and [quickstart](https://leafletjs.com/examples/quick-start/) pages.

We'll also have to make some changes so that this documentation is not immediately published to from `main`. Perhaps we'll have to deploy this to another domain (e.g. `next.leafletjs.com`). @mourner @Falke-Design WDYT?

Falke-Design:
Yeah fine for me. But I don't know if it is possible to deploy from different branches to different domains over Github.

Or maybe `v2.leafletjs.com`?

jonkoops:
I think `v2.leafletjs.com` would be nice yeah, @mourner is this something we can arrange?

mourner:
@jonkoops that'd be great! Need to figure out how to do it...

IvanSanchez:
@mourner @jonkoops It seems that we're gonna need a new repo just for `v2.leafletjs.com` (assuming we want to use github pages).

Having several branches of the same repo as different instances of github pages has been asked, and the answer was "no". See https://github.com/orgs/community/discussions/21582 .

It would be posible to do something more convoluted, such as a gitlab mirror of the whole repo, and using gitlab pages for a branch from the mirror. Unfortunately I don't know which approach is the cleanest one / the one with the lowest friction.

Falke-Design:
What if we make a subfolder `leafletjs.com/v2/` instead of a subdomain (`v2.leafletjs.com`)?

jonkoops:
[...]

PR #8837, opened 2023-02-07
Enforce `.js` file extension for module imports

jessetane:
Hi all, I know this is a massive PR and so I'm not necessarily expecting you'll want to merge it but I thought it might be worth pointing out to those who might be interested (https://github.com/Leaflet/Leaflet/issues/8352).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,7 +21,8 @@ module.exports = {
 	},
 	extends: 'mourner',
 	plugins: [
-		'@mapbox/eslint-plugin-script-tags'
+		'@mapbox/eslint-plugin-script-tags',
+		'eslint-plugin-import'
 	],
 	parserOptions: {
 		ecmaVersion: 'latest',
@@ -30,6 +31,7 @@ module.exports = {
 	rules: {
 		'consistent-return': 'off',
 		'curly': 'error',
+		'import/extensions': ['error', 'ignorePackages'],
 		'indent': ['error', 'tab', {VariableDeclarator: 0, flatTernaryExpressions: true}],
 		'key-spacing': 'off',
 		'linebreak-style': ['off', 'unix'],
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "bundlemon": "^2.0.0",
     "eslint": "^8.23.1",
     "eslint-config-mourner": "^3.0.0",
+    "eslint-plugin-import": "^2.27.5",
     "husky": "^8.0.1",
     "karma": "^6.4.1",
     "karma-chrome-launcher": "^3.1.1",
--- a/src/Leaflet.js
+++ b/src/Leaflet.js
@@ -3,22 +3,22 @@ import {version} from '../package.json';
 export {version};
 
 // control
-export * from './control/index';
+export * from './control/index.js';
 
 // core
-export * from './core/index';
+export * from './core/index.js';
 
 // dom
-export * from './dom/index';
+export * from './dom/index.js';
 
 // geometry
-export * from './geometry/index';
+export * from './geometry/index.js';
 
 // geo
-export * from './geo/index';
+export * from './geo/index.js';
 
 // layer
-export * from './layer/index';
+export * from './layer/index.js';
 
 // map
-export * from './map/index';
+export * from './map/index.js';
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "bundlemon": "^2.0.0",
         "eslint": "^8.23.1",
         "eslint-config-mourner": "^3.0.0",
+        "eslint-plugin-import": "^2.27.5",
         "husky": "^8.0.1",
         "karma": "^6.4.1",
         "karma-chrome-launcher": "^3.1.1",
@@ -401,6 +402,12 @@
       "integrity": "sha512-EYNwp3bU+98cpU4lAWYYL7Zz+2gryWH1qbdDTidVd6hkiR6weksdbMadyXKXNPEkQFhXM+hVO9ZygomHXp+AIw==",
       "dev": true
     },
+    "node_modules/@types/json5": {
+      "version": "0.0.29",
+      "resolved": "https://registry.npmjs.org/@types/json5/-/json5-0.0.29.tgz",
+      "integrity": "sha512-dRLjCWHYg4oaA77cxO64oO+7JwCwnIzkZPdrrC71jQmQtlhM556pwKo5bUzqvZndkVbeFLIIi+9TC40JNF5hNQ==",
+      "dev": true
+    },
     "node_modules/@types/lodash": {
       "version": "4.14.186",
       "resolved": "https://registry.npmjs.org/@types/lodash/-/lodash-4.14.186.tgz",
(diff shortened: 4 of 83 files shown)


jessetane:
In its current state, Leaflet is REALLY close to working out of the box without any bundling via `import * as L from '/node_modules/leaflet/src/Leaflet.js'` or even just `import * as L from 'leaflet'` with an import-map. Personally, I find not having to mess with build tooling makes the development experience just that much more magical. All that's needed is to import real file names (use .js extension) and avoid attempting to import JSON files (which are not ES modules).

My first commit here just adds the .js extensions (easy enough with a sed one liner), and the second hardcodes the version number in Leaflet.js, and that's it!

jonkoops:
Hi @jessetane, we're definitely interested in landing as close as possible support for ESM without the need for bundling. Part of this is adding the `.js` extensions. Before we land something like this I'd like to enforce this like a configuration for Rollup and ESLint rules, could you look into that?

jessetane:
I'm not super familiar with eslint but it looks like this plugin https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md does what you want, see [0376e85](https://github.com/Leaflet/Leaflet/pull/8837/commits/0376e8548d2242a83719f3824d41b286e026b561)

jessetane:
Oops, that got a little weird due to package-lock being checked in... if you want the lockfile change in a commit I can send it, just lmk

jessetane:
@jonkoops I saw https://github.com/Leaflet/Leaflet/pull/8838 and realized a file server is necessary to demonstrate "native" module loading as I was describing. I just pushed two more commits here as an example.

.eslintrc.js
```diff
@@ -30,6 +31,7 @@ module.exports = {
 	rules: {
 		'consistent-return': 'off',
 		'curly': 'error',
+		'import/extensions': ['error', 'ignorePackages'],
```
jonkoops, line 34: The documentation is not very clear to me on this, but why could this not be `always`?

debug/tests/esm.html
```diff
@@ -6,7 +6,7 @@
   <body>
     <div id="map"></div>
     <script type="module">
-      import L from '../../dist/leaflet-src.esm.js';
+      import * as  L from '../../src/Leaflet.js';
```
jonkoops, line 9: For now I'd like to leave this out of scope. Actually loading from source is problematic as we also have circular references. 
```suggestion
      import L from '../../dist/leaflet-src.esm.js';
```

package.json
```diff
@@ -42,6 +44,7 @@
     "CHANGELOG.md"
   ],
   "scripts": {
+    "debug": "serve -n",
```
jonkoops, line 47: Let's leave this out for now, should be fine just to open files on the local filesystem or use something like `npx http-server`.

src/Leaflet.js
```diff
@@ -1,24 +1,23 @@
 
-import {version} from '../package.json';
-export {version};
+export const version = '1.9.2';
```
jonkoops, line 2: I don't quite want to hard code this just yet. Let's leave it as it was before for now.

jonkoops (changes requested):
@jessetane I'd like to narrow this PR down a bit to 'just' changing the way modules are imported internally to be conformant to the standard. There is a lot of separate refactoring I'd like to get done before even considering shipping ESM natively from the source (if at all).

debug/tests/esm.html
jessetane: What's the problem with circular references? In my experience CR's are only an issue if circular access is needed synchronously at initialization time. Admittedly I did not port the entire test suite to confirm but when I tried this and a few other examples in the debug directory Leaflet seemed to work fine, and code that only works with a non-standard build tool is exactly the kind of thing I'm suggesting we take effort to avoid!

package.json
jessetane: Ya, this was just meant as an example since ES modules cannot be loaded from the filesystem by web browsers.

src/Leaflet.js
jessetane: I don't like hardcoding the version number either... the point is just that this code is non-standard, JSON files cannot be imported by web browsers because they are not ES modules.

.eslintrc.js
jessetane: imports from 3rd party modules should mapped using an [import-map](https://github.com/WICG/import-maps) instead of hardcoding filesystem paths that depend on the particular package manager being used

jessetane:
Happy to trim this PR down if any of the particular commits are useful to you - sounds like just [72da14e](https://github.com/Leaflet/Leaflet/pull/8837/commits/72da14e3317c351facaf8fe34ff2fb9bcf1fd714) and [8a5d712](https://github.com/Leaflet/Leaflet/pull/8837/commits/8a5d7128bd914547fed61f31f4e74c40355a471f) at this point? Like I say this was mostly about bringing the topic up for discussion!

.eslintrc.js
jessetane: i mean leaflet is dep free right now so I guess `always` makes sense too.. but trying to lint extensions in the first place seems funny to me, since it's not really about the extension so much as whether the file can be fetched from an http server. bare specifiers should also work when accompanied by an import-map for the reason described above.

jessetane: if the intent is to be able to see if native module loading actually works, having at least one example in debug load natively still seems like a more accurate way to start to me. but yeah, unless I'm missing something I'm pretty sure we would need some http server for that to work (zero attachment to serve, I like ecstatic for dev stuff although it's no longer maintained)

jonkoops:
Yeah, that should pretty much cover it. Thanks @jessetane!

jessetane:
rebased

src/Leaflet.js
jonkoops: This would require us to change our release process as well, which makes the change too bit for a single PR IMHO. All in due time.

package.json
jonkoops: For sure, this needs to be resolved. The development flow in this regard is less than ideal at the moment. But let's tackle that another time.

debug/tests/esm.html
jonkoops: I understand your enthusiasm to get this done, but I'd really like to give things a bit more time to bake. Moving Leaflet's distribution to ESM is already a big step forward. I also don't want to get rid the distribution just yet, as I am still on the fence over things like adding TypeScript support.

.eslintrc.js
jonkoops: The intent is to have native module loading work on the distribution first. Aligning towards ESM, such as having correct file extensions is already a step forward.

jonkoops (approved):

jonkoops:
Thanks @jessetane!

debug/tests/esm.html
jessetane: fwiw I was not necessarily suggesting doing away with dist, I just want to be able to import leaflet natively for a better dev experience in my consuming applications, and to be able to bundle upstream as desired knowing leaflet's source is standard javascript (can you tell I vote no on typescript? heh).

jonkoops: Ideally we don't adopt TypeScript. I'd prefer to adopt the [Type Annotations proposal](https://github.com/tc39/proposal-type-annotations), but that will likely take a couple of years before we can see it shipped. So pragmatism might be the reason to adopt TypeScript, at least for the short term.

jessetane: was not aware of that proposal, thanks for the link. i've just been bitten by things going out of style in js world enough times to have learned to favor standard / off-the-shelf stuff over the current hotness. ts has been around a while now but imho pragmatism in reusable modules means resisting the urge to go beyond the bleeding edge of what is standard at any given time. don't get me wrong, i love experimenting i just try to do it at the application level

jonkoops: Yeah, it's just that a lot of people (including myself) really appreciate to have type-safety as it simply eliminates a whole class of bugs. Pretty sure if we were to adopt this for Leaflet we'd find a couple of golden ones :)

jessetane: i hear that for sure, types are an amazing developer tool, but there are tradeoffs right now that don't seem worth it to me: back to build tool lock in, massive dev dep trees, less accessible code, difficulty debugging at runtime, false sense of security etc. perhaps some of these are misconceptions or interdependent, just misc observations from past encounters with ts

PR #8838, opened 2023-02-07
Convert map debug pages to ESM

jonkoops:
Converts the map debug pages (located in `debug/map`) to ESM and cleans up the code in the process. This is part of a greater effort to land ESM as a first-class citizen.


jonkoops:
Converted this to a draft, once all of the files have been converted I will mark it as ready for review. This way we can land it as a single feature rather than a bunch of PRs.

Falke-Design:
- [ ] `leaflet-src.esm.js` is not generating while `npm run watch`

jonkoops:
@Falke-Design I've created a PR to also run the ESM build in watch mode (see #8844). Since I've converted everything in the `debug/map` directory over I am marking this one ready for review.

Falke-Design (approved):
Very nice, thank you for the work!

I didn't tested them but I think you have opened every one, so ok for me to merge.

jonkoops:
> I didn't tested them but I think you have opened every one, so ok for me to merge.

Yeah I did, only issue I noticed is that some icons for the markers don't load. But this seems to be related to a bug in the icon URL detection code.

jonkoops:
Thanks for the quick review @Falke-Design!

PR #8839, opened 2023-02-09
add ScrollWheelZoom handler enableCenter

dchernyavsky:
Adds handler enableCenter to able to set ScrollWheelZoom option to 'center'

### DIFF
--- a/src/map/handler/Map.ScrollWheelZoom.js
+++ b/src/map/handler/Map.ScrollWheelZoom.js
@@ -28,6 +28,16 @@ Map.mergeOptions({
 });
 
 export const ScrollWheelZoom = Handler.extend({
+    // @method enableCenter(): this
+	// Sets ScrollWheelZoom to 'center'
+	enableCenter() {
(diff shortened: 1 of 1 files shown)


jonkoops:
Could you provide an example of what this trying to fix or accomplish?

dchernyavsky:
Need to change scrollWheelZoom dynamically, but changing prop not working without re-load component.
image

Now working (and bad) solution is re-load MapContainer component
image
image

Better way change parameters dynamically with handlers like here
image

But exist handlers for scrollWheelZoom sets value only to true or false, not 'center'

dchernyavsky:
Another way to solve this problem is re-declare exist handler enable() to be able pass argument:
enable(value = true) {
		if (this._enabled === value || value !== true && value !== 'center') { return this; }

		this._enabled = value;
		this.addHooks();
		return this;
	},

Falke-Design:
@dchernyavsky just overwrite the map options: `map.options.scrollWheelZoom = 'center'`

PR #8848, opened 2023-02-13
Add dev dependency http-server for debugging ESM

jessetane:


Falke-Design:
I have nearly no experience with ESM, can you please explain why this is needed?

package.json
```diff
@@ -43,6 +44,7 @@
     "CHANGELOG.md"
   ],
   "scripts": {
+    "debug": "http-server --port 3000 -c-1",
```
jonkoops, line 47: I think we can fall back to the default port number here.
```suggestion
    "debug": "http-server -c-1",
```

jonkoops (changes requested):
@jessetane can you run `npm install` so the lockfile is also updated?

jonkoops:
> I have nearly no experience with ESM, can you please explain why this is needed?

The CORS policy for scripts loaded with `type=module` is different. So unlike regular scripts they will not load from the file-ystem directly. For example, if I open up one of the recently converted files in the debug section I get the following error:

image

Falke-Design:
> The CORS policy for scripts loaded with `type=module` is different. So unlike regular scripts they will not load from the file-ystem directly. For example, if I open up one of the recently converted files in the debug section I get the following error:

Ou ok thank you ... now i like it even less 😄

jessetane:
agree with you @Falke-Design that not being able to load modules directly from the filesystem is annoying... does anyone know why this limitation exists? i assume there is a good reason but it's not immediately obvious to me and it didn't jump out on a quick glance at the [spec](https://tc39.es/ecma262/#sec-modules). see also: https://github.com/whatwg/html/issues/8121. maybe this will change in the future?

jessetane:
more here https://github.com/whatwg/html/issues/1888

jessetane:
the arguments there are pretty esoteric, not sure i can fully grasp them without more study but i guess i still prefer being forced to use an http server than non-standard build tools

jonkoops:
It is what it is, having a HTTP server running is not the end of the world 😉

jonkoops (approved):
LGTM

Issue #8158, opened 2022-04-18
SVG images for marker and layer control


maciek-szn:
I'm not sure if it's intended or not - I couldn't find anything about that except for #7600. `src/images` folder contains SVG images for marker and layer control. However, v1.8.0 still uses PNG images.

mourner:
It's intended. Why do you need SVG in place of images by default? Images work perfectly fine, they're better for performance, they don't need to be scaled, and they're supported by legacy browsers Leaflet still supports. You can still use custom SVG markers if needed.

Malvoz:
I like SVGs, they look great, depending on the implementation they could be easily stylable, and might even save some byte size (e.g. [this icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aplace%3A) is 300 bytes, versus current 1.43 kb).

Also, for v2 I think it's not inconceivable for default controls/icons to get a bit of a face lift? @maciek-szn do you want to suggest this feature in https://github.com/Leaflet/Leaflet/discussions/7729?

Malvoz:
Closing per above comments, feel free to continue discussing here or in https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-2599910.

PR #8855, opened 2023-02-20
Adding colors to Marker - Default Icon attempt

Falke-Design:
Based on #8766 and alternative to #8779

Adding colors to default icon over an `css filter`:

grafik

Colors: Pink, DarkBlue, Lila, Magenta, Red, Orange, Brown, Green, Mint

- `L.Icon.Default.Red()`
- `L.Icon.Default.DarkBlue()`
- `L.Icon.Default.Green()`
- ....

```
new Marker(latlng, {icon: new Icon.Default.Red()}).addTo(map);
```

The difference to #8766 is only `index.js` and `leaflet.css`.

@IvanSanchez can you please add the documation for the subclasses, if this will be get merged?


IvanSanchez:
@Falke-Design I think I messed up the "fetch a branch from a PR" thing. See https://github.com/Leaflet/Leaflet/commit/a3e6f3961785455d0f67f617683eaf96caea593e .

Falke-Design:
I just copied the docu from your commit. I hope this ok.

For me it works pretty easy to use Github CLI to edit PRs:

```
1. copy gh cli command over the code button in the top: gh pr checkout 8855
2. make changes
3. commit
4. git push --force
```

src/layer/marker/index.js
```diff
@@ -2,6 +2,48 @@ import {Icon} from './Icon.js';
...
+ * @property Red: L.Icon
+ * The default icon, but tinted red
+ * @property Orange: L.Icon
+ * The default icon, but tinted orange
+ * @property Brown: L.Icon
+ * The default icon, but tinted brown
+ * @property Green: L.Icon
+ * The default icon, but tinted green
+ * @property Mint: L.Icon
+ * The default icon, but tinted bluish green
```
Falke-Design, line 44: Should we sort it alphabetical?

src/layer/marker/index.js
```diff
@@ -2,6 +2,48 @@ import {Icon} from './Icon.js';
...
+ * @miniclass Colored default icons (Icon)
+ * @section
+ *
+ * `L.Icon.Default` has several static properties providing different coloured versions of the default icon.
+ *
+ * Usage is as: `L.marker(latlng, { icon: new L.Icon.Default.Green() })`
+ *
+ * Note that `L.Icon.Default.Blue` behaves the same as `L.Icon.Default`.
+ *
+ * @property Pink: L.Icon
```
Falke-Design, line 25: Or is the class `L.Icon.Default`?

Falke-Design (commented):
grafik

src/layer/marker/index.js
IvanSanchez: I'd rather have them ordered by hue.

src/layer/marker/index.js
IvanSanchez: That'd be correct, but I'd find that somewhat confusing, TBH. There's too many `L.Icon.Default` in that bit of docstring.

iamtekson:
It would be good to add `color` as a property, something similar to following code,
```js
L.Icon.Default(color='red');
L.Icon.Default(color='green');
L.Icon.Default(color='#0000ff');
``

File: src/layer/marker/index.js [hunk 3]
```diff
@@ -2,6 +2,48 @@ import {Icon} from './Icon.js';
...
+ * The default icon, but tinted red
+ * @property Orange: L.Icon
+ * The default icon, but tinted orange
+ * @property Brown: L.Icon
+ * The default icon, but tinted brown
+ * @property Green: L.Icon
+ * The default icon, but tinted green
+ * @property Mint: L.Icon
+ * The default icon, but tinted bluish green
+ */
```
Malvoz, line 45: Creative approach, but TBH I think this adds too many bytes for something that most wont use, it's already possible to use custom markers. Am I missing something?

Side note: OTOH if markers were `<svg>` instead of `` (which honestly I thought was the intention of https://github.com/Leaflet/Leaflet/issues/8158 / https://github.com/Leaflet/Leaflet/issues/8756) all it would require would be the option to change attributes (`fill`, `color`, etc.)?

Falke-Design: No we don't use the svg markers because #8766 is not merged. 

In my eyes we should make some things simpler for the users else we will lose more and more to other libraries. One of the things which can be really helpful (and is searched a lot), is how to change the color. I think this few bytes are worth it.

I would like to use the `stroke` and `fill` but we include the icons as a image and not as a svg element. I don't know how to change the css-classes / styles over a `` element

Vectorial1024:
Late join, but if we are concerned that we may be adding too much unpopular, rarely-used bytes to this lightweight package, then we can delegate this feature to a separate library.

I for one picked up an ancient library dating back to 2013 that's trying to do this, and right now I am working to add more color variations to the default Leaflet icon. See https://github.com/Vectorial1024/leaflet-color-markers/issues/3 .

If anything, we can redirect colored icon efforts away from Leaflet Core and let users choose whether they want these convenient colored icons (and therefore install the separate package), or don't want these (and therefore need not accept this PR).

PR #8857, opened 2023-02-21
Recommend my new plugin Leaflet.StarCircle

haelue:
I recommend my new plugin Leaflet.StarCircle, and add it into plugins.

### DIFF
--- a/docs/_plugins/markers-renderers/leaflet-starcircle.md
+++ b/docs/_plugins/markers-renderers/leaflet-starcircle.md
@@ -1,0 +1,12 @@
+---
+name: Leaflet.StarCircle
+category: markers-renderers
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Nice plugin but please update your Readme to use the same name everywhere and please consider to enable the demo page. Also I created a issue please check it: https://github.com/haelue/Leaflet.StarCircle/issues/1


1. Use the same name everywhere
grafik

2. Enable Github-Pages in the setting, then the demo can be visited over https://haelue.github.io/Leaflet.StarCircle/index.html

haelue:
@Falke-Design All modified as your wish!

Falke-Design (approved):

PR #8871, opened 2023-02-28
Convert `Class` to a JavaScript `class`

jonkoops:
Converts `Class`  into an actual [JavaScript `class`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes). This PR does not attempt to refactor the existing implementation in any other way aside from moving the methods into the class definition itself.

I tried to keep the diff as reasonable as possible, but Github is showing a lot of white-space changes. I would recommend to check out the branch and inspect the diff in your IDE.

Implements parts of #8806.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -8,113 +8,111 @@ import * as Util from './Util.js';
 
 // Thanks to John Resig and Dean Edwards for inspiration!
 
-export function Class() {}
-
-Class.extend = function ({statics, includes, ...props}) {
-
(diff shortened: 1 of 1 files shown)


mourner (approved):
Looks good 👍  GitHub has a "hide whitespace" option in the UI:

image

jonkoops:
> Looks good +1 GitHub has a "hide whitespace" option in the UI:
> 
> image

TIL, fancy.

PR #8872, opened 2023-02-28
Move initialization logic for sub-classes to `Class` constructor

jonkoops:
Moves the initialization logic for sub-classes extending `Class` into the constructor of `Class` itself. This allows classes that extend using the `extends` keyword (rather than `Class.extend()`) to perform initialization logic as well.

Implements parts of #8806.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -13,21 +13,7 @@ export class Class {
 	// [Extends the current class](#class-inheritance) given the properties to be included.
 	// Returns a Javascript function that is a class constructor (to be called with `new`).
 	static extend({statics, includes, ...props}) {
-		const NewClass = function (...args) {
-
-			Util.setOptions(this);
-
-			// call the constructor
-			if (this.initialize) {
-				this.initialize.apply(this, args);
-			}
-
-			// call all constructor hooks
-			this.callInitHooks();
-		};
-
-		// inherit parent's prototype
-		Object.setPrototypeOf(NewClass.prototype, this.prototype);
+		const NewClass = class extends this {};
 
 		// inherit parent's static properties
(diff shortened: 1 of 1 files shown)


src/core/Class.js
```diff
@@ -13,21 +13,7 @@ export class Class {
...
-				this.initialize.apply(this, args);
-			}
-
-			// call all constructor hooks
-			this.callInitHooks();
-		};
-
-		// inherit parent's prototype
-		Object.setPrototypeOf(NewClass.prototype, this.prototype);
+		const NewClass = class extends this {};
```
jonkoops, line 16: We are now also actually using the `extends` keyword to extend parent classes properly.

src/core/Class.js
```diff
@@ -13,21 +13,7 @@ export class Class {
...
-			if (this.initialize) {
-				this.initialize.apply(this, args);
-			}
-
-			// call all constructor hooks
-			this.callInitHooks();
-		};
-
-		// inherit parent's prototype
-		Object.setPrototypeOf(NewClass.prototype, this.prototype);
```
jonkoops, line 30: The `extends` keyword above removes the need to set the prototype here.

src/core/Class.js
```diff
@@ -90,6 +76,20 @@ export class Class {
 		return this;
 	}
 
+	_initHooksCalled = false;
```
jonkoops, line 79: Since this field is set later on in `callInitHooks()` I have defined it here in advance.

jonkoops:
If nobody has any objections I will go ahead and merge this in.

Falke-Design (approved):

jonkoops:
Thanks @Falke-Design

PR #8879, opened 2023-03-13
Guard for-in loops and enable guard-for-in lint rule.

### LINKED ISSUE #8843: Replace as many `for...in` loops as possible with `for...of`
Enable of the [`guard-for-in`](https://eslint.org/docs/latest/rules/guard-for-in) rule and fix as many cases as possible by replacing them with a `for...of` loop instead.

_Originally posted by @jonkoops in https://github.com/Leaflet/Leaflet/issues/8840#issuecomment-1426855332_

alope107:
Closes #8843 

- Adds checks to for-in loops to ensure they are not looping over inherited properties.
- Enables [lint rule](https://eslint.org/docs/latest/rules/guard-for-in) to ensure these guards are enabled in future code.
- Does NOT add a guard to `Util.extend` as this code purposely copies properties from source objects.

### DIFF
--- a/src/control/Control.Layers.js
+++ b/src/control/Control.Layers.js
@@ -85,11 +85,15 @@ export const Layers = Control.extend({
 		this._handlingClick = false;
 
 		for (const i in baseLayers) {
-			this._addLayer(baseLayers[i], i);
+			if (Object.hasOwn(baseLayers, i)) {
+				this._addLayer(baseLayers[i], i);
+			}
 		}
 
 		for (const i in overlays) {
-			this._addLayer(overlays[i], i, true);
+			if (Object.hasOwn(overlays, i)) {
+				this._addLayer(overlays[i], i, true);
+			}
 		}
 	},
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -40,6 +40,7 @@ module.exports = {
 		'spaced-comment': 'error',
 		'strict': 'off',
 		'wrap-iife': 'off',
+		'guard-for-in': 'error',
 		// TODO: Re-enable the rules below and fix the linting issues.
 		'no-invalid-this': 'off',
 		'prefer-object-has-own': 'error',
--- a/spec/suites/dom/DomEvent.DoubleTapSpec.js
+++ b/spec/suites/dom/DomEvent.DoubleTapSpec.js
@@ -83,10 +83,8 @@ describe('DomEvent.DoubleTapSpec.js', () => {
 			detail: 2,
 			target: container
 		});
-		for (const prop in expectedProps) {
-			if (Object.hasOwn(expectedProps, prop)) {
-				expect(event[prop]).to.be(expectedProps[prop]);
-			}
+		for (const [prop, expectedValue] of Object.entries(expectedProps)) {
+			expect(event[prop]).to.be(expectedValue);
 		}
 		expect(event.isTrusted).not.to.be.ok();
 	});
--- a/spec/suites/dom/DomEvent.DoubleTapSpec.js
+++ b/spec/suites/dom/DomEvent.DoubleTapSpec.js
@@ -84,7 +84,9 @@ describe('DomEvent.DoubleTapSpec.js', () => {
 			target: container
 		});
 		for (const prop in expectedProps) {
-			expect(event[prop]).to.be(expectedProps[prop]);
+			if (Object.hasOwn(event, prop)) {
+				expect(event[prop]).to.be(expectedProps[prop]);
+			}
 		}
 		expect(event.isTrusted).not.to.be.ok();
 	});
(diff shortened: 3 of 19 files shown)


mourner:
Thanks for the PR, this looks great!

> * Does NOT add a guard to `Util.extend` as this code purposely copies properties from source objects.

Just wondering, are there many instances of us relying on this behavior? This feels a bit awkward to do it now... Perhaps if there are just a few places like this, we could switch there to prototype inheritance via `Object.create`?

alope107:
Thanks for taking a look! A few thoughts on the `Util.extend`:

- The main place where it seems like the inherited property copying is needed in this `extend` is when working with `LatLng`. I think this may be related to how `LatLng` does not inherit from Leaflet's `Class` object, but I'm not 100% sure. There may be some opportunity to easily move away from using `Util.extend` when dealing with `LatLng`, but it's something that would require a closer look. I'm happy to look into it, but due to my inexperience with the codebase I might need a bit of support.
- Even if we do move away from relying on `Util.extend` to copy inherited properties internally, I would be wary of changing the way the function works. As I understand it, it gets exported as `L.extend` and may be used extensively by other code that depends on Leaflet. Changing it to no longer copy inherited properties would likely break other projects in confusing ways.

Because dealing with these issues looks non-trivial, I would vote to not include them in this PR and instead open a separate issue to investigate them after this code is merged. What do you think @mourner ?

src/control/Control.Layers.js
```diff
@@ -85,11 +85,15 @@ export const Layers = Control.extend({
 		this._handlingClick = false;
 
 		for (const i in baseLayers) {
```
jonkoops, line 87: Perhaps we can improve this by using `Object.entries()` instead? For example:

```js
for (const [name, layer] of Object.entries(baseLayers)) {
    this._addLayer(layer, name);
}
```

src/control/Control.Layers.js
```diff
@@ -85,11 +85,15 @@ export const Layers = Control.extend({
 		this._handlingClick = false;
 
 		for (const i in baseLayers) {
-			this._addLayer(baseLayers[i], i);
+			if (Object.hasOwn(baseLayers, i)) {
+				this._addLayer(baseLayers[i], i);
+			}
 		}
 
 		for (const i in overlays) {
```
jonkoops, line 93: Seme here, I think we can reduce the complexity of this code by using `Object.entries()`.

jonkoops (changes requested):
I think in a lot of cases here we can avoid calling `Object.hasOwn()` by looping over the result of `Object.entries()` instead (which should already check this for us).

Could you change this where it makes sense? Thanks!

alope107:
> I think in a lot of cases here we can avoid calling Object.hasOwn() by looping over the result of Object.entries() instead (which should already check this for us).

I like the idea! The challenge here is that we can only use `Object.entries` if the object is not null or undefined. We could add additional null checks before each of the loops, but I think then things start to get more cluttered than just using the `hasOwn` guard.

I went through and identified some cases where I could confidently say that the object being iterated over is not null or undefined, and in those cases I switched to using `Object.entries` like you suggested. Let me know if you see other spots where we can be sure the object is non-null. Thanks!

src/control/Control.Layers.js
alope107: We could make this switch if we were certain that `baseLayers` was not null or undefined. Is that guaranteed in this case? If not, we would have to add an additional null check, or stay with the `hasOwn` guard.

src/control/Control.Layers.js
alope107: Similarly, the viability of this change hinges on whether baseLayers in guaranteed to be not null/undefined.

alope107:
Hi @jonkoops, is there still interest in this PR? Happy to either update this one or to take a stab at a version using only `Object.entries` and a bunch of null checks if that's preferred. Thanks!

jonkoops (approved):

jonkoops:
Sorry for the recent negligence , I have been rather busy at work so I have had no time to get back to this PR.

We can always improve the code at a later point with `Object.entries()` where it makes sense. I think this is already a great improvement so I am going ahead and merge in this work.



Thanks for your contribution @alope107!

alope107:
No worries @jonkoops, the review is much appreciated! If there are any other issues or tasks you'd recommend for someone starting to get their feet wet with this repo, feel free to let me know. It was fun starting to get to know the codebase!

jonkoops:
Sure @alope107, if you are looking for new issues to pick up we do try to label those we think are good contributions with the ["help wanted"](https://github.com/Leaflet/Leaflet/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) label. Feel free to take a look at any of those.

PR #8888, opened 2023-03-16
Fix click events when latlng isn't initially set

Pierstoval:
My use-case is this: I'm using latest Leaflet (1.9.3) alongside with [Leaflet.VectorGrid](https://github.com/Leaflet/Leaflet.VectorGrid/) (which is a bit outdated but anyway).

When I set up a VectorGrid layer, set `interactive: true` to its options, and add `myVectorGridLayer.on('click', ...)` event handling (also works with mousedown, per se), I have an error that says "latlng is not defined". This is probably caused by VectorGrid not using latlng by default, so I patched VectorGrid, but in the end I realized that "mouseover" also causes this "latlng is not defined" error right in the Canvas renderer.

So here is this patch.

I don't know Leaflet internals enough to see if this fix is the right one or if I should fix something upstream, so feel free to reject, but please send explanations if you do, so that we can learn more about internals 🙏

### DIFF
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -399,11 +399,17 @@ export const Canvas = Renderer.extend({
 
 			if (candidateHoveredLayer) {
 				this._container.classList.add('leaflet-interactive'); // change cursor
+				if (!candidateHoveredLayer.getLatLng() && point) {
+					candidateHoveredLayer.setLatLng(this._map.unproject(point));
+				}
 				this._fireEvent([candidateHoveredLayer], e, 'mouseover');
 				this._hoveredLayer = candidateHoveredLayer;
 			}
 		}
 
+		if (this._hoveredLayer && !this._hoveredLayer.getLatLng() && point) {
+			this._hoveredLayer.setLatLng(this._map.unproject(point));
+		}
 		this._fireEvent(this._hoveredLayer ? [this._hoveredLayer] : false, e);
 
 		this._mouseHoverThrottled = true;


IvanSanchez:
Ok, so I see there's https://github.com/Leaflet/Leaflet.VectorGrid/pull/281 related to this.

But... what's the problem this is trying to solve? Under what condition does a pointer event not have `latlng`?

Pierstoval:
> But... what's the problem this is trying to solve? Under what condition does a pointer event not have `latlng`?

I have no idea, I guess it's caused by VectorGrid, but could not find the origin. Spent a whole week on this without finding out. But as said, I lack knowledge about internals of both libs, and that's my "current workaround" 😕

jonkoops:
Looking at the code changes I don't understand in which case this would be occurring in Leaflet. Seems like the issue is stemming from the plugin itself?

Pierstoval:
It's really hard for me to make a reproducer as it's part of an internal (quite big) library I'm working on for my client, and also hard since we're using GeoServer as data source, Svelte as framework, and Typescript as JS flavour, that's a lot to extract from private code 😭 

However, I could show it in a private not-recorded video call, that would be fine 👌

jonkoops:
I'm sorry, but we are quite swamped at the moment so video calling is not an option (at least for me). If there is no minimal reproducible example for this issue I don't feel like we can confidently merge this in.

I am going to close this PR for that reason. There is just to much uncertainty around this implementation.

If you do manage to reproduce this issue purely with Leaflet itself, feel free to report an issue with a reproducible example.

PR #8890, opened 2023-03-22
Fix tooltip focus listener if getElement is no function

Falke-Design:
If a LayerGroup contains other layers without the `getElement()` function (for example another layergroup), it will throw an error, because it loops over all layers in the LayerGroup:

https://github.com/Leaflet/Leaflet/blob/9b9c71208ee689557c2aa1170d0f0915d41b12eb/src/layer/Tooltip.js#L388-L395

Another solution (and I think the cleaner one) would be to loop over all layers and calling `_addFocusListeners` instead of `_addFocusListenersOnLayer`. But I'm a little bit scared about endless recursion.

```
 _addFocusListeners() { 
 	if (this.getElement) { 
 		this._addFocusListenersOnLayer(this); 
 	} else if (this.eachLayer) { 
 		this.eachLayer(this._addFocusListeners, this);  // <-- into _addFocusListeners instead of _addFocusListenersOnLayer
 	} 
 }, 
```

### This fix needs to be cherry-picked into `v1`


Falke-Design:
@jonkoops @mourner @IvanSanchez please review, I would like to release a new `1.9` version soon.

jonkoops (approved):

PR #8891, opened 2023-03-27
Alleviate tile gaps in chromium by using mix-blend-mode CSS

IvanSanchez:
This alleviates the dreaded tile gaps of #3575, in a clean way, with no downsides that I'm aware of.

The tile gaps are still visible if the page zoom is not 100%, however (see https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-1485077575).

Credit goes to @lapo-luchini (see https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-1484816250).

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -52,7 +52,9 @@
 	max-height: none !important;
 	width: auto;
 	padding: 0;
+	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
 	mix-blend-mode: plus-lighter;
+
 	}
 
 .leaflet-container.leaflet-touch-zoom {


Falke-Design (approved):
We should cherry pick to `v1` this and ship it with `1.9.4` too

dist/leaflet.css
```diff
@@ -52,6 +52,7 @@
 	max-height: none !important;
 	width: auto;
 	padding: 0;
+	mix-blend-mode: plus-lighter;
```
jonkoops, line 55: Perhaps we can add a comment for future reference?
```suggestion
	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
	mix-blend-mode: plus-lighter;

```

joeberkovitz:
I cherry-picked this CSS change to apply for a project that uses Leaflet 1.8 and it created problems with the colors of overlying markers. I don't know if anyone's tested this in combination with markers based on SVG file-based icons but it might be a good idea to verify that it doesn't break them.

IvanSanchez:
@joeberkovitz Could you please [open a bug report](https://github.com/Leaflet/Leaflet/issues/new/choose) with that info? I haven't done throughout testing, but if you can provide a minimal test case where colours of overlaying markers are affected, that'd be very useful.

mhirsch:
This does not seem to work in Firefox, and also has the side effect of making the "overlap" [fix](https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-150544739) from #3575 fail, since the overlapped border is lightened.

jonkoops:
@mhirsch please report this as a new issue with some steps to reproduce it.

and-rej:
Please note **this CSS property has a performance cost** associated with it.

Our client uses virtualised desktops that don't have access to a GPU and we lost 10-20 FPS when we upgraded to Leaflet 1.9.4. Took me a while to trace it back to this, but fortunately it's easy to override the style and turn this off. We've never noticed the issue it's supposed to address.

You should be able to see it for yourself. Just turn off hardware acceleration in your browser settings, show the dev tools FPS meter, and pan a map around. Toggle the CSS property between `plus-lighter` and `normal` and you should see a difference in FPS.

dist/leaflet.css
Jayhelp: My location doesn't show actually. How can I fix and update

Nekzuris:
Some downsides of this approach can be mitigated with exceptions in CSS #9560.

I can confirm a drop of around 50% FPS (120→60) without GPU acceleration. To my knowledge there is no reliable way to detect if hardware acceleration is enabled. Maybe you should add a setting to control how tile gaps are handled (plus-lighter, overlap or nothing), but ideally the control should be left to the user.

PR #8907, opened 2023-04-08
Apply mix-blend-mode only on tile images

### LINKED ISSUE #8905: Blend mode on leaflet tiles causes visual artifacts on overlap
### Checklist

### Steps to reproduce

Add a custom `GridLayer` to a Leaflet map in which the tiles of the grid:
- are permitted to exceed tile boundaries by use of CSS `overflow: visible`
- have features whose visual representations overlap features in adjacent tiles

This is a valid use case for `GridLayer` instances that are displaying an underlying vector dataset of points, rather than a set of image tiles. In this case, each point in the dataset belongs to a single tile, but will be visually displayed as an icon of some kind whose extent will go outside the tile's boundary if the point is just inside the boundary. Thus clipping must be disabled in order for icons not to be truncated at tile borders.

### Expected behavior

Overlapping markers in adjacent custom GridLayer tiles should appear with expected colors:

image

### Current behavior

Overlapping markers appear with improperly blended colors:

image

https://gist.github.com/joeberkovitz/6132abe9df80a8f114372ac3f44a1ab3

### Environment

- Leaflet version: main branch as of https://github.com/Leaflet/Leaflet/commit/0392b7f44481119169e3db4babe2eab39cf05b59 which appears to have introduced the problem
- Browser (with version): Chrome 112.0.5615.20, Safari 16.1 , Firefox 110.0.1
- OS/Platform (with version): MacOS Ventura, iOS 15

Falke-Design:
Fix #8905, `mix-blend-mode` issue with custom svg-tiles. The bug was introduced in #8891.

Needs to be cherry-picked into `v1`

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -52,10 +52,11 @@
 	max-height: none !important;
 	width: auto;
 	padding: 0;
+	}
+.leaflet-container img.leaflet-tile {
 	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
 	mix-blend-mode: plus-lighter;
-
-	}
+}
 
 .leaflet-container.leaflet-touch-zoom {
 	touch-action: pan-x pan-y;
@@ -588,7 +589,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 
 /* Printing */
-	
+
 @media print {
 	/* Prevent printers from removing background-images of controls. */
 	.leaflet-control {
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -56,7 +56,7 @@
 .leaflet-container img.leaflet-tile {
 	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
 	mix-blend-mode: plus-lighter;
-}
+	}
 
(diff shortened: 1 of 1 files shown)


dist/leaflet.css
```diff
@@ -52,10 +52,11 @@
 	max-height: none !important;
 	width: auto;
 	padding: 0;
+	}
+.leaflet-container img.leaflet-tile {
 	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
 	mix-blend-mode: plus-lighter;
-
-	}
+}
```
IvanSanchez, line 59: Nitpicking: This closing brace should be indented, for consistency with the rest.

IvanSanchez (changes requested):

IvanSanchez:
Other than my nitpicking on the CSS formatting, looks good to me. I'd like @joeberkovitz to have a look and see if #8905 is fixed satisfactorily.

IvanSanchez (approved):

joeberkovitz:
I have retested this and #8905 appears to be fixed by this PR, while still addressing the original concern of gaps between tiles.

jonkoops:
Nice, thanks for fixing this @Falke-Design

PR #8909, opened 2023-04-08
feat: add aria-label when collapsed is false

JessieFrance:
First, thank you to everyone that maintains Leaflet! This PR aims to adopt the "Proposed solution" in #7529 to provide missing contextual information. More specifically, when the `collapsed` option is set to `false`, this PR adds an `aria-label` with a value of `Layers` to the `leaflet-control-layers-list` element. Please let me know if this looks okay, and if not, I'm happy to make changes.


Falke-Design:
PR solve the issue, thank you 👍 

But I'm not sure if the aria-label should also be added when it is expanded after being collapsed. Because `<a class="leaflet-control-layers-toggle">` will be set to `display: none` too.
aria-label-control-layers

@Malvoz can you please help

JessieFrance:
Thank you for looking at this! That is an excellent point about the case where it is expanded after being collapsed. Please let me know if you want me to add this part.

sandeepvinnu (approved):

src/control/Control.Layers.js
```diff
@@ -417,6 +417,7 @@ export const Layers = Control.extend({
 	_expandIfNotCollapsed() {
 		if (this._map && !this.options.collapsed) {
 			this.expand();
+			document.querySelector('.leaflet-control-layers-list').setAttribute('aria-label', 'Layers');
```
jonkoops, line 420: I feel like there must be a better way of doing this than querying the whole document. Is this element not stored somewhere as a field on this class already?

Falke-Design: That would be `this._section`:

https://github.com/Leaflet/Leaflet/blob/007025069d7afffc03c3ebd74dbf8bdc9ebe7cce/src/control/Control.Layers.js#L189

JessieFrance: Thank you @jonkoops and @Falke-Design! I added another commit to this so that it uses the `_section` class variable instead of querying the whole document. Also @Falke-Design I added another commit for your earlier comment on the case where it is expanded after being collapsed. Please let me know if this all looks okay, and if not, I'm happy to make more changes.

JessieFrance:
Thank you @jonkoops and @Falke-Design! My apologies for not being able to fix this until now - I've added some more commits that (1) resolve the merge conflict, and (2) add some tests for the ```aria-label``` to make sure it is working as expected. Please let me know if this looks okay or if I should make changes, and many thanks again.

nfreear:
Hi,

Please take a look at the comment on [#7529](/Leaflet/Leaflet/issues/7529#issuecomment-1704346383).

A better solution for expressing expand/collapse state would be to use `aria-expanded` on the triggering element, as opposed to adding/modifying `aria-label`.

I hope that helps.

Nick

PR #8911, opened 2023-04-10
Bugfix: Fix noMoveStart option for fitBounds method

AbdullahSohail-SE:
The current implementation overrides the noMoveStart option to false value.


veitbjarsch:
Would be awesome if we could bring this adjustment into the next release.

Falke-Design:
We need to check if the docs needs to be updated. https://leafletjs.com/reference.html#zoom-options

veitbjarsch:
The docs describe the noMoveStart as inherit from panning. But it didn't worked before.

AbdullahSohail-SE:
@veitbjarsch seconded, cc: @Falke-Design

AbdullahSohail-SE:
@Falke-Design any updates?

veitbjarsch:
@mourner @jonkoops @IvanSanchez any chance to get this into the next release? Would love to use it.

src/map/Map.js
```diff
@@ -1683,7 +1683,8 @@ export const Map = Evented.extend({
 
 		Util.requestAnimFrame(function () {
 			this
-			    ._moveStart(true, false)
+			    ._moveStart(true, options.noMoveStart ?? false)
+
```
Falke-Design, line 1687:
```suggestion
			    ._moveStart(true, options.noMoveStart ?? false)
```

Falke-Design (changes requested):

Falke-Design (approved):

PR #8917, opened 2023-04-12
Set `outlineStyle` instead of `outline` when preventing outline

### LINKED ISSUE #8908: Safari 16.4 preventOutline / tests not working anymore
The new Safari version 16.4, released on 27.03, breaks the `preventOutline` test. `currentcolor` is returned instead of `none` :

grafik

I tried to test it with BrowserStack but they have only version 16.3. 
Rlease notes: https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes

In [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/CSS/outline) I found: 
grafik

We should also test if something has changed for the user.

jonkoops:
Changes the implementation of `preventOutline()` and `restoreOutline()` so that [`outline-style`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-style) is set rather than the whole [`outline`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline) property. This seems to prevent a regression introduced in Safari 16.4, and remains functional in other browsers.

Closes #8908

### DIFF
--- a/spec/suites/dom/DomUtilSpec.js
+++ b/spec/suites/dom/DomUtilSpec.js
@@ -288,22 +288,23 @@ describe('DomUtil', () => {
 	describe('#preventOutline, #restoreOutline', () => {
 		it('prevent / restore outline for the element', () => {
 			const child = document.createElement('div');
-			el.appendChild(child);
+			const originalStyle = child.style.outlineStyle = 'dotted';
+
 			child.tabIndex = 0;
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):
Looks good for me. Did you test if the outline is still removed correctly in chrome / firefox?

jonkoops:
Yes, I tested this manually in the latest versions of Chrome, Firefox and Safari. They all seem to have the expected behavior.

PR #8921, opened 2023-04-18
Convert test debug pages to ESM

jonkoops:
Converts the test debug pages (located in `debug/tests`) to ESM and cleans up the code in the process. This is part of a greater effort to land ESM as a first-class citizen. Code that was broken has been replaced with working code or removed.


IvanSanchez:
How are you planning on loading the module on browsers?

I've been starting to use `importmap`s recently, and they feel fantastic. Do have a look at https://gitlab.com/IvanSanchez/leaflet.gleo/-/blob/main/demo/sprites.html#L7

jonkoops:
Funny you mention Import Maps @IvanSanchez, because that is exactly how I was planning on distributing Leaflet 2.0! For folks that do not want to use a package manager or bundlers I think it's the best way we can ship Leaflet.

Reason why I haven't been using them yet is because up until the latest version of Safari (just released) it was not possible to use them there. I have every intention of converting all these examples to Import Maps before we ship 2.0.

Falke-Design (approved):
Looks good for me.

I think one of the next steps will be to update and check how a plugin should be created. Because then plugin-maintainers can start updating to v2

jonkoops:
@Falke-Design Yes, I'd like to start thinking about documentation. I think we need to put in a little work to ensure we're not deploying it from `main` anymore. Or perhaps we need a `v2.leaflet.com` for the new docs.

PR #8924, opened 2023-04-21
Convert vector debug pages to ESM

jonkoops:
Converts the vector debug pages (located in `debug/vector`) to ESM and cleans up the code in the process. This is part of a greater effort to land ESM as a first-class citizen. Code that was broken has been replaced with working code or removed.


debug/vector/blanket.html
```diff
@@ -1,70 +1,62 @@
...
+			});
+
+			new DebugBlanket({
+				padding: -0.1,
+				continuous: true
+			}).addTo(map);
+
+			const canvas = new Canvas({
+				padding:-0.1,
+				// continuous: true
```
Falke-Design, line 49: `continuous` is not a valid option and should be removed https://leafletjs.com/reference.html#canvas

debug/vector/blanket.html
```diff
@@ -1,70 +1,62 @@
...
+			}).addTo(map);
+			canvas._container.style.border = '2px solid red';
+			new CircleMarker([40.5, -3.6], {radius: 20, color: 'red', renderer: canvas}).addTo(map);
+
+			const svg = new SVG({
+				padding:-0.1,
+				// continuous: true,
```
Falke-Design, line 56: `continuous` is not a valid option and should be removed https://leafletjs.com/reference.html#canvas

Falke-Design (approved):
Besides the two points, which has nothing to do with this changes, everything looks good

jonkoops:
Went ahead and fixed the review comments as well, I'll merge this in :+1: Thanks again @Falke-Design!

PR #8926, opened 2023-04-23
Use import maps to load Leaflet in the `debug` directory

jonkoops:
Adds Leaflet to an [import map](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap) for each of the examples in the `debug` directory. This allows Leaflet to be imported using `import {} from 'leaflet'`.

### DIFF
--- a/debug/map/canvas.html
+++ b/debug/map/canvas.html
@@ -6,11 +6,18 @@
 		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
 		<link rel="stylesheet" href="../../dist/leaflet.css" />
 		<link rel="stylesheet" href="../css/screen.css" />
+		<script type="importmap">
+			{
+				"imports": {
+					"leaflet": "../../dist/leaflet-src.esm.js"
+				}
+			}
+		</script>
 	</head>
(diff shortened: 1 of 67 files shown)


IvanSanchez (approved):
LGTM

mourner:
@jonkoops maybe worth condensing the import map a little so that the pages don't look as verbose? E.g.:

```html
<script type="importmap">{"imports": {"leaflet": "../../dist/leaflet-src.esm.js"}}</script>
```

jonkoops:
Yeah, I was also of ways to reduce the boilerplate for the debug pages. I thin it's probably fine for now since folks will copy and paste the contents of these files. But perhaps a better solution would be to have some kind of re-usable template for the outer HTML of these pages.

IvanSanchez:
Personally, I'd rather have the importmaps be more readable, at least by now. Users should be able to see how the importmap is looking like instead of copy-pasting the magic one-liner. It's worth the 20 or so extra bytes.

jonkoops:
More exiting ESM and import map stuff on the way, working on the tests now :eyes:

Issue #8477, opened 2022-09-26
Evergreen language in the documentation


Malvoz:
- [ ] Change the wording to not mention specific Leaflet versions (and to not say "_new_ feature") in:
  - [map-panes](https://github.com/Leaflet/Leaflet/blob/399ed2105220f10b79e69b57420597e03502e611/docs/examples/map-panes/index.md?plain=1#L20)
  - [wms](https://github.com/Leaflet/Leaflet/blob/990d49773a129619f2eed09712867627c8e3cf60/docs/examples/wms/wms.md?plain=1#L146-L150)
- [ ] Change [the wording in the Video overlay's section](https://github.com/Leaflet/Leaflet/blob/c2324e52e273d6f7434102a5e49b6db2142e1b62/docs/examples/overlays/index.md?plain=1#L75-L93) to assume HTML `<video>` is supported.
- [ ] Use the terms "HTML" and "CSS" instead of "[HTML5](https://github.com/Leaflet/Leaflet/search?q=html5)" and "[CSS3](https://github.com/Leaflet/Leaflet/search?q=css3)"
- [ ] Is the slogan "A library for [_mobile-friendly_](https://github.com/Leaflet/Leaflet/search?q=an+open-source+JavaScript+library+for+mobile-friendly+interactive+maps) interactive maps" dated?

Falke-Design:
> Remove obsolete way of instantiating a circle?

Only in docs or in the code too? I would like to remove it, but many users are still using `L.circle([0,0], 400).addTo(map)`

jonkoops:
Great initiative! We'll have to go over a large part of the docs anyway for 2.0 when we drop support for the globals and other non-standard module systems.

hollowM1ke:
Hello. I'm looking for a good-first issue to contribute to. Can I work on this one? @Malvoz @Falke-Design @jonkoops

Falke-Design:
@hollowM1ke most of the points needs to be further discussed but you can try to change the following to points and make your first PR. Please make sure to read through the [CONTRIBUTING.md](https://github.com/Leaflet/Leaflet/blob/main/CONTRIBUTING.md#improving-documentation) and test the new docs

Change the wording to not mention specific Leaflet versions (and to not say "new feature") in:
    [map-panes](https://github.com/Leaflet/Leaflet/blob/399ed2105220f10b79e69b57420597e03502e611/docs/examples/map-panes/index.md?plain=1#L20)
[...]

PR #8934, opened 2023-04-27
Update docs language for HTML and CSS.

alope107:
- Use HTML and CSS terminology instead of HTML5 and CSS3.

- Part of a larger cleanup to use evergreen language. Addresses first task of #8477.


jonkoops (approved):
Yeah, I agree referring to HTML and CSS by version is a bit antiquited. Thanks!

PR #8935, opened 2023-04-27
Remove Rollup pre-proccesor from Karma runner

jonkoops:
This PR refactors the Karma configuration so that the Rollup pre-proccesor is no longer included. This is needed for a subsequent that is in the works that converts the tests to load Leaflet using import maps (and fully ESM).

This PR also:
- Removes the `spec/index.html` file in favor of using Karma's server.
- Runs Karma in watch mode by default so tests are ran when changes are made to the tests or source.
- Adds a type annotation so that IDEs can provide documentation and auto completion of the Karma config.
- Aligns the values of the Karma config as much as possible with the defaults.


mourner (approved):
Very cool to see so much code removed! 🤩

CONTRIBUTING.md
```diff
@@ -110,29 +110,32 @@ development. `dist/leaflet.js` is uglified and thus is smaller, so it's better
...
+the webpages in the `debug/` directory, and run the tests
 in a graphical browser. This requires regenerating the bundled files quickly.
 
 In order to do so, run `npm run watch` or `yarn run watch`. This will keep
 on rebuilding the bundles whenever any source file changes.
 
 ## Running the Tests
 
-To run the tests from the command line, install [Google Chrome](https://www.google.com/chrome/) then run:
+Before running the tests make sure that the source code has been built (as mentioned above). If you want to run the tests in the background while working on Leaflet it is recommended you run the build in `watch` mode. This way the tests will automatically re-run when changes to the source code are made.
```
IvanSanchez, line 121: TODO: run the tests based solely on the unbundled ES modules

jonkoops: I believe we can get there yeah, not sure if Karma can track the changes between modules. We might need to use a different solution in that case.

jonkoops: Created an issue (#8937).

Falke-Design:
> Removes the spec/index.html file in favor of using Karma's server.

I just saw it and I really don't like it. For me it was a great way to quickly create and test my tests.
@jonkoops If there is a way to bring this (in a modern way) back, I would really appreciate it.

jonkoops:
@Falke-Design I expect this to change when we move from Karma to Web Test Runner. The debugging experience there should be a lot more integrated and intuitive.

PR #8936, opened 2023-04-27
Import Leaflet in tests using JavaScript modules

jonkoops:
Changes the unit tests so they use the named exports from Leaflet loaded trough an import map.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,8 @@ module.exports = {
 		// TODO: Re-enable the rules below and fix the linting issues.
 		'no-invalid-this': 'off',
 		'prefer-object-has-own': 'error',
-		'prefer-spread': 'off'
+		'prefer-spread': 'off',
+		'no-new': 'off'
 	},
 	overrides: [
 		{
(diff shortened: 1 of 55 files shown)


jonkoops:
Interesting, tests sometimes fail. I have no idea why yet, will look at it another time (gotta sleep).

mourner:
> Interesting, tests sometimes fail. I have no idea why yet, will look at it another time (gotta sleep).

Looks like all the failures are timeouts — I guess browsers take to long to hop through all the imports and fetch everything.

Also noting that the total test time jumps from <3m to 6–8m. We should look into addressing that before shipping — perhaps this alone would be enough to keep running against the bundle. It may also be beneficial because some issues may only manifest after bundling, which is how users end up using the code anyway, so tests closer to real world.

jonkoops:
> Perhaps this alone would be enough to keep running against the bundle.

How do you mean? We are currently still running the tests against the bundled code. Or are you referring to #8937?

I am not quite sure what is causing the performance to regress so much, I'll have to take a look and investigate deeper.

mourner:
@jonkoops ah right, I missed that it's not yet implementing individual file import. But the regression is indeed suspicious — let's make sure it's not slower before we can land this.

jonkoops:
I'm closing this PR so that I can merge this work into a larger effort to start using [Web Test Runner](https://modern-web.dev/docs/test-runner/overview/). See #8939 for more information.

Issue #8939, opened 2023-04-28
Karma is deprecated


jonkoops:
### Checklist

- [X] I've searched through the [plugins](https://leafletjs.com/plugins.html) to make sure this feature isn't already available, or think it shouldn't require a plugin.
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this feature hasn't been requested already.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Motivation

Karma [has been deprecated](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes) today, so we'll likely have to migrate away from it at some point.

### Suggested solution

We might want to consider [Web Test Runner](https://modern-web.dev/docs/test-runner/overview/) as the [blog post](https://blog.angular.io/moving-angular-cli-to-jest-and-web-test-runner-ef85ef69ceca) from the Angular team suggests.

mourner:
I thought we would migrate away from Karma sooner or later anyway — it's a very old project and a pretty stale one. Web Test Runner looks pretty nice.

jonkoops:
@mourner I wonder if the performance issues that we're experiencing under #8936 might also be resolved migrating away from Karma. I think I will take a look at how much effort it might be to move away from Karma first.

jonkoops:
@Falke-Design as part of our migration to Web Test Runner we need all of our dependencies used during the testing process to be ESM compatible. Since you are maintaining `ui-event-simulator` could you review [the PR](https://github.com/Falke-Design/ui-event-simulator/pull/2) implementing the aforementioned there?

Falke-Design:
@jonkoops done

jonkoops:
@Falke-Design  Awesome thanks! Could you also publish a 2.0 for that package?

Falke-Design:
Should be already there 🤔

jonkoops:
Strange, it is not showing up [on NPM](https://www.npmjs.com/package/ui-event-simulator?activeTab=versions).

Falke-Design:
I forgot to publish ... 😄 I will do it at the evening (in 6h)

jonkoops:
[...]

PR #8952, opened 2023-05-12
Replace expect.js with Chai

jonkoops:
Replaces [expect.js](https://github.com/Automattic/expect.js/) with [Chai](https://www.chaijs.com/) for assertions in our tests. Expect is unmaintained (last commit 9 years ago and no issue activity), and does not have an ESM variant.

We'll likely be moving from Karma to Web Test Runner, which comes with [Chai by default](https://modern-web.dev/docs/test-runner/writing-tests/helper-libraries/#assertions). This moves to closing #8939.


mourner:
Nice! A little puzzled by all the test failures though, what's up with that?

jonkoops:
Yeah, gotta figure that out. Looks to be something specific to Firefox and Safari as Chrome works fine.

.github/workflows/main.yml
```diff
@@ -124,6 +124,14 @@ jobs:
...
           node-version: ${{ env.NODE_VERSION }}
 
+      - name: Set up Google Chrome
+        uses: browser-actions/setup-chrome@v1
+        if: contains(matrix.browser, 'Chrome')
+
+      - name: Set up Firefox
+        uses: browser-actions/setup-firefox@v1
+        if: contains(matrix.browser, 'Firefox')
+
```
jonkoops, line 134: Added some logic here to ensure we always get the latest version of Chrome and Firefox. Seems like we can get different versions of these browsers depending on what runner GitHub actions graces us with otherwise, which leads to inconsistent results.

jonkoops:
Looks like I missed a couple of tests specific to touch interaction that I forgot to run locally. Should be all good now.

mourner (approved):
👍

PR #8957, opened 2023-05-16
Switch to active fork of Leaflet.markercluster

KristjanESPERANTO:


jonkoops:
Since this is an extension on the Leaflet repository being replaced I don't think we want to merge this. While I acknowledge that currently our maintenance of plugins under our banner is not where it needs to be I think this is jumping the gun.

Closing this PR.

KristjanESPERANTO:
Is there a plan to improve the situation? People have to deal with issues in the "official plugins" that have been fixed in some forks for a long time.

jonkoops:
Submit a PR to fix those issues, we'll get to it when we can. Currently our main priorities are to work on Leaflet 2.0, the plugins will come later.

KristjanESPERANTO:
Okay, this prioritization is understandable. Thank you for your work! 🙂

Issue #8960, opened 2023-05-18
Layer-Control tutorial: Mapbox tiles not loading


Falke-Design:
Let us replace the Mapbox Tiles in the [Layer-Control tutorial](https://leafletjs.com/examples/layers-control/) with a free [provider](https://leaflet-extras.github.io/leaflet-providers/preview/).

![grafik](https://github.com/Leaflet/Leaflet/assets/19800037/9132859e-3e2d-4122-9237-f05f4f8d023b)

Dele-Oyelese:
Hi @Falke-Design ,

I would be happy to help out with issue. 

While reviewing the free providers it appears the [Esri.WorldImagery](https://leaflet-extras.github.io/leaflet-providers/preview/#filter=Esri.WorldImagery) maybe suitable to replace for the Satellite imaging in Layer-Control tutorial. I was unable to find a suitable Streets view in the documentation you provide or anywhere online. Let me know if you know of any Street view that could be used. 

In the mean time I will trying to familiarize myself with the code base and see how to implement the satellite view as this would be my first contribution.

Falke-Design:
@Dele-Oyelese I think the best would be to use the tiles from [OpenTopoMap](https://leaflet-extras.github.io/leaflet-providers/preview/#filter=OpenTopoMap) and [OpenStreetMap.HOT](https://leaflet-extras.github.io/leaflet-providers/preview/#filter=OpenStreetMap.HOT) instead of `Satellite` and `Streets`.

Dele-Oyelese:
Hi @Falke-Design I have opened up a pull request with the changes [requested] (https://github.com/Leaflet/Leaflet/pull/8968#issue-1722768186). Please review it and let me know if anything need to be changed at your earliest connivence no rush.  Also thank you for adding this issue, as a new contributor it was very helpful and I learned a lot about leaflet while doing it.

Dele-Oyelese:
Hi @Falke-Design I am sending a gentle nudge about this PR to see if there was any feedback regarding it or if it can be approved.

PR #8968, opened 2023-05-23
Mapbox tiles not loading 8960

Dele-Oyelese:
Fix issue #8960.

Pull request to update the [Layer Control Tutorial Page](https://leafletjs.com/examples/layers-control/). 
Updated the Satellite and Street view of the examples with titles suggested by @Falke-Design .
Changed some verbiage on the index about using a gray scale map tile as it was not being used in the example.

There was additional issue of grayscaling the mapbox legend as the lint rules do not allow double quotations on the styling so it was left out of the example.md file.
Screenshot 2023-05-23 at 1 27 54 PM


Falke-Design (approved):
Thank you. Looks good. I only changed the variable names a little bit

PR #8975, opened 2023-05-31
Import Leaflet in tests using JavaScript modules

jonkoops:
Follow up on #8936. I think we should land these changes, even if they potentially slow down test runs on CI. I've tested this locally, and I see little to no performance changes during development.

Doing a single test run, and measuring it's duration with the `time npm test` command yields the following:

**Karma (on `main`)**
```log
real	0m11.794s
user	0m0.826s
sys	0m0.229s
```

**Karma (this PR)**
```log
real	0m11.797s
user	0m0.812s
sys	0m0.251s
```

I also did some work to get Web Test Runner integrated (not part of this PR), and I have most of the test suite passing now (still working on fixing up `prosthetic-hand`). Just from some of my testing locally, I can see that it runs the tests much faster than on Karma:

**Web Test Runner**
```log
real 0m2.827s
user 0m2.233s
sys 0m0.276s
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,8 @@ module.exports = {
 		// TODO: Re-enable the rules below and fix the linting issues.
 		'no-invalid-this': 'off',
 		'prefer-object-has-own': 'error',
-		'prefer-spread': 'off'
+		'prefer-spread': 'off',
+		'no-new': 'off'
 	},
 	overrides: [
 		{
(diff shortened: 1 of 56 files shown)


mourner (approved):
Doesn't look any slower this time, let's ship! 👍 Nice work!

Minor nit: it would be nice to see what changed since you created the PR that fixed the timeouts — rebasing makes it hard to see how the work progressed, so I prefer just adding commits even if some of them fail or WIP.

PR #9005, opened 2023-06-29
Bugfix: refocus map after using layers control (#9004)

### LINKED ISSUE #9004: Map loses focus after clicking on map controls - reprise
### Checklist

### Steps to reproduce

1. click on an item in the layers control
2. try using +/- keyboard keys to zoom

### Expected behavior

The map should be refocused after using layers control.

### Current behavior

The map loses focus after using layers control.

In 2013, #2148 / 254e6fe6b9a6350cf2ee50ca8f6465bc433ac96e added code so that if you use button controls the map gets refocused, so that keyboard keys like +/- still work.

However, later in 2015,  #3275 / 8842385c376a6ba14573f85b38e7b6fe8e05e7ba broke this.  It intended to make `_refocusOnMap` only apply when it was triggered by a keyboard event, but the commit had the effect of making it _never_ apply, because the commit added a parameter to the function, without passing it at call time.

https://leafletjs.com/examples/layers-control/

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Chrome 114.0.5735.133, Firefox 113.0
- OS/Platform (with version): MacOS 13.2.1

quarl:
Fixes #9004

### DIFF
--- a/src/control/Control.js
+++ b/src/control/Control.js
@@ -104,8 +104,9 @@ export const Control = Class.extend({
 	},
 
 	_refocusOnMap(e) {
-		// if map exists and event is not a keyboard event
-		if (this._map && e && e.screenX > 0 && e.screenY > 0) {
+		// We exclude keyboard-click event to keep the focus on the control for accessibility.
+		// The position of keyboard-click events are x=0 and y=0.
+		if (this._map && e && !(e.screenX === 0 && e.screenY === 0)) {
 			this._map.getContainer().focus();
 		}
 	}
--- a/spec/suites/control/Control.LayersSpec.js
+++ b/spec/suites/control/Control.LayersSpec.js
@@ -377,4 +377,37 @@ describe('Control.Layers', () => {
 			expect(elems[4].innerHTML.trim()).to.be.equal('Marker A');
 		});
 	});
+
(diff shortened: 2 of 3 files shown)


mourner (commented):
Really nice catch, thanks for submitting a fix — looks good to me! How easy would it be to add a test case for this so that the behavior doesn't regress again in the future?

src/control/Control.js
```diff
@@ -104,8 +104,9 @@ export const Control = Class.extend({
 	},
 
 	_refocusOnMap(e) {
-		// if map exists and event is not a keyboard event
-		if (this._map && e && e.screenX > 0 && e.screenY > 0) {
+		// We exclude keyboard-click event to keep the focus on the control for accessibility.
+		// The position of keyboard-click events are x=0 and y=0.
+		if (this._map && e && !(e.screenX === 0 && e.screenY === 0)) {
```
Falke-Design, line 109: I changed this check because `e.screenX > 0 && e.screenY > 0` didn't worked on my second screen because `screenX` was `-1700`

Falke-Design (commented):
I added a test and also changed the screen coord check.

Falke-Design (approved):

Issue #9003, opened 2023-06-27
SetView 'duration' option does not work.


jia-mx:
### Checklist

- [ ] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [ ] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Expected behavior

map.setView(point, zoom, {animate: true,duration: 10});
SetView 'duration' option does not work.

### Current behavior

SetView 'duration' option does not work.

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Chrome 113.0.5672.64
- OS/Platform (with version): Windows 10 laptop

Falke-Design:
Works fine for me: https://plnkr.co/edit/fZu7HJNlWqEC94hw

Keep in mind, if the [`zoomAnimationThreshold`](https://leafletjs.com/reference.html#map-zoomanimationthreshold) is to high, it will be not animated.

jia-mx:
Thanks for your reply, but if duration is set to any value, the animation duration of the map is the same.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "Leaflet/Leaflet"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2023年7月1日(星期六) 晚上11:37
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [Leaflet/Leaflet] SetView 'duration' option does not work. (Issue #9003)





 
Works fine for me: https://plnkr.co/edit/fZu7HJNlWqEC94hw
 
Keep in mind, if the zoomAnimationThreshold is to high, it will be not animated.

Falke-Design:
Ok I see, when there is a zoom difference, the `duration` is not applied.

mohasabbasi:
4c2ff8e

movahhedi:
> From my issue #9614 

Zoom speed is too quick for my need. It is currently set to 250ms.

I found the 250ms set here:
[...]

PR #9020, opened 2023-07-07
Apply animation duration to zoom

Falke-Design:
The option `duration` is not applied to zoom animation.

Fixes:  #9003

zoom-duration


src/map/Map.js
```diff
@@ -180,11 +180,8 @@ export const Map = Evented.extend({
 		this._stop();
 
 		if (this._loaded && !options.reset && options !== true) {
-
-			if (options.animate !== undefined) {
-				options.zoom = Util.extend({animate: options.animate}, options.zoom);
-				options.pan = Util.extend({animate: options.animate, duration: options.duration}, options.pan);
-			}
```
Falke-Design, line 187: @mourner do you know, why if `if (options.animate !== undefined) {` exists? The problem is that the options like the duration are not applied until `animate: true` is passed in the options. `map.setView([50.5, 0], 9, {duration: 10, animate: true}); `

https://github.com/Leaflet/Leaflet/commit/260a869da6e6519a883bf733c7ca7caa34ce1540

mourner: @Falke-Design the original change comes in #1617, and looking at the description, I think the logic was that without explicit `animate: true`, you have no guarantee that it will animate (it could instantly reset to new location if it's far away enough), so instead the behavior is fully heuristic instead of being based on specific options. But I'm not fully sure, this PR was more than 10 years ago.

PR #9021, opened 2023-07-08
Remove dot in signature

WebSnke:


jonkoops:
Thanks for opening a PR, but I see no value in landing this change.

WebSnke:
> Thanks for opening a PR, but I see no value in landing this change.

But that's the correct way to sign a letter.

Yours truly,
WebSnke

WebSnke:
[This](https://chat.openai.com/share/96c7adae-55fb-4d01-a3f8-1ac69c89749f) is what ChatGPT thinks.

jonkoops:
I really don't care what ChatGTP thinks, it's not a reliable source of information. We do not accept AI generated contributions because they are of low quality and waste out time.

WebSnke:
> I really don't care what ChatGTP thinks, it's not a reliable source of information. We do not accept AI generated contributions because they are of low quality and waste our time.

Ok, then look at [this](https://www.indeed.com/career-advice/career-development/ways-to-sign-a-letter).
This source confirms that you would not end it with a dot.

PR #9025, opened 2023-07-11
user _limitZoom in flyTo, like we do in resetView

yohanboniface:
Thanks :)

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1418,6 +1418,24 @@ describe('Map', () => {
 
 			map.flyTo(dc, 4, {duration: 0.1});
 		});
+
+		it('flyTo should honour maxZoom', function (done) {
+			this.timeout(10000); // This test takes longer than usual due to frames
+
+			const newCenter = latLng(10, 11),
+			    maxZoom = 20;
+			map.options.maxZoom = maxZoom;
+
+			map.setView([0, 0], 0);
+
+			map.on('zoomend', () => {
+				expect(map.getCenter()).to.eql(newCenter);
+				expect(map.getZoom()).to.eql(maxZoom);
+				done();
+			});
+
+			map.flyTo(newCenter, 22);
+		});
 	});
 
 	describe('#zoomIn and #zoomOut', () => {
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1419,7 +1419,7 @@ describe('Map', () => {
 			map.flyTo(dc, 4, {duration: 0.1});
 		});
 
-		it('flyTo should honour maxZoom', () => {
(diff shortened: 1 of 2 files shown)


spec/suites/map/MapSpec.js
```diff
@@ -1418,6 +1418,24 @@ describe('Map', () => {
 
 			map.flyTo(dc, 4, {duration: 0.1});
 		});
+
+		it('flyTo should honour maxZoom', function (done) {
+			this.timeout(10000); // This test takes longer than usual due to frames
```
Falke-Design, line 1423: Can we speed up the test with setting `animate: false`? `map.flyTo(newCenter, 22, {animate:false});`

Falke-Design (changes requested):

spec/suites/map/MapSpec.js
yohanboniface: I'd say that doing so, we'll not pass in the animation code, which is where the unbounded zoom was. :/

Falke-Design: Yes that is true, sorry I didn't saw that. But please change the duration and remove the timeout. `map.flyTo(newCenter, 22, {animate: true, duration: 0.1});`

yohanboniface: Done! Thanks for the suggestion :)

spec/suites/map/MapSpec.js
```diff
@@ -1418,6 +1418,21 @@ describe('Map', () => {
 
 			map.flyTo(dc, 4, {duration: 0.1});
 		});
+
+		it('flyTo should honour maxZoom', () => {
```
Falke-Design, line 1422:
```suggestion
		it('flyTo should honour maxZoom', (done) => {
```

spec/suites/map/MapSpec.js
```diff
@@ -1418,6 +1418,21 @@ describe('Map', () => {
...
+			const newCenter = latLng(10, 11),
+			    maxZoom = 20;
+			map.options.maxZoom = maxZoom;
+
+			map.setView([0, 0], 0);
+
+			map.on('zoomend', () => {
+				expect(map.getCenter()).to.eql(newCenter);
+				expect(map.getZoom()).to.eql(maxZoom);
```
Falke-Design, line 1431:
```suggestion
				expect(map.getZoom()).to.eql(maxZoom);
				done();
```

Falke-Design (changes requested):
We need the `done` because else if `zoomend` is never called, it would be still successfull

yohanboniface:
> We need the `done` because else if `zoomend` is never called, it would be still successfull

Ah, damn, you're right!

Falke-Design (approved):
Looks good now, thank you!

PR #9029, opened 2023-07-13
Add possibility to create coverage reports

Falke-Design:
Looks like this https://florian-bischof.at/leaflet/coverage/html/


jonkoops:
Considering we'll be moving away from Karma in the near future, I wonder if it's worth it to merge this in? Still a great enhancement, and I personally have no issue with landing this now.

jonkoops (approved):

Falke-Design:
> Considering we'll be moving away from Karma in the near future, I wonder if it's worth it to merge this in?

I now, thanks. But I think it make sense to have this report. We should also add a coverage report with the new tool in the future.

PR #9033, opened 2023-07-16
Include `prosthetic-hand` from GitHub

jonkoops:
Since we still have to refactor more of `prosthetic-hand` we cannot do a release at this point. So in the mean time, we can include it directly from GitHub until we feel it's stable enough for a new release.

Needed to include  https://github.com/Leaflet/prosthetic-hand/pull/41


jonkoops:
Everything looks green, @Falke-Design can I ask you to review this one?

Falke-Design (approved):

Issue #6710, opened 2019-07-01
docs: Draggable works on all objects


tobwen:
- [X] I've looked at the [documentation](http://leafletjs.com/reference.html) to make sure the behavior is documented and expected
- [X] I'm sure this is a Leaflet code issue, not an issue with my own code nor with the framework I'm using (Cordova, Ionic, Angular, React…)
- [X] I've searched through the issues to make sure it's not yet reported

**Steps to reproduce**
1. Read docs: https://leafletjs.com/reference-1.5.0.html#draggable

**Expected behavior**
`Only works for elements that were positioned with L.DomUtil.setPosition.`

**Current behavior**
`Works on all elements`

**Environment**
n/a

**Additional context**
Please don't change the functionality, please fix the docs :-)

**Minimal example reproducing the issue**
https://codepen.io/1100101/pen/YoejoY

- [X] this example is as simple as possible
- [X] this example does not rely on any third party code

sashadev-sky:
I was wondering about this too! It doesn't work perfectly as in your example in some scenarios. For ex. I have an `ImageOverlay` that was positioned with projective transformations:

![ov](https://user-images.githubusercontent.com/41092741/61300970-04050a80-a7b1-11e9-972a-52a493319030.gif)

Since I didn't set my position with `L.DomUtil.setPosition`, it always assumes that my start point is 0 and turns the image there first before continuing to drag

Setting the position prevents this behavior.

Would love some more documentation surrounding this!

PR #9041, opened 2023-07-19
Update docs L.draggable works for any element in the DOM

AndressaBiagi:


jonkoops:
It looks like this PR has neither an issue associated with it, not a description. It's generally considered good practice to motivate proposed changes to eliminate guesswork.

Can you elaborate on why this change needs to be made? Can you provide an example of a draggable element that has not been positioned on the map first?

AndressaBiagi:
actually there is an issue with an example on codeSandbox, 'docs: Draggable
works on all objects #6710' https://github.com/Leaflet/Leaflet/issues/6710

Em dom., 23 de jul. de 2023 às 13:35, Jon Koops ***@***.***>
escreveu:

PR #9048, opened 2023-07-30
Clean up DOM event listeners when destroying Map's animation proxy

### LINKED ISSUE #9046: Memory leak: animation proxy 'transitionend' event listener never removed
### Checklist

### Steps to reproduce

Go to the constructor of `Map` and find:

```js
DomEvent.on(this._proxy, 'transitionend', this._catchTransitionEnd, this);
```

There is no removal of the event listener in the teardown code for `Map`, which means that creating/tearing down maps repeatedly will leak memory over time.

### Expected behavior

`Map.remove()` should contain the following:

```js
DomEvent.off(this._proxy, 'transitionend', this._catchTransitionEnd, this);
```

Alternatively, it could be done in an `'unload'` event listener.

### Current behavior

Memory leak.

_No response_

### Environment

- Leaflet version: 1.9.2
- Browser (with version): any
- OS/Platform (with version): M1 Max Macbook Pro

samclaus:
In the constructor for `Map`, a listener is added for the `'transitionend'` event of the animation proxy element (if zoom animation is enabled):

```js
DomEvent.on(this._proxy, 'transitionend', this._catchTransitionEnd, this);
```

This PR adds a `DomEvent.off()` call to clean up that listener inside of the `Map._destroyAnimProxy()` method.

Closes #9046.

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1646,6 +1646,7 @@ export const Map = Evented.extend({
 	},
 
 	_destroyAnimProxy() {
+		DomEvent.off(this._proxy, 'transitionend', this._catchTransitionEnd, this);
 		this._proxy.remove();
 		this.off('load moveend', this._animMoveEnd, this);
 		delete this._proxy;
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -1651,9 +1651,9 @@ export const Map = Evented.extend({
 	},
 
 	_animMoveEnd() {
-		const
-			c = this.getCenter(),
-		    z = this.getZoom();
+		const c = this.getCenter(),
+				z = this.getZoom();
+
 
 		DomUtil.setTransform(
 			this._proxy,
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -159,7 +159,6 @@ export const Map = Evented.extend({
 		// happens after starting zoom animation (propagating to the map pane), we know that it ended globally
 		if (this._zoomAnimated) {
 			this._createAnimProxy();
-			DomEvent.on(this._proxy, 'transitionend', this._catchTransitionEnd, this);
 		}
 
 		this._addLayers(this.options.layers);
@@ -781,6 +780,8 @@ export const Map = Evented.extend({
 			// Fired when the map is destroyed with [remove](#map-remove) method.
 			this.fire('unload');
 		}
+
+		this._destroyAnimProxy();
 
 		let i;
 		for (i in this._layers) {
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Thank you for making this PR. Can you please apply the requested changes.


Can you please move this code into a function `_animateProxyZoom` and then remove the `zoomanim` event in `_destroyAnimProxy` too. `this.on('zoomanim',this._animateProxyZoom, this);`
https://github.com/Leaflet/Leaflet/blob/847fbb4d62d54abd51b7681595c3d5cec0502d7a/src/map/Map.js#L1629-L1638

samclaus:
@Falke-Design May I simply add a `this.off()` call to the `Map.remove()` method? I believe the map is already unusable after removal, so that simply removing all event listeners in a single shot makes a lot of sense for performance AND JavaScript bundle size (because we don't need a ton of individual event listener removals)? Perhaps that is out of the scope of this PR, but I would gladly open another one to do that if it is okay with you. I will make the requested changes for this PR and postpone that other idea.

samclaus:
@Falke-Design I made the following changes:

1. The `Map` constructor originally had the following:

    ```js
    // zoom transitions run with the same duration for all layers, so if one of transitionend events
    // happens after starting zoom animation (propagating to the map pane), we know that it ended globally
    if (this._zoomAnimated) {
        this._createAnimProxy();
        DomEvent.on(this._proxy, 'transitionend', this._catchTransitionEnd, this);
    }
    ```
    
    I moved the `'transitionend'` listener to the end of `_createAnimProxy()` so it is grouped with the rest of the proxy setup code.
    
2. I extracted the `'zoomanim'` event closure into an `_animateProxyZoom(e)` method like you requested.

3. I changed the `_destroyAnimProxy()` from an `'unload'` listener to a direct call inside of `Map.remove()`, like you requested. Because of this change, an `if` guard is required to make sure the animation proxy was created in the first place, and I opted to place the guard inside of `_destroyAnimProxy()` instead of wrapping the method call inside of `remove()`, so that the method is idempotent and less dangerous to call, with no harm to performance or code size or readability.

4. I reordered the proxy-related methods into the order they are used:

    1. `_createAnimProxy()`
    2. `_animateProxyZoom(e)`
    3. `_animMoveEnd()`
    4. `_destroyAnimProxy()`
    
5. I cleaned up the indentation of code within the proxy-related methods, e.g.:

    ```js
    // BEFORE (very long line)
    DomUtil.setTransform(this._proxy, this.project(e.center, e.zoom), this.getZoomScale(e.zoom, 1));
    
    // NOW
    DomUtil.setTransform(
	    this._proxy,
	    this.project(e.center, e.zoom),
	    this.getZoomScale(e.zoom, 1),
    );
    ```

src/map/Map.js
```diff
@@ -1622,36 +1623,54 @@ export const Map = Evented.extend({
...
+		if (transform === this._proxy.style.transform && this._animatingZoom) {
+			this._onZoomTransitionEnd();
+		}
+	},
 
-		this._on('unload', this._destroyAnimProxy, this);
+	_animMoveEnd() {
+		const
+			c = this.getCenter(),
+		    z = this.getZoom();
```
Falke-Design, line 1653:
```suggestion
		const c = this.getCenter(),
				z = this.getZoom();

```

Falke-Design (approved):
Looks good

src/map/Map.js
```diff
@@ -1622,36 +1623,54 @@ export const Map = Evented.extend({
...
+		// workaround for case when transform is the same and so transitionend event is not fired
+		if (transform === this._proxy.style.transform && this._animatingZoom) {
+			this._onZoomTransitionEnd();
+		}
+	},
 
-		this._on('unload', this._destroyAnimProxy, this);
+	_animMoveEnd() {
+		const c = this.getCenter(),
+				z = this.getZoom();
```
Falke-Design, line 1652:
```suggestion
		const c = this.getCenter();
		const z = this.getZoom();

```

samclaus:
@jonkoops Pinging because it has been almost 3 months. Can this fix be merged?

Falke-Design:
@samclaus I know this PR is now really old but maybe you can remember why this is necessary? Is the listener not obsolete when `delete this._proxy` is called?

samclaus:
@Falke-Design I believe the event listener is never removed, which is why I made this PR. Because the event listener is bound to the Map instance, I wouldn't be surprised if instances of Map never get garbage collected. I do not know if the browser is smart enough to see that the DOM elements got removed from the page and determine that it is safe to garbage collected them plus any event listeners they have plus the Map instance. Sounds too good to be true though.

Falke-Design:
@samclaus 
> I believe the event listener is never removed, which is why I made this PR. Because the event listener is bound to the Map instance

I first thought the same but then I realised that the listener is applied to `this._proxy` and not on the `map`:
```
DomEvent.on(this._proxy, 'transitionend', this._catchTransitionEnd, this);
```

Or am I overlooking something?

samclaus:
@Falke-Design When the listener is registered, it is keeping a reference to the function/method passed, as well as "this" (aka the Map instance), in an event handler map attached to the _proxy DOM element. Because the DOM element has an event listener registered, it never gets garbage collected and neither do any objects (including the Map) referenced by it.

Falke-Design:
@samclaus thank you!

PR #9054, opened 2023-08-02
Plugins: add Leaflet.QgsMeasure

gabriel-russo:
Add leaflet-qgsmeasure info in plugins docs


Falke-Design (approved):
I recommand to create a new repository and not a fork.

gabriel-russo:
> I recommand to create a new repository and not a fork.

There is any disadvantage being a fork?

Falke-Design:
It is not a real disadvantage but some users will think, why I should use the fork when I can use the "original" library.

PR #9055, opened 2023-08-03
Add `setInteractive`

### LINKED ISSUE #5442: Feature: new method setInteractive(interactive) for interactive layers

someonewithpc:
Closes #5442 

I'll note the I couldn't get the popup to trigger the event handler (`UIEventSimulator.fire('click', popup._container);`). I'm also not sure if the line endings are correct, as I tried messing with that in many ways and some diffs show some `^M`, regardless of what I did


someonewithpc:
@Falke-Design Could you take a look at this? I had to publish a fork so my company's project could proceed, and I'd rather have it merged. Let me know if something is missing or isn't the best approach

someonewithpc:
@Malvoz Any chance you could review this? I'm currently applying a patch, to use it at my company, and I'd rather move away from that

heyman:
Just chiming in that this feature would be useful 😀

stathis1998:
@someonewithpc Hey man, very nice work adding this functionality, lost so many hours trying to make this work and then realizing its not possible. Since I don't see the team merging this, would you mind me borrowing the code ?

someonewithpc:
> would you mind me borrowing the code ?

you don't need my permission, just follow the license :)

PR #9061, opened 2023-08-11
Update docs l.draggable works on all DOM elements

AndressaBiagi:
Example: 'docs: Draggable
works on all objects [#6710](https://github.com/Leaflet/Leaflet/issues/6710)' [#6710](https://github.com/Leaflet/Leaflet/issues/6710)


Falke-Design (approved):

Falke-Design (changes requested):
Need to be changed in the source file

https://github.com/Leaflet/Leaflet/blob/35d79e012b802b61ecf958e2b956a24aae6ef62c/src/dom/Draggable.js#L14

Issue #9059, opened 2023-08-07
Tooltips get stuck while panning


BathoryPeter:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

If you have several objects with tooltip close to each other, the open tooltip and nearby tooltips will be stuck on the screen after panning the map.

1. Open JSFiddle link
2. Move cursore over the area with circle markers (a tooltip will pop up)
3. Pan the map
4. Repeat steps 2. and 3. a few times.

Nearby tooltips appears and stuck there.

![tooltip_sticks](https://github.com/Leaflet/Leaflet/assets/19800037/c7be1df0-fdc9-4160-9b0c-459ddc5dda55)

### Expected behavior

After panning, only the already opened tooltip should be seen.

### Current behavior

Several tooltips stuck on the screen.

### Minimal example reproducing the issue

https://jsfiddle.net/BathoryPeter/qztj41c3/2/

```
let markers = new L.FeatureGroup().addTo(map);

[...Array(20).keys()].map(i => {
	[...Array(30).keys()].map(j => {
    	let lat = 47 + i / 100;
      let lon = 19 + j / 100;
      
      L.circleMarker([lat, lon])
      	.bindTooltip(`${lat} ${lon}`)
        .addTo(markers);
    });
});

map.fitBounds(markers.getBounds());
```

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Vivaldi 6.1.3035 (latest stable)
- OS/Platform (with version): Ubuntu 23.04

dscarb21:
This seems to be a legit bug, has been replicated on:
Windows 10.0.19045
Chrome 115.0.5790.171

HappyViki:
I was able to reproduce stuck tooltip with 2 markers. When wiggling the map with a tooltip over another marker with a hidden tooltip, it seems to open the tooltip on the other dragged over marker after letting go of the map.

Before wiggle:
[...]

PR #9065, opened 2023-08-15
#9059 Fix - Removed convoluted map-moving / tooltip showing logic

dscarb21:
Code was deemed to be unecessary and leads to multiple tooltips showing up at once. Removing allows for clean functionality. 

Fixes: #9059

### DIFF
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -417,15 +417,6 @@ Layer.include({
 			return;
 		}
 
-		// If the map is moving, we will show the tooltip after it's done.
-		if (this._map.dragging && this._map.dragging.moving() && !this._openOnceFlag) {
-			this._openOnceFlag = true;
-			this._map.once('moveend', () => {
-				this._openOnceFlag = false;
-				this._openTooltip(e);
-			});
-			return;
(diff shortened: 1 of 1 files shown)


IvanSanchez:
You say «leads to multiple tooltips showing up at once». Under what circumstances does this happen?

Falke-Design (changes requested):
We should not remove this without having a solution for #8672, which was the purpose for this code

Falke-Design:
> You say «leads to multiple tooltips showing up at once». Under what circumstances does this happen?

@IvanSanchez I think this should fix #9059

tooltip_sticks

dscarb21:
Apolgies, didn't fully understand functionality of #8672 fix. Changed the flag that is used to "schedule" tooltip opening to a property on the map object rather than the tooltip itself. This fixes all of those tooltips displayed at once, demonstrated by @Falke-Design. Set this flag to "true" in closetooltip which gets rid of a small bug where it opens the tooltip even when the cursor has left the marker.

jonkoops:
Superseded by https://github.com/Leaflet/Leaflet/pull/9154

PR #9100, opened 2023-09-22
Possibility to refresh GridLayer tiles

yuri-karelics:
Hi, please consider the possibility to merge this PR.

**Background**
We are happy to use Leaflet in our Karelics.Cloud platform. One problem we faced with is map tiles update. If map was changed and client wants to update tiles the only available GridLayer method is 'redraw' - it deletes all tiles and creates them back. Blinking caused by tiles removal looks bad in case updates happen often.

**What was done**
In general, GridLayer logic was not touched. Just piece of code was extracted from __addTile_ into __initTileElement_ method to be reused.
Only new thing is method **_reloadTiles_**. It creates new HTML elements for each existing tile and replaces tile's HTML element with new one on success. There is no tile removal, so visually this way to update tiles looks much better.

Give me know in case some edge case was not considered. Thanks!

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -272,6 +272,17 @@ export const GridLayer = Layer.extend({
 		return document.createElement('div');
 	},
 
+	// @method refreshTiles: void
+	// Triggers creating of new tile HTML element for each existing tile
+	refreshTiles() {
+		for (const key in this._tiles) {
+			if (Object.hasOwn(this._tiles, key)) {
+				const tile = this._tiles[key];
+				this._initTileElement(tile.coords, this._tileUpdate.bind(this, tile));
+			}
+		}
+	},
+
 	// @section
 	// @method getTileSize: Point
 	// Normalizes the [tileSize option](#gridlayer-tilesize) into a point. Used by the `createTile()` method.
@@ -794,6 +805,24 @@ export const GridLayer = Layer.extend({
 		});
 	},
 
+	_initTileElement(coords, done) {
+		const el = this.createTile(this._wrapCoords(coords), done);
+
+		this._initTile(el);
+
+		// if createTile is defined with a second argument ("done" callback),
+		// we know that tile is async and will be ready later; otherwise
+		if (this.createTile.length < 2) {
+			// mark tile as ready, but delay one frame for opacity animation to happen
+			Util.requestAnimFrame(done.bind(this, null, el));
+		}
+
+		const tilePos = this._getTilePos(coords);
+		DomUtil.setPosition(el, tilePos);
+
+		return el;
+	},
+
 	_initTile(tile) {
 		tile.classList.add('leaflet-tile');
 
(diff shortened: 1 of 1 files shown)


Falke-Design:
I haven't tested it but the browser cache should break that logic, because it doesn't make a new request.
There should be another PR where we discussed that already.

I understand the need and i think if it is possible we should consider implementing a solution for this issue.

yuri-karelics:
I would say 'avoiding browser cache' is responsibility of _createTile_ method, it could be anything, not only , so maybe browser cache is not a problem at all. Responsibility of GridLayer is just to get new element and replace old one.

yuri-karelics:
@Falke-Design @mourner is there a chance to move this PR forward? I am open for discussion

yuri-apanasik:
@Falke-Design @mourner could you give me the person to contact regarding possibility to review this MR?

src/layer/tile/GridLayer.js
```diff
@@ -891,6 +908,27 @@ export const GridLayer = Layer.extend({
 		}
 	},
 
+	_tileUpdate(tile, err, el) {
```
IvanSanchez, line 911: This method is used only once (in `refreshTiles()`) and therefore I'd be tempted to hoist it in there, converting it into an anonymous function.

src/layer/tile/GridLayer.js
```diff
@@ -891,6 +908,27 @@ export const GridLayer = Layer.extend({
...
+			});
+			return;
+		}
+
+		el.classList.add('leaflet-tile-loaded');
+		tile.el.replaceWith(el);
+		tile.el = el;
+		tile.loaded = +new Date();
+
+		this.fire('tileupdate', {
```
IvanSanchez, line 926: This event needs docstrings.

src/layer/tile/GridLayer.js
```diff
@@ -272,6 +272,17 @@ export const GridLayer = Layer.extend({
 		return document.createElement('div');
 	},
 
+	// @method refreshTiles: void
+	// Triggers creating of new tile HTML element for each existing tile
```
IvanSanchez, line 276: Compare with the `redraw` method. Give users some explanation (in the docstrings) why they should use one or the other.

src/layer/tile/GridLayer.js
```diff
@@ -891,6 +908,27 @@ export const GridLayer = Layer.extend({
...
+			this.fire('tileerror', {
+				error: err,
+				tile: el,
+				coords: tile.coords
+			});
+			return;
+		}
+
+		el.classList.add('leaflet-tile-loaded');
+		tile.el.replaceWith(el);
```
IvanSanchez, line 922: I see, the element is created behind the curtains and replaced in the DOM once it's ready. Nice technique. :+1:

IvanSanchez (changes requested):

src/layer/tile/GridLayer.js
yuri-karelics: explanation extended

yuri-karelics:
@IvanSanchez just realized that probably I should create MR to 'v1' branch... or you care about cherry-picks yourself?

Falke-Design:
> @IvanSanchez just realized that probably I should create MR to 'v1' branch... or you care about cherry-picks yourself?

Hi, 
no the `main` branch is fine, we will not add new features to `v1` anymore.

yuri-karelics:
@Falke-Design @IvanSanchez is there a change to merge this PR this year? Also, do you already have a plan regarding v2 release? If no, I can create also PR to v1 branch.
P.S. We really want to remove ugly patch we are using in our code to achieve behavior implemented in this PR

yuri-karelics:
@Falke-Design @IvanSanchez @Malvoz resolved conflict, hope to see these changes released soon 👍

Issue #9096, opened 2023-09-18
Update Twitter logo and name


WebSnke:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Expected behavior

The Twitter logo and name should be updated.

### Current behavior

The website still uses the old logo and name.

### Environment

- Browser (with version): Firefox 116
- OS/Platform (with version): Fedora 38

Utkarsh3128:
Can you tell me how to run the source code on local machine?

WebSnke:
> Can you tell me how to run the source code on local machine?

Sure. You probably want to see the changes you made in your browser, right? Do you use VS Code or something else?

Utkarsh3128:
Yes sir, that's the reason I asked. I use VS Code =.

WebSnke:
> Yes sir, that's the reason I asked. I use VS Code.

Ok. I suggest that you install Live Server. It's a VS Code extension that allows you to start a local server to view your website with just one click.

Live Server:

![image](https://github.com/Leaflet/Leaflet/assets/94064167/f9db4e78-895a-4e56-ab2e-4e42a46b7a0d)


Once you installed the extension, just go back to your project and click on the 'Go Live' button in the info bar of VS Code.

![image](https://github.com/Leaflet/Leaflet/assets/94064167/75741817-ef46-4a47-8a5e-7fde18fc6c3e)

ShambhavII-Singh:
Hi can you assign this issue to me? I would like to work on this.

aialok:
Hey ! I have done the pull request with proper code and screenshot do merge it.

love-sonkar:
hello can someone help me to run this project localy im using vs Code and i have install the live server but when i start live server i got this file .
![localhost](https://github.com/Leaflet/Leaflet/assets/82786865/38f54a25-8f87-4c5b-8fdf-df907a2d1364)

aialok:
[...]

PR #9102, opened 2023-09-24
Updated Twitter Logo and Name

aialok:
### Updated twitter logo to x.com logo

- I have updated previous Twiiter svg file to new x.com svg files.
-  Deleted ununsed twiiter.png

Screenshot 2023-09-23 154259

**This is my first contribution to open source world.
Thank you !
Happy coding !**

Fixes: #9096

PR #9110, opened 2023-10-01
Logo and name update of twitter 

SajjanYadav:
Solves #9096


Falke-Design:
Thank you for your contribution but this is already fixed by #9102

PR #9114, opened 2023-10-05
Fix closest layer point edge case

### LINKED ISSUE #9098: closestLayerPoint doesn't find point on the edge between last and first point of Polygon
### Checklist

### Steps to reproduce

1. Create a triangle polygon let it be [[0,0], [1,1], [0,1]]
2. Choose a point on it's edge between last and first point -- let it be [0, 0.5]
3. Try to find closest layer point: instead of the same point, algorithm would return points on other edges.

Reason: Even though Polygon extends Polyline, it doesn't fix polyline's closestLayerPoint method, 
which doesn't take in to account this edge [length-1, 0]
https://github.com/Leaflet/Leaflet/blob/641a3820b0955f092523f7b7b842eab491dd8441/src/layer/vector/Polyline.js#L98
That's correct for polyline, but not for polygon.

### Expected behavior

since point [0, 0.5] lies on the edge [[0, 0], [0, 1]], i'd expect closestLayerPoint to return point, that would be reprojected back to [0, 0.5]

### Current behavior

closestLayerPoint returns point, that is reprojected back to [0.25, 0.25].

https://plnkr.co/edit/BmMqSBJYM3QoOiO6

### Environment

- Leaflet version: Leaflet 1.9.4+v1.d15112c,
- Browser (with version): Chrome 116.0.5845.190 (Official Build) (64-bit) (cohort: Control)
- OS/Platform (with version): Win 10

MadhukarRagaji:
This pull request addresses an edge case in the closestLayerPoint function where it did not properly consider the segment between the last and first points of a Polygon.

Details:

    Identified that the closestLayerPoint function was missing the edge case where the point could be on the segment between the last and first points of a Polygon.
    Modified the iteration logic to ensure this segment is also evaluated when finding the closest layer point.

Testing:

    Set up a local test environment with a triangle polygon and markers. Confirmed that the function now correctly identifies points on the edge between the last and first points of the Polygon.
    All existing tests pass, and no regressions were identified.

Related Issues:

    Fixes  #9098

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -88,19 +88,19 @@ export const Polyline = Path.extend({
 	// Returns the point closest to `p` on the Polyline.
 	closestLayerPoint(p) {
 		let minDistance = Infinity,
-			minPoint = null,
-			p1, p2;
+		minPoint = null,
+		p1, p2;
 		const closest = LineUtil._sqClosestPointOnSegment;
-	
+
 		for (let j = 0, jLen = this._parts.length; j < jLen; j++) {
 			const points = this._parts[j];
-	
+
 			for (let i = 1, len = points.length; i <= len; i++) {
 				p1 = points[i - 1];
 				p2 = points[i % len];
-	
+
 				const sqDist = closest(p, p1, p2, true);
-	
+
 				if (sqDist < minDistance) {
 					minDistance = sqDist;
 					minPoint = closest(p, p1, p2);
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -88,19 +88,19 @@ export const Polyline = Path.extend({
 	// Returns the point closest to `p` on the Polyline.
 	closestLayerPoint(p) {
 		let minDistance = Infinity,
-		    minPoint = null,
-		    p1, p2;
+			minPoint = null,
+			p1, p2;
 		const closest = LineUtil._sqClosestPointOnSegment;
-
+	
 		for (let j = 0, jLen = this._parts.length; j < jLen; j++) {
 			const points = this._parts[j];
-
-			for (let i = 1, len = points.length; i < len; i++) {
+	
+			for (let i = 1, len = points.length; i <= len; i++) {
 				p1 = points[i - 1];
-				p2 = points[i];
-
+				p2 = points[i % len];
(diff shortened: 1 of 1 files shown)


MadhukarRagaji:
#9098

se-ti:
This PR seems to break Polyline.closestLayerPoint, adding new edge between last and first point, and even for a Polygon this shouldn't work when clipping occures.

I'd try to construct polygon with both start and last points outside the screen. This should give reduced _parts arrays, and closestLayerPoint would find a point on the [last point; first point] edge,  which doesn't exist in original polygon.

Serge, author of #9098

Falke-Design:
> I'd try to construct polygon with both start and last points outside the screen. This should give reduced _parts arrays, and closestLayerPoint would find a point on the [last point; first point] edge, which doesn't exist in original polygon.

@se-ti you are right, good catch 👍 but this problem already oocurs on polylines too and has not directly something to do with #9098

Demo Polyline: https://plnkr.co/edit/TzzbNQf5ljXTGRV2
Demo Polygon: https://plnkr.co/edit/7gPOyNiLgGOTDxUs

If I read the docs, I would think it will return always the closest point and has nothing todo with clipping. So my suggestion is to do a refactoring and use the actual latlngs (projected to points) and not the clipped parts.
grafik

se-ti:
I agree about docs, @Falke-Design.
Probably the only aspect where current design is better -- it's much faster for large geometries like countries' borders.

Falke-Design:
If we replace `_parts` with `_rings` everything works fine. https://plnkr.co/edit/J78vmzTx82kPh1Ck
Perfomance could be still an issue, needs to be tested.

@mourner @IvanSanchez @jonkoops do you see any problems if we use `_rings` instead of `_parts` in `closestLayerPoint`? With `_parts` we have the problem that the calcuation is done with the clipped path.

se-ti:
Thought for a while: if you decide to rewrite closestLayerPoint from _parts to _rings, could you please return not only point and distance, but also ring's index, and point's index in ring. (point -- starting point of the edge, closestLayerPoint belongs to).

I used closestLayerPoint's as a second filter (after bounding box) when looking for a nearest edge of large and complex geometries. But to find nearest edge i had to rewrite closestLayerPoint for a latlngs array and call it one more time, just to find those indices.

PR #9142, opened 2023-10-22
Add help for map container keyboard shortcuts

### LINKED ISSUE #9131: Help should be provided for the map container keyboard shortcuts [accessibility]
### Checklist

## Problem ##

Labels or instructions should be provided when content requires user input.

As a keyboard user, I prefer to use the keyboard for _all_ interactions.

It was only some time after I began using Leaflet, that I discovered by chance that you can use arrow keys to pan around the map and the `+` and `-` keys to zoom in and out. They are not obvious!

There should be something like a popup to explain the keyboard shortcuts.

* WCAG 2.2 [Success Criterion 3.3.2 Labels or Instructions](https://www.w3.org/TR/WCAG22/#labels-or-instructions)

## Solution

I'd like to propose something similar to the solution in [Google Maps](https://developers.google.com/maps/documentation/javascript/overview). That is, a “Keyboard shortcuts” button that appears visually at the bottom-right of the map, next to the attribution links.

The button opens a dialog that summarises the available keyboard shortcuts. 

Screenshots to follow!

### Focus Order

Although visually the “Keyboard shortcuts” button can appear at the bottom of the map, it should be before the map container in the keyboard focus order. Otherwise, it won't be available to a keyboard user until they're past the container!

_No response_

### Environment

- Leaflet version: 1.9.4,
- Browser (with version): Chrome 114,
- OS/Platform (with version): Mac OS 10.14.6.

TaduJR:
This change resolves #9131 by adding keyboard shortcuts modal for accessibility.


Falke-Design (changes requested):
If we will add a modal like this, then it should be using the nativ [<dialog>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) methods. Additionally it should be possible to add and remove custom shortcuts.

jonkoops:
Appreciate the effort of creating a PR, but it is very unlikely to get merged until pending discussion under #9131 is resolved. I would instead recommend putting this into a dedicated plugin, which we will more than gladly list on our plugins page if you like.

Closing this for now.

PR #9147, opened 2023-10-23
Create BundlingES6LeafletDocumentation.md

### LINKED ISSUE #5807: Need documentation for bundling ES6 Leaflet modules
Since the transformation to ES6 it should be possible to import only the needed parts of Leaflet and not the whole library.

Is there any documentation, examples, or whatnot, available for this kind of usage?

It would be extremely great to have a simple guide on this which would list inter-dependencies, what you should note when importing just a piece, and so forth. Like "to have this and that functionality, import this" and "if you import that, also this is going to get imported since that depends on it".

Any thoughts?

vishnugaruda246:
Created documentation for bundling ES6 Leaflet Modules

Fixes #5807

### DIFF
--- a/BundlingES6LeafletDocumentation.md
+++ b/BundlingES6LeafletDocumentation.md
@@ -1,0 +1,98 @@
+# Prerequisites:
++ Node.js and npm: If you don't have them installed, download and install them from the official website:   https://nodejs.org/en/download/
+
++ Text Editor: You'll need a code editor to write and edit your JavaScript code. Popular options include Visual Studio Code, Sublime Text, or Atom.
+
+
+# Step 1: Create a New Project
+Open your command prompt or terminal and create a new directory for your project. Then navigate into that directory.
+
+```js
+mkdir my-leaflet-project
+cd my-leaflet-project
+```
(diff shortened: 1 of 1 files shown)


vishnugaruda246:
@Ivan-Perez plz review

Ivan-Perez:
I think you mean @IvanSanchez

vishnugaruda246:
sorry for the confussion @Ivan-Perez

shelar1423:
@jonkoops does this require any changes ? can i work on this issue too ?

jonkoops:
Firstly, thanks for taking the effort to open up a PR. This seems to be a tutorial on how to use Leaflet with a bundler,  specifically WebPack. Although this is valuable information, I don't think it's particularly unique to Leaflet, or any other libraries for that matter.

This kind of documentation is already available [for WebPack](https://webpack.js.org/concepts/), and I think that we don't have to document this separately for Leaflet.

jonkoops:
More background here: https://github.com/Leaflet/Leaflet/issues/5807#issuecomment-1783428989

PR #9157, opened 2023-11-03
Upgrade and cleanup dev dependencies

mourner:
- Remove unused / ancient `istanbul` dep.
- Get `prosthetic-hand` back to semver after releasing the latest state as `v2.0.0`.
- Upgrade all direct and transitive dependencies.


mourner:
BTW, still confused why Bundlemon doesn't do a relative comparison here:

image

Falke-Design (approved):

jonkoops:
> BTW, still confused why Bundlemon doesn't do a relative comparison here:

Yeah, I am honestly not to happy with how Bundlemon has been working out for us. Perhaps we should consider alternatives. At the time we last evaluated this there were unfortunately very little well defined options out there.

Issue #9179, opened 2023-11-23
Docstrings needed for L.Marker.getElement


IvanSanchez:
### Checklist

- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this change hasn't been suggested already.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Which documents need updating?

reference.html

### Motivation

I was looking at #9178, and had to look at the JS code to see how to get a reference to the `HTMLElement` of a `L.Marker`.

### Suggested changes

The `getElement()` method of `L.Marker` needs docstrings.

The docstrings for the `getElement()` method of `L.ImageOverlay` can serve as a starting point.

hollowM1ke:
Hello. Can I still contribute to this issue?

PR #9180, opened 2023-11-26
[9179] docstrings for L.Marker getElement()

yuri-karelics:
Closes https://github.com/Leaflet/Leaflet/issues/9179

### DIFF
--- a/src/layer/marker/Marker.js
+++ b/src/layer/marker/Marker.js
@@ -194,6 +194,9 @@ export const Marker = Layer.extend({
 		return this;
 	},
 
+	// @method getElement(): HTMLElement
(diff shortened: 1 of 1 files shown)


yuri-karelics:
@IvanSanchez I've created PR to close your issue, but not for free :) I would appreciate if you take a look on my another PR https://github.com/Leaflet/Leaflet/pull/9100 (have no clue how to move it forward)
P.S. Last 'getElement()' method without docs is inside L.Path, but it looks a bit tricky - looks like it returns some value only of SVG renderer is used. If I am right and if you want I can add some docstrings there too

IvanSanchez (approved):

Falke-Design (approved):

PR #9201, opened 2023-12-28
Honour coords.z when passed to TileLayer.getTileUrl

yohanboniface:
cf #6004 #6006

Before this change, `coords.z` would not taken into account, and was always overriden by `this._tileZoom`.
In the normal flow, `coords.z` is anyway set from `this._tileZoom` before `getTileUrl` is called (in GridLayer._update). But this prevents using `getTileUrl` out of this `_update` method. There are some scenarios where this is needed (see #5822), or (which is the case in uMap) when wanting to display a tile as a tilelayer preview.


Falke-Design:
It looks like this good be working because of:
https://github.com/Leaflet/Leaflet/blob/bbd264704b05036a281bd44513cf1760d8336b9d/src/layer/tile/GridLayer.js#L691

But I'm not sure if this makes maybe some problems with other plugins.

@IvanSanchez @mourner what do you think?

yohanboniface:
> It looks like this could be working because of:

Yep! We could also fallback to `this._tileZoom` in `_getZoomForUrl`, but it seems to me we should not have both, i.e. settings `coords.z = this._tileZoom` AND the fallback.

> But I'm not sure if this makes maybe some problems with other plugins.

That could indeed be the case. Though for this to happen, it seems that a plugin should have rewritten the entire `GridLayer._update`, which is a huge piece of internal code, so maybe it's a feature if we discover that touching such internal code breaks some external code :p

PR #9222, opened 2024-01-20
Map Unit Test for "mouseEventToLayerPoint"

andresrivas20:
The unit test provided is designed to validate the functionality of the mouseEventToLayerPoint method in Leaflet's Map. This method is responsible for converting a mouse event (like a click) on the map to a layer point, which represents a point in the coordinate system used by the map's layers.


mourner (changes requested):
Can you please fix the linting errors?

PR #9229, opened 2024-01-25
Changed Wording

hollowM1ke:
#8477 
Solution for: https://github.com/Leaflet/Leaflet/issues/8477#issuecomment-1903551247


mourner (approved):
Thanks!

Issue #9230, opened 2024-01-26
Failing Windows job on CI due to `requestAnimFrame` test


mourner:
For the last few months, one single test on the Windows CI test job has been failing, making the whole build red:

```
Chrome Headless 120.0.6099.225 (Windows 10) Util #requestAnimFrame calles a function on next frame, unless canceled FAILED
	Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

I can't for the life of me figure out why it doesn't run the supplied function. I think the only thing that changed is Windows image version — from 20231029.1.0 to 20231115.2.0. Also weird that just that test is failing. One thing I tried is testing whether the native `window.requestAnimationFrame` works fine in the test (in place of `Util.requestAnimFrame`), and it weirdly does. Not sure what else to try. Any ideas? cc @jonkoops

mourner:
After reducing the Chrome window size in CI in #9235, the failures don't seem to happen anymore. 🤷‍♂️

mourner:
Doesn't seem to be a problem anymore

PR #9231, opened 2024-01-26
Speed up Karma runner by narrowing down files Karma serves

mourner:
- Fix CI warnings about too many files open by excluding Karma from serving the whole of `node_modules`; this also seems to speed up CI runs, shaving off about a minute.
- ~~Skip `requestAnimFrame` test for now because it fails on Windows for mystery reasons https://github.com/Leaflet/Leaflet/issues/9230 — at this moment it's more important to have a functioning CI that's not red all the time.~~


mourner:
Nope, now Windows fails on a different unrelated test. I'm stumped.

jonkoops (commented):
LGTM

PR #9232, opened 2024-01-26
Unbind tooltip remove focus listeners

### LINKED ISSUE #9071: unbindTooltip does not remove focus listeners
### Checklist

### Steps to reproduce

1. Open this [jsfiddle link](https://jsfiddle.net/bielfrontera/50Lcfs96/8/)
2. Wait two seconds until the tooltip is unbinded
3. Click on the the marker. You'll see this error message at the browser console:
``` 
Tooltip.js:399 Uncaught TypeError: Cannot set properties of null (setting '_source')
    at e.<anonymous> (Tooltip.js:399:19)
    at SVGPathElement.o (DomEvent.js:108:13)
```

### Expected behavior

After the tooltip is unbinded, focus listeners should be removed (or at least, that listener should ensure that `this._tooltip` is not `null`)

### Current behavior

An error is raised if you click on a layer with an unbinded tooltip. 

https://jsfiddle.net/bielfrontera/50Lcfs96/8/

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Google Chrome v 115.0.5790.170
- OS/Platform (with version): Ubuntu 20.04

hollowM1ke:
My solution for fixes #9071

### DIFF
--- a/spec/suites/layer/TooltipSpec.js
+++ b/spec/suites/layer/TooltipSpec.js
@@ -580,4 +580,22 @@ describe('Tooltip', () => {
 		layer2.openTooltip();
 		expect(spy2.called).to.be.true;
 	});
+
+	// Related to #9071
+	it('removes focus listeners after unbinding tooltip', () => {
+		const marker = new CircleMarker([51.515, -0.09], {
+		  radius: 20,
+		  color: 'red',
+		}).addTo(map);
+
+		marker
+		  .bindTooltip('Tooltip that will be unbinded in two seconds')
+		  .openTooltip();
+
+		marker.unbindTooltip();
+
+		setTimeout(() => {
+			expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
+		}, 2000);
+	});
 });
--- a/spec/suites/layer/TooltipSpec.js
+++ b/spec/suites/layer/TooltipSpec.js
@@ -581,21 +581,35 @@ describe('Tooltip', () => {
 		expect(spy2.called).to.be.true;
 	});
 
-	// Related to #9071
-	it('removes focus listeners after unbinding tooltip', () => {
-		const marker = new CircleMarker([51.515, -0.09], {
-		  radius: 20,
-		  color: 'red',
-		}).addTo(map);
+	it('removes focus listeners after unbinding tooltip from Layer', () => {
+		const marker = new Marker([51.515, -0.09]).addTo(map);
 
 		marker
-		  .bindTooltip('Tooltip that will be unbinded in two seconds')
-		  .openTooltip();
+			.bindTooltip('Tooltip that will be unbinded')
+			.openTooltip();
+
+		expect(marker.getElement()._leaflet_focus_handler).to.be.not.undefined;
 
 		marker.unbindTooltip();
 
-		setTimeout(() => {
-			expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
-		}, 2000);
+		expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
+		expect(marker.getElement()._leaflet_focus_handler).to.be.undefined;
+	});
+
+	it('removes focus listeners after unbinding tooltip from FeatureGroup', () => {
+
+		const marker = new Marker([51.515, -0.09]);
+		const layergroup = new FeatureGroup([marker]).addTo(map);
+
+		layergroup
+			.bindTooltip('Tooltip that will be unbinded in two seconds')
+			.openTooltip();
+
+		expect(marker.getElement()._leaflet_focus_handler).to.be.not.undefined;
+
+		layergroup.unbindTooltip();
+
+		expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
+		expect(marker.getElement()._leaflet_focus_handler).to.be.undefined;
 	});
 });
--- a/src/layer/Tooltip.js
+++ b/src/layer/Tooltip.js
@@ -387,35 +387,34 @@ Layer.include({
 
 	_addFocusListeners(remove) {
 		if (this.getElement) {
-			const el = this.getElement();
-			if (el) {
-				const onOff = remove ? 'off' : 'on';
-				DomEvent[onOff](el, 'focus', () => this._handleFocus(this), this);
-				DomEvent[onOff](el, 'blur', this._handleBlur, this);
-			}
+			this._addFocusListenersOnLayer(this, remove);
 		} else if (this.eachLayer && this._map.getRenderer(this)) {
-			this.eachLayer(layer => this._addFocusListenersOnLayer(layer), this);
+			this.eachLayer(layer => this._addFocusListenersOnLayer(layer, remove), this);
 		}
-	},
-
-	_handleFocus(source) {
-		if (this._tooltip) {
-			this._tooltip._source = source;
-			this.openTooltip();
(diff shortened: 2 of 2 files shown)


Falke-Design (changes requested):
Thank you for trying to solve this issue but this solution causes other tests to fail. Please make sure to test if all the tests are successfull. You can do this with `npm run test`

grafik

hollowM1ke:
@Falke-Design it is really strange. So I've run `npm run test` on my branch and I get the following:
error1

Then I wanted to see if I get any errors on the main branch. And I get this:
error2

Would there be a reason this happens?

Falke-Design:
Did you do a re-build? `npm run watch`

hollowM1ke:
> Did you do a re-build? `npm run watch`

I've done `npm run build` and then `npm run test`. Should I do `npm run test` and then `npm run watch`?

hollowM1ke:
> > Did you do a re-build? `npm run watch`
> 
> I've done `npm run build` and then `npm run test`. Should I do `npm run test` and then `npm run watch`?
If yes then I am not sure how, because whenever I run `npm run test` it doesn't terminate, I have to force cancel it.

Falke-Design:
no, you did it correct. First build and then test.

I will check it the next days by myself. In the mean time you can merge the latest commit from `main` into your branch

src/layer/Tooltip.js
```diff
@@ -385,22 +385,38 @@ Layer.include({
...
-			}, this);
-			DomEvent.on(el, 'blur', this.closeTooltip, this);
+		if (layer.getElement) {
+			const el = layer.getElement();
+			if (el) {
+				DomEvent.on(el, 'focus', function () {
+					this._tooltip._source = layer;
+					this.openTooltip();
+				}, this);
+				DomEvent.on(el, 'blur', this.closeTooltip, this);
```
Falke-Design, line 418: We should remove the events here too

src/layer/Tooltip.js
```diff
@@ -385,22 +385,38 @@ Layer.include({
...
 
+	_handleFocus() {
+		this._tooltip._source = this;
+		this.openTooltip();
+	},
+
+	_handleBlur() {
+		this.closeTooltip();
+	},
+
```
Falke-Design, line 409: Code exists twice: here and in `_addFocusListenersOnLayer`

Falke-Design (changes requested):
Looks better now 😄 
Can you please try to add a test which covers your changes

hollowM1ke:
Hi @Falke-Design, could I get an update?

src/layer/Tooltip.js
```diff
@@ -385,22 +385,38 @@ Layer.include({
...
+	_addFocusListenersOnLayer(layer, remove) {
+		if (layer.getElement) {
+			const el = layer.getElement();
+			if (el) {
+				const onOff = remove ? 'off' : 'on';
+				DomEvent[onOff](el, 'focus', () => this._handleFocus(layer), this);
+				DomEvent[onOff](el, 'blur', this._handleBlur, this);
+			}
 		}
```
Falke-Design, line 420: I think we should do the following (or something similar if there is a better way):

1. `_addFocusListeners` does the same for a element as `_addFocusListenersOnLayer`, so we don't duplicate the code.
2. Then the `_handleBlur` don't need to be a function anymore, because we can call `closeTooltip` directly
3. We store the `named` function / listener (`el._leaflet_focus_handler`) for the focus on the element, because an anonymous function / listener can't be removed with `off`

```
	_addFocusListeners(remove) {
		if (this.getElement) {
			this._addFocusListenersOnLayer(this, remove);
		} else if (this.eachLayer && this._map.getRenderer(this)) {
			this.eachLayer(layer => this._addFocusListenersOnLayer(layer, remove), this);
		}
	},

	_addFocusListenersOnLayer(layer, remove) {
		const el = typeof layer.getElement === 'function' && layer.getElement();
		if (el) {
			const onOff = remove ? 'off' : 'on';
			if (!remove) {
				// Remove focus listener, if already existing
				el._leaflet_focus_handler && DomEvent.off(el, 'focus', el._leaflet_focus_handler, this);

				el._leaflet_focus_handler = () => {
					if (this._tooltip) {
						this._tooltip._source = layer;
						this.openTooltip();
					}
				};
			}

			el._leaflet_focus_handler && DomEvent[onOff](el, 'focus', el._leaflet_focus_handler, this);
			DomEvent[onOff](el, 'blur', this.closeTooltip, this);

			if (remove) {
				delete el._leaflet_focus_handler;
			}
		}
	},
```

spec/suites/layer/TooltipSpec.js
```diff
@@ -578,4 +578,22 @@ describe('Tooltip', () => {
...
+		  .bindTooltip('Tooltip that will be unbinded in two seconds')
+		  .openTooltip();
+
+		marker.unbindTooltip();
+
+		setTimeout(() => {
+			expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
+		}, 2000);
+	});
 });
```
Falke-Design, line 599: Please add a test for a LayerGroup too:

```
const marker = new Marker(latlng);

			var lg = new LayerGroup([marker]).addTo(map);

			lg
				.bindTooltip('Tooltip that will be unbinded in two seconds')
				.openTooltip()
				.unbindTooltip();
```

spec/suites/layer/TooltipSpec.js
```diff
@@ -578,4 +578,22 @@ describe('Tooltip', () => {
...
+
+		marker
+		  .bindTooltip('Tooltip that will be unbinded in two seconds')
+		  .openTooltip();
+
+		marker.unbindTooltip();
+
+		setTimeout(() => {
+			expect(() => UIEventSimulator.fire('focus', marker.getElement())).to.not.throw();
+		}, 2000);
```
Falke-Design, line 597: We should not wait so long. Maximum of 500ms is allowed, better is earlier

Falke-Design (changes requested):
Sorry for letting you wait so long. 
I rewrote the functions, what do you think about it?

src/layer/Tooltip.js
```diff
@@ -385,22 +385,37 @@ Layer.include({
...
 	},
 
-	_addFocusListeners() {
+	_addFocusListeners(remove) {
 		if (this.getElement) {
-			this._addFocusListenersOnLayer(this);
-		} else if (this.eachLayer) {
-			this.eachLayer(this._addFocusListenersOnLayer, this);
+			this._addFocusListenersOnLayer(this, remove);
+		} else if (this.eachLayer && this._map.getRenderer(this)) {
```
Falke-Design, line 391:
```suggestion
```

src/layer/Tooltip.js
```diff
@@ -385,22 +385,36 @@ Layer.include({
...
 	},
 
-	_addFocusListeners() {
+	_addFocusListeners(remove) {
 		if (this.getElement) {
-			this._addFocusListenersOnLayer(this);
-		} else if (this.eachLayer) {
-			this.eachLayer(this._addFocusListenersOnLayer, this);
+			this._addFocusListenersOnLayer(this, remove);
+			this.eachLayer(layer => this._addFocusListenersOnLayer(layer, remove), this);
```
Falke-Design, line 391:
```suggestion
		} else if (this.eachLayer) {
			this.eachLayer(layer => this._addFocusListenersOnLayer(layer, remove), this);
```

Falke-Design (approved):

PR #9233, opened 2024-01-28
Change running tests, on Windows, from Chrome to Firefox

Falke-Design:
It looks like the failing tests are more a problem of Chrome then of Windows. So I changed the browser to Firefox.

Fixes: #9230
and supersedes: #9231


spec/karma.conf.js
```diff
@@ -18,7 +18,8 @@ module.exports = function (/** @type {import('karma').Config} */ config) {
 		customContextFile: 'spec/context.html',
 		customDebugFile: 'spec/debug.html',
 		files: [
-			{pattern: 'node_modules/**', included: false, served: true},
+			{pattern: 'node_modules/ui-event-simulator/**/*.js', included: false, served: true},
+			{pattern: 'node_modules/prosthetic-hand/**/*.js', included: false, served: true},
```
jonkoops, line 22: Why narrow this down? Also, doesn't `prosthetic-hand` include some assets that are not just JS?

Falke-Design: Copied from @mourner #9231

mourner:
Chrome on Windows is probably the most used browser, so it doesn't feel good to not test on it (and Firefox usage is dying out from the other side). Not sure how to proceed here... Ideally we'd try to figure out why Chrome/Win is failing on CI first.

Issue #7113, opened 2020-04-26
Tests not passing in Chrome


johnd0e:
```
Map.TouchZoom
Increases zoom when pinching out‣
Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
Decreases zoom when pinching in‣
Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```
https://github.com/Leaflet/Leaflet/blob/1ae785b73092fdb4b97e30f8789345e9f7c7c912/spec/suites/map/handler/Map.TouchZoomSpec.js#L1-L87

```
Map.Drag
touch events
change the center of the map‣
Error: Script error. (:0)
reset itself after touchend‣
Error: Script error. (:0)
```

https://github.com/Leaflet/Leaflet/blob/1ae785b73092fdb4b97e30f8789345e9f7c7c912/spec/suites/map/handler/Map.DragSpec.js#L311-L446

Problem: prosthetic-hand fails: https://github.com/Leaflet/prosthetic-hand/issues/14

I'm not sure what is best way to fix tests
- perhaps it should be done in upstream (prosthetic-hand)
- but we also could adapt tests (use `pointer` hand).
- merging https://github.com/Leaflet/Leaflet/pull/7029 would fix the issue as well.

johnd0e:
```
CRS
#zoom && #scale
convert zoom to scale and viceversa and return the same values‣
Error: expected 2.4999999999999996 to sort of equal 2.5
    at Assertion.assert (file:///C:/D/repos/Leaflet/node_modules/expect.js/index.js:96:13)
    at Assertion.eql (file:///C:/D/repos/Leaflet/node_modules/expect.js/index.js:230:10)
    at Context.<anonymous> (suites/geo/CRSSpec.js:233:31)
```
https://github.com/Leaflet/Leaflet/blob/1ae785b73092fdb4b97e30f8789345e9f7c7c912/spec/suites/geo/CRSSpec.js#L229-L235

johnd0e:
Additionally, some tests are passing in desktop Chrome, but failing in headless:

```
HeadlessChrome 81.0.4044 (Windows 10.0.0) Marker.Drag drag in CSS scaled container drags a marker with mouse, compensating for CSS scale FAILED
[...]

PR #9235, opened 2024-01-29
Fix CI tests to not depend on big Chrome window size

mourner:
Removes the workaround for this https://github.com/Leaflet/Leaflet/issues/7113#issuecomment-619528577
Should make the tests slightly faster and look nicer.


Falke-Design:
Wow such a easy solution 😄

jonkoops (commented):
Nice fix, LGTM

PR #9236, opened 2024-01-29
Remove legacy timer fallbacks for `requestAnimFrame`

mourner:
Just some leftover — we no longer support IE7-8 so it's irrelevant.

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -162,33 +162,15 @@ function getPrefixed(name) {
 	return window[`webkit${name}`] || window[`moz${name}`] || window[`ms${name}`];
 }
 
-let lastTime = 0;
+export const requestFn = window.requestAnimationFrame || getPrefixed('RequestAnimationFrame');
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):

src/core/Util.js
```diff
@@ -162,33 +162,15 @@ function getPrefixed(name) {
...
-// [`window.requestAnimationFrame`](https://developer.mozilla.org/docs/Web/API/window/requestAnimationFrame),
-// otherwise it's delayed. Returns a request ID that can be used to cancel the request.
-export function requestAnimFrame(fn, context, immediate) {
-	if (immediate && requestFn === timeoutDefer) {
-		fn.call(context);
-	} else {
-		return requestFn.call(window, fn.bind(context));
-	}
+export const requestFn = window.requestAnimationFrame || getPrefixed('RequestAnimationFrame');
+export const cancelFn = window.cancelAnimationFrame || getPrefixed('CancelAnimationFrame') || getPrefixed('CancelRequestAnimationFrame');
```
jonkoops, line 166: We could probably just get rid of using the prefixed versions here, RAF is widely supported.

PR #9252, opened 2024-02-08
fix: support dashOffset for canvas

g45t345rt:
I've noticed there's a pending pull request for this issue, but it has been opened for a while.
https://github.com/Leaflet/Leaflet/pull/7867

I've added a debug map to test it out.


Falke-Design:
Thank you for your contribution but we will stick to the PR #7867

PR #9270, opened 2024-02-20
Adding new plugin leaflet.osmdatapicker to the list of plugins

amrHH:
✨ I finished a first version of a new plugin named leaflet.osmdatapicker and i wish for it to be added in the list of all the plugins. Thanks.


Falke-Design (changes requested):
Interessting plugin but we will not merge a plugin which violates the terms of OpenStreetMap. Please make sure that after loading the data you adding the OpenStreetMap attribution.

`map.attributionControl.addAttribution` https://leafletjs.com/reference.html#map-attributioncontrol

Falke-Design (approved):

PR #9271, opened 2024-02-21
9259 disable text selection

### LINKED ISSUE #9259: L.DomUtil.disableTextSelection should only disable text selection on the map container
### Checklist

### Steps to reproduce

### Expected behavior

### Current behavior

_No response_

### Environment

- Leaflet version: 
- Browser (with version): 
- OS/Platform (with version):

ventralnet:
This fixes #9259 for your review

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -84,28 +84,30 @@ const userSelectProp = ['userSelect', 'WebkitUserSelect'].find(prop => prop in d
 let prevUserSelect;
 
 // @function disableTextSelection()
-// Prevents the user from selecting text in the document. Used internally
-// by Leaflet to override the behaviour of any click-and-drag interaction on
-// the map. Affects drag interactions on the whole document.
-export function disableTextSelection() {
-	const value = documentStyle[userSelectProp];
+// Prevents the user from selecting text in the document or element passed.
+// Used internally by Leaflet to override the behaviour of any click-and-drag
+// interaction on the map.
+export function disableTextSelection(el) {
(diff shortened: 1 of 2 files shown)


ventralnet:
@IvanSanchez I didn't make those changes because of @mourner response on the original ticket.  Supposedly drag operations were performed on the entire document for reasons.  I didn't want to effect this with my changes.

src/dom/DomUtil.js
```diff
@@ -83,29 +83,32 @@ const documentStyle = document.documentElement.style;
...
+	targetStyle[userSelectProp] = 'none';
 }
 
-// @function enableTextSelection()
+// @function enableTextSelection(el: HTMLElement?)
 // Cancels the effects of a previous [`L.DomUtil.disableTextSelection`](#domutil-disabletextselection).
-export function enableTextSelection() {
+export function enableTextSelection(el) {
+	const targetStyle = el ? el.style : documentStyle;
 	if (typeof prevUserSelect === 'undefined') {
```
Falke-Design, line 107: This will cause problems, if `disableTextSelection(document)` and then `disableTextSelection(element)` is called. Try to use a `WeakMap`

Falke-Design (changes requested):

Falke-Design:
> This implementation is incomplete - see the calls to disable/enableTextSelection() in src/dom/Draggable.js and src/map/handler/Map.BoxZoom.js. Those calls should pass a reference to the map container (this._map.getContainer() IIRC).

>Supposedly drag operations were performed on the entire document for reasons. I didn't want to effect this with my changes

That is fine, let us keep the current behavior.

PR #9286, opened 2024-03-03
Add support for curved polylines/polygons using Bézier curves

serso:


src/layer/vector/Polyline.js
```diff
@@ -57,7 +57,11 @@ export const Polyline = Path.extend({
 
 		// @option noClip: Boolean = false
 		// Disable polyline clipping.
-		noClip: false
+		noClip: false,
+
+		// @option curved: Boolean = false
+		// Enable polyline Bézier curves. Curved polylines don't support clipping - each line is rendered fully.
```
IvanSanchez, line 63: This needs to reflect that bezier curves are not supported when polylines are drawn on an `L.Canvas`.

src/layer/vector/Canvas.js
```diff
@@ -284,13 +285,34 @@ export const Canvas = Renderer.extend({
...
-				p = parts[i][j];
-				ctx[j ? 'lineTo' : 'moveTo'](p.x, p.y);
+		if (layer.options.curved) {
+			for (let i = 0; i < parts.length; i++) {
+				const points = parts[i];
+
+				const cmds = curvedPathCommands(points, closed);
+				for (j = 0; j < cmds.length; j++) {
+					const cmd = cmds[j];
+					switch (cmd.type) {
```
IvanSanchez, line 295: I can't see the point of holding "command" objects. The non-bezier code can differentiate between `lineTo` and `moveTo` by `j===0`, so there is no need to do an initial `array.push` of a `M` "command" and then push `C` commands.

IvanSanchez (changes requested):

IvanSanchez:
Besides, there's plugins like https://github.com/elfalem/Leaflet.curve , so I must ask the question of what does this need to be in the Leaflet core.

src/layer/vector/Polyline.js
serso: But that's supported. See changes in `src/layer/vector/Canvas.js`. SVG drawing commands are transformed into Canvas drawing commands in Canvas.js. I have also added vector-canvas-curved.html to test that.

serso:
> so I must ask the question of what does this need to be in the Leaflet core.

My thinking was that smooth drawing (using curves and not just connecting the dots) of polylines/polygons is an essential feature and if the Leaflet core supports polylines/polygons it should also be able to draw them smoothly.

> there's plugins like https://github.com/elfalem/Leaflet.curve 

This particular plugin doesn't do what I want - it draws curves via SVG commands passed explicitly. I want to draw curved polygons, e.g. have an ordered set of dots and draw curves between smoothly (and automatically). It should be possible though  to transform the polygons into SVG commands by hand and pass the result to Leaftlet.curve.

I leave it up to the maintainers to decide whether this should be a core feature of the lib or it should be implemented externally.

src/layer/vector/Canvas.js
serso: I don't understand what you mean.
Do you want me to replace `curvedPathCommands` with in-loop calculation of Bezier curve parameters?
It would be more efficient but then the code in the loop would become more complicated.
Also, for polygons (or polylines with `closed=true`) `curvedPathCommands` modifies the first `C` command at the end of the loop to make the joint smooth. It would not be possible if `curvedPathCommands` was extracted here.

Falke-Design:
Hi @serso,

that is a really nice idea and it is much easier to use then Leaflet.curve, but I think the best would to make a plugin out of it. If you need help with creating the Plugin, just let me know. (If you don't want to create a plugin, I would be happy to create it with your code if that's ok with you)

Issue #9291, opened 2024-03-09
leafletjs.com welcome banner dead URLs


horseDeveloper:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

The first time you visit leafletjs.com (or just clear cookies and site data + refresh page).
The 2nd to last line reads: 
"It is chilling to see Leaflet being used for [documenting Russia’s war crimes](https://ukraine.bellingcat.com/), [factual reporting of the war](https://liveuamap.com/) and for coordination of humanitarian efforts [in Romania](https://refugees.ro/) and [in Poland](https://dopomoha.pl/). We commend these uses of Leaflet." 
The last two URLs are dead. 

### Expected behavior

- Remove the dead URLs?
- Separate the last 2 URLs and put them under "archived projects"?
- Put archive.org URLs?

### Current behavior

See above.

### Environment

- Leaflet version: 
- Browser (with version): Firefox 123.0.1 (64-bit)
- OS/Platform (with version): Windows 10

Falke-Design:
Thank you for reporting.

@mourner do you have some other links that you want to add instead?

mourner:
Maybe just replace "[in Romania](https://refugees.ro/) and [in Poland](https://dopomoha.pl/)" with "across Europe" without a link. We can add something back later. Not sure what happened to those two websites.

Malvoz:
Fixed in https://github.com/Leaflet/Leaflet/pull/9305.

PR #9305, opened 2024-03-28
remove dead links

Hubda987:
Removes dead links as specified in https://github.com/Leaflet/Leaflet/issues/9291


mourner (changes requested):
Also needs similar change in https://github.com/Leaflet/Leaflet/blob/main/docs/dialog/index.md

Hubda987:
> Also needs similar change in https://github.com/Leaflet/Leaflet/blob/main/docs/dialog/index.md

Thanks, I changed the index.md file as well.

mourner (approved):

PR #9316, opened 2024-04-06
Patch Leaflet to be Server-Side-Render friendly (SSR)

matthova:
This Pull request adds the eslint plugin `ssr-friendly` and implements changes to meet the new linting requirements.

I've confirmed I'm able to use the forked library with a SSR Remix app.

### DIFF
--- a/docs/docs/js/docs.js
+++ b/docs/docs/js/docs.js
@@ -2,92 +2,85 @@
 hljs.configure({tabReplace: '    '});
 hljs.initHighlighting();
 
-function setup() {
-	if (typeof document === 'undefined') {
-		return;
-	}
+const tocCopy = document.createElement('div');
+tocCopy.id = 'toc-copy';
 
-	const tocCopy = document.createElement('div');
-	tocCopy.id = 'toc-copy';
+const toc = document.querySelector('#toc');
 
-	const toc = document.querySelector('#toc');
-
-	if (toc) {
+if (toc) {
 	// top menu
-		let menus = document.querySelectorAll('#toc a');
-		let i;
+	let menus = document.querySelectorAll('#toc a');
+	let i;
 
-		for (i = 0; i < menus.length; i++) {
-			menus[i].addEventListener('click', (e) => {
-				clickOnAnchor(e);
-			});
-		}
-
-		// sidebar menu
-		tocCopy.innerHTML = toc.innerHTML;
-		document.getElementsByClassName('container')[0].appendChild(tocCopy);
-
-		menus = document.querySelectorAll('#toc-copy ul');
-		i = 0;
-
-		for (i = 0; i < menus.length; i++) {
-			menus[i].addEventListener('mouseover', function () {
-				this.previousElementSibling.classList.add('hover');
-			});
-
-			menus[i].addEventListener('mouseout', function () {
-				this.previousElementSibling.classList.remove('hover');
-			});
-
-			menus[i].addEventListener('click', (e) => {
-				clickOnAnchor(e);
-			});
-		}
-
-		const labels = document.querySelectorAll('#toc-copy h4');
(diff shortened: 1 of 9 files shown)


docs/docs/js/docs.js
```diff
@@ -2,85 +2,92 @@
 hljs.configure({tabReplace: '    '});
 hljs.initHighlighting();
 
-const tocCopy = document.createElement('div');
-tocCopy.id = 'toc-copy';
+function setup() {
```
IvanSanchez, line 5: Anything in `docs/` will only ever run in a browser. You made the blanket assumption that all `.js` files are part of the Leaflet code, but some are part of its website. These changes are therefore too broad.

src/core/Browser.js
```diff
@@ -13,54 +13,71 @@
...
-	safari,
-	mobile,
-	pointer,
-	touch,
-	touchNative,
-	retina,
-	mac,
-	linux
-};
+export default setupBrowser();
```
IvanSanchez, line 83: This breaks tree-shaking (which, from my PoV, is more important than SSR friendlyness)

src/core/Browser.js
```diff
@@ -13,54 +13,71 @@
...
-const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
+	// @property mobile: Boolean; `true` for all browsers running in a mobile device.
+	const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
 
-// @property pointer: Boolean
-// `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
-const pointer = !!window.PointerEvent;
+	// @property pointer: Boolean
+	// `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
+	const pointer = !!window.PointerEvent;
```
IvanSanchez, line 41: You may use something like `!!(window?.PointerEvent)`. Ditto for `navigator` and the like.

IvanSanchez (changes requested):

IvanSanchez:
And do read https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining .

src/core/Browser.js
matthova: Good callout. I'll see if there's a way to have both

matthova:
@IvanSanchez thank you for the feedback! I love the idea of enforcing ssr-compatibility via the lint plugin I added, however, I wonder if there's a slightly different approach we could take for this library to remove unnecessary function-wrapping. Let me do some digging

jonkoops:
> I've confirmed I'm able to use the forked library with a SSR Remix app.

I am curious about this. Why would Leaflet need to be 'SSR friendly' in order for this to work? Leaflet was always intended to be a complete client-side solution, so why even use it in an SSR context? Can Remix not opt-out of SSR for the few places one might need to include Leaflet?

matthova:
> > I've confirmed I'm able to use the forked library with a SSR Remix app.
> 
> I am curious about this. Why would Leaflet need to be 'SSR friendly' in order for this to work? Leaflet was always intended to be a complete client-side solution, so why even use it in an SSR context? Can Remix not opt-out of SSR for the few places one might need to include Leaflet?

That's a great question. I think the maintainers of the library need to own this decision. My use case, as a consumer, is that I love being able to install a package and not have to think about it or do any additional configuration. The tradeoff may not be worth it, for maintainers.

matthova:
Closing and will submit an alternate approach with optional chaining

matthova:
Alternate PR: https://github.com/Leaflet/Leaflet/pull/9317

Issue #9377, opened 2024-06-21
Trackpad pinch zoom is too sensitive


diachedelic:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

1. Open any Leaflet map on a device with a trackpad, such as a laptop.
2. Zoom the map using a pinch gesture.
3. Without removing your fingers from the trackpad, very slowly move your fingers apart or together.

### Expected behavior

The pinch gesture should change the map's zoom level in a gradual and controlled manner, as it does on a touch device.

### Current behavior

Each tiny movement of the fingers changes the zoom level by at least 1, leading to disorientatingly fast changes to the map's bounds. With almost no movement of the fingers, it is possible to take the map from the minimum to the maximum zoom.

### Minimal example reproducing the issue

https://jsfiddle.net/user2314737/tvwkgo34/

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Chrome v126, Safari v17.5 (not Firefox, which does not seem to support [pinch to zoom](https://kenneth.io/post/detecting-multi-touch-trackpad-gestures-in-javascript) on a Mac)
- OS/Platform (with version): MacOS 14.5

Note that these are only the browsers I have tested with. I believe the problem is present on any device with a trackpad.

diachedelic:
I have fixed this bug by reworking the `ScrollWheelZoom` class. If there is a chance it would be accepted, I can submit a PR.

diachedelic:
Here is my patch if anyone is desperate: https://gist.github.com/diachedelic/633712a4c3fc617e9f5e94bdf33c202b.

Falke-Design:
Hi @diachedelic interesting approach, have you tested your changes with a mouse wheel too? Are there any significant differences?

It would be nice to have an PR, but I can't say if we will merge it but we are always happy to see different approaches for the wheel zooming.

IvanSanchez:
[...]

PR #9378, opened 2024-06-21
Add maxWheelZoomSpeed option (#9377)

diachedelic:
This pull request replaces the `wheelDebounceTime` option with the more useful `maxWheelZoomSpeed` option. It also improves support for zooming via trackpad pinch gestures. It fixes issue #9377.

Further testing on a variety of devices, browsers, and operating systems would be appreciated.


Falke-Design:
I think for the trackpad it is a much better solution but not for the mouse. So we could implement it only for the trackpad but how to detect? The `ctrlKey` could be also true, when the user presses the key on the keyboard. Would this be a problem?

Tested on Windows 10, Firefox

diachedelic:
How did you find the experience using a mouse wheel?

PR #9385, opened 2024-07-02
 Support SSR (Reopen #9317)

Falke-Design:
I accidentally closed the other PR #9317 ... Look for more details into it

I added the changes from @jonkoops and made it work with Deno.

@Gugustinette can you please test this version?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,10 +110,9 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ env.NODE_VERSION }}
-          check-latest: true
           cache: npm
 
-      - name: Run Node.js ssr script
+      - name: Run Node.js SSR script
         run: node ./spec/ssr/ssr_node.mjs
 
       - name: Set up Deno
@@ -121,7 +120,7 @@ jobs:
         with:
           deno-version: v1.x
 
-      - name: Run Deno ssr script
+      - name: Run Deno SSR script
         run: deno run ./spec/ssr/ssr_deno.js
 
   test:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -100,8 +100,21 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Restore setup
+        uses: actions/cache@v4
+        with:
+          path: ./*
+          key: ${{ runner.os }}-${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          check-latest: true
+          cache: npm
+
       - name: Run Node.js ssr script
-        run: node spec/ssr/ssr_node.mjs
+        run: node ./spec/ssr/ssr_node.mjs
 
       - name: Set up Deno
         uses: denoland/setup-deno@v1
@@ -109,7 +122,7 @@ jobs:
           deno-version: v1.x
 
       - name: Run Deno ssr script
-        run: deno run spec/ssr/ssr_deno.js
+        run: deno run ./spec/ssr/ssr_deno.js
 
   test:
     needs: setup
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -95,6 +95,22 @@ jobs:
         env:
           CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
+  test-ssr:
+    needs: setup
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Run Node.js ssr script
+        run: node spec/ssr/ssr_node.mjs
+
+      - name: Set up Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: v1.x
+
+      - name: Run Deno ssr script
+        run: deno run spec/ssr/ssr_deno.js
+
   test:
     needs: setup
     runs-on: ${{ matrix.os || 'ubuntu-latest' }}
--- a/spec/ssr/ssr_node.mjs
+++ b/spec/ssr/ssr_node.mjs
@@ -1,0 +1,2 @@
+import L from '../../dist/leaflet-src.js';
+console.log(L.version);
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -40,16 +40,16 @@ const touch = touchNative || pointer;
 
 // @property retina: Boolean
 // `true` for browsers on a high-resolution "retina" screen or on any screen when browser's display zoom is more than 100%.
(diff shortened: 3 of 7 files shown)


Gugustinette:
I'll look into that asap !

mourner (approved):

Gugustinette:
Just tested the branch in [Nuxt Leaflet](https://github.com/nuxt-modules/leaflet) playground, and it works like a charm !

Here is the example code :
```vue
<template>
  <div style="height:100vh; width:100vw">
    <h1>Basic Map Exemple</h1>
    <LMap
      ref="map"
      :zoom="6"
      :max-zoom="18"
      :center="[47.21322, -1.559482]"
      :use-global-leaflet="false"
      @ready="onMapReady"
    >
      <LTileLayer
        url="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
        attribution="&amp;copy; <a href=&quot;https://www.openstreetmap.org/&quot;>OpenStreetMap</a> contributors"
        layer-type="base"
        name="OpenStreetMap"
      />
    </LMap>
  </div>
</template>

<script setup lang="ts">
import L from 'leaflet'
import { ref } from 'vue'

const map = ref(null)

// When the map is ready
const onMapReady = () => {
  console.log('Map is ready')
  console.log(L)
  console.log(map.value.maxZoom)
}
</script>
```

Before, an error was thrown as SSR environment will typically load (and optimize) imports on the server-side :

Capture d’écran 2024-07-03 à 13 36 32

After, the same code now works correctly :

Capture d’écran 2024-07-03 à 13 33 41

Leaflet instance can now be loaded in a node server context.
Huge step forward for SSR environments 👍

Falke-Design:
@mourner @jonkoops I tried to implement a test in the workflow. Can you please check it?

.github/workflows/main.yml
```diff
@@ -95,6 +95,22 @@ jobs:
         env:
           CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
+  test-ssr:
+    needs: setup
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Run Node.js ssr script
```
jonkoops, line 103: We should be setting up Node.js first, otherwise the system version on the GitHub Action runner will be used.
```suggestion
      - name: Set up Node
        uses: actions/setup-node@v4
        with:
          node-version: ${{ env.NODE_VERSION }}

      - name: Run Node.js ssr script
```

.github/workflows/main.yml
```diff
@@ -95,6 +95,22 @@ jobs:
         env:
           CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
+  test-ssr:
+    needs: setup
+    runs-on: ubuntu-latest
+
+    steps:
```
jonkoops, line 102: You'll have to restore the setup as well.
```suggestion
    steps:
      - name: Restore setup
        uses: actions/cache@v4
        with:
          path: ./*
          key: ${{ runner.os }}-${{ github.ref }}-${{ github.sha }}-setup

```

jonkoops (changes requested):

.github/workflows/main.yml
```diff
@@ -95,6 +95,35 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: ./*
+          key: ${{ runner.os }}-${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          check-latest: true
```
jonkoops, line 113: We can skip the latest check here, as the `setup` task already does this for us.
```suggestion
```

.github/workflows/main.yml
```diff
@@ -95,6 +95,35 @@ jobs:
...
+          cache: npm
+
+      - name: Run Node.js ssr script
```
jonkoops, line 116:
```suggestion
      - name: Run Node.js SSR script
```

.github/workflows/main.yml
```diff
@@ -95,6 +95,35 @@ jobs:
...
+        run: node ./spec/ssr/ssr_node.mjs
+
+      - name: Set up Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: v1.x
+
+      - name: Run Deno ssr script
```
jonkoops, line 124:
```suggestion
      - name: Run Deno SSR script
```

spec/ssr/ssr_node.mjs
```diff
@@ -0,0 +1,2 @@
+import L from '../../dist/leaflet-src.js';
```
jonkoops, line 1: Any reason the import here is to `leaflet-src.js` and the Deno one to `leaflet-src.esm.js`? I imagine we'd want both of these using the ESM version.

src/core/Browser.js
```diff
@@ -40,19 +40,21 @@ const touch = touchNative || pointer;
...
-const retina = (window.devicePixelRatio || (window.screen.deviceXDPI / window.screen.logicalXDPI)) > 1;
+const retina = typeof window === 'undefined' || typeof window.devicePixelRatio === 'undefined' ? false : window.devicePixelRatio > 1;
 
 // @property mac: Boolean; `true` when the browser is running in a Mac platform
-const mac = navigator.platform.startsWith('Mac');
+const mac = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Mac');
 
 // @property mac: Boolean; `true` when the browser is running in a Linux platform
-const linux = navigator.platform.startsWith('Linux');
+const linux = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Linux');
```
jonkoops, line 49: Not sure if this can be handled by the SSR framworks, but this could be a bit more terse using nullish coalescing. 
```suggestion
const linux = typeof navigator?.plaform === 'undefined' ? false : navigator.platform.startsWith('Linux');

```

jonkoops (approved):
Couple of small optional comments, but this is good to merge AFAIK

spec/ssr/ssr_node.mjs
Falke-Design: This was the only combination that worked.

When I try to use leaflet-src.js in deno, I get following errors:
grafik

And for node esm doesn't work:
grafik

src/core/Browser.js
Falke-Design: Node can't handle that:
grafik

spec/ssr/ssr_node.mjs
jonkoops: Makes sense, let's leave it as-is

src/core/Browser.js
jonkoops: Figured as much, let's keep it like this then

jonkoops (approved):

Teebo:
Hi, when should we expect this to be released?
v1.9.4 is still the latest in npm and it does not include this fix.
Thanks

PR #9390, opened 2024-07-08
Allmaps Leaflet plugin

mclaeysb:
This PR is adding the Allmaps Leaflet plugin to the plugins page.

Hope this matches the required format.

Thanks to the Leaflet community for building this great tool!

### DIFF
--- a/docs/_plugins/basemap-formats/allmaps-leaflet.md
+++ b/docs/_plugins/basemap-formats/allmaps-leaflet.md
@@ -1,0 +1,12 @@
+---
+name: allmaps-leaflet
+category: basemap-formats
+repo: https://github.com/allmaps/allmaps/tree/main/packages/leaflet
+author: Allmaps
+author-url: https://www.allmaps.org
+demo: https://observablehq.com/@allmaps/leaflet-plugin
+compatible-v0:
+compatible-v1: true
+---
+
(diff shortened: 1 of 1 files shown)


docs/_plugins/basemap-formats/allmaps-leaflet.md
```diff
@@ -0,0 +1,12 @@
+---
+name: allmaps-leaflet
+category: basemap-formats
```
Falke-Design, line 3: Please use the category "Non-map base layers"

Falke-Design (changes requested):

docs/_plugins/basemap-formats/allmaps-leaflet.md
mclaeysb: I consider this, but I think it's better placed in basemap-formats. Plugins in the 'Non-map base layers' like 'Leaflet-IIIF' show a (IIIF) image in ad-hoc pixel coordinates, whereas this plugin actually transforms/warps the original image to show it on the map in the right geographic location. In this way it's closer to the various W(T)MS plugins in the basemap-formats folder. See this [example](https://observablehq.com/@allmaps/leaflet-plugin). Let me know if you think the same.

Falke-Design (approved):

Issue #8659, opened 2022-11-15
Accessibility box shown on click on Google Chrome


FloChehab:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

Followup to https://github.com/Leaflet/Leaflet/pull/8247

1. Go to https://plnkr.co/edit/VQ549soZtte7FRFT?preview (on Google Chrome, this issue doesn't occur on Firefox)
2. Click on the polygon with a Tooltip
3. The accessibility box appears as in the screenshot.

![image](https://user-images.githubusercontent.com/20242665/201962304-db67d123-97da-4158-9615-905230c87c5b.png)

This is particulary inadequate when custom style is applied on click such as this : 
![Screenshot from 2022-11-15 15-50-53](https://user-images.githubusercontent.com/20242665/201963499-3ea1ef80-24fb-4fd1-abc0-36121cc93a8d.png)

### Expected behavior

No accessibility box is show (or it can be disabled by a flag).

### Current behavior

An accessibility box is show on click, whether the user wants it or not, and without any configuration from the Leaflet user.

### Minimal example reproducing the issue

https://plnkr.co/edit/VQ549soZtte7FRFT?preview

### Environment

- Leaflet version: 1.9.2
- Browser (with version): Google Chrome (107.0.5304.87) (this issue doesn't occur on Firefox)
- OS/Platform (with version): tested on Fedora 26 & latest Mac OSX

FloChehab:
Hello @Malvoz,

Here is the issue, a bit quicker than expected, I thought it would be harder to have a reproducible setup.

chrisbitmead:
Users of our app  ( http://anbg.gov.au/fap ) are suddenly reporting the same thing, and I agree it occurs on Chrome but not Firefox. It also occurs in Edge, I don't know if that's because Edge shares the Chrome rendering engine.

Falke-Design:
[...]

PR #9393, opened 2024-07-09
Prevent showing outline-box on Chromium when clicking on a vector with a tooltip

Falke-Design:
Fixes: #8659

grafik

I tried for an hour to add a test but I wasn't successfull ...

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -223,6 +223,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-container a {
 	color: #0078A8;
 	}
+/* prevent showing outline-box on Chromium when clicking on a vector with a tooltip */
+.leaflet-container path.leaflet-interactive:focus:not(:focus-visible) {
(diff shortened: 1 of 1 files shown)


dist/leaflet.css
```diff
@@ -223,6 +223,10 @@ svg.leaflet-image-layer.leaflet-interactive path {
 .leaflet-container a {
 	color: #0078A8;
 	}
+/* prevent showing outline-box on Chromium when clicking on a vector with a tooltip */
+.leaflet-container path.leaflet-interactive:focus:not(:focus-visible) {
```
Malvoz, line 227: I think we can omit `.leaflet-container `, the remaining selector is specific enough.

Malvoz (approved):

dist/leaflet.css
Falke-Design, line 227:
```suggestion
path.leaflet-interactive:focus:not(:focus-visible) {

```

PR #9410, opened 2024-07-26
Upgrade to ESLint 9+ and flat config

mourner:
Migrated all the rules to the new config format and it seems to be working with a few minor fixes. The only trouble I had was moving from `eslint-plugin-import` to `eslint-plugin-import-x` because the former seems abandoned and doesn't work with ESLint 9+.


mourner:
Not sure why this one unit test fails in all browsers, seems unrelated to the changes: 

```
Canvas #events should not fire click when dragging the map on top of it FAILED
```

Issue #9067, opened 2023-08-20
Inverting x axis of L.CRS.Simple causes L.Circle to have no radius


Tobikblom:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

Given the following CRS that extends L.CRS.Simple and inverts the x axis by inverting the value of the a coefficient in the transformation: `L.Util.extend(L.CRS.Simple, { transformation: new L.Transformation(-1, 0, -1, 0) });`
1. Create a L.Map instance with the given CRS and some specifed zoom and center.
2. Create and add a L.Circle instance with some specified radius to the map.

### Expected behavior

A circle with the specified radius is shown.

### Current behavior

A circle without any radius is shown.

### Minimal example reproducing the issue

https://plnkr.co/edit/gRcpLsOi2zBXPjn7

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Microsoft Edge Version 115.0.1901.203 (Official build) (64-bit)
- OS/Platform (with version): Windows 11 22H2

Falke-Design:
I don't know if we should fix this and how we can fix this.
If we invert the radius too, then everythig would work fine again:

`L.circle([0,0], { radius: -10 }).addTo(map);` or
`const latlng2 = crs.unproject(crs.project(this._latlng).subtract([-this._mRadius, 0])); ` 

https://github.com/Leaflet/Leaflet/blob/b6a5aeaf81d63efe83f9f022267b0b7600518841/src/layer/vector/Circle.js#L94

@mourner @IvanSanchez please share your thoughts.

bakp22:
I'm working on this.

cipherprofessor:
This looks fixed — checked current `main` (`125bda1`). `src/layer/vector/Circle.js:80` now wraps the radius calc in `Math.abs()`:

```js
this._pxRadius = Math.abs(this._point.x - map.latLngToLayerPoint(latlng2).x);
```

[...]

PR #9414, opened 2024-07-26
Solution to issue 9067 - Inverting x axis of L.CRS.Simple causes L.Circle to have no radius

bakp22:
solves #9067

## Solution to issue 9067 - Inverting x axis of L.CRS.Simple causes L.Circle to have no radius
**Identified cause**: In Circle.js, the radius was being returned as negative once it was altered through different modifications towards itself presumably because of the transformation method requirements and the project method.

**Solution**: Math.abs() surrounded the calculation of the returned radius making it positive instead of negative. “npm run test” was performed to see if this change made additional errors. All tests were successfull.

### DIFF
--- a/spec/suites/layer/vector/CircleSpec.js
+++ b/spec/suites/layer/vector/CircleSpec.js
@@ -1,5 +1,5 @@
 import {expect} from 'chai';
-import {Circle, Map} from 'leaflet';
+import {Circle, Map, Util, CRS, Transformation} from 'leaflet';
 import {createContainer, removeMapContainer} from '../../SpecHelper.js';
 
 describe('Circle', () => {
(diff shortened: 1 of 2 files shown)


IvanSanchez (approved):

Falke-Design (changes requested):
Can you please add a new test for this special case

bakp22:
We will work on this right now. Should we use an existing test suite or make one for this specific case?

bakp22:
We have been struggling to figure out how to get the circle to fully render inside our tests. The way we are creating a circle isn't calling the _project function when it is being added to a map in our test file. Could you please provide some assistance? 
Also, this case passes however continues to pass after making changes to the source code to intentionally cause a failure.

Current Code:
```js
		function createCircleWithTransformation(crs){
			container = container = createContainer();
			map = new Map(container, {crs});
			console.log(crs.distance);
			return new Circle([50,30], {radius: 10, crs: crs}).addTo(map)
		}

		

		// TEST FOR DEFAULT TRANSFORMATION
		it('returns a positive radius after default transformation', () => {
			let crs = L.Util.extend(L.CRS.Simple, {
				transformation: new L.Transformation(1, 0, 1, 0),
			});
			const circle = createCircleWithTransformation(crs)
			expect(circle.getRadius()).to.be.greaterThan(0);
			expect(circle.getRadius()).to.equal(10);
			
		});

Falke-Design (approved):

Falke-Design:
Thank you for your contirbution

PR #9437, opened 2024-08-12
Update README.md

cherylhughey:
Simple punctation and sentence flow cleanup.


Falke-Design:
Thank you for your contribution but we will keep the Readme like it is

PR #9440, opened 2024-08-17
New plugin - basemap switch control

Urban96:
New plugin featuring a basemap control with an integrated synchronized miniature map

### DIFF
--- a/docs/_plugins/layer-switching-controls/l-basemapcontrol.md
+++ b/docs/_plugins/layer-switching-controls/l-basemapcontrol.md
@@ -1,0 +1,12 @@
+---
+name: L.basemapControl
+category: layer-switching-controls
+repo: https://github.com/Urban96/L.basemapControl
+author: Daniel Urban
+author-url: https://github.com/Urban96
+demo: https://urban96.github.io/L.basemapControl/examples/example1.html
+compatible-v0:
+compatible-v1: true
+---
+
+Basemap Control with Synchronized Minimap
--- a/docs/_plugins/layer-switching-controls/l-basemapcontrol.md
+++ b/docs/_plugins/layer-switching-controls/l-basemapcontrol.md
@@ -9,4 +9,4 @@ compatible-v0:
 compatible-v1: true
 ---
 
-Basemap Control with Synchronized Minimap
+Leaflet plugin that adds a control for switching between multiple basemap layers. It features a synchronized minimap preview to show the next layer in the sequence.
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Please make the description more clear what the plugin does. 
_... Clicking on the minimap changes the base layer ..._

Also I strongly suggest you to fix following:
1. On zoom level 0 the minimap doesn't work anymore
2. Add [disableClickPorpagation](https://leafletjs.com/reference.html#domevent-disableclickpropagation) to the minimap. This prevents moving the map when the user double clicks on the minimap
3. Use the latest Leaflet version (1.9.4) in your example

Urban96:
Hello,

Thank you for the feedback! I’ve made the necessary updates: clarified the description, fixed the zoom level 0 issue, added disableClickPropagation, and updated to Leaflet 1.9.4.

docs/_plugins/layer-switching-controls/l-basemapcontrol.md
```diff
@@ -0,0 +1,12 @@
...
+category: layer-switching-controls
+repo: https://github.com/Urban96/L.basemapControl
+author: Daniel Urban
+author-url: https://github.com/Urban96
+demo: https://urban96.github.io/L.basemapControl/examples/example1.html
+compatible-v0:
+compatible-v1: true
+---
+
+Basemap Control with Synchronized Minimap
```
Falke-Design, line 12: Please change this describtion

Falke-Design (changes requested):

docs/_plugins/layer-switching-controls/l-basemapcontrol.md
Urban96: Done. I hope this description better captures what the plugin does.

docs/_plugins/layer-switching-controls/l-basemapcontrol.md
```diff
@@ -0,0 +1,12 @@
...
+category: layer-switching-controls
+repo: https://github.com/Urban96/L.basemapControl
+author: Daniel Urban
+author-url: https://github.com/Urban96
+demo: https://urban96.github.io/L.basemapControl/examples/example1.html
+compatible-v0:
+compatible-v1: true
+---
+
+Leaflet plugin that adds a control for switching between multiple basemap layers. It features a synchronized minimap preview to show the next layer in the sequence.
```
Falke-Design, line 12: @Urban96 I would suggest something like this, because for me it was at the beginning not clear that I need to click on the minimap to switch.
But if you want to keep your description, it is fine for me and I will merge it.

```suggestion
Leaflet plugin that creates a synchronized minimap preview to switch betweeen multiple basemap layers. With a click on the minimap the next basemap layer in the sequence is applied.
```

Urban96: Sure, it's updated! Thanks

PR #9475, opened 2024-09-19
Allow to set `interactive` from Polygon.setStyle

yohanboniface:
Instead of only at init.

### DIFF
--- a/src/layer/vector/SVG.js
+++ b/src/layer/vector/SVG.js
@@ -88,10 +88,6 @@ export const SVG = Renderer.extend({
 			path.classList.add(...splitWords(layer.options.className));
 		}
 
-		if (layer.options.interactive) {
-			path.classList.add('leaflet-interactive');
-		}
-
 		this._updateStyle(layer);
 		this._layers[stamp(layer)] = layer;
 	},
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):
Looks valid for me

src/layer/vector/SVG.js
```diff
@@ -148,6 +144,12 @@ export const SVG = Renderer.extend({
 		} else {
 			path.setAttribute('fill', 'none');
 		}
+
+		if (options.interactive) {
+			path.classList.add('leaflet-interactive');
```
jonkoops, line 149: Should we not also be calling `addInteractiveTarget()` and `removeInteractiveTarget()` here like we do in `DivOverlay`?

yohanboniface: Makes sense! Pushed a change :)

jonkoops (approved):
Looking good to me!

jonkoops:
Looks like we have some failing tests, not sure what is going on, but it should be resolved before we merge this in.

yohanboniface:
Can't make them fail locally :/

Could try on your side, just in case ?

Issue #9228, opened 2024-01-24
Requiring OSM attribution at runtime


opk12:
### Motivation

The [Quick Start](https://leafletjs.com/examples/quick-start/) is explicit that the OpenStreetMap attribution is mandatory. In the common case that OSM's official tile server is used, the lib can check at runtime to provide an attribution, to benefit the developer and OSM.

### Suggested solution

What about providing a default attribution, or giving an error, if the domain is any of `tile.openstreetmap.org` or `tile.osm.org` and the attribution is absent?

### Additional context

I am an OSM contributor and hear cases of missing attribution with the OSM official tile server every now and then; so many that [a repo](https://github.com/openstreetmap/tile-attribution) was recently set up to keep track of them, and this includes cases where Leaflet is used.

mourner:
Actually, providing a default OSM attribution when none is provided and `tile.openstreetmap.org` is used sounds like a neat idea. It's a common pitfall and this seems rather harmless — people who need the attribution to be empty for whatever reason (e.g. attribution is on a different visible part of the page) can do so explicitly by passing `attribution: ''`. This can be considered a breaking change though. @jonkoops @Falke-Design @IvanSanchez what do you guys think?

IvanSanchez:
@mourner Given the sheer amount of Leaflet users that (a) are newbies and (b) use OSM tiles, this seems like a nice idea indeed. The cost in terms of codebase size is negligible IMO.

(There is an alternative, which I can call "The OpenLayers Way": provide a `L.TileLayer.OSM`, a la leaflet-providers. But I think it goes against Leaflet tradition of tile URL templates all around)

Disclaimer: I have been in the OSMF (loooong ago), so I'm understandably biased in favour of OSM.

matkoniecz:
>  tile.openstreetmap.org or tile.osm.org

likely the same for a.tile.openstreetmap.org / b.tile.openstreetmap.org / c.tile.openstreetmap.org deprecated addressing https://github.com/openstreetmap/operations/issues/737

jonkoops:
I think we could parse the URL of the tile-server in general and match it against a couple of known domains:

```js
const url = new URL('https://tile.openstreetmap.org/{z}/{x}/{y}.png');
const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
const isOsmUrl = osmHosts.some((host) => url.hostname.includes(host));
```

Falke-Design:
[...]

PR #9489, opened 2024-10-01
resolves 9228 - Automatically attributes OSM if OSM tiles are used and no attribution is provided

Jouwee:
Fixes: #9228 

Automatically attributes OSM if OSM tiles are used and no attribution is provided, resolving issue 9228

Wasn't sure about the test policies, so I created one for each new behaviour added.

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -92,6 +92,18 @@ export const TileLayer = GridLayer.extend({
 
 		options = Util.setOptions(this, options);
 
+		// in case the attribution hasn't been specified, check for known hosts that require attribution
+		if (options.attribution === null && URL.canParse(url)) {
+			const urlHostname = new URL(url).hostname;
+
+			// check for Open Street Map hosts
+			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
+			const isOsmUrl = osmHosts.some(host => urlHostname.endsWith(host));
+			if (isOsmUrl) {
+				options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
+			}
+		}
+
 		// detecting retina displays, adjusting tileSize and zoom levels
 		if (options.detectRetina && Browser.retina && options.maxZoom > 0) {
 
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -98,8 +98,7 @@ export const TileLayer = GridLayer.extend({
 
 			// check for Open Street Map hosts
 			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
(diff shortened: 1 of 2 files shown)


src/layer/tile/TileLayer.js
```diff
@@ -92,6 +92,18 @@ export const TileLayer = GridLayer.extend({
 
 		options = Util.setOptions(this, options);
 
+		// in case the attribution hasn't been specified, check for known hosts that require attribution
+		if (options.attribution === null && URL.canParse(url)) {
+			const urlHostname = new URL(url).hostname;
+
+			// check for Open Street Map hosts
+			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
+			const isOsmUrl = osmHosts.some(host => urlHostname.endsWith(host));
```
IvanSanchez, line 101: I'd rather get rid of `const isOsmUrl`, and do `if (osmHosts.some(...)) {` instead.

IvanSanchez (dismissed):
Besides my nitpicking, it looks good.

src/layer/tile/TileLayer.js
Jouwee: I think the `const isOsmUrl` makes the code slightly more readable, describing what the `osmHosts.some` accomplishes. But I have no strong opinions, I can change it if you think inline it's better.

Falke-Design: I see it like you @Jouwee , it makes it more readable. But lets make the change, then I will merge it

Jouwee: Change is made. I also see that a test has failed, but it seems to be a flaky test?

Falke-Design (approved):

Falke-Design:

simon04:
`URL.canParse` is not supported in Safari Version 16.5.1 (18615.2.9.11.7) [released in July 2023](https://support.apple.com/en-us/121012).

MDN https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static categorizes this feature as "Baseline 2023
Newly available. Since ⁨December 2023⁩, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers."

Can we find a solution not breaking two year old browsers?

Downstream bug report: https://gitlab.com/albina-euregio/albina-website/-/issues/700

jonkoops:
@simon04 agreed. When is this considered 'widely available'?

simon04:
> When is this considered 'widely available'?

According to https://web.dev/baseline/,

> Newly available: The feature is supported by all of the core browsers, and is therefore interoperable.
>
> Widely available: 30 months have passed since the newly interoperable date. The feature can be used by most sites without worrying about support.

jonkoops:
Very nice, I am glad we can have something to reference when implementing new features to ensure their compatibility. We should probably document that we require 'widely available' for any contributions that do not include a polyfill. I need to add that to the release notes for v2.

PR #9499, opened 2024-10-13
Attach events to the ownerDocument of the element

plazmdk:
This will attach events to the ownerDocument and not the document from where the map was initialized in order to support showing a leaflet map in a window.open() attached map.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -114,8 +114,8 @@ export const Draggable = Evented.extend({
 		this._parentScale = DomUtil.getScale(sizedParent);
 
 		const mouseevent = e.type === 'mousedown';
-		DomEvent.on(document, mouseevent ? 'mousemove' : 'touchmove', this._onMove, this);
-		DomEvent.on(document, mouseevent ? 'mouseup' : 'touchend touchcancel', this._onUp, this);
+		DomEvent.on(this._element.ownerDocument, mouseevent ? 'mousemove' : 'touchmove', this._onMove, this);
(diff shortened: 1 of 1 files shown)


plazmdk:
@Falke-Design - what did that force push do?

Falke-Design:
@plazmdk it was a rebase on the latest commit. Didn't modified anything of your code changes

jonkoops:
Could you provide an example of the problem this fixes? I am quite curious how this would work in combination with `window.open()`.

plazmdk:
I have made a plugin for webshops that allows the administrator to view the orders on a separate map and select the "closest" on a map for shipping purposes (the shops are selling big things that are delivered on a per km basis, so grouping orders correctly matters)

I have made a [codepen](https://codepen.io/plazmdk/pen/LEYJpNw) with the issue (you can't drag'n'drop) without the change.

Falke-Design (changes requested):
If we switch to `ownerDocument` then we need o change it everywhere and there are points where we don't have access to `ownerDocument` like in `DomUtil.get` or in `Icon.Default` -> `const el = DomUtil.create('div',  'leaflet-default-icon-path', document.body);`

coryasilva:
@plazmdk Could you use an iframe instead? (might need postMessage API depending on what all you are doing)

Or I wonder about just creating a draggable dialog/model/window in the same document -- I am totally naive to your constraints..

plazmdk:
@coryasilva - this allows the user to place the map on a separate screen. I might just close this and continue using my local branch, as I am not going to invest in changing a lot of code which is not affecting the use case this solves. It has been working just fine for more than a year now :-)

Issue #9533, opened 2024-11-22
GeoJSON `MultiLineString` and `MultiPolygon` only render first one of each


piranna:
At https://github.com/Leaflet/Leaflet/blob/142f94a9ba5757f7e7180ffa6cbed2b3a9bc73c9/src/layer/GeoJSON.js#L196-L204, we see that GeoJSON both `MultiLineString` and `MultiPolygon` types are rendering only their first geometry, instead of all of them. Since we have `MultiPoint`, `GeometryCollection` and `FeatureCollection` that are being properly rendered using a `FeatureGroup` object, fix should be trivial to do, just replicate the same thing of creating the layers in a loop and group them in the `FeatureGroup` object.

Falke-Design:
Works as designed. It is not wanted, that a MultiPolygon is split into multiple standalone Polygons.

piranna:
I think here there are two issues, one of them that only first polygon is being rendered, and another ones that all of them should be rendered, collected in a FeatureGroup object. They would still apear as a single Feature, but all of them would be rendered, not only the first one as currently happens.

PR #9540, opened 2024-11-26
Fixed multiLineString and multiPolygon issue #9533

Samyak0-0:
Fixed multiLineString and multiPolygon issue fixes: #9533


simon04 (changes requested):
According to the GeoJSON feature demo #9679, everything is being rendered just fine.

Falke-Design:
It is not wanted, that a MultiPolygon is split into multiple standalone Polygons. But still thank you for your contribution

PR #9552, opened 2024-12-08
Ignored event cancel bug fix #9444

### LINKED ISSUE #9444: Ignored attempt to cancel a touchmove event with cancelable=false
### Checklist

### Steps to reproduce

Unfortunately I haven't found the way yet to consistently reproduce that issue.
It requires a touch device (or a desktop browser in mobile mode).

### Expected behavior

I'd expect no error to show up in the console while browsing a Leaflet map.

### Current behavior

I've seen the following error repeating in our logs:
```
IgnoredEventCancel: Ignored attempt to cancel a touchmove event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.
```

It comes from the `preventDefault` in the `_onMove` function: https://github.com/Leaflet/Leaflet/blob/386fbd401d88d315e0d73514b874c7707ddb7319/src/dom/Draggable.js#L143

It looks like this error could be avoided by checking whether or not the event is cancellable: https://www.uriports.com/blog/easy-fix-for-intervention-ignored-attempt-to-cancel-a-touchmove-event-with-cancelable-false/

_No response_

### Environment

- Leaflet version: 1.9.4
- Browser (with version): 
- OS/Platform (with version):

emihu:
Resolves issue #9444 

Added a check to see whether the event is cancellable before calling preventDefault.

### DIFF
--- a/src/dom/Draggable.js
+++ b/src/dom/Draggable.js
@@ -140,7 +140,7 @@ export const Draggable = Evented.extend({
 		offset.x /= this._parentScale.x;
 		offset.y /= this._parentScale.y;
 
-		if (DomEvent.cancelable) { DomEvent.preventDefault(); }
+		if (DomEvent.cancelable) { DomEvent.preventDefault(e); }
 
 		if (!this._moved) {
 			// @event dragstart: Event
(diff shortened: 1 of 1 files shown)


mickaeltr (approved):

Falke-Design:
`DomEvent` has no `cancelable`, the correct woul be `ev.cancelable`

PR #9560, opened 2024-12-14
Exceptions for plus-lighter workaround to fix white grid

Nekzuris:
The current workaround for tile gaps in Chrome has the unfortunate side effect of creating a white grid at certain browser zoom levels, see https://github.com/Leaflet/Leaflet/issues/3575#issuecomment-1881642439.
This grid can however be removed with a few exceptions in CSS for specific pixel ratios. A few very thin black lines still remain but are far less distracting. I made extensive tests [here](https://github.com/openstreetmap/iD/pull/10594#issuecomment-2540006158).

currently at 110%
now

with CSS exceptions
patch

Unfortunately this doesn't fix everything, most notably 90%, but still I think this improvement is worth 1.9.5.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -53,10 +53,15 @@
 	width: auto;
 	padding: 0;
 	}
+
+/* Workaround to remove visible grid around tile borders on Chrome */
+/* with exceptions for 110%, 90%, 80% and 67% browser zoom */
+@media not ((1.09 < -webkit-device-pixel-ratio < 1.11) or (0.89 < -webkit-device-pixel-ratio < 0.91) or (0.79 < -webkit-device-pixel-ratio < 0.81) or (0.66 < -webkit-device-pixel-ratio < 0.68)) {
 .leaflet-container img.leaflet-tile {
 	/* See: https://bugs.chromium.org/p/chromium/issues/detail?id=600120 */
 	mix-blend-mode: plus-lighter;
 	}
+}
 
 .leaflet-container.leaflet-touch-zoom {
 	touch-action: pan-x pan-y;


Nekzuris:
Specific tests for Leaflet, it's a bit different than iD

### Firefox
| Browser zoom | normal | plus-lighter |
|:---:|:---:|:---:|
| any | ✔️ | ✔️ |

### Chrome
| Browser zoom | normal | plus-lighter |
|:---:|:---:|:---:|
| 33% | black grid | black grid & blurry |
| 50% | thin black lines | blurry |
| 67% | thin black lines | white grid & blurry |
| 75% | thin black lines | blurry |
| 80% | thin black lines | white grid & blurry |
| 90% | black grid | black grid & blurry |
| 100% | ✔️ | ✔️ |
| 110% | thin black lines | white grid & blurry |
| 125-400% | thin black lines | ✔️️slight blur |

I didn't notice the blur issue with plus-lighter on iD because it's in fact blurry all the time. I think it's due to their sub-pixel translation. It's only an issue on Chrome, Firefox is perfect no matter what.

Nekzuris:
After further testing with non-standard pixel ratios, I found that the _plus-lighter_ trick only works with zoom levels that are multiples of 25 (75%, 100%, 125%...)

So instead of adding a few exceptions, I changed it to apply only to those specific pixel ratios where it’s been confirmed to work.
The blurring issue isn’t always present and is much less noticeable than the gaps anyway.

Falke-Design:
Thank you for your investigation. It sounds correct what your describing but I didn't see a difference on my computer. I will investigate deeper, if the new fix in Chromium will not work. https://issues.chromium.org/issues/40084005

Nekzuris:
The `plus-lighter` workaround has other issues, a better solution has been found here https://github.com/openstreetmap/iD/issues/10747

Falke-Design:
Thank you for sharing. Do you know how thr best way is for us to implement this workaround?

Nekzuris:
I'm not sure if it's the best way, but I tested it by adding this SVG in the HTML
```svg
<svg>
    <defs>
        <filter id="alpha-5">
            <feComponentTransfer>
                <feFuncR type="identity"></feFuncR>
                <feFuncG type="identity"></feFuncG>
                <feFuncB type="identity"></feFuncB>
                <feFuncA type="linear" slope="5"></feFuncA>
            </feComponentTransfer>
        </filter>
    </defs>
</svg>
```

and this CSS
```css
.leaflet-container img.leaflet-tile {
    filter: url(#alpha-5);
}
```

And I'm not sure why but in my testing it doesn't work as well as for iD, only fixing the grid at 90% and 110%, but not completely at other zoom levels.

Issue #9575, opened 2025-01-20
Uncaught TypeError: Cannot read properties of undefined (reading '_leaflet_pos')


Mark-Falconbridge-i2:
### Checklist

- [x] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [x] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

1. Using a Leaflet map,
2. Zoom in or out
3. Quickly afterwards remove the map

I've created a minimal example here:

https://jsfiddle.net/bevxgaqk/1/ 

Click on the zoom in button and immediately after press the Esc key to remove the map, check the browser console.

### Expected behavior

No console errors

### Current behavior

Currently you get an error reported in the browser console:

```
Uncaught TypeError: Cannot read properties of undefined (reading '_leaflet_pos')
    at Pe (DomUtil.js:247:12)
    at e._getMapPanePos (Map.js:1488:10)
    at e._getNewPixelOrigin (Map.js:1505:67)
    at e._move (Map.js:1224:28)
    at e._onZoomTransitionEnd (Map.js:1726:8)
```

### Minimal example reproducing the issue

https://jsfiddle.net/bevxgaqk/1/ 

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Any but I'm using Chrome 131
- OS/Platform (with version): Windows 10

PR #9576, opened 2025-01-20
Clear timeout on remove (#9575)

Mark-Falconbridge-i2:
Fixes: #9575

When the map was removed, it wasn't tidying up a timeout which lead to Uncaught TypeError: Cannot read properties of undefined (reading '_leaflet_pos'). The fix is to keep a handle to the timeout and clear it when the map is removed.


Falke-Design (changes requested):
Good catch and very clean solution! But please use the `main` branch as base, because `v1` is not developed anymore.

Additionally please also check the other `setTimeout` usages. For example https://github.com/Leaflet/Leaflet/blob/c511e66c20dce06ac05c43cf287b3ecdba00d924/src/map/handler/Map.ScrollWheelZoom.js#L56 fails too. This one can be easy resolved by adding `clearTimeout` into `removeHooks`: https://github.com/Leaflet/Leaflet/blob/c511e66c20dce06ac05c43cf287b3ecdba00d924/src/map/handler/Map.ScrollWheelZoom.js#L37

Mark-Falconbridge-i2:
That's the thing, I couldn't get my unit test to fail against the main branch because other changes are in there that aren't in the 1.9.4 release version. I'm happy to add that other tidy up code too.

Falke-Design:
You can try overwrite the map function and check if it is called or not:

```
const zoomTransitionSpy = sinon.spy();
map._onZoomTransitionEnd =zoomTransitionSpy ;
...
expect(zoomTransitionSpy.called).to.be(false);
```

Mark-Falconbridge-i2:
That's a bit more white box than I'd like.

Closing this pr in favour of https://github.com/Leaflet/Leaflet/pull/9577.

PR #9577, opened 2025-01-21
Clear timeouts on remove (#9575)

Mark-Falconbridge-i2:
Fixes: https://github.com/Leaflet/Leaflet/issues/9575

When the map was removed, it wasn't tidying up a timeout which lead to Uncaught TypeError: Cannot read properties of undefined (reading '_leaflet_pos'). The fix is to keep a handle to the timeout and clear it when the map is removed.

Also clears other timeouts which had a timeout  period of greater than 0.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -92,6 +92,24 @@ describe('Map', () => {
 			map.setZoom(2);
 		});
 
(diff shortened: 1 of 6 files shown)


Falke-Design (approved):

Falke-Design:

PR #9580, opened 2025-01-29
Back-port SSR support to v1

jeffreykog:
Hello! I'm using Leaflet in an Angular app, for which i have made a workaround for the lack of SSR support. But i noticed that in main a commit was made for Leaflet v2 to support SSR by making references to browser-only variables conditional. It didn't look too complicated to I cherry-picked that commit into v1 and also updated everything that was supported in the in-progress branch for v2. The two SSR test files seem to be fully functional. I have also tested this build in my Angular application and it works perfectly. I can now remove the SSR workaround and even upgrade to the new Vite build system.

Considering development velocity for v2 is low it would be really great is we can make a v1 release containing these SSR fixes. Nothing will break by doing this so it will most likely be safe for a minor release.

Thanks a lot for your great work on Leaflet.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -133,6 +133,34 @@ jobs:
           BUNDLEMON_PROJECT_ID: 61e0545915f6c3000980d0ed
           CI_COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
+  test-ssr:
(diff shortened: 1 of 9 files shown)


Falke-Design:
Thank you for your contribution but there will be no v1 release anymore. In the next weeks we will release the first alpha of v2

PR #9605, opened 2025-02-28
Implemented `trackResize` for L.Popup

### LINKED ISSUE #9588: Incorrect Popup Size Calculation with `<picture>` in Firefox
### Checklist

### Steps to reproduce

1. create marker with a popup which contains an image that will resize the popup element to be larger than the default
2. add the following code
   ```js
   document.querySelector(".leaflet-popup-pane").addEventListener("load", function (event) {
       let tagName = (event.target as HTMLElement).tagName;
       // @ts-ignore
       let popup: Popup = formationsMap._popup; // Currently open popup, if any.
   
       if ((tagName === "IMG" || tagName === "PICTURE" || tagName === "source") && popup && !popup["_updated"]) {
           popup["_updated"] = true; // prevent infinite loop
           popup.update();
       }
   }, true); // Capture the load event, because it does not bubble.
   ```
   to properly update the viewport upon the image loading
3. open site in firefox
4. position viewport so the popup would open off-screen
5. click on marker to open popup

### Expected behavior

When opening the popup on firefox, it will readjust the viewport position to contain the entire popup, as well as reposition the popup to the proper coordinates (once the image is loaded)

### Current behavior

When opening the popup on firefox, it will adjust the viewport so that the popup, without the image loaded, would be in the viewport. But since the popup has been expanded due to the image, it is in actuality out of the viewport and the popup will not be located at the proper coordinates.
Even if `popup.update()` is executed after a 1s delay (when the image is definitely loaded), it still will not update properly.

This bug applies to both Firefox Desktop and Firefox Mobile.
The proper behaviour is observed in Chromium browsers. (tested with Ungoogled Chromium on desktop & Cromite on mobile)

_No response_

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Firefox 135.0b9
- OS/Platform (with version): Arch Linux 6.13.1-arch1-1 x86_64

IvanSanchez:
Fixes #9588

I had to get rid of the manual handling of the popup's `minWidth` and `maxWidth`, because setting the popup's content's width was triggering resizes (even with the same dimensions) and entering an infinite loop. I hope it was a leftover from the IE days.

### DIFF
--- a/debug/tests/popup-resize.html
+++ b/debug/tests/popup-resize.html
@@ -1,0 +1,46 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - Popup Offset</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
+		<link rel="stylesheet" href="../../dist/leaflet.css" />
+		<link rel="stylesheet" href="../css/screen.css" />
+		<script type="importmap">
+			{
+				"imports": {
+					"leaflet": "../../dist/leaflet-src.esm.js"
+				}
+			}
+		</script>
+	</head>
+	<body>
+		<div id="map"></div>
+		<script type="module">
+			import {TileLayer, Map, Marker} from 'leaflet';
+
+			const osm = new TileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png', {maxZoom: 18});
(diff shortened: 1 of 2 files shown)


Falke-Design (approved):
Looks ok for me. 

I think it wasn't possible to create an automated test?

Falke-Design:
This causes that the Popup is rendered differently:

V1:
grafik

V2:
grafik

@IvanSanchez I'm not sure if this will cause more problems than it solves for users who want to migrate.

PR #9624, opened 2025-03-29
Refactor docs for ESM

Falke-Design:
Fixes: #8352 

No need to use the bundled version anymore:
Closes: #5807 
Closes: #7615
Closes: #5696

- Replaces all `var` with `const` / `let`
- Changing all examples to `modules`
- Replacing all factories `L.xYZ()` to `new XYZ()`
- Adding `importmap` to Tutorial
- Updating Plunkr code

PR #9626, opened 2025-03-30
Removes factory functions

Falke-Design:
Removes all Factory functions and removes all `L.` references in src, tests and docs (if not already in #9624)

`Point`, `LatLng`, `LatLngBounds` and `Bounds` needed a rewrite which is slightly different to before. Earlier the factory returned `null` but with the Class constructor it will always return a empty class object or throwing an error. For this I added the methods `validate` to `Point` and `Latlng`. Refactor can be seen in this commit: https://github.com/Leaflet/Leaflet/commit/e090b6e82aaa9e161ef39bdee65ab5626b2374e7 I'm open to changes, but this was the best I could come up with.

Sub-PR of #9624

closes: #7846


docs/examples/extending/extending-1-classes.md
```diff
@@ -174,17 +174,8 @@ Calling a method of a parent class is achieved by reaching into the prototype of
...
-    
-    function myBoxClass(name, options) {
-        return new MyBoxClass(name, options);
-    }
-    
-    
+
 ### Naming conventions
 
 When naming classes for Leaflet plugins, please adhere to the following naming conventions:
```
Falke-Design, line 181: @jonkoops how should the naming conventions for plugins be in future?

https://github.com/Leaflet/Leaflet/blob/f7b30296c74a8bd96ff89c3a6e13e8dfe96b0966/docs/examples/extending/extending-1-classes.md?plain=1#L181-L186

https://github.com/Leaflet/Leaflet/blob/c12d70f495878497637ce93309631ff6bee3dad9/PLUGIN-GUIDE.md?plain=1#L34-L38

jonkoops: > Never expose global variables in your plugin.

This is still sound.

> If you have a new class, put it directly in the L namespace (L.MyPlugin).

The namespace will no longer exist, this will have to be replaced with a regular import to whatever the import is. E.g:

```
import MyPlugin from "leaflet-my-plugin"
```

> If you inherit one of the existing classes, make it a sub-property (L.TileLayer.Banana).

This is no longer a best practice, instead they should be named something like `TileLayerBanana` and be exported separately from any other namespaces.

> Most existing plugins follow the convention of naming plugins (and repos) like this: `Leaflet.MyPluginName`. 
 You can use other forms (e.g. "leaflet-my-plugin-name"), 
 
 This is still valid, although I would consider the latter leading since it is the predominant convention for NPM packages.
 
 > just make sure to include the word "Leaflet" in the name so that it's obvious that it's a Leaflet plugin. 
 
 Seems like a good thing to keep.

Falke-Design:
@jonkoops pls review

simon04:
Hi @Falke-Design, hi all, I appreciate the momentum that leaflet development has regained. I also appreciate the code modernisation happening.

However, regarding this change, I fear that it will cause a lot of churn for every user of leaflet. Using the factory functions has been the preferred way of using leaflet (according to leaflet's own tutorials and many more found in the internet). Thus, I fear that many users will have to touch a lot of their code when upgrading to version 2.0.

Could this change be kept for a later major release of leaflet?

Falke-Design:
We decided to remove the global `L` and removing the facotries will come with that together. 

>  Thus, I fear that many users will have to touch a lot of their code when upgrading to version 2.0.

Yeah it will be a lot of work but someone doesn't want to do this work or if a plugin isn't maintained anymore, I created the polyfill script https://github.com/Falke-Design/Leaflet-V1-polyfill and as soon as this PR is merged, the factory methods will be available there too.

Falke-Design:
@jonkoops @IvanSanchez @mourner please review the last open PR of V2

simon04:
I'm integration testing for avalanche.report -- these seem to be the changes required to the own code base as well as some plugins: https://gitlab.com/albina-euregio/albina-website/-/commit/93a6f628511b6eeff677a9f1a328a841df986b36

This patch forces some plugins to remove their factory functions, too, as `L.control` is no longer defined:

```js
L.control.geonames = function (options) {
    return new L.Control.Geonames(options);
};

// Uncaught TypeError: Cannot set properties of undefined (setting 'geonames')
```

```js
L.svg.tile = function(tileCoord, tileSize, opts){
	return new L.SVG.Tile(tileCoord, tileSize, opts);
}

// Uncaught TypeError: Cannot set properties of undefined (setting 'tile')
```

simon04:
I've updated the plugin leaflet-control-geocoderfor leaflet 2.0 compatibility. The biggest pain was the removed `L.latLng` due to more advanced argument support. I also couldn't rely on the updated `L.LatLng` constructor for backwards compatibility. Would it make sense to just keep `L.latLng`, mark it as deprecated, and remove it in leaflet 3.0?

Falke-Design:
> I also couldn't rely on the updated L.LatLng constructor for backwards compatibility. Would it make sense to just keep L.latLng, mark it as deprecated, and remove it in leaflet 3.0?

Thank you for the feedback. I think it is better to force now that all the necessary changes are done, instead of hoping that the plugin will be updated with the next release again

jonkoops:
Great stuff @Falke-Design, this was another big one I wanted to land for v2!

Issue #9628, opened 2025-03-30
Replace unpkg with jsdelivr


Falke-Design:
Replace all occurrences of unpkg  with https://www.jsdelivr.com/

https://github.com/Leaflet/Leaflet/blob/c12d70f495878497637ce93309631ff6bee3dad9/docs/download.md?plain=1#L38-L39

Falke-Design:
@eshanair you can work on that. Please read through the comments of the other PR

Falke-Design:
> Let us create a new issue for this and tackle it later. I don't have the experience to make sure that everything works.
> For example the generated hash is different for [cdnjs](https://cdnjs.com/libraries/leaflet). They are using `sha512` and we `sha256` https://leafletjs.com/download.html

It looks like we already prepared it in https://github.com/Leaflet/Leaflet/pull/8459/

simon04:
Being curious: what is the rationale behind this replacement?

Falke-Design:
jsDelvir is much more used today and I also think it has a much better / modern UI to reference to

Falke-Design:
Closed by #9663

PR #9630, opened 2025-03-31
Replaced unpkg link with jsdelivr link

Sfi-codes:
Replaced unpkgs link with the jsdelivr links shown on the leaflet jsdelivr website #9628


Falke-Design (changes requested):
> Let us create a new issue for this and tackle it later. I don't have the experience to make sure that everything works.
> For example the generated hash is different for [cdnjs](https://cdnjs.com/libraries/leaflet). They are using `sha512` and we `sha256` https://leafletjs.com/download.html

_Originally posted by @Falke-Design in https://github.com/Leaflet/Leaflet/pull/9624#discussion_r2020137302_

PR #9636, opened 2025-04-08
Add missing variable initializer for .addData()

ArachnidAbby:
# Description

Was working on something at work and I noticed this is broken.

Looks like its just a simple oversight,  no big deal. Variable just needs to be initialized.

### DIFF
--- a/src/layer/GeoJSON.js
+++ b/src/layer/GeoJSON.js
@@ -102,7 +102,7 @@ export const GeoJSON = FeatureGroup.extend({
 		if (features) {
 			for (i = 0, len = features.length; i < len; i++) {
 				// only add this if geometry or geometries are set and not null
-				feature = features[i];
+				const feature = features[i];
 				if (feature.geometries || feature.geometry || feature.features || feature.coordinates) {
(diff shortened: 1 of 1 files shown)


Falke-Design:
Thank you for taking the time but the variable is initialized:

https://github.com/Leaflet/Leaflet/blob/c12d70f495878497637ce93309631ff6bee3dad9/src/layer/GeoJSON.js#L100-L105

ArachnidAbby:
> Thank you for taking the time but the variable is initialized:
> 
> https://github.com/Leaflet/Leaflet/blob/c12d70f495878497637ce93309631ff6bee3dad9/src/layer/GeoJSON.js#L100-L105

Yeah you're right, I was in a rush and missed that. I'm also generally not used to javascript having an "undefined" type so I saw an error that said "variable feature is undefined" and assumed it was the variable not being declared. Checking the source I just glanced over the fact that- yes- it was actually initialized, my data just had "undefined" values in it....

I hate javascript so much lol. Sorry for the waste of time and thanks for correcting me. You have a great rest of your day lol. Take care.

Issue #9627, opened 2025-03-30
Refactor `scrollToHeader` function


Falke-Design:
Refactor `scrollToHeader` function, which is used on the docs website to scroll to an clicked anchor link
https://github.com/Leaflet/Leaflet/blob/99355bf8a0c1d86d502ceed246496e4c641355f7/docs/docs/js/docs.js#L110-L125

- Remove the specific Chrome handling
- And [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) should be used

The specifc Chrome handling was added with this PR: #7982

jsearsy:
@Falke-Design I can create a PR for this one

jsearsy:
PR open: https://github.com/Leaflet/Leaflet/pull/9640

Falke-Design:
After trying for 2 hours to get the scrolling correct with different issues, I think it is not worth to change the current behavior. 

![scroll](https://github.com/user-attachments/assets/a737144f-ade7-43cc-ac3a-c9797920704f)

PR #9640, opened 2025-04-12
Docs | Refactor scrollToHeader function

jsearsy:
- Removes existing Chrome-specific handling
- Removes `sameAnchor` parameter and any calling arguments, as they're no longer required
- Uses [Element.scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView), replacing [Window.scrollTo](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo)

Fixes: #9627

### DIFF
--- a/docs/docs/js/docs.js
+++ b/docs/docs/js/docs.js
@@ -76,7 +76,7 @@ if (toc) {
 
 		if (elem.tagName === 'H2' || elem.tagName === 'H4') {
 			setTimeout(() => {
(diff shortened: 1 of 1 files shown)


Falke-Design:
After trying for 2 hours to get the scrolling correct with different issues, I think it is not worth to change the current behavior. 
Thank you for contributing and sorry that this will be not merged.

scroll

Issue #9445, opened 2024-08-19
Clicking videoOverlay seek circle sometimes immediately calls seekend in Chrome, moves map in Safari


reyemtm:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

Add a videoOverlay to a map with controls from a video element to a Leaflet map with a simple CRS.

In Chrome - Clicking the seek circle will sometimes immediately call seekend - you can see this in the console where mousdown is false even when the mouse is still being pressed. In the plain html5 video element this does not happen.

In Safari - clicking the seek bar appears to do nothing and the map just moves as the user attempts to seek the video.

### Expected behavior

Clicking the seek circle will always behave normally - not causing the video to immediate seek, and will not cause the map to move.

The issue with Safari is not as critical since in Chrome the user can sometimes seek the video, but not always, whereas in Safari it appears the user can never seek the video by moving the seek bar.

### Current behavior

I suspect that in Chrome, unless pressed at the exact mouse coordinates of the seek circle, the video seeks to a new time, ending the seeking event, and does not continue seeking as expected.

### Minimal example reproducing the issue

https://plnkr.co/edit/vKFmFFnQNT6cVxVf

### Environment

- Leaflet version: 1.9.4
- Chrome Version 127.0.6533.120 (Official Build) (arm64)
- Safari Version 17.6 (19618.3.11.11.5)
- OS/Platform (with version): MacOS

PR #9641, opened 2025-04-12
Fix video control / seek bar usage in safari

Falke-Design:
When the controls of a html5 video in a VideoOverlay are enabled, moving the seekbar wasn't possible it always dragged the map. With removing `vid.onmousemove = Util.falseFn;` we fix this.
The new `mousedown -> stopPropagation` logic prevents in Safari, dragging the map if the user holds down and moves inside of the VideoOverlay.

Fixes: #9445

safari_video_seekbar


simon04 (approved):
I can confirm that this PR fixes the issue in Safai Version 18.4 (20621.1.15.11.10).

Issue #7586, opened 2021-05-28
pane option disable the renderer assigned to the map


olivbrau:
Hi,
When I use Control.Layer I noticed that when I click on an item, it is drawn over everything, (so changing the order of drawing)
To correct this, I use Panes : I create 1 pane per "drawing layer", and I put 'pane' option on each item of this drawing layer (eg. on each polyline, polygon etc.)
It works well and the layer drawing order is respected.
But I noticed that if I use this 'pane' option, the items are drawn with a new SVG renderer !
I fact, I have put a Canvas renderer to my map, with option padding=0, but when I use 'pane' option, I notice that the padding is not 0. And the real probleme is that with SVG renderer, the 'tolerance' option doesn't work.

So how can, I use 'pane' option and keep my Canvas renderer ?
Is this a bug ?
Thanks a lot for your help.
Olivier

johnd0e:
> I have put a Canvas renderer to my map

Show your code please.
And make sure that you have not missed https://leafletjs.com/reference-1.7.1.html#map-prefercanvas option.

IvanSanchez:
> But I noticed that if I use this 'pane' option, the items are drawn with a new SVG renderer !

Yes, this is behaviour as expected, as per https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/Renderer.getRenderer.js . A `L.Path` in a non-default pane without an explicit `L.Renderer` spawns a new renderer.

> So how can, I use 'pane' option and keep my Canvas renderer ?

No, it's gonna be one or the other. You should consider spawning `L.Canvas` renderers *with explicit panes*, then tell your `L.Path`s to use that *renderer* (instead of telling your `L.Path`s to be in a specific pane). When dealing with `L.Path`s, an explicit renderer takes precedence over a explicit pane (the thing goes path→renderer→pane, and **not** path→pane→renderer; if there's no explicit renderer, then it's path→default-renderer-for-pane→pane).

> Is this a bug ?

No, but I'll say that this is a potential documentation improvement, since the docs don't make clear the precedence of renderers over panes.

johnd0e:
> You should consider spawning L.Canvas renderers with explicit panes

Unfortunately Leaflet is not good in handling multiple canvases with interactive features.

olivbrau:
Hi, thanks for your quick answers !
here is an example of code :
it works since if you deselect then reselect, 'green' layer in the control, **the green triangle remains between the red and blue one**.
[...]

PR #9651, opened 2025-04-18
Improve documentation about renderer and pane

Falke-Design:
Fixes: #7586

> No, it's gonna be one or the other. You should consider spawning L.Canvas renderers with explicit panes, then tell your L.Paths to use that renderer (instead of telling your L.Paths to be in a specific pane). When dealing with L.Paths, an explicit renderer takes precedence over a explicit pane (the thing goes path→renderer→pane, and not path→pane→renderer; if there's no explicit renderer, then it's path→default-renderer-for-pane→pane).
>
>    Is this a bug ?
>
> No, but I'll say that this is a potential documentation improvement, since the docs don't make clear the precedence of renderers over panes.

*Comment of [IvanSanchez](https://github.com/IvanSanchez) in https://github.com/Leaflet/Leaflet/issues/7586#issuecomment-850460317*

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -33,6 +33,7 @@ export const Layer = Evented.extend({
 	options: {
 		// @option pane: String = 'overlayPane'
 		// By default the layer will be added to the map's [overlay pane](#map-overlaypane). Overriding this option will cause the layer to be placed on another pane by default.
+		// If the `renderer` option is set, it will override the `pane` option.
 		pane: 'overlayPane',
 
 		// @option attribution: String = null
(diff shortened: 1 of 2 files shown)


src/layer/Layer.js
```diff
@@ -33,6 +33,7 @@ export const Layer = Evented.extend({
 	options: {
 		// @option pane: String = 'overlayPane'
 		// By default the layer will be added to the map's [overlay pane](#map-overlaypane). Overriding this option will cause the layer to be placed on another pane by default.
+		// If the `renderer` option is set, it will override the `pane` option.
```
simon04, line 36:
```suggestion
		// Not effective if the `renderer` option is set (the `renderer` option will override the `pane` option).
```

simon04 (approved):

PR #9650, opened 2025-04-18
Removes mozEvent warning

Falke-Design:
Before:
*MouseEvent.mozPressure should not be used. Use PointerEvent.pressure*
grafik

Now:
grafik

Added with #4493

PR #9652, opened 2025-04-19
Replace Util.extend with ES6 Object.assign or object spread

simon04:
According to MDN, both are considered Baseline: Widely available

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -167,11 +167,12 @@ export const Events = {
 	fire(type, data, propagate) {
 		if (!this.listens(type, propagate)) { return this; }
 
-		const event = Util.extend({}, data, {
+		const event = {
+			...data,
 			type,
 			target: this,
 			sourceTarget: data?.sourceTarget || this
-		});
+		};
 
 		if (this._events) {
 			const listeners = this._events[type];
@@ -294,10 +295,11 @@ export const Events = {
 
 	_propagateEvent(e) {
 		for (const p of Object.values(this._eventParents ?? {})) {
-			p.fire(e.type, Util.extend({
+			p.fire(e.type, {
 				layer: e.target,
-				propagatedFrom: e.target
-			}, e), true);
+				propagatedFrom: e.target,
+				...e
+			}, true);
 		}
 	}
 };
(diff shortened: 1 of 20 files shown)


Falke-Design (approved):
Looks good. Should we remove the function Util.extend too?

simon04:
> Should we remove the function Util.extend too?

There are a few occurrences left, in particular in `Class.js` and a nasty one in `Map.js` that i did not manage to replace: https://github.com/Leaflet/Leaflet/blob/07252ebffb191953e4c5c3197ba6c269f0b8482b/src/map/Map.js#L1426

Falke-Design:
I only found the Map occurence: https://github.com/Leaflet/Leaflet/blob/07252ebffb191953e4c5c3197ba6c269f0b8482b/src/map/Map.js#L1426 and some in the tests but all could be replaced.

Map.js:
```
const synth =  {};
for (const i in e) {
	synth[i] = e[i];
}
```

And pls fix the merge conflicts

simon04:
The change in `Map.js` might be obsolete with #9650.

Falke-Design:
Ah true ... I forgot #9650 😄

simon04:
Fixed yet another merge conflict =) ...

src/core/Events.js
Falke-Design, line 1: Can we pls keep the old format

Falke-Design:
I would like to get `Util.extend` removed completly, can you do this please?

src/core/Events.js
simon04: The pre-commit hook changed CRLF to LF. Trying to fix it...

Falke-Design: Yeah I struggled with that already too... It is on the [roadmap](https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-1577098) to fix this, but somewhere we discussed that it can destroy the git history ...

simon04:
> I would like to get `Util.extend` removed completly, can you do this please?

Done. All usages removed. All git conflicts resolved.

Falke-Design (approved):
Thank you

PR #9653, opened 2025-04-19
Use for...of

### LINKED ISSUE #7402: for...in iterating through inherited properties
Hello. I've noticed that there are many `for...in` loops that iterate over all properties, including inherited ones. This leads to errors when some other script has added properties to the prototype. For instance, I ran into this error when I tried to create a map in a page where other script had added an extra property to `Object`'s prototype:

    TypeError: Cannot read property 'fn' of undefined at NewClass._on (leaflet-src.js:510)

Adding properties to native prototypes may be a bad practice, but sometimes you don't have control over it. And it's also a good practice to use `hasOwnProperty` with `for...in` loops.

Should I push these validations? I added them in my local copy and the app is working fine now.

**Steps to reproduce**
- Alter `Object` prototype (`Object.prototype.test="test";`)
- Create a new map (`L.map()`)

**Expected behavior**
Map should be created normally.

**Current behavior**
Fails due to the unexpected property `test` in `Object`.

**Environment**
- Leaflet version: 1.7.1
- Browser (with version): Opera 73.0

**Additional context**
See intro.

https://plnkr.co/edit/iLQ7Mlm5tbvZMgyv

simon04:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

Fixes #7402 (for all but `Util.extend` and `Util.setOptions` which are used for `Class.extend`)

### DIFF
--- a/debug/vector/vector-bounds.html
+++ b/debug/vector/vector-bounds.html
@@ -44,19 +44,10 @@
 				[39.67759833072648, -104.89471435546875]
 			];
 
-			const latlngs = [];
-
-			for (let i = 0, len = pathPoints.length; i < len; i++) {
-				latlngs.push(new LatLng(pathPoints[i][0], pathPoints[i][1]));
-			}
-
+			const latlngs = pathPoints.map(p => new LatLng(p[0], p[1]));
 			const path = new Polyline(latlngs);
-			const latlngs2 = [];
 
-			for (let i = 0, len = polyPoints.length; i < len; i++) {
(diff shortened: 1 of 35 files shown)


Falke-Design (approved):
Very nice!!

Have you replaced all `for...in` loops? Or can you maybe check them to solve #7402 too?
Does it make sense to force with a lint rule for...of instead of for...in?

Some references: #8879, #8843

Falke-Design (approved):

PR #9654, opened 2025-04-19
Replace Util.getParamString with URL.searchParams

simon04:
https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams


src/layer/tile/TileLayer.WMS.js
```diff
@@ -110,9 +110,11 @@ export const TileLayerWMS = TileLayer.extend({
 		    [min.y, min.x, max.y, max.x] :
 		    [min.x, min.y, max.x, max.y]).join(','),
 		    url = TileLayer.prototype.getTileUrl.call(this, coords);
-		return url +
-			getParamString(this.wmsParams, url, this.options.uppercase) +
-			(this.options.uppercase ? '&BBOX=' : '&bbox=') + bbox;
+		const url0 = new URL(url);
```
Falke-Design, line 113: I think there is no need to keep `url`. Let us replace `url0` with `url`
```suggestion
		const url = new URL(TileLayer.prototype.getTileUrl.call(this, coords));
```

Falke-Design (changes requested):

src/layer/tile/TileLayer.WMS.js
simon04: Thanks, that's clearer.

Falke-Design (approved):

PR #9655, opened 2025-04-19
Convert Bounds, LatLng, LatLngBounds, Point, Transformation to ES6 classes

simon04:


src/geometry/Point.js
```diff
@@ -100,73 +95,71 @@ Point.prototype = {
...
 	trunc() {
 		return this.clone()._trunc();
-	},
+	}
 
 	_trunc() {
-		this.x = trunc(this.x);
-		this.y = trunc(this.y);
+		this.x = Math.trunc(this.x);
+		this.y = Math.trunc(this.y);
```
simon04, line 149: `Math.trunc` is widely supported according to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc.

Falke-Design:
Thank you but will be already fixed with #9622

simon04:
Hi @Falke-Design , the `Math.trunc` change is just a minor detail. This PR is mainly about converting the code to use native ES6 classes (please see the diff).

Falke-Design:
> native ES6 classes 
yes just saw it. I think we will take your PR and rebase the PR #9626 on it

PR #9609, opened 2025-03-02
Run test with the source files

Falke-Design:
With changing the source files to be completley runnable without any bundler, we can include them directly in the tests.

Loading the `version` from the `package.json` caused the problem. 

**Solution:** In the source files we have now the version `0.0.0` and with running rollup, we replace the version with the version from package.json.

Fixes: #8937

PR #9657, opened 2025-04-20
Export leaflet `version` via `createBanner`

simon04:
Replaces package.json import, and rollup-plugin-json, and rollup-plugin-git-version.


Falke-Design:
We made Leaflet importable without the need to bundle it. When I understand your changes correctly, the version would be now never be exported in an unbundled import

Falke-Design:
And the `preprocessors` will be removed as soon Firefox implements json import

Falke-Design:
> Changing the import to import pkg from '../package.json' with { type: 'json' } causes the entire package.json to end up in the leaflet-src.js as well as leaflet.js bundle.

Is not ideal, that is true

simon04:
> When I understand your changes correctly, the version would be now never be exported in an unbundled import

True. Sorry, I did not consider "Leaflet importable without the need to bundle it" for this PR.

`L.version` is being used, of course: https://github.com/search?q=leaflet+L.version&type=code

What about writing a `version.json` file during the build process? I'm happy to look into that...

Falke-Design:
> What about writing a version.json file during the build process? I'm happy to look into that...

There should be a better way to do this, but I have no other idea ... How are doing this other repos?

@jonkoops any ideas how we can export the `version` without including the whole package.json? https://github.com/Leaflet/Leaflet/pull/9609#issuecomment-2817039502

simon04:
Here's what OpenLayers is doing: https://github.com/openlayers/openlayers/blob/5a8bd452e8a7a0fe55cab268cdceb0b642985fdc/tasks/set-version.js updates `export const VERSION` in https://github.com/openlayers/openlayers/blob/5a8bd452e8a7a0fe55cab268cdceb0b642985fdc/src/ol/util.js#L36

Falke-Design:
@simon04 okay thank you. Can you prepare something similar?

simon04:
Superseded by https://github.com/Leaflet/Leaflet/pull/9662.

PR #9660, opened 2025-04-20
Enable `prefer-exponentiation-operator` linting rule and fix issues

simon04:
https://eslint.org/docs/latest/rules/prefer-exponentiation-operator

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation


Falke-Design (approved):
Fine for me, thank you

PR #9662, opened 2025-04-20
Export leaflet `version` as string replaced via rollup

simon04:
Supersedes #9657.


Falke-Design:
With that change we could also remove the `preprocessors` for Firefox?

simon04:
> With that change we could also remove the `preprocessors` for Firefox?

Yeah. I've removed them.

Falke-Design:
The down side of this approch is, that for the not bundled version, the version is always `latest`. 
At the moment the version in package.json will be modified with a new release and is then exported for the unbundled version too.

Falke-Design:
@simon04 did you push? I don't see any changes

simon04:
Err, not to this branch (was considering a separate branch for a moment). Pushed now.

Falke-Design (approved):
LGTM

waiting for a second approval

Falke-Design:

IvanSanchez:
@Falke-Design @simon04 Please remind me - what's the rationale for ditching the current rollup plugins approach? Compatibility issues with 2025 toolchains?

I'm kinda against using `latest`, as I'd prefer the git commit hash in that scenario.

Falke-Design:
@IvanSanchez the goal would be to be able to access the classes directly without the bundle. 
With the current approach, loading the package.json and export version we add the whole package.json to the bundle: https://github.com/Leaflet/Leaflet/pull/9609#issuecomment-2817039502

Falke-Design:
I don't like "latest" that much either. We could always overwrite the `version` variable with the actual version and commit it

IvanSanchez:
I'd rather have something like `__LEAFLET_VERSION__` as the placeholder string.

Also, can you folks try out if the `namedExports` option of the JSON rollup plugin ( https://github.com/rollup/plugins/tree/master/packages/json#namedexports ) would solve the "whole package.json in bundle" issue?

simon04:
The named export was in use prior to https://github.com/Leaflet/Leaflet/pull/9609. But browsers (running the non-bundled code while testing) do not support named imports, see https://github.com/Leaflet/Leaflet/pull/9609#issuecomment-2725847209.

Falke-Design:
Thanks for the suggestion but I just merged #9717

PR #9656, opened 2025-04-20
docs: replace unpkg with jsDelivr and update SRI hashes (fixes #9628)

eshanair:
- Replaced all Leaflet CDN references from unpkg to jsDelivr
- Updated SRI hashes to sha512 for jsDelivr CDN
- Verified all docs and examples, removed unpkg from CDN lists
- (Optional) Added a Husky pre-commit check to block unpkg.com links in docs. However, this could be removed if the team prefers to keep the workflow simpler!

PR #9663, opened 2025-04-20
docs: replace unpkg with jsDelivr in documentation (fixes #9628)

### LINKED ISSUE #9628: Replace unpkg with jsdelivr

eshanair:
Thank you @Falke-Design for your input and corrections on #9656 for #9628 (Replace unpkg with jsDelivr). I believe I've addressed all feedback in this new PR:

- Only updated current docs/examples (no old posts/plugins)
- No Husky hook (removed as requested)
- Verified via local build (`bundle exec jekyll serve`)

I apologize for the earlier formatting issues, I was more careful this time. However, I am a bit confused on what I should do for the security hashes. I noticed your comment about the hashes being generated by build/integrity.js. Should I run this script to generate new hashes for jsDelivr?

Please let me know if anything else needs adjustment. I’m happy to fix it promptly!

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -35,15 +35,15 @@ so please read the changelog carefully when upgrading to it.
 The latest stable Leaflet release is available on several CDNs. To start using it with an [importmap](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap), place the following in the `head` of your HTML code:
 
 ```html
-<link rel="stylesheet" href="https://unpkg.com/leaflet@{{ site.latest_leaflet_version }}/dist/leaflet.css" integrity="{{site.integrity_hash_css}}" crossorigin="anonymous" />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/leaflet@{{ site.latest_leaflet_version }}/dist/leaflet.css" integrity="{{site.integrity_hash_css}}" crossorigin="anonymous" />
 
 <script type="importmap">
 {
 	"imports": {
-		"leaflet": "https://unpkg.com/leaflet@{{ site.latest_leaflet_version }}/dist/leaflet.js"
+		"leaflet": "https://cdn.jsdelivr.net/npm/leaflet@{{ site.latest_leaflet_version }}/dist/leaflet.js"
 	},
 	"integrity": {
-		"https://unpkg.com/leaflet@{{ site.latest_leaflet_version }}/dist/leaflet.js": "{{site.integrity_hash_uglified}}"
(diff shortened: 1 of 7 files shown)


docs/download.md
```diff
@@ -69,7 +69,7 @@ Then, in your script, import the needed Leaflet Classes as follows:
 
 Note that the [`integrity` hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) are included for security when using Leaflet from a CDN.
 
-Leaflet is available on the following free CDNs: [unpkg](https://unpkg.com/leaflet/dist/), [cdnjs](https://cdnjs.com/libraries/leaflet), [jsDelivr](https://www.jsdelivr.com/package/npm/leaflet?path=dist).
+Leaflet is available on the following free CDNs: [cdnjs](https://cdnjs.com/libraries/leaflet), [jsDelivr](https://www.jsdelivr.com/package/npm/leaflet?path=dist).
```
Falke-Design, line 72: Let us keep unpkg here but as last one in the list and jsdelivr as first one

Falke-Design (changes requested):
In [RELEASE.md](https://github.com/Leaflet/Leaflet/blob/main/RELEASE.md) you find this step:
> Run npm run docs to generate the new docs/reference.html and update integrity hashes in docs/_config.yml

This means after we release a new version we run `npm run docs`. The script fetchs from a cdn API the hash of the just released version and put it in the `_config.yml` file. The fetch of the hash needs to be changed to the correct API on jsdelivr.

https://github.com/Leaflet/Leaflet/blob/9e199c8ad16f5181542c4d2ca9b7a6e758894aa1/package.json#L49

https://github.com/Leaflet/Leaflet/blob/9e199c8ad16f5181542c4d2ca9b7a6e758894aa1/build/integrity.js#L11-L15

To test you can execute `npm run docs` by your self

docs/download.md
eshanair: Updated the CDN list as requested - moved jsDelivr to first position and kept unpkg as last. Verified locally with npm run docs and everything looks correct.

Falke-Design (changes requested):
You need to update unpkg outside of the docs folder too. `RELEASE.md` and `integrity.js` is missing.

It is enough to just change the url path in `integrity.js` - no need to change the integrity hashes logic. jsDelivr also uses `sha256`.

eshanair:
Thanks for your patience and help. I just updated RELEASE.md and integrity.js.

Falke-Design (approved):

Falke-Design:

PR #9667, opened 2025-04-21
Use CSS nesting

simon04:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting
https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector
> Baseline 2023: Since December 2023, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers.

### DIFF
--- a/dist/leaflet.css
+++ b/dist/leaflet.css
@@ -13,93 +13,137 @@
 	position: absolute;
 	left: 0;
 	top: 0;
-	}
+}
+
 .leaflet-container {
 	overflow: hidden;
-	}
+}
+
 .leaflet-tile,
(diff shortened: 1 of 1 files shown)


Falke-Design (approved):
Sometimes harder to read for me but in general better.

Waiting for a second approval

simon04:
Would it be possible to craft a v2.0.0-alpha.2 release in order to test all recent changes? Thank you!

jonkoops:
Let's see if we can get #9906 and #9880 merged in first, they are kind of big ticket items. I hope to get some work done on it next week.

PR #9677, opened 2025-04-21
Convert everything to ES6 classes and inheritance

simon04:
- **Convert Events to ES6 class**
- **Convert Draggable to ES6 class**
- **Convert PosAnimation to ES6 class**
- **Convert Control to ES6 classes**
- **Convert Icon to ES6 classes**
- **Convert Handler to ES6 classes**
- **Convert Map to ES6 class**
- **Convert Layer to ES6 classes**
- **Use Class.setDefaultOptions**
- **Use Class.mergeOptions**

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -49,6 +49,10 @@ export class Class {
 		return NewClass;
 	}
 
+	static setDefaultOptions(options) {
+		this.prototype.options = {...this.prototype.options, ...options};
+	}
+
 	// @function include(properties: Object): this
 	// [Includes a mixin](#class-includes) into the current class.
 	static include(props) {
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -49,10 +49,6 @@ export class Class {
 		return NewClass;
 	}
 
-	static setDefaultOptions(options) {
-		this.prototype.options = {...this.prototype.options, ...options};
-	}
-
 	// @function include(properties: Object): this
 	// [Includes a mixin](#class-includes) into the current class.
 	static include(props) {
@@ -68,8 +64,7 @@ export class Class {
 	// @function mergeOptions(options: Object): this
 	// [Merges `options`](#class-options) into the defaults of the class.
 	static mergeOptions(options) {
-		this.prototype.options ??= {};
-		Object.assign(this.prototype.options, options);
+		this.prototype.options = {...this.prototype.options, ...options};
 		return this;
 	}
 
--- a/src/core/Handler.js
+++ b/src/core/Handler.js
@@ -9,10 +9,11 @@ import {Class} from './Class.js';
 // @aka L.Handler
 // Abstract class for map interaction handlers
 
-export const Handler = Class.extend({
-	initialize(map) {
+export class Handler extends Class {
+	constructor(map) {
+		super();
 		this._map = map;
-	},
+	}
 
 	// @method enable(): this
 	// Enables the handler
@@ -22,7 +23,7 @@ export const Handler = Class.extend({
 		this._enabled = true;
 		this.addHooks();
 		return this;
-	},
+	}
 
 	// @method disable(): this
 	// Disables the handler
@@ -32,7 +33,7 @@ export const Handler = Class.extend({
 		this._enabled = false;
 		this.removeHooks();
 		return this;
-	},
+	}
 
 	// @method enabled(): Boolean
 	// Returns `true` if the handler is enabled
@@ -46,12 +47,12 @@ export const Handler = Class.extend({
 	// Called when the handler is enabled, should add event hooks.
 	// @method removeHooks()
 	// Called when the handler is disabled, should remove the event hooks added previously.
-});
 
-// @section There is static function which can be called without instantiating L.Handler:
-// @function addTo(map: Map, name: String): this
-// Adds a new Handler to the given map with the given name.
-Handler.addTo = function (map, name) {
-	map.addHandler(name, this);
-	return this;
-};
+	// @section There is static function which can be called without instantiating L.Handler:
+	// @function addTo(map: Map, name: String): this
+	// Adds a new Handler to the given map with the given name.
+	static addTo(map, name) {
+		map.addHandler(name, this);
+		return this;
+	}
+}
--- a/src/layer/BlanketOverlay.js
+++ b/src/layer/BlanketOverlay.js
@@ -21,7 +21,7 @@ export class BlanketOverlay extends Layer {
 	static {
 		// @section
 		// @aka BlanketOverlay options
-		this.setDefaultOptions({
+		this.mergeOptions({
 			// @option padding: Number = 0.1
 			// How much to extend the clip area around the map view (relative to its size)
 			// e.g. 0.1 would be 10% of map view in each direction
--- a/src/layer/BlanketOverlay.js
+++ b/src/layer/BlanketOverlay.js
@@ -16,26 +16,30 @@ import {Bounds} from '../geometry/Bounds.js';
  * that rely on one single HTML element
  */
 
-export const BlanketOverlay = Layer.extend({
-	// @section
-	// @aka BlanketOverlay options
-	options: {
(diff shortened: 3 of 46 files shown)


Falke-Design:
Is `include` still working? For example: https://github.com/bbecquet/Leaflet.RotatedMarker/blob/master/leaflet.rotatedMarker.js

simon04:
> Is `include` still working?

It should, yes. I tested with `git restore --source=main src/layer/marker/*Icon*`

simon04:
Hi @Falke-Design, all but 3 unit tests are passing. What do you think? Is it worth investing more time into this PR? This PR would facilitate native ES6 class inheritance.

One advantage for debugging is that in the JavaScript console, class instances would show up as instances of say `TileLayer`, when now they show up as `NewClass`.

Falke-Design:
I'm concerned that this could break plugins. But if we can surely say that the includes (overwriting), extends, etc. is working the same, then I would like to have this merged.

src/layer/BlanketOverlay.js
```diff
@@ -145,20 +149,20 @@ export const BlanketOverlay = Layer.extend({
...
 		return size;
-	},
-	_onZoomEnd: Util.falseFn,
-	_onViewReset: Util.falseFn,
-	_onSettled: Util.falseFn,
-});
+	}
+	_onZoomEnd() {}
+	_onViewReset() {}
+	_onSettled() {}
```
simon04, line 167: Those changes were essentials to fix the remaining three tests. Otherwise, somehow, inheritance did not work.

src/layer/tile/TileLayer.WMS.js
```diff
@@ -23,48 +23,50 @@ import {Bounds} from '../../geometry/Bounds.js';
...
+			// map CRS. Don't change this if you're not sure what it means.
+			crs: null,
+
+			// @option uppercase: Boolean = false
+			// If `true`, WMS request parameter keys will be uppercase.
+			uppercase: false
+		});
+	}
 
 	initialize(url, options) {
```
simon04, line 71: Eventually, we can migrate the remaining `initialize` to proper ES6 class `constructor`

src/core/Class.js
```diff
@@ -60,6 +60,13 @@ export class Class {
 		return this;
 	}
 
+	// @function setDefaultOptions(options: Object): this
+	// Configures the [default `options`](#class-options) on the prototype of this class.
+	static setDefaultOptions(options) {
+		Util.setOptions(this.prototype, options);
+		return this;
+	}
+
```
simon04, line 69: I could not find a more concise way to set the default options including prototype inheritance with the new ES6 class syntax. A new function is required, since neither `include` not `mergeOptions` use `Object.create` to enable prototype inheritance (and tests explicitly test for this).

I also tried/considered `Map.prototype.options= {...}` after `class Map extends Evented {...}`, but then the options hidden at the bottom of the source files.

simon04 (commented):
Yay, all tests are passing now. 😊

simon04:
~Successfully~ tested this PR on albina-website (together with the plugins leaflet-geonames, leaflet-gesture-handling, leaflet.locatecontrol, leaflet.markercluster, leaflet.sync, leaflet.vectorgrid) via https://gitlab.com/albina-euregio/albina-website/-/commit/b3b720f40e5506aebbb3c131383735b0d589d375.

2025-06-15 update: investigating problems regarding `L.SVG.prototype.initialize.call` statements in leaflet.vectorgrid...

src/core/Class.js
lukewarlow: Does it not work to define static options in the classes kinda like before, and then update the setOptions util to look at them? Some variation of the below? Note the obj.constructor?.options is what gets the static options defined on the real classes?

```js
export function setOptions(obj, options) {
	const baseOptions = obj.options || obj.constructor?.options || null;

	if (!Object.hasOwn(obj, 'options')) {
		obj.options = baseOptions ? Object.create(baseOptions) : {};
	}
	for (const i in options) {
		if (Object.hasOwn(options, i)) {
			obj.options[i] = options[i];
		}
	}
	return obj.options;
}
```

lukewarlow:
Fwiw I've also been looking into how best to do this, and despite getting quite far with a slightly different approach in the end I've come to the conclusion that this is the best approach to maintain backwards compat with stuff like changing the options in the prototype (lots of tests do this).

If I can help debug any remaining issues I'm happy to as I think this would be a great change to land.

src/core/Class.js
Falke-Design: The tests in the version from @lukewarlow are all passed. Is this something that we could use instead?
https://github.com/Leaflet/Leaflet/pull/9778/files#diff-8e4854ab0e220475dc22a076300f7f19f24b4480d852e02177dfd05826e4be5d

src/core/Handler.js
```diff
@@ -8,10 +8,11 @@ import {Class} from './Class.js';
 // @class Handler
 // Abstract class for map interaction handlers
 
-export const Handler = Class.extend({
-	initialize(map) {
+export class Handler extends Class {
+	constructor(map, ...args) {
+		super(map, ...args); // for compatibility of code using `Handler.extend`
```
Falke-Design: Why do we not need this for Control?

src/map/Map.js
```diff
@@ -171,8 +174,11 @@ export const Map = Evented.extend({
 		}
 
 		this._addLayers(this.options.layers);
-	},
+	}
 
+	callInitHooks() {
+		// callInitHooks manually once this class is initialized, see constructor
+	}
```
Falke-Design: Why overwritting the `callInitHooks` method? and is `super.callInitHooks();` not calling the parent classes `callInitHooks`?

src/map/Map.js
```diff
@@ -159,7 +162,7 @@ export const Map = Evented.extend({
 			this.setView(new LatLng(options.center), options.zoom, {reset: true});
 		}
 
-		this.callInitHooks();
+		super.callInitHooks();
```
Falke-Design: Is the behavior not different if `callInitHooks` was extended on the map class?

Falke-Design:
We need to update the docs / tutorials how to extend the default options: 
```
class RotateMarker extends Marker{
	static {
		this.setDefaultOptions({
			rotate: true,
		});
	}
};
```

lukewarlow:
> The tests in the version from @lukewarlow are all passed. Is this something that we could use instead?
https://github.com/Leaflet/Leaflet/pull/9778/files#diff-8e4854ab0e220475dc22a076300f7f19f24b4480d852e02177dfd05826e4be5d

Unfortunately that doesn't work for all patterns leaflet currently tests against. Specifically some of the Popup tests related to updating the options on the prototype of Icon don't work with my model. For maximal compatability I believe this approach is best.

Doesn't mean we cant make further changes if we're okay with breaking certain behaviour.

simon04:
To increase compatibility with plugins, I've backtracked from overriding the `constructor`. Now, every subclass of `Class` uses `initialize` again. This resolves a few comments wrt. inconsistencies between `constructor` and `initialize`.

I've updated `docs/examples/extending-1-classes/index.md`.

Successfully tested this PR on albina-website (together with the plugins leaflet-geonames, leaflet-gesture-handling, leaflet.locatecontrol, leaflet.markercluster, leaflet.sync, leaflet.vectorgrid) via https://gitlab.com/albina-euregio/albina-website/-/commit/60362174a14f6ec1e8a3c6b30bb52f33dac08d96 (https://gitlab.com/albina-euregio/albina-website/-/compare/master...leaflet-all-classes).

Falke-Design:
@simon04 thank you! That is a very important change!

simon04:
Awesome, thanks to everyone involved for the review, the feedback etc. I really appreciate the steps leaflet is taking to modernise its codebase.

jonkoops:
@simon04 thanks so much for getting this landed, this is a huge chunk of what I wanted to get landed when I started the original class refactor. Amazing!

PR #9679, opened 2025-04-21
Add demo for all GeoJSON types

### LINKED ISSUE #9533: GeoJSON `MultiLineString` and `MultiPolygon` only render first one of each

simon04:
Closes #9533.

### DIFF
--- a/debug/vector/geojson-types.html
+++ b/debug/vector/geojson-types.html
@@ -1,0 +1,183 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - GeoJSON</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
+		<link rel="stylesheet" href="../../dist/leaflet.css" />
+		<script type="importmap">
+			{
+				"imports": {
+					"leaflet": "../../dist/leaflet-src.js"
+				}
+			}
+		</script>
+	</head>
+	<body>
+		<script type="module">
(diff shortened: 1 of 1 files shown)


debug/vector/geojson-types.html
```diff
@@ -0,0 +1,183 @@
...
+
+			for (const feature of features) {
+				document.body.appendChild(document.createTextNode(feature.type));
+				const div = document.createElement('div');
+				div.style.width = '800px';
+				div.style.height = '600px';
+				document.body.appendChild(div);
+				const map = new Map(div).setView([0.0, 0.0], 4);
+				new TileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png').addTo(map);
+				new GeoJSON(feature).addTo(map)
```
Falke-Design, line 177:
```suggestion
				const geojson = new GeoJSON(feature).addTo(map)
				map.fitBounds(geojson.getBounds());
```

Falke-Design (approved):

PR #9680, opened 2025-04-21
Remove `LayerGroup.invoke` in favour of existing `LayerGroup.eachLayer`

simon04:


Falke-Design:
The only occurrence that I found: https://github.com/search?q=repo%3Asimone-f%2Fwikipedia-tags-in-osm%20invoke&type=code

`invoke` was added with this commit: https://github.com/Leaflet/Leaflet/commit/e50c98863c82b034ced024103f24b92543ae1781

I'm just curious why @mourner added `invoke` when `eachLayer` already existed (`_iterateLayers`)

simon04:
@Falke-Design, @jonkoops, is this PR something to consider for version 2?

jonkoops:
Removing invoke seems fine to me, it is quite a duplicated effort, the only reason I can think of to keep it is for compatibility. I agree that `eachLayer()` can basically accomplish the same.

Issue #9644, opened 2025-04-13
L.noConflict isn't supported in V2 anymore


Falke-Design:
We define `L` in the global object, should we then not still provide `L.noConflict`?

https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/LeafletWithGlobals.js#L4-L15

@jonkoops

simon04:
This feature seems to be obsolete with the module endpoint provided by `src/Leaflet.js` ...?

Issue #9465, opened 2024-09-06
Leaflet 2.x: Backwards compatibility helper for V1


indigoxela:
### Checklist

- [X] I've searched through the [plugins](https://leafletjs.com/plugins.html) to make sure this feature isn't already available, or think it shouldn't require a plugin.
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this feature hasn't been requested already.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Motivation

Finally I found the time to play with Leaflet 2.0-dev. (For others - download can be found here: https://leafletjs.com/download.html)

Hey, it's not all that bad. Of course, I don't have info, which tasks you folks still consider crucial for an alpha (beta, rc) release. But 2.0-dev seems fully functional (standalone).

But ... and that's a huge BUT, every single plugin I use seems to break and with it the map.

`Uncaught TypeError: L.foobarWhatever is not a function`. That's because of the cleanup of (technically obsolete) utility methods in L.DomUtil, L.Util. Every single plugin I know uses at least one of them, usually more than one.

### Suggested solution

However, what if you add some sort of compatibility helper (wrappers around native JS functionality, not the IE cruft, of course).
That might make it easier to get people adopting the new version. Eventually also easier to release 2.0 in the first place. Seems like a win-win to me.

Here an example: With following plugins enabled, I only need 19 lines of code to get everything working with 2.x.

- L.Control.Zoomslider
- Leaflet.fullscreen
- leaflet.markercluster

### Alternatives considered

Yes, obviously I can fix things on my side (an integration for a CMS) by adding that compat layer myself.
But isn't it actually worth to at least consider a compat helper in Leaflet?
Be it in the main library or as a sub project (other repo)...

BTW, I'd fully understand if you decide against that. But keep in mind, that lots of plugins - how to phrase it - aren't maintained too enthusiastically. :stuck_out_tongue_winking_eye: 
Not only the more exotic ones, but also some with higher usage counts.

Falke-Design:
I created https://github.com/Falke-Design/Leaflet-V1-polyfill to make the switch to V2 easier. I hope every change is included - let it grow together with the alpha releases (hopefully in the next weeks) 🚀

jonkoops:
Very nice!

Falke-Design:
[...]

PR #9686, opened 2025-04-24
Bring back global `L` as new bundle

Falke-Design:
To provide better backward compatibility, this adds a bundle which adds Leaflet to the global `L` variable. It wasn't possible to use Leaflet without importing it in a `module`.

New bundle: `leaflet-global.js`, `leaflet-global-src.js`

Closes: #9644
Closes: #9687
See: https://github.com/Leaflet/Leaflet/issues/9465#issuecomment-2824448407

Or should we name it `leaflet-umd.js`?


Falke-Design:

docs/download.md
```diff
@@ -66,6 +66,13 @@ Then, in your script, import the needed Leaflet Classes as follows:
 	}).addTo(map);
 </script>
 ```
+#### Including Leaflet with a Global Scope
+
+The old (and not recommended) way to include Leaflet in your project without using modules is by relying on the global variable `L`.
```
IvanSanchez, line 71: Nitpick: replace "not recommended" by "no longer recommended".

IvanSanchez (approved):

jonkoops:
Late to the party here, but this all lgtm. I do believe 'umd' would be better personally, as it also allows loading the module as CommonJS, or AMD, which folks might still rely on. However, no big deal.

PR #9688, opened 2025-04-25
Add aria-keyshortcuts to map container

chcederquist:
Improves keyboard accessibility by letting assistive technology know which key shortcuts are available through aria-keyshortcuts

### DIFF
--- a/src/map/handler/Map.Keyboard.js
+++ b/src/map/handler/Map.Keyboard.js
@@ -47,6 +47,9 @@ export const Keyboard = Handler.extend({
 			container.tabIndex = '0';
 		}
 
+		// add aria-attribute for keyboard shortcuts to the container
+		container.ariaKeyShortcuts = Object.values(this.keyCodes).flat().join(' ');
+
 		on(container, {
 			focus: this._onFocus,
 			blur: this._onBlur,


src/map/handler/Map.Keyboard.js
```diff
@@ -47,6 +47,13 @@ export const Keyboard = Handler.extend({
 			container.tabIndex = '0';
 		}
 
+		// add aria-attribute for keyboard shortcuts to the container
+		let keyShortcuts = [];
+		for (const key in this.keyCodes) {
+			keyShortcuts = keyShortcuts.concat(this.keyCodes[key]);
+		}
+		container.ariaKeyShortcuts = keyShortcuts.join(' ');
```
Falke-Design, line 55: We can simplify this:
```suggestion
		container.ariaKeyShortcuts = Object.values(this.keyCodes).flat().join(' ') ;
```

Falke-Design: If we would like to make sure that no shortcut exsits multiple times, we could wrap it inside a `Set` but I think we have a differnet problem, if a keybinding is used for different commands.

`[...new Set(Object.values(keyCodes).flat())].join(' ') `

Falke-Design (changes requested):
Thank you, just the small change

src/map/handler/Map.Keyboard.js
chcederquist: I've always preferred the for...in-syntax, but I'll convert it to the one-liner. I agree the set may be overkill for something that probably won't happen, and in my opinion the one-liner starts will get too terse if it does five things at once.

chcederquist: I've implemented the change and squashed the commits

Falke-Design (approved):

Falke-Design:
Thank you for your contribution. We would have many other https://github.com/Leaflet/Leaflet/labels/accessibility issues / PRs where we need help 😉

PR #9701, opened 2025-05-03
Improve how Util.setOptions workks to ensure keys with explicit undefined values are skipped

### LINKED ISSUE #9700: Title: TypeError when icon option is undefined in L.marker (Leaflet 1.9.4)
### Checklist

- [ ] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.

### Steps to reproduce

Create map and add marker 

```ts
 let mapInstance: null | Leaflet.Map = null;
 mapInstance = Leaflet.map(mapField.value, {
      attributionControl: false,
      zoomControl: false,
    });

function renderMaker(point) {
  Leaflet.marker([point.lat, point.lng], {
          icon: undefined,
        }).addTo(mapInstance!);
}

```

### Expected behavior

Add markers with default icons

### Current behavior

Uncaught TypeError: Cannot read properties of undefined (reading 'createIcon')

_No response_

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Firefox 138.0
- OS/Platform (with version): Windows 11

MelvinManni:
closes #9700

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -58,8 +58,8 @@ export function throttle(fn, time, context) {
 // `range[1]` unless `includeMax` is set to `true`.
 export function wrapNum(x, range, includeMax) {
 	const max = range[1],
-	    min = range[0],
-	    d = max - min;
+	min = range[0],
+	d = max - min;
 	return x === max && includeMax ? x : ((x - min) % d + d) % d + min;
 }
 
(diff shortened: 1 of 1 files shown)


Falke-Design (changes requested):
Thank you for your contribution but with that change it would be not possible to overwrite a existing option with `undefined` anymore.

The better solution is to set the icon, because it is always needed. If it is undefined you can fallback to the Default Icon:

```
  Leaflet.marker([point.lat, point.lng], {
          icon: undefined || new L.Icon.Default(),
        }).addTo(mapInstance!);
```

Issue #9710, opened 2025-05-06
Replace / remove deprecated / obsolete methods


Falke-Design:
I think `touchNative` and `touch` can be removed and set always to `true` (needs to be checked):

https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/map/handler/Map.TapHold.js#L20
https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/map/handler/Map.TouchZoom.js#L19
https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/map/Map.js#L1112

Orientation is deprecated:
https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/core/Browser.js#L23

Platform is deprecated: (Replacement is not so easy)
https://github.com/Leaflet/Leaflet/blob/18ba00aee31737227430b705f8a9a8e1f116f34d/src/core/Browser.js#L46

Tosin-JD:
Please, this is my first contribution to an open-source project. Thank you for the opportunity.

Tosin-JD:
I have switched the tapHold to be true for ios and used the ? operator for the userAgentData.
userAgentData is not yet supported on non Chromium browser  [navigator.userAgentData - MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData) that is why I chose to use the userAgent. Could there be a better approach of handling it?

Amir-Golmoradi:
Hi, I’d like to take this issue!

I’m a **JavaScript / TypeScript** developer, and this will be my first contribution to Leaflet and open source in general. I’ve read the contributing guidelines and I’m setting up the project locally.

Let me know if there’s anything else I should keep in mind before I get started.

Thanks!

Falke-Design:
@Amir-Golmoradi There is already an open PR (#9711) for this issue. Please search for another one without a existing PR.

iammoizzah:
Hi, can I take this issue for Hacktoberfest?

jonkoops:
@Falke-Design looks like this one might already be closed? I see the related PR was merged, but never linked to this issue.

Falke-Design:
@jonkoops no not all issues are fixed. I added a checklist

jonkoops:
[...]

PR #9711, opened 2025-05-07
fix: Replace / remove deprecated / obsolete methods #9710

Tosin-JD:
This is my first time contributing to a public repository and I tried to address issue Fixes partially: #9710

Thank you for the oppportunity.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -19,7 +19,7 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
+const mobile = typeof Screen.orientation !== 'undefined' || userAgentContains('mobile');
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
@@ -42,10 +42,10 @@ const touch = touchNative || pointer;
 const retina = typeof window === 'undefined' || typeof window.devicePixelRatio === 'undefined' ? false : window.devicePixelRatio > 1;
 
 // @property mac: Boolean; `true` when the browser is running in a Mac platform
-const mac = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Mac');
+const mac = typeof navigator === 'undefined' || typeof navigator.userAgentData.platform === 'undefined' ? false : navigator.userAgent.indexOf('Mac') !== -1;
 
 // @property mac: Boolean; `true` when the browser is running in a Linux platform
-const linux = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Linux');
+const linux = typeof navigator === 'undefined' || typeof navigator.userAgentData.platform === 'undefined' ? false : navigator.userAgent.indexOf('Linux') !== -1;
 
 function userAgentContains(str) {
 	if (typeof navigator === 'undefined' || typeof navigator.userAgent === 'undefined') {
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -19,7 +19,7 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof Screen.orientation !== 'undefined' || userAgentContains('mobile');
+const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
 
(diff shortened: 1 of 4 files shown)


src/map/handler/Map.TapHold.js
```diff
@@ -18,7 +17,7 @@ Map.mergeOptions({
 	// @section Touch interaction options
 	// @option tapHold: Boolean
 	// Enables simulation of `contextmenu` event, default is `true` for mobile Safari.
-	tapHold: Browser.touchNative && Browser.safari && Browser.mobile,
+	tapHold: true,
```
IvanSanchez, line 20: This should be true **only** when running on iOS, not always true.

IvanSanchez (changes requested):

src/core/Browser.js
```diff
@@ -20,7 +20,7 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
+const mobile = typeof Screen.orientation !== 'undefined' || userAgentContains('mobile');
```
Falke-Design, line 23:
```suggestion
const mobile = typeof screen.orientation !== 'undefined' || userAgentContains('mobile');

```

Don't just copy paste the first thing that looks like it fits ... and it can be expected that you test before you create a PR.

1.  it is `screen.orientation` and not `Screen.orientation`
2. `screen.orientation` is also on desktop browsers supportet, `orientation` is only on mobile browsers available -> `typeof screen.orientation !== 'undefined'` will be always true

https://developer.mozilla.org/en-US/docs/Web/API/Screen/orientation
https://developer.mozilla.org/en-US/docs/Web/API/Window/orientation

src/core/Browser.js
```diff
@@ -43,10 +43,10 @@ const touch = touchNative || pointer;
 const retina = typeof window === 'undefined' || typeof window.devicePixelRatio === 'undefined' ? false : window.devicePixelRatio > 1;
 
 // @property mac: Boolean; `true` when the browser is running in a Mac platform
-const mac = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Mac');
+const mac = typeof navigator === 'undefined' || typeof navigator?.userAgentData.platform === 'undefined' ? false : navigator.userAgent.indexOf('Mac') !== -1;
```
Falke-Design, line 46: Why testing if `navigator?.userAgentData.platform` is undefined when later `navigator.userAgent` is used?

src/core/Browser.js
```diff
@@ -43,10 +43,10 @@ const touch = touchNative || pointer;
 const retina = typeof window === 'undefined' || typeof window.devicePixelRatio === 'undefined' ? false : window.devicePixelRatio > 1;
 
 // @property mac: Boolean; `true` when the browser is running in a Mac platform
-const mac = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Mac');
+const mac = typeof navigator === 'undefined' || typeof navigator?.userAgentData.platform === 'undefined' ? false : navigator.userAgent.indexOf('Mac') !== -1;
 
 // @property mac: Boolean; `true` when the browser is running in a Linux platform
-const linux = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined' ? false : navigator.platform.startsWith('Linux');
+const linux = typeof navigator === 'undefined' || typeof navigator?.userAgentData.platform === 'undefined' ? false : navigator.userAgent.indexOf('Linux') !== -1;
```
Falke-Design, line 49: Why testing if `navigator?.userAgentData.platform` is undefined when later `navigator.userAgent` is used?

Falke-Design (changes requested):

src/core/Browser.js
Tosin-JD: I have replace the check for `navigator?.userAgentData.platform` with the check for `navigator.userAgent`.

Falke-Design (approved):

src/map/handler/Map.PinchZoom.js
```diff
@@ -17,7 +16,7 @@ Map.mergeOptions({
 	// passed `'center'`, it will zoom to the center of the view regardless of
 	// where the touch events (fingers) were. Enabled for touch-capable web
 	// browsers.
```
jonkoops, line 18: This comment is now out of date, as this feature is now enabled by default everywhere.

src/map/Map.js
```diff
@@ -1117,9 +1117,8 @@ export class Map extends Evented {
 
 		this._fadeAnimated = this.options.fadeAnimation;
 
-		const classes = ['leaflet-container'];
+		const classes = ['leaflet-container', 'leaflet-touch'];
```
jonkoops, line 1120: If `leaflet-touch` is now always enabled, we should roll the CSS that exists there up into the default styles. No need to separate them any longer.

PR #9717, opened 2025-05-13
Replace `version` when `npm version` is called

Falke-Design:
With this change, the static `version` variable in `Leaflet.js` will be replaced automatically with the version of `package.json` everytime `npm version` is called.

With this solution Firefox versions before May can use the non-bundled code too. They didn't support json import: `import pkg from '../package.json' with { type: "json" };`

Solved issues with this approach:
* It is possible to use the non-bundled code and we have always included the correct version instead of `latest` or some other placeholder
* We have a static version instead of importing the version from `package.json` where the whole file was included. Named export of only the version is not possible.
     > See https://github.com/Leaflet/Leaflet/pull/9609#issuecomment-2817039502. At the moment, the entire content of package.json ends up in the dist bundle.
* We keep the git hash in the banner
* We can remove the karma preprocessor

Looks like this then:
grafik

Supersedes #9657, #9662

@simon04 please review too


build/version.js
```diff
@@ -0,0 +1,21 @@
...
+const fileContent = readFileSync(new URL('../src/Leaflet.js', import.meta.url), 'utf8');
+
+const newContent = fileContent.replace(
+	/(\/\/ !!! NEXT LINE IS AUTO-GENERATED VIA `NPM VERSION` !!!\s+export const version = ')[^']+(')/,
+	`$1${pkg.version}$2`
+);
+
+if (newContent === fileContent) {
+	console.error('Version replacement failed: pattern in Leaflet.js not found.');
+	process.exit(1); // Exit with non-zero status code
```
Falke-Design, line 16:
```suggestion
	process.exit(1);
```

build/version.js
```diff
@@ -0,0 +1,21 @@
+import {readFileSync, writeFileSync} from 'node:fs';
+
+// TODO: Replace this with a regular import when ESLint adds support for import assertions.
+// See: https://rollupjs.org/guide/en/#importing-packagejson
+const pkg = JSON.parse(readFileSync(new URL('../package.json', import.meta.url)));
+
+const fileContent = readFileSync(new URL('../src/Leaflet.js', import.meta.url), 'utf8');
+
+const newContent = fileContent.replace(
+	/(\/\/ !!! NEXT LINE IS AUTO-GENERATED VIA `NPM VERSION` !!!\s+export const version = ')[^']+(')/,
```
simon04, line 10: Do we want to also support `"` (in case we ever switch quoting style)?

```suggestion
	/(\/\/ !!! NEXT LINE IS AUTO-GENERATED VIA `NPM VERSION` !!!\s+export const version = ['"])[^'"]+(['"])/,
```

Falke-Design: Thanks for the suggestion but I think we keep it simple. If the replacement isn't working then the command will fail.

PR #9720, opened 2025-05-18
Plugin leaflet-control-geocoder is compatible with leaflet 2.0

simon04:
https://github.com/perliedman/leaflet-control-geocoder/releases/tag/3.2.0


Falke-Design (approved):

jonkoops:

PR #9721, opened 2025-05-18
Fix Leaflet Docs grammatical error

tinchomengo:
This PR corrects a fragment of Leaflet's official documentation available at https://leafletjs.com/reference.html#map-example.

Specifically, it targets the _Mouse Wheel Options_ section, where the article for the term `user` is missing, by adding the proper comma and article ('`the`') before the term `user`.

**Current Version Online:**
_Limits the rate at which a wheel can fire (in milliseconds). By default user can't zoom via wheel more often than once per 40 ms._

**Fixed Version:**
_Limits the rate at which a wheel can fire (in milliseconds). By default, the user can't zoom via wheel more often than once per 40 ms._

### DIFF
--- a/docs/reference-2.0.0.html
+++ b/docs/reference-2.0.0.html
@@ -455,7 +455,7 @@ it will zoom to the center of the view regardless of where the pointer was.</td>
 		<td><code><b>wheelDebounceTime</b></code></td>
 		<td><code>Number</code></td>
 		<td><code>40</code></td>
-		<td>Limits the rate at which a wheel can fire (in milliseconds). By default
+		<td>Limits the rate at which a wheel can fire (in milliseconds). By default, the
 user can't zoom via wheel more often than once per 40 ms.</td>
 	</tr>
 	<tr id='map-wheelpxperzoomlevel'>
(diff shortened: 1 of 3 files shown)


docs/reference.html
```diff
@@ -414,7 +414,7 @@ <h4 id='map-mouse-wheel-options'>Mouse wheel options</h4>
 		<td><code><b>wheelDebounceTime</b></code></td>
 		<td><code>Number</code></td>
 		<td><code>40</code></td>
-		<td>Limits the rate at which a wheel can fire (in milliseconds). By default
+		<td>Limits the rate at which a wheel can fire (in milliseconds). By default, the
```
Falke-Design, line 417:
```suggestion
		<td>Limits the rate at which a wheel can fire (in milliseconds). By default
```

docs/reference-2.0.0.html
```diff
@@ -455,7 +455,7 @@ <h4 id='map-mouse-wheel-options'>Mouse wheel options</h4>
 		<td><code><b>wheelDebounceTime</b></code></td>
 		<td><code>Number</code></td>
 		<td><code>40</code></td>
-		<td>Limits the rate at which a wheel can fire (in milliseconds). By default
+		<td>Limits the rate at which a wheel can fire (in milliseconds). By default, the
```
Falke-Design, line 458:
```suggestion
		<td>Limits the rate at which a wheel can fire (in milliseconds). By default
```

Falke-Design (approved):

PR #9727, opened 2025-05-19
Add readme, changelog, license to S3 artifacts

### LINKED ISSUE #9722: License file missing from official 2.x downloads
### Checklist

### Steps to reproduce

It might be on purpose, but I guess it's rather an oversight in the build process: v1.x used to also contain a LICENSE file.

But apparently v2 downloads from https://leafletjs.com/download.html don't contain any license info. So, nothing in the download contains any hint on BSD 2-Clause License.

### Expected behavior

In a download of any Open Source project I'd expect license info of some kind.

### Current behavior

I'd have to go to the repo on Github to verify, but there's not even a link to the repo...

build specific, no code example

### Environment

- Leaflet version: 2.x

IvanSanchez:
Should fix #9722

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -178,7 +178,7 @@ jobs:
 
       - name: Compress artifacts
         working-directory: dist
-        run: zip -r leaflet.zip .
+        run: zip -r leaflet.zip . ../CHANGELOG.md ../LICENSE ../README.md
 
       - name: Publish artifacts
         uses: jakejarvis/s3-sync-action@v0.5.1


Falke-Design (approved):
Looks good but i haven't tested it.

You mentioned in the issue the `package.json`, which is now not included

IvanSanchez:
> You mentioned in the issue the `package.json`, which is now not included

...because `package.json` references `src` and `dist/`, but the zipfile only includes the stuff in `dist/` at the root level.

I'm not entirely sure how that should be handled, if at all. Perhaps the NPM release should be enough for the use cases where `package.json` is important?

Falke-Design:
> I'm not entirely sure how that should be handled, if at all. Perhaps the NPM release should be enough for the use cases where package.json is important?

Fine for me

PR #9732, opened 2025-05-22
TileStitcher class (anti-gap TileLayer)

### LINKED ISSUE #3575: Space between tiles on fractional zoom levels in Webkit browsers

IvanSanchez:
My approach to fix #3575 once and for all.

This is basically #5290, but adapted to the 2.0 codebase.

While @mourner's comment at https://github.com/Leaflet/Leaflet/pull/5290#issuecomment-286386260 kinda still stands, IMO the 1-px gap has been a PITA for long enough, none of the hacky workarounds were good, and 2.0 is a good excuse to get this in.

Still a separate class since it adds ~250 lines of code that not all users might want to have.

TODO: [ ] links to the new class in the API documentation HTML.

### DIFF
--- a/debug/tests/tile-stitcher.html
+++ b/debug/tests/tile-stitcher.html
@@ -1,0 +1,204 @@
+<!doctype html>
+<html>
+	<head>
+		<meta charset="utf-8" />
+		<title>Leaflet debug page - Tile Events</title>
+		<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0" />
+		<link rel="stylesheet" href="../../dist/leaflet.css" />
+		<link rel="stylesheet" href="../css/screen.css" />
+		<style>
+			#map {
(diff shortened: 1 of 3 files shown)


jonkoops:
@IvanSanchez would love to give this an in-depth review, but I am a bit tied up for the next alpha. Seems like a nice addition though!

PR #9737, opened 2025-05-29
Use optional chaining with function calls

simon04:
Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls


Falke-Design (approved):

Falke-Design:
Thank you

Issue #9718, opened 2025-05-13
Popup width changed with v2


Falke-Design:
PR #9605 caused that the Popup is rendered differently:

V1:
![grafik](https://github.com/user-attachments/assets/7865c7ba-b34a-462e-a302-2ca5a4d17dbd)

V2:
![grafik](https://github.com/user-attachments/assets/fc4edbe9-48b0-4ec2-8928-478936ce833f)

@IvanSanchez I'm not sure if this will cause more problems than it solves for users who want to migrate.

_Originally posted by @Falke-Design in https://github.com/Leaflet/Leaflet/issues/9605#issuecomment-2869542944_

Falke-Design:
@jonkoops @IvanSanchez @mourner should we revert the PR #9605 or do we want to keep it like that?

IvanSanchez:
"It's not a bug, it's a feature". Or, in other words, now the minimum width of popups is behaving as it should have been.

Setting the min width at 100px instead of 50px (#9728) should do the trick.

mourner:
@IvanSanchez Personally I think this is a problem — I remember spending a lot of time back in the day making sure this case of popup size for one-line labels just works without additional tuning, because it's such a common case. If we are changing the default behavior people have come to rely on for a long time, we need to either have a _very good justification_ for this (which I might be missing, but it hasn't been explained), or revert. 

> "It's not a bug, it's a feature". Or, in other words, now the minimum width of popups is behaving as it should have been.

Why?

IvanSanchez:
>> "It's not a bug, it's a feature". Or, in other words, now the minimum width of popups is behaving as it should have been.
>
> Why?

Because we're letting CSS do its thing instead of reflowing content, getting the width, clamping, then reflowing the content *again*.

The bit that I overlooked is that the leaflet panes are set to zero width - so that any flow content inside a pane will try to be as narrow as possible instead of trying to fill up its `max-width`. #9765 addresses this.

mourner:
Looks like this got fixed by #9765?

PR #9765, opened 2025-06-27
CSS: set panes to 100% width so that popups fill max-width

IvanSanchez:
Addresses #9718


Falke-Design (approved):
Now we could revert https://github.com/Leaflet/Leaflet/commit/a95c2a130301a5c5078d9b4198fc846c061fddcd too

Falke-Design (approved):

PR #9777, opened 2025-07-04
Convert Evented and Handler to es6 classes

lukewarlow:
- Also throw an exception in Handlers constructor if the hook functions aren't implemented.

The plan if it's desirable is to do this across the board with all the Foo.extend({}) defined classes within leaflet. Potentially dropping that extension method entirely.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -25,7 +25,7 @@ import * as Util from './Util.js';
  * ```
  */
 
-export const Events = {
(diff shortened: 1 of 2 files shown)


lukewarlow:
Ah I just found https://github.com/Leaflet/Leaflet/pull/9677 happy to close and defer to that but perhaps it's easier to do this migration in smaller chunks?

PR #9783, opened 2025-07-04
Replace Point/LatLng.clone with copy constructor

simon04:
Fixes eslint no-constructor-return.

https://exploringjs.com/deep-js/ch_copying-class-instances.html

### DIFF
--- a/spec/suites/geo/LatLngSpec.js
+++ b/spec/suites/geo/LatLngSpec.js
@@ -120,10 +120,10 @@ describe('LatLng', () => {
 		});
 	});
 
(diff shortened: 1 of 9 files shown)


Falke-Design (commented):
I tried to do the same before but it made to much problems with our own codebase and also I think multiple plugins relay on it especially on `clone()`.

For me it is better to leave it as it is now

jonkoops:
+1, I don't think we should be removing API surface that users might rely on, unless there are good reasons to do so. Either way, we should first deprecate such APIs and add them to the 1.x release.

Issue #9792, opened 2025-07-08
Mobile example with v2.0.0-alpha fails to display accuracy-based radius


radhikalism:
### Checklist

- [x] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [x] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

1. Use the v2.0.0-alpha build.
2. Set up the [mobile example](https://github.com/Leaflet/Leaflet/blob/v2.0.0-alpha/docs/examples/mobile/example.md?plain=1).

### Expected behavior

The location pin popup description in the mobile example should be a string with a numerical radius like, "You are within 5.123 meters from this point".

### Current behavior

The location pin popup description is the string, "You are within NaN meters from this point".

### Environment

- Leaflet version: 2.0.0-alpha
- Browser (with version): Firefox 141.0b5 or Chrome 136.0.7103.113
- OS/Platform (with version): Debian 13

simon04:
Fixed via https://github.com/Leaflet/Leaflet/pull/9793?

Falke-Design:
Solved via #9793

PR #9793, opened 2025-07-08
Handle GeolocationCoordinates instance values in #locate

radhikalism:
Per #9792: The `locationfound` event data should include properties from the original [`GeolocationCoordinates`](https://developer.mozilla.org/en-US/docs/Web/API/GeolocationCoordinates) instance such as `accuracy`, when fired through `locate`.

In v2.0.0-alpha, all `GeolocationCoordinates` properties except `latitude` and `longitude` are missing in the `locationfound` event data.

As a result, the updated [mobile example](https://github.com/Leaflet/Leaflet/blob/v2.0.0-alpha/docs/examples/mobile/example.md?plain=1) does not fully work. It shows a location description like, "You are within NaN meters from this point" because `e.accuracy` is `undefined`.

### Cause

In v1.9.4, these properties are included in the `locationfound` event data because `_handleGeolocationResponse` copies them with a `for … in` loop. https://github.com/Leaflet/Leaflet/blob/d15112c9e8ac339f0f74f563959d0423d291308d/src/map/Map.js#L709

In v2.0.0-alpha, these properties are lost because `_handleGeolocationResponse` tries to copy them with a `for … of Object.keys()` loop. https://github.com/Leaflet/Leaflet/blob/00aa415832de9a7d45634291f9522856e56617a8/src/map/Map.js#L713

The `for … of Object.keys()` form does not read getter function properties defined on the prototype. But the properties of `GeolocationCoordinates` are all getter function properties.

The `latitude` and `longitude` properties do get preserved in both versions, because they are explicitly copied by name (rather than the copying loop). https://github.com/Leaflet/Leaflet/blob/00aa415832de9a7d45634291f9522856e56617a8/src/map/Map.js#L708

### Test coverage

The regression does not get detected in the specs because the `geolocationStub` `coords` in `MapSpec.js` happen to be a plain object with string-keyed own property names, rather than a mock Geolocation object instance with getters. https://github.com/Leaflet/Leaflet/blob/00aa415832de9a7d45634291f9522856e56617a8/spec/suites/map/MapSpec.js#L2302

### PR

As a proof of concept solution, this PR modifies:

1. `Map.js` `_handleGeolocationResponse` to copy getter function properties from `pos.coords` if any, as well as own object keys if any.
2. `MapSpec.js` `geolocationStub` `coords` to mock a `GeolocationCoordinates` object with getter function properties.
3. `MapSpec.js` `locate` tests to also verify `accuracy` gets passed through correctly (unlike `latitude` and `longitude` that are special).

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -2334,10 +2334,9 @@ describe('Map', () => {
 		const geolocationStub = {
 			geolocation: {
 				getCurrentPosition(onSuccess) {
-					const x = new MockGeolocationCoordinates({latitude: 50, longitude: 50, accuracy: 14});
 					onSuccess(
 						{
-							coords: x,
+							coords: new MockGeolocationCoordinates({latitude: 50, longitude: 50, accuracy: 14}),
 							timestamp: 1670000000000
 						});
 


Falke-Design:
@simon04 please take a look

src/map/Map.js
```diff
@@ -710,7 +710,18 @@ export const Map = Evented.extend({
...
+		const getterProps = Object.entries(descriptors)
+			.filter(e => ((typeof e[1].get === 'function' && e[0] !== '__proto__')))
+			.map(e => e[0]);
+
+		// Collect own properties from plain object coordinates:
+		const ownKeys = Object.keys(pos.coords);
+
+		const props = [].concat(getterProps, ownKeys);
+
+		for (const i of props) {
```
simon04: Thanks for catching this regression and updating the unit tests to cover

Simply fallback to the `for..in` loop which was used before 93509f310da0e3e15a6ca5b4b85da4c8095d339c:

```suggestion
		for (const i in pos.coords) { // do not use Object.keys here to access getters of GeolocationCoordinates
```

simon04 (approved):

radhikalism:
Updated the mock instance definition in tests to better imitate a real `GeolocationCoordinates` object, to pass tests when using `for…in` to traverse its properties.

PR #9806, opened 2025-07-25
Stop uploading builds to S3; switch to NPM registry for downloads

mourner:
For a long time, we've had Leaflet upload the build to S3 on every commit and release. However, this isn't really necessary — the releases can be downloaded from NPM or CDN, and if you need the `main` build as a contributor, you can do it from the repo. 

The upload action and reliance on S3 adds maintenance complexity — sometimes the upload breaks due to some S3 misconfigurations, sometimes I need to rotate access tokens because of some security policies, etc. So ideally, it would be great to just remove it, simpifying things.

The only thing I have reservations about is supporting `?version=main` on our Edit page — does anyone really rely on this? This seems fragile as it's not versioned, so e.g. any bug demonstration might quickly get outdated.

Another small gotcha is that TGZ (which NPM exposes) is less universal than ZIP, e.g. you can't unarchive it on Windows with built-in tools, only with apps like 7Zip (which developers always install on their machines anyway).

@Falke-Design @IvanSanchez what do you think?

### DIFF
--- a/docs/download.md
+++ b/docs/download.md
@@ -12,11 +12,11 @@ bodyclass: download-page
 		<th>Description</th>
 	</tr>
 	<tr>
-		<td><a href="https://registry.npmjs.org/leaflet/-/leaflet-1.9.4.tgz">Leaflet 1.9.4</a></td>
+		<td><a href="https://github.com/Leaflet/Leaflet/releases/download/v1.9.4/leaflet.zip">Leaflet 1.9.4</a></td>
 		<td>Stable version, released on May 18, 2023.</td>
 	</tr>
 	<tr>
-		<td><a href="https://registry.npmjs.org/leaflet/-/leaflet-2.0.0-alpha.tgz">Leaflet 2.0.0-alpha</a></td>
+		<td><a href="https://github.com/Leaflet/Leaflet/releases/download/v2.0.0-alpha/leaflet.zip">Leaflet 2.0.0-alpha</a></td>
 		<td>Prerelease version, released on May 18, 2025.</td>
 	</tr>
 </table>


docs/download.md
```diff
@@ -12,17 +12,13 @@ bodyclass: download-page
 		<th>Description</th>
 	</tr>
 	<tr>
-		<td><a href="https://leafletjs-cdn.s3.amazonaws.com/content/leaflet/v1.9.4/leaflet.zip">Leaflet 1.9.4</a></td>
+		<td><a href="https://registry.npmjs.org/leaflet/-/leaflet-1.9.4.tgz">Leaflet 1.9.4</a></td>
```
Falke-Design: Instead of providing the download as tgz via npm, we could use the release zip-asset: 

```suggestion
		<td><a href="https://github.com/Leaflet/Leaflet/releases/download/v1.9.4/leaflet.zip">Leaflet 1.9.4</a></td>
```

Falke-Design (commented):
> The only thing I have reservations about is supporting ?version=main on our Edit page — does anyone really rely on this?

No I don't think so, in the last years I never saw someone using it.

> Another small gotcha is that TGZ (which NPM exposes) is less universal than ZIP

If possible I would like to keep `zip`. 

We could make a manually triggerable workflow, which generates the zip for us. We can then download it and add it to the release.

The workflow could look like that:
```
name: Build ZIP

on:
  workflow_dispatch: # run manually

jobs:
  build-zip:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repository
        uses: actions/checkout@v4

      - name: Strip leading "v" from tag
        run: |
          RAW_VERSION="${{ github.ref_name }}"
          echo "PACKAGE_VERSION=${RAW_VERSION#v}" >> $GITHUB_ENV
          echo "Version: $PACKAGE_VERSION"

      - name: Download npm tarball
        run: |
          TARBALL_URL="https://registry.npmjs.org/leaflet/-/leaflet-${{ env.PACKAGE_VERSION }}.tgz"
          echo "Downloading $TARBALL_URL"
          curl -L "$TARBALL_URL" -o leaflet.tgz

      - name: Extract tarball
        run: |
          mkdir leaflet
          tar -xzf leaflet.tgz -C leaflet --strip-components=1

      - name: Upload ZIP as artifact
        uses: actions/upload-artifact@v4
        with:
          name: leaflet-${{ env.PACKAGE_VERSION }}
          path: leaflet
          retention-days: 1
```
grafik

grafik

mourner:
> We could make a manually triggerable workflow, which generates the zip for us. We can then download it and add it to the release.

In theory, we could go further — add a workflow that runs on creation of a GitHub release and attaches a zip file to it automatically. But let's handle that in a separate PR.

PR #9818, opened 2025-08-05
Update WMS QGIS tutorial link to updated tutorial

user1234567810:
The [previous link](https://www.qgistutorials.com/en/docs/working_with_wms.html) led to an obsolete version of the tutorial. This is the URL for the [updated tutorial](https://www.qgistutorials.com/en/docs/3/working_with_wms.html).


Falke-Design:
Thank you

PR #9837, opened 2025-08-16
Fixed the  Circle.setStyle updates radius like CircleMarker (#9)

### LINKED ISSUE #9: ability to pass HTMLElement instead of String in L.Popup constructor

vishwa123791:
Circle.setStyle now updates the radius when {radius: ...} is passed,
making it consistent with CircleMarker.setStyle.
Fixes #9.


Falke-Design:
Thank you for your PR but we will not accept KI generated content.

Issue #9119, opened 2023-10-08
The `leaflet-container` needs an accessible name


nfreear:
### Checklist

- [X] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [X] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [X] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [X] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

## Problem

Every interactive and focusable element needs a [programmatically determinable](https://www.w3.org/TR/WCAG22/#dfn-programmatically-determinable) name, role and value. Without a name and role, it is hard for a user of assistive technology to understand how to interact with it.

Currently, the accessible name is composed of the contents of the map container, which in a map with lots of markers or similar makes it hard to understand.

As an example, a map with one marker just using defaults has the name, "_Marker Zoom in Zoom out Leaflet © OpenStreetMap_".

![Map with a poor accessible name](https://github.com/Leaflet/Leaflet/assets/131235/deef5c61-203c-49af-bc38-65e83524cbcd)

* WCAG  2.2 Success Criteria: [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG22/#name-role-value)
* Users affected: Vision (users of assistive technology, screen readers).

## Proposed Solution

Give the `leaflet-container` an appropriate accessible name, by setting:

1. Use a sensible default for the `name` ("Map"),
2. Give the map implementer an option to set the `name`,
3. If `aria-label` is not set, then`aria-label="Map"`

### Options

`src/map/Map.js`:
```js
options: {
  // ...

  // @option name: String = 'Map'
  // Short text for the `aria-label` attribute of the map container.
  // [Useful for accessibility](https://leafletjs.com/examples/accessibility/).		
  name: 'Map',

  // ...
},
```

### Implementation

`src/map/Map.js`:
```js
_initLayout () {
  // ...

  if (!container.hasAttribute('aria-label')) {
    container.setAttribute('aria-label', this.options.name);
  }

  // ...
},
```

### Current behavior

I would welcome feedback on the issue and proposed solution.

If accepted, I'd like to submit a pull-request.

Thank you!

Nick

### Environment

[...]

PR #9845, opened 2025-08-17
Add i18n helpers in core (v2)

### LINKED ISSUE #9092: feat: Add placeholder function for translation/localization/i18n to Leaflet
### Checklist

### Motivation

Currently, core Leaflet and Leaflet plugins are not translated from English. This is not a big problem for visual users, as most controls like the "Zoom in" button rely on icons. However, for non-visual and low vision users, for example those relying on screen readers it is a **significant barrier**, when maps are implemented on non-English websites (accessibility).

You may argue that implementors should set the [`zoomInTitle` option](https://leafletjs.com/reference.html#control-zoom) for the control. However, that is not exposed through the [main map options](https://leafletjs.com/reference.html#map-option), making it fiddly and inconsistent.

And, there are a number of accessibility-related open bugs and pull requests that add new text strings to Leaflet. For example:

* #8909 – `aria-label = "Layers"`
* #7193 – `aria-roledescription="map"` / `aria-label="map"`

### Suggested solution

I propose adding a placeholder or "dummy" for a translation function to core Leaflet. This would be a "pass-through" function that simply returns the string that is input as the first parameter.

That is, as an arrow function or non-arrow function:
```js
L._ = L._ || ((str) => str);
```

```js
L._ = L._ || function translatePlaceholder (string) { return string; }
```

Or, in a form that handles a second `data` parameter:
```js
L._ = function translate (string, data) {
    try {
        // Do not fail if some data is missing - a bad translation should not break the app.
        string = L.Util.template(string, data);
    }
    catch (err) { } /*pass*/

    return string;
};
```

This placeholder would be called wherever needed in core Leaflet - [Control.Zoom.js](https://github.com/Leaflet/Leaflet/blob/main/src/control/Control.Zoom.js#L30):
```js
  options: {
    …
    zoomInTitle: L._('Zoom in'),
    zoomOutTitle: L._('Zoom out'),
    …
  }
```

And, it can be adopted by third-party plugins [@brunob/leaflet.fullscreen - ...js](https://github.com/brunob/leaflet.fullscreen/blob/master/Control.FullScreen.js#L207):
```js

  options: {
    …
    title: L._('Full Screen'),
    titleCancel: L._('Exit Full Screen'),
    …
  }
```

Finally, anyone wishing to implement a map on a Non-English website would use a third-party internationalization plugin, for example, [Leaflet.i18n](https://github.com/umap-project/Leaflet.i18n), which [re-implements the `L._()` function](https://github.com/umap-project/Leaflet.i18n/blob/master/Leaflet.i18n.js#L27-L40).

Benefits of this approach:

1. A standardized, consistent, documented function is available to translate EVERY string in Leaflet for core and third-party plugins.

2. A lightweight approach. No need to implement all the internationalization functionality within Leaflet. No need to add language packs to core Leaflet. Instead, rely on an external plugin and language packs/locale files as needed.

---
## Steps:

1. Implement `L._()` function in core Leaflet,
2. Adopt the `L._()` function throughout Leaflet,
3. Write unit-tests for use of the `L._()` function.
4. Document the `L._()` function in the [API docs](https://leafletjs.com/reference.html),
5. Document the `L._()` function in the [Plugin Guide](https://github.com/Leaflet/Leaflet/blob/master/PLUGIN-GUIDE.md),
6. Identify pathfinder third-party plugins to adopt the `L._()` function, for example a fullscreen plugin.
7. Create a repo. containing reference/starter language packs/locale files, for some widely used languages.
8. Blog it!

This proposal is inspired by @yohanboniface's "Leaflet.i18n" plugin, and PR #1232.

I welcome feedback on this proposal!

Thanks.

### Alternatives considered

The [Leaflet.a11y](https://github.com/nfreear/leaflet.a11y) plugin is a short-term, non-optimal alternative. The solution above is much better long-term solution.

See the open PR: #9087, to list "Leaflet.a11y" in the plugin list!

simon04:
Resurrecting #9281. Fixes #9092.

### DIFF
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -13,43 +13,43 @@ describe('I18n', () => {
 		I18n.setLocale('fr');
 	});
 
-	it('expects current locale to be fr', () => {
+	it('uses the configured locale', () => {
 		expect(I18n.locale).to.eql('fr');
 	});
 
-	it('should translate simple sentences', () => {
+	it('translates simple sentences', () => {
 		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
 	});
 
-	it('should translate sentences with a variable', () => {
+	it('translates sentences with a variable', () => {
 		expect(I18n.translate('A phrase with a {variable} to translate', {variable: 'foo'})).to.eql('Une phrase à traduire avec une foo');
 	});
 
-	it('should translate empty translations', () => {
+	it('translates empty translations', () => {
 		expect(I18n.translate('A phrase with empty translation')).to.eql('');
 	});
 
-	it('should allow to override some translations', () => {
+	it('extends existing translations', () => {
 		I18n.registerLocale('fr', {'A phrase with empty translation': 'my custom translation'});
 		expect(I18n.translate('A phrase with empty translation')).to.eql('my custom translation');
 		// Unchanged
 		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
 	});
 
-	it('should not fail if a variable is missing', () => {
+	it('preserves placeholders when omitting variables', () => {
 		expect(I18n.translate('A phrase with a {variable} to translate')).to.eql('Une phrase à traduire avec une {variable}');
 	});
 
-	it('should not fail if the translation is missing', () => {
+	it('preserves the input when a translation is missing', () => {
 		expect(I18n.translate('A missing translation')).to.eql('A missing translation');
 	});
 
-	it('should not fail if the locale is missing', () => {
+	it('preserves the input if the locale is unknown', () => {
 		I18n.setLocale('foo');
 		expect(I18n.translate('Simple phrase to translate')).to.eql('Simple phrase to translate');
 	});
 
-	it('should allow to add many locales', () => {
+	it('translates from multiple locales', () => {
 		const it = {
 			'Simple phrase to translate': 'Frase semplice da tradurre',
 			'A phrase with a {variable} to translate': 'Una frase da tradurre con {variable}',
@@ -67,7 +67,7 @@ describe('I18n', () => {
 
 describe('Map.I18n', () => {
 
-	it('should be possible to translate Leaflet strings', () => {
+	it('translates built-in Leaflet strings', () => {
 		const container = createContainer();
 		const fr = {
 			'Zoom in': 'Zoomer',
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -17,6 +17,10 @@ describe('I18n', () => {
 		expect(I18n.locale).to.eql('fr');
 	});
 
+	it('returns the configured locale', () => {
+		expect(I18n.getLocale()).to.eql('fr');
+	});
+
 	it('translates simple sentences', () => {
 		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
 	});
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -29,8 +29,8 @@ describe('I18n', () => {
 		expect(I18n.translate('A phrase with a {variable} to translate', {variable: 'foo'})).to.eql('Une phrase à traduire avec une foo');
 	});
 
-	it('translates empty translations', () => {
-		expect(I18n.translate('A phrase with empty translation')).to.eql('');
+	it('preserves the input when the translation is empty', () => {
+		expect(I18n.translate('A phrase with empty translation')).to.eql('A phrase with empty translation');
 	});
 
 	it('extends existing translations', () => {
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -13,10 +13,6 @@ describe('I18n', () => {
 		I18n.setLocale('fr');
 	});
 
-	it('uses the configured locale', () => {
-		expect(I18n.locale).to.eql('fr');
-	});
-
 	it('returns the configured locale', () => {
 		expect(I18n.getLocale()).to.eql('fr');
 	});
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -1,4 +1,5 @@
-import {locale, setLocale, registerLocale, translate} from 'leaflet';
+import {locale, setLocale, registerLocale, translate, Map} from 'leaflet';
+import {createContainer, removeMapContainer} from '../SpecHelper.js';
 
 describe('I18n', () => {
 	beforeEach(() => {
@@ -58,6 +59,28 @@ describe('I18n', () => {
 		expect(translate('Simple phrase to translate')).to.eql('Frase semplice da tradurre');
 		setLocale('fr');
 		expect(translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+});
+
+
+describe('Map.I18n', () => {
+
+	it('should be possible to translate Leaflet strings', () => {
+		const container = createContainer();
+		const fr = {
+			'Zoom in': 'Zoomer',
+			'Zoom out': 'Dézoomer',
+			'A JavaScript library for interactive maps': 'Bibliothèque JavaScript pour cartes interactives',
+		};
+		registerLocale('fr', fr);
+		setLocale('fr');
+	    const map = new Map(container);
+		map.setView([0, 0], 0);
+		expect(document.querySelector('.leaflet-control-zoom-in').title).to.eql('Zoomer');
+		expect(document.querySelector('.leaflet-control-zoom-out').title).to.eql('Dézoomer');
+		expect(document.querySelector('.leaflet-control-attribution a').title).to.include('Bibliothèque JavaScript pour cartes interactives');
+		removeMapContainer(map, container);
 	});
 
 });
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -1,0 +1,63 @@
+import {locale, setLocale, registerLocale, translate} from 'leaflet';
+
+describe('I18n', () => {
+	beforeEach(() => {
+		const fr = {
+			'Simple phrase to translate': 'Une simple phrase à traduire',
+			'A phrase with a {variable} to translate': 'Une phrase à traduire avec une {variable}',
+			'A phrase with empty translation': ''
+		};
+		registerLocale('fr', fr);
+		setLocale('fr');
+	});
+
+	it('expects current locale to be fr', () => {
+		expect(locale).to.eql('fr');
+	});
+
+	it('should translate simple sentences', () => {
+		expect(translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+	it('should translate sentences with a variable', () => {
+		expect(translate('A phrase with a {variable} to translate', {variable: 'foo'})).to.eql('Une phrase à traduire avec une foo');
+	});
+
+	it('should translate empty translations', () => {
+		expect(translate('A phrase with empty translation')).to.eql('');
+	});
+
+	it('should allow to override some translations', () => {
+		registerLocale('fr', {'A phrase with empty translation': 'my custom translation'});
+		expect(translate('A phrase with empty translation')).to.eql('my custom translation');
+		// Unchanged
+		expect(translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+	it('should not fail if a variable is missing', () => {
+		expect(translate('A phrase with a {variable} to translate')).to.eql('Une phrase à traduire avec une {variable}');
+	});
+
+	it('should not fail if the translation is missing', () => {
+		expect(translate('A missing translation')).to.eql('A missing translation');
+	});
+
+	it('should not fail if the locale is missing', () => {
+		setLocale('foo');
+		expect(translate('Simple phrase to translate')).to.eql('Simple phrase to translate');
+	});
+
+	it('should allow to add many locales', () => {
+		const it = {
+			'Simple phrase to translate': 'Frase semplice da tradurre',
+			'A phrase with a {variable} to translate': 'Una frase da tradurre con {variable}',
+			'A phrase with empty translation': ''
+		};
+		registerLocale('it', it);
+		setLocale('it');
+		expect(translate('Simple phrase to translate')).to.eql('Frase semplice da tradurre');
+		setLocale('fr');
+		expect(translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+});
--- a/spec/suites/core/I18nSpec.js
+++ b/spec/suites/core/I18nSpec.js
@@ -1,5 +1,6 @@
-import {locale, setLocale, registerLocale, translate, Map} from 'leaflet';
+import {I18n, Map} from 'leaflet';
 import {createContainer, removeMapContainer} from '../SpecHelper.js';
+import {expect} from 'chai';
 
 describe('I18n', () => {
 	beforeEach(() => {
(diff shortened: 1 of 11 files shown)


spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		};
+		I18n.registerLocale('fr', fr);
+		I18n.setLocale('fr');
+	});
+
+	it('expects current locale to be fr', () => {
+		expect(I18n.locale).to.eql('fr');
+	});
+
+	it('should translate simple sentences', () => {
```
jonkoops: Nit: instead of using 'should' for every test, try using the verb directly:

```suggestion
	it('translates simple sentences', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+	it('should translate sentences with a variable', () => {
```
jonkoops:
```suggestion
	it('translates sentences with a variable', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		expect(I18n.translate('A phrase with a {variable} to translate', {variable: 'foo'})).to.eql('Une phrase à traduire avec une foo');
+	});
+
+	it('should translate empty translations', () => {
```
jonkoops:
```suggestion
	it('translates empty translations', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		expect(I18n.translate('A phrase with empty translation')).to.eql('');
+	});
+
+	it('should allow to override some translations', () => {
```
jonkoops:
```suggestion
	it('extends existing translations', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		I18n.registerLocale('fr', {'A phrase with empty translation': 'my custom translation'});
+		expect(I18n.translate('A phrase with empty translation')).to.eql('my custom translation');
+		// Unchanged
+		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+	it('should not fail if a variable is missing', () => {
```
jonkoops:
```suggestion
	it('preserves placeholders when omitting variables', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		expect(I18n.translate('A phrase with a {variable} to translate')).to.eql('Une phrase à traduire avec une {variable}');
+	});
+
+	it('should not fail if the translation is missing', () => {
```
jonkoops:
```suggestion
	it('preserves the input when a translation is missing', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		const fr = {
+			'Simple phrase to translate': 'Une simple phrase à traduire',
+			'A phrase with a {variable} to translate': 'Une phrase à traduire avec une {variable}',
+			'A phrase with empty translation': ''
+		};
+		I18n.registerLocale('fr', fr);
+		I18n.setLocale('fr');
+	});
+
+	it('expects current locale to be fr', () => {
```
jonkoops:
```suggestion
	it('uses the configured locale', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+
+	it('should not fail if a variable is missing', () => {
+		expect(I18n.translate('A phrase with a {variable} to translate')).to.eql('Une phrase à traduire avec une {variable}');
+	});
+
+	it('should not fail if the translation is missing', () => {
+		expect(I18n.translate('A missing translation')).to.eql('A missing translation');
+	});
+
+	it('should not fail if the locale is missing', () => {
```
jonkoops:
```suggestion
	it('preserves the input if the locale is unknown', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		I18n.setLocale('foo');
+		expect(I18n.translate('Simple phrase to translate')).to.eql('Simple phrase to translate');
+	});
+
+	it('should allow to add many locales', () => {
```
jonkoops:
```suggestion
	it('translates from multiple locales', () => {
```

spec/suites/core/I18nSpec.js
```diff
@@ -0,0 +1,87 @@
...
+		I18n.setLocale('fr');
+		expect(I18n.translate('Simple phrase to translate')).to.eql('Une simple phrase à traduire');
+	});
+
+});
+
+
+describe('Map.I18n', () => {
+
+	it('should be possible to translate Leaflet strings', () => {
```
jonkoops:
```suggestion
	it('translates built-in Leaflet strings', () => {
```

src/control/Control.Zoom.js
```diff
@@ -29,17 +30,9 @@ export class Zoom extends Control {
 			// The text set on the 'zoom in' button.
 			zoomInText: '<span aria-hidden="true">+</span>',
 
-			// @option zoomInTitle: String = 'Zoom in'
-			// The title set on the 'zoom in' button.
```
jonkoops, line 33: Removing these comments here will also remove the options from the API docs.

src/core/I18n.js
```diff
@@ -0,0 +1,48 @@
...
+/*
+ * @namespace I18n
+ *
+ * Utility functions to deal with translations.
+ */
+
+export class I18n {
+	// @property messages: Object
+	// All localized strings.
+	static messages = {};
```
jonkoops: Considering the messages are always one level deep, and use string based keys and values, perhaps a [`Map` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)is a better type here? I'd also argue we'd probably want to keep this state private, and only expose methods to do the translation/register new messages.


```suggestion
	static #messages = new Map();
```

src/core/I18n.js
```diff
@@ -0,0 +1,48 @@
...
+
+	// @property locale: String
+	// The current locale code, that will be used when translating strings.
+	static locale = null;
+
+	// @function registerLocale(locale: String, messages?: Object): String
```
jonkoops: This method does not return a `String`.

```suggestion
	// @function registerLocale(locale: String, messages?: Object): undefined
```

src/core/I18n.js
```diff
@@ -0,0 +1,48 @@
...
+	// Define localized strings for a given locale, defined by `locale`.
+	static registerLocale(locale, messages) {
+		this.messages[locale] = {...this.messages[locale], ...messages};
+	}
+
+	// @function setLocale(locale: String): undefined
+	// Define or change the locale code to be used when translating strings.
+	static setLocale(locale) {
+		this.locale = locale;
```
jonkoops: What if a user changes the locale at runtime? There is no mechanism of updating existing DOM with the new labels (e.g. the user changes the locale on a page). Should we create a mechanism to handle this, or do we simply tell users to refresh the Map?

src/core/I18n.js
```diff
@@ -0,0 +1,48 @@
...
+		}
+		try {
+			// Do not fail if some data is missing
+			// a bad translation should not break the app
+			string = Util.template(string, data);
+		} catch (err) {
+			console.error('Leaflet translate', err);
+		}
+
+		return string;
```
jonkoops: Couple of notes here:

- We can do a quick opt-out here when `locale` is not set, since that effectively means localization is disabled.
- No need to fall back to `"?"` to access the messages, the optional chaining already takes care of possibly empty values.
- We should probably warn about missing translations, if localization is enabled (`locale` is not `null`).

```suggestion
		// If the locale is not set, localization is not enabled.
		if (this.locale === null) {
			return string;
		}

		const translation = this.messages[this.locale]?.[string];

		if (!translation) {
			console.warn(`No translation found for "${string}", falling back to the original string.`);
			return string;
		}

		try {
			// Do not fail if some data is missing
			// a bad translation should not break the app
			return Util.template(translation, data);
		} catch (err) {
			console.error(`Error while translating "${translation}".`, err);
			return translation;
		}
```

jonkoops:
@yohanboniface as the maintainer of `leaflet-i18n` I'd appreciate your input on this feature.

src/core/I18n.js
```diff
@@ -0,0 +1,48 @@
...
+
+	// @function registerLocale(locale: String, messages?: Object): String
+	// Define localized strings for a given locale, defined by `locale`.
+	static registerLocale(locale, messages) {
+		this.messages[locale] = {...this.messages[locale], ...messages};
+	}
+
+	// @function setLocale(locale: String): undefined
+	// Define or change the locale code to be used when translating strings.
+	static setLocale(locale) {
```
Falke-Design, line 27: maybe we should add a `getLocale()` too

src/control/Control.Zoom.js
simon04: Yes, this is intended, since `options.zoomInTitle` is no longer used below.

src/core/I18n.js
simon04: `messages` maps `locale`s to a object of all translations, for instance, `messages = {de: {'Zoom in': 'Hinein zoomen'}}` -- I don't think, we can simply change it to a `Map`

Regarding private state: maybe, yes. However, it would be the first usage of `#` in the entire codebase?

src/core/I18n.js
simon04: Good point. Since this is a static utility function, we currently do not know where the i18n strings are being used. We could consider emitting a "locale changed" event (but then `I18n` cannot be static?). However, I wonder whether it's worth the effort. At least, we should add a note that (in the current behaviour), the locale needs to be configured before instantiating a `LeafletMap`.

src/core/I18n.js
simon04: Good point. I'll do that.

src/core/I18n.js
jonkoops: Perhaps a `Map` is overkill for now, either way it is private logic that I would not want to expose as a public API right now, to avoid future refactoring becoming an issue.

> However, it would be the first usage of # in the entire codebase?

Yes, this is true. I have an intention of making all private APIs truly private (e.g. methods starting with `_`). It is standard JS syntax, so IMHO let's use it.

src/control/Control.Zoom.js
jonkoops: Ah, so your idea is to remove all values that can now instead be provided by the i18n API?

src/core/I18n.js
jonkoops: Yes, I think we can live without dynamically changing the language. Let's consider it an accepted limitation of the API.

src/control/Control.Zoom.js
simon04: Yes, that was my idea. Having two mechanisms for the same thing seems confusing to me.

src/core/I18n.js
simon04: Alright. I've made the two members of `I18n` private.

src/core/I18n.js
simon04: I've added a note to the docs.

src/core/I18n.js
Falke-Design: > I have an intention of making all private APIs truly private (e.g. methods starting with _). It is standard JS syntax, so IMHO let's use it.

Just because something is standard doesn’t mean it’s always good. If we make every "private" method truly private, we lose a lot of Leaflet’s flexibility. I’m fine with making the i18n method private, but I wouldn’t change the others.

jonkoops: We can discuss making things truly private in another moment. For now, I'd at least do it here with new code, especially if it does not extend `Class`.

nfreear:
Thank you @simon04 and @jonkoops for all your work on this. Very much appreciated!

I've found a bug - on [line 42 of `I18n.js`](https://github.com/simon04/Leaflet/blob/i18n/src/core/I18n.js#L42), the `data` parameter should be passed into the `Util.template(string)` call, when the locale is not set.

Expected JavaScript:
```js
static translate(string, data = {}) {
  if (this.#locale === null) {
    return Util.template(string, data);
  }
  ...
}
```

Apart from that, I can't find any other problems.

I'm testing the `I18n` class in [this demo page](https://github.com/nfreear/maps/blob/main/pages/leaflet.i18n.html).

(FYI, I'm the maintainer of the [Leaflet.translate](https://github.com/nfreear/leaflet.plugins) plugin, which uses @yohanboniface's `leaflet-i18n` plugin.)

simon04:
> I've found a bug - on [line 42 of `I18n.js`](https://github.com/simon04/Leaflet/blob/i18n/src/core/I18n.js#L42), the `data` parameter should be passed into the `Util.template(string)` call, when the locale is not set.

Thanks for spotting this. Fixed now.

nfreear:
Great stuff. Thanks @simon04 for responding.

I've found some more places where I believe `I18n.translate()` could and should be used:

1. [`src/layer/Popup.js` - Line 107](https://github.com/Leaflet/Leaflet/blob/main/src/layer/Popup.js#L107):
   ```js
   closeButtonLabel: 'Close popup',
   ```
   I'm not sure that the `closeButtonLabel` option is actually needed now that translation is an option.

2. [`src/control/Control.Layers.js` - Line 227](https://github.com/Leaflet/Leaflet/blob/main/src/control/Control.Layers.js#L227):
   ```js
   link.title = 'Layers';
   ```

3. [`src/control/Control.Scale.js` - Lines 97 - 113](https://github.com/Leaflet/Leaflet/blob/main/src/control/Control.Scale.js#L97-L113):
   ```js
   label = meters < 1000 ? `${meters} m` : `${meters / 1000} km`;  // Line 97.

   this._updateScale(this._iScale, `${miles} mi`, miles / maxMiles);  // Line 109.

   this._updateScale(this._iScale, `${feet} ft`, feet / maxFeet);  // Line 113.
   ```

Regarding number `3.`, though an abbreviation like "km" is widely known, for example in China, there is an alternative using Chinese characters. And, the abbreviation "mi" for miles would not be known.

I hope this all helps.

Nick

PS. Sorry I'm not in a position to formally "sign off" on this PR!

nfreear:
Regarding [`src/layer/tile/TileLayer.js` - Line 105](https://github.com/simon04/Leaflet/blob/i18n/src/layer/tile/TileLayer.js#L105), `options.attribution =…` - from a security (XSS) standpoint, the translation strings should not contain HTML markup.

I suggest that this would be better written as:

```js
options.attribution = I18n.translate('© {link}OpenStreetMap{endlink} contributors', {
    link: '<a href="http://www.openstreetmap.org/copyright">',
    endlink: '</a>',
});
```

(Finished for the evening!)

src/control/Control.Zoom.js
jonkoops: Agreed, we just need to ensure we provide information about this in an upgrade guide.

jonkoops:
+1 on the HTML suggestion, we need to find a way to handle interpolation properly.

simon04:
@nfreear, thanks for listing additional places in the code. Regarding "OpenStreetMap contributors", I found it easier to mark "OpenStreetMap contributors" for translation and wrap it in the `<a>` outside of the translation system.

Falke-Design:
Yes let's handle "OpenStreetMap contributors" separately. In my eyes there is no translation needed

nfreear:
I consider this pull request to be on the critical path for several 2.0-related things:

* [Plugin developers](https://leafletjs.com/plugins.html), who while they update their plugins for 2.0, wish to also use `I18n.translate()`,
* Accessibility: various open issues including #9119, #9116(PR), #7822, and so on that involve setting an accessible name/text alternative, may rely on calls to `I18n.translate()`.

---

FYI, @simon04, @jonkoops, @yohanboniface and @Falke-Design

PR #9854, opened 2025-08-19
Round BlanketOverlay transform offset

### LINKED ISSUE #9852: L.Path vector features are blurred and shifted in v2.0.0-alpha.1
### Checklist

### Expected behavior

Like in v1.9.4, L.Path descendants (L.Circle, L.Polyline, L.Polygon) should be rendered on initial load as crisp, clear and without any spatial shifts. 

### Current behavior

When you load a map in v2.0.0-alpha.1 with L.Path descendants (L.Circle, L.Polyline, L.Polygon), they:
- initially appear blurry AND 
- spatially slightly shifted on the map. 

If you do a zoom-in and zoom-out with the zoom control (to keep map extent the same), once the features are rendered again, crispness is back, but the spatial shift is still there (compared to v1.9.4). TileLayer seems unaffected.

For the initial load, the features are very blurry on load, and that is highlighted when using narrower stroke styles and/or more complex fills for the polygons (like hatch fills). Something is up with the final rendering of them, as the fill of polygons are also blurred, not just the outline/stroke around it. Might be an unwanted distortion?

For the shift, I'm not sure if something changed in the simplification or in the final render pipeline of the L.Path objects. It might be that the v2.0.0-alpha.1 renders features more precisely (true-to-projection) while v1.9.4 had some imperfections, I did not evaluate that specifically. However, there is a difference between the two versions for sure. Axis of the shift is not clear, it seems to depend on spatial location - most of the times they are shifted northwards, but in some cases on small zoom levels also westwards.

Static screenshot example: v2 on the left, v1.9.4 on the right, same exact crop, both after initial load (no interaction).

Image

- v2.0.0-alpha.1: https://codepen.io/balladaniel/pen/zxvWGbB
- v1.9.4: https://codepen.io/balladaniel/pen/YPyaXBY

### Environment

- Leaflet version: 2.0.0-alpha.1
- Browser (with version): Chrome 139.0.7258.128 (64-bit)
- OS/Platform (with version): Windows 11 Pro 23H2, 64-bit, build 22631.5472

IvanSanchez:
Should fix #9852.

### DIFF
--- a/src/layer/BlanketOverlay.js
+++ b/src/layer/BlanketOverlay.js
@@ -84,8 +84,9 @@ export class BlanketOverlay extends Layer {
 		const scale = this._map.getZoomScale(zoom, this._zoom),
 		viewHalf = this._map.getSize().multiplyBy(0.5 + this.options.padding),
 		currentCenterPoint = this._map.project(this._center, zoom),
-		topLeftOffset = viewHalf.multiplyBy(-scale).add(currentCenterPoint)
-			.subtract(this._map._getNewPixelOrigin(center, zoom));
+		topLeftOffset = viewHalf.multiplyBy(-scale)._add(currentCenterPoint)
+			._subtract(this._map._getNewPixelOrigin(center, zoom))
+			._round();
 
 		DomUtil.setTransform(this._container, topLeftOffset, scale);
 	}


Falke-Design (approved):
@IvanSanchez I found the difference between v1 and v2 which causes the decimal pixels, please see https://github.com/Leaflet/Leaflet/issues/9852#issuecomment-3211287047

PR #9856, opened 2025-08-24
set compatible-v2 flag for leaflet.geodesic

henrythasler:
leaflet.geodesic has a first release for leaflet 2.0.0-alpha.1 so I'd like to update the v2-flag here.


Falke-Design (approved):

jonkoops:

PR #9863, opened 2025-08-28
docs: update leaflet-bouncemarker for v2

maximeh:
Updating the documentation since the plugin has been made compatible with v2 as well.


Falke-Design (approved):

jonkoops:

PR #9865, opened 2025-09-02
Change Dependabot schedule from weekly to quarterly

simon04:
Ref: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#interval

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     rebase-strategy: disabled
     versioning-strategy: increase
     schedule:
-      interval: weekly
+      interval: quarterly
     labels:
       - dependencies
   - package-ecosystem: github-actions
@@ -14,7 +14,7 @@ updates:
     open-pull-requests-limit: 999
     rebase-strategy: disabled
     schedule:
-      interval: weekly
+      interval: quarterly
     labels:
       - dependencies
   - package-ecosystem: bundler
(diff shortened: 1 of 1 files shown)


jonkoops:
Seems a tad excessive to move away from a weekly schedule, is this really creating that much overhead?

Falke-Design:
In my opinion, `monthly` or `quarterly` is fine and helps reduce maintenance work. The PRs will be open for months anyway

jonkoops:
I sure hope we don't have to keep Dependabot PRs open that long, but yes, that has been the trend due to low availability. I merged a couple just now, since we don't have too many dependencies it is not a big deal. Looks like the only that is failing is the upgrade of Chai to the next major version, I will take a look at that.

mourner:
One thing we could do to reduce the number of open Dependabot PRs is to set up Dependabot automerge via GitHub Actions: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

Falke-Design:
> One thing we could do to reduce the number of open Dependabot PRs is to set up Dependabot automerge via GitHub Actions: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions

Yes let us implement a workflow

jonkoops:
+1 on auto-merge as long as the CI checks pass. Perhaps we can prevent major version bumps from being merged automatically, if we want to be conservative.

jonkoops:
Closing this PR, as we'd like to go the auto-merge route. I will try to keep on top of Dependabot PRs in the meantime.

PR #9866, opened 2025-09-02
Warn about removed mouse* events

simon04:
Follow-up for #9620.

### DIFF
--- a/src/core/Events.js
+++ b/src/core/Events.js
@@ -91,6 +91,10 @@ export class Evented extends Class {
 
 	// attach listener (without syntactic sugar now)
 	_on(type, fn, context, _once) {
+		if (['mousedown', 'mouseup', 'mouseover', 'mouseout', 'mousemove'].includes(type)) {
+			console.error(`The event ${type} has been removed. Use the pointer* variant instead.`);
+		}
+
 		if (typeof fn !== 'function') {
 			console.warn(`wrong listener type: ${typeof fn}`);
 			return;
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1939,6 +1939,13 @@ describe('Map', () => {
 			expect(spy.firstCall.lastArg).to.be.true;
 			expect(spy.secondCall.lastArg).to.be.true;
 		});
+
+		it('prints to console.error about deprecated mouse* events', () => {
+			const errorSpy = sinon.spy(console, 'error');
+			map.on('mouseover', () => {});
+			expect(errorSpy.calledOnce).to.be.true;
+			expect(errorSpy.firstCall.args[0]).to.eq('The event mouseover has been removed. Use the pointer* variant instead.');
+		});
 	});
 
 	describe('#getScaleZoom && #getZoomScale', () => {


src/core/Events.js
```diff
@@ -91,6 +91,10 @@ export class Evented extends Class {
 
 	// attach listener (without syntactic sugar now)
 	_on(type, fn, context, _once) {
+		if (['mousedown', 'mouseup', 'mouseover', 'mouseout', 'mousemove'].includes(type)) {
```
jonkoops: Is this something we want to log in the `Events` class? There might be perfectly valid reasons for community plugins to use these events, right?

Falke-Design: > There might be perfectly valid reasons for community plugins to use these events, right?

No not really, because only PointerEvents are handled by Leaflet

Falke-Design: After an exchange with Simon, we realized that this error log could be helpful for migrations.

jonkoops:
Perhap we can even move any future deprecations to an issue? We can add it to the `v3` milestone to remove them.

simon04:
I've added a "To be removed in leaflet 3" comment in two locations in the codebase.

src/core/Events.js
Falke-Design: The only concerns I have are when using the v1 polyfill, then we would show allways an error. Can we somehow disable it via a flag? 😄  Or can we make the array editable? Then I could remove the events in the polyfill

jonkoops: Perhaps we can make some kind of semi-global flag to disable specific deprecations? E.g. `Events.__DEPRECATION_MOUSE_EVENTS = false`, so that it can be suppressed by the polyfill.

simon04: I've introduced a `static __REMOVED_EVENTS = ['mousedown', 'mouseup', 'mouseover', 'mouseout', 'mousemove']`

src/core/Events.js
```diff
@@ -89,8 +89,15 @@ export class Evented extends Class {
 		return this;
 	}
 
+	static __REMOVED_EVENTS = ['mousedown', 'mouseup', 'mouseover', 'mouseout', 'mousemove'];
+
 	// attach listener (without syntactic sugar now)
 	_on(type, fn, context, _once) {
+		// To be removed in leaflet 3
+		if (Evented.__REMOVED_EVENTS.includes(type)) {
+			console.error(`The event ${type} has been removed. Use the pointer* variant instead.`);
```
Falke-Design:
```suggestion
			console.error(`The event ${type} has been removed. Use the PointerEvent variant instead.`);

```

When someone doesn't know PointerEvents he will maybe replace the event with `pointer*`

src/core/Events.js
```diff
@@ -89,8 +89,16 @@ export class Evented extends Class {
...
 	}
 
+	static __REMOVED_EVENTS = ['mousedown', 'mouseup', 'mouseover', 'mouseout', 'mousemove'];
+
 	// attach listener (without syntactic sugar now)
 	_on(type, fn, context, _once) {
+		// To be removed in leaflet 3
+		if (Evented.__REMOVED_EVENTS.includes(type)) {
+			console.error(`The event ${type} has been removed. Use the PointerEvent variant instead.`);
+
```
Falke-Design:
```suggestion
			console.error(`The event ${type} has been removed. Use the PointerEvent variant instead.`);
```

Falke-Design (approved):

src/map/handler/Map.PinchZoom.js
```diff
@@ -135,6 +135,7 @@ Map.addInitHook(function () {
 	this.touchZoom = this.pinchZoom;
 
 	if (this.options.touchZoom !== undefined) {
+		// To be removed in leaflet 3
```
jonkoops, line 138: Please log an issue for v3 to remove this, otherwise we might forget.

simon04: Done: https://github.com/Leaflet/Leaflet/issues/9908

Issue #9871, opened 2025-09-08
Clarify how to get a nightly snapshot (latest dev) download


indigoxela:
### Checklist

- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this change hasn't been suggested already.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Which documents need updating?

https://leafletjs.com/download.html

### Motivation

Until recently it was possible to also download the latest dev snapshot form the download page. As build, not only the source code, which is available via GitHub.

This doesn't seem to be possible anymore, so one has to wait for a tagged (pre-)release to get up to date with development, it seems.

### Suggested changes

If there's still a way, add that link back. Note: a full build, not the source code.

If this isn't possible anymore, maybe add some info, what one could do to get a "nightly snapshot", like build instructions... Not sure, how helpful that could be for non-js-devs, but mentioning possible approaches still seems useful.

Falke-Design:
I see the advantages of providing the main build as download, but maintaining it is difficult, so we removed it (#9806). We’re happy to consider other suggestions to provide the main build.

For manual builds, see:

https://github.com/Leaflet/Leaflet/blob/main/CONTRIBUTING.md#setting-up-the-build-system
https://github.com/Leaflet/Leaflet/blob/main/CONTRIBUTING.md#building-leaflet

indigoxela:
I see. To be honest, that's what I suspected, anyway.

> For manual builds, see...

Ah, get it, local build with npm.

Maybe mention that at least in a short sentence on the downloads page, linking to either that CONTRIBUTING.md, or to shorter instructions somewhere? What *testers* (like me) need is a smaller subset of the contributing info, I guess.

Something like:

- get sources from GH?
- npm install (to fetch deps)?
- npm run build
- copy stuff from "dist" folder to your project

I do have npm available, but I'm not very familiar with it (I usually only download ready-to-use stuff). And I could never contribute anything else than testing / reporting. 😉

To be honest, the current instruction "Building Leaflet from the Source" is rather confusing for me. 🙊

PR #9874, opened 2025-09-08
Automatically update nightly release with latest build

Falke-Design:
This workflow reverts partially #9806. Instead of uploading to S3 we will now have a Github nightly pre-release which updates automatically all build files with a push to the `main` branch. We can then reference on the download page to the assets in the release, like: https://github.com/Leaflet/Leaflet/releases/download/nightly/leaflet.js

This will look like this:
grafik

`GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` didn't worked on my private test repo, I needed to add a PAT-Token but it should work with GITHUB_TOKEN.

Fixes: #9871

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -184,7 +184,7 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           tag_name: dev
-          name: "Nightly Build"
+          name: Development snapshot
           prerelease: true
           files: dist/*
           generate_release_notes: false


.github/workflows/main.yml
```diff
@@ -165,6 +165,33 @@ jobs:
       - name: Run tests on ${{ matrix.browser }} with touch
         run: npm test -- --browsers ${{ matrix.browser }} --touch-browser --single-run
 
+  publish-artifacts:
+    needs: setup
+    if: github.repository_owner == 'Leaflet' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
```
jonkoops: Do we want this to trigger on tags? I would imagine we only want this to happen on the latest commit on `main`.
```suggestion
    if: github.repository_owner == 'Leaflet' && github.ref == 'refs/heads/main'
```

.github/workflows/main.yml
```diff
@@ -165,6 +165,33 @@ jobs:
...
+        with:
+          path: ./*
+          key: ${{ runner.os }}-${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Compress artifacts
+        working-directory: dist
+        run: zip -r leaflet.zip . ../CHANGELOG.md ../LICENSE ../README.md
+
+      - name: Publish Nightly Pre-release
+        if: github.ref == 'refs/heads/main'
```
jonkoops: This should already be gated by the `if` condition of the job itself, no?

.github/workflows/main.yml
Falke-Design: I just copied the old logic for publishing but yes, I think main should be enough

.github/workflows/main.yml
```diff
@@ -165,6 +165,32 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: ./*
+          key: ${{ runner.os }}-${{ github.ref }}-${{ github.sha }}-setup
+
+      - name: Compress artifacts
+        working-directory: dist
+        run: zip -r leaflet.zip . ../CHANGELOG.md ../LICENSE ../README.md
+
+      - name: Publish Nightly Pre-release
```
jonkoops:
```suggestion
      - name: Publish development snapshot
```

.github/workflows/main.yml
```diff
@@ -165,6 +165,32 @@ jobs:
...
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: nightly
```
jonkoops:
```suggestion
          tag_name: dev
```

.github/workflows/main.yml
```diff
@@ -165,6 +165,32 @@ jobs:
...
+          name: "Nightly Build"
```
jonkoops:
```suggestion
          name: Development snapshot
```

jonkoops (approved):
Since this release runs every time a commit is made on `main`, perhaps we should call it a development snapshot instead? A proper nightly would be a run that happens every night on a schedule.

Either way, I am approving, since we really need this for users doing testing. Do as you see fit 👍

Falke-Design:
development snapshot sounds better. Thank you

Falke-Design:
@mourner @IvanSanchez @jonkoops publishing failed, can you please check the settings / tokens. I have not enough permission.

> 👩‍🏭 Creating new GitHub release for tag dev...
⚠️ GitHub release failed with status: 403
{"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/releases/releases#create-a-release","status":"403"}
Skip retry — your GitHub token/PAT does not have the required permission to create a release
⚠️ Unexpected error fetching GitHub release for tag refs/heads/main: HttpError: Resource not accessible by integration - https://docs.github.com/rest/releases/releases#create-a-release
Resource not accessible by integration - https://docs.github.com/rest/releases/releases#create-a-release

jonkoops:
I think @mourner might be the only one who has permission to do so.

mourner:
@jonkoops @Falke-Design I'm not sure what to do exactly, do I need to make some kind of new secret token in the settings, or is that `GITHUB_TOKEN` some reserved internal one?

jonkoops:
@Falke-Design looks like this job [needs](
https://github.com/softprops/action-gh-release?tab=readme-ov-file#permissions) the `contents: write` permission. We currently set to `read` for the whole workflow:

https://github.com/Leaflet/Leaflet/blob/d261e1615ae03f9c909bacfebbfc3e30f5b58f00/.github/workflows/main.yml#L3-L4

We can set the permissions for this [specific job](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idpermissions) to be more permissive:

```yml
publish-artifacts:
 permissions: 
   contents: write 
```

PR #9879, opened 2025-09-13
Update leaflet-burgermenu.md

BenPortner:
Add Leaflet v2 compatibility notice

### DIFF
--- a/docs/_plugins/user-interface/leaflet-burgermenu.md
+++ b/docs/_plugins/user-interface/leaflet-burgermenu.md
@@ -6,8 +6,8 @@ author: Benjamin W. Portner
 author-url: https://github.com/BenPortner
 demo: https://benportner.github.io/leaflet-burgermenu/
 compatible-v0: unknown
-compatible-v1: v0.1.0
-compatible-v2: >v1.0.0
+compatible-v1: true
+compatible-v2: true
 ---
 
 A lightweight Leaflet plugin that adds a burger menu with nested submenus to the map interface. Hovering over menu items reveals submenus, and clicking items can trigger custom actions.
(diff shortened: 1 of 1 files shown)


docs/_plugins/user-interface/leaflet-burgermenu.md
```diff
@@ -6,8 +6,8 @@ author: Benjamin W. Portner
 author-url: https://github.com/BenPortner
 demo: https://benportner.github.io/leaflet-burgermenu/
 compatible-v0: unknown
-compatible-v1: true
-compatible-v2: false
+compatible-v1: v0.1.0
+compatible-v2: >v1.0.0
```
jonkoops: AFAIK these fields cannot be strings, is there any indication this works?

Falke-Design: We need a boolean to handle it correctly on the website

```suggestion
compatible-v1: true
compatible-v2: true
```

Falke-Design (approved):

BenPortner:
@Falke-Design @jonkoops I am a very skeptical that indicating the compatibility with a mere check mark is a good idea. Users will think that the newest version of a plugin can support both Leaflet v1 and v2. However, this will be impossible for most plugins due to Leaflet's dropping of the global L variable in the standard module. Please reconsider allowing version strings in the compatibility columns - or at least add another column where such data can be entered for the users' convenience.

jonkoops:
I have to agree. The question is whether we should be documenting that here, or if the plugin author is responsible for listing that information on their own plugin's readme. We already struggle with plugin authors listing their plugins on the site, and not maintaining them, so it is unlikely that they will keep things up-to-date as we go.

Not saying we don't want to do this, just wondering out loud what the best course of action is here.

Falke-Design:
Since most plugins will not provide a version, I don’t think it’s worth the effort to show the version in the Leaflet list. Please list the compatible versions in your own README.

jonkoops:
I agree with @Falke-Design. We can always revisit this in the future.

jonkoops:

Issue #9814, opened 2025-08-02
Treeshaking not working


Maxel01:
### Checklist

- [x] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [x] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

## Nuxt
1. ```pnpm create nuxt@latest```
2. `pnpm add leaflet@2.0.0-alpha`
3. add to app/app.vue
```js
<script>
import {Map} from "leaflet"
</script>
```
4. run `nuxi analyze`
It's not a fully working version. However the leaflet file will be loaded entirely.

## Vite
``` bash
npm init vite@latest
# choose "vanilla" as template
cd vite-project
npm install leaflet
```
Create vite.config.js
``` js
import { defineConfig } from 'vite';

export default defineConfig({});
```
edit index.html
``` html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>Leaflet Vite Demo</title>
  <link rel="stylesheet" href="https://unpkg.com/leaflet@2.0.0-alpha/dist/leaflet.css" />
  <style>
    #map {
      height: 100vh;
    }
  </style>
</head>
<body>
<div id="map"></div>
<script type="module" src="/main.js"></script>
</body>
</html>
```
create main,js
``` js
import { Map, TileLayer } from 'leaflet';

const map = new Map('map').setView([51.505, -0.09], 13);

new TileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png', {
  maxZoom: 19,
  attribution: '&copy; OpenStreetMap contributors',
}).addTo(map);
```
Test setup width `npm run dev`
Check the build: `npx vite-bundle-analyzer `

### Expected behavior

According to the release notes, ESM support and tree shaking should be available. So I’d expect only the relevant parts of the JS file to be loaded.
My Vue-Leaflet wrapper, for example, only pulls in the components I use — which keeps the bundle size down.

### Current behavior

It appears that the entire Leaflet file is loaded regardless of which parts are used, meaning true tree shaking doesn’t occur.

[...]

PR #9880, opened 2025-09-13
Expose Leaflet global for UMD version only

### LINKED ISSUE #9842: Global L is exported via ESM
When generating the dist files, we use the `LeafletWithGlobals` file. This causes the ESM module to add the global L to the window object. Should we change the export configs?

* UMD -> LeafletWithGlobals.js (with global L)
* ESM -> Leaflet.js (without global L)

https://github.com/Leaflet/Leaflet/blob/cbe2d0a1eeed70acc06bc7e9f59085aa267d883d/build/rollup-config.js#L15-L33

https://github.com/Leaflet/Leaflet/blob/cbe2d0a1eeed70acc06bc7e9f59085aa267d883d/src/LeafletWithGlobals.js#L1-L20

https://plnkr.co/edit/7wAJePQI44QmAedQ

@mourner, @IvanSanchez, @jonkoops, @simon04

jonkoops:
Changes the build so the Leaflet global (`L`) is only exposed for the UMD version of Leaflet. The exposing of the global, and the `noConflict()` mode, is now also handled fully by Rollup, instead of our own custom implementation. A bug where another global called `leaflet` was accidentally exposed (also present in v1.x), is now fixed.

Other notable changes:
- In the ESM version, namespaces are now frozen, meaning extending Leaflet built-ins is no longer possible (e.g. `L.Util.foo = 'bar';`), the UMD version still allows this for backwards compatibility.
- The code generated by Rollup is compatible with ES2015 and up, and no longer includes ES5 compatible code (just like Leaflet itself).
- All relevant documentation and code referring to the Leaflet global has been migrated to exclusively use named imports.

Closes #9842


docs/examples/accessibility/example.md
```diff
@@ -4,7 +4,7 @@ title: Accessible Markers Example
 ---
 
 <script type="module">
-	import L, {Map, TileLayer, Marker, Icon} from 'leaflet';
+	import {Map, TileLayer, Marker} from 'leaflet';
```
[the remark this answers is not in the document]
jonkoops: What are you on about? You are telling me what I did is correct, that makes no sense. Only leave comments if there is actually something to fix.

docs/examples/overlays/example-video.md
```diff
@@ -47,7 +47,7 @@ title: Video Overlay Tutorial (Video with Controls)
 				const button = DomUtil.create('button');
 				button.title = 'Play';
 				button.innerHTML = '<span aria-hidden="true">▶️</span>';
-				L.DomEvent.on(button, 'click', () => {
+				DomEvent.on(button, 'click', () => {
```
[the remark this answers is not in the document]
jonkoops: There is no need to update the import, it is already imported. Previously this was using a global by accident.

build/rollup-config.js
```diff
@@ -10,25 +11,25 @@ const release = process.env.NODE_ENV === 'release';
...
-	input: 'src/LeafletWithGlobals.js',
+export default defineConfig({
+	input: 'src/Leaflet.js',
 	output: [
 		{
 			file: pkg.exports['.'],
 			format: 'es',
 			banner,
 			sourcemap: true,
-			freeze: false
```
Falke-Design, line 22: Why not keeping `freeze: false`?

src/LeafletWithGlobals.js
```diff
@@ -1,20 +0,0 @@
-import * as L from './Leaflet.js';
-export * from './Leaflet.js';
-
-export default L;
```
Falke-Design, line 4: I'm not sure, but I would keep exporting `L`

docs/examples/accessibility/example.md
Falke-Design, line 7: I still would export `L` for "debugging" purpose

docs/examples/accessibility/example.md
```diff
@@ -15,7 +15,4 @@ title: Accessible Markers Example
 
 	const marker = new Marker([50.4501, 30.5234], {alt: 'Kyiv'}).addTo(map)
 		.bindPopup('Kyiv, Ukraine is the birthplace of Leaflet!');
-
-	globalThis.L = L; // only for debugging in the developer console
-	globalThis.map = map; // only for debugging in the developer console
```
Falke-Design, line 20: I want to keep this for debugging, as I added it explicitly for that purpose

Falke-Design (changes requested):
I’m not particularly fond of these changes, as they significantly alter the behavior of the current alpha and affect compatibility with the plugins.

I would only change applying L to globalThis. In my view, everything else should be kept (except for the docs changes L.)

jonkoops:
Let's take this one at a time, otherwise we risk having a bunch of async discussion. Firstly, let's talk about the default `L` export, I am a firm believer we should get rid of it. It is a legacy that exists because historically, Leaflet was attached to a single global in a pre-module world. If we were to create a new library today, this would not be a pattern you'd see.

It also makes it significantly harder for tooling to do proper dead-code elimination trough [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking), once the default import is consumed, it is no longer possible for a bundler such as WebPack or Vite to determine which parts of the module are used, so it must assume that all of them are used, meaning the entirety of Leaflet is bundled. Some other dead code elimination methods can still apply, but the initial pass cannot be used to shake out dead code.

I don't think the fact that we're currently shipping alphas should factor in whether API changes can still be made, that is what an alpha is for. Once we get to a release candidate phase, we should strive to no longer make any changes that could impact API stability.

jonkoops:
In regard to tree-shaking, I also see a user already reported #9814 that confirms v2 in its current state is not tree-shakable.

Falke-Design:
> Firstly, let's talk about the default L export, I am a firm believer we should get rid of it. It is a legacy that exists because historically, Leaflet was attached to a single global in a pre-module world. If we were to create a new library today, this would not be a pattern you'd see.

Of course, the goal should be to remove the global L, but the number of users which are migrating to v2 will naturally decrease once we remove L. We are an old plugin with a large fan base, and we should not make it to hard for them.


> It also makes it significantly harder for tooling to do proper dead-code elimination trough [tree-shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking), once the default import is consumed, it is no longer possible for a bundler such as WebPack or Vite to determine which parts of the module are used, so it must assume that all of them are used, meaning the entirety of Leaflet is bundled. Some other dead code elimination methods can still apply, but the initial pass cannot be used to shake out dead code.

Tree Shaking is still possible if the bundler would accesses the build file without global L (if we are able to provide a UMD script with global L and a ESM script without L)

> I don't think the fact that we're currently shipping alphas should factor in whether API changes can still be made, that is what an alpha is for. Once we get to a release candidate phase, we should strive to no longer make any changes that could impact API stability.

Sure, but I'm against removing something that users would expect us to provide, especially since I already informed them that we plan to make the release in November. We have enough time to remove these things in another major release.

jonkoops:
Closing this PR, as it has too many conflicts. It could be used as a foundation for landing #10036

Issue #9872, opened 2025-09-08
Leaflet v2: Evented mixin seems broken for global L by recent changes (3rd party plugin)


indigoxela:
### Checklist

- [x] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [x] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

Using the (already heavily patched) Leaflet.draw plugin, which seems to rely on a (forgive it I got this wrong) Evented mixin, the plugin breaks. As far as I can tell beyond repair.

I spent quite some time debugging - I'm no JS developer, so some of my findings may be wrong. It seems like something like:
```
L.Toolbar.include(L.Evented.prototype);
```
in initialize method broke.
No more events available, which results in `Uncaught TypeError: this._toolbars[L.DrawToolbar.TYPE].on is not a function` and no draw toolbar shows up.

If I comment out anything related to events attached to draw toolbars, the buttons show up again, but only partly work, as the events are missing.

Note: This broke only recently, with changes made between v2.0.0-alpha and v2.0.0-alpha.1, but my JS skills are limited, so I'm not able to figure out what exactly caused it.
It must have been _some of the last commits before v2.0.0-alpha.1_, because a dev shortly before alpha.1 still worked for me.

### Expected behavior

With Leaflet-V1-polyfill in use, I had hoped for backwards compatibility. This may not be the case, though, so if it's not possible to fix it, that should be documented. Or I'd need some guidance for a workaround.

I don't mind to further patch the 3rd party lib - I just don't want to rewrite it from scratch. 😉 

### Current behavior

3rd party library completely broken, although the very latest commit of Leaflet-V1-polyfill in use.

### Minimal example reproducing the issue

Sorry, no demo, yet, as I don't even know if this is supposed to be broken or not.

### Environment

- Leaflet version: leaflet-2.0.0-alpha.1
- Browser (with version): any
- OS/Platform (with version): (seems irrelevant) Linux
[...]

Issue #9889, opened 2025-09-16
Use JavaScript `constructor()` instead of custom `initialize()`


jonkoops:
Currently, all classes use the custom `initalize()` method left over from Leaflet v1 to handle constructors. However, JavaScript classes have their own standardized `constructor()` that should be used instead.

This is a stripped down version of the `Class` initialization logic as it exists in Leaflet right now:

```js
class Class {
	static addInitHook(fn) {
		if (!Object.hasOwn(this.prototype, '_initHooks')) {
			this.prototype._initHooks = [];
		}
		this.prototype._initHooks.push(fn);
		return this;
	}

	constructor(...args) {
		this._initHooksCalled = false;

		// call the constructor
		if (this.initialize) {
			this.initialize(...args);
		}

		// call all constructor hooks
		this.callInitHooks();
	}

	callInitHooks() {
		if (this._initHooksCalled) {
			return;
		}

		// collect all prototypes in chain
		const prototypes = [];
		let current = this;

		while ((current = Object.getPrototypeOf(current)) !== null) {
			prototypes.push(current);
		}

		// reverse so the parent prototype is first
		prototypes.reverse();

		// call init hooks on each prototype
		for (const proto of prototypes) {
			for (const hook of proto._initHooks ?? []) {
				hook.call(this);
			}
		}

		this._initHooksCalled = true;
	}
}
```

In summary when a class is constructed the following happens:

- If an `initialize()` method exists, it is called with the constructor arguments.
- If any hooks were registered trough `addInitHook()`, they will be executed, starting with the superclass, working upwards to the subclass.

For example:

```js
class Vehicle extends Class {
	initialize(brand) {
		console.log('Vehicle constructor called.');
		this.brand = brand;
	}
}

Vehicle.addInitHook(function () {
	console.log(`Vehicle init hook called with brand: ${this.brand}`);
});

// Level 2: Inherits from Vehicle
class Car extends Vehicle {
	initialize(brand, model) {
		console.log('Car constructor called.');
		Vehicle.prototype.initialize.call(this, brand);
		this.model = model;
	}
}

Car.addInitHook(function () {
	console.log(`Car init hook called with model: ${this.model}`);
});

// Level 3: Inherits from Car
class SportsCar extends Car {
	initialize(brand, model, topSpeed) {
		console.log('SportsCar constructor called.');
		Car.prototype.initialize.call(this, brand, model);
		this.topSpeed = topSpeed;
	}
}

[...]

PR #9891, opened 2025-09-16
ClassSpec: test "correct parent initialize function"

simon04:
Unit test for #9872 and #9901.

### DIFF
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -44,6 +44,7 @@ describe('Class', () => {
 					initialize();
 				}
 			});
+			Toolbar.include(Evented.prototype);
 			const DrawToolbar = Toolbar.extend({
 				initialize(options) {
 					initializeDraw();
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -35,6 +35,45 @@ describe('Class', () => {
 			expect(method.called).to.be.true;
 		});
 
+		it('calls the correct parent initialize function', () => {
+			const initialize = sinon.spy();
+			const initializeDraw = sinon.spy();
+			const initializeEdit = sinon.spy();
+			const Toolbar = Class.extend({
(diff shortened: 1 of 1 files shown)


spec/suites/core/ClassSpec.js
```diff
@@ -35,6 +35,45 @@ describe('Class', () => {
...
+			initialize.resetHistory();
+
+			new DrawToolbar();
+			sinon.assert.callOrder(initializeDraw, initialize);
+			sinon.assert.notCalled(initializeEdit);
+			initialize.resetHistory();
+			initializeDraw.resetHistory();
+
+			new EditToolbar();
+			sinon.assert.callOrder(initializeEdit, initialize);
```
jonkoops, line 74: Is this not testing the exact same thing as with `DrawToolbar` above?

jonkoops:
I am wondering if we should also add tests for the execution order of the init hooks (as described in #9889).

simon04:
I've added `Toolbar.include(Evented.prototype)` in order to test for the bugfix #9901. @Falke-Design

Falke-Design (approved):

spec/suites/core/ClassSpec.js
Falke-Design: That makes sure that after calling `Toolbar.prototype.initialize.call(this, options);` in DrawToolbar, `initialize` of EditToolbar is still called.

See #9900

jonkoops: I don't see `include()` being called here as mentioned in #9900, so why would the bug occur in this test? Both `DrawToolbar` and `EditToolbar` have the exact same internal logic and the exact same assertions.

simon04: Line 47 contains `Toolbar.include(Evented.prototype)`

jonkoops: Ahhh, totally missed it! Thanks!

PR #9804, opened 2025-07-23
v2-alpha: add in additional export to prevent shadow of global "Map

willfarrell:
this prevent shadow of global “Map” property

Closes: #9803

PR #9892, opened 2025-09-17
Rename `LeafletMap` with alias `Map`

simon04:
Avoid confusion in JavaScript debugger for LeafletMap instances.

Follow-up: #9804

### DIFF
--- a/docs/reference-2.0.0.html
+++ b/docs/reference-2.0.0.html
@@ -138,7 +138,7 @@ bodyclass: api-page
 
 
 <pre><code class="language-js">// initialize the map on the &quot;map&quot; div with a given center and zoom
-const map = new Map('map', {
+const map = new LeafletMap('map', {
 	center: [51.505, -0.09],
 	zoom: 13
 });
@@ -14873,12 +14873,12 @@ Inherits <a href="#renderer"><code>Renderer</code></a>.</p>
 
 
 <p>Use SVG by default for all paths in the map:</p>
-<pre><code class="language-js">const map = new Map('map', {
+<pre><code class="language-js">const map = new LeafletMap('map', {
 	renderer: new SVG()
 });
 </code></pre>
 <p>Use a SVG renderer with extra padding for specific vector geometries:</p>
-<pre><code class="language-js">const map = new Map('map');
+<pre><code class="language-js">const map = new LeafletMap('map');
 const myRenderer = new SVG({ padding: 0.5 });
 const line = new Polyline( coordinates, { renderer: myRenderer } );
 const circle = new Circle( center, { renderer: myRenderer, radius: 100 } );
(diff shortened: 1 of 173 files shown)


jonkoops (commented):
We should also update the documentation to use `LeafletMap` if we want to do this, as well as all the test and examples. Are we sure we want to change the default name? I would also be ok with recommending users handle conflicts with `Map` themselves:

```js
import { Map as LeafletMap } from 'leaflet';
```

simon04:
> Are we sure we want to change the default name?

I think we should. Due to the name clash, rollup transforms the code to `let Map$1 = class Map extends Evented`. As a consequence, JavaScript debugger shows instances of a Leaflet Map as `Map$1` (which lacks any clue that this is related to Leaflet).

src/map/Map.js
```diff
@@ -12,7 +12,7 @@ import {PosAnimation} from '../dom/PosAnimation.js';
 import * as PointerEvents from '../dom/DomEvent.PointerEvents.js';
 
 /*
- * @class Map
+ * @class LeafletMap
```
Falke-Design, line 15: we should mention the alias `Map` too

Falke-Design:
What if we replace Map completly? We destroy currently the plugin compatibility anyway

simon04:
> What if we replace Map completly?

Originally, I found this idea too crazy. After giving it a 2nd and 3rd thought, I like it. Having a canonical way of instantiating objects is in line with the completed removal of factory functions. And I assume any code base has only very few `new LeafletMap` constructor calls.

NB: OpenLayers still has `class Map extends BaseObject {}`

jonkoops:
@Leaflet/core any strong opinions about this? I personally don't mind leaning either way, but perhaps we don't want to make breaking changes here.

Falke-Design:
Let us keep Map as alias but rename it to LeafletMap. 

Map should be mentioned in docs and also in code as deprecated

src/map/Map.js
```diff
@@ -48,7 +48,7 @@ import * as PointerEvents from '../dom/DomEvent.PointerEvents.js';
 // @constructor LeafletMap(el: HTMLElement, options?: LeafletMap options)
```
Falke-Design: We need to document `Map` to but mark it as deprecated

Falke-Design (changes requested):
@simon04 can you please adjust

Falke-Design:
All used `Map` implementations in `debug/` and `spec/` need to be changed too

docs/reference-2.0.0.html
Falke-Design, line 1: Please revert the change here, because this file is auto generated

docs/reference.html
Falke-Design, line 1: Please revert the change here, because this file is auto generated

Falke-Design (changes requested):

spec/suites/map/handler/Map.DragSpec.js
Falke-Design, line 56: Map import is missing. Should be now LeafletMap

Falke-Design (approved):

gtyu4013-eng:
issues

PR #9894, opened 2025-09-17
Remove Event.returnValue fallback

simon04:
Event.preventDefault is baseline widely available, see https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault.

Event.returnValue is deprecated, see https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -196,7 +196,7 @@ export function disableClickPropagation(el) {
 // with page reload when a `<form>` is submitted).
 // Use it inside listener functions.
 export function preventDefault(e) {
-	e.preventDefault();
+	e.preventDefault?.();
 	return this;
 }
 


src/dom/DomEvent.js
```diff
@@ -196,11 +196,7 @@ export function disableClickPropagation(el) {
 // with page reload when a `<form>` is submitted).
 // Use it inside listener functions.
 export function preventDefault(e) {
-	if (e.preventDefault) {
-		e.preventDefault();
-	} else {
-		e.returnValue = false;
-	}
+	e.preventDefault();
```
jonkoops: At this point, perhaps we should consider getting rid of `DomEvent.preventDefault()` entirely, and call `preventDefault()` directly on the `Event`.

src/dom/DomEvent.js
```diff
@@ -190,24 +190,10 @@ export function disableClickPropagation(el) {
...
-		e.returnValue = false;
-	}
-	return this;
-}
-
 // @function stop(ev: DOMEvent): this
 // Does `stopPropagation` and `preventDefault` at the same time.
 export function stop(e) {
-	preventDefault(e);
+	e.preventDefault?.();
```
jonkoops: In which cases would a value be passed here that is not of type `Event`? It seems that the type for `e` is probably wrong, since `DOMEvent` is not a class, nor a type, but a module. Perhaps it was supposed to be `Event`? In which case, I don't see a reason to call `preventDefault()` conditionally, as it is always present. 
```suggestion
	e.preventDefault();
```

WDYT?

src/control/Control.Layers.js
```diff
@@ -430,10 +430,11 @@ export class Layers extends Control {
 	_expandSafely() {
 		const section = this._section;
 		this._preventClick = true;
-		DomEvent.on(section, 'click', DomEvent.preventDefault);
```
jonkoops, line 433: I hadn't considered that `DOMEvent.preventDefault` was used as a simple named alias for easily unsubscribing events like this. Perhaps it does serve a purpose in that way, so we should keep it?

@Leaflet/core @Leaflet/leaflet-maintainers WDYT?

src/dom/DomEvent.js
simon04: This breaks the following tests: 

```
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup toggles its visibility when marker is clicked FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at Marker._openPopup (src/layer/Popup.js:483:12)
            at Marker.fire (src/core/Events.js:186:9)
            at Map._fireDOMEvent (src/map/Map.js:1466:6)
            at Map._handleDOMEvent (src/map/Map.js:1421:8)
            at HTMLDivElement.handler (src/dom/DomEvent.js:106:13)
            at Object._dispatchEvent (node_modules/ui-event-simulator/ui-event-simulator.js:172:13)
            at Object.fire (node_modules/ui-event-simulator/ui-event-simulator.js:268:10)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:84:20)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup it should use a popup with a function as content with a FeatureGroup FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at FeatureGroup._openPopup (src/layer/Popup.js:483:12)
            at FeatureGroup.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:104:9)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup it should function for popup content after bindPopup is called FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at FeatureGroup._openPopup (src/layer/Popup.js:483:12)
            at FeatureGroup.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:134:9)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup should use a function for popup content when a source is passed to Popup FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at Marker._openPopup (src/layer/Popup.js:483:12)
            at Marker.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:158:10)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup prevents an underlying map click for Layer FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at Polygon._openPopup (src/layer/Popup.js:483:12)
            at Polygon.fire (src/core/Events.js:186:9)
            at Map._fireDOMEvent (src/map/Map.js:1466:6)
            at Map._handleDOMEvent (src/map/Map.js:1421:8)
            at HTMLDivElement.handler (src/dom/DomEvent.js:106:13)
            at Object._dispatchEvent (node_modules/ui-event-simulator/ui-event-simulator.js:172:13)
            at Object.fire (node_modules/ui-event-simulator/ui-event-simulator.js:268:10)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:296:20)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup can open a popup with enter keypress when marker has focus FAILED
        TypeError: e.preventDefault is not a function
            at Module.stop (src/dom/DomEvent.js:196:4)
            at Marker._openPopup (src/layer/Popup.js:483:12)
            at Marker._onKeyPress (src/layer/Popup.js:506:9)
            at Marker.fire (src/core/Events.js:186:9)
            at Map._fireDOMEvent (src/map/Map.js:1466:6)
            at Map._handleDOMEvent (src/map/Map.js:1421:8)
            at HTMLDivElement.handler (src/dom/DomEvent.js:106:13)
            at Object._dispatchEvent (node_modules/ui-event-simulator/ui-event-simulator.js:172:13)
            at Object.fire (node_modules/ui-event-simulator/ui-event-simulator.js:268:10)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:306:20)
  Chrome Headless 140.0.0.0 (Mac OS 10.15.7) ERROR
  Some of your tests did a full page reload!
```

Falke-Design:
~I have in mind that for Leaflet custom events we added a property there to prevent it in a later execution. This fits to the error that `e.preventDefault` is undefined.~

Edit: I mixed it up with `_leaflet_disable_click`

src/dom/DomEvent.js
Falke-Design: This happens for all dom events which are wrapped to `originalEvent` (Map, Draggable):
https://github.com/Leaflet/Leaflet/blob/72086841b659e184b37254753b60bc82df0d5d26/src/map/Map.js#L1453-L1455

https://github.com/Leaflet/Leaflet/blob/72086841b659e184b37254753b60bc82df0d5d26/src/map/Map.js#L1466

This results in 
```
e.preventDefault == undefined
e.originalEvent.preventDefault == function
```

src/control/Control.Layers.js
Falke-Design: Both is fine for me

src/map/handler/Map.TapHold.js
```diff
@@ -47,7 +47,7 @@ export class TapHold extends Handler {
 			if (!this._isTapValid()) { return; }
 
 			// prevent simulated mouse events https://w3c.github.io/touch-events/#mouse-events
-			DomEvent.on(document, 'pointerup', DomEvent.preventDefault);
+			DomEvent.on(document, 'pointerup', DomEvent.stop);
```
Falke-Design: Are we sure that this doesn't change the behavior?

src/control/Control.Layers.js
jonkoops: My vote would be to keep it around in that case, it is actually useful, and I see no need to break public APIs if we can avoid it.

src/map/handler/Map.TapHold.js
jonkoops: Pretty sure it does, let's keep it the way it was before.

src/control/Control.Layers.js
Falke-Design: Sounds good

src/dom/DomEvent.js
```diff
@@ -190,24 +190,10 @@ export function disableClickPropagation(el) {
...
-// with page reload when a `<form>` is submitted).
-// Use it inside listener functions.
-export function preventDefault(e) {
-	if (e.preventDefault) {
-		e.preventDefault();
-	} else {
-		e.returnValue = false;
-	}
-	return this;
-}
```
Falke-Design, line 205: @simon04 please keep the method DomEvent.preventDefault around for quick usage in event listeners. `DomEvent.on(document, 'pointerup', DomEvent.preventDefault);`

The usages which can be replaced with `e.preventDefault()` are fine.

jonkoops:
I want to take a quick look at the cases where `preventDefault()` is called conditionally. But this is likely good to go.

jonkoops:
So what I am confused about is why we are passing things that are not `Event` to this function in the first place. The test failures in https://github.com/Leaflet/Leaflet/pull/9894#discussion_r2362603177 seem to indicate a larger issue that raises questions for me:

1. Should we have been calling `.preventDefault()` on `originalEvent` if present? (e.g. we do have some special handling for [stopping events](https://github.com/Leaflet/Leaflet/blob/fd3001ebf5c8c10d0e47616877b30f47f0b4dcc3/src/dom/DomEvent.js#L168-L170)
2. Why does this `originalEvent` even exist in the first place? Could we possibly get rid of it?
3. Is it even the responsibility of `DomUtil.preventDefault()` to check for the existence of `preventDefault()`, or should it have never been called in the first place? (perhaps this should have been handled in `DomEvent.stop()`?)

@Leaflet/core @Leaflet/leaflet-maintainers any insights into this?

Either way, I am approving this PR as this discussion is out of scope for what is being delivered here, so feel free to merge. I would still like to get some answers.

jonkoops (approved):

src/dom/DomEvent.js
jonkoops: Discussion continued here https://github.com/Leaflet/Leaflet/pull/9894#issuecomment-3328127823

PR #9895, opened 2025-09-17
Remove Event.cancelBubble fallback

simon04:
Event.cancelBubble is deprecated, see https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelBubble

### DIFF
--- a/src/dom/DomEvent.js
+++ b/src/dom/DomEvent.js
@@ -163,12 +163,13 @@ function removeOne(obj, type, fn, context, id) {
 // ```
 export function stopPropagation(e) {
 
-	if (e.stopPropagation) {
+	if (e instanceof Event) {
 		e.stopPropagation();
 	} else if (e.originalEvent) {  // In case of Leaflet event.
 		e.originalEvent._stopped = true;
 	} else {
-		e.cancelBubble = true;
+		// custom Leaflet event
+		console.debug('Unsupported event', e);
 	}
 
 	return this;


src/dom/DomEvent.js
```diff
@@ -163,12 +163,13 @@ function removeOne(obj, type, fn, context, id) {
 // ```
 export function stopPropagation(e) {
 
-	if (e.stopPropagation) {
+	if (e instanceof Event) {
 		e.stopPropagation();
 	} else if (e.originalEvent) {  // In case of Leaflet event.
 		e.originalEvent._stopped = true;
 	} else {
```
jonkoops, line 170: Should we not be removing this branch entirely?

src/dom/DomEvent.js
```diff
@@ -163,12 +163,13 @@ function removeOne(obj, type, fn, context, id) {
...
 
-	if (e.stopPropagation) {
+	if (e instanceof Event) {
 		e.stopPropagation();
 	} else if (e.originalEvent) {  // In case of Leaflet event.
 		e.originalEvent._stopped = true;
 	} else {
-		e.cancelBubble = true;
+		// custom Leaflet event
+		console.debug('Unsupported event', e);
```
Falke-Design, line 172:
```suggestion
		console.warn('Unsupported event', e);
```

simon04: This will trigger three warnings when running the test suite. Maybe we should "fix" the relevant code first?

```
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup it should use a popup with a function as content with a FeatureGroup FAILED
        console.error Unsupported event
            at stopPropagation (src/dom/DomEvent.js:172:9)
            at Module.stop (src/dom/DomEvent.js:212:2)
            at FeatureGroup._openPopup (src/layer/Popup.js:483:12)
            at FeatureGroup.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:104:9)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup it should function for popup content after bindPopup is called FAILED
        console.error Unsupported event
            at stopPropagation (src/dom/DomEvent.js:172:9)
            at Module.stop (src/dom/DomEvent.js:212:2)
            at FeatureGroup._openPopup (src/layer/Popup.js:483:12)
            at FeatureGroup.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:134:9)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Popup should use a function for popup content when a source is passed to Popup FAILED
        console.error Unsupported event
            at stopPropagation (src/dom/DomEvent.js:172:9)
            at Module.stop (src/dom/DomEvent.js:212:2)
            at Marker._openPopup (src/layer/Popup.js:483:12)
            at Marker.fire (src/core/Events.js:186:9)
            at Context.<anonymous> (spec/suites/layer/PopupSpec.js:158:10)
```

Falke-Design: That is because they are just fake events for testing. We could add the property  `originalEvent: {}` to them

Falke-Design: You will get a error for this events for the changes in #9894 too

jonkoops: Couple of notes, first off, I think we can just go ahead and `throw` here if an event is invalid, a [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) seems most appropriate. Secondly, let's fix those tests by constructing an actual `Event` object.

PR #9897, opened 2025-09-19
TileLayer: default to strict-origin-when-cross-origin referrer-policy

IvanSanchez:
Sets the default value of the TileLayer referrerPolicy to `strict-origin-when-cross-origin` (which should already be the default anyway).

This is an alternative to https://github.com/Leaflet/Leaflet/pull/9883/

Fixes: #10156

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -78,13 +78,13 @@ export class TileLayer extends GridLayer {
 			// Refer to [CORS Settings](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) for valid String values.
 			crossOrigin: false,
 
-			// @option referrerPolicy: Boolean|String = false
-			// Whether the referrerPolicy attribute will be added to the tiles.
-			// If a String is provided, all tiles will have their referrerPolicy attribute set to the String provided.
-			// This may be needed if your map's rendering context has a strict default but your tile provider expects a valid referrer
-			// (e.g. to validate an API token).
-			// Refer to [HTMLImageElement.referrerPolicy](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy) for valid String values.
-			referrerPolicy: false
+			// @option referrerPolicy: String = 'strict-origin-when-cross-origin'
+			// Sets the tiles' `referrerPolicy` attribute set to the String provided. See
+			// [`HTMLImageElement.referrerPolicy`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/referrerPolicy) for possible values.
+			//
+			// Tile providers might require you to use a specific referrer policy (e.g. to verify API tokens or prevent abuse).
+			// Specifically, if you are using tiles from `tile.openstreetmap.org` or `tile.osm.org`, you [should not change this value](https://operations.osmfoundation.org/policies/tiles/).
+			referrerPolicy: 'strict-origin-when-cross-origin'
 		});
 	}
 
@@ -93,14 +93,19 @@ export class TileLayer extends GridLayer {
 
 		this._url = url;
 
-		// in case the attribution hasn't been specified, check for known hosts that require attribution
-		if (this.options.attribution === null) {
-			const urlHostname = new URL(url, location.href).hostname;
+		options = Util.setOptions(this, options);
 
-			// check for Open Street Map hosts
-			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
-			if (osmHosts.some(host => urlHostname.endsWith(host))) {
-				this.options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
+		// Set required OpenStreetMap attribution if it hasn't been specified; upgrade to HTTPS so the referrer policy works.
+		const tileUrl = new URL(this._url, location.href);
+		const urlHostname = tileUrl.hostname;
+		const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
+		if (osmHosts.some(host => urlHostname.endsWith(host))) {
+			if (options.attribution === null) {
+				options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
+			}
+			if (tileUrl.protocol === 'http:') {
+				tileUrl.protocol = 'https:';
+				this._url = tileUrl.toString();
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


IvanSanchez:
I feel a tiny bit slightly better setting the default to `strict-origin-when-cross-origin` for **all** tile domains instead of just for OSM. After all, as we discussed at #9883, this *should* be already the default.

cc @Firefishy @tomhughes @jonkoops @Falke-Design

src/layer/tile/TileLayer.js
```diff
@@ -94,14 +94,14 @@ export class TileLayer extends GridLayer {
...
-		if (options.attribution === null && URL.canParse(url)) {
+		// Set required OpenStreetMap attribution if it hasn't been specified; upgrade to HTTPS so the referrer policy works.
+		if (URL.canParse(url)) {
 			const urlHostname = new URL(url).hostname;
-
-			// check for Open Street Map hosts
-			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
-			if (osmHosts.some(host => urlHostname.endsWith(host))) {
-				options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
+			if ((/^tile\.(openstreetmap|osm)\.org$/i).test(urlHostname)) {
```
jonkoops: Honestly. I prefer a list of origins, as that is much more readable than a regex, but that is my personal opinion.

src/layer/tile/TileLayer.js
```diff
@@ -94,14 +94,14 @@ export class TileLayer extends GridLayer {
...
+				if (options.attribution === null) {
+					options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
+				}
+				this._url = url.replace(/^http:\/\//, 'https://');
```
jonkoops: Are we upgrading HTTP protocol to HTTPS implicitly here? Is that needed? If we already parse the URL, perhaps we can set the [`protocol`](https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol) of it instead?

jonkoops (commented):
Generally agree with just making this the default, just a couple of notes/questions in the implementation.

src/layer/tile/TileLayer.js
jonkoops: I see the discussion in #9883 now, upgrading to HTTPS seems fine to me, my point about using `URL` still stands. Perhaps we should also log a warning to the user.

Falke-Design (dismissed):
I also prefer a array of url.
Also a warning about changing to https makes sense in my eyes

Falke-Design:
I prefer this over #9883

Firefishy:
> I prefer this over #9883

I don't mind which gets accepted. The outcome is the same for me.

src/layer/tile/TileLayer.js
Falke-Design: I just run a performance test and the URL protocol usage was 50% slower.

```
// Setup
let str = 'http://tile.openstreetmap.org'; 
let u = new URL(str); 

// Test 1
let url = str.replace(/^http:\/\//, 'https://');

// Test 2 - 50% slower
u.protocol = 'https:'; 
let url = u.toString();
```

src/layer/tile/TileLayer.js
```diff
@@ -94,17 +94,15 @@ export class TileLayer extends GridLayer {
...
+		const urlHostname = new URL(url, location.href).hostname;
+		if ((/^tile\.(openstreetmap|osm)\.org$/i).test(urlHostname)) {
+			if (options.attribution === null) {
 				options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
 			}
-		}
-
+			this._url = url.replace(/^http:\/\//, 'https://');
+    }
+
```
Falke-Design:
```suggestion
		}
```

src/layer/tile/TileLayer.js
```diff
@@ -94,17 +94,14 @@ export class TileLayer extends GridLayer {
...
-			// check for Open Street Map hosts
-			const osmHosts = ['tile.openstreetmap.org', 'tile.osm.org'];
-			if (osmHosts.some(host => urlHostname.endsWith(host))) {
+		// Set required OpenStreetMap attribution if it hasn't been specified; upgrade to HTTPS so the referrer policy works.
+		const urlHostname = new URL(url, location.href).hostname;
+		if ((/^tile\.(openstreetmap|osm)\.org$/i).test(urlHostname)) {
+			if (options.attribution === null) {
 				options.attribution = '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors';
 			}
+			this._url = url.replace(/^http:\/\//, 'https://');
```
Falke-Design:
```suggestion
			if (url.startsWith('http://')) {
				this._url = url.replace(/^http:\/\//, 'https://');
				console.warn('Tile requests (tile.openstreetmap.org / tile.osm.org) have been switched to HTTPS to ensure correct referrer-policy behavior.');
			}
```

@jonkoops like that?

jonkoops: Something like that, yes. Perhaps we can even clean up this code by using `URL()` and then changing the `protocol` property.

Falke-Design: > I just run a performance test and the URL protocol usage was 50% slower.
> 
> ```
> // Setup
> let str = 'http://tile.openstreetmap.org'; 
> let u = new URL(str); 
> 
> // Test 1
> let url = str.replace(/^http:\/\//, 'https://');
> 
> // Test 2 - 50% slower
> u.protocol = 'https:'; 
> let url = u.toString();
> ```

https://github.com/Leaflet/Leaflet/pull/9897#discussion_r2596262344

jonkoops: Commented there.

src/layer/tile/TileLayer.js
jonkoops: Does the performance matter here? The URL will be parsed once for each `TileLayer`, I don't see users adding thousands of tile layers each second.

Falke-Design: Nah, it doesn’t really matter. Both solutions work for me so I was wondering which one has better performance. If you’d prefer to use the URL approach, feel free to update the PR so it can be merged.

jonkoops: I can see if I can dedicate some time to this after the 2.x release. Right now I want to focus on getting a release candidate out of the window.

simon04:
Hi @jonkoops, to me it seems that the discussion has converged towards using `new URL` (as performance is not an issue), and not using regex, right? So this PR could be finished rather easily...

jonkoops:
Yes, that would be my take.

Falke-Design:
@IvanSanchez @simon04 @jonkoops please review again

mourner (approved):
This looks good to me, let's land this.

Issue #9900, opened 2025-09-20
Class.include and extends is not working properly in v2


Falke-Design:
Follow up: #9872

```
const ClassExtended = Class.extend({
	initialize() {
		console.log('ClassExtended init');
		ClassExtended.include(Evented.prototype); // <-- This line causes the problem
	}
});

const ClassExt1 = ClassExtended.extend({
	initialize() {
		console.log('ClassExt1 init');
		ClassExtended.prototype.initialize.call(this);
	}
});
const ClassExt2 = ClassExtended.extend({
	initialize() {
		console.log('ClassExt2 init');
		ClassExtended.prototype.initialize.call(this);
	}
});

new ClassExt1();
new ClassExt2();
```

current behavior:
```
	ClassExt1 init
	ClassExtended init
	ClassExt2 init
```

Expected:
```
	ClassExt1 init
	ClassExtended init
	ClassExt2 init
	ClassExtended init
```

PR #9901, opened 2025-09-20
Remove initialize placeholder in Class

Falke-Design:
Fixes: #9900

The problem is that when we call `ClassExtended.include(Evented.prototype);`, it overwrites 
`ClassExtended.initialize` with `Evented.initialize`, which contains no code. This happens 
because `Evented` extends from `Class` but does not define its own `initialize` method.

So everything worked as "expected".

```
const ClassExtended = Class.extend({
	initialize() {
		console.log('ClassExtended init');
		ClassExtended.include(Evented.prototype); // <-- This line causes the problem
	}
});
```

I added the code [here](https://github.com/Leaflet/Leaflet/commit/47167af0e4676a6c09eb0752e6015c1df6687302#diff-8e4854ab0e220475dc22a076300f7f19f24b4480d852e02177dfd05826e4be5d) but it was just for documentation. 

@simon04 please review

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -115,11 +115,6 @@ export class Class {
 		this.callInitHooks();
 	}
 
-	initialize(/* ...args */) {
-		// Override this method in subclasses to implement custom initialization logic.
-		// This method is called automatically when a new instance of the class is created.
-	}
-
 	callInitHooks() {
 		if (this._initHooksCalled) {
 			return;


simon04 (approved):
Amazing. Great that you found the root cause for this problem.

At some point, a new alpha release would be handy for testing. :)

Falke-Design:
> At some point, a new alpha release would be handy for testing. :)

@simon04 Yes, I would like to make one too. But since @jonkoops wants to break backward compatibility, there won’t be a new alpha until he has committed his changes (#9842, #9889, ...). To be honest, I also don’t think he will be able to release it in November anymore.

jonkoops:
I'd like to avoid breaking backwards compatibility where possible, but some things might be unavoidable. I'll try to get my changes landed ASAP, so we can start tagging some new alphas.

PR #9902, opened 2025-09-21
Use super to access methods on superclass

simon04:
Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super


jonkoops (approved):
Great little enhancement!

PR #9906, opened 2025-09-23
Use standard `constructor()` instead of `initialize()`

### LINKED ISSUE #9889: Use JavaScript `constructor()` instead of custom `initialize()`

jonkoops:
Changes all built-in Leaflet classes from the custom `initialize()` method to a standardized `constructor()`. I have tried to maintain backwards compatibility with existing Leaflet code as much as possible, but there were some cases where breaking changes could not be avoided (see below).

Previously, we were using the `constructor()` of the super-most class (`Class`) to run any initialization hooks, however due to fundamental incompatibilities (see #9889) of mixing `constructor()` and `initialize()`, a new mechanism for running initialization hooks had to be introduced.

The new implementation makes use of a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to intercept the constructor of any class wrapped by the new `withInitHooks()` function. Once intercepted, the class will be constructed [using reflection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) upon instantiation, after which the hooks registered with `addInitHook()` will be trigged, exactly in the same manner as they were before.

Wrapping classes that require initialization hooks with `withInitHooks()` is required, because there is no way to hook into the [`extends` keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/extends), leaving us otherwise unable to intercept their constructor. For this reason, all built-in Leaflet classes that extend from `Class` are wrapped manually, ensuring some degree of backwards compatibility:

```js

// Remains functional due to being wrapped with `withInitHooks()` by default.
Map.addInitHook(function () {});
```

### Missing `initialize()` method on Leaflet built-ins

Since this change replaces all `initialize()` methods from the built-in Leaflet classes with a standard `constructor()`, it is no longer possible to call this method as if it were the constructor:

```js
import { Marker } from 'leaflet';

class MyMarker extends Marker {
  initialize(latlng, options) {
    // Though `initialize()` will still be called, it is no longer possible to call `initialize()`
    // on the super-class, as it no longer exists.
    Marker.prototype.initialize.call(this, latlng, options);
  }
}

// Instead a `constructor` must be used by `MyMarker`.
class MyMarker extends Marker {
  constructor(latlng, options) {
    super(latlng, options);
  }
}
```

Or, with 'legacy' classes using `Class#extend()`:

```js
import { Marker } from 'leaflet';

const MyMarker = Marker.extend({
  initialize(latlng, options) {
    // Also here the `initialize()` method does not exist on the super class.
    Marker.prototype.initialize.call(this, latlng, options);
  }
});

// Instead a `constructor` must be used by `MyMarker`.
const MyMarker = Marker.extend({
  constructor(latlng, options) {
    super.constructor(latlng, options);
  }
});
```

As far as I can tell, there is no way to preserve the old behavior here, and this will be a necessary breaking change, but I am open to suggestions. Perhaps we could somehow patch classes that use the `Class#extend()` method, but I have not yet found a way to do so without introducing weird bugs or edge cases.

### No initialization hooks for classes using `extends` keyword

Extending classes using the `extends` keyword no longer guarantees initialization hooks will be triggered:

```js
import { Class } from 'leaflet';

class Foo extends Class {}

// Will throw an error due to the inability to intercept the
// constructor of `Foo` to execute the initialization hooks.
Foo.addInitHook(function () {});
```

To fix this, users can wrap their classes manually with `withInitHooks()`:

```js
import { Class, withInitHooks } from 'leaflet';

const Foo = withInitHooks(class Foo extends Class {});

// Will work, as we can intercept the constructor with `withInitHooks()`.
Foo.addInitHook(function () {});

// There is also no requirement to extend `Class`, calling `withInitHooks()` on
// a regular class will also work.
const Bar = withInitHooks(class Bar {});
```

The 'legacy' classes using `Class#extend()` are automatically wrapped with `withInitHooks()`, and will therefore keep working as expected:

```js
import { Class } from 'leaflet';

const Foo = Class.extend({});

// Will work, as we can intercept the constructor when calling `Class#extend()`.
Foo.addInitHook(function () {});
```

Closes #9889

### DIFF
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -1,400 +1,443 @@
 import {expect} from 'chai';
-import {Class, Evented} from 'leaflet';
+import {Class, Evented, withInitHooks} from 'leaflet';
 import sinon from 'sinon';
 
 describe('Class', () => {
-	describe('#extend', () => {
-		let Klass,
-		props,
-		constructor,
-		method;
+	it('sets default options', () => {
+		const instance = new Class();
+		expect(Object.hasOwn(instance, 'options')).to.be.true;
+		expect(instance.options).to.eql({});
+	});
 
-		beforeEach(() => {
-			constructor = sinon.spy();
-			method = sinon.spy();
+	it('calls \'initialize()\' automatically with constructor arguments', () => {
+		const spy = sinon.spy();
+		const warnSpy = sinon.spy(console, 'warn');
 
-			props = {
-				statics: {bla: 1},
-				includes: {mixin: true},
-
-				initialize: constructor,
-				foo: 5,
-				bar: method
-			};
-			Klass = Class.extend(props);
-		});
-
-		it('creates a class with the given constructor & properties', () => {
-			const a = new Klass();
-
-			expect(constructor.called).to.be.true;
-			expect(a.foo).to.eql(5);
-
-			a.bar();
-
-			expect(method.called).to.be.true;
-		});
-
-		it('calls the correct parent initialize function', () => {
-			const initialize = sinon.spy();
-			const initializeDraw = sinon.spy();
-			const initializeEdit = sinon.spy();
-			const Toolbar = Class.extend({
-				initialize() {
-					initialize();
-				}
-			});
-			Toolbar.include(Evented.prototype);
-			const DrawToolbar = Toolbar.extend({
-				initialize(options) {
-					initializeDraw();
-					Toolbar.prototype.initialize.call(this, options);
-				}
-			});
-			const EditToolbar = Toolbar.extend({
-				initialize(options) {
-					initializeEdit();
-					Toolbar.prototype.initialize.call(this, options);
-				}
-			});
-
-			new Toolbar();
-			sinon.assert.calledOnce(initialize);
-			sinon.assert.notCalled(initializeDraw);
-			sinon.assert.notCalled(initializeEdit);
-			initialize.resetHistory();
-
-			new DrawToolbar();
-			sinon.assert.callOrder(initializeDraw, initialize);
-			sinon.assert.notCalled(initializeEdit);
-			initialize.resetHistory();
-			initializeDraw.resetHistory();
-
-			new EditToolbar();
-			sinon.assert.callOrder(initializeEdit, initialize);
-			sinon.assert.notCalled(initializeDraw);
-		});
-
-		it('inherits parent classes\' constructor & properties', () => {
-			const Klass2 = Klass.extend({baz: 2});
-
-			const b = new Klass2();
-
-			expect(b instanceof Klass).to.be.true;
-			expect(b instanceof Klass2).to.be.true;
-
-			expect(constructor.called).to.be.true;
-			expect(b.baz).to.eql(2);
-
-			b.bar();
-
-			expect(method.called).to.be.true;
-		});
-
-		it('does not modify source props object', () => {
-			expect(props).to.eql({
-				statics: {bla: 1},
-				includes: {mixin: true},
-
-				initialize: constructor,
-				foo: 5,
-				bar: method
-			});
-		});
-
-		it('supports static properties', () => {
-			expect(Klass.bla).to.eql(1);
-		});
-
-		it('does not merge \'statics\' property itself', () => {
-			expect('statics' in Klass.prototype).to.be.false;
-		});
-
-		it('inherits parent static properties', () => {
-			const Klass2 = Klass.extend({});
-
-			expect(Klass2.bla).to.eql(1);
-		});
-
-		it('overrides parent static properties', () => {
-			const Klass2 = Klass.extend({statics: {bla: 2}});
-
-			expect(Klass2.bla).to.eql(2);
-		});
-
-		it('includes the given mixin', () => {
-			const a = new Klass();
-			expect(a.mixin).to.be.true;
-		});
-
-		it('does not merge \'includes\' property itself', () => {
-			expect('includes' in Klass.prototype).to.be.false;
-		});
-
-		it('includes multiple mixins', () => {
-			const Klass2 = Class.extend({
-				includes: [{mixin: true}, {mixin2: true}]
-			});
-			const a = new Klass2();
-
-			expect(a.mixin).to.be.true;
-			expect(a.mixin2).to.be.true;
-		});
-
-		it('includes Evented mixins', () => {
-			const Klass2 = Class.extend({
-				includes: Evented.prototype
-			});
-			const a = new Klass2();
-
-			expect(a.on).to.eql(Evented.prototype.on);
-			expect(a.off).to.eql(Evented.prototype.off);
-		});
-
-		it('includes inherited mixins', () => {
-			class A {
-				foo() {}
+		class TestClass extends Class {
+			initialize(...args) {
+				spy(...args);
 			}
-			class B extends A {
-				bar() {}
+		}
+
+		new TestClass(1, 2, 3);
+		console.warn.restore();
+
+		expect(spy.calledWith(1, 2, 3)).to.be.true;
+		expect(warnSpy.calledWith('The \'initialize()\' method is deprecated, use a class constructor instead.')).to.be.true;
+	});
+});
+
+describe('Class#extend', () => {
+	let Klass,
+	props,
+	constructor,
+	method;
+
+	beforeEach(() => {
+		constructor = sinon.spy();
+		method = sinon.spy();
+
+		props = {
+			statics: {bla: 1},
+			includes: {mixin: true},
+
+			initialize: constructor,
+			foo: 5,
+			bar: method
+		};
+		Klass = Class.extend(props);
+	});
+
+	it('creates a class with the given constructor & properties', () => {
+		const a = new Klass();
+
+		expect(constructor.called).to.be.true;
+		expect(a.foo).to.eql(5);
+
+		a.bar();
+
+		expect(method.called).to.be.true;
+	});
+
+	it('calls the correct parent initialize function', () => {
+		const initialize = sinon.spy();
+		const initializeDraw = sinon.spy();
+		const initializeEdit = sinon.spy();
+		const Toolbar = Class.extend({
+			initialize() {
+				initialize();
 			}
-			const Klass2 = Class.extend({
-				includes: B.prototype
-			});
-			const a = new Klass2();
-
-			expect(a.bar).to.eql(B.prototype.bar);
-			expect(a.foo).to.eql(B.prototype.foo);
+		});
+		Toolbar.include(Evented.prototype);
+		const DrawToolbar = Toolbar.extend({
+			initialize(options) {
+				initializeDraw();
+				Toolbar.prototype.initialize.call(this, options);
+			}
+		});
+		const EditToolbar = Toolbar.extend({
+			initialize(options) {
+				initializeEdit();
+				Toolbar.prototype.initialize.call(this, options);
+			}
 		});
 
-		it('grants the ability to include the given mixin', () => {
-			Klass.include({mixin2: true});
+		new Toolbar();
+		sinon.assert.calledOnce(initialize);
+		sinon.assert.notCalled(initializeDraw);
+		sinon.assert.notCalled(initializeEdit);
+		initialize.resetHistory();
 
-			const a = new Klass();
-			expect(a.mixin2).to.be.true;
+		new DrawToolbar();
+		sinon.assert.callOrder(initializeDraw, initialize);
+		sinon.assert.notCalled(initializeEdit);
+		initialize.resetHistory();
+		initializeDraw.resetHistory();
+
+		new EditToolbar();
+		sinon.assert.callOrder(initializeEdit, initialize);
+		sinon.assert.notCalled(initializeDraw);
+	});
+
+	it('inherits parent classes\' constructor & properties', () => {
+		const Klass2 = Klass.extend({baz: 2});
+
+		const b = new Klass2();
+
+		expect(b instanceof Klass).to.be.true;
+		expect(b instanceof Klass2).to.be.true;
+
+		expect(constructor.called).to.be.true;
+		expect(b.baz).to.eql(2);
+
+		b.bar();
+
+		expect(method.called).to.be.true;
+	});
+
+	it('does not modify source props object', () => {
+		expect(props).to.eql({
+			statics: {bla: 1},
+			includes: {mixin: true},
+
+			initialize: constructor,
+			foo: 5,
+			bar: method
+		});
+	});
+
+	it('supports static properties', () => {
+		expect(Klass.bla).to.eql(1);
+	});
+
+	it('does not merge \'statics\' property itself', () => {
+		expect('statics' in Klass.prototype).to.be.false;
+	});
+
+	it('inherits parent static properties', () => {
+		const Klass2 = Klass.extend({});
+
+		expect(Klass2.bla).to.eql(1);
+	});
+
+	it('overrides parent static properties', () => {
+		const Klass2 = Klass.extend({statics: {bla: 2}});
+
+		expect(Klass2.bla).to.eql(2);
+	});
+
+	it('includes the given mixin', () => {
+		const a = new Klass();
+		expect(a.mixin).to.be.true;
+	});
+
+	it('does not merge \'includes\' property itself', () => {
+		expect('includes' in Klass.prototype).to.be.false;
+	});
+
+	it('includes multiple mixins', () => {
+		const Klass2 = Class.extend({
+			includes: [{mixin: true}, {mixin2: true}]
+		});
+		const a = new Klass2();
+
+		expect(a.mixin).to.be.true;
+		expect(a.mixin2).to.be.true;
+	});
+
+	it('includes Evented mixins', () => {
+		const Klass2 = Class.extend({
+			includes: Evented.prototype
+		});
+		const a = new Klass2();
+
+		expect(a.on).to.eql(Evented.prototype.on);
+		expect(a.off).to.eql(Evented.prototype.off);
+	});
+
+	it('includes inherited mixins', () => {
+		class A {
+			foo() {}
+		}
+		class B extends A {
+			bar() {}
+		}
+		const Klass2 = Class.extend({
+			includes: B.prototype
+		});
+		const a = new Klass2();
+
+		expect(a.bar).to.eql(B.prototype.bar);
+		expect(a.foo).to.eql(B.prototype.foo);
+	});
+
+	it('grants the ability to include the given mixin', () => {
+		Klass.include({mixin2: true});
+
+		const a = new Klass();
+		expect(a.mixin2).to.be.true;
+	});
+
+	it('merges options instead of replacing them', () => {
+		const KlassWithOptions1 = Class.extend({
+			options: {
+				foo1: 1,
+				foo2: 2
+			}
+		});
+		const KlassWithOptions2 = KlassWithOptions1.extend({
+			options: {
+				foo2: 3,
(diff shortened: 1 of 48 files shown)


jonkoops:
I've marked this PR as a draft, as it is intended to open up discussion about the implementation first and foremost. @simon04 I would love your input as well, considering your involvement in recent modernization of the codebase in relation to classes and standardized JavaScript features.

spec/suites/core/ClassSpec.js
jonkoops, line 1: I will be adding a bunch of tests for various scenarios later on, once we have settled on an API.

Falke-Design:
Would be `initialize` still called? And is the behavior still the same only calling initialize on super class would not work?

```js
const MyMarker = Marker.extend({
  initialize(latlng, options) {
   // do something
  }
});
```

Just to confirm: extending a class which has initHooks works but initHooks are only called for the super class (Bar) and not for the extended class (Foo)? Edit: just saw that Bar is no class anymore. Extending classes from plugins will be become very hard
```js
const Bar = withInitHooks(class Bar {});

const Foo extends Bar {}
```

jonkoops:
> Would be `initialize` still called? And is the behavior still the same only calling initialize on super class would not work?

Yes, `initialize()` would still be called, I intentionally left the [constructor behavior](https://github.com/jonkoops/Leaflet/blob/e20b76bd049a507b21c94753c75e4c9f68f254ba/src/core/Class.js#L91-L99) of `Class` in-place to preserve backwards compatibility. As long as the `initialize()` method of the class does not make any calls to a super class' `initialize()` method, things should keep working. We might consider logging a deprecation message when this happens to inform the user to migrate over.

> Just to confirm: extending a class which has initHooks works but initHooks are only called for the super class (Bar) and not for the extended class (Foo)?

Initialization hooks can be registered on any class that is wrapped with `withInitHooks()`, or is created by calling `Class#extend()`. For example:

```js
const A = withInitHooks(class A {});

A.addInitHook(function () {
  console.log('Init hook A called');
});

const B = withInitHooks(class B extends A {});

B.addInitHook(function () {
  console.log('Init hook B called');
});

// 'Init hook A called'
// 'Init hook B called'
new B();
```

This calls all the hooks in the chain, because all classes are wrapped in `withInitHooks()`. However, even if `B` was not wrapped, it would still execute all the hooks registered on its super classes:

```js
const A = withInitHooks(class A {});

A.addInitHook(function () {
  console.log('Init hook A called');
});

class B extends A {}

// 'Init hook A called'
new B();
```

> just saw that Bar is no class anymore. Extending classes from plugins will be become very hard

I think you are referring to the fact that `Bar` does not extend `Class` in my example? That is intentional, it is to show that classes no longer need to extend from `Class`, but they still can. It is an example to show that the `withInitHooks()` functionality can be de-coupled from `Class` (if so desired).

As for backwards compatibility, the `Class#extends()` automatically applies `withInitHooks()`, so as long as that is used by v1.x plugins over the new `extends` keyword, they should remain functional (a lot of old plugins should still work this way). I think the lack of `initialize()` is the bigger backwards compatibility issue.

Falke-Design:
> We might consider logging a deprecation message when this happens to inform the user to migrate over.

Yes lets do this.

Thanks for the explaination. I was confused with extending a class into `const` (`const B = withInitHooks(class B extends A {});`). 

If we replace `initialize` with `constructor` this PR is fine for me. The workaround `withInitHooks` is not really pretty but if there is no other way then we need to do it like this

simon04:
How to upgrade the following code to the new technology proposed in this pull request? https://github.com/Leaflet/Leaflet.VectorGrid/blob/2ac9a8b72285e28465384d0c303acdc08e05f95e/src/Leaflet.Renderer.SVG.Tile.js#L3-L16

First attempt does not work since `BlanketOverlay` calls `super()` (without arguments) and thus `initialize: function (tileCoord, tileSize, options)` is called without arguments, i.e., `(undefined, undefined, undefined)`

```diff
diff --git a/dist/Leaflet.VectorGrid.bundled.js b/dist/Leaflet.VectorGrid.bundled.js
index f587bbcf147e55e3fbafebc9af46363a3aad710c..1fc46f882a1fff2e47b7c4f5840966fd60db9778 100644
--- a/dist/Leaflet.VectorGrid.bundled.js
+++ b/dist/Leaflet.VectorGrid.bundled.js
@@ -1637,7 +1637,7 @@ var VectorTile = vectortile;
 L.SVG.Tile = L.SVG.extend({
 
 	initialize: function (tileCoord, tileSize, options) {
-		L.SVG.prototype.initialize.call(this, options);
+		L.Util.setOptions(this, options);
 		this._tileCoord = tileCoord;
 		this._size = tileSize;
```

Second attempt does not work either:

```diff
diff --git a/dist/Leaflet.VectorGrid.bundled.js b/dist/Leaflet.VectorGrid.bundled.js
index f587bbcf147e55e3fbafebc9af46363a3aad710c..4004a67da82c7c5cc0ac7858e419398122811899 100644
--- a/dist/Leaflet.VectorGrid.bundled.js
+++ b/dist/Leaflet.VectorGrid.bundled.js
@@ -1636,8 +1636,8 @@ var VectorTile = vectortile;
 
 L.SVG.Tile = L.SVG.extend({
 
-	initialize: function (tileCoord, tileSize, options) {
-		L.SVG.prototype.initialize.call(this, options);
+	constructor: function (tileCoord, tileSize, options) {
+		super(options);
 		this._tileCoord = tileCoord;
 		this._size = tileSize;
```

```
[PARSE_ERROR] Error: Super calls are not permitted outside constructors or in nested functions inside constructors.
      ╭─[ node_modules/leaflet.vectorgrid/dist/Leaflet.VectorGrid.bundled.js:1640:3 ]
      │
 1640 │        super(options);
      │        ───────┬──────  
      │               ╰──────── 
──────╯
```

jonkoops:
@simon04 you'll have to call `super.constructor()` instead of `super()` for 'legacy' classes (using `Class#extend()`), see the example in the PR description.

jonkoops:
@Falke-Design do we have some docs for users and plugin authors to migrate to Leaflet v2? I am thinking of setting one up in this PR if there isn't one already, so we can contribute to it as we collect changes along the way.

Falke-Design:
No not really. We have the blog post and the Polyfill Repo

simon04:
> you'll have to call `super.constructor()` instead of `super()` for 'legacy' classes (using `Class#extend()`)

Thank you for the pointer. It seems that I'm still missing (or misunderstanding) something:

```diff
diff --git a/spec/suites/core/ClassSpec.js b/spec/suites/core/ClassSpec.js
index 0ec6d7c2..f3c82013 100644
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -1,5 +1,5 @@
 import {expect} from 'chai';
-import {Class, Evented} from 'leaflet';
+import {Class, Evented, GridLayer} from 'leaflet';^M
 import sinon from 'sinon';
 
 describe('Class', () => {
@@ -281,6 +281,18 @@ describe('Class', () => {
 
                        expect(spy1.called).to.be.true;
                });
+
+               it('calls constructor', () => {
+                       const VectorGridProtobuf = GridLayer.extend({
+                               constructor(url, options) {
+                                       super.constructor(options);
+                                       this.url = url;
+                               }
+                       });
+                       const l = new VectorGridProtobuf('https://example.com/', {maxZoom: 42});
+                       expect(l.url).to.eql('https://example.com/');
+                       expect(l.options.maxZoom).to.eql(42);
+               });
        });
 
        describe('#extend', () => {
```

```
Chrome Headless 140.0.0.0 (Mac OS 10.15.7) Class #extend calls constructor FAILED
        AssertionError: expected undefined to deeply equal 'https://example.com/'
            at Context.<anonymous> (spec/suites/core/ClassSpec.js:293:21)
Chrome Headless 140.0.0.0 (Mac OS 10.15.7): Executed 1086 of 1100 (1 FAILED) (skipped 14) (6.093 secs / 5.836 secs)
TOTAL: 1 FAILED, 1085 SUCCESS
```

spec/suites/core/ClassSpec.js
jonkoops: Added a bunch of tests to verify the behavior of hooks, plus some general restructuring.

jonkoops:
@simon04 looks like you found a bug, the constructors on 'legacy' classes don't seem to run properly. I'll do some research as to why, and provide a fix and some regression tests.

jonkoops:
It seems some limitations might apply in how a `constructor()` is defined when using the `class` syntax, no matter what I do, it will not call the new constructor when it is re-defined by assigning `prototype.constructor`. I'll have to do some deeper digging.

See:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/constructor#calling_super_in_a_constructor_bound_to_a_different_prototype
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor

jonkoops:
I have been doing some research into this, and I don't think we can make `initialize()` backwards compatible. Take the following example code that we'd hypothetically want to refactor that we might find in the wild:

```js
// Minimal implementation of 'Class' as we have it in Leaflet now.
class Class {
	static extend(props) {
		const ChildClass = class extends this {};

		Object.assign(ChildClass.prototype, props);

		return ChildClass;
	}

	constructor(...args) {
		if (typeof this.initialize === 'function') {
			this.initialize(...args);
		}
	}
}

// Hypothetical class exported from our library.
const Person = Class.extend({
	initialize(name) {
		this.name = name;
	},

	greet() {
		console.log(`${this.name} says hello!`);
	}
});

// Implementation by a user, extending a class imported from our library.
const SuperHero = Person.extend({
	initialize(name, superPower) {
		Person.prototype.initialize.call(this, name);
		this.superPower = superPower;
	},

	useSuperPower() {
		console.log(`${this.name} uses ${this.superPower}!`);
	}
});

const superman = new SuperHero('Superman', 'laser eyes');
superman.greet(); // "Superman says hello!"
superman.useSuperPower(); // "Superman uses laser eyes!"
```

In this scenario, the `Person` class is a hypothetical API provided by the library that is consumed by the users' code, and the `SuperHero` class is a piece of implementation from users' code, or a plugin, etc. We now refactor the `Person` class to the new `class` syntax, leaving the `SuperHero` implementation intact (this will simulate a user upgrading the library):

```js
class Person extends Class {
	constructor(name) {
		super();
		this.name = name;
	}

	greet() {
		console.log(`${this.name} says hello!`);
	}
}
```

This will break `SuperHero`, when its `initialize()` method is called it attempts to call `Person.prototype.initialize()`, which no longer exists. Ignorantly, I thought calling `Person.prototype.constructor()` directly would be an option:

```js
const SuperHero = Person.extend({
	initialize(name, superPower) {
		// TypeError: class constructors must be invoked with 'new'"
		Person.prototype.constructor.call(this, name);
	},
});
```

However, as you can see, this will throw an error, as class constructors [must be called with `new`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Class_ctor_no_new). We could choose to construct the instance directly using the `new` keyword, or use `Reflect.construct()`:

```js
const SuperHero = Person.extend({
	initialize(name, superPower) {
		const instance = new Person(name);
		// Alternatively:
		const instance = Reflect.construct(Person, [name]);
	},
});
```

Now we have a constructed instance, but since it is not bound to the `this` of `SuperHero`, it is just a regular `Person` instance. As far as I can tell, there is no way to actually call a constructor of a parent class using the `this` value of a child.

Another issue is that due to the fact that the `constructor()` of `Class` is called last, the constructor of `Person` is already executing, the execution order is now:

- :arrow_right: `new SuperHero('Superman', 'laser eyes')`
- :arrow_right: `Person` constructor
- :arrow_right: `super()` call to `Class` constructor
- :arrow_right: Call to `initialize()` method of `SuperHero`
- :arrow_left: back to `Person` constructor (sets `name`)

Not only does the parent constructor (`Person`) get called before the `initialize()` method of `SuperHero`, the `name` property is also not available, as `super()` [must be called](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Super_not_called) before accessing `this`. This also means we cannot define the constructor on the `NewClass` in `Class.extend()`, as we'd still have to call super, causing all the constructors to fire before we call `initialize()`.

What we want, is to intercept the constructor of any class definition using `Class.extend()`, and make sure we call the `initialize()` before any other parent constructor, as `initialize()` will be responsible for calling the constructor, whist preserving the prototype hierarchy.

Logically, one would think that it would be possible to accomplish this using a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) and [`Reflect`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect), but we are still limited by the fact that even when intercepted, we cannot call a constructor with a custom `this` value.

Considering all this, unless someone has a suggestion on how to get around these limitations, I would propose we drop `initialize()` entirely, and replace it with a regular constructor. Since this prevents `Class.extend()` from actually being able to call super classes, I think it is time we drop it entirely.

@Leaflet/core @Leaflet/leaflet-maintainers WDYT?

Falke-Design:
Thank you for the explanation. 

I still don’t see the benefit of forcing the use of a constructor. We’re breaking so many things for very little gain. We tried implementing `constructor`, but it’s not possible to maintain the same behavior, so let’s keep `initialize`. We’ve already modernized a lot - there’s no need to force it everywhere just to say that we’re using 100% of the JS class implementation, especially when many things no longer work for users and migrating will be a huge pain. Many plugins will never work again because they won’t be updated and no polyfill is possible. In the end, I think many users will choose not to update and stay with version 1.9

Falke-Design:
@jonkoops, we need a solution. I’ve invested a lot of time getting v2 ready, and it’s frustrating that we haven’t moved forward in the past six weeks because we’re stuck on this.

jonkoops:
Yes, I agree. I am working on updating the documentation to match what is currently in `main`. We'll need that before we can get to release candidates, no matter what we decide to do here.

As for the work under this PR, I am still waiting for more @Leaflet/core members to comment on how we want to proceed with it. Specifically, I would like to get @mourner and @IvanSanchez's takes.

In my opinion, I think it is still best if we continue by dropping `initialize()` entirely, and adopting standardized constructors. It might be a breaking change, but that is what major versions are for, and it will set the API for V2 in stone, which is likely a version we'll be supporting a long time. I don't want to make any big breaking changes to the core class system in any major version that follows, so I prefer to make them now and move on.

mourner:
> It might be a breaking change, but that is what major versions are for, and it will set the API for V2 in stone, which is likely a version we'll be supporting a long time. I don't want to make any big breaking changes to the core class system in any major version that follows, so I prefer to make them now and move on.

@jonkoops I generally agree with this — it's better to break early than to allow legacy stuff pile on. This is the same reason I was hoping we'd be able to drop init hooks (which we go to great length to enable in a compatible way), but I guess that would be a bigger hassle for plugins to get compatible. 

But it's not a strong opinion — I think it's more valuable to make a decision quickly (whichever way it goes) than deliberate forever on a perfect solution that might not come.

jonkoops:
I'm going to go ahead and do the work to move to using `constructor()` only, and keep initialization hooks intact. We'll have to have a separate discussion on features like initialization hooks, and `Class.include()`.

jonkoops:
Closing this, and raising a separate PR with the new implementation.

PR #9910, opened 2025-09-24
Reduce exported API surface from package

jonkoops:
Reduces the exported API surface of the package to only include the main module and stylesheet. This is a follow-up of #9658, which increased the exported surface to include everything under the `src` and `dist` directories. These directories are implementation details of the package, and should not be leaked as part of the public API surface.

This PR allows bundlers such as Vite that support package exports to include Leaflet as follows:

```js
import 'leaflet/styles.css';
```

The original PR provided the following motivation (aside from not being able to load the stylesheet):

> In addition, I (used to) import source files (such as `import { Map } from "leaflet/src/map/Map";`) because this strategy allows for smaller bundle size. Thus, please also cover the `src` files in `exports` in `package.json`.

This is not a valid strategy of optimizing bundle size. Aside from leaking the entire API surface, including functionality we might want to consider private, or not yet read for public consumption, this does not actually result in a smaller bundle size if any other imports are present. For example:

```js
// Some plugin might import this.

// And application code this.
import { Map } from "leaflet/src/map/Map";
```

In the above scenario, both `leaflet/src/map/Map.js` and `leaflet/dist/leaflet-src.js` are now included, leading to code duplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,9 +34,8 @@
   },
   "type": "module",
   "exports": {
-    "./src/": "./src/",
-    "./dist/": "./dist/",
-    ".": "./dist/leaflet-src.js"
+    ".": "./dist/leaflet-src.js",
+    "./styles.css": "./dist/leaflet.css"
   },
   "style": "dist/leaflet.css",
   "files": [


jonkoops:
@simon04 FYI

simon04:
Due to #9814, I'm selectively importing certain Leaflet classes to minimise the bundle size, for instance:

```js
import "leaflet/src/control/Control.Attribution";
import { Layers } from "leaflet/src/control/Control.Layers";
import { Scale } from "leaflet/src/control/Control.Scale";
import "leaflet/src/control/Control.Zoom";
import { LayerGroup } from "leaflet/src/layer/LayerGroup";
import { Tooltip } from "leaflet/src/layer/Tooltip";
import { CircleMarker } from "leaflet/src/layer/vector/CircleMarker";
import "leaflet/src/layer/vector/Renderer.getRenderer";
import { Map } from "leaflet/src/map/Map";
import "leaflet/src/map/handler/Map.BoxZoom";
import "leaflet/src/map/handler/Map.DoubleClickZoom";
import "leaflet/src/map/handler/Map.Drag";
import "leaflet/src/map/handler/Map.Keyboard";
import "leaflet/src/map/handler/Map.PinchZoom";
import "leaflet/src/map/handler/Map.ScrollWheelZoom";
import "leaflet/src/map/handler/Map.TapHold";
```

Without `"./src/": "./src/"` this strategy will no longer be possible.

jonkoops:
I understand that it is not possible to selectively import certain parts of Leaflet like this, but this was never supposed to be a feature and only worked incidentally because v1 never defined an export surface. There are problems with doing it in this manner, such as code duplication (see PR description).

If we are going to support proper dead code elimination, that should happen through #9814 and be done in a way we can support going forward, without exposing implementation details of the package that might be prone to changes.

Falke-Design:
I have no knowledge about that, so I can't say much. 

I do not understand why is a big issue, because at the end it is the same object. If someone uses the direct import then he accepts to prevent Tree Shaking

```
// Some plugin might import this.

// And application code this.
import { Map } from "leaflet/src/map/Map";
```

jonkoops:
It is not the same, as the `leaflet` import specifier links to `./dist/leaflet-src.js`, which is an entirely different piece of code than `./src/map/Map.js`. Since they are two different entry-points, the bundler will bundle both.

For example:
```js
import { Map as MapA } from "leaflet";
import MapB from "leaflet/src/map/Map";

MapA === MapB; // false
```

package.json
```diff
@@ -34,9 +34,8 @@
   },
   "type": "module",
   "exports": {
-    "./src/": "./src/",
-    "./dist/": "./dist/",
```
Falke-Design, line 38: why dist not needed anymore?

jonkoops: We don't want to export anything that isn't part of the public API surface (see my comments above). Note, this does not impact folks that just grab the files from the NPM package. This only affects tools that implement the import specifier, such as bundlers (Vite, WebPack), or Node.js.

Falke-Design: But would it not make sense for /images too?

jonkoops: It might, let's see if this works correctly. If we need to expose certain entry points, we can do so selectively (e.g. only expose the images).

PR #9918, opened 2025-09-29
Plugin: Update leafletrastercoords.md

commenthol:
confirming leaflet-rastercoords compatibility with leaflet v2


jonkoops:
Hi @commenthol, we're holding off on merging further PRs with 2.0 support for plugins, as we are likely making some more API changes while in alpha. See https://github.com/Leaflet/Leaflet/issues/9842#issuecomment-3302673291. Once we're in a beta/release candidate stage, I think we can commit to landing these again.

PR #9919, opened 2025-09-30
Register mixins for code splitting

### LINKED ISSUE #9814: Treeshaking not working

simon04:
Attempt to resolve #9814

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "type": "module",
   "exports": {
     ".": "./dist/leaflet-src.js",
+    "./src/": "./src/",
     "./styles.css": "./dist/leaflet.css"
   },
   "style": "dist/leaflet.css",
--- a/src/LeafletWithGlobals.js
+++ b/src/LeafletWithGlobals.js
@@ -1,6 +1,22 @@
 import * as L from './Leaflet.js';
 export * from './Leaflet.js';
 
+// register mixins
+L.AttributionControl.register();
+L.Control.register();
+L.DivOverlay.register();
+L.Layer.register();
+L.MapBoxZoom.register();
+L.MapDoubleClickZoom.register();
+L.MapDrag.register();
+L.MapKeyboard.register();
+L.MapPinchZoom.register();
+L.MapScrollWheelZoom.register();
+L.MapTapHold.register();
+L.Popup.register();
+L.Tooltip.register();
+L.ZoomControl.register();
+
 export default L;
 
 const oldL = getGlobalObject().L;
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -123,18 +123,20 @@ export class Attribution extends Control {
 
 		this._container.innerHTML = prefixAndAttribs.join(' <span aria-hidden="true">|</span> ');
 	}
-}
 
-// @namespace Map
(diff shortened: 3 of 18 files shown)


Maxel01:
Hey, I have seen your progress.
I tested a single component (`Popup`) with treeshaking.
First of all, there is a need to use the `Leaflet.js` as the entry file `LeafletWithGlobals `registers the modules directly which does not really make sense. see #9880

The problem with `Popup` is still the static part with `this.setDefaultOptions` which is a side effect. I found an experimental option in rollup: `experimentalLogSideEffects: true`.
This logs the first sideEffect of every file and could help to identify the problems.
I proposed to use mixins or some kind of plugin system in #9814, in my opinion it could be a little bit overkill to add mixins to all the control and map handlers as they are used anyway (`Control.Layers` is a good example for the usage of mixins as it is a rather big file and is not always used). In `control/index.js` these components are used anyway:
```js
Control.Layers = Layers;
Control.Zoom = Zoom;
Control.Scale = Scale;
Control.Attribution = Attribution;
```
but the components are registered in `LeafletWithGlobals` which increases the complexity without any benefit (at the moment as they are not treeshakeable yet). Primarily, this means that the shortcut `Control.*` shouldn't be available anymore (or add it in the register function if possible).
```js
static register() {
    Control.Layers = Layers
    ....
}
```
Let me know if I can assist you in any way.

PR #9924, opened 2025-10-03
Update leaflet-WorldMiniMap.md

maneoverland:
Updated for leaflet-V2.


maneoverland:
Dont know, if the pull-request was registered by the maintainers of leaflet.

jonkoops:
Hi @maneoverland, we're holding off on merging further PRs with 2.0 support for plugins, as we are likely making some more API changes while in alpha. See https://github.com/Leaflet/Leaflet/issues/9842#issuecomment-3302673291. Once we're in a beta/release candidate stage, I think we can commit to landing these again.

PR #9928, opened 2025-10-04
Fix wrong identifiers in docs

### LINKED ISSUE #9927: Errors in leaflet-src.js doc
### Checklist

- [x] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this change hasn't been suggested already.

### Which documents need updating?

leaflet-src.js

### Motivation

I think I have found two errors in the inline documentation in leaflet-src.js:
pointerEventToLatLng where the @method value is wrong
VideoOverlay.controls property where the @option value is wrong

### Suggested changes

Change the values for above mentioned tags so they reflect the correct method/property name.

simon04:
Closes #9927.

### DIFF
--- a/src/layer/VideoOverlay.js
+++ b/src/layer/VideoOverlay.js
@@ -35,7 +35,7 @@ export class VideoOverlay extends ImageOverlay {
 			// On some browsers autoplay will only work with `muted: true`
 			autoplay: true,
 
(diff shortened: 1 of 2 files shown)


jonkoops (approved):

jonkoops:
LGTM, thanks!

PR #9952, opened 2025-10-22
Remove user-agent detection for `Browser.mobile`

jonkoops:
This removes the user-agent detection code for `Browser.mobile`, which was originally added (ea680ce21af1e36b136c6108c0d51962e220b9a5) to support older versions of Firefox mobile. This is no longer needed, as Firefox mobile now supports the `orientation` property.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -19,7 +19,7 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof orientation !== 'undefined' || userAgentContains('mobile');
+const mobile = typeof orientation !== 'undefined';
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).


IvanSanchez:
@jonkoops How concerned are we with breaking plugins?

Falke-Design:
Reference: #9710

jonkoops:
@IvanSanchez this should not break anything, see the commit I mentioned and the reference @Falke-Design linked (https://github.com/Leaflet/Leaflet/issues/9710#issuecomment-3432324028).

IvanSanchez:
@jonkoops @Falke-Design Since I see that `Browser.mobile` is being used at... https://github.com/Leaflet/Leaflet/blob/aa7dea21dfdae7248acced89e9675f93306ff604/src/layer/tile/GridLayer.js#L95

...to degrade functionality on a low-performance scenario, I think it makes more sense to *add* a new item to `L.Browser`, based on the WebGL performance attribute.

The idea is copy-pasting the code from https://www.khronos.org/webgl/public-mailing-list/public_webgl/1703/msg00022.php ...

```js
     canvas.addEventListener('webglcontextlost`, function() {});
     canvas.addEventListener('webglcontextrestored`, function() {});
     const gl = canvas.getContext("webgl", { powerPreference: "high-performance" });
     const havePerf = gl.getContextAttributes().powerPreference === "high-performance";
     if (!havePerf) {
         alert("Dear user, this site requires a gamer ready GPU. Sucks to be you");
    }
```

...and store that into `L.Browser.highPerformanceGPU` or something like that.

These days we don't really care about Leaflet being on a cellphone or tablet anymore - we care about (a) whether the GPU can waste cycles on animations and (b) whether the computer has a metered internet connection to load more or less data (`navigator.network.saveData`).

Perhaps we can assume that Leaflet runs on "mobile" if it has low WebGL performance **or** is using a metered connection?

jonkoops:
@IvanSanchez yes, we can likely remove this in the future. For now, I just want to get rid of redundant code. I'd honestly just default `updateWhenIdle` to `true`, I don't think there are any modern phones that would have issues.

PR #9954, opened 2025-10-23
Update leaflet-polydraw.md, Enable Leaflet 2.x compatibility

AndreasOlausson:
Adds Leaflet 2.x compatibility flag for plugin testing.

- Enables v2 path handling
- Keeps Leaflet 1.x behavior unchanged
- Verified with both 1.9.4 and 2.0.0-alpha.1 using switch-version.sh

### DIFF
--- a/docs/_plugins/edit-geometries/leaflet-polydraw.md
+++ b/docs/_plugins/edit-geometries/leaflet-polydraw.md
@@ -7,7 +7,7 @@ author-url: https://github.com/AndreasOlausson
 demo: https://polydraw.ao-tech.se
 compatible-v0:
 compatible-v1: true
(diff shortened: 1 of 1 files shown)


jonkoops:
Hi @AndreasOlausson, we're holding off on merging further PRs with 2.0 support for plugins, as we are likely making some more API changes while in alpha. See https://github.com/Leaflet/Leaflet/issues/9842#issuecomment-3302673291. Once we're in a beta/release candidate stage, I think we can commit to landing these again.

AndreasOlausson:
No problem, that makes perfect sense. I’ll keep the branch updated and reopen once Leaflet 2.0 moves into beta. Thanks for the clarification!

PR #9962, opened 2025-10-25
Fix: type selector "svg" cannot be directly after nesting selector "&"

simon04:
```
▲ [WARNING] Cannot use type selector "svg" directly after nesting selector "&" [css-syntax-error]

CSS nesting syntax does not allow the "&" selector to come before a type selector. You can wrap this selector in ":is(...)" as a workaround. This restriction exists to avoid problems with SASS nesting, where the same syntax means something very different that has no equivalent in real CSS (appending a suffix to the parent selector).
```

### DIFF
--- a/src/leaflet.css
+++ b/src/leaflet.css
@@ -208,7 +208,7 @@
 
 .leaflet-zoom-animated {
 	transform-origin: 0 0;
-	&svg {
+	svg& {
 		will-change: transform;
 	}
 }


jonkoops:
@simon04 are these warnings produced by the linter? Can we fail it if these kinds of issues are found?

simon04:
The error has been reported by angular-cli while building https://gitlab.com/albina-euregio/albina-admin-gui against leaflet f8fef451. Due to #9910, a patching leaflet was required, see https://gitlab.com/albina-euregio/albina-admin-gui/-/commit/55af6126c093030674347a362b63a41ea4f15a2a.

jonkoops (approved):

PR #9973, opened 2025-11-08
Refactor Draggable class for improved clarity

rvambadi69:


rvambadi69:
CHANGE SUMMARY:
Previously, disableTextSelection() and enableTextSelection() affected the entire document,
which unintentionally blocked text selection in all parts of the page when interacting with the map.

This update adds an optional `el` parameter (HTMLElement) to both functions,
 allowing text selection to be disabled and re-enabled only for a specific target element
 such as the map container.

 If no element is provided, the behavior defaults to the previous global document-wide behavior
to maintain backward compatibility.

 This change fixes issue #9259 and ensures text selection is only disabled inside the map area,
without interfering with input fields or other interactive UI elements on the page.

simon04 (changes requested):
Useless PR. `DomUtil.disableTextSelection` does not accept an argument.

jonkoops:
Same garbage as #9979. Please stop wasting our time.

PR #9982, opened 2025-11-11
Reformat JavaScript code using prettier

simon04:
https://prettier.io/

```sh
$ npx prettier --write --ignore-unknown (git ls-files '**/*.js')
```

We can add a pre-commit hook if this code formatting style is appreciated, see https://prettier.io/docs/precommit


jonkoops:
To be honest, I don't think we should make large, sweeping code-style changes for the sake of it. It makes the Git history harder to parse, and existing patches harder to apply. We already have a linter set up with rules that conflict with Prettier's defaults.

If we are going to move to a different system in the future, I'd like to move to [Vite Plus](https://viteplus.dev/), or [Deno](https://docs.deno.com/runtime/), which are toolchains that come with bundling, API documentation, formatter, linter, type-checker, etc. included. This removes the need to juggle various packages and integrate them together.

simon04:
> It makes the Git history harder to parse

This is easily solved using https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revrev

> and existing patches harder to apply

I think the ongoing refactoring for 2.0.0 already did.

> I'd like to move to [Vite Plus](https://viteplus.dev/)

From https://viteplus.dev/#feature-format, to https://oxc.rs/, to https://oxc.rs/docs/guide/usage/formatter.html -- the result is basically the same (see https://github.com/simon04/Leaflet/commit/3eaf10f01f23e41c8b80dcbc11bf10525e7645e0).

Falke-Design:
@simon04 we would need a solution for users which are not using the cli. Maybe the `--ignore-revs-file <file> ` can be setup via default config?

>  > and existing patches harder to apply
>
> I think the ongoing refactoring for 2.0.0 already did.

I think that too. 

We already had issues with line endings which would be fixed by that too. https://github.com/Leaflet/Leaflet/discussions/7729#discussioncomment-1577098

simon04:
> Maybe the --ignore-revs-file <file> can be setup via default config?

Yes: Using the `blame.ignoreRevsFile` config option: https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-filefile

The leaflet repository would include a file (sometimes named `.git-blame-ignore-revs`), and a `git config blame.ignoreRevsFile .git-blame-ignore-revs` would configure git to exclude the listed revisions for git-blame.

PR #9987, opened 2025-11-14
Posted three plugsin

almamigratoria-netizen:
Three plugins for Leaflet v2.  Tested on 2.0.0-alpha1.
- Leaflet-SVGMarkers
- Leaflet-OpenMeteo
- Leaflet-LayersTree


Falke-Design:
We currently paused merging V2 plugins because we are redesigning some APIs in V2. After the API is stable we will continue merging after a check from the plugin maintainers

jonkoops:
We're currently not accepting PRs advertising 2.x support as the API is still in flux. Once we reach the beta/RC state, this will be allowed again. Sorry for the inconvenience.

almamigratoria-netizen:
No, that's all right.  Thanks for getting back to me.  It's appreciated, and I don't mind the delay.  You guys are making my plugins possible, so...

I tried to use 100% new Leaflet idioms, but between starting the first one and finishing the last one concensus on best practice had morphed a bit... These are all still good to go with current Leaflet 2 practices, but I'll keep and eye on when  you say things are stabilized.  By that time I'm likely to have one or two more.

Issue #9990, opened 2025-11-22
Leaflet does not work on Safari 16.5 (July 2023)


simon04:
`URL.canParse` was introduced in https://github.com/Leaflet/Leaflet/pull/9489

My comment in https://github.com/Leaflet/Leaflet/pull/9489#issuecomment-3566165192

> `URL.canParse` is not supported in Safari Version 16.5.1 (18615.2.9.11.7) [released in July 2023](https://support.apple.com/en-us/121012).
> 
> MDN https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static categorizes this feature as "Baseline 2023 Newly available. Since ⁨December 2023⁩, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers."
> 
> Can we find a solution not breaking two year old browsers?

I think leaflet should restrict itself to "Baseline Widely available" features.

Do we want to incorporate the following eslint plugin, https://github.com/3ru/eslint-plugin-baseline-js?

mourner:
Yeah, Safari 17 is too high of a requirement for a core Leaflet feature — should definitely be fixed (`canParse` isn't even necessary in that PR), and it's good idea to add a ESLint plugin for catching those things. Want to PR this?

PR #9668, opened 2025-04-21
Enable `@eslint/css` linting

simon04:
https://eslint.org/blog/2025/02/eslint-css-support/

https://github.com/eslint/css

```
$ eslint .

/Users/simon/src/Leaflet/dist/leaflet.css
   23:2  warning  Property 'user-select' is not a widely available baseline feature         css/use-baseline
  228:2  warning  Property 'outline' is not a widely available baseline feature             css/use-baseline
  532:2  warning  Property 'user-select' is not a widely available baseline feature         css/use-baseline
  603:3  warning  Property 'print-color-adjust' is not a widely available baseline feature  css/use-baseline

/Users/simon/src/Leaflet/docs/docs/css/main.css
   829:14  warning  Value 'break-word' of property 'word-break' is not a widely available baseline feature  css/use-baseline
  1020:15  warning  Value 'break-word' of property 'word-break' is not a widely available baseline feature  css/use-baseline
  1116:23  warning  Selector 'has' is not a widely available baseline feature                               css/use-baseline

/Users/simon/src/Leaflet/docs/docs/css/normalize.css
  97:3  warning  Property 'outline' is not a widely available baseline feature  css/use-baseline
```

Fixes #7806.

PR #9991, opened 2025-11-22
Enable `eslint-plugin-baseline-js` linting

simon04:
https://github.com/3ru/eslint-plugin-baseline-js

Relates to https://github.com/Leaflet/Leaflet/issues/9990.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -87,7 +87,11 @@ export default [
 		rules: {
 			'baseline-js/use-baseline': ['error', {
 				available: 'widely',
-				includeWebApis: {preset: 'auto'},
+				includeWebApis: {preset: 'auto', ignore: [
+					// According to https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio, the feature is only partially supported in Safari:
+					// In Safari on iOS, the devicePixelRatio does not change when the page is zoomed. See bug https://webkit.org/b/124862.
+					'devicepixelratio',
+				]},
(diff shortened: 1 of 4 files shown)


mourner (approved):
Surprised `devicePixelRatio` is not baseline, isn't it supported since forever?

simon04:
According to https://developer.mozilla.org/en-US/docs/Web/API/Window/devicePixelRatio, the feature is only partially supported in Safari:

> This feature is not Baseline because it does not work in some of the most widely-used browsers.

> In Safari on iOS, the devicePixelRatio does not change when the page is zoomed. See [bug 124862](https://webkit.org/b/124862).

mourner:
@simon04 sound a bit contrived of a reason for removing it from baseline... I wonder if we can add exceptions like this to the plugin config instead of adding eslint comment disables.

simon04:
> I wonder if we can add exceptions like this to the plugin config instead of adding eslint comment disables.

I finally managed to do so (in the 3rd attempt). It's confusing: https://github.com/3ru/eslint-plugin-baseline-js/issues/58

Generally speaking, in your opinion, when do you think leaflet 2.0 will be released? To me it seems that everything is blocked because a decision is pending in PR `#9906`.

jonkoops:
Should we also be linting the CSS? See https://web.dev/blog/eslint-baseline-integration

simon04:
I think we're already doing so: https://github.com/Leaflet/Leaflet/pull/9668

Issue #9930, opened 2025-10-06
Fractional zoom level in tile URLs.


vagran:
### Checklist

- [x] I've looked at the [documentation](https://leafletjs.com/reference.html) to make sure the behavior isn't documented and expected.
- [x] I'm sure this is an issue with Leaflet, not with my app or other dependencies (Angular, Cordova, React, etc.).
- [ ] I've searched through the [current issues](https://github.com/Leaflet/Leaflet/issues) to make sure this hasn't been reported yet.
- [x] I agree to follow the [Code of Conduct](https://github.com/Leaflet/Leaflet/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Steps to reproduce

When having non-integer `zoomSnap` option and calling `map.redraw()` method, it may result in fractional zoom level in tile URL, like `2.4396229271192125/0/2`. It can happen after view is set by `setView()`, `fitBounds()` or similar methods. Zooming in and out by mouse wheel normalizes the situation and map is loaded.
Seems this place in code is missing `Math.round()` call:
https://github.com/Leaflet/Leaflet/blob/2f5594810c26aa87019582d381960f1b10c1bf62/src/layer/tile/GridLayer.js#L235

### Expected behavior

Tiles are always requested with integer coordinates.

### Current behavior

Tiles may be requested with fractional Z coordinate.

### Environment

- Leaflet version: 1.9.4

vagran:
Workaround with monkey-patching if somebody encountered this bug:
```js
    L.GridLayer.prototype.redraw = function() {
        if (this._map) {
			this._removeAllTiles();
			const tileZoom = this._clampZoom(Math.round(this._map.getZoom()));
			if (tileZoom !== this._tileZoom) {
				this._tileZoom = tileZoom;
				this._updateLevels();
			}
			this._update();
		}
		return this;
    }
```

PR #9997, opened 2025-11-24
Fix fractional tile zoom

Adamnson:
## Summary

This PR fixes an issue where `GridLayer.redraw()` requests tiles using a
fractional zoom level when `zoomSnap` allows non-integer zooms (e.g., 0.1).
This leads to tile URLs like `/2.4/x/y.png`, which tile servers reject.

## What was happening

- When zoomSnap is fractional, `map.getZoom()` returns values like `2.4`.
- `GridLayer.redraw()` uses this zoom value directly.
- Tile URLs generated from `{z}` contain fractional zooms.
- Most servers expect integer zoom levels → 400 errors.

This behavior matches the problem documented in #9930.

## Fix

Before clamping the zoom, the value is rounded:

```js
const tileZoom = this._clampZoom(Math.round(this._map.getZoom()));

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -228,7 +228,7 @@ export var GridLayer = Layer.extend({
 	redraw: function () {
 		if (this._map) {
 			this._removeAllTiles();
-			var tileZoom = this._clampZoom(this._map.getZoom());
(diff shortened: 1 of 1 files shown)


Adamnson:
Will reopen with updated main

PR #9998, opened 2025-11-24
Fix fractional zoom level used in GridLayer.redraw()

### LINKED ISSUE #9930: Fractional zoom level in tile URLs.

Adamnson:
GridLayer.redraw() currently uses this._map.getZoom() directly, which can contain fractional values when zoomSnap is set to a fractional increment (e.g., 0.1). This causes tile URLs to be generated with fractional zoom levels, such as "/2.4/x/y.png", which tile servers do not accept.

To ensure tile requests always use integer zoom levels, this patch rounds the zoom value before clamping it. This results in correct tile URLs like "/2/x/y.png" during redraw, while preserving normal zoom behavior.

Fixes #9930.

### DIFF
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -232,7 +232,7 @@ export class GridLayer extends Layer {
 	redraw() {
 		if (this._map) {
 			this._removeAllTiles();
-			const tileZoom = this._clampZoom(this._map.getZoom());
+			const tileZoom = this._clampZoom(Math.round(this._map.getZoom()));
 			if (tileZoom !== this._tileZoom) {
 				this._tileZoom = tileZoom;
 				this._updateLevels();


src/layer/tile/GridLayer.js
```diff
@@ -232,7 +232,7 @@ export class GridLayer extends Layer {
 	redraw() {
 		if (this._map) {
 			this._removeAllTiles();
-			const tileZoom = this._clampZoom(this._map.getZoom());
+			const tileZoom = this._clampZoom(Math.round(this._map.getZoom()));
```
jonkoops, line 235: Just thinking out loud. Instead of rounding the number, perhaps we should ceil it? That way, the resolution of the tile will always be high enough, and not get blurry. Then again, you might want to get a lower zoom level tile, not quite sure.

Adamnson: I approached this mainly from a data-sanitization perspective, so round() felt like the most natural choice.

If you’d prefer using ceil() instead, I’m happy to update the PR accordingly.

Falke-Design:
Related PRs: #6118, #6505

Falke-Design:
Here we added the rounding only for the TileLayer:  #8613

Adamnson:
Hi!
Just bumping this — the fix addresses fractional zoom handling in GridLayer.redraw().

Happy to refine or follow up if there are concerns. Thanks!

PR #10006, opened 2025-12-02
Remove `Class.extend()` entirely

jonkoops:
Removes `Class.extend()` entirely, as it is fundamentally incompatible with regular class constructors (see #9906). This change is a prerequisite to a follow-up PR, which will replace all `initialize()` methods with a proper class constructor. The documentation has also been re-written to reflect the removal of `Class.extend()`.

### DIFF
--- a/docs/examples/extending-1-classes/index.md
+++ b/docs/examples/extending-1-classes/index.md
@@ -5,200 +5,175 @@ title: Extending Leaflet, Class Theory
 
 ## Extending Leaflet
 
-Leaflet has literally hundreds of plugins. These expand the capabilities of Leaflet: sometimes in a generic way, sometimes in a very use-case-specific way.
+Leaflet has literally hundreds of plugins. These expand the capabilities of Leaflet: sometimes in a generic way, sometimes in a very use-case-specific way. Part of the reason there are so many plugins is that Leaflet is easy to extend. This tutorial will cover the most commonly used ways of doing so.
 
-Part of the reason there are so many plugins is that Leaflet is easy to extend. This tutorial will cover the most commonly used ways of doing so.
-
-Please note that this tutorial assumes you have a good grasp of:
-
-* [JavaScript](https://developer.mozilla.org/en-US/Learn/JavaScript)
-* [DOM handling](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Introduction)
-* [Object-oriented programming](https://en.wikipedia.org/wiki/Object-oriented_programming) (understanding concepts like classes, instances, inheritance, methods and properties)
-
+Please note that this tutorial assumes you have a good grasp of [JavaScript](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting), and especially [JavaScript classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_classes).
 
 ## Leaflet architecture
 
-Let's have a look at a simplified UML Class diagram for Leaflet 1.0.0. There are more than 60 JavaScript classes, so the diagram is a bit big. Luckily we can make a zoomable image with a `L.ImageOverlay`:
+Let's have a look at a simplified UML Class diagram for Leaflet. There are more than 60 classes, so the diagram is a bit big. Luckily we can make a zoomable image with a `L.ImageOverlay`:
 
 {% include frame.html url="class-diagram.html" %}
 
-
 From a technical point of view, Leaflet can be extended in different ways:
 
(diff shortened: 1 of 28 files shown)


docs/examples/extending-1-classes/index.md
```diff
@@ -5,200 +5,175 @@ title: Extending Leaflet, Class Theory
...
+* Method and property names should be in `lowerCamelCase`.
+* Private properties and methods start with an underscore (`_`). This indicates they're internal and shouldn't be used directly.
+
+### Setting default options
+
+All classes that extend from `L.Class` can be provided with options, which will automatically be merged when constructed:
+
+```js
+class MyBox extends Class {
+	static {
```
IvanSanchez, line 61: This is the first time I see this syntax (executable statements inside a `static` block). I think it'd be a good idea to link to some documentation (MDN?) about this JS feature.

src/core/Class.leafdoc
```diff
@@ -3,22 +3,22 @@
 
 Class powers the OOP facilities of Leaflet and is used to create almost all of the Leaflet classes documented here.
 
-In addition to implementing a simple classical inheritance model, it introduces several special properties for convenient code organization — options, includes and statics.
+It implements a simple classical inheritance model using JavaScript's standard `extends` keyword, and introduces several special properties for convenient code organization — options and init hooks.
```
IvanSanchez, line 6: I kinda disagree with this, because `L.Class` right now is mostly legacy. Not sure how to word it :-S

jonkoops: Yeah, `L.Class` will progressively be less present. Perhaps we even remove it in a future version. For now, it still contains several APIs likely used extensively by plugins.

docs/examples/extending-1-classes/index.md
jonkoops: I added a link to the relevant MDN page :+1:

docs/examples/extending-1-classes/index.md
```diff
@@ -5,200 +5,175 @@ title: Extending Leaflet, Class Theory
...
+	}
+});
+
+const instance = new MyLayer();
+console.log(instance.incrementCount()); // Outputs "2"
+```
+
+Note: Use `.include()` sparingly, as modifying base classes can have unexpected side effects. As a general rule of thumb, try to extend existing classes instead.
+
+### Initialization hooks
```
IvanSanchez, line 132: I'm wondering if we should rephrase this whole section, and mark "init hooks" as legacy.

Nowadays I'd expect any initialization to happen in the constructor, and the developer is supposed to call `super.constructor()`.

simon04:
I fear we're changing to much for the Leaflet 2 release. Plugin authors will have a hard job to support Leaflet 1.x and 2.0 without maintaining two codebases, or dropping 1.x support entirely, or ignoring 2.0 support. And Leaflet users cannot upgrade to 2.0 until all plugins are compatible.

To me, the situation seems very similar to the Python 2 to 3 migration debacle. :worried:

Maxel01:
In my opinion, it is not possible to maintain plugin compatibility with both versions. It's possible to use v1 plugins with polyfill, but I don't think it's possible to use v2 plugins in leaflet v1. The codebase needs to be splitted anyway.

It might be best to modernize the core of leaflet and deprecate all legacy functions. These deprecations could be removed in the next major version.
Another possibility is to remove these functions like `Class.extend` now and add them to the polyfill. In that way, it would still be possible to use v1 plugins in v2.

The leaflet plugin system is indeed a problem. Many plugins are still in use but the maintainers abandoned many of them some time ago.

docs/examples/extending-1-classes/index.md
Falke-Design: How would you solve, for example, extending the Marker initialization without creating a new class? Many plugins rely on this logic, and it is very useful for cross-plugin interactions.

Falke-Design:
@Maxel01 

> It's possible to use v1 plugins with polyfill [...]

By forcing a switch to the default JavaScript class system, many features will stop working and cannot be re-added through a polyfill. `Class.extend` could be provided as a polyfill but with the next PR `initialize` will be removed and then `extend` will no longer work. See: #9906

Maxel01:
@Falke-Design I understand....

jonkoops:
I don't think we can get around making this change, unfortunately. As discussed under #9906, standardized JavaScript classes deviate too much from Leaflet's own implementation, to a point where we need to make a choice between modernizing the code, or essentially having our own implementation forever incompatible with standard JS.

jonkoops:
I am merging this in, it has been discussed in great detail enough, and I think the core team is in agreement that we need to make this cut.

PR #9648, opened 2025-04-18
Add blog and migration information for V2

Falke-Design:
Fixes: #9625

PR #10011, opened 2025-12-08
Migrate build from rollup to rolldown

simon04:
https://rolldown.rs/

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,6 @@
   "devDependencies": {
     "@eslint/css": "^0.14.1",
     "@mapbox/eslint-plugin-script-tags": "^1.0.0",
-    "@rollup/plugin-json": "^6.1.0",
-    "@rollup/plugin-terser": "^0.4.4",
     "bundlemon": "^3.1.0",
     "chai": "^6.2.2",
     "eslint": "^9.39.2",
@@ -28,7 +26,7 @@
     "mocha": "^11.7.5",
     "prosthetic-hand": "^2.0.0",
     "requirejs": "^2.3.8",
-    "rollup": "^4.54.0",
+    "rolldown": "^1.0.0-beta.57",
     "sinon": "^21.0.1",
     "ssri": "^13.0.0",
     "ui-event-simulator": "^2.0.0"
@@ -51,7 +49,7 @@
     "docs": "node ./build/docs.js && node ./build/integrity.js",
     "test": "karma start ./spec/karma.conf.cjs",
     "coverage": "karma start ./spec/karma.conf.cjs --coverage --single-run",
-    "build": "rollup --config build/rollup-config.js",
+    "build": "rolldown --config build/rolldown-config.js",
     "watch": "npm run build -- --watch",
     "lint": "eslint .",
     "lintfix": "npm run lint -- --fix",
--- a/build/rolldown-config.js
+++ b/build/rolldown-config.js
@@ -1,7 +1,6 @@
-import json from '@rollup/plugin-json';
-import terser from '@rollup/plugin-terser';
+// @ts-check
 import {readFileSync} from 'node:fs';
-import {defineConfig} from 'rollup';
+import {defineConfig} from 'rolldown';
 import pkg from '../package.json' with {type: 'json'};
 
 const STATIC_ASSETS = [
@@ -12,7 +11,7 @@ const STATIC_ASSETS = [
 	'images/marker-shadow.svg'
 ];
 
-/** @type {import('rollup').OutputPlugin} */
+/** @type {import('rolldown').Plugin} */
 const staticAssetsPlugin = {
 	name: 'static-assets',
 	generateBundle() {
@@ -29,13 +28,13 @@ const banner = `/* @preserve
  */
 `;
 
-/** @type {import('rollup').OutputOptions} */
+/** @type {import('rolldown').OutputOptions} */
 const commonOptions = {
 	banner,
 	sourcemap: true
 };
 
-/** @type {import('rollup').OutputOptions} */
+/** @type {import('rolldown').OutputOptions} */
 const umdOptions = {
 	...commonOptions,
 	name: 'L',
@@ -49,7 +48,6 @@ const umdOptions = {
 
 export default defineConfig({
 	input: './src/Leaflet.js',
-	plugins: [json()],
 	output: [
 		{
 			...commonOptions,
@@ -59,7 +57,7 @@ export default defineConfig({
 		{
 			...commonOptions,
 			file: './dist/leaflet.js',
-			plugins: [terser()],
+			minify: true
 		},
 		{
 			...umdOptions,
@@ -68,7 +66,7 @@ export default defineConfig({
 		{
 			...umdOptions,
 			file: './dist/leaflet-global.js',
-			plugins: [terser()],
+			minify: true
 		}
 	]
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,8 +11,6 @@
       "devDependencies": {
         "@eslint/css": "^0.14.1",
         "@mapbox/eslint-plugin-script-tags": "^1.0.0",
-        "@rollup/plugin-json": "^6.1.0",
-        "@rollup/plugin-terser": "^0.4.4",
         "bundlemon": "^3.1.0",
         "chai": "^6.2.2",
         "eslint": "^9.39.2",
@@ -33,7 +31,7 @@
         "mocha": "^11.7.5",
         "prosthetic-hand": "^2.0.0",
         "requirejs": "^2.3.8",
-        "rollup": "^4.54.0",
+        "rolldown": "^1.0.0-beta.57",
         "sinon": "^21.0.1",
         "ssri": "^13.0.0",
         "ui-event-simulator": "^2.0.0"
@@ -419,21 +417,21 @@
       }
     },
     "node_modules/@emnapi/core": {
-      "version": "1.4.5",
-      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.4.5.tgz",
-      "integrity": "sha512-XsLw1dEOpkSX/WucdqUhPWP7hDxSvZiY+fsUC14h+FtQ2Ifni4znbBt8punRX+Uj2JG/uDb8nEHVKvrVlvdZ5Q==",
+      "version": "1.7.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.7.1.tgz",
+      "integrity": "sha512-o1uhUASyo921r2XtHYOHy7gdkGLge8ghBEQHMWmyJFoXlpU58kIrhhN3w26lpQb6dspetweapMn2CSNwQ8I4wg==",
       "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
-        "@emnapi/wasi-threads": "1.0.4",
+        "@emnapi/wasi-threads": "1.1.0",
         "tslib": "^2.4.0"
       }
     },
     "node_modules/@emnapi/runtime": {
-      "version": "1.4.5",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.4.5.tgz",
-      "integrity": "sha512-++LApOtY0pEEz1zrd9vy1/zXVaVJJ/EbAF3u0fXIzPJEDtnITsBGbbK0EkM72amhl/R5b+5xx0Y/QhcVOpuulg==",
+      "version": "1.7.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.7.1.tgz",
+      "integrity": "sha512-PVtJr5CmLwYAU9PZDMITZoR5iAOShYREoR45EyyLrbntV50mdePTgUn4AmOw90Ifcj+x2kRjdzr1HP3RrNiHGA==",
       "dev": true,
       "license": "MIT",
       "optional": true,
@@ -442,9 +440,9 @@
       }
     },
     "node_modules/@emnapi/wasi-threads": {
-      "version": "1.0.4",
-      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.0.4.tgz",
-      "integrity": "sha512-PJR+bOmMOPH8AtcTGAyYNiuJ3/Fcoj2XN/gBEWzDIKh254XO+mM9XoXHk5GNEhodxeMznbg7BlRojVbKN+gC6g==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.1.0.tgz",
+      "integrity": "sha512-WI0DdZ8xFSbgMjR1sFsKABJ/C5OnRrjT06JXbZKexJGrDuPTzZdDYfFlsgcCXCyf+suG5QU2e/y1Wo2V/OapLQ==",
       "dev": true,
       "license": "MIT",
       "optional": true,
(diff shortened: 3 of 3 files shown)


src/Leaflet.js
```diff
@@ -1,5 +1,4 @@
-// !!! NEXT LINE IS AUTO-GENERATED VIA `NPM VERSION` !!!
-export const version = '2.0.0-alpha.1';
+export {version} from '../package.json' with {type: 'json'};
```
IvanSanchez: If I recall correctly, one of the issues with the `rollup` build was that getting the version field from `package.json` did import *all* of package.json into the bundle.

I guess you've tested that rolldown tree-shakes the rest of package.json, right?

simon04: I did, indeed. Since https://github.com/rolldown/rolldown/releases/tag/v1.0.0-beta.44 "JSON default imports are now tree-shakeable", too.

```
> grep -wC5 version dist/leaflet-src.js
        return target;
};

//#endregion
//#region package.json
var version = "2.0.0-alpha.1";

//#endregion
```

simon04:
Here's how the bundle sizes change:

```
# rollup
> /usr/bin/ls -l dist/*.js
-rw-r--r-- 1 simon simon 137005 2025-12-09 07:43 dist/leaflet-global.js
-rw-r--r-- 1 simon simon 411403 2025-12-09 07:43 dist/leaflet-global-src.js
-rw-r--r-- 1 simon simon 144988 2025-12-09 07:43 dist/leaflet.js
-rw-r--r-- 1 simon simon 398799 2025-12-09 07:43 dist/leaflet-src.js

# rolldown removes the comments from the source files
-rw-r--r-- 1 simon simon 139276 2025-12-09 07:44 dist/leaflet-global.js
-rw-r--r-- 1 simon simon 223422 2025-12-09 07:44 dist/leaflet-global-src.js
-rw-r--r-- 1 simon simon 138690 2025-12-09 07:44 dist/leaflet.js
-rw-r--r-- 1 simon simon 215264 2025-12-09 07:44 dist/leaflet-src.js
```

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
...
+			minify: true,
+			sourcemap: true,
+			freeze: false
+		},
+		{
+			file: './dist/leaflet-global.js',
+			name: 'leaflet',
+			format: 'umd',
+			banner,
+			minify: false,
```
Falke-Design:
```suggestion
			minify: true,
```

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
...
+			sourcemap: true,
+			freeze: false,
+			esModule: false
+		},
 		{
 			file: './dist/leaflet-global-src.js',
 			name: 'leaflet',
 			format: 'umd',
 			banner,
+			minify: true,
```
Falke-Design:
```suggestion
			minify: false,
```

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
...
 	output: [
 		{
 			file: pkg.exports['.'],
 			format: 'es',
 			banner,
+			minify: false,
 			sourcemap: true,
 			freeze: false
 		},
+		{
```
Falke-Design: Currently these outputs are added only if `release` is `true`.

Falke-Design (changes requested):

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
...
-/** @type {import('rollup').RollupOptions} */
-const config = {
+export default defineConfig({
 	input: 'src/LeafletWithGlobals.js',
 	output: [
 		{
 			file: pkg.exports['.'],
 			format: 'es',
 			banner,
+			minify: false,
```
jonkoops: Let's not provide a value if we can use the default instead.
```suggestion
```

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
-import json from '@rollup/plugin-json';
 import {readFileSync} from 'node:fs';
-import rollupGitVersion from 'rollup-plugin-git-version';
+import {defineConfig} from 'rolldown';
 import {simpleGit} from 'simple-git';
+import pkg from '../package.json' with {type: 'json'};
```
jonkoops: Using the `with` keyword is not supported by the ECMAScript version configured by the linter.  @mourner added an explicit version under 54f49b5bd42f6996c0bbf59b7570a589f06fc95e, but I think we can simply omit it and fall back to the default, which is always the latest version.

Please remove the following sections of the ESLint configuration:

https://github.com/Leaflet/Leaflet/blob/e226f77aa0339bf7e46027e87430a19c42b50f43/eslint.config.js#L13-L17

https://github.com/Leaflet/Leaflet/blob/e226f77aa0339bf7e46027e87430a19c42b50f43/eslint.config.js#L114-L121

package.json
```diff
@@ -52,17 +49,14 @@
...
-    "build": "npm run rollup && npm run uglify && npm run uglify-global",
+    "build": "npm run rolldown",
     "lint": "eslint .",
     "lintfix": "npm run lint -- --fix",
-    "rollup": "rollup -c build/rollup-config.js",
-    "watch": "rollup -w -c build/rollup-config.js",
-    "uglify": "uglifyjs dist/leaflet-src.js -c -m -o dist/leaflet.js --source-map filename=dist/leaflet.js.map --source-map content=dist/leaflet-src.js.map --source-map url=leaflet.js.map --comments",
-    "uglify-global": "uglifyjs dist/leaflet-global-src.js -c -m -o dist/leaflet-global.js --source-map filename=dist/leaflet-global.js.map --source-map content=dist/leaflet-global-src.js.map --source-map url=leaflet-global.js.map --comments",
+    "rolldown": "rolldown -c build/rollup-config.js",
+    "watch": "rolldown -w -c build/rollup-config.js",
```
jonkoops: Let's not add a script to run a single command, we can also reduce the watch now that it can call `build` directly.

```suggestion
    "build": "rolldown -c build/rollup-config.js",
    "watch": "npm run build -- -w",
```

package.json
```diff
@@ -52,17 +49,14 @@
     "docs": "node ./build/docs.js && node ./build/integrity.js",
     "test": "karma start ./spec/karma.conf.cjs",
     "coverage": "karma start ./spec/karma.conf.cjs --coverage --single-run",
-    "build": "npm run rollup && npm run uglify && npm run uglify-global",
+    "build": "npm run rolldown",
```
jonkoops:
```suggestion
```

build/rollup-config.js
jonkoops: This is the default, so we can omit this option entirely.

build/rollup-config.js
```diff
@@ -1,38 +1,53 @@
-import json from '@rollup/plugin-json';
 import {readFileSync} from 'node:fs';
-import rollupGitVersion from 'rollup-plugin-git-version';
```
jonkoops: @IvanSanchez if we merge this, you can probably deprecate this package.

build/rollup-config.js
jonkoops: The build is much faster, even with these changes, but I agree, the production build should be conditional.

jonkoops:
I am a big supporter of using Rolldown, it is much faster! Couple of notes though:

- I am seeing some validation errors when running the build.
- Assets are emitted multiple times.

build/rollup-config.js
IvanSanchez: @jonkoops Noted!

IvanSanchez: @jonkoops On second thought: my rollup git version plugin appended the git branch and commit hash to the version, which I found useful. Do we want to put work into getting that feature back in?

jonkoops: The commit hash should be sufficient. To be very honest, I don't exactly recall why we did this. For nightly versions I imagine?

simon04: We could get rid of `simple-git`, too, if we perform an `npm version ...` prior to an `npm build` in `.github/workflows/main.yml`:

```
> git describe --always
v2.0.0-alpha.1-91-ga96e00d4
> npm version --no-git-tag-version $(git describe --always)
v2.0.0-alpha.1-91-ga96e00d4
```

simon04: > We could get rid of `simple-git`, too, if we perform an `npm version ...` prior to an `npm build` in `.github/workflows/main.yml`

Done in 5a966ea09c54e7d9375c936c80bc9c9d65d67259.

.github/workflows/main.yml
```diff
@@ -21,6 +23,9 @@ jobs:
           check-latest: true
           cache: npm
 
+      - name: Set version in package.json using `git describe`
+        run: npm version --no-git-tag-version $(git describe --always)
```
jonkoops: Running `git describe --always` locally results in a string like `v2.0.0-alpha.1-95-g517d5cc0`, which is a great unique identifier, but it is not a valid [semantic version](https://semver.org/). It does seem to work, though, and `npm version` trims the invalid parts of the string, so I guess it is fine?

That said, we should try and skip this step when we're doing a tagged release, can we make this conditional like the `publish-npm` job?

build/rollup-config.js
jonkoops: This sounds like a good approach, I consider this resolved. @IvanSanchez wdyt?

build/rollup-config.js
```diff
@@ -1,38 +1,49 @@
...
+export default defineConfig({
 	input: 'src/LeafletWithGlobals.js',
 	output: [
 		{
 			file: pkg.exports['.'],
 			format: 'es',
 			banner,
+			sourcemap: true
+		},
+		release && {
```
jonkoops: If `false` is passed here I still see chunks being emitted, I think we should make this actually conditional instead of passing placeholders:

```js
/** @type {import('rolldown').OutputOptions[]} */
const output = [{
	file: pkg.exports['.'],
	format: 'es',
	banner,
	sourcemap: true,
}];

if (release) {
	output.push({
		file: './dist/leaflet.js',
		format: 'es',
		banner,
		minify: true,
		sourcemap: true
	}, {
		file: './dist/leaflet-global.js',
		name: 'leaflet',
		format: 'umd',
		banner,
		minify: true,
		sourcemap: true,
		esModule: false
	}, {
		file: './dist/leaflet-global-src.js',
		name: 'leaflet',
		format: 'umd',
		banner,
		sourcemap: true,
		esModule: false
	});
}
```

build/rollup-config.js
```diff
@@ -1,38 +1,49 @@
...
 			file: './dist/leaflet-global-src.js',
 			name: 'leaflet',
 			format: 'umd',
 			banner,
 			sourcemap: true,
-			freeze: false,
 			esModule: false
 		}
 	],
 	plugins: [
```
jonkoops: Passing this plugin duplicates it for each output, which means the static assets are emitted multiple times. This can be fixed by adding the `plugins` to the first output in the array instead.

src/Leaflet.js
```diff
@@ -1,5 +1,5 @@
-// !!! NEXT LINE IS AUTO-GENERATED VIA `NPM VERSION` !!!
-export const version = '2.0.0-alpha.1';
+import pkg from '../package.json' with {type: 'json'};
+export const version = pkg.version;
```
jonkoops, line 2: I believe this can be shortened to:

```js
export {version} from '../package.json' with {type: 'json'};
```

simon04: I've already tried this, but it broke all unit tests. See 3f86bcc66644443cecbf56c91b9b180831b2252c.

build/rollup-config.js
simon04: Amazing. Did not know that each output can have it's own plugins. Fixed via c9397c442f2d48fa2097ba5043651b42e6fad79f.

build/rollup-config.js
simon04: Good catch. I wonder whether the 25ms time difference between 1 and 4 outputs is worth the added complexity? I would strive for simplicity and build all 4 outputs unconditionally.

```
✔ rolldown v1.0.0-beta.53 Finished in 74.81 ms                                                                                                                                                                          
✔ rolldown v1.0.0-beta.53 Finished in 100.24 ms                                                                                                                                                                         
```

.github/workflows/main.yml
simon04: `git describe` on tag v1.9.4 yields `v1.9.4` (as intended), so we should be fine?

```
> git checkout v1.9.4
HEAD is now at d15112c9 1.9.4

> git describe --always
v1.9.4
```

build/rollup-config.js
jonkoops: You know what, I agree, this is so fast I don't think we really need to care. Let's leave it as-is.

jonkoops:
@simon04 I'm going to do a quick squash and add some of the discussed changes so we can get this merged in.

simon04:
@jonkoops, amazing, thank you.

How can I make myself useful for other outstanding issues or pull requests? I've got plenty of spare time and Tyrol offers too little snow for outdoor activities at the moment... ;-)

jonkoops:
What would really help us with the v2 release is if we can get a list of all the commits that we have added since v1 that introduce new features and/or breaking changes, and to collect those in a single issue with the related PR and issues linked. We'll need this to get to the release candidate phase, so that we can provide users with a proper migration guide and release announcement blog post.

simon04:
> get a list of all the commits that we have added since v1 that introduce new features and/or breaking changes

I've worked on a public API diff in https://github.com/Leaflet/Leaflet/pull/9648#issuecomment-2938464747. Maybe this should be updated again?

Some migration instructions are covered in the blog post on 2.0.0 alpha: https://leafletjs.com/2025/05/18/leaflet-2.0.0-alpha.html

jonkoops:
Nice, yes, I think we need an update of that. Perhaps you could log an issue for the 2.x migration guide and blog post and put the information there?

simon04:
I've updated https://gist.github.com/simon04/b5e9cfbcf2583b822f829e0e678f1370 to the current main branch `2.0.0-alpha.1-87-gf33fbefb`.

src/Leaflet.js
jonkoops: Got it, ok let's consider this resolved since Rolldown manages to eliminate the rest of the code.

jonkoops:
So I have been doing some digging, and it seems that the bundle that is currently emitted by Rollup (on `main`) as well as the one emitted here by Rolldown, do not work properly for UMD setups. I'm going to raise a PR to fix this on `main` first, and then we'll see how we can adapt this to Rolldown.

jonkoops:
@simon04 I had to do some fixes to the original Rollup build (see #10027, #10028, #10029) to make sure all the various distributions work correctly. It looks like Rolldown might not support all the options required to make this work (yet), specifically the [`noConflict`](https://rollupjs.org/configuration-options/#output-noconflict) and [`amd.id`](https://rollupjs.org/configuration-options/#output-amd-id) options, so I am marking this PR as draft for now until these are implemented. Perhaps you could take a look at logging some issues in the Rollup project to add support for these options, or perhaps even fix them?

Even without Rolldown, I think this PR has some valuable additions, so I will be splitting the chunks of the work you have done here into separate PRs that we can land individually.

jonkoops:
@IvanSanchez I've removed `rollup-plugin-git-version` under 1b3875fb50b6c10cfdfbec67ad4be3151eec3b5c (accidentally pushed to `main`, set up a branch protection rule to prevent that from happening again). Since we're using the script @Falke-Design wrote to set the version under #9717, we are no longer importing the package.json, and thus this plugin never did anything since.

You might want to deprecate `rollup-plugin-git-version` since AFAIK we were the only users.

build/rollup-config.js
jonkoops: Marking this as resolved, the plugin was not used anyway, see https://github.com/Leaflet/Leaflet/pull/10011#issuecomment-3697212016

.github/workflows/main.yml
jonkoops: We no longer need this, as I have removed the git related logic entirely in other PRs.

jonkoops:
@simon04 I have split out the work you did here, and some other enhancements into separate PRs that have landed on `main` (see #10027, #10028, #10029, #10037, #10038, 1b3875fb50b6c10cfdfbec67ad4be3151eec3b5c, #10041, #10042), that are all applied to the existing Rollup build. I have rebased this PR so the only work left is now purely the migration from Rollup to Rolldown.

Looking at the final build, it seems that Rolldown is missing several features that would have to be implemented in order for this PR to be merged, namely:

- The `freeze` option is not supported, meaning the exports of the ESM build are mutable.
- The `amd.id` option is not supported, meaning the AMD distribution doesn't work (see `test/debug/dist-amd.html`).
- The `noConflict` option is not supported, meaning the `noConflict()` method is not added to the namespace for the UMD distribition.

simon04:
Hi @jonkoops, amazing, thank you for your work. Considering that builds are performed only so often, and that the rollup build is sufficiently fast (914ms on an Intel i5-14600K), I think we can de-prioritize the migration to rolldown.

jonkoops:
Agreed, it would still be nice to move in the future once these features are implemented. I believe the goal is to be fully Rollup compatible in the long run, so it should 'just work' in the long run. It would be nice if someone could report these compatibility issues on their repo, and add them here for tracking. I am currently too tied down on other stuff to do so :wink:

simon04:
I've submitted the following upstream feature requests:

* https://github.com/rolldown/rolldown/issues/7722
* https://github.com/rolldown/rolldown/issues/7723
* https://github.com/rolldown/rolldown/issues/7724

jonkoops:
Appreciate it!

PR #10021, opened 2025-12-21
chore: add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

claude used Leaflet for [my little project](https://github.com/yarikoptic/strava-backup) and I decided to say Thank you! ;)


docs/_plugins/bookmarked-pan-zoom/leaflet-showall.md
```diff
@@ -2,7 +2,7 @@
 name: Leaflet.ShowAll
 category: bookmarked-pan-zoom
 repo: https://github.com/florpor/Leaflet.ShowAll
-author: Mor Yariv
+author: More Yariv
```
Falke-Design: That is a name - not misspelled

yarikoptic: sorry for missing that ! added skip to skip all such lines , redone, forcepushed

PR #10029, opened 2025-12-25
Remove default export from ESM distribution

jonkoops:
Removes the default `L` export from the ESM distribution, opting for named exports only. This makes it easier to tree-shake code that is imported from the package in the future. This also freezes the ESM distribution, so exported modules cannot be modified, which is expected for modules, and prevents side effects from being introduced by external code.

The UMD distribution is unaffected, and still allows external code to modify the `L` namespace and its members, and still exposes the `L` global.

This PR is a split off from #10027 and #9880.


jonkoops:
@Leaflet/core @Leaflet/leaflet-maintainers could I get a review for this one so I can continue fixing up the build?

jonkoops:
I am going to bypass the review and just get this in, I want to start making some progress on the last impediments here. If there are any latent comments, do let me know, and I will address them in a separate PR.

PR #10030, opened 2025-12-27
Plugins: Update Leaflet.DirectionMarker for v2

schierlm:


jonkoops:
Hi @schierlm, we're holding off on merging further PRs with 2.0 support for plugins, as we are likely making some more API changes while in alpha. See https://github.com/Leaflet/Leaflet/issues/9842#issuecomment-3302673291. Once we're in a beta/release candidate stage, I think we can commit to landing these again.

schierlm:
Thanks for the feedback. 

Would have been nice to update (or at least mention) all those issues you spammed into hundreds of repositories. Now I finally found time to fully update my plugin to V2-alpha1 and thorougly test it. Would have held this off (committed a WIP fix before that comment and that should be enough for "early adopters") if I knew I'll have to do this again later and more changes to be expected.

Merry Christmas anyway :)

PR #10041, opened 2025-12-30
Replace `uglify-js` scripts with `@rollup/plugin-terser`

jonkoops:
Replaces the minification scripts using `uglify-js` with the official `@rollup/plugin-terser` package, reducing the complexity of the scripts, and speeding up the build.  Cleans up the Rollup configuration so it is less verbose by eliminating options that can safely be removed in order to fall to defaults, and by placing common options in shared defaults.

Additionally, the static asset plugin is now only used for the first output, ensuring assets are only emitted once for the entire build.


jonkoops:
I'm going to merge this in as a rapid fire of a couple or PRs related to the build system, if there are any comments let me know, and I will address them in another PR.

PR #10042, opened 2025-12-30
Replace custom versioning with import of package version

jonkoops:
Replaces the version script introduced under #9717 with a JSON import of the package file. Even though this feature is only newly available in Baseline, this should not be an issue, as we bundle up the code and Rollup inlines the version number. This also removes the logic to use a specific branch and commit from Git for development builds, as this is largely left unused (e.g. the nightly doesn't actively use this).


jonkoops:
I'm going to merge this in as a rapid fire of a couple or PRs related to the build system, if there are any comments let me know, and I will address them in another PR.

@simon04 FYI, this takes some of your work from #10011. So I added you as a co-author of the commit.

PR #10044, opened 2026-01-01
Use standard `constructor()` instead of `initialize()` with `async callInitHooks()`

simon04:
A proof-on-concept to use standard `constructor()` instead of `initialize()` (based on PR https://github.com/Leaflet/Leaflet/pull/9906) with `async callInitHooks()`.

Making `callInitHooks()` an `async function`, and awaiting for a Promise to be resolved in the `Class` constructor body (`_initHooksReady.resolve()`), seems to ensure that it is called after all constructor bodies are executed.

### DIFF
--- a/spec/suites/map/handler/Map.KeyboardSpec.js
+++ b/spec/suites/map/handler/Map.KeyboardSpec.js
@@ -21,6 +21,7 @@ describe('LeafletMap.Keyboard', () => {
 		map = new LeafletMap(container, {
 			zoomAnimation: false	// If true, the test has to wait extra 250msec
 		});
+		await map.callInitHooks();
 
 		// make keyboard-caused panning instant to cut down on test running time
 		map.panBy = function (offset) { return LeafletMap.prototype.panBy.call(this, offset, {animate: false}); };
@@ -31,7 +32,6 @@ describe('LeafletMap.Keyboard', () => {
 		// However this can only happen with trusted events (non-synthetic), so
 		// this bit has to be faked.
 		map.keyboard._onFocus();
-		await map.callInitHooks();
 	});
 
 	afterEach(() => {
--- a/spec/suites/map/handler/Map.PinchZoomSpec.js
+++ b/spec/suites/map/handler/Map.PinchZoomSpec.js
@@ -227,7 +227,8 @@ describe('LeafletMap.PinchZoom', () => {
 			.down().moveBy(-200, 0, 500).up();
 	});
 
-	it('Layer is rendered correctly while pinch zoom when zoomAnim is false', (done) => {
+	it('Layer is rendered correctly while pinch zoom when zoomAnim is false', async () => {
+		const {promise, resolve} = Promise.withResolvers();
 		map.remove();
 
 		map = new LeafletMap(container, {
@@ -235,6 +236,7 @@ describe('LeafletMap.PinchZoom', () => {
 			inertia: false,
 			zoomAnimation: false
 		});
+		await map.callInitHooks();
 
 		map.setView([0, 0], 8);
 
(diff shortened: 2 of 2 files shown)


src/core/Class.js
```diff
@@ -74,12 +74,16 @@ export class Class {
...
 	}
 
-	callInitHooks() {
-		if (this._initHooksCalled) {
+	async callInitHooks() {
+		if (!this._initHooksReady) { // _initHooksCalled
 			return;
 		}
+		await this._initHooksReady.promise;
+		this._initHooksReady = undefined;
```
simon04, line 89: Await for Promise to be resolved in the `Class` constructor body.

src/core/Class.js
```diff
@@ -74,12 +74,16 @@ export class Class {
 
 		// call all constructor hooks
 		this.callInitHooks();
+
+		this._initHooksReady.resolve();
```
simon04, line 78: Resolve the Promise to async trigger `callInitHooks`.

simon04:
@jonkoops, I would like to hear your thoughts in this "trick". :scientist:

jonkoops:
@simon04 try adding the [following test](https://github.com/jonkoops/Leaflet/blob/784602179c6eda476d3e076bc8b63ecff4b5cf2f/spec/suites/core/ClassSpec.js#L342-L406) and making it pass with the changes you have here, modifying it to use a default class instead of the custom function. This test should accurately reflect how init hooks should be triggered.

simon04:
> try adding the [following test](https://github.com/jonkoops/Leaflet/blob/784602179c6eda476d3e076bc8b63ecff4b5cf2f/spec/suites/core/ClassSpec.js#L342-L406) and making it pass with the changes you have here

Done. Most tests pass; some might need some additional `await` tuning...

jonkoops:
The main thing is that the initialization hooks should execute directly after a class has been constructed, the user should not have to manually call methods, or await anything themselves.

Falke-Design:
@reblace maybe you have some input for this issue too #9906

Goal is to use `constructor` instead of `initialize` but keep `callInitHooks`

reblace:
Chiming in since Falke-Design tagged us — we've been looking at this from the tree-shaking angle and have some relevant context on the migration effort, though the core team will have a much better sense of the plugin ecosystem than we do.

**Our read: the migration cost is real but bounded, and in my opinion healthy plugins will manage it.**

The dominant `initialize()` pattern in the wild looks straightforward to migrate:

```js
// Old
L.SomeLayer = L.Layer.extend({
  initialize: function(options) {
    L.Util.setOptions(this, options);
    this._thing = computeSomething();
  }
});

// New
class SomeLayer extends Layer {
  constructor(options) {
    super(options);
    this._thing = computeSomething();
  }
}
```

That's a mechanical rename. The cases that look harder — plugins that call `ParentClass.prototype.initialize.call(this, args)` — I suspect actually become *simpler* with native constructors (`super(args)`), not harder.

The genuine breakage is plugins that mix `Class.extend()` with direct prototype calls into a parent's `initialize`. In my opinion those plugins already need a larger rethink to adopt native `class` syntax anyway, so `initialize()` removal isn't the marginal cost — it's part of the same migration.

On the ecosystem concern: the plugin landscape appears to be a mix of actively maintained projects and ones that haven't seen a commit in years. The actively maintained ones will likely follow Leaflet wherever it goes; the dormant ones probably won't be updated regardless of how gentle the migration path is. My guess is a decision like this is unlikely to change that calculus either way — but you all know the ecosystem far better than we do.

One small suggestion that might ease the transition: logging a deprecation warning (not an error) when `initialize()` is detected in the base `Class` constructor during the alpha period. That would give active plugin authors an automatic signal without breaking anything prematurely.

jonkoops:
Closing this PR, as I don't think using async will solve our problem here, as it comes with its own caveats we have discussed here.

@Falke-Design @reblace could I ask you to move your comment to #9889? I'd like to keep all discussion related to this migration on the issue, to avoid information getting fragmented all over the place.

PR #10045, opened 2026-01-04
Mark leaflet-gpx as compatible with v2.x

mpetazzoni:
The Leaflet GPX plugin (in its `main` branch) is now compatible with v2.x, verified against the latest Leaflet alpha release.


jonkoops:
Hi @mpetazzoni, we're holding off on merging further PRs with 2.0 support for plugins, as we are likely making some more API changes while in alpha. See https://github.com/Leaflet/Leaflet/issues/9842#issuecomment-3302673291. Once we're in a beta/release candidate stage, I think we can commit to landing these again.

PR #10051, opened 2026-01-05
Consistently call super class `initialize()` method

jonkoops:
As part of landing standard JavaScript class constructors, we must consistently call the constructor of the super class. This is currently not the case with the existing `initialize()` method, so this PR adds these changes first, before migrating fully over to standard constructors. This is to ensure the size of the PR doesn't get too big and its contents remain reviewable.

### DIFF
--- a/src/control/Control.Attribution.js
+++ b/src/control/Control.Attribution.js
@@ -1,7 +1,6 @@
 
 import {Control} from './Control.js';
 import {LeafletMap} from '../map/Map.js';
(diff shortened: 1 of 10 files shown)


jonkoops:
@Leaflet/core @Leaflet/leaflet-maintainers could I get a review for this PR, so I can land the rest of the refactoring required for 2.x?

jonkoops:
Thanks for the review, Ivan!

PR #10053, opened 2026-01-05
Fixing infinite zoom issue with flyTo() with specific bounds values

itslevo:
Issue is adequately demonstrated by the following snippet: https://plnkr.co/edit/BgxmmrJcHynLivtV?preview

Seemingly benign coordinate and zoom values result in a wild camera trajectory, followed by an imminent `NaN` value crash.

While I'm not entirely sure of all the underlying calculus, the issue basically comes down to the value returned by the `r(1)` function call produced by taking a `log()` of the `Math.sqrt(b * b + 1) - b` equation. The chosen constant of 0.000000001 does not, unfortunately, cut off the entire range of values that result in `Math.log(sq)` returning values < -18. As far as I can tell, values of `b` that cause trouble begin somewhere around `b = 29826170` (see graph below)

Screenshot 2026-01-05 at 17 17 24

Putting 29826170 into the equation results in a value of 1.4901161193847656e-8, or 0.000000014901161193847656. The newly chosen constant value of 0.000000015 therefore fulfils the condition of `sq < 0.000000015`.

Need recommendation for a better test case, as in case of the fail condition, the test merely starts throwing the `NaN` error. Don't know if this is the best possible way to test this. Thank you.

### DIFF
--- a/spec/suites/map/MapSpec.js
+++ b/spec/suites/map/MapSpec.js
@@ -1459,7 +1459,7 @@ describe('Map', () => {
 		});
 
 		it('should handle parameters leading to Math.log(sq) issue', function (done) {
-			container.style.width =
+			container.style.width = '1024px';
 			container.style.height = '1024px';
 			container.style.visibility = 'visible';
 


IvanSanchez:
@itslevo The linter is complaining about a missing semicolon, please check that.

spec/suites/map/MapSpec.js
```diff
@@ -1457,6 +1457,29 @@ describe('Map', () => {
 
 			map.flyTo(newCenter, 22, {animate: true, duration: 0.1});
 		});
+
+		it('should handle parameters leading to Math.log(sq) issue', function (done) {
+			container.style.width =
```
jonkoops: Why is the right-hand side of this assignment not populated? If the value is the same as the height, it is fine to assign it the same value explicitly.

itslevo: I believe the assignment happens as `container.style.width = container.style.height = '1024px';` but I will write it out more explicitly.

jonkoops: Thanks!

jonkoops:
I am merging this in, thanks for the fix @itslevo! If you'd like to land your work in the 1.x series of Leaflet as well, please backport your work by opening a PR to the [`v1`](https://github.com/Leaflet/Leaflet/tree/v1) branch.

Issue #10047, opened 2026-01-05
Stop exposing modules on other classes


jonkoops:
Currently, a lot of classes exported by Leaflet are attached to other classes (e.g. `LeafletMap.Drag`, `Control.Layers`, etc.). This means that they are not tree-shakable (see #9814), and our modules have side effects, making it hard to eliminate dead code.

As part of this issue, all classes exported by Leaflet must not have other classes attached to them, and these classes must be individually importable. For example, instead of exporting `Control.Layers`, the class should be exported as `LayersControl`.

mourner:
Definitely worth doing. The current naming is a relic of namespacing classes in pre-ES era, no longer relevant.

jonkoops:
@simon04 I think all tasks for this issue have been accomplished, right? Or do we still have places left where we are modifying the exports?

simon04:
I think, we're done. Thanks for your work and your reviews!

jonkoops:
Awesome, then I am closing this one. Thanks a lot for the help!

PR #10056, opened 2026-01-05
TileLayer: stop exposing modules on other classes

simon04:

### DIFF
--- a/docs/examples/wms/index.md
+++ b/docs/examples/wms/index.md
@@ -25,7 +25,7 @@ When somebody publishes a WMS service, most likely they link to something called
 
 	http://ows.mundialis.de/services/service?request=GetCapabilities
 
-Leaflet does not understand WMS `GetCapabilities` documents. Instead, you have to create a `L.TileLayer.WMS` layer, provide the base WMS URL, and specify whatever WMS options you need.
+Leaflet does not understand WMS `GetCapabilities` documents. Instead, you have to create a `L.TileLayerWMS` layer, provide the base WMS URL, and specify whatever WMS options you need.
 
 The base WMS URL is simply the `GetCapabilities` URL, without any parameters, like so:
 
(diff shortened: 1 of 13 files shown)


docs/examples/wms/index.md
```diff
@@ -25,7 +25,7 @@ When somebody publishes a WMS service, most likely they link to something called
 
 	http://ows.mundialis.de/services/service?request=GetCapabilities
 
-Leaflet does not understand WMS `GetCapabilities` documents. Instead, you have to create a `L.TileLayer.WMS` layer, provide the base WMS URL, and specify whatever WMS options you need.
+Leaflet does not understand WMS `GetCapabilities` documents. Instead, you have to create a `L.TileLayerWMS` layer, provide the base WMS URL, and specify whatever WMS options you need.
```
mourner, line 28: Just wondering, should we replace all the instances of `L.` in the docs for the relevant classes while we're at it? Since we don't use it in the code, it probably doesn't make sense referencing it this way too.

```suggestion
Leaflet does not understand WMS `GetCapabilities` documents. Instead, you have to create a `TileLayerWMS` layer, provide the base WMS URL, and specify whatever WMS options you need.
```

mourner (approved):

docs/examples/wms/index.md
simon04: Fine by me, but there seem to plenty of occurrences of `L...` in all docs/examples, and thus be better handled by a separate PR?

mourner: Yeah, separate's fine too

simon04: I'm happy to provide a follow-up PR once this series of PRs is merged. :)

jonkoops: Yes, let us handle this as a separate PR. Contribution is welcome :)

jonkoops (approved):
LGTM, thanks again!

PR #10058, opened 2026-01-05
Handler: stop exposing modules on other classes

simon04:


simon04:
I've rebased in order to resolve merge conflicts.

spec/suites/map/handler/TapHoldHandlerSpec.js
```diff
@@ -4,7 +4,7 @@ import sinon from 'sinon';
 import UIEventSimulator from 'ui-event-simulator';
 import {createContainer, removeMapContainer} from '../../SpecHelper.js';
 
-describe('LeafletMap.TapHoldSpec.js', () => {
+describe('TapHoldSpecHandler', () => {
```
jonkoops: We can omit the 'spec' part here, like with the other specs.

```suggestion
describe('TapHoldHandler', () => {
```

jonkoops:
Thanks again @simon04, merging this in.

PR #10059, opened 2026-01-05
Icon: stop exposing modules on other classes

simon04:


Falke-Design (approved):

simon04:
> Can we change the name to `DefaultIcon`? 

Done. :smile:

PR #10063, opened 2026-01-07
Rename WMSTileLayer for consistency

simon04:
Follow-up for https://github.com/Leaflet/Leaflet/pull/10056.

### DIFF
--- a/build/leafdoc-templates/html.hbs
+++ b/build/leafdoc-templates/html.hbs
@@ -42,7 +42,7 @@ bodyclass: api-page
 		<h4>Raster Layers</h4>
 		<ul>
 			<li><a href="#tilelayer">TileLayer</a></li>
-			<li><a href="#tilelayer-wms">TileLayerWMS</a></li>
+			<li><a href="#wmstilelayer">WMSTileLayer</a></li>
 			<li><a href="#imageoverlay">ImageOverlay</a></li>
 			<li><a href="#videooverlay">VideoOverlay</a></li>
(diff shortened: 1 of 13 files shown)


build/leafdoc-templates/html.hbs
```diff
@@ -42,7 +42,7 @@ bodyclass: api-page
 		<h4>Raster Layers</h4>
 		<ul>
 			<li><a href="#tilelayer">TileLayer</a></li>
-			<li><a href="#tilelayer-wms">TileLayerWMS</a></li>
```
jonkoops, line 45: Do we need to update the `href` here? I still see references to `#tilelayer-wms` in one of the examples, and I am sure there will be external links referencing it as well. If we need to update the `href` because the title has changed, let's also update that example.

src/layer/tile/WMSTileLayer.js
jonkoops, line 1: Looks like this file has some unrelated line-ending changes, could you revert those?

build/leafdoc-templates/html.hbs
IvanSanchez: Yes, we do. The anchor names are derived from the class names (in lowercase).

jonkoops: Gotcha, thanks!

src/layer/tile/WMSTileLayer.js
simon04: I've rebased onto main.

jonkoops (approved):

jonkoops:
Thanks @simon04!

PR #10068, opened 2026-01-10
Remove emptyImageUrl and asscociated logic

### LINKED ISSUE #9363: Change setup around invisible pixel
### Checklist

### Motivation

I am working on a project where we are targeting a strict content security policy and we are having some issues with Leaflet. The issue is described here on a now closed issue:
https://github.com/Leaflet/Leaflet/issues/6113

In short - Leaflet creates empty pixels using 'data:', which we can not allow in our web applications with our strict CSP.

A solution was proposed in the closed issue, but it does not seem to work anymore. It would be awesome to have a solution where the invisible pixel is not set using 'data:'.

### Suggested solution

Would be great if the Leaflet package declares the empty pixel as a .gif file instead.

### Alternatives considered

_No response_

### LINKED ISSUE #10050: Remove `emptyImageUrl` and asscociated logic
The `emptyImageUrl` property was added to deal with old mobile WebKit bugs, and is no longer required. It should be eliminated, as well as all the associated logic that comes with it. See the linked discussion for more information.

### Discussed in https://github.com/Leaflet/Leaflet/discussions/10048

<div type='discussions-op-text'>

<sup>Originally posted by **williwyg** January  5, 2026</sup>
Could you please consider changing the emptyImageUrl to "//:0". This will serve the same purpose as the current setting but will allow a strict CSP header to be used with img-src 'self' </div>

SerAbin1:
Remove legacy emptyImageUrl workaround

Closes #10050 
Closes #9363

### DIFF
--- a/src/layer/tile/TileLayer.js
+++ b/src/layer/tile/TileLayer.js
@@ -252,7 +252,7 @@ export class TileLayer extends GridLayer {
 				tile.onerror = Util.falseFn;
 
 				if (!tile.complete) {
-					tile.src = Util.emptyImageUrl;
+					tile.setAttribute('src', '');
 					const coords = this._tiles[i].coords;
 					tile.remove();
 					delete this._tiles[i];
@@ -272,13 +272,13 @@ export class TileLayer extends GridLayer {
 		if (!tile) { return; }
 
 		// Cancels any pending http requests associated with the tile
-		tile.el.setAttribute('src', Util.emptyImageUrl);
+		tile.el.setAttribute('src', '');
 
 		return super._removeTile(key);
 	}
 
 	_tileReady(coords, err, tile) {
-		if (!this._map || (tile && tile.getAttribute('src') === Util.emptyImageUrl)) {
+		if (!this._map || (tile && tile.getAttribute('src') === '')) {
 			return;
 		}
 
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -117,8 +117,4 @@ export function template(str, data) {
 	});
 }
 
-// @property emptyImageUrl: String
-// Data URI string containing a base64-encoded empty GIF image.
-// Used as a hack to free memory from unused images on WebKit-powered
-// mobile devices (by setting image `src` to this string).
-export const emptyImageUrl = 'data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=';
+


docs/reference.html
jonkoops, line 1: Can you remove these changes, we try not to update the documentation in PRs like this.

src/layer/tile/TileLayer.js
```diff
@@ -252,7 +252,6 @@ export class TileLayer extends GridLayer {
 				tile.onerror = Util.falseFn;
 
 				if (!tile.complete) {
-					tile.src = Util.emptyImageUrl;
```
jonkoops, line 255: It looks like removing an element is not enough for browsers to cancel the request for the image. In order to speed up tile loading on slow connections, we should still cancel the request. I have done a little research, and found out that setting the source to an empty string works consistently across all browsers we support (Chrome, Firefox and Safari).

Please change this line to:
```js
tile.setAttribute('src, '');
```

src/layer/tile/TileLayer.js
```diff
@@ -271,14 +270,11 @@ export class TileLayer extends GridLayer {
 		const tile = this._tiles[key];
 		if (!tile) { return; }
 
-		// Cancels any pending http requests associated with the tile
-		tile.el.setAttribute('src', Util.emptyImageUrl);
-
```
jonkoops, line 276: Please restore the code here as well, and use an empty string as the value.

src/layer/tile/TileLayer.js
```diff
@@ -271,14 +270,11 @@ export class TileLayer extends GridLayer {
...
 		return super._removeTile(key);
 	}
 
 	_tileReady(coords, err, tile) {
-		if (!this._map || (tile && tile.getAttribute('src') === Util.emptyImageUrl)) {
```
jonkoops, line 281: Please restore this code, and check against an empty string source.

src/layer/tile/TileLayer.js
SerAbin1: Okay, thanks. I have updated the PR.

jonkoops:
Thanks again @SerAbin1!

mourner:
@jonkoops I did a little digging in the code where that hack was first introduced, and back then (15 years ago!) it had the following comment:

image

I'm pretty sure that since then all those issues were fixed, but commenting just in case — hopefully Android / iOS now frees memory correctly when setting to an empty string?

jonkoops:
> hopefully Android / iOS now frees memory correctly when setting to an empty string?

I did a little testing on my old Android phone (on Chrome), and I could not see any memory pressure from uncollected images. I have not tested on iOS, but I assume the result is the same. Those old browsers were notorious for being extremely leaky, this was when it was still all built on top of crappy WebKit browsers stuck on specific versions.

PR #10069, opened 2026-01-10
Replace `setDefaultOptions` with `static defaultOptions`

simon04:
An attempt to tackle https://github.com/Leaflet/Leaflet/issues/9814.

Quoting from https://github.com/Leaflet/Leaflet/issues/9814#issuecomment-3730229170

> It seems that the [static init block](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks) of `class WMSTileLayer` `static { this.prototype.defaultWmsParams... this.setDefaultOptions... }` is preventing tree-shaking. Commenting out this block, removes the entire `class WMSTileLayer` from the bundle.

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -11,6 +11,7 @@ export class Class {
 	// @function include(properties: Object): this
 	// [Includes a mixin](#class-includes) into the current class.
 	static include(props) {
+		this._initDefaultOptions();
 		const parentOptions = this.prototype.options;
 		for (const k of getAllMethodNames(props)) {
 			this.prototype[k] = props[k];
@@ -33,16 +34,10 @@ export class Class {
 		}
 	}
 
-	// @function setDefaultOptions(options: Object): this
-	// Configures the [default `options`](#class-options) on the prototype of this class.
-	static setDefaultOptions(options) {
-		Util.setOptions(this.prototype, options);
-		return this;
-	}
-
 	// @function mergeOptions(options: Object): this
 	// [Merges `options`](#class-options) into the defaults of the class.
 	static mergeOptions(options) {
+		this._initDefaultOptions();
 		this.prototype.options ??= {};
 		Object.assign(this.prototype.options, options);
 		return this;
@@ -62,9 +57,21 @@ export class Class {
 		return this;
 	}
 
+	static _initDefaultOptions(proto = this.prototype) {
+		if (!proto) { return; }
+		Class._initDefaultOptions(Object.getPrototypeOf(proto));
+		const options = proto.constructor.defaultOptions;
+		if (!options) { return; }
+		if (Object.hasOwn(proto, 'options')) { return; }
+		Util.setOptions(proto, options);
+	}
+
 	constructor(...args) {
 		this._initHooksCalled = false;
 
+		for (const proto of this._prototypes) {
+			Class._initDefaultOptions(proto);
+		}
 		Util.setOptions(this);
 
 		// call the constructor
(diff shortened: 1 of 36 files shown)


IvanSanchez:
Doesn't this technique overwrite the parent class' default options? I'm wondering if the code should rather look like...

```
class WMSTileLayer {
  static defaultOptions {
    crs: null,
    uppercase: false,
    ...super.constructor.defaultOptions
  }

  // etc
```

(or something like that)

IvanSanchez:
Oh, wait, that's what `getDefaultOptions()` does. Nevermind.

simon04:
The problem however is that `static defaultOptions` does not mix well with the former `prototype.options`.

Is is the reason why the not-yet-converted Marker/Icon/DivIcon tests fail: `Marker` specifies `pane: 'markerPane'` via `static { this.setDefaultOptions(...) }`. In `Util.setOptions`, first the `prototype.options` are taken into account (via `Object.create`), and afterwards all `static defaultOptions` are set (possibly overwriting the former).

simon04:
> does not mix well with the former `prototype.options`.

Neither did it in Leaflet 1.9 since directly modifying `prototype.options` was never encouraged and the canonical way was to use `Class.extend({options:...})`.

So here's the latest draft:
- `Class.setDefaultOptions` is removed
- `static defaultOptions` is used in the entire code base
- `Class._initDefaultOptions` is called by the constructor, by `Class.include`, and `Class.mergeOptions` and copies `static defaultOptions` to `prototype.options` -- so the observed behaviour of `prototype.options` is as it was before

src/core/Class.js
```diff
@@ -62,9 +57,19 @@ export class Class {
...
+		if (Object.hasOwn(proto, 'options')) { return; }
+		const options = proto.constructor.defaultOptions;
+		if (!options) { return; }
+		Util.setOptions(proto, options);
+	}
+
 	constructor(...args) {
 		this._initHooksCalled = false;
 
+		Class._initDefaultOptions(Object.getPrototypeOf(this));
```
IvanSanchez, line 72: Looping through the prototype chain *every time* an object is instantiated looks like a potential performance issue to me.

There should be a way to cache the default options for a class without calculating them every time. Perhaps the `static defaultOptions = {...super.constructor.defaultOptions}` technique would be the right way for this? As a `static`, it's evaluated only once per class.

IvanSanchez (changes requested):

src/core/Class.js
simon04: > There should be a way to cache the default options for a class without calculating them every time.

It more or less already is via the following line in `_initDefaultOptions`

```
if (Object.hasOwn(proto, 'options')) { return; }
```

Possibly this line could even be moved one line up before the recursive call.

> Perhaps the `static defaultOptions = {...super.constructor.defaultOptions}` technique

This technique should work, however it makes specifying default options a bit cumbersome/fragile (in particular for plugins). Omitting `super.constructor.defaultOptions` once will break the inheritance chain.

jonkoops:
I'll have to take an in-depth look at this before merging, so marking this as a draft for now. In general, I want to get the API for `Class` sorted out soon, including some changes I am working on that relate to adopting the standard constructor. I'll unfortunately be quite busy at work starting next week, so I am going to be putting in nights and weekends.

simon04:
Here's a benchmark using `hyperfine`

```js
// bench.js
import {Marker, DivIcon} from './src/Leaflet.js';

for (let i = 0; i < 9_000_000; i++) {
	new Marker([12, 12], {icon: new DivIcon()});
}
```

```
> bun --version
1.3.5

> hyperfine -p 'git checkout main' 'bun bench.js' -p 'git checkout static-defaultOptions' 'bun bench.js'
Benchmark 1: bun bench.js // v2.0.0-alpha.1-114-g6975c419
  Time (mean ± σ):      1.851 s ±  0.048 s    [User: 1.920 s, System: 0.122 s]
  Range (min … max):    1.763 s …  1.926 s    10 runs

Benchmark 2: bun bench.js // v2.0.0-alpha.1-119-gd73c3fdd
  Time (mean ± σ):      1.977 s ±  0.047 s    [User: 2.038 s, System: 0.132 s]
  Range (min … max):    1.906 s …  2.080 s    10 runs

Summary
  bun bench.js ran
    1.07 ± 0.04 times faster than bun bench.js
```

Overall, callInitHooks (1296ms) is more expensive than setOptions (499ms) on Marker+DivIcon combined when profiling in Firefox 146 (total runtime 3150ms).

PR #10070, opened 2026-01-11
Add test for Canvas#_containsPoint()

antoniotp2:
This PR adds test coverage for Canvas `_containsPoint`, including points inside and outside polygons.

Fixes https://github.com/Leaflet/Leaflet/issues/5165

### DIFF
--- a/spec/suites/layer/vector/CanvasSpec.js
+++ b/spec/suites/layer/vector/CanvasSpec.js
@@ -297,7 +297,8 @@ describe('Canvas', () => {
 			requestAnimationFrame(() => { done(); });
 		});
 	});
-	describe('Canvas _containsPoint', () => {
+	
+	describe('#_containsPoint', () => {
 		it('detects point inside polygon', () => {
 			const polygon = new Polygon([
 				[0, 0],


IvanSanchez:
@antoniotp2 Please check the linter errors, it seems to be complaining about indentation.

antoniotp2:
I have fix the indentation issue

spec/suites/layer/vector/CanvasSpec.js
```diff
@@ -297,4 +297,56 @@ describe('Canvas', () => {
 			requestAnimationFrame(() => { done(); });
 		});
 	});
+	describe('Canvas _containsPoint', () => {
```
jonkoops: Couple of things:

1. Let's add a newline before the `describe()` block, like the other ones.
2. No need to add 'Canvas' before here, as this is already wrapped in a `describe('Canvas')` block (I see this was copied from the test above, it can be removed there as well).
3. The common convention for methods seems to be a `#` prefix, so the name should be `#_containsPoint` to be consistent.

jonkoops:
@antoniotp2 for future reference, please use a descriptive name for your pull request, e.g. 'Add test for `Canvas#_containsPoint()`', instead of a generic 'Update X'. That makes it easier to understand in the future what the commit was about at a glance.

Please also use the rebase strategy for keeping your PR up-to-date, to avoid merge commits, and keep the history clean.

Thanks!

antoniotp2:
Thanks a lot for the quick feedback and your time. I have fixed the formatting and indentation
and i updated the describe block naming to follow the conventions.

IvanSanchez:
@antoniotp2 Unfortunately the linter complains, and the tests fail.

Please update/sync your fork, do a `git pull`, and run `npm run lint` and `npm run test` locally. You should be able to see some errors there.

Falke-Design:
Never worked, because `renderer._containsPoint` is not a function. Should be `polygon._containsPoint`

Issue #10046, opened 2026-01-05
Investigate and possibly resolve code duplication in `Circle`


jonkoops:
Currently, `Circle` extends from `CircleMarker`, and overrides certain methods to use a custom `_mRadius` property for the radius. However, it seems that the logic is largely comparable to the super-class, which uses its own `_radius` property. In fact, both these properties are used within the `Circle` class, somewhat interchangeably.

As part of this issue, we need to determine if the logic in `Circle` is actually the same as `CircleMarker` and we can unify the `_mRadius` and `_radius` properties, and if so, unify them into a single property, eliminating redundant methods in the process (e.g. `getRadius()`, `setRadius()`, `setStyle()`, etc.).

jonkoops:
@mourner @IvanSanchez perhaps you know some of the history around this, and why the implementation is the way it is?

mourner:
@jonkoops looking at the old commits, I think it split back when both classes were extremely small and simple — here https://github.com/Leaflet/Leaflet/commit/27c6aba9328b8d00f0d10b0c931b7d7c9d314184, and the separate properties were because `_radius` dictated the actual pixel radius the vector object rendered with, and `_mRadius` was radius in meters which in turn determined the value `_radius` (which changed on every zoom or with latitude change). At the time it was just easier that way — see how `CircleMarker` was just a few lines of code that made `_radius` static, nothing else. 

Now, after all the changes both classes went through in years, if you can find a way to refactor / simplify this, I think it's worth doing — the only gotcha to be aware of is if any plugins depend on the private propeties for custom logic, which unfortunately happens sometimes.

jonkoops:
Thanks for the insights, this is certainly something we can look into once we get 2.x out the door then.

> the only gotcha to be aware of is if any plugins depend on the private propeties for custom logic, which unfortunately happens sometimes.

This is actually something that I would like to be more explicit about going forward with 2.x, I would consider any of these properties (or anything underscored) as not part of our commitment to a public API going forward. Meaning that if plugins, or users rely on them, there is no guarantee they will continue to exist.

We could also start implementing true [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements), which would make this an explicit contract. But that would mean a rather large refactor, that I would rather not get into right now.

@Leaflet/core WDYT?

tomasmusil:
[...]

PR #10072, opened 2026-01-11
Deduplicate code in `Circle`

tomasmusil:
Attempts to deduplicate code in `Circle` by renaming `_mRadius` to `_radius` and creating new property `_pxRadius` used both in `Circle` and `CircleMarker` (see #10046).

### DIFF
--- a/src/layer/vector/CircleMarker.js
+++ b/src/layer/vector/CircleMarker.js
@@ -8,7 +8,7 @@ import {Bounds} from '../../geometry/Bounds.js';
  * @class CircleMarker
  * @inherits Path
  *
- * A circle of a fixed size with radius specified in pixels. Extends `Path`.
+ * A circle of a fixed size with radius specified in CSS pixels. Extends `Path`.
  */
 
 // @constructor CircleMarker(latlng: LatLng, options?: CircleMarker options)
@@ -22,7 +22,7 @@ export class CircleMarker extends Path {
 			fill: true,
 
 			// @option radius: Number = 10
-			// Radius of the circle marker
+			// Radius of the circle marker, in CSS pixels
 			radius: 10
 		});
 	}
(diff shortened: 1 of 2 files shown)


spec/suites/layer/vector/CircleSpec.js
```diff
@@ -55,7 +55,77 @@ describe('Circle', () => {
...
 			const circle = new Circle([0, 0], {radius: 200}).addTo(map);
 
-			expect(circle._radius).to.eql(3200);
+			expect(circle._radius).to.eql(200);
+			expect(circle._pxRadius).to.eql(3200);
+		});
+	});
+});
+
+describe('Circle', () => {
```
jonkoops: Let's not duplicate `describe()` blocks with the same name, can you move these tests into the block that already exists?

jonkoops:
@IvanSanchez if you have the time, could I ask you for a pass on this as well?

spec/suites/layer/vector/CircleSpec.js
musilto9: Sure, I did that because of the custom `beforeEach`, should I move it into the `describe('#_radius', ...)` or rename the top-level `describe()`?

jonkoops: If the logic differs between the `beforeEach()`  perhaps we can move some of that setup work into each individual test where it differs? If this is too much of a hassle, don't worry, this is more of a nit.

src/layer/vector/CircleMarker.js
```diff
@@ -22,7 +22,7 @@ export class CircleMarker extends Path {
 			fill: true,
 
 			// @option radius: Number = 10
-			// Radius of the circle marker, in pixels
```
IvanSanchez, line 25: I'd leave the docstrings as they were (or even change them to "CSS pixels" as opposed to "device pixels" for hi-dpi displays).

tomasmusil: I removed the units here and in `setRadius()` as to not confuse the reader in docs where in `Circle` options/methods inherited from `CircleMarker` it would say that radius is in pixels even though it is actually in meters. The actual correct units are in description of the whole class.

I can revert it if you don't feel like that's an issue, alternatively I could leave the original docstrings in `Circle` even though they wouldn't actually document any code and be there only to override the inherited ones.

What do you think?

IvanSanchez: I think it's better to have docstrings in both `Circle` and `CircleMarker` (even if there's no matching JS code). That will prevent the "inherited from" documentation from appearing, and we can have "meters" in one place and "CSS pixels" in another.

Otherwise it's confusing (from my point of view).

tomasmusil: Understand, I reverted the docstrings.

Falke-Design (approved):

Falke-Design:

PR #10074, opened 2026-01-14
Fix scale overlapping with attributions on mobile devices

theAnuragMishra:
This pr fixes https://github.com/openstreetmap/openstreetmap-website/issues/6705 by shifting the scale up by `34px` on mobile devices (width less than `520px`).

### DIFF
--- a/src/leaflet.css
+++ b/src/leaflet.css
@@ -192,6 +192,14 @@
 	}
 }
 
+@media screen and (max-width: 520px){
+	.leaflet-bottom {
+		.leaflet-left{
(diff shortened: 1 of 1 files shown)


IvanSanchez:
This approach won't work for the general case. It doesn't take into account cases where there's no attribution control visible, or the attribution control is bigger than 34px high.

It'd be worth to look into CSS grids, though.

theAnuragMishra:
Oh okay. I gave it a look and didn't find the attribution control height changing. A better solution will require changing a lot of css given how it is currently designed.

PR #10087, opened 2026-01-30
Lower polyline smoothFactor from 1.0 to 0.5

### LINKED ISSUE #10086: Trying to understand why small gaps are appearing between shapefile boundaries in my map
I am creating a choropleth map of US states using Leaflet 1.9.4. My state boundaries are based on state-level shapefiles from the US Census Bureau that I then simplified using the [`simplify_coverage()`](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.simplify_coverage.html#geopandas.GeoSeries.simplify_coverage) method within GeoPandas in Python. (This method helps prevent gaps from appearing in between regions.)

When I zoom in on my Leaflet map, boundaries (like the Illinois-Wisconsin border in the following image) appear as expected:

Image

However, at lower zoom levels, I sometimes see small gaps emerge between boundaries. Notice the white sliver between Illinois and Wisconsin within the following screenshot:

Image

I'm not noticing this behavior within an alternative set of maps that I created in Plotly.py, so I'm thinking that this issue might be Leaflet-related. I'm still relatively new to both Leaflet and JavaScript, so any explanation of what might be going wrong here--and whether there's a way to prevent this behavior--would be greatly appreciated. Thank you!

IvanSanchez:
Should fix #10086. cc @kburchfiel

I wonder if `window.devicePixelRatio ` should be taken into account for this. Do slivers appear at hi-dpi screens?

### DIFF
--- a/src/layer/vector/Polyline.js
+++ b/src/layer/vector/Polyline.js
@@ -54,10 +54,11 @@ export class Polyline extends Path {
 		// @section
 		// @aka Polyline options
 		this.setDefaultOptions({
-			// @option smoothFactor: Number = 1.0
-			// How much to simplify the polyline on each zoom level. More means
-			// better performance and smoother look, and less means more accurate representation.
-			smoothFactor: 1.0,
+			// @option smoothFactor: Number = 0.5
+			// How much to simplify the polyline on each zoom level (Douglas–Peucker
+			// epsilon, in pixels). Higher values mean faster rendering but less
+			// accuracy; lower values mean more accuracy but slower rendering.
+			smoothFactor: 0.5,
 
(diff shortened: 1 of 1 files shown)


mourner:
The new default sounds reasonable to me — it does seem to fix the given example. Simplification will remove slightly less points but that should be OK — it's been almost 15 years since the previous default, devices and browsers were a whole lot weaker then.

> I wonder if `window.devicePixelRatio ` should be taken into account for this. Do slivers appear at hi-dpi screens?

Not sure but intuitively I'd say not — hi-DPI should affect the relative scale of the slivers but not whether they appear or not (that's purely algorithmic).

PR #10100, opened 2026-02-08
feat: setup vitepress documentation

abdul-alhasany:
This is the first PR to migrate documentation into VitePress. 

This the vitepress setup. To run, `pnpm install` then `pnpm docs:dev`

#9980 #10099


Falke-Design:
Saddly not the right time - thank you anyways!

See #10099

PR #10108, opened 2026-02-12
Fix: Correct spelling of suppressEvent parameter

pierreeurope:
Fixed typo in `_move` method parameter name from `supressEvent` to `suppressEvent` (suppress has double 'p').

**Changes:**
- Renamed parameter from `supressEvent` to `suppressEvent`
- Updated the conditional check to use the corrected parameter name

This is a private method so there are no breaking changes.


pierreeurope:
Duplicate of #10105

PR #10120, opened 2026-02-12
docs: update mobile example to use viewport units

### LINKED ISSUE #7342: Mobile Example Outdated

pierreeurope:
Update the mobile tutorial example to use `100vh` instead of `100%` for the map height.

The Chrome mobile bug that previously prevented using `vh` units was fixed years ago. Using `100vh` is more reliable for fullscreen maps as it doesn't depend on parent elements having explicit heights.

Also removes the parenthetical note about the workaround from the tutorial text.

Fixes #7342.

### DIFF
--- a/docs/examples/mobile/example.md
+++ b/docs/examples/mobile/example.md
@@ -6,7 +6,7 @@ css: "body {
 		margin: 0;
 	}
 	#map {
-		height: 100%;
+		height: 100vh;
 		width: 100vw;
 	}"
 ---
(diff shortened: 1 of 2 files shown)


jonkoops:
@pierreeurope I'll try to get to your PRs in due time, but we are currently focusing on a Leaflet 2.x release. It will likely be after that we have bandwith.

pierreeurope:
No worries :) good luck!

PR #10124, opened 2026-02-12
docs: add missing JSDoc annotations for public methods

pierreeurope:
Adds `@method` JSDoc annotations for public methods that previously lacked proper API documentation.

**Methods documented:**

### Layer.js
- `addInteractiveTarget(targetEl)` — registers a DOM element as an interactive target
- `removeInteractiveTarget(targetEl)` — unregisters an interactive target

These methods are used by custom layer implementations to handle events on layer elements.

### Marker.js
- `update()` — updates marker position based on current coordinates

Useful when programmatically changing marker coordinates.

### Tooltip.js
- `setOpacity(opacity)` — sets the tooltip opacity

Public API method for controlling tooltip visibility.

### Circle.js
- `setStyle(options)` — overridden to also handle radius updates

Important to document this override behavior so developers know `radius` can be set via `setStyle()`.

---

These methods were all functional but lacked proper JSDoc annotations, making them difficult to discover in the API documentation. This PR improves API discoverability without changing any functionality.

### DIFF
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -78,11 +78,16 @@ export class Layer extends Evented {
 		return this._map.getPane(name ? (this.options[name] || name) : this.options.pane);
 	}
 
+	// @method addInteractiveTarget(targetEl: HTMLElement): this
+	// Registers a DOM element as an interactive target for this layer.
+	// Used internally to handle events on layer elements.
 	addInteractiveTarget(targetEl) {
 		this._map._targets[Util.stamp(targetEl)] = this;
 		return this;
 	}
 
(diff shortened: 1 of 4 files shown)


IvanSanchez:
I see you opened 22 separate PRs in the span of 80 minutes. 

I gotta ask: @pierreeurope , are you a human?

pierreeurope:
Hi ! Yes I'm a human (as far as I can tell) but I am indeed using Claude code to look at some issues on the open source project that I've used in the past (we were building maphub.co) so it's definitely going much faster than I would by hand, but so far on most projects it has become really good at solving small bugs here and there

pierreeurope:
Closing this out, sorry for the PR spam.

PR #10133, opened 2026-02-20
Listen to pointer move and up events from containing document

### LINKED ISSUE #9218: When clicking to drag a map that has been opened in a popout window, the dragging events will only resolve if the mouse pointer is over/in the parent window
### Checklist

### Steps to reproduce

1. Create a page that, instead of embedding the map directly into the page, allows the user to open a window with a map in it.
2. Try to move the map around by clicking + dragging.

### Expected behavior

When clicking and dragging in the child window, the map moves. When releasing the mouse in the child window, the drag behaviouor is ended.

### Current behavior

When clicking and dragging in the child window nothing changes. After initiating drag, when movin the mouse in/over the parent window, the map in the child window starts moving. The drag event is only ended after refocusing the parent window.

https://plnkr.co/edit/ZtQYoVNQI6V6CBQl

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Edge 120.0.2210.133
- OS/Platform (with version): Windows 10 Enterprise 19045.3693

markmeeus:
This change attaches the required mouse event handlers for the Draggable module.
When leaflet is rendered into a dom element which is not in the current document, like a window opened with window.open, it should listen to the events on the document which contains the dom element, not the one in which the code is running.

This Fixes #9218


src/dom/Draggable.js
```diff
@@ -106,7 +106,7 @@ export class Draggable extends Evented {
 		// Cache the scale, so that we can continuously compensate for it during drag (_onMove).
 		this._parentScale = DomUtil.getScale(sizedParent);
 
-		DomEvent.on(document, 'pointermove', this._onMove, this);
+		DomEvent.on(this._element.ownerDocument, 'pointermove', this._onMove, this);
 		DomEvent.on(document, 'pointerup pointercancel', this._onUp, this);
```
IvanSanchez: You seem to have left out some event handlers. Shouldn't the `ownerDocument` logic affect all of them?

markmeeus: Indeed, my bad.
Fixed it now

src/dom/Draggable.js
IvanSanchez: I'm fairly sure this should be changed to `ownerDocument` as well.

IvanSanchez:
I'm re-reading this code, and it seems to me like it implements a IE9/IE10-compatible way of performing pointer capture (see https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture ).

@jonkoops @Falke-Design Given that we've deprecating all support for old IE, should we implement pointer capture instead?

src/dom/Draggable.js
markmeeus: Yes, you are totally right. This causes the cursor to change.

jonkoops:
@IvanSanchez My stance is that we can use any API that is 'widely available' under Baseline without any sort of fallbacks, which this method is. So if pointer capture solves the same problem in a more elegant manner, I think we should go for it.

PR #10179, opened 2026-04-09
Replace static {} blocks with static defaultOptions for tree-shaking support

### LINKED ISSUE #9814: Treeshaking not working

reblace:
Fixes #9814.

## Problem

`static {}` class initializer blocks are treated as unconditional side effects by bundlers. Even with `"sideEffects": false`, Rollup cannot tree-shake individual classes because their static blocks mutate prototypes at module-evaluation time. As simon04 demonstrated in the issue, `WMSTileLayer` appears in bundles even when never imported.

## Solution

**Replace all `static { this.setDefaultOptions({...}) }` blocks** (26 files) with `static defaultOptions = {...}` class fields. Class fields that don't call external functions are recognized as side-effect-free, allowing bundlers to eliminate unused classes.

**Update `Util.setOptions`** to resolve options inheritance at construction time rather than at class-definition time:
- Walks the prototype chain collecting each class's `static defaultOptions`
- Uses a lazy `Object.setPrototypeOf` call to link each class's `defaultOptions` to its parent's — this preserves the live-inheritance semantics of the original `Object.create(prototype.options)` system, so modifications to `SomeClass.defaultOptions` are immediately visible in existing instances that haven't overridden that property

**Add `"sideEffects": false` to `package.json`** to enable module-level tree-shaking for consumers importing from Leaflet source.

**WMSTileLayer**: the `prototype.defaultWmsParams` assignment is moved to a `static defaultWmsParams` class field; `initialize` accesses it via `this.constructor.defaultWmsParams` so subclasses can override it.

## Result

Importing only `{ LeafletMap, TileLayer }` from source produces a ~167 kB bundle vs ~390 kB for the full build. `WMSTileLayer` and other unused classes are absent from the output (verified with `rollup --treeshake`).

## Test changes

Three spec files accessed `SomeClass.prototype.options` directly (which is no longer populated at class-definition time). Updated to use `SomeClass.defaultOptions` — the correct public API going forward. Behaviour is equivalent because the new `setOptions` implementation preserves live inheritance from `defaultOptions`.

All existing tests pass (1 pre-existing failure unrelated to this change: `console.warning is not a function` in a third-party test dependency).

### DIFF
--- a/src/core/Util.js
+++ b/src/core/Util.js
@@ -86,11 +86,48 @@ export function splitWords(str) {
 // Merges the given properties to the `options` of the `obj` object, returning the resulting options. See `Class options`.
 export function setOptions(obj, options) {
 	if (!Object.hasOwn(obj, 'options')) {
-		obj.options = obj.options ? Object.create(obj.options) : {};
+		// Collect static defaultOptions from the prototype chain.
+		// This supports tree-shaking: each class declares only its own defaults as a
+		// static class field (`static defaultOptions = {...}`), and inheritance is
+		// resolved here at construction time rather than at class-definition time.
+		const staticDefaults = [];
+		const protoOptions = [];
+		let proto = obj;
+		while ((proto = Object.getPrototypeOf(proto)) !== null) {
+			const ctor = proto.constructor;
+			if (ctor && Object.hasOwn(ctor, 'defaultOptions')) {
+				staticDefaults.unshift(ctor.defaultOptions);
+			}
+			// Also collect prototype.options set by plugins via mergeOptions/include
+			if (Object.hasOwn(proto, 'options')) {
+				protoOptions.unshift(proto.options);
+			}
+		}
+		if (staticDefaults.length > 0) {
+			// Link each defaultOptions object to its parent via prototype chain (once per pairing).
+			// This creates live inheritance: modifying a class's defaultOptions is immediately
+			// visible to all instances that haven't overridden that property.
+			for (let i = 1; i < staticDefaults.length; i++) {
+				if (Object.getPrototypeOf(staticDefaults[i]) === Object.prototype) {
+					Object.setPrototypeOf(staticDefaults[i], staticDefaults[i - 1]);
+				}
+			}
+			// Instance options inherit directly from the most-derived defaultOptions object
+			obj.options = Object.create(staticDefaults[staticDefaults.length - 1]);
+			// Layer in any plugin-added options (from mergeOptions/include)
+			for (const pluginOpts of protoOptions) {
+				Object.assign(obj.options, pluginOpts);
+			}
+		} else {
+			// Legacy path: options set via setDefaultOptions on the prototype
+			obj.options = obj.options ? Object.create(obj.options) : {};
+		}
 	}
-	for (const i in options) {
-		if (Object.hasOwn(options, i)) {
-			obj.options[i] = options[i];
+	if (options) {
+		for (const i in options) {
+			if (Object.hasOwn(options, i)) {
+				obj.options[i] = options[i];
+			}
 		}
 	}
 	return obj.options;
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "ui-event-simulator": "^2.0.0"
   },
   "type": "module",
+  "sideEffects": false,
   "exports": {
     ".": "./dist/leaflet-src.js",
     "./styles.css": "./dist/leaflet.css"
--- a/spec/suites/layer/PopupSpec.js
+++ b/spec/suites/layer/PopupSpec.js
@@ -247,10 +247,10 @@ describe('Popup', () => {
 		const latlng = center;
 		const offset = new Point(20, 30);
 
-		const autoPanBefore = Popup.prototype.options.autoPan;
-		Popup.prototype.options.autoPan = false;
-		const popupAnchorBefore = DefaultIcon.prototype.options.popupAnchor;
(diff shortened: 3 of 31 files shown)


IvanSanchez:
This is clever, but I don't like the idea of iterating through the prototype chain on object instantiation.

Instead, I'm thinking something along the lines of:

```
class Foo {
  static defaultOptions = {
    fooOpt: 'fo'
  }

  constructor(options={}){
    // Spread operator does the trick of the current Util.setOptions()
    this.options = {
      ...this.constructor.defaultOptions,
      ...options
    };
  }
}

class Bar extends Foo {
  static defaultOptions = {
    barOpt: 'ba',
    ...super.defaultOptions
  }
}

let bar = new Bar({quxOpt: "qu"});
console.log(bar.options);
// outputs { barOpt: "ba", fooOpt: "fo", quxOpt: "qu" }

let bar2 = new Bar({barOpt: "rab"});
console.log(bar2.options);
// outputs { barOpt: "rab", fooOpt: "fo" }
```

reblace:
Thanks for the feedback — your approach is cleaner and I'd be happy to rewrite it that way. Before doing so, I want to make sure we're aligned on one behavioral trade-off.

**Live inheritance of class defaults**

With the `...super.defaultOptions` spread at class-definition time, each subclass gets a flat snapshot of its parent's defaults baked in when the class is loaded. A subsequent mutation like:

```js
Layer.defaultOptions.pane = 'newPane';
```

would not propagate to subclasses (Marker, TileLayer, etc.) because their `defaultOptions` were already spread at definition time.

The current implementation (and our PR's approach) preserves live inheritance — changes to a parent class's defaults are visible in all subclasses and subsequently created instances that haven't overridden that property.

This pattern does appear within the existing test suite (e.g. tests that temporarily patch `Popup.prototype.options.autoPan` or `DefaultIcon.prototype.options.shadowUrl` and then restore them), though it's not clear whether that's intentional public API behavior or just a test implementation detail.

**The core question:** is mutating class-level defaults after load time a pattern Leaflet intends to support going forward? If not, your simpler approach is the right call and we'll rewrite it. If yes, the prototype-chain approach is necessary to preserve that semantics.

Also worth noting: your approach drops `.include()`-based plugin options entirely (the constructor spread never touches `prototype.options`), whereas our approach still collects those at construction time. But that's likely a separate concern from what this PR is trying to solve.

src/core/Util.js
```diff
@@ -86,11 +86,48 @@ export function splitWords(str) {
...
+				}
+			}
+			// Instance options inherit directly from the most-derived defaultOptions object
+			obj.options = Object.create(staticDefaults[staticDefaults.length - 1]);
+			// Layer in any plugin-added options (from mergeOptions/include)
+			for (const pluginOpts of protoOptions) {
+				Object.assign(obj.options, pluginOpts);
+			}
+		} else {
+			// Legacy path: options set via setDefaultOptions on the prototype
```
Falke-Design, line 122: `this.setDefaultOptions` is new with v2-alpha, we can remove the fallback

Falke-Design (commented):
Thank you for your contribution, this is very helpful!

Alternative to: #10069 - @simon04 can you please review this one too


> The current implementation (and our PR's approach) preserves live inheritance — changes to a parent class's defaults are visible in all subclasses and subsequently created instances that haven't overridden that property.

I think we should keep that behavior, this is pretty useful for plugins

>Also worth noting: your approach drops .include()-based plugin options entirely (the constructor spread never touches prototype.options), whereas our approach still collects those at construction time. But that's likely a separate concern from what this PR is trying to solve.

I'm still a big fan of `.include()` and we should keep it in my eyes

PR #10190, opened 2026-04-19
Accessibility / WCAG 1.4.1: Increase color contrast

mfranzke:
## Problem description

The text and links within the attribution container (CSS selector `.leaflet-control-attribution`) only get differentiated by `color`, which doesn't fit the necessary contrast ratio of at least 3:1, according to [WCAG Success Criteria 1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html).

Further information on the topic are included in the [WCAG Technique G183: Using a contrast ratio of 3:1 with surrounding text and providing additional visual cues on hover for links or controls where color alone is used to identify them](https://www.w3.org/WAI/WCAG22/Techniques/general/G183)

## Steps to reproduce

- Open https://leafletjs.com/ and go an included map that even also includes contents in the attribution container
- Get the `color` CSS property values for both the text and links in this area. Text `color` is `#333333`, link `color` is `#0078A8`
- The calculation shows that these values measure at a ratio of 2.55823:1, which is less than the required 3:1

image

## Solution

Implemented another color that get's suggested by the following online tool (please don't get confused by their labels "background" and "foreground" color, we're essentially measuring two text CSS property `color` values): https://app.contrast-finder.org/result.html?foreground=%23333333&background=%230078a8&ratio=3&isBackgroundTested=true&algo=Rgb&lang=de

### DIFF
--- a/src/leaflet.css
+++ b/src/leaflet.css
@@ -291,7 +291,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	outline-offset: 1px;
 
 	a {
-		color: #0078a8;
+		color: #0284B7;
 	}
 }
 


mourner:
I think you calculated the contrast incorrectly — you measure it as if it were a background color with a `#333333` text color according to that link, and the proposed color leads to even lower contrast (it's lighter). 

With the existing color as the foreground and `#cccccc` as the background (which is white blended on black with 0.8 opacity), it comes out as 3.08 (>3), and in practice it's higher because attribution is usually blended on lighter colors.

mfranzke:
@mourner thanks a lot for your feedback. I think that my description might have caused a misunderstanding (the labels background and foreground color in the tool https://app.contrast-finder.org/might enforce this as well) here – I didn't want to report anything related to text- and **background** color in the attribution container, but about the **text and link colors**. I should have added further information to the description and just did that. If there are any further question, please don't hesitate to further comment and disuss on this.

mourner:
@mfranzke ah, I see the issue now — thanks for the clarification. However, the proposed change, while solving the contrast between the link and the text, decreases contrast of the link over the background — from ~4.5 (AA minimum) to ~3.8. If we were to choose whether to have a more readable link or a more discerning link from text, the first seems more important for accessibility overall. At the same time we don't want to increase overall contrast of the attribution block because it's intentionally in the background, not taking the focus away from the map too much.

mfranzke:
> @mfranzke ah, I see the issue now — thanks for the clarification. However, the proposed change, while solving the contrast between the link and the text, decreases contrast of the link over the background — from ~4.5 (AA minimum) to ~3.8. If we were to choose whether to have a more readable link or a more discerning link from text, the first seems more important for accessibility overall. At the same time we don't want to increase overall contrast of the attribution block because it's intentionally in the background, not taking the focus away from the map too much.

@mourner could catch, thank you. So I do see two remaining options, that we have on this:
- Use another signal instead of color. They could e.g. get the regular link underline.
- Ignore this issue. As the text that it differentiates to is mainly the copyright sign and "contributors", it might be acceptable to not do any changes.

Which of these do you prefer ?

mourner:
> * Ignore this issue. As the text that it differentiates to is mainly the copyright sign and "contributors", it might be acceptable to not do any changes.

Let's leave things as is — I think it's good enough.

Issue #9869, opened 2025-09-07
The targeted release date for Leaflet V2.0 is November 2025


Falke-Design:

Spixmaster:
May I ask for an update regarding the release timeframe?

jonkoops:
We just started merging the big breaking changes that we need for a stable (see #10006). I am hoping we can get a new alpha out quickly after I make some of the required changes. Once this is done, we will start moving to release candidates, which will require us to have a migration guide in place, as well as updated documentation.

indigoxela:
So, the targeted release date was November. But breaking changes continue, huge ones still in the queue, apparently.

To my understanding removing initialize() will require all plugins to undergo another rewrite, which renders all "compatible-v2" feedback so far futile.

And most of all: removing intialize() entirely will break the Leaflet-V1-polyfill concept, forcing all integrations with incompatible plugins to stick with v1.

_Heretic question:_ isn't it a little _late_ in the ongoing preparation for v2 for such a far-reaching change? Wouldn't it make sense to push the removal of initialize() to v3 instead?
Otherwise refactoring goes on and on and *nobody* gets any modernization.

To my understanding, plugin maintainers have been notified about the "upcoming" major release and tested with what was in main branch several months ago. Would they even know, that they have to start over?

Never underestimate stability and reliability. Not only regarding the software itself, but also regarding resource planning for maintenance tasks (like getting compatible with the main library).
Leaflet isn't just a playground for students. It's the base for many integrations, used by hundreds of thousands of projects. It's also part of a large ecosystem of plugins, where hundreds of people are involved.

So please make their maintenance tasks not harder than necessary. Stop refactoring and postpone further (deeply) breaking changes to the next major.

jonkoops:
It doesn't make sense to me to push the breaking changes for the class system half into 2.x and another pass into 3.x. As we've discussed in detail in the related PRs and issues, there is no way to make the existing systems interoperable, so a hard break is needed. I really wish it was different, but this is the reality of the situation.

The intention is for Leaflet 2.x to be a long-lived release, like 1.x was. I do not want to put out Leaflet 2.x, only to then do another breaking release a couple of months down the line. As far as I am concerned, it should be stable at least for a year or two, but other maintainers might differ in opinion.

[...]

PR #10202, opened 2026-05-07
Use standard `constructor()` instead of `initialize()`

### LINKED ISSUE #9889: Use JavaScript `constructor()` instead of custom `initialize()`

jonkoops:
Replaces all `initialize()` methods with standard JavaScript `constructor()` across all built-in Leaflet classes.

A new `withInitHooks()` decorator using a `Proxy` is introduced to preserve `addInitHook()` functionality.

Previously, the `constructor()` of `Class` was responsible for calling `initialize()`, followed by walking the prototype chain to trigger init hooks. This was fundamentally incompatible with standard JavaScript constructors, as `super()` must be called before accessing `this`, meaning hooks could not access properties set after `super()` (see #9889 for detailed analysis).

The new approach wraps each class with `withInitHooks()`, which uses a `Proxy` to intercept construction via `Reflect.construct()`. Hooks registered with `addInitHook()` are triggered after each level's constructor completes, preserving the original execution order.

Given the following class hierarchy where both classes are wrapped and have hooks registered:

```js
const A = withInitHooks(class A {
  constructor() { console.log('1. A constructor'); }
});
A.addInitHook(function () { console.log('2. A hook'); });

const B = withInitHooks(class B extends A {
  constructor() { super(); console.log('3. B constructor'); }
});
B.addInitHook(function () { console.log('4. B hook'); });

new B();
// 1. A constructor
// 2. A hook
// 3. B constructor
// 4. B hook
```

The following diagram illustrates the construction order when calling `new B()`:

```mermaid
sequenceDiagram
    participant User
    participant B_Proxy
    participant B_Constructor
    participant A_Proxy
    participant A_Constructor
    participant A_Hooks
    participant B_Hooks

    User->>B_Proxy: new B(args)
    B_Proxy->>B_Constructor: Reflect.construct()
    B_Constructor->>A_Proxy: super(args)
    A_Proxy->>A_Constructor: Reflect.construct()
    A_Constructor-->>A_Proxy: instance
    A_Proxy->>A_Hooks: hook.call(instance)
    A_Hooks-->>B_Constructor: return
    B_Constructor-->>B_Proxy: instance
    B_Proxy->>B_Hooks: hook.call(instance)
    B_Hooks-->>User: instance
```

## Breaking changes

- **`initialize()` is removed from all built-in classes.** Subclasses that override `initialize()` and call `ParentClass.prototype.initialize.call(this, ...)` must migrate to `constructor()` with `super(...)`.
- **Classes using `addInitHook()` must be wrapped with `withInitHooks()`.** Built-in classes are wrapped by default. Plugin authors extending Leaflet classes who wish to register init hooks on their own class must wrap it:
  ```js
  import { Layer, withInitHooks } from 'leaflet';

  const MyLayer = withInitHooks(class MyLayer extends Layer {
    constructor(options) {
      super(options);
    }
  });

  MyLayer.addInitHook(function () { /* ... */ });
  ```

## Migration guide

```js
// Before
class MyMarker extends Marker {
  initialize(latlng, options) {
    Marker.prototype.initialize.call(this, latlng, options);
    this.custom = true;
  }
}

// After
class MyMarker extends Marker {
  constructor(latlng, options) {
    super(latlng, options);
    this.custom = true;
  }
}
```

Closes #9889

### DIFF
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -7,7 +7,7 @@ import * as Util from './Util.js';
 
 // Thanks to John Resig and Dean Edwards for inspiration!
 
-export class Class {
+export const Class = withInitHooks(class Class {
 	// @function include(properties: Object): this
 	// [Includes a mixin](#class-includes) into the current class.
 	static include(props) {
@@ -50,55 +50,47 @@ export class Class {
 
 	// @function addInitHook(fn: Function): this
 	// Adds a [constructor hook](#class-constructor-hooks) to the class.
-	static addInitHook(fn, ...args) { // (Function) || (String, args...)
+
+	constructor() {
+		Util.setOptions(this);
+	}
+});
+
+// @function withInitHooks(Object TargetClass): Function
+// Decorates a class to be able to use [constructor hooks](#class-constructor-hooks) functionality.
+/**
+ * @template T
+ * @param {T} TargetClass
+ * @returns {T}
+ */
+export function withInitHooks(TargetClass) {
+	const hooks = [];
+	const proxy = new Proxy(TargetClass, {
+		construct(target, args, newTarget) {
+			const instance = Reflect.construct(target, args, newTarget);
+
+			for (const hook of hooks) {
+				hook.call(instance);
+			}
+
+			return instance;
+		}
+	});
+
+	TargetClass.addInitHook = function addInitHook(fn, ...args) {
+		if (this !== proxy) {
+			const className = this.name || '(anonymous)';
+			throw new Error(`The 'addInitHook()' method can only be called on classes wrapped with 'withInitHooks()'. Try wrapping your class:\n\nconst ${className} = withInitHooks(class ${className} { ... });\n`);
+		}
+
 		const init = typeof fn === 'function' ? fn : function () {
 			this[fn].apply(this, args);
 		};
 
-		if (!Object.hasOwn(this.prototype, '_initHooks')) { // do not use ??= here
-			this.prototype._initHooks = [];
-		}
-		this.prototype._initHooks.push(init);
+		hooks.push(init);
+
 		return this;
-	}
+	};
 
-	constructor(...args) {
-		this._initHooksCalled = false;
-
-		Util.setOptions(this);
-
-		// call the constructor
-		if (this.initialize) {
-			this.initialize(...args);
-		}
-
-		// call all constructor hooks
-		this.callInitHooks();
-	}
-
-	callInitHooks() {
-		if (this._initHooksCalled) {
-			return;
-		}
-
-		// collect all prototypes in chain
-		const prototypes = [];
-		let current = this;
-
-		while ((current = Object.getPrototypeOf(current)) !== null) {
-			prototypes.push(current);
-		}
-
-		// reverse so the parent prototype is first
-		prototypes.reverse();
-
-		// call init hooks on each prototype
-		for (const proto of prototypes) {
-			for (const hook of proto._initHooks ?? []) {
-				hook.call(this);
-			}
-		}
-
-		this._initHooksCalled = true;
-	}
+	return proxy;
 }
--- a/spec/suites/core/ClassSpec.js
+++ b/spec/suites/core/ClassSpec.js
@@ -1,8 +1,14 @@
 import {expect} from 'chai';
-import {Class} from 'leaflet';
+import {Class, withInitHooks} from 'leaflet';
 import sinon from 'sinon';
 
 describe('Class', () => {
+	it('sets default options', () => {
+		const instance = new Class();
+		expect(Object.hasOwn(instance, 'options')).to.be.true;
+		expect(instance.options).to.eql({});
+	});
+
 	describe('#extends', () => {
 		it('merges options instead of replacing them', () => {
 			class KlassWithOptions1 extends Class {
@@ -28,37 +34,6 @@ describe('Class', () => {
 			expect(a.options.foo3).to.eql(4);
 		});
 
-		it('inherits constructor hooks', () => {
-			const spy1 = sinon.spy(),
-			spy2 = sinon.spy();
-
-			class Class1 extends Class {}
-			class Class2 extends Class1 {}
-
-			Class1.addInitHook(spy1);
-			Class2.addInitHook(spy2);
-
-			new Class2();
-
-			expect(spy1.called).to.be.true;
-			expect(spy2.called).to.be.true;
-		});
-
-		it('does not call child constructor hooks', () => {
-			const spy1 = sinon.spy(),
-			spy2 = sinon.spy();
-
-			class Class1 extends Class {}
-			class Class2 extends Class1 {}
-
-			Class1.addInitHook(spy1);
-			Class2.addInitHook(spy2);
-
-			new Class1();
-
-			expect(spy1.called).to.be.true;
-			expect(spy2.called).to.eql(false);
-		});
 	});
 
 	describe('#include', () => {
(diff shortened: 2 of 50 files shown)


IvanSanchez:
At first sight, I must say this is a pretty clever solution :heart_eyes:

src/core/Class.js
```diff
@@ -7,7 +7,7 @@ import * as Util from './Util.js';
 
 // Thanks to John Resig and Dean Edwards for inspiration!
 
-export class Class {
+export const Class = withInitHooks(class Class {
```
IvanSanchez, line 10: I'm not sure that doing the init-hooks-proxy on the base `Class` is not a good idea. One of the goals of this is to get rid of extra code paths when instantiating subclasses, right?

IvanSanchez (changes requested):
I think that we could get away with removing all calls to `withInitHooks` in the core, and provide a compatibility wrapper, which would be just all the exports from `L` but wrapped in `withInitHooks`. That way there would be a purist Leaflet 2 package, and a Leaflet-compat one for transitioning.

mourner:
Love the constructor change — very nice to see Leaflet getting closer to modern ES idioms!

> I think that we could get away with removing all calls to `withInitHooks` in the core, and provide a compatibility wrapper, which would be just all the exports from `L` but wrapped in `withInitHooks`.

That's a pretty good idea. For me personally, the whole `addInitHook` API now feels wrong and an architectural mistake — I did it years before ES classes where even remotely a thing, and plugins hooking directly into core class constructors to augment their logic feels like an anti-pattern. The correct idiom for that kind of thing is to either subclass core classes with your own logic + `super`, or hook into specific events in a map lifecycle to insert custom logic — that way there's much less chance of interfering with other plugins or changes on the Leaflet side.

I know Leaflet geoman depends on those hooks a lot, but if we're breaking API with 2.0 anyway, why not break just a little bit further to make our code elegant, modern, simple and free of uglyish hacks like all that Reflect voodoo? Plugins like geoman can afford to add compat layers or invest in refactoring for a more robust plugin architecture. cc @Falke-Design

Falke-Design:
I know that you all see `addInitHook` as an anti-pattern but to much plugins relay on this helpful and nice feature in my eyes. It is not just geoman.

> I've cloned almost all leaflet plugins and searched for occurrences of `addInitHook` excluding `leaflet.js,leaflet-src.js,bundle.js,*min*,leaflet-0.*js,leaflet-1.*js,dist/,docs/,vendor*` -- this is the result:
> 
> ```
>       1 BingLayer.addInitHook
>       5 Circle.addInitHook
>       5 CircleMarker.addInitHook
>       1 cls.addInitHook
>       1 DataLayer.addInitHook
>       1 DistortableCollection.addInitHook
>       1 DistortableImageOverlay.addInitHook
>       1 Evented.addInitHook
>       2 Feature.addInitHook
>       1 FeatureGroup.addInitHook
>       1 GeoJSON.addInitHook
>       1 ImageOverlay.addInitHook
>       1 Keymapper.addInitHook
>       1 LayerGroup.addInitHook
>      79 Map.addInitHook
>      11 Marker.addInitHook
>       1 Path.addInitHook
>       1 Pointer.addInitHook
>       1 Poly.addInitHook
>       1 Polygon.addInitHook
>       7 Polyline.addInitHook
>       1 polylineEditor.addInitHook
>       3 Rectangle.addInitHook
>       1 SimpleShape.addInitHook
>       1 SvgObject.addInitHook
>       1 Textbox.addInitHook
>       1 this.addInitHook
>       2 TileLayer.addInitHook
>       2 var addInitHook
>       4 Yandex.addInitHook
> ```
> ...
 _Originally posted by @simon04 in [#9889](https://github.com/Leaflet/Leaflet/issues/9889#issuecomment-4076699240)_


> The correct idiom for that kind of thing is to either subclass core classes with your own logic + super, or hook into specific events in a map lifecycle to insert custom logic

For plugins like Geoman, Rotate-Marker, etc., it is essential to add code to the base classes like Marker, Polyline, etc. for cross-plugin compatibility.

If we take a look at the `addInitHook`s in Geoman: https://github.com/geoman-io/leaflet-geoman/blob/e811eb097f8d5c38b14fb48983a1a680be478f16/src/js/L.PM.js#L115

It would be possible to refactor this by listening to `layeradd`, but this only works if the layer is actually added to the map. If a layer is created but (for now) not added to the map, the function will never be called - and this currently works with addInitHook. Or am I overlooking some feature that would make this achievable?

Anyway, I’ll check whether Geoman really needs the addInitHook logic for layers that are not added to the map yet — but there are other plugins out there as well.


Maybe we need to dig into usages of `addInitHook` to understand if there are logics that can't be replaced with some other code

simon04:
Me in https://github.com/Leaflet/Leaflet/issues/9889#issuecomment-3315536709

> I dislike the complexity added with `withInitHooks`. In the end we're only changing names (`initialize` to `constructor`) without gaining any real benefit.
> 
> A huge strength of Leaflet always has been to be simple to use and simple to extend. I think we should adhere to this design principle.

This proxy concept clutters JS debugging consoles. Instead of showing objects as class instances, they are instances of a proxy object and the actual class implementation is somewhere hidden.

----

I would like to propose (yet another) alternative. Classes `C` that wish to support init hooks are required to call `this.callInitHooks(C)` somewhere appropriate in their constructor:

```diff
diff --git a/src/core/Class.js b/src/core/Class.js
index a27ec63f..d3098a99 100644
--- a/src/core/Class.js
+++ b/src/core/Class.js
@@ -73,32 +73,15 @@ export class Class {
 		}
 
 		// call all constructor hooks
-		this.callInitHooks();
+		this.callInitHooks(Class);
 	}
 
-	callInitHooks() {
-		if (this._initHooksCalled) {
-			return;
+	// @method callInitHooks(cls: subclass of Class): this
+	// All subclasses C that want to support init hooks need to call this.callInitHooks(C) in their constructor.
+	callInitHooks(cls) {
+		for (const hook of cls.prototype._initHooks ?? []) {
+			hook.call(this);
 		}
-
-		// collect all prototypes in chain
-		const prototypes = [];
-		let current = this;
-
-		while ((current = Object.getPrototypeOf(current)) !== null) {
-			prototypes.push(current);
-		}
-
-		// reverse so the parent prototype is first
-		prototypes.reverse();
-
-		// call init hooks on each prototype
-		for (const proto of prototypes) {
-			for (const hook of proto._initHooks ?? []) {
-				hook.call(this);
-			}
-		}
-
-		this._initHooksCalled = true;
+		return this;
 	}
 }
diff --git a/src/map/Map.js b/src/map/Map.js
index 0eb1e564..3b1e1dff 100644
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -168,7 +168,7 @@ export class LeafletMap extends Evented {
 			this.setView(new LatLng(options.center), options.zoom, {reset: true});
 		}
 
-		this.callInitHooks();
+		this.callInitHooks(LeafletMap);
 
 		// don't animate on browsers without hardware-accelerated transitions or old Android
 		this._zoomAnimated = this.options.zoomAnimation;
```

src/core/Class.js
jonkoops: This is nothing more than a backwards compatibility move; some plugins/user code might be hooking into the lowest level class.

jonkoops:
> I think that we could get away with removing all calls to `withInitHooks` in the core, and provide a compatibility wrapper, which would be just all the exports from `L` but wrapped in `withInitHooks`. That way there would be a purist Leaflet 2 package, and a Leaflet-compat one for transitioning.

I'd rather make a clean cut. If we decide to remove this functionality, we are so far down the road with breaking changes for v2 that providing some kind of 'compatible' export would simply be ceremonial. I'll be [volunteering to maintain 1.x](https://github.com/Leaflet/Leaflet/issues/9869#issuecomment-3687647477) for the foreseeable future (backporting bug fixes and CVEs); that, to me, is a better way of supporting users stuck because of API incompatibilities.

> I know Leaflet geoman depends on those hooks a lot, but if we're breaking API with 2.0 anyway, why not break just a little bit further to make our code elegant, modern, simple and free of uglyish hacks like all that Reflect voodoo? 

I am inclined to agree with this; at that point the class system has so many breaking changes that the 'migration' surface we offer might as well be removed entirely, and we can start with a clean slate. We might then as well:

- Drop `Class` entirely, together with `Class#include()`.
- Move `Class#setDefaultOptions()` and `Class#mergeOptions()` into `Util`.
- Change all `_` 'private' fields to actually be [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements).

I think the more important question, though, is if we make these sorts of changes, how can we provide a migration path for plugins that depend on these APIs? I think we'd have to come up with a new, well-designed public API that allows plugins to hook into core components (e.g., `Map`, `Layer`) and their various lifecycles and events without relying on private APIs. I'd like to get rid of this idea that external components can modify the internal state (e.g., `this`) of a built-in class.

Perhaps the way forward is to remove these `Class` functionalities and then take the most common plugin use cases as mentioned by @Falke-Design and develop an API around those concerns that can provide them a path to 2.x compatibily.

Falke-Design:
If we provide some hook / event which is fired after creation of a layer, no matter if it is added to the map or not, I think we can get rid of callInitHooks. All my personal useses would be covered with a `layer instanceof Marker`:

```
_someAvailableGlobalRef_.on('init', (e)=>{
  if(e.layer instanceof Marker){
    e.layer.pm = new Geoman(e.layer)
    e.layer.once("add", () => {
  }
});
```
or
```
Marker.on('init', (e)=>{
  e.layer.pm = new Geoman(e.layer)
  e.layer.once("add", () => {
});
```


To your other points:
- Drop `Class` entirely, together with `Class#include()`. -> I always liked the easy and clean way to add properties and custom funktions. To get rid of it, we need cover the functionality by antoher way, for example with the new init-event, and some other way to add functions to a base class like Marker.

- Move `Class#setDefaultOptions()` and `Class#mergeOptions()` into `Util`. -> Fine for me

- Change all `_` 'private' fields to actually be [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements). -> I’m strongly against this approach. If someone is using our private fields (`_`), there is usually a reason for it - often because we did not provide sufficient public methods to achieve the same goal. Users are aware that these fields may change at any time, but in their specific use case, using them may be the only viable solution. In my view, plugin ecosystems and customization tend to work better without such restrictions. Over the years, I have already worked with several unmaintained Leaflet plugins where the only way to fix bugs, was to override internal functions using .include or by patching internal fields directly. If we want to keep maintenance burden low, an open approach may be more sustainable in the long run.

simon04:
> If we provide some hook / event which is fired after creation of a layer

+1

That's a very neat idea and a concise concept since we have all methods already in place via `Evented`. The init hooks for the various core map handlers could most likely be converted to "init" handles of `LeafletMap`.

Thus, we would only add two `fire('init', ...)` (in LeafletMap and Layer).

Btw: it just stroke me that the [first 2.0 alpha](https://github.com/Leaflet/Leaflet/releases/tag/v2.0.0-alpha) has been released 13+ months ago. I wonder whether users, people or companies, are expecting a subsequent at some point in time...?

```diff
diff --git a/src/layer/Layer.js b/src/layer/Layer.js
index 7958559b7..d6c50dff1 100644
--- a/src/layer/Layer.js
+++ b/src/layer/Layer.js
@@ -44,6 +44,12 @@ export class Layer extends Evented {
 		});
 	}
 
+	constructor() {
+		super();
+		// ...
+		this.fire('init', this);
+	}
+
 	/* @section
 	 * Classes extending `Layer` will inherit the following methods:
 	 *
diff --git a/src/map/Map.js b/src/map/Map.js
index d32eca949..2c150e95b 100644
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -179,6 +179,7 @@ export class LeafletMap extends Evented {
 		}
 
 		this._addLayers(this.options.layers);
+		this.fire('init', this);
 	}
 
 	// @section Methods for modifying map state
```

```
 git grep addInitHook src/map
src/map/handler/BoxZoomHandler.js:LeafletMap.addInitHook('addHandler', 'boxZoom', BoxZoomHandler);
src/map/handler/DoubleClickZoomHandler.js:LeafletMap.addInitHook('addHandler', 'doubleClickZoom', DoubleClickZoomHandler);
src/map/handler/DragHandler.js:LeafletMap.addInitHook('addHandler', 'dragging', DragHandler);
src/map/handler/KeyboardHandler.js:LeafletMap.addInitHook('addHandler', 'keyboard', KeyboardHandler);
src/map/handler/PinchZoomHandler.js:LeafletMap.addInitHook('addHandler', 'pinchZoom', PinchZoomHandler);
src/map/handler/PinchZoomHandler.js:LeafletMap.addInitHook(function () {
src/map/handler/ScrollWheelZoomHandler.js:LeafletMap.addInitHook('addHandler', 'scrollWheelZoom', ScrollWheelZoomHandler);
src/map/handler/TapHoldHandler.js:LeafletMap.addInitHook('addHandler', 'tapHold', TapHoldHandler);
```

Falke-Design:
I don't think that a plugin will be able to listen on the init event if it is fired on the layer /map which is just created. The event needs to be fired on some global object, where a plugin can register on

jonkoops:
> If we provide some hook / event which is fired after creation of a layer, no matter if it is added to the map or not, I think we can get rid of callInitHooks. All my personal useses would be covered with a `layer instanceof Marker`.

So the proposal would be to have a more narrow API that only targets layers? A global event system that is triggered the moment a layer is constructed? If that can cover the existing plugins' use cases then I'd be more than happy to accept something along those lines. It would also make it more explicit which objects are supposed to be hooked like this, without it being on every single class.

Perhaps we could simply handle this as a static method on `Layer` itself though? That would be a logical hook up point for plugin authors.

As for the point about private fields (and by extension `Class#include()`), the reason why I mention this is precisely the one raised by @Falke-Design:

> If someone is using our private fields (`_`), there is usually a reason for it - often because we did not provide sufficient public methods to achieve the same goal. Users are aware that these fields may change at any time, but in their specific use case, using them may be the only viable solution. 

This essentially means that all `_` APIs are a public API, whether we want it or not, and that means people build and rely on it. Hence, we cannot evolve the private API, even though we soft agree these are 'dangerous' paths, we end up with no motivation to build a better public API. I believe these APIs should be private going forward, and if there is no suitable public API, we must create one. Doing so will result in a better product.

simon04:
> I don't think that a plugin will be able to listen on the init event if it is fired on the layer /map which is just created.

Right. I tested whether it's feasible to make it a static function `LeafletMap.on('init', ...)` and have `Evented` handle that.

Here's what I came up with: https://github.com/Leaflet/Leaflet/pull/10275

simon04:
Hi all, I would love to get some feedback on my PR https://github.com/Leaflet/Leaflet/pull/10275.

On this refactor and the road to version 2.0: is there a target timeline, and do we have a process for deciding which refactoring path to pursue?

Issue #9773, opened 2025-07-02
Consider migrating from innerHTML to avoid TrustedTypes errors


lukewarlow:
Currently leaflet makes use of innerHTML (and potentially other similar XSS sinks), if it could move away from these to safe APIs such as innerText or manually building the DOM tree this would make it easier to use Leaflet in an environment that enforces Trusted Types.

SanatKulkarni:
Hi Luke, I have used Leaflet during my 2nd year in University and it has really been helpful to me ever since, I am excited to contribute if possible, I have identified some components where migration can take place from innerHTML to avoid TrustedTypes errors:

In Control.Attribution.js 
- Issue: in the _update() method, in line 121:
`this._container.innerHTML = prefixAndAttribs.join(' <span aria-hidden="true">|</span> ');`
is using innerHTML to dynamically set the HTML content, which is violating the Trusted Types security policies. 

- Fix: 
We clear the container with textContent=''. Create DOM elements programatically using document.createElement() and document.createElementNS() since the Ukranian flag SVG is bugging out if we do not do this. We can then add the text content using textContent or createTextNode() instead of the HTML strings that are currently being used. Building seperator spans manually rather than injecting the HTML and constructing the Ukranian flag SVG and link elements step by step. 

This will eliminate the XSS risks while maintaining the functionality and appearance.

Just a small change, I have identified others as well. 

I would get started on solving these problems if you could respond or assign the issue to me. Happy to help :D

Falke-Design:
Related discussions:

- #6465
- #7580
- #8423 
- #9429 

I think we could replace some innerHTML usages but the following will be not replaceable:

- https://github.com/Leaflet/Leaflet/blob/main/src/layer/marker/DivIcon.js#L53
- https://github.com/Leaflet/Leaflet/blob/main/src/layer/DivOverlay.js#L278
- https://github.com/Leaflet/Leaflet/blob/main/src/control/Control.Attribution.js#L121

> ...therefore the content is 100% under control of the consumer developer.
> As such, it seems to me that it is the consumer developer's responsibility how they use Leaflet library and call such Leaflet method with data sanitized at app level, in the rare case where this data comes from arbitrary source (e.g. end user text input or fetched from remote service).

 _Originally posted by @ghybs in [#6465](https://github.com/Leaflet/Leaflet/issues/6465#issuecomment-452950618)_

lukewarlow:
[...]

PR #10205, opened 2026-05-11
Document handling untrusted HTML in popups/tooltips

mourner:
We're getting hounded about this CVE https://www.cve.org/CVERecord?id=CVE-2025-69993, which is in it's essence not a valid vulnerability issue since `bindPopup` and similar methods accept HTML input **by design**, and not passing untrusted user input to it or sanitizing it is the responsibility of application developers rather than a JS library (with many similar precedents existing for jQuery, React, etc.).

However, we could do a better job of documenting that aspect in relevant sections. This PR makes a pass at updating the docs — both the source and the generated files, just so that we can update it everywhere with one PR merge for now.

Related: #9773

### DIFF
--- a/docs/examples/geojson/index.md
+++ b/docs/examples/geojson/index.md
@@ -168,6 +168,12 @@ new GeoJSON(geojsonFeature, {
 	onEachFeature: onEachFeature
 }).addTo(map);</code></pre>
 
+<p><strong>Security note.</strong> <code>bindPopup</code> renders its string argument as HTML. If your GeoJSON comes from untrusted sources (user uploads, third-party APIs, etc.), sanitize <code>popupContent</code> with a library like <a href="https://github.com/cure53/DOMPurify">DOMPurify</a>, or build a DOM element via <code>textContent</code> and bind that instead:</p>
+
+<pre><code>const el = document.createElement('div');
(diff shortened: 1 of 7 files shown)


mourner:
Merging this without a review so that it's live ASAP (to avoid more unwarranted attacks about that CVE), but let's follow up with other PRs if you have concerns.

Falke-Design:
@mourner FYI If you want to have this on the website right now, you need to add the changes to the `v1` branch

mourner:
@Falke-Design yep, already did

jonkoops (commented):
I'd rather not recommend specific libraries, as these might end up unmaintained, and are therefore also a possible attack vector. Ideally, this recommendation would be the[ Sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API) built into browsers, and will be updated as part of the platform itself, with a fallback for unsupported agents.

We might also want to rename these methods in 2.x (see https://github.com/Leaflet/Leaflet/issues/10214#issuecomment-4429932526)

jonkoops:
Incorporated the feedback in a PR https://github.com/Leaflet/Leaflet/pull/10215

PR #9243, opened 2024-02-05
Migrate from Karma to Vitest

jonkoops:
Closes #8939. Alternative implementation to #9242, just as a means to evaluate our options.

PR #10207, opened 2026-05-11
Switch from Karma to Vitest

### LINKED ISSUE #8939: Karma is deprecated

mourner:
Closes #8939. Supercedes #9243 and #9242.

- [x] Get no-touch pass green on CI
- [x] Get touch pass green on CI
- [x] See if we can simplify/improve the CI matrix set up
- [x] See if we can fix occasional webkit job hangs

---

Karma has been unmaintained since 2023 and pulled along a long tail of legacy infrastructure (RequireJS, standalone Mocha, custom launchers, browser plugins per browser). This PR replaces it with **Vitest browser mode** on the Playwright provider, modernizing the setup end-to-end.

### Highlights

- **One modern stack.** Karma + 7 `karma-*` plugins + RequireJS + standalone Mocha all removed. Vitest + `@vitest/browser-playwright` handle browser launching, test isolation, source maps, and reporting in a single integrated runner.
- **Real browser engines via Playwright.** Tests run in Chromium, Firefox, and WebKit (and a Firefox retina/`deviceScaleFactor: 2` variant) — same provider used across the JS ecosystem, with versioned browser binaries managed by Playwright instead of system Chrome/Firefox.

### Library-side leak fixes uncovered during the migration

Three real bugs surfaced as silent WebKit hangs while debugging CI flakes — all fixed here:

- `PinchZoomHandler._onPointerEnd` early-return leaked `pointermove`/`pointerup`/`pointercancel` listeners; `removeHooks` didn't clean up mid-pinch.
- `TapHoldHandler.removeHooks` left mid-tap listeners attached.
- `DomEvent.PointerEvents.cleanupPointers()` cleared its map without releasing pointer captures, pinning detached DOM elements.

### Follow-ups for a separate PR:

- Sinon → Vitest `vi.fn()` / `vi.spyOn` migration (~400 call sites).
- Move global chai matchers + `skipIfTouch` helpers from `SpecHelper.js` into `setup.js`.
- Migrate remaining `(done) => …` specs to `async`.


IvanSanchez:
@mourner What's the workflow for running the tests interactively in a browser (setting breakpoints and all that)? I recall having a `specrunner.html` at some point, and I'm missing that.

mourner:
@IvanSanchez there's a pretty neat [Vitest UI mode](https://vitest.dev/guide/ui) for that, no need for custom-maintained hacks.

Falke-Design (approved):
not tested locally but the code looks good

mourner:
Apologies, a bit of a rush to merge this because I couldn't even get the Karma stuff to work properly locally for me anymore, so I decided it's time for a change.

IvanSanchez:
> there's a pretty neat [Vitest UI mode](https://vitest.dev/guide/ui) for that, no need for custom-maintained hacks.

@mourner As far as I can see, that's just a UI for the underlying runner, which in turn runs a browser in the background.

I mean: when I run `npm test --ui`, a new tab opens in my default Firefox, and playwright spawns a Chromium process in the background. The tests run in the background browser, not the one I'm looking at.

I might need to do a bit more research about this.

mourner:
@IvanSanchez yep, I addressed that with this earlier commit: https://github.com/Leaflet/Leaflet/pull/10207/changes/5eb9e3750e015cd159f9c68153dea5aaa56fcb9c — it's a bit unusual but I think it's worth getting used to.

IvanSanchez:
@mourner Thanks for adding that to the docs :grinning:

mourner:
As a follow up, I also landed various cleanups in #10210, and switched to running tests against `src/Leaflet.js` instead of the built file in https://github.com/Leaflet/Leaflet/commit/34ada17342b717f1879e8b6f60a57bb0ad9cd633.

jonkoops:
Nice, been meaning to move away from Karma since a while, nice cleanup.

PR #10208, opened 2026-05-11
fix: default icon imagePath was treated as undefined if empty

Rag404:
I'm in a situation where I want to set the default `iconUrl` to some full URL, and have `imagePath` set to an empty string.

However, doing so results in leaflet trying to auto-detect a path instead of simply using the empty string.

I don't know if this behavior was intended, but if it wasn't, a simple fix would be to have `IconDefault` checking if `imagePath` is nullish and not falsly.


Rag404:
In the meantime, a workaround I found is to set `DefaultIcon.prototype.imagePath` to `new String('')` instead of `''` as the object is truthy.

IvanSanchez (approved):
LGTM.

Gotta love very small changes with a clearly defined reason for them :heart_eyes:

PR #10215, opened 2026-05-12
Recommend Sanitizer API instead of DOMPurify

jonkoops:
Follow-up to #10205. Rather than recommending a specific third-party library (DOMPurify), point users to the browser's built-in [Sanitizer API](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Sanitizer_API) with a note about fallbacks for unsupported browsers.

### DIFF
--- a/docs/examples/geojson/index.md
+++ b/docs/examples/geojson/index.md
@@ -168,7 +168,7 @@ new GeoJSON(geojsonFeature, {
 	onEachFeature: onEachFeature
 }).addTo(map);</code></pre>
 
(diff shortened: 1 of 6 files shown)


IvanSanchez (approved):

mourner (commented):
This looks good but IMO "with fallback" hides a lot of complexity — you can't just use the API as in example and then discover your app doesn't work in Safari. I'd lean on keeping this section as small and simple as possible — maybe leave the text content example, just drop the DOMPurify reference.

Also worth noting our website is built from the `v1` branch so this will need a follow-up PR

PR #10217, opened 2026-05-16
Replace twitter with mastodon

jjimenezshaw:
Replace the twitter (a.k.a. X) link in the main webpage (v2.html), and replace it with Mastodon.
https://mapstodon.space/@leaflet

The svg was taken from https://joinmastodon.org/branding

image


mourner:
Huh, I didn't even know we had that 😅 is that you @IvanSanchez? Cool! Not sure we should replace Twitter though — still a huge following there and the account isn't dead yet

jjimenezshaw:
I can also leave the X button. There is a link in the `<div class="social-buttons">` at the bottom... using iframes :see_no_evil:

IvanSanchez:
> Huh, I didn't even know we had that 😅 is that you @IvanSanchez? Cool! 

Yup, that's me. I'm happy to share the account's password with the rest of the team if we can agree in some secure means of communication (Signal?)

jjimenezshaw:
If you want I can remove the div with the iframes and embedded script
https://github.com/Leaflet/Leaflet/blob/main/docs/index.html#L188-L196

jonkoops:
> still a huge following there and the account isn't dead yet

I'd rather get off the Nazi tech bro platform, to be honest; staying there just feeds into its perceived value. Perhaps we can announce we moved to Mastodon there. Either way, this change has a +1 from me.

jonkoops (approved):
LGTM, but I want other maintainers to provide their take.

Falke-Design:
If we make a statement on Twitter, that we will no longer be available there, I'm okay with switching completly to mastodon

mourner (changes requested):
> If we make a statement on Twitter, that we will no longer be available there

Why would we leave? We have a huge following there (28.6k). We're still regularly tagged in posts about maps. I don't use Mastodon. So I'm against simply replacing one thing with another.

jonkoops:
Superseded by #10265.

PR #10224, opened 2026-05-19
Add Leaflet.AI.GeoJSON plugin

EricDalnas:
Adds an entry for [Leaflet.AI.GeoJSON](https://github.com/EricDalnas/leaflet.ai.geojson) 
to the **Dynamic/Custom data loading** section of the plugins page.

**What it does:**  
A lightweight Leaflet control that lets users query an LLM (Gemini, OpenAI, OpenRouter, etc.) 
for geographic data and renders the results as toggleable GeoJSON layers. Also supports 
navigation queries like "Zoom to Tokyo", named per-query layers with individual or bulk 
clear, a collapsible chat UI, model picker, and settings panel.

**Demo:** https://ericdalnas.github.io/leaflet.ai.geoJson/examples/proxy-azure/demo.html

### DIFF
--- a/docs/_plugins/dynamic-custom-data-loading/leaflet.AI.GeoJSON.md
+++ b/docs/_plugins/dynamic-custom-data-loading/leaflet.AI.GeoJSON.md
@@ -1,0 +1,17 @@
+---
+name: Leaflet.AI.GeoJSON
+category: dynamic-custom-data-loading
+repo: https://github.com/EricDalnas/leaflet.ai.geojson
+author: Eric Dalnas
+author-url: https://github.com/EricDalnas
+demo: https://ericdalnas.github.io/leaflet.ai.geoJson/examples/proxy-azure/demo.html
+compatible-v1: true
+compatible-v2: false
+---
+
+A lightweight Leaflet control that lets users query an LLM (Gemini, OpenAI,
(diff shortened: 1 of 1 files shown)


IvanSanchez:
I'm against adding to the list based on three reasons.

On one hand: There's ethical and ecological implications of LLMs. I cannot ignore those, and by default I am against LLM usage.

On the other hand: The naming scheme for Leaflet plugins implies/mandates that the name should be "Leaflet.GeoJSON.AI" instead.

On the other hand: The results are, to be honest, bullshit.

A quick Overpass query to primary schools in OSM yields about 450 in the Madrid metropolitan area:

image

Whereas your tool (via whatever model doesn't return a "quota exceeded" error) displays either 3 or 4 depending on the query:

image

On top of that, **all** results are "hallucinated". Of those seven results, only one is next to an actual school:

image

The only word I have to describe those results is "bullshit".

EricDalnas:
No problem, it was just more a POC on pulling in geojson from an LLM. I appreciate the feedback.

IvanSanchez:
@EricDalnas Apologies if I've been too blunt but, pal, the results are **bad**.

EricDalnas:
Sure, it's all dependent on the model. It can also search for things like the El Toro skateboard gap or the MLk I have a dream speech location that aren't in overpass  though.

PR #10227, opened 2026-05-22
fix: remove stale browser-compat guards and comments (refs #8477)

Gilmoursa:
Picks up the two remaining unchecked items from #8477.

## Changes

**`src/layer/vector/Canvas.js`**
Removes the `if (ctx.setLineDash)` feature-detection guard. `setLineDash` now has [97%+ browser support](https://caniuse.com/mdn-api_canvasrenderingcontext2d_setlinedash) and the guard is no longer needed.

**`src/map/Map.js`**
Removes a stale inline comment referencing "browsers without hardware-accelerated transitions or old Android". The code simply assigns `this.options.zoomAnimation` and the comment no longer reflects reality.

## Already done in prior PRs

- `src/layer/vector/Path.js` — `dashArray`/`dashOffset` old-browser note: already removed
- `src/layer/vector/Circle.js` — obsolete instantiation method: already removed

### DIFF
--- a/src/map/Map.js
+++ b/src/map/Map.js
@@ -170,7 +170,6 @@ export class LeafletMap extends Evented {
 
 		this.callInitHooks();
 
-		// don't animate on browsers without hardware-accelerated transitions or old Android
 		this._zoomAnimated = this.options.zoomAnimation;
 
 		// zoom transitions run with the same duration for all layers, so if one of transitionend events
--- a/src/layer/vector/Canvas.js
+++ b/src/layer/vector/Canvas.js
@@ -326,10 +326,8 @@ export class Canvas extends Renderer {
 		}
 
 		if (options.stroke && options.weight !== 0) {
-			if (ctx.setLineDash) {
-				ctx.lineDashOffset = Number(options.dashOffset ?? 0);
-				ctx.setLineDash(options._dashArray ?? []);
-			}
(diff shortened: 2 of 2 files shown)


mourner (approved):

src/map/Map.js
```diff
@@ -170,7 +170,6 @@ export class LeafletMap extends Evented {
 
 		this.callInitHooks();
 
-		// don't animate on browsers without hardware-accelerated transitions or old Android
 		this._zoomAnimated = this.options.zoomAnimation;
```
jonkoops, line 173: Should this have been removed as well then?

Gilmoursa: That comment describes current behavior rather than a browser workaround, so I left it in. It explains the reasoning behind the _createAnimProxy() call: that all layers share the same transition duration, which is why listening for a single transitionend on the map pane is sufficient. Still accurate as far as I can tell.

jonkoops: Ah, so this was a leftover from previous work that had a conditional around `this._zoomAnimated = this.options.zoomAnimation;`?

Gilmoursa: That's my understanding too. The comment looks like it was left behind when the conditional was removed, so it no longer reflected what the code was actually doing.

jonkoops: Thanks for validating. Yes, that does look to be the case; see 3f5ea5a

PR #10245, opened 2026-06-05
Add `viewBox` attribute to logo

bartvanandel:
Fixes aspect ratio when rendering in certain viewers, e.g., VS Code extension `vitaliymaz.vscode-svg-previewer` is affected.

Without `viewBox`, some viewers stretch the image into the bounding box that the viewer provides, instead of fitting the image with its correct aspect ratio.


jonkoops (commented):
Thanks for the fix! This is a good improvement.

While you're at it, could you apply the same treatment to the other SVGs in the repo that are missing `viewBox`? There are a few more:

- `src/images/layers.svg` (26 x 26, no viewBox)
- `docs/docs/images/forum-round.svg` (46 x 46, no viewBox)
- `docs/docs/images/github-round.svg` (46 x 46, no viewBox)
- `docs/docs/images/sprite.svg` (400 x 200, no viewBox)
- `docs/docs/images/x-round.svg` (has viewBox but is missing `height`)

Would be nice to make them all consistent in one go.

bartvanandel:
I'll have a look when I'm back at my laptop this weekend!

bartvanandel:
@jonkoops Done. Found some others as well (not including embedded SVG).

I allowed VSCode to format the files, as there was no consistent formatting applied anyway. From their content, I can see that they were sourced from different places, their code is quite inconsistent. But doesn't cause any problems that I could see, so I didn't modify them any further.

bartvanandel:
NB: somewhat related: #10246

PR #10248, opened 2026-06-10
Fix: disableTextSelection should only affect map container, not whole…

### LINKED ISSUE #9259: L.DomUtil.disableTextSelection should only disable text selection on the map container

adityao3o8:
Fixes #9259

This PR changes `disableTextSelection()` and `enableTextSelection()` to accept a map container element parameter instead of affecting the entire document.

**Before:** Called with no parameters, disabled text selection globally
**After:** Called with a map container element, disables text selection only on that element

This prevents unintended side effects on other page elements during map interactions.

### DIFF
--- a/src/dom/DomUtil.js
+++ b/src/dom/DomUtil.js
@@ -1,6 +1,5 @@
 import * as DomEvent from './DomEvent.js';
 import {Point} from '../geometry/Point.js';
-
 /*
  * @namespace DomUtil
  *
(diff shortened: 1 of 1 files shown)


mourner:
Please see the discussion in the issue.

IvanSanchez:
Besides, https://github.com/Leaflet/Leaflet/pull/10150 should already help to capture/release pointer events properly, so there should be no scenario where text selection misvbehaves.

PR #10249, opened 2026-06-12
fix: replace deprecated navigator.platform in Browser.mac and Browser.linux

ezforhamza:
## What

Replaces the deprecated `navigator.platform` API in `Browser.mac` and `Browser.linux` with `userAgentContains()` calls, which use `navigator.userAgent` — the MDN-recommended replacement.

## Changes

```js
// Before
const mac = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined'
  ? false : navigator.platform.startsWith('Mac');

const linux = typeof navigator === 'undefined' || typeof navigator.platform === 'undefined'
  ? false : navigator.platform.startsWith('Linux');

// After
const mac = userAgentContains('macintosh');
const linux = userAgentContains('linux') && !userAgentContains('android');
```

## Why

`navigator.platform` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform) and browsers are actively removing it. `navigator.userAgent` is the recommended replacement for platform detection.

The `!userAgentContains('android')` guard in `linux` preserves the existing behaviour — Android devices report `linux` in their user-agent, so without it `Browser.linux` would incorrectly return `true` on Android.

## What this does NOT change

`Browser.mobile` — left untouched. As noted in the review of this PR, `screen.orientation` is available on desktop browsers too and cannot be used as a mobile signal. The existing `typeof orientation !== 'undefined'` check is the correct approach for now.

Relates to the deprecation warnings surfaced in #9710.

### DIFF
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -19,7 +19,7 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof orientation !== 'undefined';;
+const mobile = typeof orientation !== 'undefined';
 
 // @property pointer: Boolean
 // `true` for all browsers supporting [pointer events](https://msdn.microsoft.com/en-us/library/dn433244%28v=vs.85%29.aspx).
--- a/src/core/Browser.js
+++ b/src/core/Browser.js
@@ -19,7 +19,12 @@ const chrome = userAgentContains('chrome');
 const safari = !chrome && userAgentContains('safari');
 
 // @property mobile: Boolean; `true` for all browsers running in a mobile device.
-const mobile = typeof orientation !== 'undefined';
+const mobile = typeof window !== 'undefined' && (
+	// Standard Screen Orientation API
+	(typeof window.screen !== 'undefined' && typeof window.screen.orientation !== 'undefined') ||
+	// Deprecated fallback for older browsers
+	typeof window.orientation !== 'undefined'
+);
 
 // @property pointer: Boolean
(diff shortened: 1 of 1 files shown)


IvanSanchez:
@ezforhamza Have you *actually* tested this yourself, on a big PC tower with a big (>24inch) monitor? Is a PC with a big monitor a "mobile"?

IvanSanchez:
See also conversation at https://github.com/Leaflet/Leaflet/pull/9952

ezforhamza:
Thank you for the review and for pointing me to #9952.

You're right — `screen.orientation` is available on desktop browsers too (Chrome, Firefox, Edge all support it), so my change to `Browser.mobile` would incorrectly return `true` on any modern desktop. That breaks the mobile detection entirely.

I'll revert the `mobile` change and narrow this PR down to only what is actually valid: replacing the deprecated `navigator.platform` usage in `Browser.mac` and `Browser.linux` with `userAgentContains()` calls, which is a straightforward fix that doesn't affect mobile detection at all.

Sorry for the oversight — will update shortly.

jonkoops:
`navigator.platform` isn't actually spec-deprecated; the MDN banner is editorial and has been [disputed repeatedly](https://github.com/whatwg/html/issues/11558). Both `navigator.platform` and `navigator.userAgent` are part of Chrome's [UA reduction effort](https://www.chromium.org/updates/ua-reduction/), so swapping one for the other doesn't change much. Longer term we'd rather move toward feature detection instead of reshuffling which detection API we use.

Also, this PR reads like it was entirely AI-generated. Please don't use LLMs to submit drive-by contributions to projects you're not familiar with.

PR #10250, opened 2026-06-12
Implement Browser.reducedMotion and conditionally disable animations

IvanSanchez:
While reviewing #10249 I went back to the deprecation of `Browser.mobile` (#9952) and realized that we're using `.mobile` to disable tile loads during animations, but animations overall shouldn't depend on mobile/desktop or even on GPU performance, but rather in the `prefers-reduced-motion` media query or the (upcoming/experimental) `navigator.preferences.reducedMotion`.

This PR does that, and includes the changes from (and supersedes) #10119.

Manually tested on FFX by setting an `about:config` flag, doesn't seem to break anything.

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -791,19 +791,24 @@ describe('GridLayer', () => {
 			clock.tick(250);
 		});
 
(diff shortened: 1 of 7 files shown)


IvanSanchez:
Hhhhmmmm, the `flyTo()` tests are broken, I gotta check why a bit later.

mourner (approved):

jonkoops (approved):

PR #10094, opened 2026-02-03
Add plugin: Leaflet.CityAutocompleteUSA

slbarriosdev:
Closes #9965

This PR adds **Leaflet.CityAutocompleteUSA** to the official Leaflet plugins list.

- Category: Geocoding
- Demo: https://slbarriosdev.github.io/leaflet-city-autocomplete-usa/demo/
- GitHub: https://github.com/slbarriosdev/leaflet-city-autocomplete-usa
- NPM: https://www.npmjs.com/package/leaflet-city-autocomplete-usa
- License: MIT

PR #10258, opened 2026-06-17
Fix missing front matter delimiter in city-autocomplete-usa.md

slbarriosdev:
The plugin entry was missing the opening `---` delimiter for the YAML front matter, so Jekyll could not parse the metadata fields (name, category, repo, etc.) and the plugin never appeared in the rendered plugins table on leafletjs.com/plugins.html.

This adds the missing `---` on line 1 to match the format used by other plugin entries (e.g. esri-leaflet-geocoder.md).


slbarriosdev:
Friendly ping on this one — it's a single-line fix and all checks are green.

The entry for Leaflet.CityAutocompleteUSA is missing the opening `---` of its YAML front matter, so Jekyll never parses the metadata and the plugin has never shown up in the Geocoding table on leafletjs.com/plugins.html, even though #10094 was merged back in April. You can see the symptom in the Preview tab on main: the whole block renders as plain text.

I've just updated the branch against current main, so this keeps `compatible-v2: false` from #10290 and only adds the missing delimiter on line 1. Happy to close and reopen from a fresh branch if that's easier to review. Thanks!

PR #10274, opened 2026-06-23
Prevent iOS long-press save-image menu on markers

### LINKED ISSUE #6865: iOS 13.1 - contextmenu/longtap issue on Markers

hassams:
## Summary
- Adds `-webkit-touch-callout: none;` to the existing rule on `.leaflet-tile`, `.leaflet-marker-icon`, `.leaflet-marker-shadow` that already disables `user-select` and `-webkit-user-drag`.
- This stops iOS Safari from showing the "Save Image" / "Copy" context menu when long-pressing a marker icon, matching the behavior already in place for long-press on the map itself.

Fixes #6865.

A previous attempt (#7042) bundled this same fix with an unrelated, more controversial change (`user-select: none` on `.leaflet-control:first-child`) and was discarded because of that second part. This PR is scoped to only the `-webkit-touch-callout` fix.

## Test plan
- [x] `npm run lint` passes
- [x] `npm run build` succeeds
- [x] `npm test -- --run` — all 1082 tests pass, no regressions (CSS-only change)
- [x] Verified the rule change matches the repro steps in #6865 (long-press on a marker no longer offers to save the icon image on iOS Safari)


jonkoops (commented):
This also disables long presses on links. Are we sure that these classes are not applied to anything a user might want to long-press?

hassams:
> This also disables long presses on links. Are we sure that these classes are not applied to anything a user might want to long-press?

Good catch @jonkoops, thank you!

`-webkit-touch-callout` is an inherited property, so setting it on `.leaflet-marker-icon` would cascade into child elements — meaning any `<a>` tag inside a DivIcon would also lose its long-press "Open Link / Copy Link" callout on iOS, which is undesirable.

The "Save Image" callout only fires on `` elements. Since the default icon is an `` but DivIcon is a `<div>`, we can scope the rule to target only `` markers:

```css
.leaflet-tile,
img.leaflet-marker-icon,
.leaflet-marker-shadow {
    user-select: none;
    -webkit-user-drag: none;
    -webkit-touch-callout: none;
}
```

This leaves `div.leaflet-marker-icon` (DivIcon) untouched, so links inside DivIcons keep their native iOS callout behaviour. I'll push an update with this change.

**Manual test steps (iOS Safari — real device or Xcode Simulator):**

1. Serve the repo locally (`npm run debug`) and open `debug/map/markers.html` in iOS Safari.
2. **Long-press a map tile** → no "Save Image" menu should appear.
3. **Long-press the default marker icon (blue pin)** → no "Save Image" menu should appear.
4. **Long-press the link inside the DivIcon** → the native iOS "Open Link / Copy Link" callout **should** still appear.

Steps 2–3 confirm the original fix for #6865. Step 4 confirms the regression raised here does not occur.

jonkoops (approved):
The existing rule already applies `user-select: none` to `.leaflet-marker-icon` (including DivIcons), so adding `-webkit-touch-callout: none` to the same selector is consistent. No need to scope it to `img.leaflet-marker-icon`. LGTM as-is.

PR #10276, opened 2026-06-25
Fix LatLng.validate throwing on null instead of returning false

chatman-media:
`LatLng.validate(null)` throws `TypeError: Cannot use 'in' operator to search for 'lat' in null` instead of returning `false`.

`typeof null === 'object'` is `true`, so the object branch reaches `'lat' in null` and throws. This is asymmetric with `validate(undefined)`, which correctly returns `false`. The fix guards the object check with `lat !== null`.

As a side effect, `new LatLng(null)` now throws the intended descriptive `Invalid LatLng object` error (from the constructor's `validate` call) instead of the opaque internal `TypeError`.

Added a `validate` spec covering valid inputs and invalid inputs (`undefined`/`null`/`'foo'`/short array) returning `false` without throwing. The `null` case fails without this fix.

### DIFF
--- a/spec/suites/geo/LatLngSpec.js
+++ b/spec/suites/geo/LatLngSpec.js
@@ -33,6 +33,22 @@ describe('LatLng', () => {
 		});
 	});
 
+	describe('validate', () => {
+		it('returns true for valid inputs', () => {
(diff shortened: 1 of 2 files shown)


jonkoops (approved):
Good catch! LGTM

PR #10279, opened 2026-06-27
My leaflet plugin

Spoogle-search:
This document provides an overview of the Poligonostic Leaflet Radar plugin, including installation instructions and a demo link. All true!


Spoogle-search (commented):
This is a good, complete plugin. I just need you to see if it is okay.

PR #10333, opened 2026-08-07
Wrap non-square tile columns by the tile width, not the tile height

### LINKED ISSUE #9326: Tall custom sized tiles are not wrapped correctly
### Checklist

### Steps to reproduce

I'm trying to get custom sized tiles from http://www.ign.es/wms-inspire/ign-base and display them in our app, with wrap option enabled. 

### Expected behavior

All tiles should be displayed properly forming whole map instead of repeated fragments 
map2

### Current behavior

There is no problem when tiles width is greater than height (for example 2048 width and 1024 height), but with the opposite configuration tiles are displayed incorrectly.
map

After some debugging session I recognized that there is a problem with wrapping X coords of tiles. CRS that we are using there have wrapLng (-180, 180) so we are setting _wrapX in GridLayer.js:
```
this._wrapX = crs.wrapLng && !this.options.noWrap && [
			Math.floor(map.project([0, crs.wrapLng[0]], tileZoom).x / tileSize.x),
			Math.ceil(map.project([0, crs.wrapLng[1]], tileZoom).x / tileSize.y)
		];

this._wrapY = crs.wrapLat && !this.options.noWrap && [
			Math.floor(map.project([crs.wrapLat[0], 0], tileZoom).y / tileSize.y),
			Math.ceil(map.project([crs.wrapLat[1], 0], tileZoom).y / tileSize.x)
		];
```
I don't understand why it's using y value of tileSize to calculate max range of X, shouldn't it use only tileSize.x for _wrapX, and tileSize.y for _wrapY? 
After changing it locally to:
```
this._wrapX = crs.wrapLng && !this.options.noWrap && [
			Math.floor(map.project([0, crs.wrapLng[0]], tileZoom).x / tileSize.x),
			Math.ceil(map.project([0, crs.wrapLng[1]], tileZoom).x / tileSize.x)
		];
this._wrapY = crs.wrapLat && !this.options.noWrap && [
			Math.floor(map.project([crs.wrapLat[0], 0], tileZoom).y / tileSize.y),
			Math.ceil(map.project([crs.wrapLat[1], 0], tileZoom).y / tileSize.y)
		];
```
everything is working properly.

_No response_

### Environment

- Leaflet version: 1.9.4
- Browser (with version): Chrome 123.0.6312.123, Firefox 124.0.2 (64 bits), Edge 123.0.2420.97
- OS/Platform (with version): Windows 11 laptop

Jaybhade:
Fixes #9326.

`_resetGrid()` turns the projected world extent into a tile range so `_wrapCoords()` can wrap tile coordinates. The upper bound of `_wrapX` divides an **x** pixel coordinate by `tileSize.y`, and the lower bound of `_wrapY` divides a **y** pixel coordinate by `tileSize.x`:

```js
this._wrapX = crs.wrapLng && !this.options.noWrap && [
	Math.floor(map.project([0, crs.wrapLng[0]], tileZoom).x / tileSize.x),
	Math.ceil(map.project([0, crs.wrapLng[1]], tileZoom).x / tileSize.y)   // <- tile height
];
```

Square tiles hide it, which is why it went unnoticed. With `tileSize: [256, 512]` at zoom 1 the world is 512px wide — two columns — but `_wrapX` comes out as `[0, 1]`, so `wrapNum()` collapses every column to `x = 0` and the map repeats the westernmost tile column across the whole viewport, as in the screenshots in #9326.

Both lines had their axis swapped in cf51113, the commit that introduced non-square tiles (#3570): the scalar `tileSize` was expanded to `tileSize.x`/`tileSize.y` and the four call sites ended up as `.x`, `.y`, `.x`, `.y` instead of `.x`, `.x`, `.y`, `.y`. This restores what the scalar version computed.

Reproduced with a `GridLayer` whose tiles paint their own coords, 800x400 container, `tileSize: [256, 512]`, zoom 1:

| | `_wrapX` | tile columns requested |
| --- | --- | --- |
| `main` | `[0, 1]` | `0` |
| this branch | `[0, 2]` | `0, 1` |

The added spec fails on `main` with `expected [ '0:0:1', '0:0:1' ] to deeply equal [ '0:0:1', '1:0:1' ]`.

`_wrapY` is only reachable through a custom CRS, since none of the built-in ones define `wrapLat`, so the spec covers `_wrapX` only — the `_wrapY` line is the same typo and now converts a y coordinate with the y tile size.

Full suite passes locally on `chromium`, `chromium-retina` and `webkit` (1085 passed, 14 skipped); `firefox` wouldn't launch on my machine for unrelated reasons.

### DIFF
--- a/spec/suites/layer/tile/GridLayerSpec.js
+++ b/spec/suites/layer/tile/GridLayerSpec.js
@@ -1067,6 +1067,31 @@ describe('GridLayer', () => {
 				done();
 			});
 		});
+
+		it('When false, wraps the columns of non-square tiles by the tile width', (done) => {
+			container.style.width = '512px';
+			container.style.height = '512px';
+
+			const grid = new GridLayer({
+				attribution: 'Grid Layer',
+				tileSize: new Point(256, 512),
+				noWrap: false
+			});
+			const loadedTileKeys = [];
+
+			grid.createTile = function (coords) {
+				loadedTileKeys.push(`${coords.x}:${coords.y}:${coords.z}`);
+				return document.createElement('div');
+			};
+
+			map.addLayer(grid).setView([0, 0], 1);
+
+			grid.on('load', () => {
+				// At zoom 1 the world is 512px wide, i.e. two columns of 256px tiles
+				expect(loadedTileKeys).to.eql(['0:0:1', '1:0:1']);
+				done();
+			});
+		});
 	});
 
 	describe('Sanity checks for infinity', () => {
--- a/src/layer/tile/GridLayer.js
+++ b/src/layer/tile/GridLayer.js
@@ -608,10 +608,10 @@ export class GridLayer extends Layer {
 
 		this._wrapX = crs.wrapLng && !this.options.noWrap && [
 			Math.floor(map.project([0, crs.wrapLng[0]], tileZoom).x / tileSize.x),
-			Math.ceil(map.project([0, crs.wrapLng[1]], tileZoom).x / tileSize.y)
+			Math.ceil(map.project([0, crs.wrapLng[1]], tileZoom).x / tileSize.x)
 		];
 		this._wrapY = crs.wrapLat && !this.options.noWrap && [
-			Math.floor(map.project([crs.wrapLat[0], 0], tileZoom).y / tileSize.x),
+			Math.floor(map.project([crs.wrapLat[0], 0], tileZoom).y / tileSize.y),
 			Math.ceil(map.project([crs.wrapLat[1], 0], tileZoom).y / tileSize.y)
 		];
 	}


fallenmi (approved):
I reproduced the non-square wrapping defect against exact base `c96f31a7`: with 256×512 tiles at zoom 1, both requested columns collapse to `x=0`. Exact head `b05f2eb9` restores the expected `x=0,1` sequence, and the axis-specific divisors also match the corresponding projected coordinates for vertical wrapping.

The focused `GridLayer` file passed 38/38 tests, and the full Chromium suite passed 1,086 tests with 14 skips. Lint completed with only the nine unchanged CSS `!important` warnings, and the diff check is clean.

Reviewed with OpenAI Codex assistance; I independently ran the exact base/head oracle and full exact-head suite.

PR #10372, opened 2026-09-09
Preserve marker transform origin

### LINKED ISSUE #3800: Transform-origin on .leaflet-zoom-animated breaks TransformMarker
Since Leaflet doesn't support rotation of markers (yet), I am using @jacobtoye's [transformmarker](https://github.com/jacobtoye/Leaflet.transformmarker). This worked fine in 0.7.3, but broke in 1.0.0 beta 1: markers aren't being rotated around their center point* anymore.

After a bit of digging, I found that 1.0.0 sets a `transform-origin` property of `0 0` on the `leaflet-zoom-animated` class. This causes markers to be rotated around their top-left corner rather than the correct point.

The fix for this is simple. However, since the transform marker is not an official Leaflet product, I am not sure whether this should be "fixed" in Leaflet itself, or in transformmarker, on on a per-project basis. Can a Leaflet-core contributor please advise?

---

\* center point in my case, but in reality this will depend on the marker's `iconAnchor` point.

deepujain:
Fixes #3800

The broad `.leaflet-zoom-animated` rule forces `transform-origin: 0 0` on marker icons. Marker rotation plugins therefore rotate around the top-left corner instead of their configured origin.

Exclude marker icons from that transform-origin rule while retaining it for the other zoom-animated elements that need it. A browser regression verifies that zoom-animated markers no longer receive the forced origin.

Validation:

- `npx vitest --run --project=chromium spec/suites/layer/marker/MarkerSpec.js`
- `npx vitest --run --project=chromium` (1085 passed, 14 skipped)
- `npm run lint`
- `npm run build`
- `node ./spec/ssr/ssr_node.js`

### DIFF
--- a/spec/suites/layer/marker/MarkerSpec.js
+++ b/spec/suites/layer/marker/MarkerSpec.js
@@ -26,6 +26,13 @@ describe('Marker', () => {
 		removeMapContainer(map, container);
 	});
 
+	it('does not override the transform origin of zoom-animated markers', () => {
+		const marker = new Marker([0, 0]).addTo(map);
+
+		expect(marker._icon.classList.contains('leaflet-zoom-animated')).to.be.true;
+		expect(getComputedStyle(marker._icon).transformOrigin).to.not.equal('0px 0px');
+	});
+
 	describe('#setIcon', () => {
 
 		it('set the correct x and y size attributes', () => {
--- a/src/leaflet.css
+++ b/src/leaflet.css
@@ -207,7 +207,7 @@
 	}
 }
 
-.leaflet-zoom-animated {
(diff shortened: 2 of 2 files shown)


lbesecker195 (approved):
Verified that Leaflet itself never scales or rotates marker icons: `Marker._setPos` and `_animateZoom` both go through `DomUtil.setPosition`, which emits only `translate3d`, so `transform-origin` on `.leaflet-marker-icon` has no effect on core behavior and only matters to rotation plugins, which is exactly what #3800 asked for. The `:not()` exclusion is narrowly scoped and the spec test pins it. LGTM.

